PR #33158, opened 2025-03-12
New workflow for sdk breaking change labeling

raych1:
### Workflow and Event Handling Enhancements:
* [`.github/src/context.js`]: Added support for the `check_run` event and removed the requirement to find an 'issue-number' artifact for certain events. 
* [`.github/workflows/update-labels.yaml`]: Included the "SDK Breaking Change Labels (Preview)" workflow in the list of upstream workflows.

### SDK Label Management:
* [`.github/src/sdk-types.js`]: Defined `SdkName` types and mapped SDK labels for breaking change labels.
* [`.github/workflows/sdk-breaking-change-labels.yaml`]: Created a new workflow for processing SDK breaking change labels.
* [`.github/workflows/src/sdk-breaking-change-labels.js`]: Implemented the logic to fetch check run data and determine label actions based on artifacts.

### SDK Generation Improvements:
* [`eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml`]: Updated the SDK generation pipeline to publish breaking change label artifacts conditionally. 
* [`eng/tools/spec-gen-sdk-runner/src/commands.ts`]: Enhanced logging and artifact processing during SDK generation, and added handling for breaking change label artifacts.


raych1:
@mikeharder here are the test runs for `update-label` actions. I will check if test coverage for 'context.js' needs to be extended.
[Test run without issue_number](https://github.com/raych1/azure-rest-api-specs/actions/runs/13801334667/job/38604178253)
image

[Test run with issue_number](https://github.com/raych1/azure-rest-api-specs/actions/runs/13801448838/job/38604497653)
image

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
+name: SDK Breaking Change Labels (Preview)
+
+on:
+  check_run:
```
mikeharder, line 4: Do you need to trigger on check_run:completed, or would check_suite:completed be sufficient?

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      check_run_id:
+        description: 'Check run ID to process'
+        required: false
+        default: '38335408598'
+        type: string
+
```
mikeharder, line 13:
```suggestion
permissions:
  contents: read
```

Start with these minimal permissions, and only add more if needed.

.github/src/context.js
```diff
@@ -96,7 +96,7 @@ export async function extractInputs(github, context, core) {
         context.payload
       );
 
-    let issue_number;
+    let issue_number = NaN;
```
mikeharder, line 99:
```suggestion
    let issue_number;
```

Better to require each "if" block to set this explicitly to prevent errors.

.github/src/context.js
```diff
@@ -191,12 +192,6 @@ export async function extractInputs(github, context, core) {
           }
         }
       }
```
mikeharder, line 194:
```suggestion
      }
      if (!issue_number) {
        core.info(`Could not find 'issue-number' artifact, which is required to associate the triggering workflow run with a PR`);
        issue_number = NaN;
      }
```

.github/src/sdk-types.js
mikeharder, line 1: Is there a benefit, either currently or planned, of defining these types in a separate file under .github/src?  If not, please move them into the existing JS file for the workflow.

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  process-check-run:
```
mikeharder, line 15:
```suggestion
  sdk-breaking-change-labels:
```

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: SDK Breaking Change Labels
```
mikeharder, line 16:
```suggestion
    name: SDK Breaking Change Labels (Preview)
```

.github/workflows/src/update-labels.js
```diff
@@ -20,6 +20,12 @@ export default async function updateLabels({ github, context, core }) {
     run_id = run_id || inputs.run_id;
   }
 
+  if (Number.isNaN(issue_number) || Number.isNaN(run_id)) {
+    core.info(`skipping run because issue_number or run_id is not a valid number. issue_numer: ${issue_number}, run_id: ${run_id}`);
+    return;
+  }
+  const pullRequestUrl = `https://github.com/${owner}/${repo}/pull/${issue_number}`;
+  core.info(`pull request url: ${pullRequestUrl}`);
```
mikeharder, line 28:
```suggestion
```

I think all this code can be deleted.  The decision on how to handle undefinded issue_number should be deferred as late as possible, until we know if it's necessary or not.

mikeharder:
@raych1: Do you have an end-to-end examples of all the pipelines and workflows running together?  I'd like to trace how the information flows from devops to github.

mikeharder (changes requested):
See comments

.github/workflows/src/update-labels.js
raych1: Why not return as early as possible?

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  check_run:
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      check_run_id:
+        description: 'Check run ID to process'
+        required: false
+        default: '38335408598'
```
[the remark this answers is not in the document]
raych1: removed

.github/workflows/sdk-breaking-change-labels.yaml
raych1: As there are a few check runs in the 'Azure Pipeline' suite, I don't want the triggering needs to wait until all the check runs complete.

raych1:
> @raych1: Do you have an end-to-end examples of all the pipelines and workflows running together? I'd like to trace how the information flows from devops to github.

@mikeharder these are two PRs for the end to end testing.
[PR from forked repo](https://github.com/raych1/azure-rest-api-specs/pull/7)
[PR from same repo](https://github.com/raych1/azure-rest-api-specs/pull/11)

.github/workflows/src/update-labels.js
raych1: Reported error in the later steps.

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: SDK Breaking Change Labels
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            .github
+      
+      - name: Setup Node and install deps
+        uses: ./.github/actions/setup-node-install-deps
```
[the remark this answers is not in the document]
raych1: I thought 'node settings' need it. I removed this step.

.github/workflows/src/update-labels.js
```diff
@@ -103,6 +97,13 @@ export async function updateLabelsImpl({
   core.info(`labelsToAdd: ${JSON.stringify(labelsToAdd)}`);
   core.info(`labelsToRemove: ${JSON.stringify(labelsToRemove)}`);
 
+  if ((labelsToAdd.length > 0 || labelsToRemove.length > 0) && Number.isNaN(issue_number)) {
```
mikeharder, line 100:
```suggestion
  if ((labelsToAdd.length > 0 || labelsToRemove.length > 0) && !issue_number) {
```

Convention in JS is to use `!foo` to determine if a value is "falsy".  This includes things like `""`, `NaN`, `0`, `null`, and `undefined`.  Typically, you want to treat these all like "false", rather than specifically checking for `NaN`.

.github/src/sdk-types.js
```diff
@@ -0,0 +1,50 @@
...
+ * @typedef {'azure-sdk-for-go' | 'azure-sdk-for-java' | 'azure-sdk-for-js' | 'azure-sdk-for-net' | 'azure-sdk-for-python'} SdkName
+ */
+
+/**
+ * SDK labels mapping for breaking change labels
+ * @type {Object.<SdkName, {
+ *   breakingChange: string | undefined,
+ *   breakingChangeApproved: string | undefined,
+ *   breakingChangeSuppression: string | undefined,
+ *   breakingChangeSuppressionApproved: string | undefined
```
mikeharder, line 13:
```suggestion
 *   breakingChange: string,
 *   breakingChangeApproved: string,
 *   breakingChangeSuppression: string,
 *   breakingChangeSuppressionApproved: string
```

Do these need to support undefined?  All keys appear to be defined in all objects below.

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,57 @@
...
+    name: SDK Breaking Change Labels (Preview)
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            .github
+      
+      - name: Setup Node and install deps
+        uses: ./.github/actions/setup-node-install-deps
```
mikeharder, line 27:
```suggestion
```

You left a comment that you removed this, but it's still here.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+    } else {
+      core.info(`No open PRs found for commit ${checkRun.head_sha}`);
+    }
+  } catch (error) {
+    core.error(`${wfName}: Error searching for PRs with commit ${checkRun.head_sha}: ${error.message}`);
+    throw error;
+  }
+
+  return { issueNumber };
+}
```
mikeharder, line 186: Add trailing newline.  Your IDE might be removing this.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments
+ * @returns {Promise<{labelName: string, labelAction: LabelAction, issueNumber: number, state: string}>}
+ */
+export async function getLabelAndAction({ github, context, core }) {
+  let issue_number = NaN;
+  let check_run_id = process.env.CHECK_RUN_ID || context.payload.check_run?.id;  
+  let labelAction;
+  let labelName = "";
+
```
mikeharder, line 16: Please use the pattern defined here for extracting inputs:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff-preview.js

Top-level function called by the workflow YAML, that extracts inputs and calls a helper method to do the actual work.  This makes our workflows more consistent, and allows you to test the business logic separately from the logic to extract the inputs.

Thus, I think a lot of the code below should be moved into context.js.  Any workflow that needs to trigger on a DevOps pipeline and extract inputs from it will need similar code, right?

.github/workflows/src/sdk-breaking-change-labels.js
mikeharder, line 1: Add unit tests to achieve 100% code coverage.  If a particular block of code is hard to test, I might be able to help, or we can suppress that bit of code from coverage if needed.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+    throw new Error('No check run ID provided');
+  }
+  core.info(`Fetching check run with ID: ${check_run_id}`);
+  
+  // First fetch the check run with specific ID
+  const checkRunData = await github.rest.checks.get({
+    owner: context.repo.owner,
+    repo: context.repo.repo,
+    check_run_id: parseInt(check_run_id, 10)
+  });
```
mikeharder, line 28:
```suggestion
  const checkRun = await github.rest.checks.get({
    owner: context.repo.owner,
    repo: context.repo.repo,
    check_run_id: parseInt(check_run_id, 10)
  }).data;
```

Simplify

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+  
+  // Use the check run data we retrieved instead of from the payload
+  const checkRun = checkRunData.data;
+  if (!checkRun) {
+    throw new Error('No check run data found in the API response');
+  }
+  core.info(`Check run name: ${checkRun.name}`);
+  if (!checkRun.name.includes("(SDK Generation Job)")) {
```
mikeharder, line 36: Is this available in the payload (and could be filtered in yaml instead of JS)?

.github/workflows/sdk-breaking-change-labels.yaml
```diff
@@ -0,0 +1,57 @@
...
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  sdk-breaking-change-labels:
+    name: SDK Breaking Change Labels (Preview)
+    runs-on: ubuntu-24.04
```
mikeharder, line 19:
```suggestion
    if: contains(github.event.check_run.name, '(SDK Generation Job)') && github.event.check_run.check_suite.app.name == 'Azure Pipelines`
    runs-on: ubuntu-24.04
```

Filter as aggressively as possible based on the properties of `github.event` before allocating an agent.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+        core.info(`Artifacts URL: ${artifactsUrl}`);
+        
+        // Call Azure DevOps REST API to list artifacts
+        try {          
+          // API URL to list artifacts
+          const apiUrl = `https://dev.azure.com/${organization}/${project}/_apis/build/builds/${buildId}/artifacts?api-version=7.0`;              
+          core.info('Calling Azure DevOps API to list artifacts...');
+          
+          // Use Node.js fetch to call the API
+          const response = await fetch(apiUrl, {
```
mikeharder, line 70: Does this require authentication?  Have you tested it for private pipelines (repo specs-pr) as well as public?

mikeharder, line 70: Does this require authentication?  Have you tested it for private pipelines (repo specs-pr) as well as public?

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * Retrieves the PR number associated with a specific commit SHA
+ * @param {Object} params
+ * @param {import('github').CheckRun} params.checkRun - The check run object containing head_sha
+ * @param {typeof import("@actions/core")} params.core - GitHub Actions core for logging
+ * @param {import('github-script').GitHub} params.github - GitHub API client
+ * @returns {Promise<{issueNumber: number}>} - The PR number or NaN if not found
+ */
+async function getIssueNumber({checkRun, core, github})
```
mikeharder, line 148: I think this should definitely be a shared helper.

mikeharder (changes requested):
see comments

.github/workflows/sdk-breaking-change-labels.yaml
weshaggard: Yes if we can do that level of filtering we should.

.github/workflows/src/sdk-breaking-change-labels.js
weshaggard: It will need authentication for the private repo.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+      // Extract organization and project from URL
+      const urlParts = new URL(checkRun.details_url);
+      const pathSegments = urlParts.pathname.split('/').filter(Boolean);
+      
+      if (pathSegments.length >= 2) {
+        const organization = pathSegments[0]; // e.g., azure-sdk
+        const project = pathSegments[1]; // e.g., fadad8ea-1e34-4d7a-9016-fd29b14e8b94
+
+        // Construct artifacts URL
+        const artifactsUrl = `https://dev.azure.com/${organization}/${project}/_build/results?buildId=${buildId}&view=artifacts`;
```
weshaggard, line 60: As we talked let's try to split on `_build` and take the first part to use in the other calls.

.github/workflows/src/sdk-breaking-change-labels.js
```diff
@@ -0,0 +1,186 @@
...
+        core.info(`Artifacts URL: ${artifactsUrl}`);
+        
+        // Call Azure DevOps REST API to list artifacts
+        try {          
+          // API URL to list artifacts
+          const apiUrl = `https://dev.azure.com/${organization}/${project}/_apis/build/builds/${buildId}/artifacts?api-version=7.0`;
```
weshaggard, line 66: As we chatted if we can avoid listing the artifacts and can pull one by name we should do that.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -195,15 +195,30 @@ extends:
...
+                $artifactPath = "$(System.DefaultWorkingDirectory)/$(AddBreakingChangeLabelArtifactPath)"
+                Write-Host "Artifact Name: $artifactName"
+                Write-Host "Artifact Path: $artifactPath"
+                if (Test-Path $artifactPath) {
+                  Get-ChildItem $artifactPath -Recurse
+                }
+
+          - template: /eng/common/pipelines/templates/steps/publish-1es-artifact.yml
+            parameters:
+              ArtifactName: $(AddBreakingChangeLabelArtifactName)
```
weshaggard, line 213: As we chatted lets only have one artifact that has a well-known name and not have the possibility of having multiple which can be confusing and require more downloading than needed.

.github/workflows/sdk-breaking-change-labels.yaml
raych1: Updated.

.github/workflows/src/sdk-breaking-change-labels.js
raych1: Added tests with 100% coverage.

.github/workflows/src/sdk-breaking-change-labels.js
raych1: Made some input properties as optional in context.js and added common logic to there for check_run case.

.github/workflows/src/sdk-breaking-change-labels.js
raych1: Added.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: Updated to use well-known artifact name then read the content of it to determine the language and label action.

.github/workflows/src/sdk-breaking-change-labels.js
raych1: Updated to download specific artifact.

raych1:
@mikeharder all the comments were addressed, and could you have another review?

.github/test/mocks.js
```diff
@@ -12,3 +12,4 @@ export function createMockLogger() {
     isDebug: vi.fn().mockReturnValue(false),
   };
 }
+
```
mikeharder, line 15:
```suggestion
```

mikeharder:
@raych1: One issue I see early is your changes don't pass the new linting tools I added.

mikeharder (changes requested):
Please fix linting issues first:

https://github.com/Azure/azure-rest-api-specs/actions/runs/13929427110/job/38982515798?pr=33158#step:4:16

raych1:
@mikeharder latest test run:
[PR from same repo](https://github.com/raych1/azure-rest-api-specs/pull/11#issuecomment-2734978531)
[PR from forked repo](https://github.com/raych1/azure-rest-api-specs/pull/7#issuecomment-2734978417)

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "azure-rest-api-specs",
   "devDependencies": {
+    "@autorest/openapi-to-typespec": "0.10.12",
```
mikeharder, line 4:
```suggestion
    "@autorest/openapi-to-typespec": "0.10.13",
```

good to sort this correctly, but delete the dupplicate below

.github/workflows/src/context.js
```diff
@@ -210,6 +210,36 @@ export async function extractInputs(github, context, core) {
       issue_number: issue_number,
       run_id: payload.workflow_run.id,
     };
+  } else if (context.eventName === "check_run") {
```
mikeharder, line 213:
```suggestion
  } else if (context.eventName === "check_run" && context.payload.action === "completed") {
```

I'm not sure this is the exact property name, but I think this should be scoped to only "completed" events, until we need otherwise.

.github/workflows/src/context.js
```diff
@@ -210,6 +210,36 @@ export async function extractInputs(github, context, core) {
...
+    let checkRun = context.payload.check_run;
+
+    // Extract the ADO build ID and project URL from the check run details URL
```
mikeharder, line 216: What if the check_run didn't come from ADO?  I think we should handle this gracefully rather than fail.

.github/workflows/src/issues.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Retrieves the PR number associated with a specific commit SHA
+ * @param {Object} params
+ * @param {String} params.head_sha - The head_sha
+ * @param {typeof import("@actions/core")} params.core - GitHub Actions core for logging
+ * @param {import("@octokit/core").Octokit & import("@octokit/plugin-rest-endpoint-methods/dist-types/types.js").Api} params.github - GitHub API client
+ * @returns {Promise<{issueNumber: number}>} - The PR number or NaN if not found
+ */
+export async function getIssueNumber({ head_sha, core, github }) {
```
mikeharder, line 9: Could this be moved into context.js, and returned automatically as the value of "output.issue_number" instead of NaN?

.github/workflows/test/issues.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+const mockGithub = createMockGithub();
+const mockCore = createMockCore();
+
+describe("getIssueNumber", () => {
+  beforeEach(() => {
+    // Reset mocks
+    vi.clearAllMocks();
+  });
+  it("should return NaN when head_sha is missing", async () => {
```
mikeharder, line 13: Should this throw instead of returning NaN?  Is it ever valid to call without a head_sha?

.github/workflows/test/issues.test.js
```diff
@@ -0,0 +1,98 @@
...
+      core: mockCore,
+    });
+
+    // Verify result
+    expect(result.issueNumber).toBeNaN();
+
+    // Verify appropriate message was logged
+    expect(mockCore.info).toHaveBeenCalledWith(
+      "No head_sha found in check run",
+    );
```
mikeharder, line 27: I'd remove expectations about logging, it will be more work to update if we change the logs.  Logs should be flexible.

.github/workflows/test/issues.test.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    // Call function
+    const result = await getIssueNumber({
+      head_sha: "abc",
+      github: mockGithub,
+      core: mockCore,
+    });
+
+    // Verify result uses first PR
```
mikeharder, line 49: Why correct to use first PR?

mikeharder (approved):
Approved pending fix to package.json

package.json
raych1: updated.

.github/workflows/src/context.js
raych1: At this point, we only have the check_run from ADO. We could enhance it when there are other run types.

.github/workflows/src/issues.js
raych1: As talked offline, the reason is that we want to save the call of searching issues if there are no label action to make.

.github/workflows/src/context.js
raych1: Leave it to handle by each individual workflow template.

.github/workflows/test/issues.test.js
raych1: sounds good.

.github/workflows/test/issues.test.js
raych1: Multiple PRs happens when users create multiple PRs based on the same commit sha. It's an uncommon user case and usually users should keep only one of them open. For now, just return the first PR and log warnings.  We could revisit this logic when there are more scenarios using this API in the future.

PR #33208, opened 2025-03-14
add @@clientName for iotoperations to avoid JS SDK breaking

kazrael2119:


MaryGao (approved):
Approved and this is to mitigate the enum name breaking in JS. Should have no impact for REST or other emitters.

PR #33244, opened 2025-03-17
locked spec-gen-sdk version

raych1:


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -42,7 +42,7 @@ extends:
           - name: PythonVersion
             value: '3.13'
           - name: SpecGenSdkVersion
-            value: 'latest'
+            value: '0.2.2'
```
mikeharder, line 45: Proper fix is to use a package-lock.json to pin the package and all its transitive deps.  Can we do this instead?

mikeharder (changes requested):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: @mikeharder I added `spec-gen-sdk` to dev dependencies of root 'package.json'.

mikeharder (approved):
Looks good for this repo, but you may also want to fix how this gets installed in eng/common.

https://github.com/Azure/azure-sdk-tools/blob/0c82e8b23a833286bed8975fe79903a5c19c9357/eng/common/pipelines/templates/stages/archetype-spec-gen-sdk.yml#L43-L44

raych1:
> Looks good for this repo, but you may also want to fix how this gets installed in eng/common.
> 
> https://github.com/Azure/azure-sdk-tools/blob/0c82e8b23a833286bed8975fe79903a5c19c9357/eng/common/pipelines/templates/stages/archetype-spec-gen-sdk.yml#L43-L44

This template under eng/common was deprecated and will be deleted later.

mikeharder:
> > Looks good for this repo, but you may also want to fix how this gets installed in eng/common.
> > https://github.com/Azure/azure-sdk-tools/blob/0c82e8b23a833286bed8975fe79903a5c19c9357/eng/common/pipelines/templates/stages/archetype-spec-gen-sdk.yml#L43-L44
> 
> This template under eng/common was deprecated and will be deleted later.

@raych1: Is tool spec-gen-sdk only used in repo azure-rest-api-specs now?  If so, can we move the source code from tools directly into this repo, which would simplify things?

PR #33257, opened 2025-03-17
[.github] Add Windows to test matrix

mikeharder:
- Most actions only use Linux at runtime, but dev on Windows needs to be supported

More complete fix is planned: https://github.com/Azure/azure-rest-api-specs/issues/33193

### DIFF
--- a/.github/workflows/github-test.yaml
+++ b/.github/workflows/github-test.yaml
@@ -21,7 +21,11 @@ defaults:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        os: [ubuntu, windows]
+
+    runs-on: ${{ fromJSON('{"ubuntu":"ubuntu-24.04", "windows":"windows-2022"}')[matrix.os] }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/github-test.yaml
```diff
@@ -16,12 +16,15 @@ permissions:
 
 defaults:
   run:
-    shell: bash
```
mikeharder, line 19: I think this was always superfluous, but especially if we are testing on windows

.github/workflows/github-test.yaml
```diff
@@ -16,12 +16,15 @@ permissions:
...
     working-directory: ./.github
 
 jobs:
   test:
-    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        os: [ubuntu, windows]
+
+    runs-on: ${{ fromJSON('{"ubuntu":"ubuntu-24.04", "windows":"windows-2022"}')[matrix.os] }}
```
benbp, line 27: I always wanted dynamic object literals and hash lookups in devops templating. Useful that you can do this here.

benbp (approved):

PR #33258, opened 2025-03-17
[.github] Use TypeScript for type-checking of JS sources

### LINKED ISSUE #33250: [EngSys] Use TypeScript to verify type annotations for pure JS code
## Config
```
{
  "compilerOptions": {
    "allowJs": true,    // Allow compiling JavaScript files.
    "checkJs": true,    // Enable type checking on JavaScript files.
    "noEmit": true,     // Do not emit any output files.
    "strict": true      // Enable all strict type-checking options.
  },
  "include": [
    "./src/**/*.js"     // Adjust the glob to match your JavaScript source files.
  ]
}
```

## Run
```tsc```

### PR DESCRIPTION
**Pull request by mikeharder**
- Fixes #33250

### CODE REVIEW

File: .github/workflows/github-test.yaml [hunk 1]
```diff
@@ -37,6 +37,9 @@ jobs:
           node-version: 20.x
           working-directory: ./.github
 
+      - name: Check types
+        run: npx tsc && echo "Type checking completed successfully!"
+
       - name: Run tests
         run: npm run test:ci
```
benbp, line 44: Would it be better to embed this into an npm run command so everything can still be executed locally + reduce config growth?

mikeharder: Good call.  I think I'll just use the standard script name "lint", which we already use under eng/tools:

https://github.com/Azure/azure-rest-api-specs/blob/752a5c0d1f303bf950ccef3f8840f601a45ef6f5/.github/workflows/_reusable-eng-tools-test.yaml#L61

We could potentially refactor this even further, to use the same YAML for both eng/tools and .github with options, but they might be too different to be worth it.

benbp: Looks good to me. We could also consider just `npm ci` which runs `test:ci and lint` under the hood. Thought it might be nice to have lint as its own step status.

benbp (approved):

.github/workflows/github-test.yaml
mikeharder: `npm ci` is already a reserved command.  You're correct we could have a meta-command with a different name to run both "test:ci" and "lint", but I think the convention in most JS projects is to **not** do this, and instead have a few top-level commands like "build", "lint", "test".

PR #33277, opened 2025-03-18
Add asset-emitter to dependencies to avoid deduplicaiton to old compiler

witemple-msft:


mikeharder:
@witemple-msft: You will probably want to add this dep to main as well.  I think we try to list all TSP package deps, even if not strictly required since they'd be pulled in as transitive deps.

weshaggard:
@witemple-msft for my information what is the asset-emitter?

PR #33295, opened 2025-03-19
Update models.tsp to include safety evals raw data support

posaninagendra:
# Choose a PR Template

### DIFF
--- a/specification/ai/Azure.AI.Projects/evaluations/models.tsp
+++ b/specification/ai/Azure.AI.Projects/evaluations/models.tsp
@@ -44,6 +44,18 @@ model InputDataset extends InputData {
   id: string;
 }
 
+@doc("Dataset as source for evaluation.")
+model RawDataset extends InputData {
+  type: "dataset";
+
+  @doc("Evaluation input raw data batch id")
+  id: string;
+
+  @doc("The collection of context messages associated with this chat completions that need to be evaluated.")
+  @minItems(1)
+  value: ChatCompletions[];
+}
+
 @doc("Evaluation Definition")
 @resource("runs")
 model Evaluation {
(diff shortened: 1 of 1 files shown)


specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -44,6 +44,18 @@ model InputDataset extends InputData {
   id: string;
 }
 
+@doc("Dataset as source for evaluation.")
+model RawDataset extends InputData {
+  type: "dataset";
+
+  @doc("Evaluation input raw data batch id")
+  id: string;
```
janhavi13, line 52: What is the purpose of this id? Is the raw data going to be saved somewhere?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -44,6 +44,18 @@ model InputDataset extends InputData {
   id: string;
 }
 
+@doc("Dataset as source for evaluation.")
+model RawDataset extends InputData {
+  type: "dataset";
```
janhavi13, line 49: type is polymorphic. The type "dataset" is already used to indicate dataset asset. Can we please use another type here?

janhavi13 (changes requested):

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +90,119 @@ model Evaluation {
...
+
+  #suppress "@azure-tools/typespec-autorest/union-unsupported" "External API shape is defined in OpenAPI 3.0 as oneOf."
+  @doc("The contents of the user message, with available input types varying by selected model.")
+  content: string | ChatMessageContentItem[];
+}
+
+@doc("A request chat message representing response or action from the assistant.")
+model ChatRequestAssistantMessage extends ChatRequestMessage {
+  @doc("The chat role associated with this message, which is always 'assistant' for assistant messages.")
+  role: ChatRole.assistant;
```
singankit, line 203: Should this be user ?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
posaninagendra: The id is a way for customers and on service side to link all calls related to a dataset. For example, user with a huge file will write a batch script to run and this id can be used to link all these batch calls. Not required, just optional id. We could instead take this as header as well, I have no strong preference.

janhavi13: How will users link these runs based on ID? Is this exposed in client today?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -44,6 +44,18 @@ model InputDataset extends InputData {
   id: string;
 }
 
+@doc("Dataset as source for evaluation.")
+model RawDataset extends InputData {
```
janhavi13, line 48: What is the difference between this contract and what we have in local SDK today? Can we compare and maintain parity?

posaninagendra: You mean SubmitAnnotation API contract? It has parity with that, the property names are modified to make it more appropriate to the public REST API standard.

specification/ai/Azure.AI.Projects/evaluations/models.tsp
posaninagendra: We removed id as it is confusing, to correlated batch runs across a dataset can be done through API header something like "x-ms-dataset-id". This is purely for logging purpose on service side to correlated various calls related to a single dataset evaluation.

specification/ai/Azure.AI.Projects/evaluations/models.tsp
janhavi13: parity with the properties exposed by evaluate() in azure-ai-evaluation SDK.
Ideally we should be able to use same contract

PR #33296, opened 2025-03-19
Uprade to typespec 0.67.0

timotheeguerin:


specification/app/data-plane/Microsoft.App.DynamicSessions/preview/2024-10-02-preview/DynamicSessions.json
```diff
@@ -375,6 +374,10 @@
       "get": {
         "operationId": "SessionResourceFiles_GetContent",
         "description": "Get the content of the file.",
+        "produces": [
```
timotheeguerin, line 377: This seems unexpected, feel like it should just be octet-stream

timotheeguerin: ah no because of the error response

markcowl (approved):

PR #33395, opened 2025-03-21
Avere Azure Storage Cache Auto Import 2025-07-01 API Spec

Aman-Jain-14:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


TimLovellSmith:
@Aman-Jain-14 please fill in the PR template questionnaire (PR description)

TimLovellSmith:
>       "description": "An auto import job instance. Follows Azure Resource Manager standards: https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/resource-api-reference.md",

True but that not being a customer facing doc,  I think this comment probably isn't what we want in the public documentation for all the Azure resources.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2567 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>       "x-ms-azure-resource": true,

nit, can remove x-ms-azure-resource when you inherit TrackedResource already

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2568 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>           "x-ms-client-flatten": true,

not blocking, I think, but last i heard, flattening is NOT recommended, one reason is it sometimes causes aliasing bugs.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2577 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>               ]

nit - just remove this x-ms-mutability every time you specify it to be fully mutable, since this is the default.
```
  "x-ms-mutability": [
                "read",
                "create",
                "update"
              ]
```

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2612 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>             "adminStatus": {

was there a reason you want to call it adminStatus like API consistency?
Otherwise...
 I'm thinking other Azure APIs probably called it something different.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2597 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>             "adminStatus": {

Like, would any of these be clearer?
```
"enablement": "Enabled"
```
```
"runnable": "Disabled" 
```
```
"runsAutomatically": "Enabled" | "Disabled"
```

---
In reply to: [2779955672](https://github.com/Azure/azure-rest-api-specs/pull/33395#issuecomment-2779955672) [](http://example.com/codeflow?ancestors=2779955672)

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2597 in 8084f9d. [](commit_id = 8084f9d1103cfd837285b1a15b478213767f04ac, deletion_comment = False)

TimLovellSmith:
>             "adminStatus": {

[Or even the isEnabled=true|false pattern is *technically* still an option, though we discourage bools in general.]

---
In reply to: [2779957777](https://github.com/Azure/azure-rest-api-specs/pull/33395#issuecomment-2779957777) [](http://example.com/codeflow?ancestors=2779957777,2779955672)

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2597 in 8084f9d. [](commit_id = 8084f9d1103cfd837285b1a15b478213767f04ac, deletion_comment = False)

TimLovellSmith:
>             "adminStatus": {

PS: what does it mean by 'administrative status'? Really?

---
In reply to: [2779959661](https://github.com/Azure/azure-rest-api-specs/pull/33395#issuecomment-2779959661) [](http://example.com/codeflow?ancestors=2779959661,2779957777,2779955672)

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2597 in 8084f9d. [](commit_id = 8084f9d1103cfd837285b1a15b478213767f04ac, deletion_comment = False)

TimLovellSmith:
>               "description": "How the auto import job will handle conflicts. For example, if the auto import job is trying to bring in a directory, but a file is at that path, how it handles it. Fail indicates that the auto import job should stop immediately and not do anything with the conflict. Skip indicates that it should pass over the conflict. OverwriteIfDirty causes the auto import job to delete and re-import the file or directory if it is a conflicting type, is dirty, or is currently released. OverwriteAlways extends OverwriteIfDirty to include releasing files that had been restored but were not dirty. Please reference https://learn.microsoft.com/en-us/azure/azure-managed-lustre/ for a thorough explanation of these resolution modes.",

This is one example, but a tricky one. What about just if the file already exists, is that the same?

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2626 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>               "x-ms-enum": {

doc quality: I would highly encourage adding detailed descriptions of each option and its behavior in the x-ms-enum.values section

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2635 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                 "OverwriteIfDirty",

What does overwriteIfDirty mean? Overwrite if the destination file is dirty? Source file? Dirty in what sense? Do you have this status exposed in some particular way?

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2632 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                 "OverwriteIfDirty",

is it supposed to be like 'OverwriteIfNewer'?

---
In reply to: [2779962083](https://github.com/Azure/azure-rest-api-specs/pull/33395#issuecomment-2779962083) [](http://example.com/codeflow?ancestors=2779962083)

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2632 in 8084f9d. [](commit_id = 8084f9d1103cfd837285b1a15b478213767f04ac, deletion_comment = False)

TimLovellSmith:
>                 "create"

Why not let people change the strategy later? What if their import job is stuck because they chose 'Fail' and later they realize they wanted to change it to OverwriteAlways to get unstuck??

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2641 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                 "create"

Why not let people change the enableDeletions later? What if their import job is stuck because they chose to disable it and they change their minds?

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2650 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>               "description": "Total non-conflict-oriented errors (e.g., OS errors) Import will tolerate before exiting with failure. -1 means infinite. 0 means exit immediately on any error.",

Infinite sounds like a bad idea to me. Just saying.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2656 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>               "description": "The status of the auto import job.",

Inquiring minds want to know: is this related to its provisioningState at all, or completely independent? Can you use ProvisioningState to determine whether your job finished, or do you have to ignore provisioningState, and check 'status' instead? Or check both??

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2664 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                   "format": "int32"

Consider using bigger ints. Some day int32 could be limiting. There are more than int32 people on this planet. 

[But Int64 likely NEVER will be.]

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2699 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                       "description": "Total number of blobs imported during auto import.",

is that per second? per minute?

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2811 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                       "description": "Rate of blob import during auto import.",

is that per second? per minute?

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2817 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>         "tags": {

Instead of defining tags yourself, you're recommended to inherit the definition of tags via inheriting from TrackedResource.

If you already inherited TrackedResource, then great.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2863 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>               "$ref": "#/definitions/AutoImportJobUpdate"

more of an optional feedback but:

consider not defining separate AutoImportJob and AutoImportJobUpdate type definitions, but instead just using readonly, x-ms-mutability etc to specify which properties are updatable using PATCH

(this works fine for many APIs, and creates easier to understand SDK types - ymmv)

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:1282 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                     "preexistingSymlinks": {

I think this is an authoring BUG:
x-ms-flatten flattening is going to break you here
these property names (preexistingSymlinks etc) have the same names as properties as a level above, and due to flattening will ALIAS the property so they are both flattened onto a single level, but that can't work.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2803 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>                   "x-ms-client-flatten": true,

I think this is an authoring BUG:
x-ms-flatten flattening is going to break you here
these property names (preexistingSymlinks etc) have the same names as properties as a level above, and due to flattening will ALIAS the property so they are both flattened onto a single level, but that can't work.

---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2768 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

TimLovellSmith:
>         "nextLink": {

very nit but, usual convention is to put 'value' before 'nextLink'. [Shouldn't matter at all, it just looks funny to me]




---
Refers to: specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/amlfilesystem.json:2892 in 6fdff4d. [](commit_id = 6fdff4d5112e9a6d871028c1a776a9144c86e256, deletion_comment = False)

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/AscOperations_Get.json
```diff
@@ -0,0 +1,19 @@
...
+  "parameters": {
+    "subscriptionId": "00000000-0000-0000-0000-000000000000",
+    "location": "westus",
+    "operationId": "testoperationid",
+    "api-version": "2025-07-01"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "id": "/subscriptions/id/locations/westus/ascOperations/testoperationid",
```
TimLovellSmith, line 11: >locations [](http://example.com/codeflow?start=33&length=9)

why isn't there providers/ in this resource id?

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/autoImportJobs_CreateOrUpdate.json
```diff
@@ -0,0 +1,132 @@
...
+            "importedFiles": 8000,
+            "importedDirectories": 1000,
+            "importedSymlinks": 1000,
+            "preexistingFiles": 800000,
+            "preexistingDirectories": 80000,
+            "preexistingSymlinks": 10000,
+            "totalErrors": 10,
+            "totalConflicts": 5,
+            "blobSyncEvents": {
+              "importedFiles": 500,
```
TimLovellSmith, line 113: >importedFiles [](http://example.com/codeflow?start=15&length=13)

(just wondering, would blobsync events imported files actually give a different number of imported files from status imported files?)

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/autoImportJobs_CreateOrUpdate.json
```diff
@@ -0,0 +1,132 @@
...
+          "Dept": "ContosoAds"
+        },
+        "name": "autojob1",
+        "location": "eastus",
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/scgroup/providers/Microsoft.StorageCache/amlfilesystems/fs1/autoImportJobs/autojob1",
+        "type": "Microsoft.StorageCache/amlFilesystem/autoImportJob",
+        "properties": {
+          "adminStatus": "Enable",
+          "provisioningState": "Succeeded",
+          "autoImportPrefixes": [
```
TimLovellSmith, line 90: >autoImportPrefixes [](http://example.com/codeflow?start=11&length=18)

side remark: If in the future, you ever need the flexibility of per prefix configuration, this json format may be hard to extend.

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/autoImportJobs_CreateOrUpdate.json
```diff
@@ -0,0 +1,132 @@
...
+            "/"
+          ],
+          "conflictResolutionMode": "Skip",
+          "enableDeletions": false,
+          "maximumErrors": 0,
```
TimLovellSmith, line 95: >maximumErrors [](http://example.com/codeflow?start=11&length=13)

possible point of usability confusion:

does maximumErrors=0 mean no retries, or does maximumErrors=1 no retries?

If they mean the same, why have two different values?

Aman-Jain-14:
@TimLovellSmith regarding `adminStatus`, yes we are mostly doing it for API consistency. AutoImport is a part of our HSM operation suite which has 5 operations:
Manual Import
Manual Export
Auto Export
Auto Import
Bi-Directional Sync

Out of them we have implemented and released 2 features already (Manual Import and Auto Export) and both of them use `adminStatus`.

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/AscOperations_Get.json
Aman-Jain-14: The model validation didn't throw any errors for this.

TimLovellSmith: Right, I'm flagging it as an area for possible improvement.
Anyway, it is pretty weird to put a nonrealistic uri in the example - operation results should always be in a provider namespace shouldn't they.

specification/storagecache/resource-manager/Microsoft.StorageCache/stable/2025-07-01/examples/autoImportJobs_CreateOrUpdate.json
Aman-Jain-14: @TimLovellSmith Yes, they can have different values. The imported files in the status field is a result of initial or full scan, while the imported files in the blobSyncEvents section is the number of files imported while monitoring change feeds after the full scan was concluded.

Aman-Jain-14:
@TimLovellSmith thank you so much for your detailed review and comments. For some reason GitHub doesn't let me reply to all the comments, so I'll try to summarise my responses here while categorising the comments to some degree:

Regarding `x-ms-flatten`, I've removed that qualifier from the `blobSyncEvents` subsection of the auto import status. Hopefully that will address the issues about fields under `blobSyncEvents` subsection.

I've replaced `int32` with `int64` for all the integer status fields.

Fixed the `nextLink` and `value` ordering.

Regarding AutoImport status vs. ProvisioningState, unfortunately users cannot look at ProvisioningState to understand the state of the job itself. ProvisioningState denotes if the job was able to start or if it failed during initial stage, i.e., the synchronous part of the job. Once the control plane receives the status that the data plane layer was able to successfully start the job and the job enters the long running asynchronous phase, the user needs to poll AutoImport status to know the current state of the job. 

Regarding `maximum-errors`, there is no notion of retrying if an error is encountered. `maximum-errors` denotes the number of errors the long running auto import job will see, and still continue to progress and run. If the number of errors seen exceeds the `maximum-errors` given by user, the job reports a failure and is halted. We also noted that the idea of `infinite` is not appealing but a user can accept to not let any errors affect the progress of the auto import job itself while monitoring the logging container for the nature of the errors. Long running jobs can encounter a number of errors which may or may not affect the overall outcome of the job. This nature is also consistent with the same field defined in manual import.

Regarding json nature of `autoImportPrefixes`, we understand the concern but we have followed json formatting for similar fields in manual import an auto export features. File system admins are also more familiar and used to json formatting when using automated solutions like az cli, terraform etc.

Regarding letting users changing some strategies/nature of the job later, we do not allow the nature of the job to be changed once it has started. The only field that the user can alter is `adminStatus` to disable the job itself. Once the job is disabled, it cannot be re-enabled. To change any of the user given parameters, the user will have to disable/delete the current job and start a new job with new set of parameters.

Regarding issues with the conflict resolution modes, I have updated the link that the users can refer to for a better/detailed understanding of the modes. We understand that these modes can seem convoluted but we have tried our best to explain them in the document leveraging some understanding on the user's part regarding the Lustre filesystem, HSM and AutoImport in general. Even the folks on the control plane team found it hard to grasp the nuances but as we talked more with the data plane team and their experience with other file system admins who held varying degree of understanding of the Lustre filesystem, the fields and their usage were in-line with the wider industry usage.

gary-x-li (approved):

brpanask (approved):

gary-x-li (approved):

PR #33399, opened 2025-03-21
Version-less Pending Upload API

needuv:
# Choose a PR Template


specification/ai/Azure.AI.Projects/datasets/routes.tsp
```diff
@@ -51,11 +52,23 @@ interface Datasets
...
+   * Create or start a pending upload of a dataset for a specific version
    */
   startPendingUpload is ServicePatterns.VersionedResourceAction<
     DatasetVersion,
     PendingUploadRequest,
     PendingUploadResponse
   >;
+
+  /**
+   * Create or start a pending upload of a dataset, dataset version will be generated by service
```
dargilco, line 64: Period instead of command: "Create or start a pending upload of a dataset. Dataset version will be generated by service". 

Can we say anything about the format of the version string? e.g. is it "x.y.z"? The customer really does not know what to expect.

specification/ai/Azure.AI.Projects/datasets/routes.tsp
```diff
@@ -51,11 +52,23 @@ interface Datasets
       ListDatasetVersionsParameters
     > {
   /**
-   * Create or start a pending upload of a dataset
+   * Create or start a pending upload of a dataset for a specific version
    */
   startPendingUpload is ServicePatterns.VersionedResourceAction<
```
dargilco, line 57: Is it intentional that there is a difference between the two ServicePatterns used for these two operations?
  startPendingUpload                         is ServicePatterns.VersionedResourceAction<>
  startPendingUploadAutoIncrement is ServicePatterns.BuildingBlocks.RepeatableCoreOps.ResourceAction<>

dargilco (approved):

specification/ai/Azure.AI.Projects/datasets/routes.tsp
needuv: Updated.

Describing the autoincrement logic here would result in a very long docstring, and we have to do it in multiple places wherever autoincremented versions are used. Once we have Microsoft learn documentation, I think it would be better to put a link to that instead. I'll open a task for it.

specification/ai/Azure.AI.Projects/datasets/routes.tsp
needuv: Yes, the versioned service pattern is used for any resource actions under `/versions`

The one from `BuildingBlocks` operates at the top-level resource (i.e `/datasets`)

PR #33402, opened 2025-03-21
Use original file name when searching files

raych1:
- There are some services under specification folder adding `README.md` to describe its service. This would be mistakenly considered as 'readme.md' to be used for SDK generation. This PR changes the file comparison code to use the original file name when searching files.
Here is an example README.md:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/Communication.JobRouter/README.md

- Log warning issues in pipeline result.

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/utils.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/utils.ts
@@ -178,7 +178,7 @@ export function findParentWithFile(
     try {
       const absolutePath = path.resolve(specRepoFolder, currentPath);
       const files = fs.readdirSync(absolutePath);
-      if (files.some((file) => searchFile.test(file.toLowerCase()))) {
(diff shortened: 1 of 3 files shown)


eng/tools/spec-gen-sdk-runner/src/utils.ts
```diff
@@ -178,7 +178,7 @@ export function findParentWithFile(
     try {
       const absolutePath = path.resolve(specRepoFolder, currentPath);
       const files = fs.readdirSync(absolutePath);
-      if (files.some((file) => searchFile.test(file.toLowerCase()))) {
+      if (files.some((file) => searchFile.test(file))) {
```
mikeharder, line 181: Fine for now, but eventually I think this should move to (or use an existing) shared JS helper.

mikeharder (approved):

PR #33473, opened 2025-03-25
Gate updates for the fleet-2025-04-01-preview API

dvadas:
* Remove GateTargetType and the matching union.
  * Using the @disciminated annotation (https://typespec.io/docs/standard-library/discriminated-types) did not work once I added a second type to the union (gave warning and did not generate any swagger output for the type).
  * Instead, I've just put the updateRun target properties directly on GateTarget, making the oneof implicit.
* Remove the regex pattern from the displayName property in GateProperties.

### DIFF
--- a/specification/containerservice/Fleet.Management/gate.tsp
+++ b/specification/containerservice/Fleet.Management/gate.tsp
@@ -66,13 +66,13 @@ model GateProperties {
   @maxLength(100)
   displayName?: string;
 
-  @doc("The target that the Gate is controlling, e.g. an Update Run.")
-  @visibility(Lifecycle.Create, Lifecycle.Read)
-  target: GateTarget;
-
   @doc("The type of the Gate determines how it is completed.")
   @visibility(Lifecycle.Create, Lifecycle.Read)
   gateType: GateType;
+
+  @doc("The target that the Gate is controlling, e.g. an Update Run.")
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  target: GateTarget;
 
   @doc("The state of the Gate.")
   @visibility(Lifecycle.Create, Lifecycle.Update, Lifecycle.Read)
--- a/specification/containerservice/Fleet.Management/gate.tsp
+++ b/specification/containerservice/Fleet.Management/gate.tsp
@@ -89,25 +89,9 @@ union GateTargetType {
   UpdateRun: "UpdateRun",
 }
 
-@doc("The type of staged rollout that the Gate is targeting.")
-@added(Versions.v2025_04_01_preview)
-union GateTargetProperties {
-  @doc("The Gate is controlling the staged rollout of an Update Run.")
-  updateRunProperties: UpdateRunGateTargetProperties,
-}
-
-// A wrapper is needed because the union does not create one itself. Without it, properties of
-// all different types would end up in a single `properties` model.
-@doc("Wrapper object around the Update Run Gate Target Properties.")
+@doc("The properties of the Update Run that the Gate is targeting.")
 @added(Versions.v2025_04_01_preview)
 model UpdateRunGateTargetProperties {
-  @doc("Properties specific to the Update Run.")
-  updateRun: UpdateRunGateTarget;
-}
-
-@doc("The Gate is targeting an Update Run staged rollout.")
-@added(Versions.v2025_04_01_preview)
-model UpdateRunGateTarget {
   @doc("The name of the Update Run.")
   @visibility(Lifecycle.Read)
   @pattern("^[a-z0-9]([-a-z0-9]*[a-z0-9])?$")
@@ -160,20 +144,16 @@ union GateTargetId {
   updateRun: UpdateRunResourceId,
 }
 
-@doc("The target that the Gate is controlling, e.g. an Update Run.")
+@doc("The target that the Gate is controlling, e.g. an Update Run. Exactly one of the properties objects will be set.")
 @added(Versions.v2025_04_01_preview)
 model GateTarget {
   @doc("The resource id that the Gate is controlling the rollout of.")
   @visibility(Lifecycle.Create, Lifecycle.Read)
   id: GateTargetId;
 
-  @doc("The type of the resource being targeted by this Gate.")
+  @doc("The properties of the Update Run that the Gate is targeting.")
   @visibility(Lifecycle.Create, Lifecycle.Read)
-  type: GateTargetType;
-
-  @doc("Properties specific to the target of this Gate.")
-  @visibility(Lifecycle.Create, Lifecycle.Read)
-  properties: GateTargetProperties;
+  updateRunProperties?: UpdateRunGateTargetProperties;
 }
 
 @doc("The provisioning state of the Gate resource.")
--- a/specification/containerservice/Fleet.Management/gate.tsp
+++ b/specification/containerservice/Fleet.Management/gate.tsp
@@ -45,6 +45,9 @@ union GateState {
   @doc("A Pending Gate will continue to block the staged rollout process it is controlling.")
   Pending: "Pending",
 
+  @doc("A Skipped Gate means that the staged rollout process it is controlling was skipped.")
+  Skipped: "Skipped",
+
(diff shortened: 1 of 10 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
```diff
@@ -3436,60 +3434,23 @@
...
-    "GateTargetProperties": {
-      "type": "object",
-      "x-nullable": false,
-      "allOf": [
-        {
-          "$ref": "#/definitions/UpdateRunGateTargetProperties"
-        }
-      ]
-    },
-    "GateTargetType": {
```
serbrech, line 3475: did you compile the typespec?
why is GateTargetType removed from the json?

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -37,12 +37,12 @@ model UpdateStage {
 
   @doc("A list of Gates that will be created before this Stage is executed.")
   @added(Versions.v2025_04_01_preview)
-  @extension("x-ms-identifiers", ["displayName"])
+  @extension("x-ms-identifiers", #["displayName"])
```
serbrech, line 40: what's the hash for?

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -37,12 +37,12 @@ model UpdateStage {
...
   beforeGates?: GateConfiguration[];
 
   @doc("A list of Gates that will be created after this Stage is executed.")
   @added(Versions.v2025_04_01_preview)
-  @extension("x-ms-identifiers", ["displayName"])
+  @extension("x-ms-identifiers", #["displayName"])
```
serbrech, line 45: AFAIK, the x-ms-identifier needs to be unique in the array.
is that goingto be the case for displayName?

specification/containerservice/Fleet.Management/update/common.tsp
serbrech: https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/missing-x-ms-identifiers/#_top

specification/containerservice/Fleet.Management/update/common.tsp
serbrech: https://azure.github.io/autorest/extensions/#x-ms-identifiers

serbrech: I think that the right approach here is :
> If there is no appropriate identifying properties, leave it as an empty array.

specification/containerservice/Fleet.Management/gate.tsp
```diff
@@ -160,20 +143,16 @@ union GateTargetId {
   updateRun: UpdateRunResourceId,
```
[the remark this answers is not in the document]
dvadas: There's a prior discussion about this here: https://github.com/Azure/azure-rest-api-specs/pull/33074#discussion_r1988238921

Trying it now (i.e. adding `string` and suppressing the warning), it fails to generate anything:

```
$ git diff
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
@@ -3447,10 +3447,7 @@
         "id"
       ]
     },
-    "GateTargetId": {
-      "$ref": "#/definitions/UpdateRunResourceId",
-      "x-nullable": false
-    },
+    "GateTargetId": {},
     "GateType": {
```

So maybe we should just use a plain string type and remove the union? We might also get advice on this during the ARM review.

CC @serbrech

specification/containerservice/Fleet.Management/update/common.tsp
dvadas: After upgrading the Typspec compiler to 0.67 you get this error:

```
update/common.tsp:40:34 - error expect-value: Is a tuple type, but is being used as a value here. Use #[] to create an array value.
> 40 |   @extension("x-ms-identifiers", ["displayName"])
     |             
```

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
dvadas: It's removed from the tsp as well (was at gate.tsp:172). This was based on the conversation with @matthchr (see https://github.com/Azure/azure-rest-api-specs/pull/33074#discussion_r1997923736).

The downsides of the type are:
  * it's redundant
  * it allows the object to get into an invalid state (e.g. `type=UpdateRun`, `properties=someOtherType`).

Some relevant quotes from that thread:

>  (I'm not 100% sure I've ever under stood the need for the discriminator field from a pure REST modelling perspective, I think it exists to help the clients figure out how to deserialize the polymorphic type, but if you're not actually going to do polymorphism and instead just going to define mutually exclusive fields, which I also happen to prefer as I think it's simpler in REST, I don't think you want/need a type field at all.

> I think you need to decide what value type is adding for you. You can just as easily write this code:
>
> ```
> if target.updateRunProperties != null {
>   // process update run target
> } else if target.otherTargetProperties != null {
>   // process other target type
> }
> ```

specification/containerservice/Fleet.Management/update/common.tsp
dvadas: Agreed. Done.

specification/containerservice/Fleet.Management/gate.tsp
```diff
@@ -160,20 +143,16 @@ union GateTargetId {
...
 
-  @doc("The type of the resource being targeted by this Gate.")
-  @visibility(Lifecycle.Create, Lifecycle.Read)
-  type: GateTargetType;
-
-  @doc("Properties specific to the target of this Gate.")
+  @doc("The properties of the Update Run that the Gate is targeting.")
   @visibility(Lifecycle.Create, Lifecycle.Read)
-  properties: GateTargetProperties;
+  updateRunProperties?: UpdateRunGateTargetProperties;
```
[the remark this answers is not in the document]
dvadas: Done.

specification/containerservice/Fleet.Management/gate.tsp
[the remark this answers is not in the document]
circy9: I am ok not addressing the comment. I just realized it is not a regular string.

circy9 (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
matthchr: I'm very curious @serbrech's take on this.

specification/containerservice/Fleet.Management/gate.tsp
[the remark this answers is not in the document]
serbrech: different unions.
union for string generates an enum, while this is a "oneof"

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2025-04-01-preview/fleets.json
serbrech: I think the difference is that with a union and discriminator, the same json property name can have different values entirely?
I'm all for avoiding "polymorphism" though :)

serbrech: see the guidelines on "polymorphism"
https://github.com/microsoft/api-guidelines/blob/c2ef3c5cba631dd1d2bfcd6d88d69e7db944ca3e/azure/ConsiderationsForServiceDesign.md#avoid-surprises

serbrech (approved):

PR #33488, opened 2025-03-25
Support new parameters in pipeline

raych1:
**Pipeline**
- Added `ApiVersion` and `SdkReleaseType` parameters to the pipeline configuration. These parameters are now passed to the pipeline steps
- Included `ApiVersion` and `SdkReleaseType` parameters in the template and added validation checks to ensure they are set correctly

**Runner**
- Modified the `parseArguments` and `prepareSpecGenSdkCommand` functions to handle the new `apiVersion` and `sdkReleaseType` parameters.


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -17,6 +14,14 @@ parameters:
     type: string
     default: ''
     displayName: 'Path to API specification file (TypeSpec/OpenAPI)'
+  - name: ApiVersion
+    type: string
+    default: ''
```
benbp, line 19: Given you're checking for empty on this below, do you want to also remove `default` here to make it required?

benbp (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: It's required param in the entry yml file.

eng/pipelines/spec-gen-sdk.yml
```diff
@@ -16,9 +14,20 @@ parameters:
...
+  - name: SdkReleaseType
+    type: string
+    values:
+      - 'beta'
+      - 'stable'
+    default: 'beta'
+    displayName: 'SDK release type'
   - name: SkipPullRequestCreation
     type: boolean
     default: false
```
weshaggard, line 29: Is this the right default? I kind of think we shouldn't do a pull request by default.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -30,7 +35,7 @@ extends:
     Use1ESOfficial: false
     stages:
     - stage: Build
-      displayName: 'SDK Generation'
+      displayName: 'SDK Generation Preview'
```
weshaggard, line 38: Didn't we want to call this `[Ingore testing]` as a prefix?

raych1: I will add [TEST IGNORE] to `spec-gen-sdk - pull request` pipeline name. But, for job as it's used for release scenario I would rather add 'preview' instead of [TEST IGNORE].

weshaggard (approved):

eng/pipelines/spec-gen-sdk.yml
raych1: It should have a default value otherwise pull request trigger would fail. Downstream language side would reply on 'api-version' to determine if this parameter needs to be respected or not. So, setting it with 'beta' won't matter.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: I forgot that the default must be set to support PR trigger scenario. However, even if I leave default value as empty string, it would still display the parameter in pipeline UI as required.

eng/pipelines/spec-gen-sdk.yml
weshaggard: I was asking about the SkipPullRequestCreation default and wondering if it should be flipped.

PR #33495, opened 2025-03-26
Update tspconfig with kebab case style option

v-jiaodi:
related issue https://github.com/Azure/autorest.typescript/issues/3067


MaryGao (approved):

MaryGao:
Bypass the rule since only JS emitter config changes.

PR #33499, opened 2025-03-26
test js breaking detection if remove apiversion enum in main.tsp

kazrael2119:


kazrael2119:
close as js automation works well

PR #33535, opened 2025-03-27
Migrate to new multipart

timotheeguerin:
Wip, to review all changes


specification/ai/data-plane/Face/preview/v1.2-preview.1/Face.json
```diff
@@ -6981,25 +6981,6 @@
         "livenessOperationMode"
       ]
     },
-    "CreateLivenessWithVerifySessionMultipartContent": {
```
timotheeguerin, line 6984: This one I htink is correct, this model is now not used anywhere in the swagger due to params being inline

specification/translation/data-plane/Azure.AI.DocumentTranslation/stable/2024-05-01/openapi.json
```diff
@@ -728,17 +726,12 @@
       "description": "Document Translate Request Content",
       "properties": {
         "document": {
-          "type": "string",
-          "format": "byte",
+          "type": "object",
```
timotheeguerin, line 729: the reason for those kind of changes is that now this model is an unreachable model where the httppart is now an empty object

specification/ai/data-plane/Face/stable/v1.2/Face.json
```diff
@@ -475,22 +475,65 @@
...
         "parameters": [
           {
-            "$ref": "#/parameters/CreateLivenessSessionContent.livenessOperationMode"
-          },
-          {
-            "$ref": "#/parameters/CreateLivenessSessionContent.deviceCorrelationIdSetInClient"
-          },
-          {
-            "$ref": "#/parameters/CreateLivenessSessionContent.enableSessionImage"
+            "name": "livenessOperationMode",
```
timotheeguerin, line 478: here the parameters were moved to be inline, it doesn't really matter and I don't think its worth to cheat those back in with some spread as the definition name at least would need to change

PR #33536, opened 2025-03-27
Disable SDL in unofficial 1es template

raych1:
Followed the [document](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/sdlanalysis/disablingtools) disabling all the tools. The job will still run, but the the time is on the 'check out' step and all the SDL jobs are skipped.

[Test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4696606&view=results)
image

For comparison, this is [another run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4696257&view=results) with SDL enabled.
image


weshaggard:
We want to see about skipping the entire stag so that we can eliminate the noise in the checks list. I don't see anything in the documentation but looking at the source of the templates perhaps https://dev.azure.com/azure-sdk/1ESPipelineTemplates/_git/1ESPipelineTemplates?path=/v1/1ES.Pipeline.yml&version=GBmain&line=436&lineEnd=437&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents it seems like we should be able to disable it if we don't have any releaseJobs.

raych1:
> We want to see about skipping the entire stag so that we can eliminate the noise in the checks list. I don't see anything in the documentation but looking at the source of the templates perhaps https://dev.azure.com/azure-sdk/1ESPipelineTemplates/_git/1ESPipelineTemplates?path=/v1/1ES.Pipeline.yml&version=GBmain&line=436&lineEnd=437&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents it seems like we should be able to disable it if we don't have any releaseJobs.

It had same run result after set job type to 'buildJob'.  Actually1es team has a pending [work item](https://dev.azure.com/mseng/1ES/_workitems/edit/2253084) in backlog to disable 'guardian' tasks when `enableAllTools` option is set to false.

weshaggard:
If we cannot figure out a way to disable the entire SDLSources stage then is it even worth taking this change? We might want to post a question to the 1ESPT teams channel to see if we have some option to disable this stage completely.

eng/pipelines/templates/stages/1es-redirect.yml
```diff
@@ -22,21 +22,29 @@ extends:
...
-        image: windows-2022
-        os: windows
-      psscriptanalyzer:
-        compiled: true
-        break: true
-      policy: M365
-
-    stages: ${{ parameters.stages }}
\ No newline at end of file
+    ${{ if and(parameters.Use1ESOfficial, eq(variables['System.TeamProject'], 'internal')) }}:
```
weshaggard, line 25: We should not duplicate this condition. Instead we should just move the settings under the conditions above where we are pulling in the template.

eng/pipelines/templates/stages/1es-redirect.yml
```diff
@@ -22,21 +22,29 @@ extends:
...
+          os: windows
+        psscriptanalyzer:
+          compiled: true
+          break: true
+        policy: M365
+      stages: ${{ parameters.stages }}
+    ${{ else }}:
+      sdl:
+        enableAllTools: false
+        sourceAnalysisPool:
```
weshaggard, line 46: Let's add a comment here linking to the issue about removing this entire stage.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -57,6 +57,7 @@ extends:
           os: linux
 
         templateContext:
+          type: 'buildJob'
```
weshaggard, line 60: I don't think this is needed and can be removed.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -174,14 +175,14 @@ extends:
 
               if [ "${{ parameters.SpecBatchTypes }}" != "" ]; then
                 optional_params="$optional_params --rm ${{ parameters.SpecBatchTypes }}"
-                sdk_gen_info="$sdk_gen_info, SpecBatchTypes: ${{ parameters.SpecBatchTypes }}"
+                sdk_gen_info="$sdk_gen_info SpecBatchTypes: ${{ parameters.SpecBatchTypes }}"
```
weshaggard, line 178: Is this fixing another unrelated bug?

eng/pipelines/templates/stages/1es-redirect.yml
raych1: It will have syntax errors, i.e. parameters section cannot be put within 'if' condition.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: yes, removing extra comma. Do you want me to have separate PR?

eng/pipelines/templates/stages/1es-redirect.yml
weshaggard: We are already splitting the parameters here so just do it above is all I was suggesting.

raych1: Are you suggesting this one, it has syntax errors.
```yaml
extends:
  ${{ if and(parameters.Use1ESOfficial, eq(variables['System.TeamProject'], 'internal')) }}:
    template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
    parameters:
      settings:
          skipBuildTagsForGitHubPullRequests: true
        sdl:
          ${{ if and(parameters.GenerateBaselines, eq(variables['Build.SourceBranchName'], 'main'), eq(variables['System.TeamProject'], 'internal')) }}:
            autobaseline:
              isMainPipeline: true
              disableAutoBaselineOnNonDefaultBranches: true
              enableForGitHub: true
          sourceAnalysisPool:
            name: azsdk-pool
            image: windows-2022
            os: windows
          psscriptanalyzer:
            compiled: true
            break: true
          policy: M365
        stages: ${{ parameters.stages }}
  ${{ else }}:
    template: v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates
    parameters:
      sdl: 
        enableAllTools: false 
        sourceAnalysisPool:
          name: azsdk-pool
          image: windows-2022
          os: windows
      stages: ${{ parameters.stages }}
```

eng/pipelines/templates/stages/1es-redirect.yml
raych1: Added comment.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
weshaggard: No, it is fine I just wanted to make sure it wasn't a mistake.

eng/pipelines/templates/stages/1es-redirect.yml
weshaggard: Yes that is what I'm suggesting. It looks like you are just missing the indention of the parameters section which needs one more level of indention.

raych1: ADO doesn't allow to do that. 'extends' must have 'template' and 'parameter' on same level.
https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/extends?view=azure-pipelines

raych1: Interesting, the pipeline works in my new test run with the above yaml content. It didn't work last time in which I may mess up something. Can you review the latest commit?

raych1: This is the [latest test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4706574&view=results).

weshaggard (approved):

eng/pipelines/templates/stages/1es-redirect.yml
```diff
@@ -19,24 +19,31 @@ parameters:
 extends:
   ${{ if and(parameters.Use1ESOfficial, eq(variables['System.TeamProject'], 'internal')) }}:
     template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+    parameters:
+      settings:
+          skipBuildTagsForGitHubPullRequests: true
```
weshaggard, line 24: @raych1 you need to add this to the unofficial parameters as well.

PR #33546, opened 2025-03-27
[TypeSpec Validation] Fail if spec folder contains extra tsp-generated swagger files

### LINKED ISSUE #32861: [TypeSpecValidation] Fail if spec folder contains extra tsp-generated swagger files
As the title, can we provide a validation to make sure the TypeSpec and swagger are consistent in all versions not just the latest version? 
In this case https://github.com/Azure/azure-rest-api-specs/pull/30670/files#r1811975454 the service team is removing a version enum value, to really reflect the changes, we should also remove the corresponding swaggers for 2024-09-15-preview, but those swaggers are still there https://github.com/nonsocode/azure-rest-api-specs/tree/nchukwuogor/aio-ga/specification/iotoperations/resource-manager/Microsoft.IoTOperations/preview/2024-09-15-preview, 

 From my understanding, removing a version enum value in TypeSpec equals to deprecate an api version, which should involve the API stewardship board to review. And all the api reference documentation related to that api version, should be removed, as that API reference document is generated by our swagger, therefore, we have to remove those swaggers as well. /cc @JeffreyRichter

mikeharder:
- Fixes #32861

Example failure: https://github.com/Azure/azure-rest-api-specs/actions/runs/14213542174/job/39825063555?pr=33653


eng/tools/typespec-validation/src/rules/compile.ts
```diff
@@ -14,21 +17,127 @@ export class CompileRule implements Rule {
...
+            (p) => normalize(p),
+          );
+
+          // Filter to files generated by TypeSpec
+          const tspGeneratedSwaggers = await filterAsync(
+            allSwaggers,
+            async (swaggerPath: string) => {
+              const swaggerText = await host.readFile(swaggerPath);
+              const swaggerObj = JSON.parse(swaggerText);
+              return swaggerObj["info"]?.["x-typespec-generated"];
```
qiaozha, line 97: there're some cases where the swagger was generated from cadl 
```suggestion
              return swaggerObj["info"]?.["x-typespec-generated"] || swaggerObj["info"]?.["x-cadl-generated"];
```

qiaozha (approved):

eng/tools/typespec-validation/src/rules/compile.ts
```diff
@@ -14,21 +17,127 @@ export class CompileRule implements Rule {
...
+          const lines = stripAnsi(stdout).split(/\r?\n/);
+
+          // TODO: Use helpers in /.github once they support platform-specific paths
+          const outputSwaggers = lines
+            // Remove leading and trailing whitespace
+            .map((l) => l.trim())
+            // Skip blank lines
+            .filter((l) => l !== "")
+            // Skip first and last lines (version and success messages)
+            .slice(1, -1)
```
timotheeguerin, line 60: is this one necessary if you already try to filter by `.json` later, feel like this is something more likely to break the logic

eng/tools/typespec-validation/src/rules/compile.ts
```diff
@@ -14,21 +17,127 @@ export class CompileRule implements Rule {
...
+            // Normalize to platform-specific path
+            .map((l) => normalize(l))
```
timotheeguerin, line 62: your choice here but I do find that its much easier to not have to deal with windows path and always working only with `/` makes it much better. Typespec should only ever return unix style path. 
We have our own internal path utils that does the logic (forked from typescript) but also the package `pathe` just does that too

mikeharder: I have a bigger issue tracking this for all the engsys code in this repo: https://github.com/Azure/azure-rest-api-specs/issues/33193

My conclusion was to try to match the behavior of the Node `fs` APIs, which I believe return platform-specific paths.

Are you suggesting working internally with only posix paths?  Couldn't this create problems, unless you can guarantee all related code is also assuming this?

mikeharder: OK, if this is what TypeSpec does, I'm willing to give it a shot.

mikeharder: Actually, I decided to keep using platform-specific paths in engsys code for now.  It's safer since we mix a lot of languages and tools.

PR #33565, opened 2025-03-28
fix: spec for list keys api functions webapp

munishsuri:
# Choose a PR Template

Response from Spec of List Keys currently points to String dictionary , which is not correct, it should be function keys , as depicted in the screenshot.

1. Changing the Spec to Correct Info
2. Creating a definition for function key

Screenshot 2025-03-27 at 3 42 06 PM


timotheeguerin:
This PR has been closed because it has had no activity for more than a year. If this change is still needed, please reopen the PR.

PR #33566, opened 2025-03-28
Address some additional TypeSpec issues

dargilco:
1) Add `AuthType` property on the Connection object. Values were copied from the TypeSpec we currently use (`feature/azure-ai-projects` branch)

2) Align on the names of operators with version and without version. Before this PR we have:
[/datasets and /indexes] createVersion
[/datasets and /indexes] create
[/datasets] startPendingUpload
[/datasets] startPendingUploadAutoIncrement

This PR updates the above two to the following:
[datasets] startPendingUploadVersion
[datasets] startPendingUpload

Such that we are consistent -- operator names that require the caller to specify a version end with "Version".  Operator names that do not required a version (it's auto-generated by the service) will not have "Version" at the end.

3) In the /deployment "list" operator, remove the option to include models deployed to connections. Since that will not be ready for build.


needuv (approved):

specification/ai/Azure.AI.Projects/connections/models.tsp
```diff
@@ -56,3 +59,23 @@ union ConnectionType {
   @doc("Custom Keys")
   Custom: "CustomKeys",
 }
+
+@doc("The authentication type used by the connection")
+union AuthenticationType {
+  string,
+
+  @doc("API Key authentication")
```
JeffreyRichter, line 67: Due to Microsoft's SFI initiative, Azure services are supposed to support only Entra ID now and NOT support api-key or SAS (unless it is delegated SAS). If you need to support api-key or non-delegated SAS for getting-started scenarios (but not for customer long-term production use scenarios), then a short expiration (perhaps a week or so) on the key/sas is required.

dargilco: @JeffreyRichter thanks for the note, I'll follow up with you offline on this.

PR #33581, opened 2025-03-31
Remove some options  in tspconfig for mgmt plane

v-jiaodi:
remove: `azure-sdk-for-js`, `generate-metadata` , `hierarchyClient`, `enableOperationGroup`


MaryGao (approved):
Approved since they are only JS mpg SDKs, no real impact.

maririos:
@MaryGao and @v-jiaodi by deleting these properties you need to make sure to:

- Remove these properties from the TypeSpec template too. Otherwise any new team that uses that template will have these properties: https://github.com/Azure/typespec-azure/blob/main/eng/feeds/arm/tspconfig.yaml
- Update the JS documentation: https://github.com/Azure/azure-sdk-for-js/blob/main/documentation/Generate-code-from-TypeSpec.md#how-to-configure-tspconfigyaml
- Would it be ok to add a rule to the Tspconfig validation about not using these 2 properties? https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts

FYI @wanlwanl @raych1

MaryGao:
@maririos thanks for the information! 

@v-jiaodi could you have a pr to update for item 1 and item 2? I believe for item 3 we have already updated.

v-jiaodi:
> @maririos thanks for the information!
> 
> @v-jiaodi could you have a pr to update for item 1 and item 2? I believe for item 3 we have already updated.

 - item 1 : https://github.com/Azure/typespec-azure/pull/2541
 - item 2: https://github.com/Azure/azure-sdk-for-js/pull/33952

MaryGao:
@maririos Could you help understand why we have templates in typespec-azure repo? Is it because release planner uses them?

 Here is the pr to update the template also: https://github.com/Azure/typespec-azure/pull/2541/files.

PR #33593, opened 2025-03-31
[Storage] Omit response headers for certain headers

vincenttran-msft:
The changes to the header traits was applied in a multi-filtered fashion.

At the top-level in the template, we removed the following:
- request ID
- client request ID

On a **per-API basis**, but appears in multiple, we looked for:
- date
- accept-ranges
- structured_body_type

-----------------------------------------------

The only other specific additions was the following:
**Create_container**
- EtagResponseHeader
- LastModifiedResponseHeader

Note: This was done to get `create` to return `Response<()>`


specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -42,8 +42,6 @@ namespace Storage.Blob {
     @header("Content-Type")
     contentType: ResponseMediaType;
 
-    ...RequestIdResponseHeader;
-    ...ClientRequestIdHeader;
```
vincenttran-msft, line 46: These are off the template (so applies to all)

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -184,9 +182,6 @@ namespace Storage.Blob {
       },
       {
         @statusCode statusCode: 201;
-        ...EtagResponseHeader;
-        ...LastModifiedResponseHeader;
-        ...DateResponseHeader;
```
vincenttran-msft, line 189: `create_container`

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -205,7 +200,6 @@ namespace Storage.Blob {
         ...MetadataHeaders;
         ...EtagResponseHeader;
         ...LastModifiedResponseHeader;
-        ...DateResponseHeader;
```
vincenttran-msft, line 208: `get_container_properties`

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -246,7 +240,6 @@ namespace Storage.Blob {
       },
       {
         @statusCode statusCode: 202;
-        ...DateResponseHeader;
```
vincenttran-msft, line 249: `delete_container`

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -610,10 +603,6 @@ namespace Storage.Blob {
         @header("x-ms-range-get-content-crc64")
         rangeGetContentCrc64?: boolean,
 
-        /** Specifies the response content should be returned as a structured message and specifies the message schema version and properties. */
-        @header("x-ms-structured-body")
-        structuredBodyType?: string,
-
```
vincenttran-msft, line 616: `download` blob API
This removes the parameter on download API to return result as structured message
image

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -652,8 +641,6 @@ namespace Storage.Blob {
         ...LeaseStatusResponseHeader;
         ...VersionIdResponseHeader;
         ...IsCurrentVersionResponseHeader;
-        ...AcceptRangesResponseHeader;
-        ...DateResponseHeader;
```
vincenttran-msft, line 656: `download` blob API 
response headers changes 1

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -665,8 +652,6 @@ namespace Storage.Blob {
         ...ImmutabilityPolicyExpiresOnResponseHeader;
         ...ImmutabilityPolicyModeResponseHeader;
         ...LegalHoldResponseHeader;
-        ...StructuredBodyTypeResponseHeader;
-        ...StructuredContentLengthResponseHeader;
```
vincenttran-msft, line 669: `download` blob API 
response headers changes 2

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -717,8 +702,6 @@ namespace Storage.Blob {
...
```
vincenttran-msft, line 721: `download` blob API 
response headers changes 3

I still don't fully understand this portion, we can ask Joel once we get closer to a finished state on this proposed changes. If I had to guess, one of them is on the literal download response, the other one is in the body which is a part of the download response (the body response in the download response contained a dupe?)

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -773,8 +756,6 @@ namespace Storage.Blob {
           ...ContentLanguageResponseHeader;
           ...CacheControlResponseHeader;
           ...BlobSequenceNumberResponseHeader;
-          ...DateResponseHeader;
-          ...AcceptRangesResponseHeader;
```
vincenttran-msft, line 777: `get_blob_properties`

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -837,7 +818,6 @@ namespace Storage.Blob {
         },
         {
           @statusCode statusCode: 202;
-          ...DateResponseHeader;
```
vincenttran-msft, line 840: `delete` blob API

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -1890,8 +1870,6 @@ namespace Storage.Blob {
             ...ImmutabilityPolicyModeParameter;
             ...LegalHoldOptionalParameter;
             ...ContentCrc64Parameter;
-            ...StructuredBodyPut;
-            ...StructuredContentLengthParameter;
```
vincenttran-msft, line 1894: `upload` API

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -1903,11 +1881,9 @@ namespace Storage.Blob {
             ...LastModifiedResponseHeader;
             ...ContentMd5ResponseHeader;
             ...VersionIdResponseHeader;
-            ...DateResponseHeader;
             ...IsServerEncryptedResponseHeader;
             ...EncryptionKeySha256ResponseHeader;
             ...EncryptionScopeResponseHeader;
-            ...StructuredBodyTypeResponseHeader;
```
vincenttran-msft, line 1910: `upload` API

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -1992,18 +1968,14 @@ namespace Storage.Blob {
...
           },
           {
             @statusCode statusCode: 201;
-            ...DateResponseHeader;
             ...ContentMd5ResponseHeader;
             ...ContentCrc64ResponseHeader;
             ...IsServerEncryptedResponseHeader;
             ...EncryptionKeySha256ResponseHeader;
             ...EncryptionScopeResponseHeader;
-            ...StructuredBodyTypeResponseHeader;
```
vincenttran-msft, line 2006: `stage_block` API

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -2091,7 +2063,6 @@ namespace Storage.Blob {
             ...ContentMd5ResponseHeader;
             ...ContentCrc64ResponseHeader;
             ...VersionIdResponseHeader;
-            ...DateResponseHeader;
```
vincenttran-msft, line 2094: `commit_block_list`

vincenttran-msft:
Hi @jhendrixMSFT Joel, this should represent the changes we want to make for Storage. I did have one question for the changes I made to `download`. I saw that there are several responses for download (one for partial content, one for error case), so just wanted to call out to make sure that the way I handled omitting the responses there was correct for that "complex" case. Thanks!

specification/storage/Microsoft.BlobStorage/routes.tsp
jhendrixMSFT: Was the parameter incorrect?

specification/storage/Microsoft.BlobStorage/routes.tsp
jhendrixMSFT: The first response is to handle the HTTP 200 case (if no status code is specified, 200 is implied).

The second response is to handle the HTTP 206 case for downloading a specified range. Note the `@statusCode statusCode: 206` on line 672.

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -184,9 +182,6 @@ namespace Storage.Blob {
       },
       {
         @statusCode statusCode: 201;
-        ...EtagResponseHeader;
```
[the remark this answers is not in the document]
vincenttran-msft: Typically in the other language SDKs, `create` and `delete` generally don't have a response, so this was done to get the effect of `Response<()>`. Then customers could still get it directly from headers if needed 😄

specification/storage/Microsoft.BlobStorage/routes.tsp
vincenttran-msft: Hi Joel, for context-- the parameter is not incorrect but structured message is only for content validation (which is NYI) and there is discussion internally on Storage's end on how we want to implement content validation- so we decided it would be best to not expose the current impl details to customers especially since it may not represent the final cut.

PR #33625, opened 2025-04-01
Add new APC API version

SallyAnneBennett:
Add new APC API version, changing the following:

- Moving Application properties to Gateway from Operator Api Connections, and some properties from Application properties to Application Owner properties
- Add many properties required by Operators for using the Operate API (which this API passes the properties through to)
- Add properties required to change our integration with Marketplace, and remove properties that were needed but are no longer relevant
- Make properties required that should have been required previously
- Remove properties relating to User Managed credentials since we have never supported them
- Some minor fixes to documentation strings


AndrewNivenMSFT (approved):

scomtott (approved):

SallyAnneBennett:
Note to ARM reviewer:
- This change has previously been signed off with "ARMSignedOff", but we realised we didn't need a field that we'd added in this new version and so I removed it and that appears to have requested ARM review again
- There are a few changes in this PR which are not exactly as ARM would ideally have them (e.g. boolean not ENUM), but we are a passthrough API to the TMF931 Operate API and in a number of places I have matched the ARM API to the Operate API rather than require a translation. (We'd prefer not to be a passthrough API at all, but there are fields our Partner companies require that we can't get otherwise.)

AndrewNivenMSFT (approved):

scomtott (approved):

PR #33671, opened 2025-04-02
Added fixtures for spec-gen-sdk-runner tests

### LINKED ISSUE #33071: [spec-gen-sdk-runner] Remove dependency from unit tests to specification/contosowidgetmanager
The unit tests currently depend on `specification/contosowidgetmanager`, which is undesirable for a few reasons:

1. You need to clone this folder to run the tests.  The tool isn't self-contained.
https://github.com/Azure/azure-rest-api-specs/blob/8a0168458930c86636a76bcd7acfdc9c81291bfc/.github/workflows/spec-gen-sdk-runner-test.yaml#L30-L32
2. Changes to the sample could break your tests.  These should be decoupled.
3. Your tests fail in RPSaaSMaster, since this branch doesn't have the contoso data-plane. (example: https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/13711484022)

Please update your tests to use a sample spec under your own test folder.  This could even be a copy of contoso if you want. 
 Example:

https://github.com/Azure/azure-rest-api-specs/tree/main/eng/tools/typespec-requirement/test/specification

raych1:
* Added test fixtures
* Updated tests to use test fixtures removing dependency on the specification folder (differences exist across branches)

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/test/fixtures/specification/contosowidgetmanager/Contoso.Management/main.tsp
+++ b/eng/tools/spec-gen-sdk-runner/test/fixtures/specification/contosowidgetmanager/Contoso.Management/main.tsp
@@ -1,0 +1,35 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/versioning";
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-azure-resource-manager";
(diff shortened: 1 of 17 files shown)


raych1:
@mikeharder can you review this PR?

mikeharder (approved):

PR #33684, opened 2025-04-02
Update enterprisePolicy.json

cmerino01:
Updating `VirtualNetworkPropertiesList` definition. 

We are not actually breaking the API, rather the spec was wrong from the get go and we can't change it since it's already shipped and in use code so we have to fix the documentation here.
image


cmerino01:

ryan-mcculloch (approved):

PR #33701, opened 2025-04-03
Update enterprisePolicy.json - updating VirtualNetworkPropertiesList

cmerino01:
We are not actually breaking the API, rather the spec was wrong from the get go and we can't change it since it's already shipped and in use code so we have to fix the documentation here.

image


cmerino01:
Awaiting approval for proper entitlement.

specification/powerplatform/suppressions.yaml
```diff
@@ -0,0 +1,3 @@
+- tool: TypeSpecRequirement
+  path: powerplatform/resource-manager/Microsoft.PowerPlatform/preview/2020-10-30-preview/enterprisePolicy.json
```
mikeharder, line 2:
```suggestion
  path: resource-manager/Microsoft.PowerPlatform/preview/2020-10-30-preview/enterprisePolicy.json
```

path is relative to suppressions.yaml

yirsu (approved):

yirsu (approved):

yirsu (approved):

yirsu (approved):

PR #33782, opened 2025-04-08
[Swagger Avocado] Convert to GitHub Workflow

mikeharder:
- Workflow is very simple, because external tool `avocado` includes its own code to:
  - Detect files changed in a PR
  - Ignore (certainly types of) errors also appearing in the base branch
- Console output is sufficient to verify tool ran on correct folders and show any errors

# Before
image
https://github.com/Azure/azure-rest-api-specs/pull/33067/checks?check_run_id=40150938473

# After
image
https://github.com/mikeharder/azure-rest-api-specs/actions/runs/14325792724/job/40150937940?pr=17

### DIFF
--- a/.github/workflows/avocado.yml
+++ b/.github/workflows/avocado.yml
@@ -1,0 +1,36 @@
+name: "[TEST-IGNORE] Swagger Avocado"
+
+on: pull_request
+
+permissions:
+  contents: read
+
+jobs:
+  avocado:
+    name: "[TEST-IGNORE] Swagger Avocado"
+
+    runs-on: ubuntu-24.04
+
+    steps:
+      # Default checkout of head branch with depth=1 should be sufficient,
+      # since avocado has internal code to clone and diff the base branch
+      # (passed via env var SYSTEM_PULLREQUEST_TARGETBRANCH).
+      - uses: actions/checkout@v4
+
+      - name: Setup Node and install deps
+        uses: ./.github/actions/setup-node-install-deps
+
+      - name: Run Avocado
+        run: |
+          npm exec --no avocado \
+          --excludePaths \
+            "/common-types/" \
+            "/scenarios/" \
+            "/package.json" \
+            "/package-lock.json" \
+            "/cadl/examples/" \
+          --includePaths \
+            "data-plane" \
+            "resource-manager"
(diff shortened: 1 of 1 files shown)


.github/workflows/avocado.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  avocado:
+    name: "[TEST-IGNORE] Swagger Avocado"
+
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Must include all branches, for git branch logic in Avocado to work correctly
```
weshaggard, line 17: For my information what branches other than the current one does Avocado actually look at?

.github/workflows/avocado.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Run Avocado
+        run: |
+          npm exec --no avocado \
+          --excludePaths \
+            "/common-types/" \
+            "/scenarios/" \
+            "/package.json" \
+            "/package-lock.json" \
+            "/cadl/examples/" \
```
weshaggard, line 31: Do we still have cadl paths?

weshaggard (approved):

.github/workflows/avocado.yml
mikeharder: Avocado has logic similar to BreakingChanges and LintDiff, where it runs the tool before and after the PR, and attempts to exclude (some) errors if they also occur before the PR.

Longer-term, I think this "feature" should be removed, and all Avocado suppressions should be in suppressions.yaml.  Then, Avocado can run the same both locally and in CI.

Short-term, if we want the check to work unchanged, we need to clone all branches, so Avocado can run `git checkout <pr-base-branch>`.  Unlike other checks, Avocado has the code to checkout and change branches baked right into the tool.

.github/workflows/avocado.yml
mikeharder: We do.  Random example:

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/voiceservices/resource-manager/Microsoft.VoiceServices/cadl/examples/Contacts_Get.json

PR #33827, opened 2025-04-09
Fix the @bodyRoot nested with @body issue for scvmm

kazrael2119:
fixes https://github.com/Azure/autorest.typescript/pull/3136

The old way to set `StopVirtualMachineOptions` as optional is not recommanded and we may have [future linter to limit the usage](https://github.com/microsoft/typespec/issues/5993). 

In this pr we updated the spec to an customized template with optional body operation. This change would produce more correct client code but without any REST change.


kazrael2119:
approved for JS as these breakings are introduced by convertion to typespec

kazrael2119:
azp run [TEST-IGNORE] spec-gen-sdk - js - pullrequest

specification/scvmm/ScVmm.Management/custom.tsp
```diff
@@ -0,0 +1,64 @@
+using Http;
```
MaryGao, line 1: no import ?

specification/scvmm/ScVmm.Management/VirtualMachineInstance.tsp
```diff
@@ -100,13 +101,9 @@ interface VirtualMachineInstances {
   list is ArmResourceListByParent<VirtualMachineInstance>;
   /** The operation to power off (stop) a virtual machine instance. */
   @summary("Implements the operation to stop a virtual machine.")
-  stop is ArmResourceActionNoResponseContentAsync<
+  stop is CustomizedArmResourceActionNoResponseContentAsync<
     VirtualMachineInstance,
-    {
-      /** The content of the action request */
-      @body
-      body?: StopVirtualMachineOptions;
```
MaryGao, line 108: @kartikverma020 Feel free to take a look at the change and let me know if any concern!

The old patter to set `StopVirtualMachineOptions` as optional is not recommanded and we may have [future linter to limit the usage](https://github.com/microsoft/typespec/issues/5993). 

In this pr we updated the spec to an customized template with optional body operation. This change would only produce more correct client method but without any REST change.

MaryGao (approved):

weidongxu-microsoft (approved):

weidongxu-microsoft:
We seem not released GA from tsp, hence the possible reason of SDK break.

MaryGao:
> We seem not released GA from tsp, hence the possible reason of SDK break.

@weidongxu-microsoft yeah, this is one side. Another side is that no emitters really support bodyRoot nested body case, so somehow it won't impact generated code. 

JS breaking is introduced by from swagger to typespec. @yuchao why Python has breaking?

weidongxu-microsoft:
> > We seem not released GA from tsp, hence the possible reason of SDK break.
> 
> @weidongxu-microsoft yeah, this is one side. Another side is that no emitters really support bodyRoot nested body case, so somehow it won't impact generated code.
> 
> JS breaking is introduced by from swagger to typespec. @yuchao why Python has breaking?

Latter assertion likely false. Emitter does not need to handle body or bodyRoot, if they follow the http lib.

msyyc (approved):
Looks good to python.

PR #33829, opened 2025-04-09
Another round of updates to /connections, /datasets & /indexes operations

dargilco:
Updates to Connection operations:
- Add a boolean property `isDefault` in the `Connection` object to indicate if the connection is tagged as the default one for its connection type. 
- Add a new `getWithCredentials` operation (route `/connections/{name}/withCredentials`) to get a `Connection` object with credentials populated. Credentials can be the connection's api-key for authentication, or for Application Insights, the Application Insights connection string.
- Update list of Connection Types: Add `AzureStorageAccount` and `AppConfig`.

Dataset operations and Indexes operations:
- Removed the option to create an un-versioned dataset/index. The service is not yet ready to support this. Comment out the relevant sections until the service is ready.


specification/ai/Azure.AI.Projects/connections/models.tsp
```diff
@@ -27,15 +33,85 @@ model Connection {
...
-  authType: AuthenticationType;
+  isDefault: boolean;
 
   @doc("Metadata of the connection")
   @visibility(Lifecycle.Read)
   metadata: Record<string>;
 }
 
+@doc("A connection with API key authentication")
+model ConnectionWithApiKeyAuth extends Connection {
```
needuv, line 46: do we need these strong typed models? Can't we just have base connection with optional credentials?

PR #33937, opened 2025-04-15
[.github] Change exec and git funcs to return {stdout, stderr}

mikeharder:
- Change exec and git funcs to return {stdout: string, stderr: string}
- Update all consumers of exec and git funcs
- Add execNpm() helper to shared
- TypeSpecValidation: Remove host.runFile() and host.runNpm()

### DIFF
--- a/eng/tools/typespec-validation/src/tsv-runner-host.ts
+++ b/eng/tools/typespec-validation/src/tsv-runner-host.ts
@@ -1,40 +1,12 @@
 import { globby, Options as GlobbyOptions } from "globby";
-import { dirname, join } from "path";
 import { simpleGit } from "simple-git";
 import { RuleResult } from "./rule-result.js";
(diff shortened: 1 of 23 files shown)


eng/tools/typespec-validation/src/tsv-runner-host.ts
```diff
@@ -1,42 +1,14 @@
...
 export class TsvRunnerHost implements TsvHost {
   gitOperation(folder: string): IGitOperation {
     return simpleGit(folder);
   }
 
-  runFile(file: string, args: string[], cwd?: string): Promise<[Error | null, string, string]> {
-    return runFile(file, args, cwd);
-  }
-
-  runNpm(args: string[], cwd?: string): Promise<[Error | null, string, string]> {
```
mikeharder, line 17: moved to exec.js

PR #33949, opened 2025-04-15
Update tspconfig with kebab case style

v-jiaodi:
related to https://github.com/Azure/autorest.typescript/issues/3067


MaryGao (approved):
Approved since this is only JS config change.

PR #34034, opened 2025-04-17
API for create agent evaluation run

YusakuNo1:
# Choose a PR Template

# API: Create agent evaluation

Sample request:
`POST /runs:runAgent`
```json
{
  "threadid": "thread_[unique-id]",                    # Optional, after agent API V2 migration from the server side, the user don't need to fill it
  "runid": "run_[unique-id]",
  "evaluators": {
    "similarity": {
      "id": "SimilarityEvaluator"
    },
    "protectedmaterial": {
      "id": "ProtectedMaterialEvaluator"
    }
  },
  "samplingconfiguration": {
    "name": "my-sampling-config",
    "samplingpercent": 10.0,                                # Percentage of sampling per hour (0-100)
    "maxrequestrate": 100                                   # Maximum tokens per hour (0-10000)
  },
  "redactionconfiguration": {
    "redactscoreproperties": true                          # Redact the score properties
  },
  "appinsightsconnectionstring": "[connection-string]"
}
```

Sample response:
`HTTP 201`
```json
{
  "id": "run_[unique-id]",           # For agent API v1, it's completion Id; for agent API v2, it's the run Id
  "status": "Running"
}
```


specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -42,4 +42,30 @@ interface Evaluations {
...
+      @doc("Agent evaluation with asynchronous run")
+      @body
+      evaluation: AgentEvaluationRequest;
+    },
+    AgentLongRunningOperationResponse
+  >;
+
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  @doc("Get agent evaluation run results.")
+  @route("agents/runs/{runId}")
```
priyankamehtani, line 61: should this have results somewhere in the route?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,102 @@ model Evaluation {
...
+
+  @doc("Maximum request rate per hour.")
+  @visibility(Lifecycle.Read)
+  maxRequestRate: float;
+}
+
+@doc("Evaluation request for agent run.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model AgentEvaluationRequest {
```
priyankamehtani, line 102: Can this be merged with the evaluation object? We can add an object for agent specific properties similar to the EvaluationTarget property.

YusakuNo1: It seems like [the EvaluationTarget in 1DP private branch](https://github.com/Azure/azure-rest-api-specs/blob/f2b08267fae6cd19cd52ad76e60fe996e70cf317/specification/ai/Azure.AI.Projects/evaluations/models.tsp#L94) is different? Also in 1DP API public and private branch, it seems like only [blob URI](https://github.com/Azure/azure-rest-api-specs/blob/f2b08267fae6cd19cd52ad76e60fe996e70cf317/specification/ai/Azure.AI.Projects/evaluation-results/models.tsp#L21) is defined, but not the result? If it's defined somewhere, let's sync them

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
YusakuNo1: We reviewed the other design from [this PR](https://github.com/Azure/azure-rest-api-specs/pull/33683/files#diff-3b3b3ce074d90d63892e53c2dc6ee7ea1a111e56ad4455a42130a7121162fd04) for evaluation results, it's not merge and we are not sure `agents/runs/{runId}/results` is good or not... "evaluation run" is more like "evaluation metadata", while the real evaluation result is the most important thing? As this API is temporary, I'm fine with `agents/runs/{runId}/results`

YusakuNo1: Resolve as it's not relevant anymore

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,102 @@ model Evaluation {
   @doc("Evaluators to be used for the evaluation.")
   evaluators: Record<EvaluatorConfiguration>;
 }
+
+@doc("Definition for sampling strategy.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model SamplingConfiguration {
+  @doc("Name of the sampling strategy.")
+  @visibility(Lifecycle.Read)
```
johanste, line 87: It is rare to have a model with only `Lifecycle.Read` properties.

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,102 @@ model Evaluation {
...
+  name: string;
+
+  @doc("Percentage of sampling per hour.")
+  @visibility(Lifecycle.Read)
+  samplingPercent: float;
```
johanste, line 92: What is the range? E.g. does 99.2 mean 99.2% sampling rate or is it an error?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
priyankamehtani: Is this a synchronous API? My understanding was that it was async and the client had to poll for the LRO to complete.

YusakuNo1: It's async, the get API will be private API, only used by our UI, until we have the proper public API

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,48 @@ model Evaluation {
...
+  samplingConfiguration?: SamplingConfiguration;
+
+  @doc("Optional and temporary way to pass the AppInsights connection string to the evaluator. When this string is not null, the evaluation results will be logged to Azure AppInsights.")
+  appInsightsConnectionString?: string;
+}
+
+@doc("Long running operation response for agent evaluation.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model AgentLongRunningOperationResponse {
```
johanste, line 119: Wasn't this supposed to be removed?

johanste, line 119: This is not adhering to Azure's long running operation schema. I thought this was supposed to be removed per our discussion last week?

johanste (approved):

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,94 @@ model Evaluation {
   @doc("Evaluators to be used for the evaluation.")
   evaluators: Record<EvaluatorConfiguration>;
 }
+
+@doc("Definition for sampling strategy.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model SamplingConfiguration {
```
janhavi13: Does SamplingConfiguration apply to only input that comes from app insights? If so, my recommendation is to add a new polymorphic type for Input Data called "traces" or "agent_input" and move this into that polymorphic object,

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,94 @@ model Evaluation {
...
+  samplingPercent: float;
+
+  @doc("Maximum request rate per hour (0 to 1000).")
+  maxRequestRate: float;
+}
+
+@doc("The redaction configuration will allow the user to control what is redacted.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model RedactionConfiguration {
```
janhavi13: This also seems specific to output configuration. Maybe we want to add OutputConfiguration instead?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,94 @@ model Evaluation {
...
+  @doc("Redact score properties. If not specified, the default is to redact in production.")
+  redactScoreProperties?: boolean;
+}
+
+@doc("Evaluation request for agent run.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model AgentEvaluationRequest {
```
janhavi13, line 107: I would probably structure this as close to Evaluation Object even if we want to keep it separate for Build. 
1. Create new input data type object that takes runId, threadId, samplingConfiguration, appInsightConnectionString

janhavi13: Here is an example of how to do it - https://msdata.visualstudio.com/Vienna/_git/vienna?path=/src/azureml-api/src/RAISvc/Contracts/Evaluations/InputData.cs&version=GBmaster&line=28&lineEnd=29&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -78,3 +78,94 @@ model Evaluation {
...
+}
+
+@doc("Evaluation response for agent evaluation run.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v_latest)
+model AgentEvaluation {
+  @doc("Identifier of the agent evaluation run.")
+  id: string;
+
+  @doc("Status of the agent evaluation. Options: Running, Succeeded, Failed.")
```
janhavi13, line 166: I would recommend using the same status as cloud evaluation to make it easy to merge post Build. 
Here is what cloud evaluation uses today  - https://msdata.visualstudio.com/Vienna/_git/vienna?path=/src/azureml-api/src/RunHistory/Contracts/RunStatus.cs&version=GBmaster&_a=contents

janhavi13 (changes requested):

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: Rename `SamplingConfiguration` to `AgentEvaluationSamplingConfiguration`, do you think it's better? From the long term, `AgentEvaluationSamplingConfiguration` gates all the agent related evaluations, which outputs to AppInsights.

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: Renamed `RedactionConfiguration` to `AgentEvaluationRedactionConfiguration`. Let's discuss offline for the naming

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: offline discussion:

user the following status
    public enum RunStatus
    {
        // Running - The job started to run in the compute target.
        // Completed - Job completed successfully. This reflects that both the job itself and output collection states completed successfully
        // Failed - Job failed.
   }

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: Thanks for the comment, let's separate fields by input annd output:
1. Eval inputs: runId, threadId, samplingConfiguration and redactionConfiguration. These fields are specific to agent, and the current request object is call "AgentEvaluationRequest", maybe for the purpose of this API, they are already separated? -- in the future, once we create the base model "EvaluationBase", which is the parent of AgentEvaluation, and then AgentEvaluation will still have these fields.
2. AppInsights. There are some confusion between the old API and this API: for schedule eval, AppInsights is input; for agent eval, AppInsights is output.

For #1, let's discuss from the previous 2 comments; for #2, I suggest to keep appInsightsConnectionString in the class for now, which is simple enough from 1RP point of view?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: Discussed offline, we will cover this by the evaluation API convergence project

specification/ai/Azure.AI.Projects/evaluations/models.tsp
YusakuNo1: Discussed offline, we will cover this by the evaluation API convergence project

janhavi13 (approved):

PR #34065, opened 2025-04-21
[computeschedule] Remove python configuration for swagger

msyyc:
Python SDK is already generated from typespec so no need to keep configuration for swagger.


ChenxiJiang333 (approved):

msyyc:
We need retrigger https://github.com/Azure/azure-rest-api-specs/pull/33742  after this PR merged.

PR #34095, opened 2025-04-21
Repurposed breaking chagne artifact to be spec-gen-sdk artifact

### LINKED ISSUE #10404: Refactor SDK breaking change label artifact
Repurposed the artifact to include more information. The new information could be used to determine if the spec-gen-sdk check is required.

raych1:
This PR repurposed the SDK breaking change artifact to be the spec-gen-sdk artifact, which includes additional information such as spec types in the PR, required settings. This information will be used to determine whether `spec-gen-sdk` check is required or not.

* Renamed the breaking change artifact to `spec-gen-sdk-artifact` across the codebase, including variable names, file paths, and related logic. This ensures consistency and reflects the broader purpose of the artifact. (`.github/workflows/src/sdk-breaking-change-labels.js`, `eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml`, `eng/tools/spec-gen-sdk-runner/src/commands.ts`)

* Added logic to detect and differentiate between management plane and data plane specifications based on file paths. This information is stored in the new `SpecGenSdkArtifactInfo` interface. (`eng/tools/spec-gen-sdk-runner/src/commands.ts`)

* Introduced a new `SpecGenSdkArtifactInfo` interface to encapsulate metadata about the spec-gen-sdk generation process, such as whether the specification belongs to the management or data plane, the programming language, and the label action. (`eng/tools/spec-gen-sdk-runner/src/types.ts`) 

This is a [test run](https://github.com/raych1/azure-rest-api-specs/actions/runs/14581553550) for the SDK breaking change labels workflow, ensuring no impact to it.

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/types.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/types.ts
@@ -29,3 +29,14 @@ export type VsoLogs = Map<
     warnings?: string[];
   }
 >;
+
+/**
+ * Represents the result of the spec-gen-sdk generation process.
+ */
+export interface SpecGenSdkArtifactInfo {
+  managementPlane: boolean;
+  dataPlane: boolean;
+  language?: string;
+  labelAction?: boolean;
+  isSpecGenSdkCheckRequired?: boolean;
+}
--- a/.github/workflows/src/sdk-breaking-change-labels.js
+++ b/.github/workflows/src/sdk-breaking-change-labels.js
@@ -62,7 +62,7 @@ export async function getLabelAndActionImpl({
   let issue_number = NaN;
   let labelAction;
   let labelName = "";
-  const artifactName = "spec-gen-sdk-breaking-change-artifact";
+  const artifactName = "spec-gen-sdk-artifact";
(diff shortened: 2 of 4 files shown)


eng/tools/spec-gen-sdk-runner/src/types.ts
```diff
@@ -29,3 +29,14 @@ export type VsoLogs = Map<
     warnings?: string[];
   }
 >;
+
+/**
+ * Represents the result of the spec-gen-sdk generation process.
+ */
+export interface SpecGenSdkArtifactInfo {
+  managementPlane: boolean;
```
weshaggard, line 37: Should this just be type so there is no chance it can have a conflict where both mgmt and data plane are set to try?

weshaggard (approved):
One suggestion but otherwise looks good.

eng/tools/spec-gen-sdk-runner/src/types.ts
raych1: It's intended to be separated field as a spec PR can include both management plane specs and data plane specs.

PR #34127, opened 2025-04-22
[Do Not Merge] Adding system data changes 

apoorv1107tiwari01:
# Choose a PR Template

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/stable/2024-07-01/networkSecurityPerimeter.json
+++ b/specification/network/resource-manager/Microsoft.Network/stable/2024-07-01/networkSecurityPerimeter.json
@@ -1526,7 +1526,7 @@
       },
       "allOf": [
         {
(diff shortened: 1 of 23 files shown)


specification/network/resource-manager/Microsoft.Network/stable/2024-07-01/networkSecurityPerimeter.json
```diff
@@ -1910,11 +1910,6 @@
           "description": "Resource type."
         }
       },
-      "allOf": [
-        {
-          "$ref": "./network.json#/definitions/Resource"
-        }
-      ],
       "description": "Resource that is onboarded to use network security perimeter. Also referred as perimeter associable resource."
```
mentat9, line 1913: This is a good change: it's removing duplicate definitions. But why not go farther and use v6 common-types ProxyResource definition (and remove the locally defined properties (name, id, type))?

PR #34128, opened 2025-04-22
Add visibility read to response collection properties

JoshLove-msft:
We should confirm that these models will never be used as inputs in the future.

### DIFF
--- a/specification/eventgrid/Azure.Messaging.EventGrid/main.tsp
+++ b/specification/eventgrid/Azure.Messaging.EventGrid/main.tsp
@@ -110,6 +110,7 @@ namespace Microsoft.EventGrid {
   @doc("Details of the Receive operation response.")
   model ReceiveResult {
     @doc("Array of receive responses, one per cloud event.")
+    @visibility(Lifecycle.Read)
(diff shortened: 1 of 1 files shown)


lmazuel (approved):

lmazuel (approved):

weshaggard:
@JoshLove-msft any reason you didn't update the swagger files with this change? The TypeSpec Validation failure is complaining about that.

JoshLove-msft:
> @JoshLove-msft any reason you didn't update the swagger files with this change? The TypeSpec Validation failure is complaining about that.

Oops, forgot about that. Looks like it will get updated in https://github.com/Azure/azure-rest-api-specs/pull/34182.

JoshLove-msft:
> > @JoshLove-msft any reason you didn't update the swagger files with this change? The TypeSpec Validation failure is complaining about that.
> 
> Oops, forgot about that. Looks like it will get updated in #34182.

But part of the reason is the amount of CI stages that are failing but titled with "IGNORE". It makes it easy to miss issues when there are so many false positives.

PR #34163, opened 2025-04-23
Microsoft.Migrate Seperating out operations API

v-anishrao:
# Choose a PR Template


v-anishrao:
## Rationale for adding `Versioning-Approved-BugFix` label 

This previous PR #33791 authored by me had operations API defined in each of the sub services. This is wrong as the RP has only one operations API at the RP level which is used to display all the operations from every sub service. 

This PR simply removes operations defined by every other sub service to a common place attached to the Same API-Version which is accurate. 
Followed an example from here: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/workloads/Workloads.Operations.Management/models.tsp

Moreover, this started causing problems for the other pipelines like ApiDoc which is also fixed with this PR.

srilipta-swain (approved):

PR #34165, opened 2025-04-23
Updated typespec for Java SDK

jayantjha:
# Choose a PR Template

### DIFF
--- a/specification/ai/Azure.AI.Agents/tools/models.tsp
+++ b/specification/ai/Azure.AI.Agents/tools/models.tsp
@@ -332,9 +332,6 @@ model RunStepFileSearchToolCall extends RunStepToolCall {
   @doc("The object type, which is always 'file_search'.")
   type: "file_search";
 
-  @doc("The ID of the tool call. This ID must be referenced when you submit tool outputs.")
-  id: string;
-
   @doc("For now, this is always going to be an empty object.")
   @encodedName("application/json", "file_search")
   fileSearch: RunStepFileSearchToolCallResults;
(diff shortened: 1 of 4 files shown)


specification/ai/Azure.AI.Agents/tools/models.tsp
```diff
@@ -332,9 +332,6 @@ model RunStepFileSearchToolCall extends RunStepToolCall {
   @doc("The object type, which is always 'file_search'.")
   type: "file_search";
 
-  @doc("The ID of the tool call. This ID must be referenced when you submit tool outputs.")
```
jhakulin, line 335: what is the reason for removal?

specification/ai/Azure.AI.Agents/tspconfig.yaml
```diff
@@ -33,6 +33,13 @@ options:
     packageDetails:
       name: "@azure/ai-agents"
     flavor: azure
+  "@azure-tools/typespec-java":
+    package-dir: "azure-ai-agents"
```
jhakulin, line 37: azure-ai-agents-persistent

specification/ai/Azure.AI.Agents/tools/models.tsp
jayantjha: @jhakulin The `RunStepToolCall` already contains an `id` field. Since `RunStepFileSearchToolCall` inherits from `RunStepToolCall`, it creates a duplicate in Java. I didn't think it should create a problem in other languages, but if it does, then probably need to find a way to ignore it for Java

nick863: I have just checked Python and C# and did not found anomalies.
As a remark, it seems C# code generator better cheks syntax. ID in this class seems to be a copy-paste error.

nick863 (approved):

PR #34179, opened 2025-04-23
Added status check for spec-gen-sdk

raych1:
This pull request introduces a new GitHub Actions workflow and enhances the Azure Pipelines SDK generation process with improved artifact handling, status reporting, and breaking change detection. The changes also include updates to the spec-gen-sdk-runner to better track execution results and provide detailed summaries.

* Added a new workflow, `.github/workflows/spec-gen-sdk-status.yml`, to monitor the status of Azure Pipelines SDK generation jobs and update commit statuses accordingly. This workflow uses the `spec-gen-sdk-status` script to fetch artifact data and summarize results.

* Introduced `SkipRenamingArtifact` and updated conditions in `publish-1es-artifact.yml` and `archetype-spec-gen-sdk.yml` to handle failed attempts and prevent unnecessary artifact renaming.
* Enhanced artifact generation logic in `spec-gen-sdk-runner` to include detailed execution results, breaking change detection, and management plane spec tracking. Added support for required checks per SDK language. 

* Added `getAzurePipelineArtifact` function in `.github/workflows/src/artifacts.js` to fetch artifacts from Azure DevOps with retry logic.
* Introduced `writeToActionsSummary` in `.github/workflows/src/github.js` to append markdown summaries to the GitHub Actions summary file.

* Refactored `generateSdkForSpecPr` in `commands.ts` to aggregate execution results, track breaking changes, and include management plane spec information in artifacts. 
* Updated artifact structure in `types.ts` to include execution results and required check settings for supported SDK languages.

@weshaggard here're some test PRs:

1.[successful status](https://github.com/raych1/azure-rest-api-specs/pull/39)
summary page:
image

2.[failed status for mgmt spec PR](https://github.com/raych1/azure-rest-api-specs/pull/41)
image

3.[successful status for data plane spec PR](https://github.com/raych1/azure-rest-api-specs/pull/40)
image


eng/common/pipelines/templates/steps/publish-1es-artifact.yml
```diff
@@ -12,10 +12,11 @@ parameters:
   ArtifactPath: ''
   CustomCondition: true
   SbomEnabled: true
+  SkipRenamingArtifact: false
```
raych1, line 15: This file change will be reverted in this PR and I will create a PR in azure-sdk-tools repo later.

eng/tools/spec-gen-sdk-runner/src/commands.ts
```diff
@@ -497,15 +506,20 @@ function generateArtifact(
       fs.mkdirSync(specGenSdkArtifactAbsoluteFolder, { recursive: true });
     }
     // Write artifact
-    artifactInfo.language = commandInput.sdkLanguage;
-    artifactInfo.labelAction = shouldLabelBreakingChange;
+    const artifactInfo: SpecGenSdkArtifactInfo = {
+      language: commandInput.sdkLanguage,
+      result,
+      labelAction: hasBreakingChange,
+      isSpecGenSdkCheckRequired:
```
raych1, line 513: Making required as long as a spec PR contains management plane specs.

weshaggard: This is fine for now but if the logic gets more involved we should factor it out into its own method.

eng/tools/spec-gen-sdk-runner/src/index.ts
```diff
@@ -24,5 +24,5 @@ export async function main() {
   if (statusCode !== 0) {
     console.log("##vso[task.complete result=Failed;]");
   }
-  exit(statusCode);
+  exit(0);
```
weshaggard, line 27: Why are you changing this to 0 all the time?

eng/tools/spec-gen-sdk-runner/src/types.ts
```diff
@@ -34,9 +34,29 @@ export type VsoLogs = Map<
  * Represents the result of the spec-gen-sdk generation process.
  */
 export interface SpecGenSdkArtifactInfo {
-  managementPlane: boolean;
-  dataPlane: boolean;
-  language?: string;
+  language: string;
+  result: string;
```
weshaggard, line 38: Is result a string or a bool?

raych1: It's a string with values: ["succeeded", "failed" ...]

.github/workflows/spec-gen-sdk-status.yml
```diff
@@ -0,0 +1,33 @@
+name: "[TEST IGNORE] spec-gen-sdk status"
```
JoshLove-msft, line 1: I've noticed that this check is failing in Event Grid specs for a few languages. But the title is [TEST IGNORE]. Are failures expected here?

raych1: You could ignore this check status as this is a newly added check which is in testing phase. However, I noticed event grid typespecs fails in the TSV validation. Do you know what's the issue?

JoshLove-msft: This should be fixed if you sync your branch from main. But regarding these statuses, do they block regular merging?

JoshLove-msft: I also wonder if they could show a warning status instead of a failure if they are still in testing phase. Ideally, we would be able to know if a PR can be merged or not based on the CI checks - otherwise we end up with alert fatigue and can miss errors that we shouldn't be ignoring (as in the case you mention with the TSV validation).

weshaggard: These do not block merging as they are not required checks. Please just ensure the required checks are passing and watch the "Next Steps to Merge" comment for what needs to be done.

.github/workflows/spec-gen-sdk-status.yml
```diff
@@ -0,0 +1,33 @@
...
+  statuses: write
+
+jobs:
+  spec-gen-sdk-status:
+    # Only run this job when the check run is from Azure Pipelines SDK Generation job in the azure-sdk/public(id: 29ec6040-b234-4e31-b139-33dc4287b756) project
+    if: |
+      github.event.check_run.check_suite.app.name == 'Azure Pipelines' &&
+      contains(github.event.check_run.name, 'SDK Generation') &&
+      endsWith(github.event.check_run.external_id, '29ec6040-b234-4e31-b139-33dc4287b756')
+    name: "[TEST IGNORE] spec-gen-sdk - set status"
```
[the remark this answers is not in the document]
raych1: It's to follow the rule of prefixing with it for any new or rewritten checks. I am planning to remove it when we remove it from all the `spec-gen-sdk` checks.

.github/workflows/src/github.js
```diff
@@ -120,3 +120,25 @@ export const CommitStatusState = {
...
   SUCCESS: "success",
 };
+
+/**
+ * Writes content to the GitHub Actions summary
+ * @param {string} content - Markdown content to add to the summary
+ * @param {typeof import("@actions/core")} core - GitHub Actions core library
+ */
+export async function writeToActionsSummary(content, core) {
+  const summaryFilePath = process.env.GITHUB_STEP_SUMMARY;
```
weshaggard, line 130: Why are we using an env to pass this information instead of just passing it as a parameter?

weshaggard: Looks like that is a github concept. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions

Can you not use core.summary method instead?

raych1:
@weshaggard this is the [test run](https://github.com/raych1/azure-rest-api-specs/actions/runs/14671234572/job/41178015295) with the code to list artifact then fetch the failed artifacts.  This is the [spec PR](https://github.com/raych1/azure-rest-api-specs/pull/41)

.github/workflows/src/github.js
raych1: Updated the code to use `core.Summary`.

.github/workflows/src/spec-gen-sdk-status.js
```diff
@@ -0,0 +1,218 @@
...
+  for (const checkRun of checkRuns) {
+    core.info(
+      `Processing check run: ${checkRun.name} (${checkRun.conclusion})`,
+    );
+    const buildInfo = getAdoBuildInfoFromUrl(checkRun.details_url);
+    const ado_project_url = buildInfo.projectUrl;
+    const ado_build_id = buildInfo.buildId;
+    let artifactName = "spec-gen-sdk-artifact";
+    const artifactFileName = `${artifactName}.json`;
+    const result = await getAzurePipelineArtifact({
```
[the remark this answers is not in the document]
weshaggard: https://github.com/Azure/azure-rest-api-specs/pull/34179/files#diff-410f1d9b7cf1ea63bd905c016d27c33f001af44ae4f67162acb8d68d0831391eR87 Looks like that is the case.

weshaggard (approved):

raych1:
TSV validation fails in the `tspconfig.yaml` validation for some services. we should fix them in a separated PR otherwise it would always fail when TSV validate against all the TypeSpecs.

weshaggard:
Yes the TSV all failures are known and @wanlwanl is looking at those.

PR #34182, opened 2025-04-23
[EG] Readonly arrays

l0lawrence:
something happening with tspconfig

### DIFF
--- a/specification/eventgrid/data-plane/Microsoft.EventGrid/stable/2018-01-01/GeneratedSystemEvents.json
+++ b/specification/eventgrid/data-plane/Microsoft.EventGrid/stable/2018-01-01/GeneratedSystemEvents.json
@@ -427,6 +427,7 @@
           "items": {
             "$ref": "#/definitions/AcsChatThreadParticipantProperties"
           },
+          "readOnly": true,
           "x-ms-identifiers": []
         }
       },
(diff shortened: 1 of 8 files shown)


specification/eventgrid/data-plane/Microsoft.EventGrid/stable/2018-01-01/GeneratedSystemEvents.json
```diff
@@ -1349,18 +1352,19 @@
...
           "items": {
             "$ref": "#/definitions/AcsRouterCommunicationError"
           },
+          "readOnly": true,
           "x-ms-identifiers": []
         }
       },
       "required": [
         "innererror",
-        "details"
```
l0lawrence, line 1363: why is this changing names

JoshLove-msft (approved):

specification/eventgrid/data-plane/Microsoft.EventGrid/stable/2018-01-01/GeneratedSystemEvents.json
JoshLove-msft: Maybe we are applying a rename and the new decorator is somehow preventing the rename from working.

JoshLove-msft: Doesn't look like this is being renamed in propertyNameOverride. But the currently generated name is correct. I wonder if the old version was manually modified.

lmazuel (approved):

PR #34198, opened 2025-04-24
Switch AppInsights connection string to be mandatory for SDKs and add…

YusakuNo1:
… top level error field for get agent evaluation results API

# Choose a PR Template

There are 2 fields updated:
1. Update `appInsightsConnectionString` in `AgentEvaluationRequest` from optional to mandatory. As AI Foundry SDKs are using this TypeSpec to generate the SDK, while other than AppInsights, there are no way to log user errors for Build, we want to flip it to mandatory
2. Add optional `error` field in `AgentEvaluation`. There are 3 types of errors for the request, this error is for the (ii) type from the list below:
    1. Immediately failure before sending back to the user, we will send non-200 status code with reason
    2. After the response to the user with 200, the code may still fail, e.g. when talking agent service for chat history, the paginated requests may take very long time, it's not good to combine it with type (i) above
    3. For any evaluator failures, the error will be shown within the array of evaluator results

### DIFF
--- a/specification/ai/Azure.AI.Projects/evaluations/models.tsp
+++ b/specification/ai/Azure.AI.Projects/evaluations/models.tsp
@@ -120,7 +120,7 @@ model AgentEvaluationRequest {
   @doc("Redaction configuration for the evaluation.")
   redactionConfiguration?: AgentEvaluationRedactionConfiguration;
 
-  @doc("Optional and temporary way to pass the AppInsights connection string to the evaluator. When this string is not null, the evaluation results will be logged to Azure AppInsights.")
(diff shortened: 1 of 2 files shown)


specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -121,7 +121,7 @@ model AgentEvaluationRequest {
   redactionConfiguration?: AgentEvaluationRedactionConfiguration;
 
   @doc("Optional and temporary way to pass the AppInsights connection string to the evaluator. When this string is not null, the evaluation results will be logged to Azure AppInsights.")
```
janhavi13, line 123: Update this comment to say it is required instead of optional?

janhavi13 (approved):

PR #34201, opened 2025-04-24
fix mismatch Versions for IotFirmwareDefense.Management

kazrael2119:


weidongxu-microsoft (approved):

specification/fist/IotFirmwareDefense.Management/main.tsp
```diff
@@ -35,5 +35,5 @@ enum Versions {
    */
   @useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
   @useDependency(Azure.Core.Versions.v1_0_Preview_1)
-  v2024_08_07: "2025-04-01-preview",
+  v2025_04_01_preview: "2025-04-01-preview",
```
MaryGao, line 38: Fix the issue that the version enum key and value is in-correct, @TheDonLuke let me know if any concern!

weidongxu-microsoft:
I guess we now need .NET dev to add their config.

specification/fist/IotFirmwareDefense.Management/main.tsp
TheDonLuke: No there shouldn't be any concern, it was oversight, thank you for addressing it!

MaryGao (approved):

PR #34259, opened 2025-04-25
Add support for dual TypeSpec directory validation during transition to new structure

mario-guerra:
## Background
As we're transitioning to a new directory structure, we need the tool to support validating both old and new structures during the migration period to ensure a smooth transition.

## Changes
This PR modifies the TypeSpec validation tool to:

1. Add command line options to control which directory structure validation runs:
   - `--validate-old-structure` (shorthand: `-o`) - Controls validation of the old structure
   - `--validate-new-structure` (shorthand: `-n`) - Controls validation of the new structure
   - Both options default to `true` to enable dual validation

2. Add detailed logging to help diagnose validation issues
   - Logs when validation starts/finishes
   - Shows which directories are being checked
   - Records file and directory existence checks
   - Clearly marks validation success/failure at each step

3. Implement special handling during the transition period:
   - When both validations are enabled, the overall validation passes if either structure is valid
   - This enables teams to migrate at their own pace without breaking validation checks

## Testing
The new rule checking has been tested on both old and new directory structures to ensure compatibility during the transition period. In theory at least, it should be a drop-in replacement for the existing checker.


mario-guerra:
This is a work in progress. The rule checking works when I run it manually against directories representing old and new structure, but I need to work on the test to get the mocks set up correctly.

mario-guerra:
I believe this PR is now ready for review.

eng/tools/typespec-validation/src/index.ts
```diff
@@ -18,11 +18,27 @@ export async function main() {
       type: "string",
       short: "f",
     },
+    validateOldStructure: {
```
mikeharder, line 21: I think whether a particular spec is using the "old" or "new" structure will need to be configuration in the repo itself.  For example, TSV could default to requiring the new structure, but use the suppressions mechanism to "configure" the FolderStructure rule.  Something like:

```
suppressions.yaml
- tool: TypeSpecValidation
  rule: FolderStructure
  sub-rule: FolderStructureV2
  reason: Not ready to migrate to TypeSpec folder structure v2
```

The rule would then treat "suppress v2" as "enforce v1".

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -8,61 +9,180 @@ import { RuleResult } from "../rule-result.js";
 import { Rule } from "../rule.js";
 import { fileExists, normalizePath, readTspConfig } from "../utils.js";
 
+// Use promisify as a workaround for fs/promises compatibility
+const readFile = promisify(fs.readFile);
```
mikeharder, line 13: what issue with fs/promises needs a workaround?

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -125,16 +255,13 @@ export class FolderStructureRule implements Rule {
...
-
-      // The path specified in the import must either start with "./" or "../", or be an absolute path.
-      // The path should either point to a directory, or have an extension of either ".tsp" or ".js".
-      // https://typespec.io/docs/language-basics/imports/
-      //
-      // We don't bother checking if the path has an extension of ".tsp" or ".js", because a directory
-      // is also valid, and a directory could be named anything.  We only care if the path is under
-      // $teamFolder, so we just treat anything that looks like a relative or absolute path,
-      // as a path.
+      const matches: string[] = [];
```
mikeharder, line 258: why did you change this code?

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -155,10 +285,719 @@ export class FolderStructureRule implements Rule {
...
+
+    // Check if this is the root of the RPNamespace
+    const isRpNamespaceRoot = folderStruct.length === 2;
+    log(`[New] Is RP namespace root? ${isRpNamespaceRoot}`);
+
+    if (isRpNamespaceRoot) {
+      log(`[New] Validating RP namespace root folder`);
+      // Validate the RP namespace root folder
+      // Check for cspell.yaml
+      const cspellExists = await fileExists(path.join(folder, "cspell.yaml"));
```
mikeharder, line 334: I think cspell.yaml should be optional, for example if you don't need any suppressions?

eng/tools/typespec-validation/src/rules/folder-structure.ts
mikeharder, line 1: This is a lot of custom code to write, maintain, and test.  I will try to think of a more succinct way to express this, perhaps a tiny DSL.

eng/tools/typespec-validation/src/index.ts
mario-guerra: Yep, we could do it this way and I think it makes sense. Does this need more discussion with other people in order for us to decide to go this route, or can we assume "this is the way"?

eng/tools/typespec-validation/src/rules/folder-structure.ts
mario-guerra: That's an artifact of a previous iteration, removing it.

PR #34270, opened 2025-04-27
Remove un-necessary configurations for typespec-ts

v-jiaodi:
The pr is mainly to update un-necessary configurations for typespec-ts.

Since JS emitters properly handles the default value for `generate-sample` and `generate-test` so we don't need to set it explicitly in configurations. Please note [the relevant templates for DPG and MPG are already updated](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L29-L35).

We'd like to clean up the existing configurations so that we could keep them up-to-date.

resolved https://github.com/Azure/typespec-azure/pull/2541/files#r2057393146


specification/applicationinsights/ApplicationInsights.LiveMetrics/tspconfig.yaml
```diff
@@ -28,11 +28,9 @@ options:
   "@azure-tools/typespec-ts":
     package-dir: "livemetrics-rest"
     generate-metadata: true
-    generate-test: true
     package-details:
-      name: "@azure-rest/livemetrics-rest"
```
MaryGao, line 33: can we do this?

MaryGao (approved):

MaryGao:
Applied the options with latest templates https://github.com/Azure/azure-rest-api-specs/pull/34550

MaryGao:
Approved for Python breaking since this is false alarm.

tadelesh (approved):

specification/cognitiveservices/OpenAI.Inference/tspconfig.yaml
```diff
@@ -37,7 +37,6 @@ options:
   "@azure-tools/typespec-ts":
     package-dir: "openai"
     generate-metadata: false
-    generate-test: false
     is-modular-library: true
```
MaryGao, line 40:
```suggestion
    generate-test: false
    is-modular-library: true
```

specification/ai/Face/tspconfig.yaml
```diff
@@ -29,7 +29,6 @@ options:
   "@azure-tools/typespec-ts":
     package-dir: "ai-vision-face-rest"
     generate-metadata: true
-    generate-test: true
```
maririos, line 32: (same but tests) 
what does removing the property does? does the emitter by default generates the tests?

MaryGao: yes, exactly. emitters handles the default with more properly
- `generate-test` would be `true` by default if emitter detects there is no test folder; if already having test folders this would be false.

MaryGao (approved):

maririos (approved):

lirenhe:
SDK owners already checked the SDK failures and those are not related to this PR.

PR #34278, opened 2025-04-28
[TSP Migration][hardwaresecuritymodules] TypeSpec migrated from swagger

melina5656:


specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,248 @@
...
+    CloudHsmCluster,
+    {
+      /**
+       * Azure storage Resource Uri
+       */
+      @bodyRoot
+      backupRequestProperties?: BackupRequestProperties;
+    },
+    ArmResponse<BackupResult>,
+    LroHeaders = ArmAsyncOperationHeader &
```
JiaqiZhang-Dev, line 125: Should here need to specify `FinalResult`?  same with `validateRestoreProperties`, `restore`, `validateBackupProperties` operations

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/tspconfig.yaml
```diff
@@ -0,0 +1,50 @@
...
+    generate-sample: true
+    flavor: "azure"
+  "@azure-tools/typespec-java":
+    package-dir: "azure-resourcemanager-hardwaresecuritymodules"
+    namespace: "com.azure.resourcemanager.hardwaresecuritymodules"
+    service-name: "Hardwaresecuritymodules" # human-readable service name, whitespace allowed
+    flavor: azure
+  "@azure-tools/typespec-ts":
+    package-dir: "arm-hardwaresecuritymodules"
+    is-modular-library: true
```
v-jiaodi, line 33:
```suggestion
   typespec-title-map:
      HardwareSecurityModulesClient: AzureDedicatedHSMResourceProvider
```

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,240 @@
...
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+
+namespace Microsoft.HardwareSecurityModules;
+/**
+ * Resource information with extended details.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property"
+model CloudHsmCluster
+  is Azure.ResourceManager.TrackedResource<CloudHsmClusterProperties> {
```
catalinaperalta: Why is the suppression over the entire model? It'll suppress this issue over every new property added. Should we be defining this in a different way?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,240 @@
...
+  >;
+
+  /**
+   * Gets the private link resources supported for the Cloud Hsm Cluster.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @get
+  @action("privateLinkResources")
+  @operationId("CloudHsmClusterPrivateLinkResources_ListByCloudHsmCluster")
+  listByCloudHsmCluster is ArmResourceActionSync<
```
catalinaperalta: Why isnt this operation in a CloudHsmClusterPrivateLinkResources interface?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,240 @@
...
+
+  /**
+   * Gets the backup operation status of the specified Cloud HSM Cluster
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-operation"
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header"
+  @operationId("CloudHsmClusterBackupStatus_Get")
+  @get
+  cloudHsmClusterBackupStatusGet(
```
catalinaperalta: Same question and it applies to other operations that needed the operationId decorator, why aren't they in an interface that corresponds to the name in the prefix of the operationId? In this case I would expect this operation to be in a CloudHsmClusterBackupStatus interface. Mostly calling this out because it may change the way operations are generated downstream in other language emitters unless there were transforms

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,240 @@
...
+    ...ResourceInstanceParameters<
+      Resource,
+      Azure.ResourceManager.Foundations.DefaultBaseParameters<CloudHsmCluster>
+    >,
+
+    #suppress "@azure-tools/typespec-azure-core/documentation-required"
```
catalinaperalta: We should add missing docs instead of suppressing them

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,143 @@
...
+   * Properties of the dedicated HSM
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property"
+  @Azure.ResourceManager.Private.conditionalClientFlatten
+  @Azure.ResourceManager.Private.armResourcePropertiesOptionality(false)
+  properties: DedicatedHsmProperties;
+}
+
+@armResourceOperations
+interface DedicatedHsms {
```
catalinaperalta, line 44: Nit: Should this interface name match the prefix used in the operationIds of the routes within it?
```suggestion
interface DedicatedHsm {
```

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,143 @@
...
+  sku: Sku;
+
+  ...Azure.ResourceManager.AvailabilityZonesProperty;
+
+  /**
+   * Properties of the dedicated HSM
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property"
+  @Azure.ResourceManager.Private.conditionalClientFlatten
+  @Azure.ResourceManager.Private.armResourcePropertiesOptionality(false)
```
catalinaperalta: We should not be using elements from the Private namespaces, please remove or find an equivalent decorator. Otherwise, let's open issues to discuss if there's any missing functionality in the existing arm library

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,143 @@
...
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("DedicatedHsm_ListBySubscription")
+  listBySubscription is ArmListBySubscription<
+    DedicatedHsm,
+    Parameters = {
+      /**
+       * Maximum number of results to return.
+       */
+      @query("$top")
+      $top?: int32;
```
catalinaperalta, line 107: Nit: since we already have the wire name in the query decorator, why not just leave this with a good param name in typespec?
```suggestion
      top?: int32;
```

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,853 @@
...
+
+using TypeSpec.Rest;
+using TypeSpec.Http;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Foundations;
+
+namespace Microsoft.HardwareSecurityModules;
+
+@@nextLink(CommonTypes.PrivateLinkResourceListResult.nextLink);
+@@Azure.Core.items(CommonTypes.PrivateLinkResourceListResult.value);
```
catalinaperalta, line 14: Why arent these decorators directly over those properties?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,853 @@
...
+
+/**
+ * State of security domain activation
+ */
+union ActivationState {
+  string,
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "For backward compatibility"
```
catalinaperalta: Nit: instead of suppressing documentation we should take this opportunity to add missing docs

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
pshao25: Yes, you are right. @melina5656 Please remove this suppression.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
pshao25: Because this operation operates on the resource CloudHsmCluster.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
pshao25: Because we group operations by resource, not by operation group. The downstream issue will be handled by https://github.com/Azure/typespec-azure/issues/2683

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
pshao25: Maybe, @melina5656 check if all the operations in this interface have an operationid starts with `DedicatedHsm_` , if so, try to rename to check if there is name conflict.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
pshao25: We can. It just cannot be automatically handled by converter because its original name is `$top`.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
pshao25: Because this model is from library. I've no ideas why this model doesn't have nextlink and value. Will talk to Mark and Ali.

MattGertz (approved):
Approving on behalf of conversion team so service team can continue work.

emmeliaAra:
Hi @mcgallan, I am running tsp compile .  on this branch but it fails with these errors: 
I saw the changes made in this commit: https://github.com/Azure/azure-rest-api-specs/pull/34278/commits/29f6a685113e55abd157d9e87c13e9614efe28e0 
Can you please help to resolve ? I am temporarily commenting these lines so that I can continue working on the validation errors of the PR 
image

mcgallan:
> Hi @mcgallan, I am running tsp compile . on this branch but it fails with these errors: I saw the changes made in this commit: [29f6a68](https://github.com/Azure/azure-rest-api-specs/commit/29f6a685113e55abd157d9e87c13e9614efe28e0) Can you please help to resolve ? I am temporarily commenting these lines so that I can continue working on the validation errors of the PR image

Hi @emmeliaAra The @clientLocation decorator is newly added in "@azure-tools/typespec-client-generator-core": "0.57.1". The modifications I made have been verified locally without any issues. I believe you can avoid this problem by pulling the latest changes first and then running npm install to ensure the version is 0.57.1.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,142 @@
...
+  get is ArmResourceRead<DedicatedHsm, Error = DedicatedHsmError>;
+
+  /**
+   * Create or Update a dedicated HSM in the specified subscription.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @operationId("DedicatedHsm_CreateOrUpdate")
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<
+    DedicatedHsm,
+    LroHeaders = ArmLroLocationHeader &
```
tadelesh: Lose final result here.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
TimLovellSmith: +1

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
```diff
@@ -0,0 +1,259 @@
...
+    jobId: string,
+  ): (BackupResult &
+    Azure.Core.RequestIdResponseHeader) | ArmAcceptedResponse<ExtraHeaders = ArmLroLocationHeader &
+    Azure.Core.RequestIdResponseHeader> | ErrorResponse;
+  /**
+   * Gets the restore operation status of the specified Cloud HSM Cluster
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-operation" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
TimLovellSmith, line 199: Why do we need these suppressions on a 'get'?

Is this something particular to do with it being a get of *operation status* particularly?

Does this make them *false positives*? Or is there something weird about how we modeling the operation status part of the API? What would be normal?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,142 @@
...
+using TypeSpec.OpenAPI;
+
+namespace Microsoft.HardwareSecurityModules;
+/**
+ * Resource information with extended details.
+ */
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-core/no-private-usage" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+@Azure.ResourceManager.Private.armResourceInternal(DedicatedHsmProperties)
+@TypeSpec.Http.Private.includeInapplicableMetadataInPayload(false)
```
TimLovellSmith, line 19: >includeInapplicableMetadataInPayload [](http://example.com/codeflow?start=23&length=36)

This is super confusing, why would we need includeInapplicableMetadataInPayload false? Justifications need to be provided, who is providing them?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
```diff
@@ -0,0 +1,142 @@
...
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @operationId("DedicatedHsm_Delete")
+  delete is ArmResourceDeleteWithoutOkAsync<
+    DedicatedHsm,
+    Error = DedicatedHsmError
+  >;
+
+  /**
+   * The List operation gets information about the dedicated hsms associated with the subscription and within the specified resource group.
```
TimLovellSmith: >hsms [](http://example.com/codeflow?start=61&length=4)

nit comment should say 'HSMs' (capitalization)

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
TimLovellSmith: It should definitely have 'value' - some of the APIs don't declare nextLink today... but I think probably really all list apis should.

[https://learn.microsoft.com/en-us/rest/api/azureml/private-link-resources/list?view=rest-azureml-2025-06-01&tabs=HTTP#privatelinkresourcelistresult](https://learn.microsoft.com/en-us/rest/api/azureml/private-link-resources/list?view=rest-azureml-2025-06-01&tabs=HTTP#privatelinkresourcelistresult)

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+   */
+  @visibility(Lifecycle.Read)
+  provisioningState?: PrivateEndpointConnectionProvisioningState;
+
+  /**
+   * The group ids for the private endpoint resource.
+   */
+  @visibility(Lifecycle.Read)
+  groupIds?: string[];
+}
```
TimLovellSmith, line 387: Private endpoint connections being a standard contract across many RPs, should be modeled with a common types contract - can we do it?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+   * The reason for approval/rejection of the connection.
+   */
+  description?: string;
+
+  /**
+   * A message indicating if changes on the service provider require any updates on the consumer.
+   */
+  actionsRequired?: string;
+}
+
```
TimLovellSmith: Private endpoint connections being a standard contract across many RPs, should be modeled with a common types contract - can we do it?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+  @visibility(Lifecycle.Read)
+  type?: string;
+
+  /**
+   * Azure Resource Manager metadata containing createdBy and modifiedBy information.
+   */
+  @visibility(Lifecycle.Read)
+  systemData?: SystemData;
+}
+
```
TimLovellSmith, line 428: @MattGertz @rkmanda @raosuhas

Redeclaring such a custom 'Resource' type in typespec (or anything in common types, but Resource most obviously) really should need a *justification* or suppression. It shouldn't just be allowed by default.

Which raises the question to me, why don't we get a linter error for this? Seems like an obvious gap! 

sort of but not really [ARMBlockingComment]

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+   * The private link resource required member names.
+   */
+  @visibility(Lifecycle.Read)
+  requiredMembers?: string[];
+
+  /**
+   * The private link resource private link DNS zone name.
+   */
+  requiredZoneNames?: string[];
+}
```
TimLovellSmith: Private endpoint connections being a standard contract across many RPs, should be modeled with a common types contract - can we do it?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+   */
+  @visibility(Lifecycle.Read)
+  message?: string;
+
+  /**
+   * Contains more specific error that narrows down the cause. May be null.
+   */
+  @visibility(Lifecycle.Read)
+  innererror?: Error;
+}
```
TimLovellSmith, line 693: Is this supposed to correspond to ErrorResponse in common types? Or specifically be custom?

Can we maybe also tighten up typescript linting to require justifications for defining custom error types like this?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+model EndpointDetail {
+  /**
+   * An IP Address that Domain Name currently resolves to.
+   */
+  ipAddress?: string;
+
+  /**
+   * The port an endpoint is connected to.
+   */
+  port?: int32;
```
TimLovellSmith, line 758: >int32 [](http://example.com/codeflow?start=9&length=5)

nit, tighter doc:
Consider possibly documenting minValue 0, maxValue 65535
since its a port number

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+
+  /**
+   * The protocol used for connection
+   */
+  protocol?: string;
```
TimLovellSmith, line 763: >string [](http://example.com/codeflow?start=13&length=6)

nit, is protocol treated as an enum in practice? what values are allowed?

Improve docs or typing

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
```diff
@@ -0,0 +1,851 @@
...
+   */
+  @visibility(Lifecycle.Read)
+  operation?: string;
+
+  /**
+   * The short, localized friendly description of the operation; suitable for tool tips and detailed views.
+   */
+  @visibility(Lifecycle.Read)
+  description?: string;
+}
```
TimLovellSmith: is this the operations metadata contract? Again I don't think we want RPs redeclaring it in their typespecs, what can we do about that?

tadelesh:
The `id` property of `APIEntityReference` and `NetworkInterface` has been changed to `resourceID` since last stable version. Is it by design?

emmeliaAra:
> The `id` property of `APIEntityReference` and `NetworkInterface` has been changed to `resourceID` since last stable version. Is it by design?

As discussed offline we rename the property to resolve a LintDiff validation. this change is already approved from previous PRs.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
emmeliaAra: not sure why this was added by the typespec team but I tried to remove it and check the difference and a new definition DedicatedHsmCreateOrUpdate was created. This definition was passed used as the input parameter for the put Api

image

msyyc:
azure-mgmt-hardwaresecuritymodules is not GAed yet, so the breakings are accepable for python.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
pshao25: It's just its properties type "PrivateLinkResourceProperties" is not the same as that in common type.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: This is used only for Dedicated hsm Service. This is an older service and the validations were introduced after the model was created for this service. The error code model of the service does not match any of the ErrorCode descriptions from the common types. Making a change now will result in a breaking change for the service. Considering that this service will be announced for deprecation soon (once Cloud hsm service goes to GA) we believe it is not worth the engineering effort.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: From previous swagger, this is not an enum

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
emmeliaAra: Removing the supressions creates raises some warnings. Adding @melina5656 to help with this

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: PrivateEndpointConnections property is using all the common types except for PrivateEndpointConnectionProvisioningState. the common types for that definition has these 4 states: "Succeeded", "Creating", "Deleting", "Failed" and from our RP we also return some additional states

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: removed PrivateLinkServiceConnectionState from models since it is not used. In private endpoint connection properties we are using PrivateLinkServiceConnectionState from the common types
image

After compilation no errors and not changes in the openapi.json 

@melina5656  can you please confirm this ?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: Private Link Resource and Private Link Resource Properties in not used.  we are using PrivateLinkResourceListResult from the common types
image
After compilation no errors and not changes in the openapi.json 

@melina5656  can you please confirm this ?

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: updated

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
emmeliaAra: This is again in not used. After compilation no errors and not changes in the openapi.json

@melina5656 can you please confirm this ?

emmeliaAra:
Hi @TimLovellSmith, thanks for reviewing. I've made the changes. Could you please review again ?

emmeliaAra:
Hi @mcgallan can you please check the changes in this commit ? https://github.com/Azure/azure-rest-api-specs/commit/1e492e9f0e4ca6c87c4f500a233869aad2d973d6

After this commit the .NET SKD pipeline change started to fail https://github.com/Azure/azure-rest-api-specs/pull/34278/checks?check_run_id=45042667044

emmeliaAra (approved):

pshao25 (approved):

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/CloudHsmCluster.tsp
pshao25: As we synced, we don't do any improvement on the existing spec.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/DedicatedHsm.tsp
pshao25: This is handled by `@clientLocation`

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
pshao25: There was a bug previously, so we have to add these. But I think we could remove it now. @melina5656 please remove it.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
pshao25: @emmeliaAra We just honor what you defined in the original swagger. You could go ahead to take any improvements that you think makes sense.

specification/hardwaresecuritymodules/HardwareSecurityModules.Management/models.tsp
pshao25: @emmeliaAra Same as above.

PR #34286, opened 2025-04-28
Change namespace to managedNamespace

purpleran7:
Original api doc: https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/747319/Namespace-as-Service

Doc explain why we're making this renaming:
https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/personalplayground/776319/'Managed-Namespace'-ARM-resource-type-rename


purpleran7:
I'm trying to created on the " to open a PR for a Control Plane (ARM) API". But its not working...

matthchr (approved):

PR #34363, opened 2025-04-30
[EngSys] Fix tsconfigs to prevent overwrite errors

### LINKED ISSUE #34122: [EngSys] TS5055: Cannot write file 'index.d.ts' because it would overwrite input file.
If you build the repo multiple times, you might get this error:

```
npm error error TS5055: Cannot write file 'eng/tools/typespec-validation/dist/src/index.d.ts' because it would overwrite input file.
```

I suspect this is a small bug in our tsconfig files.

mikeharder:
- fixes #34122

### DIFF
--- a/eng/tools/tsconfig.json
+++ b/eng/tools/tsconfig.json
@@ -7,9 +7,14 @@
     // override "importHelpers:true" in root tsconfig.json
     "importHelpers": false,
 
-    // required to use project references
-    "composite": true,
+    // base isn't a build unit itself
+    "composite": false,
+
+    // required to reference @azure-tools/specs-shared which doesn't declare types
+    "allowJs": true,
   },
+  // don't compile anything here
+  "files": [],
   "references": [
     { "path": "./specs-model" },
     { "path": "./suppressions" },
--- a/eng/tools/tsp-client-tests/tsconfig.json
+++ b/eng/tools/tsp-client-tests/tsconfig.json
@@ -1,6 +1,12 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./dist",
-  }
+    "composite": true,
+    "outDir": "dist",
+    "rootDir": "."
+  },
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
+  ]
 }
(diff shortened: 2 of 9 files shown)


eng/tools/tsconfig.json
```diff
@@ -7,9 +7,14 @@
...
-    // required to use project references
-    "composite": true,
+    // base isn't a build unit itself
+    "composite": false,
+
+    // required to reference @azure-tools/specs-shared which doesn't declare types
+    "allowJs": true,
   },
+  // don't compile anything here
+  "files": [],
```
mikeharder, line 17: I believe some errors were caused by files being compiled by both the root and leaf projects.

eng/tools/tsconfig.json
```diff
@@ -7,9 +7,14 @@
     // override "importHelpers:true" in root tsconfig.json
     "importHelpers": false,
 
-    // required to use project references
-    "composite": true,
+    // base isn't a build unit itself
```
mikeharder, line 10: I believe some errors were caused by files being compiled by both the root and leaf projects.

eng/tools/tsconfig.json
```diff
@@ -7,9 +7,14 @@
...
+    "composite": false,
```
mikeharder, line 11: I believe some errors were caused by files being compiled by both the root and leaf projects.

eng/tools/tsp-client-tests/tsconfig.json
```diff
@@ -1,6 +1,12 @@
...
-    "outDir": "./dist",
-  }
+    "composite": true,
+    "outDir": "dist",
+    "rootDir": "."
+  },
+  "include": [
+    "src/**/*.ts",
+    "test/**/*.ts"
+  ]
```
mikeharder, line 11: All sub-projects should include these configs by default.  They cannot be set in the base config, because paths are relative to tsconfig.json.

eng/tools/tsp-client-tests/tsconfig.json
```diff
@@ -1,6 +1,12 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "outDir": "./dist",
-  }
+    "composite": true,
```
mikeharder, line 4: Must be set in sub-projects, but should not be set in default, to avoid potential double-compilation.

PR #34365, opened 2025-04-30
[EngSys] npm update

mikeharder:


.github/shared/package.json
```diff
@@ -27,6 +27,7 @@
   "devDependencies": {
     "@eslint/js": "^9.22.0",
     "@tsconfig/node20": "^20.1.4",
+    "@types/debug": "^4.1.12",
```
mikeharder, line 30: was always missing, but masked somehow

PR #34368, opened 2025-04-30
Add depedency @typespec-migration/swagger-compare

pshao25:
Compare two sets of swaggers: 1) original swaggers before we do TypeSpec migration, 2) generated swagger from TypeSpec.

I want to use this tool to give user a better visualized experience when they do the comparison.


pshao25:
Close in favor of https://github.com/Azure/azure-rest-api-specs/pull/34438

PR #34435, opened 2025-05-04
Release metricalerts microsoft.insights preview/2024 03 01-preview 

yashmuel:
# Choose a PR Template


specification/monitor/resource-manager/readme.md
```diff
@@ -32,12 +32,20 @@ title: MonitorClient
 description: Monitor Management Client
 openapi-type: arm
 openapi-subtype: rpaas
-tag: package-2025-01-01-preview
+tag: package-preview-2024-03-01-preview
```
yashmuel, line 35: what is this change?

specification/monitor/resource-manager/Microsoft.Insights/preview/2024-03-01-preview/metricAlert_API.json
```diff
@@ -0,0 +1,1263 @@
...
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "schema": {
+              "$ref": "#/definitions/MetricAlertResource"
+            }
+          },
+          "default": {
+            "description": "BadRequest",
```
rkmanda, line 293: This defines the error contract for all kinds of errors, not just for bad requests

rkmanda:
why is a 2024 March version being released now?

rkmanda:
> why is a 2024 March version being released now?

please change the version string to a latest value closer to the current date to release this

yashmuel:
> > why is a 2024 March version being released now?
> 
> please change the version string to a latest value closer to the current date to release this

We can't the version is already in our code, we've started developing it a while ago.

yashmuel:
@rkmanda We're trying to change from swagger 2.0 to openApi 3.0.0, but the validation expects swagger 2.0 and it fails on:
Missing required property: swagger

is there a way to change the validation to run the proper validation for openApi instead of Oav?

mentat9:
@yashmuel - Your PR is not ready for ARM review. Please address these issues first:
1. Select and fill out the correct template at the top.
2. Address these failing checks: Breaking Change(Cross-Version), Swagger ModelValidation, Swagger SemanticValidation.

yashmuel:
> @yashmuel - Your PR is not ready for ARM review. Please address these issues first:
> 
> 1. Select and fill out the correct template at the top.
> 2. Address these failing checks: Breaking Change(Cross-Version), Swagger ModelValidation, Swagger SemanticValidation.

@mentat9 

The reason those validations are failing is because we have to use OpenApi3 instead of swagger 2.0.
Is there a way to make the PR run the validations for OpenApi3?

The only other option we have is to generate typespec that would require significant effort on our end, and since the controllers already exist and the feature is in private preview, we wouldn’t gain any value from using TypeSpec at this stage.

PR #34438, opened 2025-05-05
Add package to validate TypeSpec migration result

pshao25:
As titled. The user story will be updated in https://github.com/Azure/typespec-azure/pull/2593


mikeharder:
Also add the new project to this file:

https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/tsconfig.json

eng/tools/typespec-migration-validation/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "@azure-tools/typespec-migration-validation",
+  "private": true,
+  "type": "module",
+  "main": "dist/src/index.js",
+  "bin": {
+    "tsmv": "cmd/tsmv.js"
+  },
+  "dependencies": {
+    "@azure-tools/typespec-autorest": ">=0.44.0 <1.0.0",
+    "fs": "^0.0.1-security",
```
mikeharder, line 11:
```suggestion
```

I think this is a bug an you intend to use `fs` from the node standard library.

eng/tools/typespec-migration-validation/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "devDependencies": {
+    "@types/jest": "^29.4.0",
+    "@types/json-diff": "^1.0.3",
+    "@types/node": "^18.19.86",
+    "@types/yargs": "^17.0.33",
+    "@typescript-eslint/eslint-plugin": "^5.54.0",
+    "@typescript-eslint/parser": "^5.54.0",
+    "eslint": "^8.35.0",
+    "jest": "^29.5.0",
```
mikeharder, line 23: convert to use `vite` like the other folders under eng/tools

eng/tools/typespec-migration-validation/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "ts-jest": "^29.0.5",
+    "ts-node": "^10.9.1",
```
mikeharder, line 25:
```suggestion
```

I don't think you need this dep.

eng/tools/typespec-migration-validation/scripts/download-main.ps1
mikeharder, line 1: Why is this written in powershell instead of JS?  I think it's more complex to mix languages like this.

pshao25: It's like a pre step before we could run this package. I agree it's better to rewrite it by ts. But considering our limited time, we prefer to keep it as it is.

eng/tools/typespec-migration-validation/package.json
pshao25: I deleted jest. When we add tests I'll add vite.

mikeharder (approved):
Approving this PR, but in a future PR please add at least a "hello world" test and a github workflow to ensure your package at least builds successfully.  You can copy this workflow:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/lintdiff-test.yaml

This is the full PR which added LintDiff:

https://github.com/Azure/azure-rest-api-specs/pull/32697

PR #34482, opened 2025-05-06
Upgrade typespec 1.0

timotheeguerin:


AlitzelMendez (approved):

markcowl (approved):

markcowl (changes requested):
I am a little worried about the change of additionalProperties.allOf.$ref to $ref,  the spectral linter has issues with sibling $ref properties, if I'm not mistaken, and we would not want this to result in API validation errors.

markcowl (approved):
For ARM, the changes are only for userAssignedIdentity in PATCH, and in the common-types has an x-nullable that is a sibling of $ref here: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/managedidentity.json#L90-L92

markcowl (approved):

AlitzelMendez (approved):

PR #34496, opened 2025-05-07
[TSP Migration][maintenance] TypeSpec migrated from swagger

melina5656:

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind. Here is the list of your anti-pattern:
- The model definition for pageable operation response should have a `nextLink` property.
- `resourcegroups` segment in your path will be normalized to `resourceGroups`


specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
```diff
@@ -0,0 +1,144 @@
...
+         * Resource type
+         */
+        @path resourceType: string,
+
+        /**
+         * Resource name
+         */
+        @path resourceName: string,
+
+        ...Azure.ResourceManager.Legacy.Provider,
```
markcowl: This seems like it's in the wrong place

specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
```diff
@@ -0,0 +1,144 @@
...
+
+        /**
+         * Resource provider name
+         */
+        @path @segment("providers") providerName: string,
+
+        /**
+         * Resource parent type
+         */
+        @path resourceParentType: string,
```
markcowl: This is a really problematic pattern - it allows the extension to apply to any resource that is the direct child of a resource group resource.  It disallows any more deeply nexted resource, or a resource homed in the subscription or location.

IMO, we should consider correcting this API to use Scope - at the very least, we should add a set of scope-based operations

specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
```diff
@@ -0,0 +1,144 @@
...
+  extends Azure.ResourceManager.Legacy.CustomLegacyOperations<
+      {
+        ...ApiVersionParameter,
+        ...SubscriptionIdParameter,
+        ...ResourceGroupParameter,
+
+        /**
+         * Resource provider name
+         */
+        @path @segment("providers") providerName: string,
```
markcowl: Again, a problematic pattern that applies only to resource group homed resource, rather than using 'scope' for this

specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
```diff
@@ -0,0 +1,144 @@
...
+    > {}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation"
+@armResourceOperations
+interface ApplyUpdateOperationGroup {
+  /**
+   * Track maintenance updates to resource
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("ApplyUpdates_Get")
```
markcowl: We should consider using the new `@moveTo` decorator in TCGC when this is released.

specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
```diff
@@ -0,0 +1,144 @@
...
+        KeyName = "applyUpdateName",
+        SegmentName = "applyUpdates",
+        NamePattern = ""
+      >>,
+      MaintenanceError
+    > {}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation"
+@armResourceOperations
+interface ApplyUpdateOperationGroup {
```
markcowl, line 122: Why can't this just be called 'ApplyUpdates'  or rename the interface?  The `@operationId` decorators will have to be cleaned up in any case, so I don't think we should leave this.

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * Configuration Assignment
+ */
+#suppress "@azure-tools/typespec-azure-core/no-private-usage"
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance"
+@Azure.ResourceManager.Private.armResourceInternal(
+  ConfigurationAssignmentProperties
+)
+@TypeSpec.Http.Private.includeInapplicableMetadataInPayload(false)
+model ConfigurationAssignment extends Foundations.ProxyResource {
```
markcowl, line 22: Why can't we just use `ProxyResource<T>` for this, or ExtensionResource<T>?

This is an extension resource, we should not use private decorators unless absolutely necessary.

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+
+        /**
+         * Resource provider name
+         */
+        @path @segment("providers") providerName: string,
+
+        /**
+         * Resource parent type
+         */
+        @path resourceParentType: string,
```
markcowl: See comments above about why this pattern is problematic

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+
+        /**
+         * Resource provider name
+         */
+        @path @segment("providers") providerName: string,
+
+        /**
+         * Resource parent type
+         */
+        @path resourceType: string,
```
markcowl: See comments above about why this pattern is problematic

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+    > {}
+
+@armResourceOperations
+interface ConfigurationAssignmentOperationGroup {
+  /**
+   * Get configuration assignment for resource..
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("ConfigurationAssignments_Get")
+  get is ConfigurationAssignmentOperationGroupOps.Read<ConfigurationAssignment>;
```
markcowl: Why can't we just put these operations in the previous interface and avoid all of the `@operationId` decorators we will have to replace in the future?

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+  delete is ConfigurationAssignmentsForSubscriptionOps.DeleteSync<
+    ConfigurationAssignment,
+    Response = ArmResponse<ConfigurationAssignment> | ArmDeletedNoContentResponse
+  >;
+
+  /**
+   * [UNSUPPORTED] Get configuration assignment within a subscription. This API is not implemented yet.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("ConfigurationAssignmentsWithinSubscription_List")
```
markcowl: `@moveTo`

specification/maintenance/Maintenance.Management/ConfigurationAssignment.tsp
```diff
@@ -0,0 +1,301 @@
...
+
+  /**
+   * Register configuration for resource.
+   */
+  createOrUpdate is ConfigurationAssignmentsForResourceGroupOps.CreateOrUpdateSync<ConfigurationAssignment>;
+
+  /**
+   * Register configuration for resource.
+   */
+  @patch(#{ implicitOptionality: false })
```
markcowl, line 313: This shouldn't be necessary.  The template sets implicit optionality to false

specification/maintenance/Maintenance.Management/MaintenanceConfiguration.tsp
```diff
@@ -0,0 +1,151 @@
...
+        Resource = MaintenanceConfiguration,
+        KeyName = "resourceName",
+        SegmentName = "maintenanceConfigurations",
+        NamePattern = ""
+      >>,
+      MaintenanceError
+    > {}
+
+@armResourceOperations
+interface MaintenanceConfigurationOperationGroup {
```
markcowl, line 92: Why not use the appropriate name?

markcowl: Note that I would prefer using the same interface and *not* using `@operationId`  over splitting things into two interfaces

markcowl (dismissed):
y initial thought is that the original swagger for this is a bit problematic.  It is describing extension resources (which are a special pattern - instead of having a fixed resource id,  they apply to another ARM resource, and the prefix of the resource ID is the id of the target resource).
 
That is, a standard resource has an id like:  
 
`/subscription/{subscriptionId}/providers/Microsoft.Foo/bars/{barName}`
 
But these have ids like:
 
`/subscription/{subscriptionId}/providers/Microsoft.Foo/bars/{barName}/providers/Microsoft.Baz/widgets/{widgetName}`
 
Where the prefix is the target resource.
 
In swagger and typespec, we normally handle this by having a {scope} parameter, so that, instead of parameterizing weird values like the target provider namespace and resource type, clients can use the target resource id as a parameter.
 
`/{scope}/providers/Microsoft.Baz/widgets/{widgetName}`
 
But this spec went in another direction and has highly parameterized paths, which has two main issues:
It doesn't fully describe the API - extensions can apply to any resource with any set of parents, and this describes a subset of that (a large subset)
It leads  to SDK methods with a number of rather esoteric parameters (like providerName) that the user will need to extract from a resource id,  rather than simpler parameters that represent commonly understood concepts (like resource ids).
 
Here's an example:
`/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{providerName}/{resourceParentType}/{resourceParentName}/{resourceType}/{resourceName}/providers/Microsoft.Maintenance/applyUpdates/{applyUpdateName}`
 
The TypeSpec spec does a pretty good job of reproducing this, but we have a design issue slated for next sprint for how to handle the old patterns that were used for extension resources, which would help us put more information into this spec (knowing this is an extension resource is useful information in multiple contexts), and likely make the TypeSpec a bit simpler. We also likely want to add a {scope} parameter to make sure we have fully captured the actual api.
 
It looks like there are a limited number of specs (4 others in the main branch: Batch, EventGrid, WorkloadMonitor, OperationsManagement) that do something similar to this, so feels like we don't need to rush this - even the specs that follow the more common pattern are not that numerous.
 
I am happy with you conveying this, but also happy to push this into the chat if that would be helpful.

ramoka178:
> y initial thought is that the original swagger for this is a bit problematic. It is describing extension resources (which are a special pattern - instead of having a fixed resource id, they apply to another ARM resource, and the prefix of the resource ID is the id of the target resource).
> 
> That is, a standard resource has an id like:
> 
> `/subscription/{subscriptionId}/providers/Microsoft.Foo/bars/{barName}`
> 
> But these have ids like:
> 
> `/subscription/{subscriptionId}/providers/Microsoft.Foo/bars/{barName}/providers/Microsoft.Baz/widgets/{widgetName}`
> 
> Where the prefix is the target resource.
> 
> In swagger and typespec, we normally handle this by having a {scope} parameter, so that, instead of parameterizing weird values like the target provider namespace and resource type, clients can use the target resource id as a parameter.
> 
> `/{scope}/providers/Microsoft.Baz/widgets/{widgetName}`
> 
> But this spec went in another direction and has highly parameterized paths, which has two main issues: It doesn't fully describe the API - extensions can apply to any resource with any set of parents, and this describes a subset of that (a large subset) It leads to SDK methods with a number of rather esoteric parameters (like providerName) that the user will need to extract from a resource id, rather than simpler parameters that represent commonly understood concepts (like resource ids).
> 
> Here's an example: `/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{providerName}/{resourceParentType}/{resourceParentName}/{resourceType}/{resourceName}/providers/Microsoft.Maintenance/applyUpdates/{applyUpdateName}`
> 
> The TypeSpec spec does a pretty good job of reproducing this, but we have a design issue slated for next sprint for how to handle the old patterns that were used for extension resources, which would help us put more information into this spec (knowing this is an extension resource is useful information in multiple contexts), and likely make the TypeSpec a bit simpler. We also likely want to add a {scope} parameter to make sure we have fully captured the actual api.
> 
> It looks like there are a limited number of specs (4 others in the main branch: Batch, EventGrid, WorkloadMonitor, OperationsManagement) that do something similar to this, so feels like we don't need to rush this - even the specs that follow the more common pattern are not that numerous.
> 
> I am happy with you conveying this, but also happy to push this into the chat if that would be helpful.

Please address this issue and also the open comments and Swagger LintDiff failures.

ArthurMa1978 (approved):

ArthurMa1978 (approved):

specification/maintenance/resource-manager/Microsoft.Maintenance/preview/2023-10-01-preview/Maintenance.json
```diff
@@ -2831,354 +2420,339 @@
...
+        "properties": {
+          "$ref": "#/definitions/MaintenanceConfigurationProperties",
+          "description": "Gets or sets properties of the resource",
+          "x-ms-client-flatten": true
+        }
+      },
       "allOf": [
         {
-          "$ref": "#/definitions/Resource"
+          "$ref": "../../../../../common-types/resource-management/v5/types.json#/definitions/TrackedResource"
```
v-jiaodi, line 2541: `location` should be optional for  `MaintenanceConfiguration`

specification/maintenance/Maintenance.Management/ApplyUpdate.tsp
pshao25: As we've already synced offline, resolve all these issues.

PR #34499, opened 2025-05-07
confidential ledger split service specific readme file

ivarprudnikov:
Confidential Ledger data-plane docs render incorrectly in https://learn.microsoft.com/en-us/rest/api/data-plane/confidentialledger/operation-groups?view=rest-data-plane-confidentialledger-2022-05-13

Split service specific readme file to generate 3 separate services instead of bundling them in one readme file


ivarprudnikov:
Will attempt to do it differently as this refactoring proves to be quite time intensive
https://github.com/MicrosoftDocs/azure-docs-rest-apis/pull/6847

weshaggard:
/cc @mikeharder 

Avocado currently complains if there isn't a readme.md directly in the data-plane folder but I'm not sure if that causes any real issues if they are under the service folder directly in a case where there are multiple services. 

LintDiff cannot handle this type of change because it is trying to produce a before and after diff of the issues and it has no way to map the 1 to 2 split we are doing here. 

ApiDocPreview currently only runs for a single service api version and only triggers if there are json files changes so it didn't do anything for this PR.

SDK generation is currently doing nothing as it doesn't see any sdk configuration in the split readmes. We need to figure out if these should be using the new TypeSpec code generation or the older readme based autorest generation and ensure things are generated correctly. 

@JeffreyRichter these are some things that we will need to investigate further with your layout proposal.

weshaggard:
@mikeharder is looking into addressing the tooling issues caused by the readme refactorings.

PR #34524, opened 2025-05-08
Add TypeSpec migration validation workflow

pshao25:
# Choose a PR Template


.github/workflows/typespec-migration-validation.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  typespec-validation:
+    name: TypeSpec Validation
+    runs-on: ubuntu-24.04
+
+    steps:
```
mikeharder, line 13: I think you can add a first step, which uses github-script to call a REST API to list the labels on the issue, and no-op if you don't have a `typespec-conversion-*` label.

mikeharder (changes requested):
see comment

.github/workflows/typespec-migration-validation.yaml
```diff
@@ -0,0 +1,27 @@
+name: TypeSpec Migration Validation
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
```
mikeharder, line 5:
```suggestion
    types: [labeled, opened, reopened, synchronize, unlabeled]
```

sort alphabetically, and add "reopened" which is included in default trigger

.github/workflows/typespec-migration-validation.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  typespec-migration-validation:
+    name: TypeSpec Migration Validation
+    if: contains(github.event.pull_request.labels.*.name, 'typespec-conversion-w1') || contains(github.event.pull_request.labels.*.name, 'typespec-conversion-w2')
```
mikeharder, line 13: Have you tested if this works on a PR from a fork?  According to the docs, the pull_request webhook event payload is empty for PRs from forks.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

mikeharder: I verified this does work!  This is a great find, because it means we can condition workflows on PR labels, without needing to run code on an agent.  And it automatically re-runs on every commit, or when the label is changed.

mikeharder (approved):

live1206 (approved):

PR #34546, opened 2025-05-09
chore: add new Ubuntu2204 OSSKU enum value to 2025-03-01

anujmaheshwari1:
This PR adds the new OSSKU's Ubuntu2204 to the 2025-03-01 API spec.

It uses the exact same format and wording as the approved API doc: https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/748089/Add-Ubuntu2204-and-Ubuntu2404-to-ossku


raiderss:
Brilliant implementation! At Eyes Store, we're also exploring similar technologies for our FiveM Tebex marketplace.

FumingZhang (approved):

PR #34566, opened 2025-05-09
add more guidance including .net naming violation fix

raych1:


maririos (approved):
while we understand why we need it and/or how to integrate it, this is fine

PR #34569, opened 2025-05-09
Update TypeSpec API Specification for Microsoft.CodeTransparency

praveenkuttappan:
This pull request includes changes to the TypeSpec API specification for the Microsoft.CodeTransparency service, including updates to the CreateEntryPending model and other related operations. It aims to enhance the API's compliance with the IETF SCITT draft.


weidongxu-microsoft:
Unsure if you want an offline review (if the only change is add 1 property).

PR #34625, opened 2025-05-13
Updated the required check condition

raych1:
Allow warnings result while checking the required CI check result. As long as the spec-gen-sdk check doesn't have failures, we should succeed the status check.

### DIFF
--- a/.github/workflows/src/spec-gen-sdk-status.js
+++ b/.github/workflows/src/spec-gen-sdk-status.js
@@ -177,7 +177,7 @@ async function processResult({ checkRuns, core }) {
     const shortLanguageName = language.split("-").pop();
     const executionResult = artifactJsonObj.result;
     const isSpecGenSdkCheckRequired = artifactJsonObj.isSpecGenSdkCheckRequired;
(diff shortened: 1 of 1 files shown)


.github/workflows/src/spec-gen-sdk-status.js
```diff
@@ -177,7 +177,7 @@ async function processResult({ checkRuns, core }) {
     const shortLanguageName = language.split("-").pop();
     const executionResult = artifactJsonObj.result;
     const isSpecGenSdkCheckRequired = artifactJsonObj.isSpecGenSdkCheckRequired;
-    if (isSpecGenSdkCheckRequired && executionResult !== "succeeded") {
+    if (isSpecGenSdkCheckRequired && !(executionResult === "succeeded" || executionResult === "warning")) {
```
weshaggard, line 180: Should we check for exectuionResult === "failed" instead?

raych1: I agree since `notEnabled` state is already validated in TSV. We don't need to block here.

mikeharder (approved):

PR #26576, opened 2023-11-06
Johnkoh msft azuredatatransfer microsoft.azure data transfer 2023 10 11 preview

johnkoh-msft:
# ARM (Control Plane) API Specification Update Pull Request 

## PR review workflow diagram

Please understand this diagram before proceeding. It explains how to get your PR approved & merged.

diagram

[1] [ARM review queue] (for **merge** queues, see [4])  
The PRs are processed by time opened, ascending. Your PR may show up on 2nd or later page. 
If you addressed Step 1 from the diagram and your PR is not showing up in the queue, ensure the label `ARMChangesRequested` 
is removed from your PR. This should cause the label `WaitForARMFeedback` to be added.  
[2] https://aka.ms/azsdk/support/specreview-channel  
[3] [List of SDK breaking changes approvers] in pinned Teams announcement  
[4] [public repo merge queue], [private repo merge queue] (for **ARM review** queue, [1])

If you need further help with anything, see `Getting help` section below.

## Purpose of this PR

What's the purpose of this PR? Check all that apply. This is **mandatory**!

  - [ ] New API version. (Such PR should have been generated with [OpenAPI Hub](https://aka.ms/openapihub)).
  - [ ] Update existing version for a new feature. (This is applicable only when you are revising a private preview API version.)
  - [ ] Update existing version to fix swagger quality issues in S360.
  - [x] Other, please clarify:
    - Our API specs have previously been merged into RPSaaSMaster in the PR repository. We are looking to move this new API version to the public repo. To see the changes between this new API version and the last API version that was approved for RPSaaSMaster, please view this closed PR in the PR repo: https://github.com/Azure/azure-rest-api-specs-pr/pull/15345

## Due diligence checklist

To merge this PR, you **must** go through the following checklist and confirm you understood 
and followed the instructions by checking all the boxes:

- [x] I confirm this PR is modifying Azure Resource Manager (ARM) related specifications, and not data plane related specifications.
- [x] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including
  [ARM resource provider contract](https://github.com/Azure/azure-resource-manager-rpc) and
  [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md) (estimated time: 4 hours).  
  I understand this is required before I can proceed to Step 2, "ARM Review", for this PR.

### Breaking changes review (Step 1)

- If the automation determines you have breaking changes, i.e. Step 1 from the diagram applies to you,
[...]

PR #34640, opened 2025-05-13
Convert Azure Data Transfer to typespec (2025-04-11-preview)

Reprezent:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Convert existing [OpenAPI spec to TypeSpec spec](https://aka.ms/typespec/conversion) (do not combine this with implementing changes for a new API version).
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/azuredatatransfer/resource-manager/Microsoft.AzureDataTransfer/preview/2025-04-11-preview/azuredatatransfer.json
```diff
@@ -2262,208 +3364,586 @@
           "$ref": "#/definitions/apiFlowOptions",
           "description": "The API Flow configuration options for Azure Data Transfer API Flow type."
         },
-        "customerManagedKeyVaultUri": {
+        "customerManagedKeyVaultUri": {
+          "type": "string",
+          "format": "uri",
```
fzkhan, line 3369: Our previous GA spec [mis-spelled](https://github.com/Azure/azure-rest-api-specs/blob/8dd69b3eb4198fc42ff4baf9c217508f5e09f4e6/specification/azuredatatransfer/resource-manager/Microsoft.AzureDataTransfer/stable/2024-09-27/azuredatatransfer.json#L2142) this as `url`. Maybe changing this format is what is causing it to flag as a breaking change? preview specs do not allow breaking changes over prev GA specs, so we should consider reverting this.

P.S. I tried [removing ](https://github.com/Azure/azure-rest-api-specs/pull/34598/commits/c2bb2c026aa2b22f76fbd29dd06d52e66713d078) it in my test PR for a new GA spec and it still got flagged as a breaking change: #34598

Reprezent: Yea I'm going to classify this as a bugfix under the justification as to have us better showcase the intent of the of the specification (as everything is named Uri and the server will reject anything that does not conform to RFC 3986). Regardless of a GA or preview change, this will cause a breaking change to be flagged.

TimLovellSmith:
I think you raised an open question about how many enum definitions to have. The answer depends, I think on whether they're conceptually representing distinct things, e.g. states of different state machines, and are not truly interchangable, or might hypothetically diverge in future.

specification/azuredatatransfer/AzureDataTransfer.Management/models.tsp
```diff
@@ -0,0 +1,1398 @@
...
+@@clientName(PipelineStatus, "pipelineStatus");
+@@clientName(SchemaStatus, "schemaStatus");
+@@clientName(ValidateSchemaStatus, "validateSchemaStatus");
+@@clientName(ListFlowsByPipelineConnection, "listFlowsByPipelineConnection");
+@@clientName(InternalMetadataProperties, "internalMetadataProperties");
+@@clientName(OperationStatusProperties, "operationStatusProperties");
+@@clientName(PendingFlow, "pendingFlow");
+@@clientName(PendingConnection, "pendingConnection");
+@@clientName(Subscriber, "subscriber");
+@@clientName(PipelineConnection, "pipelineConnection");
```
TimLovellSmith, line 1398: I'm puzzled about the naming, were things already named lowercase, counter to usual convention?

specification/azuredatatransfer/AzureDataTransfer.Management/models.tsp
```diff
@@ -0,0 +1,1398 @@
...
+ */
+model PipelineConnectionProperties {
+  ...Record<unknown>;
+
+  /**
+   * Internal metadata of the connection inside pipeline.
+   */
+  internalMetadata?: InternalMetadataProperties;
+}
+
```
TimLovellSmith, line 739: so you plan to allow users to set arbitrary metadata on here forever? How is it used, anyway? Will someone be subjecting it to validation other than your RP?

Reprezent: Looking at this a bit deeper, this type definition is a bit misleading as each of the underlying models (see directly below) are all read only, so a user cannot actually modify these. Some part of me feels like this is due to some voodoo magic we have to do with a `Pipeline` and a `Connection` as they can live in separate subscriptions, but let me double check.

Reprezent: After discussing with the team, this is because of how we have to link a `Pipeline` and a `Connection` across Entra ID Tenant and ARM Subscription boundaries. The team went with this design [(back in 2023)](https://github.com/Azure/azure-rest-api-specs/pull/26576) to be able to copy some of the `internalProperties` set by our RP on a `Connection` as a copy within the `Pipeline` to ensure we can correctly correlate in this esoteric scenario. As for more as to why its schemeless, looks like the idea here is to not lock down the schema as the internal required fields could change internally for us (as that's the point of the internal metadata).

On the `so you plan to allow users to set arbitrary metadata on here forever?` question, I've went ahead and confirmed with the team that all of these should be read only. I've went ahead and updated all the models to have a lifecycle of `Read` and explicitly set the return to be a read version of the model. I've removed the one instance of this being used in a Patch request, as that property is ignored and should not be allowed. I consider this breaking change to be a bugfix to ensure a user cannot modify an RP level property.

johnkoh-msft (approved):

PR #31523, opened 2024-11-13
Restructure KubernetesConfiguration directories to a suite of services

dipti-pai:
This PR splits the Kubernetes Configuration Services into a service group with 5 services. The swaggers for each service group has no new changes except FluxConfiguration Service which includes addition of a new provider field. ARM changes were reviewed by ARM team and signed off in PR #30956 . The service group restructuring has been discussed with the Azure Breaking Changes Board. 

These are the 5 services - 

- Flux Configuration Service 
     -- new API version: stable 2024-11-01 version
     -- base version specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/preview/2024-04-01-preview/fluxconfiguration.json

- Extensions Service 
    -- new API version: stable 2024-11-01
    -- base API version: specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/stable/2023-05-01

- Extension Types Service 
    -- new API version: preview 2024-11-01-preview
    -- base API version: specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/preview/2023-05-01-preview/extensionTypes.json

- Private Link Scopes Service 
   -- new API version: preview 2024-11-01-preview
   -- base API version: specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/preview/2022-04-02-preview/privateLinkScopes.json

- Operations Service 
   -- new API version: stable 2024-11-01
   -- base API version: specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/stable/2023-05-01/operations.json

For new reviewers, a lot of the discussions in this PR is regarding the Operations service. Summarizing the conclusion after discussion with the ARM team and Breaking changes board team - 

1) Operations will be a separate service with its own docs/sdk/etc.
2) Operations API will have a new version whenever any of the sub-services (flux configuration/extension/extensionTypes/private link scopes) has a new version.

# ARM (Control Plane) API Specification Update Pull Request 

> [!TIP]
> Overwhelmed by all this guidance? See the `Getting help` section at the bottom of this PR description.

## PR review workflow diagram

Please understand this diagram before proceeding. It explains how to get your PR approved & merged.

spec_pr_review_workflow_diagram

## Purpose of this PR

What's the purpose of this PR? Check the specific option that applies. This is **mandatory**!

  - [ ] New resource provider.
[...]

PR #34674, opened 2025-05-14
Restore readme file for monolithic KubernernetesConfiguration service

dipti-pai:

Context: KubernetesConfiguration service was split into a suite of services in [this ](https://github.com/Azure/azure-rest-api-specs/pull/31523) PR. Since, the older monolithic KubernetesConfiguration service is not expected to version again, the main readme.md and the language specific readmes were removed in [this ](https://github.com/Azure/azure-rest-api-specs/pull/33256) PR.

It was recently brought to our attention that removing the readme.md makes the older ARM APIs non-discoverable via bicep/arm templates. This PR reverts the older commit that deleted the readme.md

## Purpose of this PR

## Due diligence checklist

## Getting help

-


specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+``` yaml $(tag) == 'package-preview-2024-04'
+input-file:
+  - Microsoft.KubernetesConfiguration/stable/2023-05-01/extensions.json
+  - Microsoft.KubernetesConfiguration/preview/2024-04-01-preview/fluxconfiguration.json
+  - Microsoft.KubernetesConfiguration/stable/2023-05-01/kubernetesconfiguration.json
+  - Microsoft.KubernetesConfiguration/stable/2023-05-01/operations.json
+suppressions:  
+  - code: ResourceNameRestriction
+    from: fluxconfiguration.json
+    reason: Existing service contract needs to be backward compatible, pattern validation exists in RP.
```
TimLovellSmith, line 47: But why don't you document the pattern validation in your spec also?

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+  - code: DeleteResponseCodes
+    from: fluxconfiguration.json
+    reason: Existing service contract needs to be backward compatible, force delete does synchronous delete and returns 200.
+  - code: LroLocationHeader
+    from: fluxconfiguration.json
+    reason: Existing service contract needs to be backward compatible, 202 operations return Azure-Async-Operation header.
```
TimLovellSmith, line 53: This suppression reason doesn't really make sense to me! The thing you're being asked to do and suppressing is return a Location header, since that is what the RPC says you should do when you return 202. You can still also return a Azure-AsyncOperation header for backwards compat.

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+  - code: AvoidAdditionalProperties
+    from: fluxconfiguration.json
+    reason: Existing service contract needs to be backward compatible.
```
TimLovellSmith, line 56: Yes, but what was the reason for this property being additionalProperties in the first place? Sometimes its better to model it with a proper schema than keep using additionalProperties. Why is that the case here?

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+  - code: PatchResponseCodes
+    from: fluxconfiguration.json
+    reason: Existing service contract needs to be backward compatible.
```
TimLovellSmith, line 59: >Synchronous PATCH must have 200 return code and LRO PATCH must have 200 and 202 return codes.

What exactly are you doing to be backwards compatible? You normally really shouldn't be returning anything other than 200 and 202.

specification/kubernetesconfiguration/resource-manager/readme.md
TimLovellSmith, line 47: Suppressions should be scoped with filters to match them to the path of the violation, so that you don't get false negatives where one suppression direction suppresses genuine issues by suppressing multiple issues

dipti-pai:
Hi @TimLovellSmith Thanks for the review, can we address the suppressions in a subsequent iteration of the API version? This PR is not introducing a new API version OR changing any specs, it is just restoring the readme file that previously existed so that bicep/other tools that depend on it are functional again.

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+---
+
+## Suppression
+
+``` yaml
+directive:
+  - suppress: LongRunningResponseStatusCode
+    reason: The validation tools do not properly recognize 202 as a supported response code.
```
TimLovellSmith, line 209: Is this a current tooling issue, or an old, fixed tooling issue? Can we remove this supression?

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+    from: extensions.json
+  - suppress: TopLevelResourcesListBySubscription
+    reason: 'Microsoft.KubernetesConfiguration is a proxy resource provider under Microsoft.Kubernetes'
+    from: extensions.json
+  - suppress: TopLevelResourcesListBySubscription
+    reason: 'Microsoft.KubernetesConfiguration is a proxy resource provider under Microsoft.Kubernetes'
+    from: extensionTypes.json
+  - suppress: BodyTopLevelProperties
```
TimLovellSmith, line 217: Can you please remove this suppression for sytsemData being at top level.
You shouldn't be getting linter errors for $systemData if you are declaring it *correctly*. Likely, there is something else wrong with your declaration that we should understand.

specification/kubernetesconfiguration/resource-manager/readme.md
```diff
@@ -0,0 +1,359 @@
...
+    reason: 'Microsoft.KubernetesConfiguration is a proxy resource provider under Microsoft.Kubernetes'
+    from: kubernetesconfiguration.json
+  - suppress: LongRunningResponseStatusCode
+    reason: The validation tools do not properly recognize 202 as a supported response code.
+    from: extensions.json
+  - suppress: TopLevelResourcesListBySubscription
+    reason: 'Microsoft.KubernetesConfiguration is a proxy resource provider under Microsoft.Kubernetes'
+    from: extensions.json
+  - suppress: LongRunningResponseStatusCode
+    reason: The validation tools do not properly recognize 202 as a supported response code.
```
TimLovellSmith, line 310: hopefully we no longer need this

specification/kubernetesconfiguration/resource-manager/readme.md
TimLovellSmith: (or this is an old suppression that you don't really need with new improved tools?)

TimLovellSmith:
Approving for expediency since this API set / readme is just being restored for tooling backcompat and will be replaced.

dipti-pai:
> Approving for expediency since this API set / readme is just being restored for tooling backcompat and will be replaced.

Thank you @TimLovellSmith, appreciate the prompt review and feedback.

TimLovellSmith (approved):

PR #34677, opened 2025-05-14
[contoso] Move/copy readme.md for folder structure v2

mikeharder:
## Changelog
1. `data-plane/readme.md` moved one level down to `data-plane/Azure.Contoso.WidgetManager/readme.md`
2. `resource-manager/readme.md` copied to `resource-manager/Microsoft.Contoso/readme.md`
    - `resource-manager/readme.md`: Represents "ARM Schema Manifest" file (under RP namespace in final state)
    - `resource-manager/Microsoft.Contoso/readme.md`: Represents SDK/service level file (under RP namespace and service in final state)

## Related
* #34679
* #34821
* #34677
* #34823


JeffreyRichter (approved):

specification/contosowidgetmanager/resource-manager/Contoso.Widget/readme.md
```diff
@@ -0,0 +1,48 @@
+# containerstorage
```
weshaggard, line 1: How does this readme related to the one in the WidgetManagment folder?

JeffreyRichter: The readme.md file in the Contoso.Widget folder is the union of the content in each service name folder's readme.md file.  This union file is required by RPSaaS. I just learned yesterday that they will get rid of the need for this file "someday".

weshaggard: Should it be an import of the other files instead of a copy of the content?

Also, I suspect there may be other tools that require the readme at this level as well.

JeffreyRichter: I don't think the RPaaS schema validator will work if the top-level readme file imports (references) the lower-level readme files. I don't think other tools need this top-level readme because I had one team remove it entirely (because I didn't know about the RPaaS schema validation checker) and the schema validation checker was the only thing that broke.

JeffreyRichter: I also know that docs & SDKs will refer to the lower-level readmes and not the top-level readme.

mikeharder: What is the "RPaaS schema validator"?  Where does it run?  If the inputs are from the specs repo, shouldn't it be running as part of PR validation?

How are all the specs currently using "folder structure v1" passing the "RPaaS schema validator"?  Or is "RPaaS schema validator" a new feature that will only be enabled on specs using "folder structure v2"?

JeffreyRichter: Whenevere an ARM client requests enters the ARM RPaaS service, this service grabs the swagger files and compares the incoming HTTP client request's JSON body against the swagger schema definitions and fails the request if the validation fails. RP service responses are also validated against the schema.

RPaaS uses the private repo (not the public repo) and has URL links to the readme.md file which has the links to all the swaggers it needs to do the runtime validation. 

The schema validation has been happening for a few years now with the v1 folder structure and must continue to work with the v2 folder structure.

mikeharder: @JeffreyRichter: I'm still confused.  If the schema validation already works with the folder structure v1, why is the extra (merged) readme.md necessary for folder structure v2?

For resource-manager swagger and readmes, I thought your claim was the repo already supports both v1 and v2 (without the extra readme).  Some resource-manager specs using a single level of nesting (v1), but some already use two levels of nesting (I believe this is identical to v2, minus the extra readme.md).

So what exactly is introducing the requirement for the extra (merged) readme.md?

mikeharder: Or is splitting (and re-merging) the readmes the entire point of folder structure v2?  For example, this spec already has two levels of folders under resource-manager, but only a single readme.md directly under resource-manager.

https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager

So, while this spec already has their **swaggers** using the v2 folder structure, their **readme.md** is still using the v1 structure?  Is this why this spec works with "schema validation" today, but requires an extra (merged) readme after the readmes are split and moved down into sub-folders?

JeffreyRichter: The high priority goal is to make each service self-contained. 
So, each service has its own folder and ALL things related to this service are in that folder: tsp files, preview/stable swaggers, readme.md files, etc. And from these files, docs & SDKs are produced. If a service is later retired, then we can just delete the service folder and everything related to that service is deleted in one fell swoop.

For control-plane services, they are under an extra folder: the RP namespace folder (eg: Microsoft.Compute). data-plane doesn't have this extra folder. And, control-plane services all need this extra readme.md file because of RPaaS schema validation which validates all operations within the RP namespace regardless of how the operations are split across multiple services. The RPaaS team told me that they hope to get rid of the requirement for this extra readme.md file in the future so this is required as a stop gap until then.

mikeharder: Got it, so it's what I said in my last paragraph.  Some specs already in main split their **swaggers** into self-contained folders, but they still use a single `readme.md` at the top-level of `resource-manager`.

For these specs (like compute), the only change in v2 will be splitting the existing `readme.md` into multiple files, one per "service", and one extra readme per "RP namespace" (hopefully temporary).

PR #34727, opened 2025-05-15
[spec-model] Public APIs return Map<string, V> instead of Set<V>

mikeharder:


.github/shared/src/spec-model.js
```diff
@@ -79,22 +80,20 @@ export class SpecModel {
   }
 
   /**
-   * Given a swagger file, return the swagger files that are affected by the
+   * Given a swagger file, return the swagger files that are affected by
```
mikeharder, line 83:
```suggestion
   * Given a swagger file, return the swagger files that are affected by the
```

danieljurek (approved):
Approved. 

We discussed the possibility of adding more backpointers from objects like `Tag`. These can be convenient in some cases but also require making assumptions about a given object that cannot be easily enforced. 

For example, if I receive a Tag object can I be sure that the Tag was created as part of a Spec Model and the backpointers are always populated? A more robust pattern involves ensuring that you're building the spec model or passing in the spec model. 

Overall changes to `processChanges.ts` in LintDiff were minimal and resulted in slightly cleaner code.

mikeharder:
> We discussed the possibility of adding more backpointers from objects like `Tag`. These can be convenient in some cases but also require making assumptions about a given object that cannot be easily enforced.
> 
> For example, if I receive a Tag object can I be sure that the Tag was created as part of a Spec Model and the backpointers are always populated? A more robust pattern involves ensuring that you're building the spec model or passing in the spec model.

I'm happy to add backpointers (required or optional) in another PR.

PR #34746, opened 2025-05-16
Remove 1es-redirect

weshaggard:
We don't have an official production pipelines in repo so removing the unnecessary 1es-redirect.


benbp (approved):

benbp (approved):

weshaggard:
@raych1 why isn't the artifacts getting published? https://dev.azure.com/azure-sdk/public/_build/results?buildId=4877175&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692

Is there some reason SpecGenSdkArtifactName wouldn't get set? Also, we should make that variable defined in the pipeline yml and potentially pass it in so that we don't get `Publish $(SpecGenSdkArtifactName)` in the pipeline step name.

raych1:
@weshaggard the `spec-gen-sdk-artifact` is only published in pull request scenario. We might add a 'if' condition to include this publish step only when build reason is 'PullRequest'.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -29,257 +29,251 @@ parameters:
...
-                  -repoName "$(SdkRepoOwner)/$(SpecRepoName)" `
-                  -pullRequestNumber "$(System.PullRequest.PullRequestNumber)" 
-              displayName: 'Detect API Change and Create API Review'
-              condition: and(succeeded(), ne(variables['SpecGenSdkArtifactPath'], ''), ne(variables['SpecGenSdkArtifactName'], ''))
\ No newline at end of file
+      - template: /eng/common/pipelines/templates/steps/publish-artifact.yml
+        parameters:
+          ArtifactName: "spec-gen-sdk-logs"
+          ArtifactPath: "$(System.DefaultWorkingDirectory)/out/logs"
+          CustomCondition: and(succeeded(), ne(variables['SpecGenSdkArtifactPath'], ''), ne(variables['SpecGenSdkArtifactName'], ''))
```
raych1, line 279:
```suggestion
          CustomCondition: and(succeeded(), ne(variables['StagedArtifactsFolder'], ''))
```

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -29,257 +29,251 @@ parameters:
...
-                  -specGenSDKArtifactPath "$(System.DefaultWorkingDirectory)/$(SpecGenSdkArtifactPath)/$(SpecGenSdkArtifactName).json" `
-                  -apiviewArtifactName "$(PipelineArtifactsName)" `
-                  -buildId "$(Build.BuildId)" `
-                  -commitish "$(System.PullRequest.SourceCommitId)" `
-                  -repoName "$(SdkRepoOwner)/$(SpecRepoName)" `
-                  -pullRequestNumber "$(System.PullRequest.PullRequestNumber)" 
-              displayName: 'Detect API Change and Create API Review'
-              condition: and(succeeded(), ne(variables['SpecGenSdkArtifactPath'], ''), ne(variables['SpecGenSdkArtifactName'], ''))
\ No newline at end of file
+      - template: /eng/common/pipelines/templates/steps/publish-artifact.yml
```
raych1, line 275: Why not keep this step along with other's artifact publish step? This step should be run even if APIView creation step fails.

weshaggard: I just put this at the end like it was before but I'm fine with publishing them together given the retry doesn't currently work anyway.

raych1:
> Why do we only publish for PR? Is that because we only do that for ApiView? Wouldn't those artifacts be interesting in other contexts as well?

This `spec-gen-sdk-artifact` currently includes the information of breaking change and ApiView, which are specifically relevant in the PR context. At this point, it’s not used in release or batch run scenarios.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: Right, retry won't work if ApiView step fails.

raych1 (approved):

PR #34769, opened 2025-05-19
Support reporting the duplicated SDK config issue in the batch run of spec-gen-sdk pipeline

raych1:
This pull request introduces significant updates to the `spec-gen-sdk-runner` tool, focusing on improving the handling of specification configurations, enhancing batch processing, and updating the underlying code structure. Key changes include the introduction of a new `SpecConfigs` type, the addition of a utility function to group specification paths, and updates to batch SDK generation logic to better manage configurations.

### Enhancements to Specification Configuration Handling:
* **New `SpecConfigs` Type**: Added a new `SpecConfigs` type to represent paired `readme.md` and `tspconfig.yaml` paths, improving clarity and reducing ambiguity in spec configuration handling (`eng/tools/spec-gen-sdk-runner/src/utils.ts`).
* **Grouping Spec Configurations**: Introduced the `groupSpecConfigPaths` function to group `readme.md` and `tspconfig.yaml` files by service, enabling more structured processing of specs (`eng/tools/spec-gen-sdk-runner/src/spec-helpers.ts`).

### Improvements to Batch SDK Generation:
* **Refactored `getSpecPaths`**: Updated the `getSpecPaths` function to return an array of `SpecConfigs`, replacing the previous flat string array approach. This change supports more complex batch run scenarios (`eng/tools/spec-gen-sdk-runner/src/command-helpers.ts`).
* **Enhanced Logging and Reporting**: Added detailed logging for duplicated SDK configurations and updated the markdown report to include a count of such cases (`eng/tools/spec-gen-sdk-runner/src/commands.ts`). [[1]](diffhunk://#diff-d7be89bc13d39990960756a431e7e422649175ec72b22237a9ebabf6b41c7e4eR269-R274) [[2]](diffhunk://#diff-d7be89bc13d39990960756a431e7e422649175ec72b22237a9ebabf6b41c7e4eR288-R302)

### Code Structure and Test Updates:
* **File Renaming and Refactoring**: Renamed `change-files.ts` to `spec-helpers.ts` and moved relevant functions for better organization (`eng/tools/spec-gen-sdk-runner/src/spec-helpers.ts`).
* **Test Snapshot Updates**: Updated test snapshots to reflect the new batch processing logic, including changes in the successful specs count (`eng/tools/spec-gen-sdk-runner/test/src/__snapshots__/commands.test.ts.snap`).

This is a [test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4880157&view=ms.vss-build-web.run-extensions-tab)


eng/tools/spec-gen-sdk-runner/src/commands.ts
```diff
@@ -187,29 +195,52 @@ export async function generateSdkForBatchSpecs(batchType: string): Promise<numbe
...
+  let specConfigPath = "";
+  let stagedArtifactsFolder = "";
 
   // Generate SDKs for each spec
-  for (const specConfigPath of specConfigPaths) {
-    logMessage(`Generating SDK from ${specConfigPath}`, LogLevel.Group);
+  for (const specConfigs of specConfigsArray) {
+    if (specConfigs.tspconfigPath && specConfigs.readmePath) {
+      logMessage(
+        `Generating SDK from ${specConfigs.tspconfigPath} and ${specConfigs.readmePath}`,
```
weshaggard, line 213: Why are we generating from both? I would expect us to just warn and use the tsp configuration.

weshaggard: Or are we just passing both not necessarily generating both as I suspect spec-gen-sdk is only generating from tsp.

raych1: We only generate from TypeSpec if both configs exist and report the warnings of duplicated config issue. Here is just the logging to differentiate between one input config and two input configs.
image

weshaggard: OK I think this log message might cause confusion and we should either remove it and leave it up to spec-gen-sdk tool to log and in the runner we just pass whichever parameters we have.

raych1: I knew that it's a little confusion, but I haven't come with a better solution. The group log is written by runner, and runner passes spec configs to spec-gen-sdk tool. Then spec-gen-sdk tool validate spec configs and determine the right configuration to be used for code generation. Currently users are able to know the details when expanding the logs.
What do you suggest for the group log title?

weshaggard: I would just remove the and part. In fact, just remove this entire if block and let the below if/else write generating for tsp (if tsp is set ignoring the readme) and then generating from the readme, if only the readme is present.

raych1: @weshaggard I updated it to log the service folder as group title. It's not perfect but it should not be fine for now?

weshaggard: Let's start with that as it should at least help with some confusion.

eng/tools/spec-gen-sdk-runner/src/spec-helpers.ts
```diff
@@ -182,3 +183,180 @@ export function detectChangedSpecConfigFiles(commandInput: SpecGenSdkCmdInput):
...
+): SpecConfigs[] {
+  const emptyArray: string[] = [];
+  const safeTspConfigs = tspconfigs ?? emptyArray;
+  const safeReadmes = readmes ?? emptyArray;
+
+  if (skipUnmatchedReadme) {
+    // Explicitly remove the specific readme file from the array
+    // as we don't want to include it for typespec related batch runs
+    const indexToRemove = safeReadmes.indexOf(
+      "specification/awsconnector/resource-manager/readme.md",
```
weshaggard, line 207: Probably worth adding a comment as to why this is removed.

weshaggard (approved):

eng/tools/spec-gen-sdk-runner/src/spec-helpers.ts
raych1: The issue here is 'awsconnector' service folder contains a bunch of sub services and all of them don't enable SDK emitters in their 'tspconfig.yaml'. This causes every sub service to generate code based on parent readme.md.
With further thoughts, I plan to add an option to 'spec-gen-sdk' tool to skip SDK generation for TypeSpecs batch runs if the paired 'readme.md' has the SDK configuration but 'tspconfig.yaml' don't have the SDK configuration.

weshaggard (approved):

PR #34804, opened 2025-05-20
Added Logic App Extension to stable api version

Shubhrmcf-07:
# Logic Apps Extension on Microsoft.App RP

This PR aims to introduce logic app capabilities on top of container apps as a new offering. We already offer our services on Microsoft.Web RP and introducing a new SKU here. 

This spec is already a part of the preview API versions and is released to the customers as Logic App (Hybrid) offering. 

Key APIs introduced : 

- Logic App List Connections : Gets the connections for a particular logic app
- Logic App Get Workflows : Lists the workflows present on the app.
- Logic App Get Workflow : Get the workflow properties for a specific workflow on the app


specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+          }
+        }
+      }
+    },
+    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.App/containerApps/{containerAppName}/providers/Microsoft.App/logicApps/{logicAppName}/invoke": {
+      "post": {
+        "tags": [
+          "LogicApps"
+        ],
+        "summary": "Proxies a the API call to the logic app backed by the container app.",
```
razvanbadea-msft: > a [](http://example.com/codeflow?start=27&length=2)

nit: remove fix typo #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_Get.json
```diff
@@ -0,0 +1,19 @@
...
+    "logicAppName": "testcontainerApp0",
+    "api-version": "2025-07-01"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "id": "/subscriptions/8efdecc5-919e-44eb-b179-915dca89ebf9/resourceGroups/examplerg/providers/Microsoft.App/containerApps/testcontainerApp0/providers/Microsoft.App/logicApps/testcontainerApp0",
+        "name": "testcontainerApp0",
+        "type": "Microsoft.App/logicApps",
+        "properties": {}
```
razvanbadea-msft: >{} [](http://example.com/codeflow?start=22&length=2)

update this example with more realistic properties #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+          "x-ms-client-flatten": true
+        }
+      },
+      "allOf": [
+        {
+          "$ref": "../../../../../common-types/resource-management/v5/types.json#/definitions/ProxyResource"
+        }
+      ]
+    },
+    "LogicAppProperties": {
```
razvanbadea-msft: >LogicAppProperties [](http://example.com/codeflow?start=5&length=18)

How will this be used without a schema #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource name."
+        },
+        "type": {
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource type."
+        },
+        "kind": {
```
razvanbadea-msft, line 350: >kind [](http://example.com/codeflow?start=9&length=4)

what is the difference between kind and type? can it be any string for both? #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+          }
+        },
+        "nextLink": {
+          "description": "Link to next page of resources.",
+          "type": "string",
+          "readOnly": true
+        }
+      }
+    },
+    "WorkflowFiles": {
```
razvanbadea-msft, line 602: >WorkflowFiles [](http://example.com/codeflow?start=5&length=13)

why is this not an array?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+      "properties": {
+        "appSettings": {
+          "$ref": "#/definitions/Object",
+          "description": "Application settings of the workflow."
+        },
+        "files": {
+          "$ref": "#/definitions/WorkflowFiles",
+          "description": "Files of the app."
+        },
+        "filesToDelete": {
```
razvanbadea-msft, line 660: >filesToDelete [](http://example.com/codeflow?start=9&length=13)

are the files listed here from files. You should expand more the description about this collections and how it will be used

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_Get.json
Shubhrmcf-07: ### so far we do not have any properties in the extension body, this has been the placeholder for things we will introduce in the future. Its the same as preview versions, it is already supported in preview with same schema.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: As I mentioned before, this does not have any schema/properties defined as yet, this is supported the same way in preview version and is an object for things we are planning to introduce in properties.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: Type is the resource type, e.g containerApp
Kind specifies a specific kind of containerApp, ex. kind=workflowApp for logic apps

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: This is the same schema as the one we have for web apps, we are migrating to container apps and do not want to change anything. This is not supposed to be an array, it is supposed to be an object with key as filename, and value as file content.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: no, this is an array of file names, corresponding file names are checked in the file system and deletes the corresponding files. Again, we are already supporting this extension in preview versions with same schema, and on our existing web offering also this is the same way. This is not a new schema we are introducing.

krishnajaju (approved):

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
gary-x-li: Please update the description for each property to make it clearer.

Shubhrmcf-07: updated

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_Get.json
gary-x-li: If you don't have any concrete definition, you should remove it. Defining it without any schema and restrict it in the future will break customers. This is especially true for stable versions since you'll need to support it for a long time.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
gary-x-li: Similarly as above, this is an anti-pattern.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
gary-x-li: You need to define a schema for it.

Shubhrmcf-07: Added detailed description and examples on usage as an object, cannot define schema as the files are sent with filenames as keys and file content as values (which is consistent with our existing offerings on Microsoft.Web)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,742 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "Files of the app to delete."
+        }
+      },
+      "description": "The workflow filter."
+    },
+    "Object": {
```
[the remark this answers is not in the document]
Shubhrmcf-07: yes, we are using it to represent a generic json object (as used in other api specs as well), using it to represent response of an invoke call which may vary based on headers, as well as using it to represent files to add/modify in deployartifacts where the object will be of the form { <filename> : <filecontent> } and schema cannot be determined at spec.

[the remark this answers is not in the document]
rkmanda: If you know its of the form { <filename> : <filecontent> } what is preventing you from defining a type with those 2 properties

rkmanda:
pl visit the ARM API review office hours to discuss the need for a generic object.. aka.ms\armofficehoursinfo

specification/app/resource-manager/Microsoft.App/preview/2025-02-02-preview/ContainerApps.json
```diff
@@ -723,7 +723,8 @@
         "kind": {
           "type": "string",
           "enum": [
-            "workflowapp"
+            "workflowapp",
+            "functionapp"
```
najian: 2025-02-02-preview is already released, we'd better do new changes only in 2025-07-01 folder

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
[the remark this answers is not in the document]
Shubhrmcf-07: because the filenames and paths can be anything, not fixed.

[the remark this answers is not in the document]
Shubhrmcf-07: it is similar to how other API specs are using it

Shubhrmcf-07:
The generic object is valid in the sense that we don't have the filename/path and filecontent at the spec. We are using it in 2 places : 

1.) Represents files in payload of deploy artifacts : 

We do not have the filename/path and file content at spec, so it is represented as just a generic object and can be anything. For example : 

{
"files" : {
"Stateful1/workflow.json" : {},
"Stateful2/input.txt" : "hello",
"Artifacts/test.xml" : "<root>...</root>"
}

so it has to be generic.

2.) Represents the response of the invoke API : 

The invoke APIs in logic app are used to invoke runtime operations and can have different outputs based on the headers sent. So there is no definite pattern there. Hence, it is required.

We have already discussed this spec thoroughly and currently offering it in our preview versions, we do not want to make a lot of changes, with some of the changes in this PR we are already modifying the contract upto a certain extent and not desirable. 

In fact, this is the same contract we are using in our offering on Microsoft.Web and do not want to deter from that.

Shubhrmcf-07:
The LintDiff error here is unrelated to the files changed in this PR. Can be ignored.

mentat9:
> The generic object is valid in the sense that we don't have the filename/path and filecontent at the spec. We are using it in 2 places :
> 
> 1.) Represents files in payload of deploy artifacts :
> 
> We do not have the filename/path and file content at spec, so it is represented as just a generic object and can be anything. For example :
> 
> { "files" : { "Stateful1/workflow.json" : {}, "Stateful2/input.txt" : "hello", "Artifacts/test.xml" : "..." }
> 
> so it has to be generic.
> 
> 2.) Represents the response of the invoke API :
> 
> The invoke APIs in logic app are used to invoke runtime operations and can have different outputs based on the headers sent. So there is no definite pattern there. Hence, it is required.
> 
> We have already discussed this spec thoroughly and currently offering it in our preview versions, we do not want to make a lot of changes, with some of the changes in this PR we are already modifying the contract upto a certain extent and not desirable.
> 
> In fact, this is the same contract we are using in our offering on Microsoft.Web and do not want to deter from that.

Was there a modeling discussion @rkmanda requested here: https://github.com/Azure/azure-rest-api-specs/pull/34804#issuecomment-2937889135?

I share the concerns of earlier ARM reviewers. Unschematized JSON is a major ARM anti-pattern and RPs are not allowed to publish APIs that use it anymore, aside from a few scenarios that aren't otherwise practical to schematize. You are creating new APIs with this PR, so current guidance applies: it's not enough that it's more convenient to keep a previously published pattern.

Unschematized objects are vectors for bypassing Azure API versioning. This makes them hard for clients to use. They also can't be supported by several important ARM core features. You need a strong justification to use this pattern in a new API. From what I can tell, the two examples you described above are unlikely to qualify:
1. The first one is easily represented as an array of &lt;string&gt;, &lt;string&gt; pairs (fileName, fileContents). This is fully schematized and supported by all ARM features.
2. The second requires your service to parse the response and return a schematized object representing the result, rather than just dumping the unparsed response to the client. Doing this will make your API better.

If you want to pursue using this pattern, please schedule ARM modeling office hours to go through the scenario and describe the importance of keeping this pattern in a new API. If you have already done that, please provide details.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,748 @@
...
+        }
+      }
+    }
+  },
+  "definitions": {
+    "LogicApp": {
+      "type": "object",
+      "description": "A logic app extension resource",
+      "properties": {
+        "id": {
```
mentat9: >id [](http://example.com/codeflow?start=9&length=2)

Remove these top-level property definitions (id, name, type). They are already defined in the common-types ProxyResource definition.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
mentat9: That doesn't necessarily mean the API is idempotent. What does the API do if the "files" array includes files that are already uploaded? Likewise, what does the API do if the "filesToDelete" array includes files that don't exist? Idempotency says that if I send a request twice I will get the same result both times.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,748 @@
...
+    },
+    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.App/containerApps/{containerAppName}/providers/Microsoft.App/logicApps/{logicAppName}/listWorkflowsConnections": {
+      "post": {
+        "tags": [
+          "LogicApps"
+        ],
+        "summary": "Gets logic app's connections.",
+        "operationId": "LogicApps_ListWorkflowsConnections",
+        "parameters": [
+          {
```
ajaygit158, line 275: Why isnt this being modelled as a GET ?

najian (approved):

raosuhas:
> > The generic object is valid in the sense that we don't have the filename/path and filecontent at the spec. We are using it in 2 places :
> > 1.) Represents files in payload of deploy artifacts :
> > We do not have the filename/path and file content at spec, so it is represented as just a generic object and can be anything. For example :
> > { "files" : { "Stateful1/workflow.json" : {}, "Stateful2/input.txt" : "hello", "Artifacts/test.xml" : "..." }
> > so it has to be generic.
> > 2.) Represents the response of the invoke API :
> > The invoke APIs in logic app are used to invoke runtime operations and can have different outputs based on the headers sent. So there is no definite pattern there. Hence, it is required.
> > We have already discussed this spec thoroughly and currently offering it in our preview versions, we do not want to make a lot of changes, with some of the changes in this PR we are already modifying the contract upto a certain extent and not desirable.
> > In fact, this is the same contract we are using in our offering on Microsoft.Web and do not want to deter from that.
> 
> Was there a modeling discussion @rkmanda requested here: [#34804 (comment)](https://github.com/Azure/azure-rest-api-specs/pull/34804#issuecomment-2937889135)?
> 
> I share the concerns of earlier ARM reviewers. Unschematized JSON is a major ARM anti-pattern and RPs are not allowed to publish APIs that use it anymore, aside from a few scenarios that aren't otherwise practical to schematize. You are creating new APIs with this PR, so current guidance applies: it's not enough that it's more convenient to keep a previously published pattern.
> 
> Unschematized objects are vectors for bypassing Azure API versioning. This makes them hard for clients to use. They also can't be supported by several important ARM core features. You need a strong justification to use this pattern in a new API. From what I can tell, the two examples you described above are unlikely to qualify:
> 
> 1. The first one is easily represented as an array of <string>, <string> pairs (fileName, fileContents). This is fully schematized and supported by all ARM features.
> 2. The second requires your service to parse the response and return a schematized object representing the result, rather than just dumping the unparsed response to the client. Doing this will make your API better.
> 
> If you want to pursue using this pattern, please schedule ARM modeling office hours to go through the scenario and describe the importance of keeping this pattern in a new API. If you have already done that, please provide details.


We discussed this in Office hours. Both these patterns are incorrect and should not be used. We discussed options to fix both of these scenarions. Unfortunately both of these are already released apis that customers are using in the current namespace that were missed in the original ARM review. 

For the first one the current design can be suppressed. 

For the second one that I would highly recommend looking into retiring the api since it contains a number of ARM anti patterns and is still in preview stage. But if there are timne constraints then it can be approved as well.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: Again, parity with existing offering, and were also advised to keep it this way from container apps side.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: This is not just the resource property, it is the structured definition being used as the exact body for create request as well as the response to multiple other calls, we are using these in response for further computations so need to have this structure (it is the same as in preview version as well - which is already Live since 2024-02-02-preview)

TimLovellSmith:
Can you please fix the PR description?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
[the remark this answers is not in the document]
TimLovellSmith: @shubhrmcf-07 That doesn't actually prevent you from fixing it, IIUC.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
TimLovellSmith: @Shubhrmcf-07 Did you respond to the concern here?

TimLovellSmith:
> For the second one that I would highly recommend looking into retiring the api since it contains a number of ARM anti patterns and is still in preview stage. But if there are timne constraints then it can be approved as well.

Based on what I'm reading in the discussion here,
For the second one, sounds like you should really fix this 'second' API design now, *before* you put this in a stable version.

You can't easily fix it afterwards.

(Aren't there already known issues caused by the existing API that you based this on? Why waste the opportunity to fix them?)

Shubhrmcf-07:
Removing the concerning specs for now for better modelling in the future

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
```diff
@@ -699,6 +702,22 @@
           ],
           "description": "The fully qualified resource ID of the resource that manages this resource. Indicates if this resource is managed by another Azure resource. If this is present, complete mode deployment will not delete the resource if it is removed from the template since it is managed by another resource."
         },
```
TimLovellSmith, line 707: Wouldn't our naming guidelines be to use PascalCase - or at least camelCase? #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,567 @@
...
+            "name": "logicAppName",
+            "in": "path",
+            "description": "Name of the Logic App, the extension resource.",
+            "required": true,
+            "type": "string",
+            "pattern": "^[-\\w\\._\\(\\)]+$"
+          },
+          {
+            "name": "resource",
```
TimLovellSmith, line 83: > "name": "resource", [](http://example.com/codeflow?start=11&length=20)

Not sure why change it, you can't really "create or update" by PUT with an empty body. Can you? #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,567 @@
...
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource name."
+        },
+        "type": {
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource type for ex. Microsoft.App/logicApps/workflows."
+        },
+        "kind": {
```
TimLovellSmith, line 350: You defined an enum for 'kind' elsewhere, but not here - can you change this to an enum? I'd consider its a better form of documentation and not really breaking to change it from string that was only allowed by server to be 2 fixed values.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
TimLovellSmith: If it is supposed to be an object with key as filename, and value as file content, as you say, then I'd think it should be modeled as a dictionary - you don't specify *how* the value is interpreted the file content in that short description, but your API docs really should explain that kind of thing right?

i.e. I"d say why not model it as additionalProperties?

    "WorkflowFiles": {
      "description": "Gets or sets the files.",
      "type": "object",
      "additionalProperties": {
           "$ref": "#/definitions/FileContent"
      }
}

With a suppression to explain it, and schema description etc of the FileContent?

praveensri (approved):

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: Removed this for better modelling in future

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
Shubhrmcf-07: this is the way we have it in our backend for existing offering on Microsoft.Web RP and also in the preview version for this one.

Shubhrmcf-07: since these are not related to spec and just values, we can keep it this way as enum values.

Shubhrmcf-07: both for control plane and data plane we have ordinal validation for this

Shubhrmcf-07: k8se.microsoft.com/appKind: workflowapp

^from our pods

Shubhrmcf-07: even for functionapp (which is not our service), it is the same way. this is how it is in preview as well and not in our purview to change it for them as well.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: This API is to create an extension resource on top of Container App resource. Basically the container app already exists and this is just to register a logic app extension resource on top of it. So no explicit body is needed, and the parameters are covered in the api path itself. The resource here in body refers to the top level properties like id, name, type, which are already part of the ProxyResource type as mentioned in another comment above so dont need to define those explicitly.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
TimLovellSmith: So, the argument here is back compatibility with existing public preview versions, did I understand correctly? @shubhrmcf-07

I reckon it goes against naming guidelines. And being back compat with preview versions is not a hard requirement.

But it is good if you are backwards compatible with preview versions from the point of preview customers, and I don't know how many preview customers you have.

In the end, it is your decision.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
TimLovellSmith: Sorry looks like in my comment I pasted the wrong text(?!). 
What I meant is, not sure why you are making required false, for the body of the resource.

But I think I understand now, your intention is to make the entire body optional, because there are no required properties.

TimLovellSmith:
>           "enum": [

nit, docs here could be made more complete
How is client cert forwarded? Where are cert details? Is client certificate checked for validity. when it is accepted, required, or ignored?

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1074 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

TimLovellSmith:
>         },

nit, this doc could be clarified: how are secret names interpreted? Is it just looking it up from a SecretsCollection somewhere?

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1116 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

TimLovellSmith:
>         },

nit, the meaning/value/format of this string is unclear to me

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1108 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

TimLovellSmith:
>           "description": "Resource ID of a managed identity to authenticate with Azure Key Vault, or System to use a system-assigned identity.",

nit, putting 'System' in quotes could be clearer

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1158 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

TimLovellSmith:
>           "description": "Resource ID of a managed identity to authenticate with Azure Key Vault, or System to use a system-assigned identity.",

Nit saying 'Either resouce id of...' could be clearer

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1158 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
```diff
@@ -699,6 +702,22 @@
...
+        "kind": {
+          "type": "string",
+          "enum": [
+            "workflowapp",
+            "functionapp"
+          ],
+          "x-ms-mutability": [
+            "read",
+            "create"
+          ],
```
TimLovellSmith, line 717: > If supported, the resource provider must validate and persist this value. [](http://example.com/codeflow?start=182&length=74)

Can we take out the text " If supported, the resource provider must validate and persist this value."?

That doesn't make sense to customers who are trying to understand the API

TimLovellSmith, line 717: >Metadata used to render different experiences for resources of the same type; [](http://example.com/codeflow?start=26&length=77)

nit the description 'Metadata used to render different experiences for resources of the same type;' is a general description of what property 'kind' is, and accurate, but probably not super useful in this context.

You could explain that it is metadata to declare whether the (container app?) is a workflow app or a function app, and which set of properties it should have

TimLovellSmith:
>     },

Are all these properties valid for both logic apps and function apps?

Or do we need to model it as discriminated property set?

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:844 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+            "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/ResourceGroupNameParameter"
+          },
+          {
+            "name": "containerAppName",
+            "in": "path",
+            "description": "Name of the Container App.",
+            "required": true,
+            "type": "string",
+            "pattern": "^[-\\w\\._\\(\\)]+$"
+          },
```
TimLovellSmith, line 228: optional nit, there's a less verbose factoring, you could consider having the parameter definitions in just one place, and using $ref to them - that helps ensure you do them consistently

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+        "name": {
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource name."
+        },
+        "type": {
+          "type": "string",
+          "readOnly": true,
+          "description": "Gets the resource type for ex. Microsoft.App/logicApps/workflows."
+        },
```
TimLovellSmith: Instead of declaring these properties id/name/type, shouldn't you inherit ProxyResource?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+      "type": "object",
+      "description": "A logic app extension resource",
+      "allOf": [
+        {
+          "$ref": "../../../../../common-types/resource-management/v5/types.json#/definitions/ProxyResource"
+        }
+      ]
+    },
+    "WorkflowEnvelope": {
+      "description": "Workflow properties definition.",
```
TimLovellSmith: >Workflow properties definition. [](http://example.com/codeflow?start=22&length=31)

Isn't this level of object modeling *the whole resource*, and not just the set of resource properties?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+        "value": {
+          "description": "Collection of resources.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/WorkflowEnvelope"
+          }
+        },
+        "nextLink": {
+          "description": "Link to next page of resources.",
+          "type": "string",
```
TimLovellSmith, line 399: nit option to use "format": "uri"

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+        "state": {
+          "description": "Gets or sets the workflow health state.",
+          "enum": [
+            "NotSpecified",
+            "Healthy",
+            "Unhealthy",
+            "Unknown"
+          ],
+          "type": "string",
+          "x-ms-enum": {
```
TimLovellSmith, line 425: >enum [](http://example.com/codeflow?start=16&length=4)

is it readonly?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+            "name": "WorkflowHealthState",
+            "modelAsString": false
+          }
+        },
+        "error": {
```
TimLovellSmith, line 430: >error [](http://example.com/codeflow?start=9&length=5)

is it readonly?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+          "description": "Message template.",
+          "type": "string"
+        },
+        "parameters": {
+          "description": "Parameters for the template.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
```
TimLovellSmith, line 475: Do you only support unnamed, ordered parameters?
Just wondering if that makes it easy to change the parameter sets without calls breaking.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+    "ErrorEntity": {
+      "description": "Body of the error response returned from the API.",
+      "type": "object",
+      "properties": {
+        "extendedCode": {
+          "description": "Type of error.",
+          "type": "string"
+        },
+        "messageTemplate": {
+          "description": "Message template.",
```
TimLovellSmith, line 466: Is the message template in some format that requires post-processing to get the actual message?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,576 @@
...
+      "type": "object",
+      "description": "JSON object.",
+      "properties": {}
+    },
+    "ErrorEntity": {
+      "description": "Body of the error response returned from the API.",
+      "type": "object",
+      "properties": {
+        "extendedCode": {
+          "description": "Type of error.",
```
TimLovellSmith, line 462: any well known or example error codes you wish to document?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+      "managedBy": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.Web/sites/testcontainerAppKind",
+      "kind": "workflowapp",
+      "properties": {
+        "managedEnvironmentId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.App/managedEnvironments/testmanagedenv3",
+        "configuration": {
+          "activeRevisionsMode": "Single",
+          "ingress": {
+            "external": true,
+            "targetPort": 80,
+            "allowInsecure": true
```
TimLovellSmith: >allowInsecure [](http://example.com/codeflow?start=13&length=13)

consider putting insecure 'false' instead of true in the examples, since 
a) people are likely to copy them, and we'd like to b) push people into the pit of security success

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+              "exposedPort": 0,
+              "transport": "auto",
+              "traffic": [
+                {
+                  "weight": 100,
+                  "latestRevision": true
+                }
+              ],
+              "customDomains": null,
+              "allowInsecure": true,
```
TimLovellSmith: >allowInsecure [](http://example.com/codeflow?start=15&length=13)

consider putting insecure 'false' instead of true in the examples, since 
a) people are likely to copy them, and we'd like to b) push people into the pit of security success

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
```
TimLovellSmith: >allowInsecure [](http://example.com/codeflow?start=15&length=13)

consider putting insecure 'false' instead of true in the examples, since 
a) people are likely to copy them, and we'd like to b) push people into the pit of security success

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+      "location": "East Us",
+      "managedBy": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.Web/sites/testcontainerAppKind",
+      "kind": "workflowapp",
+      "properties": {
+        "managedEnvironmentId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg/providers/Microsoft.App/managedEnvironments/testmanagedenv3",
+        "configuration": {
+          "activeRevisionsMode": "Single",
+          "ingress": {
+            "external": true,
+            "targetPort": 80,
```
TimLovellSmith: >     "targetPort": 80, [](http://example.com/codeflow?start=7&length=22)

consider making your example show a secure HTTPS configuration instead of HTTP. Or at least having two examples, one secure, one insecure.

TimLovellSmith:
>         "customDomains": {

I see its not modeled as x-ms-nullable, so it should be omitted in example responses, not returning null

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1029 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

TimLovellSmith:
>         "customDomains": {

Or you should update to x-ms-nullable if you actually return 'null' values from the service!

---
In reply to: [3237507417](https://github.com/Azure/azure-rest-api-specs/pull/34804#issuecomment-3237507417) [](http://example.com/codeflow?ancestors=3237507417)

---
Refers to: specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json:1029 in d5c0569. [](commit_id = d5c0569282cf09a99e47e11cf2e346cee4251b0c, deletion_comment = False)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+                  "latestRevision": true
+                }
+              ],
+              "customDomains": null,
+              "allowInsecure": true,
+              "ipSecurityRestrictions": null,
+              "corsPolicy": null,
+              "clientCertificateMode": null,
+              "stickySessions": null,
+              "additionalPortMappings": null
```
TimLovellSmith: 'null' value shouldn't be returned in your response unless
your property is annotated nullable

What you would do instead is just omit the value from the response.

Please check your example against your actual service behavior!

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+                {
+                  "weight": 100,
+                  "latestRevision": true
+                }
+              ],
+              "customDomains": null,
+              "allowInsecure": true,
+              "ipSecurityRestrictions": null,
+              "corsPolicy": null,
+              "clientCertificateMode": null,
```
TimLovellSmith: >      "clientCertificateMode": null, [](http://example.com/codeflow?start=8&length=36)

clientCertificateMode being a non nullable enum value, it should not be null, it should be an actual value, or unspecified.
If you have a default behavior, consider returning the value which best explains the default behavior

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+              "stickySessions": null,
```
TimLovellSmith: >"stickySessions": null, [](http://example.com/codeflow?start=14&length=23)

Does your service actually default to 'sticky', or 'none'? It is one of those right?
Return whichever one you actually default to, not null!

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+              "additionalPortMappings": null
+            },
+            "registries": null,
+            "dapr": null,
+            "runtime": null,
+            "maxInactiveRevisions": 100,
+            "service": null
```
TimLovellSmith: do you have examples covering all your optional properties that are null in this example? Its good documentation to have examples covering example usage of all your properties.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+          },
+          "template": {
+            "revisionSuffix": "",
+            "terminationGracePeriodSeconds": null,
```
TimLovellSmith: > "terminationGracePeriodSeconds": null, [](http://example.com/codeflow?start=11&length=39)

[ARMBlockingComment] 
again, I think null in responses should require x-ms-nullable on the property

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
```diff
@@ -0,0 +1,200 @@
...
+                "image": "default/logicapps-base:latest",
+                "name": "logicapps-container",
+                "resources": {
+                  "cpu": 1.0,
+                  "memory": "2Gi",
+                  "ephemeralStorage": "4Gi"
+                }
+              }
+            ],
+            "initContainers": null,
```
TimLovellSmith, line 103: >initContainers [](http://example.com/codeflow?start=13&length=14)

where is initContainers declared?

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_ListConnections.json
```diff
@@ -0,0 +1,44 @@
...
+        "id": "/subscriptions/8efdecc5-919e-44eb-b179-915dca89ebf9/resourceGroups/examplerg/providers/Microsoft.App/containerApps/testapp2/providers/Microsoft.App/logicApps/testapp2/workflowconfigurations/connections",
+        "name": "testapp2/connections",
+        "type": "Microsoft.App/logicApps/workflowsconfiguration",
+        "properties": {
+          "files": {
+            "connections.json": {
+              "managedApiConnections": {
+                "office365": {
+                  "api": {
+                    "id": "string"
```
TimLovellSmith: a more realistic looking example might be helpful? #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_ListConnections.json
```diff
@@ -0,0 +1,44 @@
...
+                  },
+                  "connection": {
+                    "id": "string"
+                  },
+                  "connectionRuntimeUrl": "string",
```
TimLovellSmith: >string [](http://example.com/codeflow?start=43&length=6)

a more realistic looking example might be helpful? #Resolved

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: yes, getting from runtime

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: yes

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
Shubhrmcf-07: not ours to update, already mentioned as optional but not owned by us: https://github.com/Azure/azure-rest-api-specs/blob/release-app-Microsoft.App-stable/2025-07-01/specification/app/resource-manager/Microsoft.App/stable/2025-07-01/CommonDefinitions.json

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
Shubhrmcf-07: these properties are not associated with our resource (logic app) and at container app level, we don't need it for logic app, our aim with this example is to demonstrate proper payload for updating the kind property, other than that these properties are not ours to update in terms for definitions, you can find the values in other examples related to container apps. in fact, we don't need this in the request for workflowapp.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
Shubhrmcf-07: again, the aim of this file is not to specify all properties of container apps, we are not responsible for putting these values as nullable or getting sample values, these property definitions and examples can already be found in other places and are not ours to update.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: it is used at multiple places in multiple formats, basically in some of the requests like ListWorkflows, the response has a part with this WorkflowEnvelope array along with other properties, then while updating connections or updating workflows, this schema is just a set of resource properties which might be just one part of the response, updated description.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: already documented in -> https://github.com/Azure/azure-rest-api-specs/blob/7757286399f92099f616a02463d7d8a9ad7e70cf/specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_ListWorkflows.json

This is only the error for workflow health which is just a part of the response, not for the entire call. example documented in above file

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: this is actually coming from our runtime and the parameters are already set, its not a whole error response for the call but just a part of the get workflow response, example documented here => https://github.com/Azure/azure-rest-api-specs/blob/7757286399f92099f616a02463d7d8a9ad7e70cf/specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_ListWorkflows.json

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: see below

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: updated, but started seeing lintdiff errors

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
TimLovellSmith: @shubhrmcf-07 I'm not sure what you are saying here.

If these properties are not valid in the logic app scenario, it seems to me that you might want to leverage pattern like 'discriminated types', in order to make that more explicit in your swagger.

TimLovellSmith:
my mistake, target is development branch

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
```diff
@@ -0,0 +1,525 @@
...
+    },
+    "WorkflowEnvelope": {
+      "description": "Schema for the workflow object.",
+      "type": "object",
+      "allOf": [
+        {
+          "$ref": "../../../../../common-types/resource-management/v5/types.json#/definitions/ProxyResource"
+        }
+      ],
+      "properties": {
```
TimLovellSmith, line 349: Seems like a discriminator kind.

Does it make sense to leverage "x-ms-discriminator-value" with this?

[https://github.com/Azure/autorest/blob/main/docs/extensions/swagger-extensions-examples/x-ms-discriminator-value/readme.md](https://github.com/Azure/autorest/blob/main/docs/extensions/swagger-extensions-examples/x-ms-discriminator-value/readme.md)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
```diff
@@ -699,6 +705,22 @@
           ],
           "description": "The fully qualified resource ID of the resource that manages this resource. Indicates if this resource is managed by another Azure resource. If this is present, complete mode deployment will not delete the resource if it is removed from the template since it is managed by another resource."
         },
+        "kind": {
```
TimLovellSmith, line 708: Seems like a discriminator kind.

Does it make sense to leverage "x-ms-discriminator-value" with this?

[https://github.com/Azure/autorest/blob/main/docs/extensions/swagger-extensions-examples/x-ms-discriminator-value/readme.md](https://github.com/Azure/autorest/blob/main/docs/extensions/swagger-extensions-examples/x-ms-discriminator-value/readme.md)

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/LogicApps_ListConnections.json
```diff
@@ -0,0 +1,44 @@
...
+    "resourceGroupName": "testrg123",
+    "containerAppName": "testapp2",
+    "logicAppName": "testapp2",
+    "api-version": "2025-07-01"
+  },
+  "responses": {
+    "200": {
+      "headers": {},
+      "body": {
+        "id": "/subscriptions/8efdecc5-919e-44eb-b179-915dca89ebf9/resourceGroups/examplerg/providers/Microsoft.App/containerApps/testapp2/providers/Microsoft.App/logicApps/testapp2/workflowconfigurations/connections",
```
TimLovellSmith: >/8efdecc5-919e-44eb-b179-915dca89ebf9 [](http://example.com/codeflow?start=29&length=37)

note, I recommend using obviously fake subscription ids though

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/examples/ContainerApps_Kind_CreateOrUpdate.json
Shubhrmcf-07: Removed these as per the other comments and validating against the actual response. The are the container app level properties and as an extension resource on top of container apps this is not ours to update. The example shows the relevant properties that refer to the kind property update, other things are not relevant in this example. I removed those and can be found in other container apps examples

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/ContainerApps.json
Shubhrmcf-07: I doubt that its valid here because there are not different resources but properties of the same containerapp resource itself, we are not defining a specific app for Logic App or Functions App to make it different. For a particular base containerapp itself we register as Logic App or Function App, the default base container app is the resource itself which is same for both. Those are not separate inherited resources from the container app resource but they are a kind of container app resource.

specification/app/resource-manager/Microsoft.App/stable/2025-07-01/LogicAppsExtension.json
Shubhrmcf-07: Again, this is the same resource. To use this property we would likely have something to inherit the base workflow, but these are properties on the workflow resource itself, it does not create a new resource with different properties. Exisitng piece should be right here.

PR #34816, opened 2025-05-20
Update Microsoft.CodeTransparency TypeSpec with location header for TransparentStatement

l0lawrence:
Added location header to TransparentStatement model to provide proper IETF SCITT draft compliance.


l0lawrence:
testing workflow

PR #34817, opened 2025-05-20
Add sharedModelList property to WidgetSuite model

l0lawrence:
This pull request adds a new property `sharedModelList` to the WidgetSuite model in the Contoso.WidgetManager TypeSpec project. This property is an array of FakedSharedModel objects that allows storing multiple shared models for a widget.


l0lawrence:
testing workflow

PR #34844, opened 2025-05-21
Added conditional headers to Get and Set Blob Tags

seanmcc-msft:


specification/storage/data-plane/Microsoft.BlobStorage/stable/2026-02-06/blob.json
```diff
@@ -12076,6 +12100,44 @@
       },
       "description": "Specify a SQL where clause on blob tags to operate only on blobs with a matching value."
     },
+    "IfBlobMatch": {
```
jalauzon-msft, line 12103: Any reason you created new definitions for these instead of using the existing ones above? These ones don't include the `x-ms-parameter-grouping` which is important.

seanmcc-msft: These headers include "blob" in the header name, the existing conditional headers do not.  I'll add the parameter grouping.

weirongw23-msft (approved):

PR #34871, opened 2025-05-22
[app config] migrate to new c# generator

jorgerangel-msft:
This PR adds enables the generation of the azure.data.appconfiguration c# library using the service's typespec and the new csharp generator.


specification/appconfiguration/AppConfiguration/tspconfig.yaml
```diff
@@ -26,6 +26,7 @@ options:
     model-namespace: false
     flavor: azure
   "@azure-typespec/http-client-csharp":
+    package-name: Azure.Data.AppConfiguration
```
JoshLove-msft: I think this the package-name is defaulted to the namespace, so not strictly necessary.

JoshLove-msft (approved):

jsquire (approved):

PR #34878, opened 2025-05-22
[EngSys] Support folder structure v2

mikeharder:
- [test/Avocado] Ignore TSP `examples` folder
- [TSV/FolderStructure] Detect v1/v2 and validate accordingly
  - Regex for v2 RP Namespace: `/^[A-Za-z0-9\.]+$/`
  - Regex for v2 Service (DP or RM): `/^[A-Za-z0-9]+$/`
- [spec-gen-sdk-runner] Add "under /resource-manager/" to RM detection

## Related
- https://github.com/Azure/azure-rest-api-specs/pull/34823


.github/workflows/avocado-code.yml
```diff
@@ -29,6 +29,7 @@ jobs:
             "/package.json" \
             "/package-lock.json" \
             "/cadl/examples/" \
+            '(?=/examples/)(?!(?:/stable/|/preview/))' \
```
weshaggard, line 32: Do you know where these paths are rooted from? If they are rooted from the readme then we might not need the stable/preview bits or we can do something like "data-plane/examples".

mikeharder: https://github.com/Azure/avocado/blob/643c09d18fa84a74965ad65d86ace01d6b8fc39a/src/index.ts#L822

```
exclude.some(item => v.path.search(item) !== -1)
```

`v.path` is the fully-qualified path of the file which caused the error (e.g. `/home/foo/specs/specification/foo/.../bar.json`).

`exclude` is an array of strings, interpreted as regexes.  If any exclude regex matches the fully-qualified path, the error is ignored.

There may be a better regex, but it's tricky.  We need to exclude paths like this:

```
# TypeSpec Examples
specification/foo/data-plane/Foo/examples/e1.json
specification/foo/resource-manager/Microsoft.Foo/FooManagement/examples/e1.json
```

But not paths like this:

```
# Swagger Examples
specification/foo/data-plane/Foo/stable/2025-01-01/examples/e1.json
specification/foo/resource-manager/Microsoft.Foo/FooManagement/preview/2025-02-01-preview/examples/e1.json
```

So, my regex excludes paths that include `/examples/`, but don't include `/stable/` or `/preview/`.  This means the file is a TypeSpec example, since TypeSpec examples should **never** include `/stable/` or `/preview/` in their path (assuming we never allow a service or org to use these names).  And Swagger examples should **always** include `/stable/` or `/preview/` in their path, since these are the only two valid parent directories for Swaggers.

eng/tools/typespec-validation/src/index.ts
```diff
@@ -57,6 +57,11 @@ export async function main() {
   let success = true;
   for (let i = 0; i < rules.length; i++) {
     const rule = rules[i];
+
+    if (rule instanceof FolderStructureRule) {
+      continue;
```
mikeharder, line 62: Temporarily disable rule, until it allows v2.

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -83,45 +59,114 @@ export class FolderStructureRule implements Rule {
...
+          errorOutput += `Invalid folder structure: TypeSpec for data-plane specs or shared code must be in a folder NOT ending with '.Management'`;
+          success = false;
+        }
+      }
+    } else if (structureVersion === 2) {
+      if (!tspConfigExists) {
+        errorOutput += `Invalid folder structure: Spec folder must contain tspconfig.yaml.`;
+        success = false;
+      }
+
```
mikeharder, line 124:
```suggestion

      // TODO: Decide if v2 should validate folder names and structure to ensure specs align, or allow a more loose structure to ease
      // migration from v1 to v2.

```

eng/tools/spec-gen-sdk-runner/src/utils.ts
```diff
@@ -409,7 +409,7 @@ export function groupPathsByService(
     }
 
     const info = serviceMap.get(serviceName)!;
-    if (folderPath.endsWith(".Management")) {
+    if (folderPath.endsWith(".Management") || folderPath.includes("resource-manager")) {
```
mikeharder, line 412: @raych1: Do you still want this merged to main as part of this PR, even though you already merged #35108?

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -83,45 +59,117 @@ export class FolderStructureRule implements Rule {
...
+      // TODO: Decide if v2 should validate folder names and structure to ensure specs align, or allow a more loose structure to ease
+      // migration from v1 to v2.
+
+      const specType = folder.includes("data-plane") ? "data-plane" : "resource-manager";
+      if (specType === "data-plane") {
+        if (folderStruct.length !== 4) {
+          errorOutput +=
+            "Invalid folder structure: TypeSpec for data-plane specs must be in a folder exactly one level under 'data-plane', like 'specification/foo/data-plane/Foo'.";
```
mikeharder:
```suggestion
            "Invalid folder structure: TypeSpec for data-plane specs must be in a folder exactly one level under 'data-plane', like 'specification/foo/data-plane/FooAnalytics'.";
```

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -83,45 +59,117 @@ export class FolderStructureRule implements Rule {
...
+      if (specType === "data-plane") {
+        if (folderStruct.length !== 4) {
+          errorOutput +=
+            "Invalid folder structure: TypeSpec for data-plane specs must be in a folder exactly one level under 'data-plane', like 'specification/foo/data-plane/FooAnalytics'.";
+          success = false;
+        }
+      } else if (specType === "resource-manager") {
+        if (folderStruct.length !== 5) {
+          errorOutput +=
+            "Invalid folder structure: TypeSpec for resource-manager specs must be in a folder exactly two levels under 'resource-manager', like 'specification/foo/resource-management/Microsoft.Foo/FooManagement'.";
```
mikeharder:
```suggestion
            "Invalid folder structure: TypeSpec for resource-manager specs must be in a folder exactly two levels under 'resource-manager', like 'specification/foo/resource-management/Microsoft.Foo/Foo'.";
```

eng/tools/spec-gen-sdk-runner/src/utils.ts
raych1: It's ok to merge it as it won't introduce bad impact.

PR #34909, opened 2025-05-23
[SpecModel] Improve SpecModelError

mikeharder:
- Move SpecModelError to separate file
- Document properties
- Add more tests


danieljurek (approved):
Thanks for adding the member variables directly in the class definition. I'd left them out.

PR #34913, opened 2025-05-23
[SpecModel] Add backpointers, resolve paths against backpointers, move "new Swagger" from Readme to Tag

mikeharder:
- move code calling "new Swagger()" from Readme to Tag to align responsibility
- main benefit of backpointers is calling code doesn't need to try/catch/enhance/throw to add error context
- may be useful in the future for additional exceptions, logging, etc.
- properties are optional and backwards-compatible

### DIFF
--- a/.github/shared/src/spec-model-error.js
+++ b/.github/shared/src/spec-model-error.js
@@ -24,9 +24,9 @@ export class SpecModelError extends Error {
    * @param {string} message
    * @param {Object} [options]
    * @param {Error} [options.cause]
-   * @param {string} [options.source]
-   * @param {string} [options.readme]
-   * @param {string} [options.tag]
+   * @param {string} [options.source] Path to file that caused the error
+   * @param {string} [options.readme] Path to readme that caused the error (if known)
+   * @param {string} [options.tag] Name of tag that caused the error (if known)
    */
   constructor(message, options) {
     super(message, { cause: options?.cause });
--- a/.github/shared/src/swagger.js
+++ b/.github/shared/src/swagger.js
@@ -48,7 +48,7 @@ export class Swagger {
    * @param {Tag} [options.tag]
    */
   constructor(path, options) {
-    this.#path = resolve(path);
+    this.#path = resolve(options?.tag?.readme?.path ?? "", path);
     this.#logger = options?.logger;
     this.#tag = options?.tag;
   }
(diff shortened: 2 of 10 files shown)


.github/shared/src/spec-model-error.js
```diff
@@ -24,9 +24,9 @@ export class SpecModelError extends Error {
    * @param {string} message
    * @param {Object} [options]
    * @param {Error} [options.cause]
-   * @param {string} [options.source]
-   * @param {string} [options.readme]
-   * @param {string} [options.tag]
+   * @param {string} [options.source] Path to file that caused the error
+   * @param {string} [options.readme] Path to readme that caused the error (if known)
+   * @param {string} [options.tag] Name of tag that caused the error (if known)
```
mikeharder, line 29: Only adding descriptions

.github/shared/src/swagger.js
```diff
@@ -70,6 +70,8 @@ export class Swagger {
             {
               cause: error,
               source: error.source,
+              tag: this.#tag?.name,
+              readme: this.#tag?.readme?.path,
```
mikeharder, line 74: Backpointers allow specifying the full error context immediately

.github/shared/src/swagger.js
```diff
@@ -38,19 +38,19 @@ export class Swagger {
   /** @type {Map<string, Swagger> | undefined} */
   #refs;
 
-  /** @type {SpecModel | undefined} backpointer to owning SpecModel */
-  #specModel;
+  /** @type {Tag | undefined} Tag that contains this Swagger */
+  #tag;
```
mikeharder, line 42: Swagger already had a backpointer, albeit an unprecise one only pointing to SpecModel

.github/shared/src/swagger.js
```diff
@@ -105,15 +107,22 @@ export class Swagger {
...
    * @param {ToJSONOptions} [options]
    * @returns {Promise<Object>}
    */
   async toJSONAsync(options) {
     return {
       path:
-        options?.relativePaths && this.#specModel
-          ? relative(this.#specModel.folder, this.#path)
+        options?.relativePaths && this.#tag?.readme?.specModel
+          ? relative(this.#tag?.readme?.specModel.folder, this.#path)
```
mikeharder, line 125: backpointers allow walking all the way to root

PR #34937, opened 2025-05-27
Add ARM TypeSpec migration instructions

pshao25:
As titled


.github/instructions/mgmt-migration.instructions.md
```diff
@@ -1,11 +1,103 @@
...
+     - The string value (e.g., "2023-01-01") is what will be substituted for {version} in the output path
+   - The value of {version-status} is either stable or preview
+   - The value of {service-name} is determined from the `main.tsp` file:
+     - The namespace of the main TypeSpec file.
+     - For example: `namespace Microsoft.Compute;`
+     - The service name is `Microsoft.Compute` in this case.
+
+2. Download the latest specification as a baseline by running the following command:
+
+```powershell
```
josefree, line 89: the {} part in the guidance copilot agent provided base on this description is confusing. could we provide an example of the powershell command using a service
image

josefree: I ran the script like this, but not sure what should be expected.
image

josefree: I am confused why I need to download the swagger of latest version from main again, if before the conversion (in a newly created branch) it was confirmed to have the latest commit.  Instead, I wonder if I should

josefree: seems this is the proper command to execute. It is generated when only adding comparison.instructions.md as reference
image

.github/instructions/comparison.instructions.md
```diff
@@ -0,0 +1,76 @@
...
+
+2. Second, determine the location of the original Swagger folder:
+   - The original Swagger is typically located at `sparse-spec\specification\{service-folder-name}`
+   - Use the innermost folder (e.g., `sparse-spec\specification\storageactions\resource-manager\Microsoft.StorageActions\stable\2023-01-01`) as the original Swagger folder path
+   - Check if this folder exists by running:
+     ```powershell
+     Test-Path sparse-spec\specification\{service-folder-name}
+     ```
+   - If the folder doesn't exist, instruct the user to run:
+     ```powershell
```
josefree, line 30: the script execution failed  
image

josefree: how to root cause the failure?

PR #32748, opened 2025-02-21
Migrate compute latest swagger to TypeSpec

pshao25:
This PR is migrating Compute(ComputeRP, GalleryRP, DiskRP) REST API definitions to TypeSpec. 
This is PR contains all the latest Compute versions translated to TypeSpec, and Swagger files generated again from the TypeSpec files.

Original Swagger files were divided in to resource types - virtualmachine.json, images.json etc. 
TypeSpec generated Swagger file has 1 json - Compute.json 
Because of the different structure, LintDiff and BreakingChange CI checks are failing.

However, the ONLY real breaking change is: DiskRP's "PurchasePlan" is now "DiskPurchasePlan". 
There was a conflicting definition from Compute also called "PurchasePlan" 
See [Original Definition link](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/DiskRP/stable/2024-03-02/diskRPCommon.json#L417) [Renamed Definition link](https://github.com/Azure/azure-rest-api-specs/blob/migration/compute/specification/compute/resource-manager/Microsoft.Compute/DiskRP/stable/2024-03-02/DiskRP.json#L3168)
[Compute's definition](https://github.com/Azure/azure-rest-api-specs/blob/migration/compute/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-11-01/ComputeRP.json#L16002)
The small breaking change above is already mitigated at the SDK level. 

This migration PR has been thorough validated at every level - through generating all SDKs, and even at track 1 SDK (because Compute has track 1 dependency for PowerShell still) 
See PRs: 
.NET https://github.com/Azure/azure-sdk-for-net/pull/48196
Python https://github.com/Azure/azure-sdk-for-python/pull/40600
Go https://github.com/Azure/azure-sdk-for-go/pull/24459
Java https://github.com/Azure/azure-sdk-for-java/pull/45013
Js https://github.com/Azure/azure-sdk-for-js/pull/33054
PowerShell (track 1 SDK) https://github.com/Azure/azure-powershell/pull/27827

PRs for latest GA'd releases:
ComputeRP: https://github.com/Azure/azure-rest-api-specs/pull/32787
DiskRP: https://github.com/Azure/azure-rest-api-specs/pull/29399
GalleryRP: https://github.com/Azure/azure-rest-api-specs/pull/31254

## For validation
[...]

PR #34953, opened 2025-05-27
[servicefabricmanagedclusters] sdk configuration changes for java and dotnet

iliu816:

## Purpose of this PR

## Due diligence checklist

## Getting help


a-santamaria (approved):

specification/servicefabricmanagedclusters/ServiceFabricManagedClusters.Management/back-compat.tsp
```diff
@@ -380,3 +380,6 @@ using Microsoft.ServiceFabric;
   armResourceIdentifier[],
   "csharp"
 );
+
+@@scope(OperationStatus.get, "!csharp");
```
live1206, line 384: this should be moved to client.tsp

pshao25: back-compat.tsp will be imported by main.tsp. It is to ensure the swagger equivalence because swagger is emitted from main.tsp. If your previous swagger has `x-ms-client-name`, that corresponding TypeSpec code should be in back-compat.tsp.

client.tsp is only used to generate SDK. It will not be read by swagger emitter, so your generated swagger doesn't have any content from client.tsp. 

It's better to put these into client.tsp since it doean't impact swagger generation at all.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/preview/2024-11-01-preview/servicefabricmanagedclusters.json
```diff
@@ -8152,11 +8152,13 @@
         },
         "settings": {
           "type": "object",
-          "description": "Json formatted public settings for the extension."
+          "description": "Json formatted public settings for the extension.",
+          "additionalProperties": {}
         },
```
rkmanda, line 8155: please book a slot in the ARM API review office hours to explain why your scenario needs additionalProperties

specification/servicefabricmanagedclusters/ServiceFabricManagedClusters.Management/Operations.tsp
```diff
@@ -33,7 +33,13 @@ interface OperationResults {
     operationId: string,
   ):
     | OkResponse
-    | ArmAcceptedResponse<ExtraHeaders = ArmLroLocationHeader>
+    | (AcceptedResponse & {
+        /**
+         * The URL to get the status of a completed long-running operation.
+         */
+        @header("Location")
+        location?: string;
```
weidongxu-microsoft: nit, Would `LocationHeader` do?

specification/servicefabricmanagedclusters/ServiceFabricManagedClusters.Management/NodeType.tsp
```diff
@@ -904,14 +904,14 @@ model VMSSExtensionProperties {
...
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Day 0 property"
+  settings?: Record<unknown>;
 
   /**
    * The extension can contain either protectedSettings or protectedSettingsFromKeyVault or no protected settings at all.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "Day 0 property"
-  protectedSettings?: {};
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Day 0 property"
+  protectedSettings?: Record<unknown>;
```
weidongxu-microsoft: A notice that I found the unmerged migration of compute (which contains VMSS) uses `unknown`.

PR https://github.com/Azure/azure-rest-api-specs/pull/32748
tsp
https://github.com/Azure/azure-rest-api-specs/blob/710a6e41bd2d2a659615833dfebdf67a22fb0732/specification/compute/Compute.Management/models.tsp#L2989-L2999

`unknown` would affect Swagger too, make them
```
  "settings": {
  }
```
without `type: object` -- basically it means `any`

Loop @pshao25 for advise too.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/preview/2024-11-01-preview/servicefabricmanagedclusters.json
iliu816: I had used record<unknown> based on advice from 

https://teams.microsoft.com/l/message/19:084875bb626242ed95f3ac8dddfaa12a@thread.skype/1747972523177?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1747972523177&teamName=Azure%20SDK&channelName=SDK%20release%20support&createdTime=1747972523177

Would appreciate insight from @weidongxu-microsoft on whether it should be changed from record<unknown> to unknown, which would remove the additionalProperties. Awaiting further discussion below about Compute team's use of unknown and whether that is the correct approach.

Further discussion about unknown vs record<unknown> that makes me believe we should be using unknown:
https://github.com/Azure/typespec-azure/issues/1725#issuecomment-2448645928

settings and protectedSettings are represented as JObjects in our code:
https://msazure.visualstudio.com/One/_git/winfab-RP?path=/src/CSMTemplate/Model/VMExtensionResourcePropertites.cs&version=GBdevelop&_a=contents&line=14&lineStyle=plain&lineEnd=15&lineStartColumn=1&lineEndColumn=1

specification/servicefabricmanagedclusters/ServiceFabricManagedClusters.Management/Operations.tsp
iliu816: I think so. The only change in the output swagger is capitalization ("Location" -> "location") and the description. This might've been introduced swagger conversion tool and then not modified.

weidongxu-microsoft (approved):

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/preview/2024-11-01-preview/servicefabricmanagedclusters.json
weidongxu-microsoft: Yes, AFAIK, https://github.com/Azure/typespec-azure/issues/1725#issuecomment-2448645928 is the recommended migrate solution given by TypeSpec dev.

Strictly speaking, `unknown` means any JSON type, include JSON object, but could also be JSON string or numeric or boolean or array.
`Record<unknown>` means JSON object.

MaryGao: Is the translation from TypeSpec `unknown` to swagger `{}`  correct?

From [table](https://github.com/Azure/typespec-azure/issues/1725#issuecomment-2448645928), it seems it could be `{}` or `{"type": "object"}` in swagger side? I can understand that Autorest has [never made a clear difference between {} and {type: object}](https://github.com/Azure/typespec-azure/issues/1725#issuecomment-2448088391) but it doesn't mean the openapi has no difference for them.

{3694F6C7-F406-4A23-A707-ED19AA2D200C}

weidongxu-microsoft: I assume the dev on typespec-autorest already figured that out.

Not expert but here is one [stackoverflow](https://stackoverflow.com/questions/32841298/openapi-what-schema-to-accept-any-complex-json-value)

iliu816: Hi @rkmanda, we are no longer adding "additionalproperties" to the model. Were there any other points of concern about this PR blocking approval?

a-santamaria (approved):

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/preview/2024-11-01-preview/servicefabricmanagedclusters.json
mentat9: The ARM concern here isn't specifically about "additionalProperties": {}, rather it is the fact that the API uses an unschematized object. This is a major ARM anti-pattern. The suggestion of booking office hours was to discuss better, more ARM and client-friendly options for modeling this API object. So the more general question of "why does this API require a property that has no schema?" remains.

iliu816: From my understanding, this is an existing model in our spec that is also present in previous versions. It is a pass through for Microsoft.Compute. The change introduced in this PR is to align with how Java SDK client generation handles the Typespec. The api surface on the service side hasn't changed, and it would be a breaking change to change the behavior.

a-santamaria: @mentat9 we can  go to ARM office hours to discuss further. But this shouldn't block this pr because that property has been there since it was introduced and we are following the same pattern as VMSS https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachinescalesets?tabs=bicep&pivots=deployment-language-arm-template#virtualmachinescalesetextensionproperties-1. Each extension owner can define their own model for this property.

a-santamaria: Discussed today with @mentat9 during ARM API Review Office hours. here are the meeting notes:
- SFMC manages the VMSS resource on behalf of the customers. 
- These properties are pass-through to VMSS
-  These properties are unschematized object following the same definition as [VMSS on their api](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachinescalesets?tabs=bicep&pivots=deployment-language-arm-template#virtualmachinescalesetextensionproperties-1) because each extension owner can define their own model for setting and protectedSettings.
- In this pr we are not changing that behavior. Those properties have been defnied as unschematized since the begining of the service [see the first public api 2021-01-01](https://github.com/Azure/azure-rest-api-specs/blob/f005e2723d97279b42445fafe13b9b0cad379a52/specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/stable/2021-05-01/nodetype.json#L816) before we were using TypeSpec.
- The change on this pr is to fix the sdk generation from TypeSpec. Chaning the TS definition from {} to unknown

@mentat9 as discussed please approve the pr to be able to merge

mentat9:
@iliu816, @a-santamaria - I happened to save a link to this PR from office hours, but for future reference your PR is invisible to reviewers until you remove the ARMChangesRequested label.

qiaozha:
@mentat9 could you also approve the suppression? I think ARM reviewers need to sign off this.

mentat9:
> @mentat9 could you also approve the suppression? I think ARM reviewers need to sign off this.

Yes, thanks for the reminder: I overlooked adding the approved suppression label.

PR #35093, opened 2025-06-04
NVAInVnet swagger changes

sbhosalemsft:
# Choose a PR Template


specification/network/resource-manager/Microsoft.Network/stable/2024-10-01/networkVirtualAppliance.json
```diff
@@ -1764,6 +1783,88 @@
...
+      "maxItems": 3,
+      "items": {
+        "$ref": "#/definitions/NvaInterfaceConfigurationsProperties"
+      }
+    },
+    "NvaInterfaceConfigurationsProperties": {
+      "type": "object",
+      "properties": {
+        "subnet": {
+          "readOnly": false,
```
mentat9, line 1801: >type [](http://example.com/codeflow?start=9&length=4)

[ARMBlockingComment] Don't use ARM top-level property names within the property bag. You could use nicType or interfaceType or something else that fits better.

specification/network/resource-manager/Microsoft.Network/stable/2024-10-01/networkVirtualAppliance.json
```diff
@@ -1764,6 +1783,88 @@
...
+                },
+                {
+                  "value": "AdditionalPrivateNic",
+                  "description": "An additional private NIC type"
+                },
+                {
+                  "value": "AdditionalPublicNic",
+                  "description": "An additional public NIC type"
+                }
+              ]
```
mentat9, line 1836: >name [](http://example.com/codeflow?start=9&length=4)

[ARMBlockingComment] Rename to something different from top-level properties (name, type, id, systemData, etc.).

specification/network/resource-manager/Microsoft.Network/stable/2024-10-01/networkVirtualAppliance.json
```diff
@@ -1764,6 +1783,88 @@
...
+        },
+        "name": {
+          "type": "string",
+          "maxLength": 70,
+          "description": "Specifies the name of the interface. Maximum length is 70 characters."
+        }
+      },
+      "description": "Specifies input parameters required NVA in VNet interface configuration."
+    },
+    "NvaInVnetSubnetReferenceProperties": {
```
mentat9, line 1849: >id [](http://example.com/codeflow?start=9&length=2)

[ARMBlockingComment] Rename to something different from top-level properties (name, type, id, systemData, etc.). Recommend something the includes resourceId in the name, like resourceId or subnetResourceId, or something you like better.

sbhosalemsft: Thank you for the feedback  @mentat9 . Referenced resources in our model are currently represented using the ResourceType.Id pattern. For example:

1. "virtualHub": { "id": "/subscriptions/.../virtualHubs/RahulTestHub" }
2. "internetIngressPublicIps": [{ "id": "/subscriptions/.../publicIPAddresses/slbip1" }]

To maintain alignment with this established convention—especially within Hybrid Networking—we’ve applied the same approach for the new subnet reference field. Given that the current naming is consistent with existing patterns and already integrated into the codebase, we’d prefer to retain it as is.

image
image
image
image

specification/network/resource-manager/Microsoft.Network/stable/2024-10-01/networkVirtualAppliance.json
sbhosalemsft: image

Thank you for the feedback. We understand the concern regarding the use of ARM top-level property names. However, in this case, the "type" property is already established in a related configuration(screenshot above) and is used to indicate the NIC type (PublicNic or PrivateNic). The new property being introduced is closely tied to the existing one, and both are interdependent.

To maintain consistency and parity across the API model, we would prefer to retain the name type for this property. This aligns with the current structure and avoids introducing divergence in naming conventions within the same context.

Please let us know if you have any further concerns or suggestions.

specification/network/resource-manager/Microsoft.Network/stable/2024-10-01/networkVirtualAppliance.json
sbhosalemsft: image

Thank you for the feedback. We understand the guidance around avoiding ARM top-level property names. However, the name property is already used in the existing API for related configurations, and the new property being introduced is closely aligned with that structure (screenshot above).

To maintain consistency and parity across the API surface, especially for features that are interdependent, we would prefer to retain the use of name in this context.

Please let us know if you have any concerns with this approach.

sbhosalemsft:
Closing as new base branch was created by Prajwal. Will reopen PR with new base branch

PR #35130, opened 2025-06-05
Remove old emitter from EG config

JoshLove-msft:


raych1:
@JoshLove-msft I would suggest holding off this PR change as `tsp-client` would validate `package-dir` config in the 'tspconfig.yaml'.

jeremymeng (approved):

JoshLove-msft:
I filed https://github.com/Azure/azure-sdk-tools/issues/10823

PR #35131, opened 2025-06-05
Updates for `oav-runner`

scbedd:
- [x] No longer run against files that don't start with `specification/`
- [x] Add new option at startup to pass a file that contains a _list_ of files that were changed. (This is for easy repro)
- [x] Output the list of changed files before starting to parse. To allow easier diagnosis.

### DIFF
--- a/eng/tools/oav-runner/src/cli.ts
+++ b/eng/tools/oav-runner/src/cli.ts
@@ -8,7 +8,9 @@ import {
   ReportableOavError,
 } from "./formatting.js";
 
+import { resolve } from "path";
 import { parseArgs, ParseArgsConfig } from "node:util";
+import { exit } from "node:process";
 
 export async function main() {
   const config: ParseArgsConfig = {
@@ -19,6 +21,12 @@ export async function main() {
         multiple: false,
         default: process.cwd(),
       },
+      fileList: {
+        type: "string",
+        short: "f",
+        multiple: false,
+        default: undefined,
+      },
     },
     allowPositionals: true,
   };
@@ -27,6 +35,24 @@ export async function main() {
   // this option has a default value of process.cwd(), so we can assume it is always defined
   // just need to resolve that here to make ts aware of it
   const targetDirectory = opts.targetDirectory as string;
+
+  let fileList: string[] | undefined = undefined;
+  if (opts.fileList !== undefined) {
+    const fileListPath = resolve(opts.fileList as string);
+    try {
+      const fs = await import('node:fs/promises');
+      const fileContent = await fs.readFile(fileListPath, { encoding: 'utf-8' });
+      fileList = fileContent.split('\n')
+        .map(line => line.trim())
+        .filter(line => line.length > 0);
+      console.log(`Loaded ${fileList.length} files from ${opts.fileList}`);
+    } catch (error) {
+      console.error(`Error reading file list from ${opts.fileList}: ${error instanceof Error ? error.message : String(error)}`);
+      console.error('User provided file list thatis invalid or not found.');
+      console.error("Please ensure the file exists and is readable, or do not provide the option 'fileList'")
+      exit(1);
+    }
+  }
 
   // first positional is runType
   const [runType] = positionals;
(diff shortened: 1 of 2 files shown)


eng/tools/oav-runner/src/cli.ts
```diff
@@ -28,6 +36,24 @@ export async function main() {
...
+    try {
+      const fs = await import('node:fs/promises');
+      const fileContent = await fs.readFile(fileListPath, { encoding: 'utf-8' });
+      fileList = fileContent.split('\n')
+        .map(line => line.trim())
+        .filter(line => line.length > 0);
+      console.log(`Loaded ${fileList.length} files from ${opts.fileList}`);
+    } catch (error) {
+      console.error(`Error reading file list from ${opts.fileList}: ${error instanceof Error ? error.message : String(error)}`);
+      console.error('User provided file list thatis invalid or not found.');
```
scbedd:
```suggestion
      console.error('User provided file list that is not found.');
```

eng/tools/oav-runner/src/cli.ts
```diff
@@ -28,6 +36,24 @@ export async function main() {
   // just need to resolve that here to make ts aware of it
   const targetDirectory = opts.targetDirectory as string;
 
+  let fileList: string[] | undefined = undefined;
+  if (opts.fileList !== undefined) {
+    const fileListPath = resolve(opts.fileList as string);
+    try {
+      const fs = await import('node:fs/promises');
```
mikeharder: nit: move import to top of file

eng/tools/oav-runner/src/cli.ts
```diff
@@ -28,6 +36,24 @@ export async function main() {
...
+      const fileContent = await fs.readFile(fileListPath, { encoding: 'utf-8' });
+      fileList = fileContent.split('\n')
+        .map(line => line.trim())
+        .filter(line => line.length > 0);
+      console.log(`Loaded ${fileList.length} files from ${opts.fileList}`);
+    } catch (error) {
+      console.error(`Error reading file list from ${opts.fileList}: ${error instanceof Error ? error.message : String(error)}`);
```
mikeharder, line 50: I think it's slightly better to wrap and re-throw, like we do here:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/swagger.js#L79
https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/spec-model-error.js

eng/tools/oav-runner/src/runner.ts
```diff
@@ -11,20 +11,30 @@ import {
...
-): Promise<[number, string[], ReportableOavError[]]> {
-  let errors: ReportableOavError[] = [];
-
-  const changedFiles = await getChangedFiles({
-    cwd: rootDirectory,
-  });
+export async function preCheckFiltering(rootDirectory: string, fileList?: string[]): Promise<string[]> {
+  const changedFiles = fileList?.length
+    ? fileList
+    : await getChangedFiles({ cwd: rootDirectory });
```
mikeharder:
```suggestion
  const changedFiles = fileList ?? await getChangedFiles({ cwd: rootDirectory });
```

Shorter and I think sufficient, unless it's important to special-case empty array.

eng/tools/oav-runner/src/runner.ts
```diff
@@ -182,9 +190,18 @@ export async function processFilesToSpecificationList(
           const swaggerModel = new Swagger(
             path.join(rootDirectory, swaggerFile),
           );
-          const exampleSwaggers = await swaggerModel.getExamples();
-          const examples = [...exampleSwaggers].map((e) => e[1].path);
-          cachedSwaggerSpecs.set(swaggerFile, examples);
+          try {
+            const exampleSwaggers = await swaggerModel.getExamples();
+            const examples = [...exampleSwaggers].map((e) => e[1].path);
```
mikeharder:
```suggestion
            const examples = exampleSwaggers.keys()
```

mikeharder (approved):

PR #35145, opened 2025-06-06
Update package path for JS SDKs

v-jiaodi:


MaryGao (approved):

MaryGao:
This is only JS config change and I don't think we need ARM review.

PR #35169, opened 2025-06-09
Evaluation Service FDP cancel delete endpoints

shrutiyer:


specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -51,4 +51,22 @@ interface Evaluations {
...
     ResourceCreatedResponse<AgentEvaluation>
   >;
+
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  @doc("Cancel an evaluation run by name")
+  @route("runs:cancel/{name}")
+  @patch
+  @useAuth(BearerAuth)
+  @added(Versions.v2025_05_15_preview)
+  cancel is Azure.Core.Foundations.Operation<
```
johanste: Should be a ResourceAction<Evaluation, ...>.

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -51,4 +51,24 @@ interface Evaluations {
...
+      Evaluation & {
+        @Azure.Core.pollingLocation()
+        @doc("The location for monitoring the evaluation state.")
+        @TypeSpec.Http.header("operation-location")
+        operationLocation: string;
+      }
+  >;
+
+  @doc("Delete an evaluation run by name")
+  @Azure.Core.pollingOperation(Evaluations.delete)
```
johanste: You can't poll on the location you just deleted since that will return a 404 once the delete is completed. You need a separate status monitor url.

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -68,7 +68,7 @@ model Evaluation {
 
   @doc("Status of the evaluation. It is set by service and is read-only.")
   @visibility(Lifecycle.Read)
-  status?: string;
+  status?: EvaluationStatus;
```
johanste: Under what circumstances would there not be a status returned from the service? Presumably this should be:

```suggestion
  status: EvaluationStatus;
```

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -51,4 +51,24 @@ interface Evaluations {
     },
     ResourceCreatedResponse<AgentEvaluation>
   >;
+
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  @doc("Cancel an evaluation run by name")
+  @post
+  @Azure.Core.pollingOperation(Evaluations.cancel)
+  cancel is EvaluationsOperations.ResourceAction<
```
johanste, line 58: You want to use LongRunningResourceAction if you intend to adhere to azures standard long running operations.

shrutiyer: I think we will be able to avoid LRO and have a simpler user experience for our callers. Have removed this LRO and gone with a simpler implementation

specification/ai/Azure.AI.Projects/evaluations/models.tsp
shrutiyer: Reverted my change. This was left as optional for a different use case where local runs submitted might not have a status. But that is outside the scope of this proposal

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
shrutiyer: This is removed :) I think we will be able to avoid LRO and have a simpler user experience for our callers. Have removed this LRO and gone with a simpler implementation

johanste (approved):

johanste (approved):

PR #35182, opened 2025-06-09
Add script to fix title/descriptions for service Swaggers

matthchr:
While many tools consuming the Swagger specifications _used_ to honor title + description for fields in the `definitions` section, they no longer do. When discussing this with the SDK teams, the consensus was that we should move away from title/description to just putting everything in description, as this is what typespec is eventually going to do as well.

This script automates the title + description -> description migration.


mikeharder (changes requested):
@matthchr: Can you provide more context on this change?  We typically don't commit scripts like this, especially not written in Python (this repo uses JS).

matthchr:
Let me add you to the internal mail thread.

matthchr:
Added you to the internal mail thread. Also updated the PR description to be clearer.

PR #35208, opened 2025-06-10
Fix schema registry tsp

JoshLove-msft:


specification/schemaregistry/SchemaRegistry/main.tsp
```diff
@@ -32,7 +32,7 @@ using Azure.Core;
   "The Schema Registry service endpoint.",
   {
     @doc("The Schema Registry service endpoint, for example 'my-namespace.servicebus.windows.net'.")
-    fullyQualifiedNamespace: url,
+    fullyQualifiedNamespace: string,
```
JoshLove-msft, line 35: I believe this should be a string as we prepend https on L31.

lmazuel (approved):

PR #35210, opened 2025-06-10
Add CRP to the namespace API

nwnt:
Add CRP to the namespace API


specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -159,13 +331,32 @@ model FleetManagedNamespaceStatus {
...
+  @doc("The annotations for the fleet managed namespace.")
+  annotations?: Record<string>;
+
   @visibility("read")
   @doc("The status of the last operation.")
   provisioningState?: FleetManagedNamespaceProvisioningState;
 
-  @visibility("read")
-  @doc("The profile of the fleet managed namespace.")
-  fleetManagedNamespaceProfile?: FleetManagedNamespaceProfile;
```
audrastump, line 168: Why are we removing the managed namespace profile?

Also, currently the swagger parses from the managed namespace profile - I think you'll need to update the code here if you want to remove this layer: https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/fleet/apis/v20250302naar/convert_from_swagger_to_proto.go&version=GBmaster&line=214&lineEnd=215&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +116,204 @@ union EgressPolicy {
...
+  @doc("Using a fleet hub and kubefleet ClusterResourcePlacement.")
+  Hub: "Hub",
+}
 
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Records is the most semantically correct choice for this use-case"
-  @doc("The annotations for the fleet managed namespace.")
-  annotations?: Record<string>;
+@added(Versions.v2025_03_02_naar)
+union OrchestratorPolicy {
+  hubPolicy: HubPolicy,
```
audrastump: I think you need to add a comment above for all fields like this

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: This profile is no longer there on the newly proposed API.  All the fields inside this object have been lifted up.

ryanzhang-oss (commented):
How hard is it to create this PR? My main concerns is how often do we need to sync the CRP API to our namespace API

nwnt:
> How hard is it to create this PR? My main concerns is how often do we need to sync the CRP API to our namespace API

I believe any changes would almost always require a new API version..

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -159,13 +373,32 @@ model FleetManagedNamespaceStatus {
 @doc("A fleet managed namespace.")
 @added(Versions.v2025_03_02_naar)
 model FleetManagedNamespaceProperties {
```
circy9: Could you package all the single cluster namespace properties into one data structure?



FleetManagedNamespaceProperties {
    ManagedNamespaceProperties {
      labels;
      annotations;
      defaultResourceQuota
      defaultNetworkPolicy
      adoptionPolicy
      deletePolicy
    }
}

circy9: Please follow the tradition of other tsp files. We always define data structures top down rather than bottom up.

https://github.com/Azure/azure-rest-api-specs/blob/8a733a2c8bfc1fbb76601e6e6d67566694f4c031/specification/containerservice/Fleet.Management/fleetmember.tsp


Start with something like, and then define the properties.

@doc("A member of the Fleet. It contains a reference to an existing Kubernetes cluster on Azure.")
@resource("members")
@parentResource(Fleet)
model FleetMember is ProxyResource<FleetMemberProperties> {

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+
+  string,
+}
+
 @added(Versions.v2025_03_02_naar)
-model FleetManagedNamespaceProfile {
-  @doc("Action if Kubernetes managed namespace with the same name already exists.")
-  adoptionPolicy?: AdoptionPolicy;
+@doc("The orchestrator profile to manage the lifecycle of the namespace.")
+model OrchestratorProfile {
```
circy9: Orchestrator describes who will do the job (which customers care less about as it is implementation details), rather than how the job will be done.

I would suggest using:

```
PropagationPolicy {
  type: Placement | Direct
  PlacementProfile | DirectProfile
}
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+  @doc("The default ClusterResourcePlacement (CRP) policy configuration.")
+  defaultClusterResourcePlacement: ClusterResourcePlacement;
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("The desired state of ClusterResourcePlacement")
+model ClusterResourcePlacement {
+  @doc("Policy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
+  policy?: ClusterResourcePlacementPolicy;
+}
```
circy9: What is the reason to have so many levels of indirections?

hubpolicy contains one ClusterResourcePlacement which contains one ClusterResourcePlacementPolicy.

Will this work?
```
PropagationPolicy {
  type: Placement | Direct
  PlacementProfile | DirectProfile
}

// This is the same as ClusterResourcePlacementSpec.
PlacementProfile {
  PlacementPolicy // == PlacementPolicy in ClusterResourcePlacementSpec.
}
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+
+  @doc("ClusterNames contains a list of names of MemberCluster to place the selected resources. Only valid if the placement type is PickFixed.")
+  clusterNames?: string[];
 
+  @doc("Affinity contains cluster affinity scheduling rules. Defines which member clusters to place the selected resources. Only valid if the placement type is PickAll.") // don't forget to update this when we add PickN.
+  affinity?: ClusterResourcePlacementAffinity;
+
+  @doc("If specified, the ClusterResourcePlacement's Tolerations. Tolerations cannot be updated or deleted.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  tolerations?: CrpToleration[];
```
circy9: Please make sure we use the same names as the upstream api. It should be Toleration, not CrpToleration. It will be easier to keep the two APIs in sync.

Also avoid using acronyms in field names or data structure names.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: @circy9 we did consider this, and on the one hand, we agree with keeping everything we inherit from AKS in a single struct.  The thing I don't like about that is that in our API, adoptionPolicy and deletePolicy have an effect wider than ManagedNamespaceProperties.  They govern the application of the AKS bits **and also** the CRP, for example.

Another possibility would be to put everything except adoptionPolicy and deletePolicy into ManagedNamespaceProperties.  How would you feel about that?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: I'm fine with PropagationPolicy and Placement.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: The reason for the nesting here is:
a) match the naming/approach for **default**ResourceQuota, **default**NetworkPolicy
b) allow an extension point if in the future we want the placement propagation policy to place something else in addition to or instead of a CRP.

My personal take is that those reasons outweigh the nesting annoyance; WDYT @circy9 ?

ryanzhang-oss:
> > How hard is it to create this PR? My main concerns is how often do we need to sync the CRP API to our namespace API
> 
> I believe any changes would almost always require a new API version..

Let me rephase my question, how many hours would it take me to write this PR without any bug (as it's a perfect copy paste of CRP API)?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+  @doc("matchExpressions is a list of label selector requirements. The requirements are ANDed.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  matchExpressions?: MetaV1LabelSelectorRequirement[];
+}
 
-  @doc("The default network policy for the fleet managed namespace.")
-  defaultNetworkPolicy?: NetworkPolicy;
+@added(Versions.v2025_03_02_naar)
+@doc("A label selector requirement is a selector that contains values, a key, and an operator that relates the key and values.")
+model MetaV1LabelSelectorRequirement {
```
ryanzhang-oss: is there any reason not just call it LabelSelectorRequirement?

jim-minter: Yes, the intention is to fix all of these up.

> Let me rephase my question, how many hours would it take me to write this PR without any bug (as it's a perfect copy paste of CRP API)?

The intention is to get this homogeneous enough such that the process of translating future API updates into this typespec could in principle be subsequently automated.

nwnt: This is now fixed.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: Will rearrange this once we have finalized the API.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: Fixed.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
circy9: There are two relationships:
1. fleetmanagednamespace -> CRP
2. managednamespace -> namespace

Do we think adoptionPolicy and deletePolicy applies to both, and there won't be differences between the two relationships?

If so, we can put those policies out of managednamespace.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+model ClusterResourcePlacement {
+  @doc("PlacementPolicy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
+  policy?: PlacementPolicy;
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("PlacementPolicy contains the rules to select target member clusters to place the selected resources. Only clusters that are both joined and satisfying the rules will be selected. You can only specify at most one of the two fields: ClusterNames and Affinity.")
+model PlacementPolicy {
+  @doc("Type of placement.")
+  placementType?: PlacementType;
```
circy9: Please comment on what the default behavior is for any optional field.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
-  adoptionPolicy?: AdoptionPolicy;
+@doc("The propagation to be used for provisioning the namespace among the fleet.")
+model PropagationPolicy {
+  @doc("The type of the policy to be used.")
+  type: PropagationType;
 
-  @doc("Delete options of a fleet managed namespace.")
-  deletePolicy?: DeletePolicy;
+  @doc("The propagation profile to be used for the propagation.")
+  propagationProfile: PropagationProfile;
```
circy9: Can this be optional?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+}
+
 @added(Versions.v2025_03_02_naar)
-model FleetManagedNamespaceProfile {
-  @doc("Action if Kubernetes managed namespace with the same name already exists.")
-  adoptionPolicy?: AdoptionPolicy;
+@doc("The propagation to be used for provisioning the namespace among the fleet.")
+model PropagationPolicy {
+  @doc("The type of the policy to be used.")
+  type: PropagationType;
```
circy9: Can this be optional?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+union PropagationProfile {
+  @doc("Use ClusterResourcePlacement.")
+  PlacementProfile,
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("The configuration profile for default ClusterResourcePlacement for placement.")
+model PlacementProfile {
+  @doc("The default ClusterResourcePlacement policy configuration.")
+  defaultClusterResourcePlacement: ClusterResourcePlacement;
```
circy9: Can this be optional?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+  @doc("ClusterSelectorTerms is a list of cluster selector terms. The terms are `ORed`.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  clusterSelectorTerms: ClusterSelectorTerm[];
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("ClusterSelectorTerm allows specifying what clusters in the fleet are in scope. If you specify both label and property selectors in the same term, the results are AND'd.")
+model ClusterSelectorTerm {
+  @doc("LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected.")
+  labelSelector?: ClusterLabelSelector;
```
circy9: What is the reason to use ClusterLabelSelector instead of the original LabelSelector?

I could be reused if it is named as LabelSelector as we won't customize the data structure for clusters at all.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+
+  @doc("PropertySelector is a property query over all joined member clusters. Clusters matching the query are selected.")
+  propertySelector?: ClusterPropertySelector;
```
circy9: Same here. Any reason behind adding Cluster as the prefix for the data structure?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+
+  string,
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("PropertySelector helps user specify property requirements when picking clusters for resource placement.")
+model ClusterPropertySelector {
+  @doc("MatchExpressions is an array of PropertySelectorRequirements. The requirements are AND'd.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  matchExpressions: ClusterPropertySelectorRequirement[];
```
circy9: Any reason for adding a Cluster prefix to ClusterPropertySelectorRequirement? But not for PropertySelectorOperator?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -120,28 +121,241 @@ union EgressPolicy {
...
+  @doc("Le (Less than or equal to): a cluster’s observed value of the given property must be less than or equal to the value in the matcher before it can be picked for resource placement.")
+  Le: "Le",
+
+  string,
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("Toleration allows ClusterResourcePlacement to tolerate any taint that matches <key,value,effect> using the matching operator <operator>")
+model Toleration {
+  @doc("Key is the taint key that the toleration applies to.")
```
circy9: Why are we not copying all the comments over? Especially what the default behavior is when it is unset.

Source: https://github.com/Azure/fleet/blob/main/apis/placement/v1/clusterresourceplacement_types.go

specification/containerservice/Fleet.Management/fleetnamespace.tsp
circy9: The current structure LGTM.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: Yes, I do.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,229 @@
+import "@azure-tools/typespec-azure-resource-manager";
```
jim-minter: Please can you make this file include only things that are strict subset of placement.kubernetes-fleet.io/v1 ?

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,229 @@
...
+@doc("The configuration profile for default ClusterResourcePlacement for placement.")
+model PlacementProfile {
+  @doc("The default ClusterResourcePlacement policy configuration.")
+  defaultClusterResourcePlacement?: ClusterResourcePlacement;
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("The desired configuration of ClusterResourcePlacement")
+model ClusterResourcePlacement {
+  @doc("PlacementPolicy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
```
jim-minter: this docstring does not correspond with the underlying API - please can you re-crosscheck them?

nwnt: I used the spec because initially I thought it was going to be like the spec than the full CRP object.  Corrected this now.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
+using TypeSpec.Versioning;
+
+namespace Microsoft.ContainerService;
+
+@added(Versions.v2025_03_02_naar)
+@doc("ClusterResourcePlacement is used to select cluster scoped resources, including built-in resources and custom resources, and placement them onto selected member clusters in a fleet.")
+model ClusterResourcePlacement {
```
jim-minter: I think we should call this ClusterResourcePlacementSpec

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
...
+model ClusterAffinity {
+  @doc("If any affinity requirements specified by this field are not met at scheduling time, the resource will not be scheduled onto the cluster.")
+  requiredDuringSchedulingIgnoredDuringExecution?: ClusterSelector;
+}
+@added(Versions.v2025_03_02_naar)
+@doc("ClusterSelector allows specifying cluster selector terms.")
+model ClusterSelector {
+  @doc("ClusterSelectorTerms is a list of cluster selector terms. The terms are `ORed`.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  clusterSelectorTerms: ClusterSelectorTerm[];
```
jim-minter: some array types are written `foo: Bar[];`, and some are written `foo?: Bar[];` (with a `?`) - is this right?

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
...
+  @doc("If specified, the ClusterClusterResourcePlacement's Tolerations. Tolerations cannot be updated or deleted.")
+  @OpenAPI.extension("x-ms-identifiers", [])
+  tolerations?: Toleration[];
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("PlacementType identifies the type of placement.")
+union PlacementType {
+  @doc("PickAll selects all clusters that match the specified requirements.")
+  PickAll: "PickAll",
```
jim-minter: In the source, the enum name is like PickAllPlacementType (the value is "PickAll").  What happens if the tsp is made to match?

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
...
+  placementType?: PlacementType;
+
+  @doc("ClusterNames contains a list of names of MemberCluster to place the selected resources. Only valid if the placement type is PickFixed.")
+  clusterNames?: string[];
+
+  @doc("Affinity contains cluster affinity scheduling rules. Defines which member clusters to place the selected resources. Only valid if the placement type is PickAll.") // don't forget to update this when we add PickN.
+  affinity?: Affinity;
+
+  @doc("If specified, the ClusterClusterResourcePlacement's Tolerations. Tolerations cannot be updated or deleted.")
+  @OpenAPI.extension("x-ms-identifiers", [])
```
jim-minter: what does this do, and why is it needed?

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
...
+model ClusterResourcePlacement {
+  @doc("PlacementPolicy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
+  policy?: PlacementPolicy;
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("PlacementPolicy contains the rules to select target member clusters to place the selected resources. Only clusters that are both joined and satisfying the rules will be selected. You can only specify at most one of the two fields: ClusterNames and Affinity.")
+model PlacementPolicy {
+  @doc("Type of placement. Default is PickAll.")
+  placementType?: PlacementType;
```
jim-minter: I think Liqian already commented on this, but I wonder why we are making something optional here that is not optional in the source API.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
+using TypeSpec.Versioning;
+
+namespace Microsoft.ContainerService;
+
+@added(Versions.v2025_03_02_naar)
+@doc("ClusterResourcePlacement is used to select cluster scoped resources, including built-in resources and custom resources, and placement them onto selected member clusters in a fleet.")
```
jim-minter: "placement them" is an introduced grammatical error.  Honestly, I would rather we slavishly copy/paste the docstrings for now with no special changes.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
```diff
@@ -0,0 +1,193 @@
...
+  @doc("LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected.")
+  labelSelector?: LabelSelector;
+
+  @doc("PropertySelector is a property query over all joined member clusters. Clusters matching the query are selected.")
+  propertySelector?: PropertySelector;
+}
+
+@added(Versions.v2025_03_02_naar)
+@doc("A label selector is a label query over a set of resources. The result of matchLabels and matchExpressions are ANDed. An empty label selector matches all objects. A null label selector matches no objects.")
+model LabelSelector {
```
jim-minter: this comes from metav1 - is there really no way this can be reflected in our tsp namespacing?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -34,44 +35,51 @@ model FleetManagedNamespace
...
-}
+  @doc("The default network policy for the fleet managed namespace.")
+  defaultNetworkPolicy?: NetworkPolicy;
 
-@doc("The provisioning state of the fleet managed namespace resource")
-union FleetManagedNamespaceProvisioningState {
-  string,
-  ResourceProvisioningState,
+  @doc("Action if Kubernetes managed namespace with the same name already exists.")
+  adoptionPolicy?: AdoptionPolicy;
```
jim-minter: do we want adoptionPolicy and deletePolicy to be optional/defaulted?  Is it better to make them explicit?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -92,84 +100,97 @@ model ResourceQuota {
...
   string,
 }
 
-@doc("The egress policy options for a managed namespace.")
-union EgressPolicy {
-  @doc("Allow all egress traffic from the managed namespace.")
-  AllowAll: "AllowAll",
+@doc("Action if Kubernetes managed namespace with the same name already exists.")
+union AdoptionPolicy {
+  @doc("If k8s managed namespace already exists, do not take over the existing managed namespace to be managed by ARM.")
```
jim-minter: Think I'd globally remove the terms k8s/Kubernetes from docstrings if possible.

also, the object on the k8s side is a Namespace, it is not a managed namespace.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -92,84 +100,97 @@ model ResourceQuota {
...
+  string,
+}
 
-  @doc("The default resource quota for the fleet managed namespace.")
-  defaultResourceQuota?: ResourceQuota;
+@added(Versions.v2025_03_02_naar)
+@doc("The propagation to be used for provisioning the namespace among the fleet.")
+model PropagationPolicy {
+  @doc("The type of the policy to be used. Default is Placement.")
+  type?: PropagationType;
```
jim-minter: questioning the logic on this being optional

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -92,84 +100,97 @@ model ResourceQuota {
...
 }
 
-@doc("A fleet managed namespace.")
 @added(Versions.v2025_03_02_naar)
-model FleetManagedNamespaceProperties {
-  @visibility("read")
-  @doc("The status of the last operation.")
-  provisioningState?: FleetManagedNamespaceProvisioningState;
+@doc("The available Propagation profiles.")
+union PropagationProfile {
```
jim-minter: I don't straightforwardly understand what this is doing - please could you show me a json example?

I think what I'm expecting to see is as follows -- is this what it's doing?

```json
"propagationPolicy": {
  "type": "Placement",
  "placementProfile": {
    "defaultClusterResourcePlacement": {}
  }
}
```

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: It seems this is linted by `@azure-tools`.  Without it, this warning is shown:

>Missing identifying properties of objects in the array item, please add @OpenAPI.extension("x-ms-identifiers", [<prop>]) to specify it. If there are no appropriate identifying properties, please add @OpenAPI.extension("x-ms-identifiers",[]). [@azure-tools/typespec-azure-resource-manager/missing-x-ms-identifiers]

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: Hmm I'm seeing `PlacementType` is optional in the source API:

```go
	// Type of placement. Can be "PickAll", "PickN" or "PickFixed". Default is PickAll.
	// +kubebuilder:validation:Enum=PickAll;PickN;PickFixed
	// +kubebuilder:default=PickAll
	// +optional
	PlacementType PlacementType `json:"placementType,omitempty"`
```
https://github.com/kubefleet-dev/kubefleet/blob/25d684b039ec2b65fd26ddb01afa8048fd6fa6ea/apis/placement/v1/clusterresourceplacement_types.go#L137

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: I actually copied them slavishly because I couldn't come up with anything that would make it better.

https://github.com/kubefleet-dev/kubefleet/blob/25d684b039ec2b65fd26ddb01afa8048fd6fa6ea/apis/placement/v1/clusterresourceplacement_types.go#L43C8-L43C17

nwnt: This is not a concern anymore though.  As suggested by you earlier, I have made the field `ClusterResourcePlacementSpec`, also with its description, so this grammatical error has been removed.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: since the PropagationPolicy field in FleetManagedNamespaceProperties can be optional, wouldn't all its fields be optional using the same values?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: Yes that's what it's doing, but also allowing to add more types to it in the future.  The union expression allows us to construct a "one of" relationship here.  My understanding is we allow only one to be chosen (from https://github.com/Azure/azure-rest-api-specs/pull/35210#discussion_r2145855897).  Let me know if that's not correct.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: The values are still going to be the same.  It looks like TSP just adds more metadata so that code generation can use to name the enum possible values.  I think this is what we want, so this is now updated.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: ? is to denote optionality of the field.

specification/containerservice/Fleet.Management/clusterresourceplacement.tsp
nwnt: done - separated metav1 and corev1 into their own package.

specification/containerservice/Fleet.Management/kubefleet/apis/placement/v1/types.tsp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace KubeFleet.Apis.Placement.V1;
+
+@doc("ClusterResourcePlacementSpec defines the desired state of ClusterResourcePlacement.")
+model ClusterResourcePlacementSpec {
+  @doc("PlacementPolicy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
+  policy?: PlacementPolicy;
+}
+
+@doc("PlacementPolicy contains the rules to select target member clusters to place the selected resources. Only clusters that are both joined and satisfying the rules will be selected. You can only specify at most one of the two fields: ClusterNames and Affinity.")
```
jim-minter: If none is specified, all the joined clusters are selected.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -34,47 +36,54 @@ model FleetManagedNamespace
   ...Azure.ResourceManager.Foundations.ArmTagsProperty;
```
jim-minter, line 36: please check \@key and \@segment above - should not have the fleet prefix IMO

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -34,47 +36,54 @@ model FleetManagedNamespace
...
-}
+  @doc("The default network policy for the fleet managed namespace.")
+  defaultNetworkPolicy?: NetworkPolicy;
 
-@doc("The provisioning state of the fleet managed namespace resource")
-union FleetManagedNamespaceProvisioningState {
-  string,
-  ResourceProvisioningState,
+  @doc("Action if the managed namespace with the same name already exists. Default is Never.")
+  adoptionPolicy?: AdoptionPolicy = "Never";
```
jim-minter:
```suggestion
  adoptionPolicy: AdoptionPolicy;
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -34,47 +36,54 @@ model FleetManagedNamespace
...
 
-  @doc("The provisioning state of a fleet managed namespace being created.")
-  Creating: "Creating",
+  @doc("Delete options of a fleet managed namespace. Default is Keep.")
+  deletePolicy?: DeletePolicy = "Keep";
```
jim-minter:
```suggestion
  deletePolicy: DeletePolicy;
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: No, I think you should either be able to have no PropagationPolicy (nil), or a PropagationPolicy with a type.  PropagationPolicy without a type not allowed.

```suggestion
  type: PropagationType;
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -89,87 +98,100 @@ model ResourceQuota {
...
+@added(Versions.v2025_03_02_naar)
+@doc("The propagation to be used for provisioning the namespace among the fleet.")
+model PropagationPolicy {
+  @doc("The type of the policy to be used. Default is Placement.")
+  type?: PropagationType;
 
-  @doc("The default network policy for the fleet managed namespace.")
-  defaultNetworkPolicy?: NetworkPolicy;
+  @doc("The propagation profile to be used for the propagation. Default is PlacementProfile.")
+  propagationProfile?: PropagationProfile;
```
jim-minter: Still not 100% on how this works.  Please could you generate a Go client from this PR and put it somewhere so that we can see what it looks like?

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: @nwnt this one is still active - need to introduce ManagedNamespaceProperties that does not include adoptionPolicy / deletePolicy

specification/containerservice/Fleet.Management/kubefleet/apis/placement/v1/types.tsp
```diff
@@ -0,0 +1,127 @@
...
+
+using TypeSpec.Versioning;
+using K8s.Io.Api.Core.V1;
+using K8s.ApiMachinery.Apis.Meta.V1;
+
+namespace KubeFleet.Apis.Placement.V1;
+
+@doc("ClusterResourcePlacementSpec defines the desired state of ClusterResourcePlacement.")
+model ClusterResourcePlacementSpec {
+  @doc("PlacementPolicy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
```
jim-minter:
```suggestion
  @doc("Policy defines how to select member clusters to place the selected resources. If unspecified, all the joined member clusters are selected.")
```

specification/containerservice/Fleet.Management/kubefleet/apis/placement/v1/types.tsp
```diff
@@ -0,0 +1,127 @@
...
+
+@doc("PlacementPolicy contains the rules to select target member clusters to place the selected resources. Note that only clusters that are both joined and satisfying the rules will be selected. You can only specify at most one of the two fields: ClusterNames and Affinity. If none is specified, all the joined clusters are selected.")
+model PlacementPolicy {
+  @doc("Type of placement. Can be \"PickAll\", \"PickN\" or \"PickFixed\". Default is PickAll.")
+  placementType?: PlacementType;
+
+  @doc("ClusterNames contains a list of names of MemberCluster to place the selected resources. Only valid if the placement type is PickFixed.")
+  clusterNames?: string[];
+
+  @doc("Affinity contains cluster affinity scheduling rules. Defines which member clusters to place the selected resources. Only valid if the placement type is PickAll.") // don't forget to update this when we add PickN.
```
jim-minter:
```suggestion
  @doc("Affinity contains cluster affinity scheduling rules. Defines which member clusters to place the selected resources. Only valid if the placement type is \"PickAll\" or \"PickN\".")
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: Here it is

https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/12704758

specification/containerservice/Fleet.Management/fleetnamespace.tsp
nwnt: `ManagedNamespaceProperties` can't be used because we renamed `FleetManagedNamespace` to it.  I have instead created a new section called `NamespaceProperties`.

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: Apologies for lack of clarity, I think the model names should be prefixed with Fleet, but I think that key and segment maybe shouldn't be? (see FleetUpdateStrategy)

jim-minter: Using FleetManagedNamespaceProperties at the top level enables us to use ManagedNamespaceProperties for the enclosed struct, which maybe helps us too

specification/containerservice/Fleet.Management/kubefleet/apis/placement/v1/types.tsp
```diff
@@ -0,0 +1,127 @@
...
+  operator: PropertySelectorOperator;
+
+  @doc("Values are a list of values of the specified property which Fleet will compare against the observed values of individual member clusters in accordance with the given operator. At this moment, each value should be a Kubernetes quantity. For more information, see https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity. If the operator is Gt (greater than), Ge (greater than or equal to), Lt (less than), or `Le` (less than or equal to), Eq (equal to), or Ne (ne), exactly one value must be specified in the list.")
+  values: string[];
+}
+
+@doc("PropertySelectorOperator is the operator that can be used with PropertySelectorRequirements.")
+union PropertySelectorOperator {
+  @doc("PropertySelectorGreaterThan dictates Fleet to select cluster if its observed value of a given property is greater than the value specified in the requirement.")
+  PropertySelectorGreaterThan: "Gt",
```
jim-minter:
```suggestion
  Gt: "Gt",
```
...

specification/containerservice/Fleet.Management/fleetnamespace.tsp
```diff
@@ -34,47 +36,62 @@ model FleetManagedNamespace
...
+@added(Versions.v2025_03_02_naar)
+model FleetManagedNamespaceProperties {
+  @visibility("read")
+  @doc("The status of the last operation.")
+  provisioningState?: FleetManagedNamespaceProvisioningState;
 
-  @doc("If there is an identical managed namespace, take over the existing managed namespace to be managed by ARM.")
-  IfIdentical: "IfIdentical",
+  @doc("The namespace properties for the fleet managed namespace.")
+  managedNamespaceProperties: ManagedNamespaceProperties;
```
jim-minter:
```suggestion
  managedNamespaceProperties?: ManagedNamespaceProperties;
```

specification/containerservice/Fleet.Management/fleetnamespace.tsp
jim-minter: Right, I think this should read:

```
@added(Versions.v2025_03_02_naar)
@doc("The propagation to be used for provisioning the namespace among the fleet.")
model PropagationPolicy {
  @doc("The type of the policy to be used. Default is Placement.")
  type: PropagationType;

  @doc("The profile to be used for propagation via placement.")
  placementProfile?: PlacementProfile;
}

@added(Versions.v2025_03_02_naar)
@doc("The configuration profile for default ClusterResourcePlacement for placement.")
model PlacementProfile {
  @doc("The default ClusterResourcePlacement policy configuration.")
  defaultClusterResourcePlacement?: ClusterResourcePlacementSpec;
}
```

jim-minter (approved):

PR #35235, opened 2025-06-11
Client library customization of input param name when creating a Dataset or Index (instead of generic variable name "body")

dargilco:
Before this change, emitted SDKs had:
- Dataset and Index methods named createOrUpdateVersion with an argument named "body".  
- Dataset method named startPendingUploadVersion with an argument named "body".

The first two were defined in the TypeSpec  interface named `VersionedOperations` in the file `servicepatterns.tsp`. The third was also defined in that file (but not part of the interface, since only Datasets needs it).

In order to customize the emitted method argument name, @iscai-msft showed me how to define all the request parameters in a new TypeSpec template named `InputParameters`, including what will become the emitted SDK method argument name. Thank you, Isabella!

This PR:
- Does not change the REST API definition. It just affects the value of the "name" field in the "in: body" section of the swagger. This is not something you see on the wire, it's just a friendly name for the request body. It is used by the SDK emitters to determine the method input variable name.
-  Updates the Dataset and Index operators to use this new template, with a customized SDK argument name.  That takes care of the createOrUpdateVersion methods
- Since the body of the startPendingUploadVersion call is only used by Datasets, I decided to move that out of the template and place it directly in the Datasets folder, so I can have unique docs strings (and of course, a better variable name than "body"). I also removed the definition of the body of the getCredential call from the template into the Dataset, since Dataset is the only one using it. If/when we have a different route that uses these bodies, we can pull that back into the interface.
- I made some minor doc string changes.
- Delete commented out TypeSpec lines associated with "un-versioned" operations. Not sure when and if the service will support this.


jianingwang123:
Hi @dargilco,
According to [link1](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:508),[link2](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:1101),[link3](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:1372),[link4](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:1887),[link5](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:2119),[link6](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:2326),[link7](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:2553),[link8](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:2778),[link9](https://github.com/Azure/azure-rest-api-specs/actions/runs/15614076509/job/43982199596?pr=35235#step:4:3265), your `tspconfig.yaml` is incorrectly configured. Kindly update your `tspconfig.yaml` as per the details provided below.
Please let us know if you need any further assistance.
```
  - Failed to find "options.@azure-tools/typespec-java.package-dir". Please add "options.@azure-tools/typespec-java.package-dir".
  - Failed to find "options.@azure-tools/typespec-ts.package-dir". Please add "options.@azure-tools/typespec-ts.package-dir".
  - Failed to find "options.@azure-tools/typespec-ts.package-details.name". Please add "options.@azure-tools/typespec-ts.package-details.name".
  - Failed to find "options.@azure-tools/typespec-go.generate-fakes". Please add "options.@azure-tools/typespec-go.generate-fakes".
  - Failed to find "options.@azure-tools/typespec-go.inject-spans". Please add "options.@azure-tools/typespec-go.inject-spans".
  - Failed to find "options.@azure-tools/typespec-go.service-dir". Please add "options.@azure-tools/typespec-go.service-dir".
  - Failed to find "options.@azure-tools/typespec-go.package-dir". Please add "options.@azure-tools/typespec-go.package-dir".
  - The value of options.@azure-tools/typespec-python.generate-test "false" does not match "true". Please update the value of "options.@azure-tools/typespec-python.generate-test" to match "true".
  - The value of options.@azure-tools/typespec-python.generate-sample "false" does not match "true". Please update the value of "options.@azure-tools/typespec-python.generate-sample" to match "true".
Kindly refer to the link for more details....
```

iscai-msft (commented):
looks good from a body-parameter renaming angle

iscai-msft:
@JeffreyRichter the breaking change is coming from body-param renames in the swagger, but it doesn't correspond to a breaking change on the pipeline, are you able to help unblock this? Thanks!

JeffreyRichter:
I see an exception from the tool saying that swagger is invalid: https://github.com/Azure/azure-rest-api-specs/pull/35235/checks?check_run_id=44047200631.
You should fix this first before we discuss any specifics.

johanste:
> I see an exception from the tool saying that swagger is invalid: https://github.com/Azure/azure-rest-api-specs/pull/35235/checks?check_run_id=44047200631. You should fix this first before we discuss any specifics.

That is a tool bug: https://github.com/Azure/autorest/blob/e57564401c49ed6eb6085d5c474dceb772f47391/packages/libs/autorest-schemas/example-schema.json#L49. And it is incorrectly complaining about the example, not about the swagger being incorrect. The changes in the PR are not REST API affecting. They are making intended changes to generated code.

dargilco:
@JeffreyRichter based on the new information from @johanste  and @iscai-msft, do you mind taking a 2nd look? Thanks!

JeffreyRichter (approved):

PR #35255, opened 2025-06-12
[Storage] Test Date Common Header Theory

vincenttran-msft:


specification/storage/data-plane/Microsoft.BlobStorage/stable/2025-11-05/blob.json
```diff
@@ -1054,11 +1064,6 @@
                 "type": "string",
                 "description": "Indicates the range of bytes returned in the event that the client requested a subset of the blob by setting the 'Range' request header."
               },
-              "Date": {
-                "type": "string",
-                "format": "date-time-rfc7231",
-                "description": "UTC date/time value generated by the service that indicates the time at which the response was initiated"
-              },
```
catalinaperalta: @vincenttran-msft some of the operations dont fit into the template, so they're defined using plain tsp. Doing the mass removal of the data header seems like it affected this operation. Please revert for this operation.

specification/storage/data-plane/Microsoft.BlobStorage/stable/2025-11-05/blob.json
```diff
@@ -1411,11 +1416,6 @@
...
```
catalinaperalta: Same comment as above, should be reverted

PR #35259, opened 2025-06-13
Enable SdkTspConfigValidation in TypeSpecValidation and Suppress Existing Failures

wanlwanl:
# Background
## Terms
- TypeSpecValidation: A GitHub pull request check that used to validate typespec related specifications, including `tspconfig.yaml`.
- TypeSpecValidation-All: A Github regular check that used to validate typespec related specifications, including `tspconfig.yaml`.
- SdkTspConfigValidation: A rule to validation emitter options and parameters in `tspconfig.yaml`, makes sure they are present and corect.

## Azure SDK Policy
According to [Azure SDK consistent guideline](https://azure.github.io/azure-sdk/general_introduction.html#consistent) - "**There should be feature parity across target languages. This is more important than feature parity with the service**."

# Objective
- For releasing SDK, we require the configurations for all 5 tier 1 languages (Java, Python, .NET, JS and GO) to be **present** and **corrent** in the tspconfig.yaml file for management plane and dataplane. The purpose of validating configurations for all code emitters is to identify the spec quality issue that can lead to SDK generation failure **as early as possible**.

# Problem
- This spec repo already has incorrect or missing emitter options or emitters in `tspconfig.yaml` for many specifications.
- Enabling `SdkTspConfigValidation` rule without any handling will cause `TypeSpecValidation-All` fail, which is not allowed by EngSys team.

# Solution
- This pull request enable the `SdkTspConfigValidation` rule to making sure the configurations for all 5 tier 1 languages (Java, Python, .NET, JS and GO) to be **present** and **corrent** in the tspconfig.yaml file for management plane and dataplane. 
- This pull request also adds suppressions for existing specifications that has missing or incorrect emitter options or parameters in `tspconfig.yaml` to make `TypeSpecValidation-All` pass for regular monitoring.

# Limitation
- The major limitation in this PR is to make `TypeSpecValidation-All` pass for regular monitoring while `TypeSpecValidation` failed in pull requests to make sure service team has chance to be aware the problem and then fix it. To make `TypeSpecValidation-All` pass in regular monitoring and `TypeSpecValidation` fail in pull requests, we need to wait for the [Add property "if" to scope suppressions](https://github.com/Azure/azure-rest-api-specs/issues/35153) feature get supported.  

# Next Steps
- Support [Add property "if" to scope suppressions](https://github.com/Azure/azure-rest-api-specs/issues/35153) feature to remove the major limitation
- Guadually remove suppressions for the updated `tspconfig.yaml` once the spec authors get errors for `tspconfig.yaml` about emitter options and parameters and fix them when they update specs in pull requests


specification/suppressions.yaml
mikeharder, line 1: This is a **lot** of suppressions, and I'm not sure it's worth suppressing each sub-rule.  How about just suppressing all of `SdkTspConfigValidation` for every existing spec not currently passing?  Or, maybe we suppress at the language level (so you can suppress all of `go`), but not list each sub-rule within each language?

mikeharder (changes requested):
Blocking until we have consensus agreement on the design

specification/suppressions.yaml
timotheeguerin: Writing my comments here too

I assume that if a service get an exemption to release a SDK with the current form they'd ahve to suppress 1 rule per field that was supposed to be provided? If so I do think we really need at least be able to disable per for a whole emitter otherwise thoses service will keep accumlating suppressions as more field get added


and given we do that I think this PR could be simplified to have
	1.	Whole emitter suppression if the emitter was already not defined
	2.	Keep specific rule suppression if the values are just missing/wrong

catalinaperalta: I havent been fully tuned into the discussions about these validations, but should some of these validations be suppressed if they are missing/wrong? Will teams still have to fix these values whenever they open new prs (how will they know what needs to be fixed?) or will this hide the issues in tspconfig.yaml settings?

wanlwanl:
First of all, the most important thing in this PR is to enable `SdkTspconfigValiation` rule to make sure the new coming spec cannot check in problematic tspconfig. I wonder if there's changes to unblock this PR sooner.

Actually, I don't care about the sub rules if https://github.com/Azure/azure-rest-api-specs/issues/35153 feature is supported.
The blocker to me is we cannot distingush PR or non-PR. We want to make sure service team get notified that validation in tspcofig is failed in new PR. Details about which option is violating is not a must. We have to change the suppression to `TypeSpecValidationAll` tool with the scope feature (#35153) anyway. Suppression by emitter level is useless then.
 
For now, since we can only add suppression for `TypeSpecValidation` tool, I have to make the suppression as detailed as possible based on existing features. Suppress on emitter level is rare, since service team should release all 5 languages according to azure sdk's policy, I don't think it worth support.

mikeharder:
> First of all, the most important thing in this PR is to enable `SdkTspconfigValiation` rule to make sure the new coming spec cannot check in problematic tspconfig. I wonder if there's changes to unblock this PR sooner.

I am (mostly) OOF through Weds, back on Thu.  Hope to implement #35153 by end of week.

@wanlwanl: An option for you could be to enable these as permanent suppressions first (so they are ignored for both TSVAll and PRs), then scope them to only TSVAll once #35153 is done.

> For now, since we can only add suppression for `TypeSpecValidation` tool, I have to make the suppression as detailed as possible based on existing features. Suppress on emitter level is rare, since service team should release all 5 languages according to azure sdk's policy, I don't think it worth support.

The number of suppressions in this PR is unmanageable, both for the file as a whole, and even for a single spec.  This may be a signal, that we are signing up spec owners for a lot of work they may not want to do. There are 1509 files in suppressions.yaml, meaning at least 1509 "work items" for spec owners to remove them.  Are we planning to help spec authors get to green on this, like we are funding TSP conversions for existing specs?  We may need to, for this to be successful.

Looking at the suppressions you added, a few suggestions:

1. In your rule, make `sub-rules` be a list of regexes instead of matching the exact string.  Then you suppress say all rules for `go` by suppressing `@azure-tools/typespec-go*`.

2. For awsconnector, you can also use globs for the paths (I believe), so instead of listing each sub-folder, use `awsconnector/*/tspconfig.yaml`.

3. How are the suppressions grouped?  At the top of the file I see multiple unrelated specs grouped under suppressions, then later in the file, it appears to be one section per spec?  I think we should standardize on one suppression per "org name", the top-level folder under "specification".

mikeharder:
> > First of all, the most important thing in this PR is to enable `SdkTspconfigValiation` rule to make sure the new coming spec cannot check in problematic tspconfig. I wonder if there's changes to unblock this PR sooner.
> 
> I am (mostly) OOF through Weds, back on Thu. Hope to implement #35153 by end of week.

@wanlwanl: An option for you could be to enable these as permanent suppressions first (so they are ignored for both TSVAll and PRs), then scope them to only TSVAll once #35153 is done.

wanlwanl:
> Are we planning to help spec authors get to green on this, like we are funding TSP conversions for existing specs? We may need to, for this to be successful

Tried to help spec authors, but you already rejected. Since I have to add place holders which you and Wes didn't like.

wanlwanl:
> In your rule, make sub-rules be a list of regexes instead of matching the exact string. Then you suppress say all rules for go by suppressing @azure-tools/typespec-go*.

Eventually, the suppression in the central suppression filed will be removed, and Azure SDK policy does not allow not release a specific languanges unless it is granted. Which means this scenario is pretty rare. I don't see the value to implement this for long term.

wanlwanl:
> How are the suppressions grouped? At the top of the file I see multiple unrelated specs grouped under suppressions, then later in the file, it appears to be one section per spec? I think we should standardize on one suppression per "org name", the top-level folder under "specification".

If a `tspconfig.yaml` doesn't add all emitters in `options`, then it will be a single item. 
The later one are the other cases, e.g. part of the options is missing, which cannot merge into the above item.

wanlwanl:
> or awsconnector, you can also use globs for the paths (I believe), so instead of listing each sub-folder, use awsconnector/*/tspconfig.yaml.

What if `awsconnector` adds a new SDK? I don't want to suppress the new SDK in PR.
If I add `awsconnector/*/tspconfig.yaml`, when a new SDK is in PR to check in like `awsconnector/NEW/tspconfig.yaml`, then it's also get suppressed, which is unexpected.

weshaggard:
>What if awsconnector adds a new SDK? I don't want to suppress the new SDK in PR.
If I add awsconnector/*/tspconfig.yaml, when a new SDK is in PR to check in like awsconnector/NEW/tspconfig.yaml, then it's also get suppressed, which is unexpected.

awsconnector is example where I don't believe we plan to release any SDK's and so this will likely be a permanent suppression for all those. Have we actually looked at what you are adding here to see which ones should translate to work and which ones are just noise? I know our goal is to have all the SDK's but that isn't a reality for all these.

>Tried to help spec authors, but you already rejected. Since I have to add place holders which you and Wes didn't like.

Injecting placeholders is not helping them it is just moving the work from a suppression to invalid config in the tspconfig.yaml file. We will need to come up with a plan for how we expect spec authors to do this work or do the real work for them.

Based on the suppressions do we have any idea of how many tspconfig files are actually passing? It would seem like the percentage is pretty low and if that is the case it seems like we might be setting up partners teams for a high amount of work.

I do like @mikeharder's suggestion about enabling some wildcard support to help reduce the cost and noise of maintaining this.

specification/suppressions.yaml
```diff
@@ -54,3 +54,977 @@
     - sphere/resource-manager/Microsoft.AzureSphere/preview/2022-09-01-preview/azuresphere.json
     - splitio/resource-manager/SplitIO.Experimentation/preview/2024-03-01-preview/splitio.json
     - translation/data-plane/Azure.AI.DocumentTranslation/preview/2023-11-01-preview/openapi.json
+- tool: TypeSpecValidation
```
mikeharder: I think suppressions should be organized as one group per spec.  Unrelated specs shouldn't share suppressions.

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -624,4 +623,23 @@ export class SdkTspConfigValidationRule implements Rule {
...
+
+  private isKeyPathSuppressed(keyPath: string): boolean {
+    // Direct match
+    if (this.suppressedKeyPaths.has(keyPath)) {
+      return true;
+    }
+
+    // Only check for wildcard at the end (format: prefix.*)
+    for (const suppressedPath of this.suppressedKeyPaths) {
+      if (suppressedPath.endsWith('.*')) {
```
mikeharder, line 635: Why use a custom special-case syntax instead of regex?

mikeharder (approved):
Approved, pending the following:

1. Suppressions are grouped by spec
2. PR updating RPSaaSMaster is ready to merge at the same time as this one.
3. Change rule to use regex instead of special-casing `.*` (can be in a separate PR).

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
skywing918: The validation is dealing with configuration key paths (like options.emitter.property). A simple prefix matching with .* suffix is much safer than allowing arbitrary regex patterns that could be complex to validate and potentially dangerous.

skywing918:
> Approved, pending the following:
> 
> 1. Suppressions are grouped by spec
> 2. PR updating RPSaaSMaster is ready to merge at the same time as this one.
> 3. Change rule to use regex instead of special-casing `.*` (can be in a separate PR).

for track: https://github.com/Azure/azure-rest-api-specs/issues/35365

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
wanlwanl: @mikeharder Actually, I don't want it to be permanent, since ignore a language in releasing sdk is not what we want.
I don't want to courage people to suppress the specific entire emitter options when it's not necessary. I don't think reviewer will not add make sure the suppression is minimal. Could you add me to codeowner to every spec's suppression file and this suppression file as well?

I'd like to remove this feature in the future, and let service team add exactly the option they want to suppress.

specification/suppressions.yaml
```diff
@@ -2,9 +2,10 @@
 # Remove suppressions as specs are fixed
 
 - tool: TypeSpecValidation
+  reason: Remove specs contain incorrect or missing emitter options or parameters
```
mikeharder:
```suggestion
```

specification/suppressions.yaml
```diff
@@ -3,8 +3,8 @@
 
 - tool: TypeSpecValidation
   rules: [Compile]
-  sub-rules: [ExtraSwagger]
-  reason: Extra swaggers left behind when versions were removed from *.tsp files
+  sub-rules:
+    - ExtraSwagger
```
mikeharder:
```suggestion
  sub-rules: [ExtraSwagger]
  reason: Extra swaggers left behind when versions were removed from *.tsp files
```

mikeharder (approved):

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
mikeharder: There is nothing to validate and no security boundary here.  It's simpler and clearer to use an existing language like Regex, than inventing a custom language.  All other tools in EngSys use Regex for this.

specification/suppressions.yaml
```diff
@@ -54,3 +54,2293 @@
...
+- tool: TypeSpecValidation
+  reason: Remove specs contain incorrect or missing emitter options or parameters
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-go.package-dir
+    - options.@azure-tools/typespec-java.namespace
+    - options.@azure-tools/typespec-ts.package-details.name
+    - options.@azure-tools/typespec-ts.package-dir
+  paths: [widget/resource-manager/Microsoft.Widget/Widget/tspconfig.yaml]
+- tool: TypeSpecValidation
```
mikeharder, line 2323: How do we need suppressions in the v2 widget sample that was just added?  @wanlwanl: Please create a PR updating the widget sample to remove these suppressions.

specification/suppressions.yaml
```diff
@@ -54,3 +54,2293 @@
...
+    - options.@azure-tools/typespec-ts.*
+    - parameters.service-dir.default
+  paths: [containerstorage/ContainerStorage.Management/tspconfig.yaml]
+- tool: TypeSpecValidation
+  reason: Remove specs contain incorrect or missing emitter options or parameters
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-go.flavor
+    - options.@azure-tools/typespec-java.namespace
+  paths: [contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml]
```
mikeharder, line 1652: @wanlwanl: Same comment on this sample as widget.

specification/suppressions.yaml
skywing918: Track By: https://github.com/Azure/azure-rest-api-specs/issues/35389

specification/suppressions.yaml
skywing918: Track By: https://github.com/Azure/azure-rest-api-specs/issues/35389

PR #35294, opened 2025-06-16
Confirm new oav version

scbedd:
I'm expecting this to pass, as the new version of oav `3.6.1` is now consumed in production devops task + repo `package.json`.


scbedd:
Looking good.

PR #35325, opened 2025-06-17
Extended change file code to return values by categories

raych1:
- This change retrieves the status of modified files, which will be handled differently during breaking change checks.
- Added `scenario` filter.
- Added tests.

### DIFF
--- a/.github/shared/src/changed-files.js
+++ b/.github/shared/src/changed-files.js
@@ -238,6 +238,20 @@ export function swagger(file) {
     typeof file === "string" &&
     json(file) &&
     (dataPlane(file) || resourceManager(file)) &&
-    !example(file)
+    !example(file) &&
+    !scenario(file)
+  );
+}
+
+/**
+ * @param {string} [file]
+ * @returns {boolean}
+ */
+export function scenario(file) {
+  return (
+    typeof file === "string" &&
+    json(file) &&
+    specification(file) &&
+    file.includes("/scenarios/")
   );
 }
--- a/.github/shared/src/changed-files.js
+++ b/.github/shared/src/changed-files.js
@@ -52,7 +52,7 @@ export async function getChangedFiles(options = {}) {
  * @param {import('./logger.js').ILogger} [options.logger]
  * @returns {Promise<{additions: string[], modifications: string[], deletions: string[], renames: {from: string, to: string}[], total: number}>}
  */
-export async function getCategorizedChangedFiles(options = {}) {
+export async function getChangedFilesStatuses(options = {}) {
   const {
     baseCommitish = "HEAD^",
     cwd,
--- a/.github/shared/src/changed-files.js
+++ b/.github/shared/src/changed-files.js
@@ -71,78 +71,81 @@ export async function getCategorizedChangedFiles(options = {}) {
       modifications: /** @type {string[]} */ ([]),
       deletions: /** @type {string[]} */ ([]),
       renames: /** @type {{from: string, to: string}[]} */ ([]),
-      total: 0
+      total: 0,
     };
 
     if (result.trim()) {
       const lines = result.trim().split("\n");
-      
+
       for (const line of lines) {
         const parts = line.split("\t");
         const status = parts[0];
-        
+
         switch (status[0]) {
-          case 'A':
+          case "A":
             categorizedFiles.additions.push(parts[1]);
             break;
-          case 'M':
+          case "M":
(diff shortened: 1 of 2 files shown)


.github/shared/src/changed-files.js
```diff
@@ -44,6 +44,125 @@ export async function getChangedFiles(options = {}) {
...
 
+/**
+ * @param {Object} [options]
+ * @param {string} [options.baseCommitish] Default: "HEAD^".
+ * @param {string} [options.cwd] Current working directory.  Default: process.cwd().
+ * @param {string} [options.headCommitish] Default: "HEAD".
+ * @param {import('./logger.js').ILogger} [options.logger]
+ * @returns {Promise<{additions: string[], modifications: string[], deletions: string[], renames: {from: string, to: string}[], total: number}>}
+ */
+export async function getCategorizedChangedFiles(options = {}) {
```
weshaggard:
```suggestion
export async function getChangedFilesStatuses(options = {}) {
```

.github/shared/src/changed-files.js
```diff
@@ -44,6 +44,125 @@ export async function getChangedFiles(options = {}) {
...
+    const categorizedFiles = {
+      additions: /** @type {string[]} */ ([]),
+      modifications: /** @type {string[]} */ ([]),
+      deletions: /** @type {string[]} */ ([]),
+      renames: /** @type {{from: string, to: string}[]} */ ([]),
+      total: 0,
+    };
+
+    if (result.trim()) {
+      const lines = result.trim().split("\n");
```
weshaggard, line 78: While I don't like all the parsing I'm OK with this approach for now but we still might consider adding a option to the existing getChangedFiles to filter to the different types.

weshaggard (approved):

.github/shared/src/changed-files.js
raych1: Sounds good.

.github/shared/src/changed-files.js
```diff
@@ -122,6 +241,20 @@ export function swagger(file) {
...
+    !example(file) &&
+    !scenario(file)
+  );
+}
+
+/**
+ * @param {string} [file]
+ * @returns {boolean}
+ */
+export function scenario(file) {
```
mikeharder, line 253: @raych1: What is the purpose of the "scenarios" folder?  I'm not aware of an official use in folder structure v1 or v2.

.github/shared/src/changed-files.js
```diff
@@ -44,6 +44,125 @@ export async function getChangedFiles(options = {}) {
...
 
+/**
+ * @param {Object} [options]
+ * @param {string} [options.baseCommitish] Default: "HEAD^".
+ * @param {string} [options.cwd] Current working directory.  Default: process.cwd().
+ * @param {string} [options.headCommitish] Default: "HEAD".
+ * @param {import('./logger.js').ILogger} [options.logger]
+ * @returns {Promise<{additions: string[], modifications: string[], deletions: string[], renames: {from: string, to: string}[], total: number}>}
+ */
+export async function getChangedFilesStatuses(options = {}) {
```
mikeharder, line 55: Do you really needed the statuses?  I was hoping we could simplify by just using a flat list of changed files.

If you do need statuses, instead of adding another, we should probably update the existing API to return statuses.

I will refactor this in a later PR.

raych1: @mikeharder 
This is [the place](https://github.com/Azure/azure-rest-api-specs/pull/35353/files#diff-1ee83d4d127867a6377ccbcc8cdc38f4a9d8a938e4d958f1c20938a22613bbedR52) in breaking change tool to get changed files in statuses. Based on different statues, the tool needs to create dummy file for comparison.

As a shared library, it would be better to return them in different lists instead of requiring each caller to parse the result.

We still have other usages which don't care about the status. I'd rather keep both APIs but extending the first API to take an option of status filter.

.github/shared/src/changed-files.js
raych1: It was used by the scenario test which we've retired. However, there're still some files haven't been cleaned up, such as [this one](https://github.com/Azure/azure-rest-api-specs/tree/4cb330be555a994ed507acf739e3349b7cb76d85/specification/appconfiguration/resource-manager/Microsoft.AppConfiguration/stable/2022-05-01/scenarios).

Do you think we just help service team clean them up or leave them as is? From tooling perspective, breaking change tool doesn't need to run against these files.

PR #35379, opened 2025-06-19
Return all status changes in the change file function

### LINKED ISSUE #10907: [SDK Validation] Code generation is skipped when a spec PR only contain deleted files
This is the default behavior in identifying the change files of a spec PR - any deletion files will be ignored.

We might need to include the deletion files as any file deletion could impact the code generated.

[Sample spec PR](https://github.com/Azure/azure-rest-api-specs/pull/35288)

raych1:
SDK Validation is supposed to run with any kind of changes, such as added, copied, modified, deleted, renamed, and type changed files.

The second commit is for test purposes and will be reverted.


eng/tools/spec-gen-sdk-runner/src/utils.ts
```diff
@@ -177,8 +177,9 @@ export function getChangedFiles(
   specRepoPath: string,
   baseCommitish: string = "HEAD^",
   targetCommitish: string = "HEAD",
-  diffFilter: string = "d",
 ): string[] | undefined {
+  // set diff filter to include added, copied, modified, deleted, renamed, and type changed files
+  const diffFilter = "ACMDRT";
   const scriptPath = path.resolve(specRepoPath, "eng/scripts/ChangedFiles-Functions.ps1");
```
mikeharder, line 183: Should this be updated to use changed-files.js?

mikeharder (approved):

rkmanda:
dont have much context for the change.. but assuming its related to SDK tooling, I am approving it

eng/tools/spec-gen-sdk-runner/src/utils.ts
raych1: I will make this change when we adopt the spec-shared library.

PR #35386, opened 2025-06-20
do not suppression for contosowidgetmanager and widget

skywing918:
# Choose a PR Template


eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -149,7 +149,7 @@ class TspconfigEmitterOptionsSubRuleBase extends TspconfigSubRuleBase {
 }
 
 function isManagementSdk(folder: string): boolean {
-  return folder.includes(".Management");
+  return folder.includes("/resource-manager/") || folder.includes(".Management");
```
mikeharder, line 152: Good catch, this is needed for all specs using "folder structure v2"

specification/suppressions.yaml
```diff
@@ -1643,13 +1643,6 @@
     - options.@azure-tools/typespec-ts.*
     - parameters.service-dir.default
   paths: [containerstorage/ContainerStorage.Management/tspconfig.yaml]
-- tool: TypeSpecValidation
-  reason: Remove specs contain incorrect or missing emitter options or parameters
-  rules: [SdkTspConfigValidation]
-  sub-rules:
-    - options.@azure-tools/typespec-go.flavor
-    - options.@azure-tools/typespec-java.namespace
```
mikeharder, line 1651: Why was this suppressed, if the code had it specified?  Did we have a bug in a tool we used to auto-add these suppressions?

Typically, we ignore unused suppressions, but maybe for SdkTspConfigValidation, you should fail on them, since we have so many now?

mikeharder (approved):

PR #35398, opened 2025-06-20
Set model-namespace to true

JoshLove-msft:
The default value for model-namespace is false in the new generator.

### DIFF
--- a/specification/keyvault/Security.KeyVault.Administration/tspconfig.yaml
+++ b/specification/keyvault/Security.KeyVault.Administration/tspconfig.yaml
@@ -38,6 +38,7 @@ options:
     include-api-view-properties: false
   "@azure-typespec/http-client-csharp":
     namespace: Azure.Security.KeyVault.Administration
+    model-namespace: true
   "@azure-tools/typespec-csharp":
     package-dir: "Azure.Security.KeyVault.Administration"
     namespace: "Azure.Security.KeyVault.Administration"


heaths (changes requested):
This is a breaking change for the existing SDK that has already GA'd:

https://learn.microsoft.com/en-us/dotnet/api/azure.security.keyvault.administration

JoshLove-msft:
> This is a breaking change for the existing SDK that has already GA'd:
> 
> https://learn.microsoft.com/en-us/dotnet/api/azure.security.keyvault.administration

This is a new generator that we are switching to. The default in the old generator was true.

heaths:
Okay, but the models aren't in a separate namespace now. So either it never worked, or somehow we were working around that and this new setting may break it. @JonathanCrd can you look into this?

Of course, I'd feel a lot better if APIView would've created one for .NET above. I only see JS and Python right now. 😕

JoshLove-msft:
> Okay, but the models aren't in a separate namespace now. So either it never worked, or somehow we were working around that and this new setting may break it. @JonathanCrd can you look into this?
> 
> Of course, I'd feel a lot better if APIView would've created one for .NET above. I only see JS and Python right now. 😕

Yeah, it looks like every public model was customized to be in the root namespace. The net effect of having model-namespace as true is that the models will be generated under the Models directory. We will make sure there are no API changes when switching generators.

JoshLove-msft:
> Of course, I'd feel a lot better if APIView would've created one for .NET above. I only see JS and Python right now. 😕

This is still a work in progress. I am just setting up the config so that it can be used with the new generator. For now, it is still using the old one.

heaths (approved):
Talked offline. Sounds like this could be broken if the subsequent change isn't made very soon, though. /cc @JonathanCrd

JonathanCrd:
> Yeah, it looks like every public model was customized to be in the root namespace. 

Correct, public models are being customized on the convenience layer.

> The net effect of having model-namespace as true is that the models will be generated under the Models directory.

Can you elaborate on this? As you can see [here](https://github.com/JonathanCrd/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/Models), the generated models in KeyVault Administration are being generated in a `/Models` directory already.

JoshLove-msft:
> Can you elaborate on this? As you can see [here](https://github.com/JonathanCrd/azure-sdk-for-net/tree/main/sdk/keyvault/Azure.Security.KeyVault.Administration/src/Generated/Models), the generated models in KeyVault Administration are being generated in a `/Models` directory already.

Yes, I want to retain this behavior when switching to the new generator. The default value for this setting has flipped from the old generator (true) to the new generator(false). Hence in order to make the migration as seamless as possible I am setting the config to true.

PR #35423, opened 2025-06-23
Updating the union so the two fields we need to rename are explicitly named

richardpark-msft:
The clientName decorator can't be applied to unnamed fields within a union. This change gives them a name (the same as they would have, normally), which allows them to be renamed like any other type.


lmazuel (approved):

jhendrixMSFT (approved):

richardpark-msft:
@Kishp01, @rajeshka, or @shankarsama can one of you approve this? It doesn't change your API service, it's needed to enable a rename I'm doing for Go.

PR #35427, opened 2025-06-23
Update uri to url for API Management properties

samvaity:


lmazuel (approved):

specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/propertyNameOverrideCsharp.tsp
```diff
@@ -252,3 +252,110 @@ using Microsoft.EventGrid.SystemEvents;
 
 @@clientName(AcsRouterWorkerSelector.state, "SelectorState", "csharp");
 @@clientName(AcsRouterWorkerSelector.value, "labelValue", "csharp");
+
+@@clientName(ApiManagementUserCreatedEventData.resourceUri,
```
srnagar: Do we need to repeat these in csharp when we already have these overrides in the common file which only excludes `go`?

JoshLove-msft: We would need to either repeat them here or also exclude csharp in the common file.

lmazuel (approved):

PR #35443, opened 2025-06-24
Update copilot instructions and prompts

praveenkuttappan:
Update prompts and copilot instructions to provide value in expected format and also to add additional instruction in namespace approval step.


.github/prompts/typespec-to-sdk.prompt.md
```diff
@@ -103,7 +103,8 @@ Your goal is to guide user through the process of generating SDKs from TypeSpec
 1. Run `/create-release-plan`
 2. If SDK PRs exist, link them to the release plan
 3. Run `/sdk-details-in-release-plan` to add languages and package names to the release plan
-4. If TypeSpec project is for management plane, Run `/verify-namespace-approval` to check package namespace approval
+4. If TypeSpec project is for management plane, Run `/verify-namespace-approval` to check package namespace approval.
+This step should not check package readiness to verify namespace approval for management plane SDK.
```
maririos, line 107: not sure i follow this direction. You mean on the release readiness the verify approval for management plane shouldn't run?
would it be better to move this to that step? or the mcp method description itself?

maririos (approved):

.github/prompts/typespec-to-sdk.prompt.md
praveenkuttappan: Good point. We cannot move to MCP description because it is triggered by agent that runs the step to verify namespace approval and `verify-namespace-approval` is a prompt and not MCP. We have already added namespace approval as skipped in MCP that checks for package readiness. Agent got confused and used package readiness instead of using verify namespace approval prompt.

PR #35463, opened 2025-06-25
mgmt, signalr, convert to TypeSpec

weidongxu-microsoft:

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind. This is the list you need to pay attention:

**1 things remain on REST**
- `text/json` is not added to content-type, as it is untypical for ARM CreateOrUpdate to use this content-type.

**1 thing remains for some SDK, it does not affect REST**
- Path parameter order change is not customized in SDK. This has no impact on REST.

doc https://tspwebsitepr.z22.web.core.windows.net/typespec-azure/prs/2700/docs/migrate-swagger/01-get-started/

---

# Summary of changes

breakdown from [Swagger BreakingChange](https://github.com/Azure/azure-rest-api-specs/pull/35463/checks?check_run_id=46703257843)

## 1003 - RequestBodyFormatNoLongerSupported

`text/json` is removed from API `consumes`. All standard ARM API should consume `application/json`.

## 1017 - ReferenceRedirection

This is equivalent in Swagger, hence benign.
old
```
      "properties": {
        "type": {
          "enum": [
            "ThrottleByJwtSignatureRule",
            "ThrottleByUserIdRule",
            "ThrottleByJwtCustomClaimRule"
          ],
          "type": "string",
          "x-ms-enum": {
            "name": "ClientConnectionCountRuleDiscriminator",
            "modelAsString": true
          }
        }
      },
```
new
```
      "properties": {
        "type": {
          "$ref": "#/definitions/ClientConnectionCountRuleDiscriminator"
        }

    "ClientConnectionCountRuleDiscriminator": {
      "type": "string",
      "enum": [
        "ThrottleByJwtSignatureRule",
        "ThrottleByUserIdRule",
        "ThrottleByJwtCustomClaimRule"
      ],
      "x-ms-enum": {
        "name": "ClientConnectionCountRuleDiscriminator",
        "modelAsString": true,
      ...
    }
```

## 1023 - TypeFormatChanged

One case of this is due to `@secret` decorator on e.g.
```
  /**
   * The primary access key.
   */
  @secret
  primaryKey?: string;
```
The generated Swagger would contain `"format": "password"`, along with `"x-ms-secret": true`. I assume this be benign.

Another case of this is that `nextLink` in paged result now be `"format": "uri"`. This is ARM standard.

## 1025 - RequiredStatusChange

`value` in paged result must be "required". This is ARM standard.

## 1029 - ReadonlyPropertyChanged

This is false alert. `provisioningState` is `readonly` before/after. Nothing changes.

## 1034 - AddedRequiredProperty

Same as 1025, `value` in paged result must be "required".

## 1042 - ChangedParameterOrder

This does not affect REST API.
SDK may want to check later whether we want to mitigate in client.tsp (again, this won't affect REST API)


weidongxu-microsoft:
@TD2106 
Can you take a quick look on the migration of SignalR, from Swagger to TypeSpec? SDK dev is helping with the migration.
Let us know your opinion on below 2 items (they are included in PR description).

1. `List_Operations` is changed to use [shared operation from typespec-azure-resource-manager](https://github.com/Azure/azure-rest-api-specs/blob/a389e270849f2ff1b96c21d7d2e7de91fcf53757/specification/signalr/SignalRService.Management/main.tsp#L52). Some model was removed (e.g. `OperationProperties`), as a result.
2. `text/json` is not added to content-type, as it is untypical for ARM CreateOrUpdate to use this content-type.

For item (1), if [`OperationProperties` property/schema](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json#L2630-L2633) is of importance, I can update the TypeSpec to include it; if not, we can use the standard `Operations_List` from common lib.

For item (2), personally I see no reason to include `text/json` as an input for ARM CreateOrUpdate (all API uses `application/json`). I didn't see any other resource doing this.

TD2106:
@bjqian can you give some inputs for the above comment from @weidongxu-microsoft ?

weidongxu-microsoft:
Hi @bjqian,

Can you give me some feedback?

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
```diff
@@ -2,92 +2,138 @@
...
-        ],
+        "description": "Checks that the resource name is valid and is not already in use.",
         "parameters": [
           {
-            "in": "path",
-            "name": "location",
-            "description": "The region",
-            "required": true,
-            "type": "string"
+            "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/ApiVersionParameter"
```
bjqian, line 110: The parameter order has been changed. Isn't that a breaking change?
Should we keep the original order?

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
```diff
@@ -102,63 +148,50 @@
...
-            "type": "string"
-          },
           {
             "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/ApiVersionParameter"
           },
           {
             "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/SubscriptionIdParameter"
+          },
+          {
+            "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/LocationParameter"
```
bjqian, line 161: Is this change to common type required?

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
```diff
@@ -257,18 +284,25 @@
             "$ref": "../../../../../common-types/resource-management/v5/types.json#/parameters/ResourceGroupNameParameter"
           },
           {
-            "$ref": "#/parameters/ResourceName"
+            "name": "resourceName",
```
bjqian, line 287: Why do we flattern this parameter?

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
weidongxu-microsoft: It is not a break on REST API level. The path/query parameters is on URL, and the URL would be same, regardless of the order of Swagger parameter definition.

On SDK side, we are checking whether customization in client.tsp would make SDK unchanged, WIP https://github.com/Azure/azure-rest-api-specs/pull/36279 <- Done, merged back to this PR.

I will also check with Powershell/CLI about path param order (I'd assume CLI depends on Python SDK). If they still generate from Swagger, and the order matters, we may also use directive in readme.md to customize the order.

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
weidongxu-microsoft: This is by using standard LocationParameter in arm lib.
```
  list is ArmProviderActionSync<
    Response = SignalRUsageList,
    Scope = SubscriptionActionScope,
    Parameters = LocationParameter
  >;
```
and Swagger version of it be https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json#L656-L664

Swagger before/after should be same, except `description` and one more constraint of `minLength=1` (which is expected).

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
weidongxu-microsoft: This should be equivalent (though I indeed do not know how typespec-autorest generate this part).

The API uses standard arm lib template
```
  get is ArmResourceRead<SignalRResource>;
```
And this parameter comes from here https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/models.tsp#L35

weidongxu-microsoft:
When I am skimming over LintDiff errors, this one appears to be genuine (API definition be same in main branch, without x-ms-pageable).

##  XmsPageableForListCalls

- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json#L2084-L2129
- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json#L2084-L2127

specification/signalr/resource-manager/Microsoft.SignalRService/preview/2025-01-01-preview/signalr.json
weidongxu-microsoft: Confirmed from Powershell dev, the order does not matter for them.

bjqian (approved):

ArthurMa1978 (approved):

PR #35465, opened 2025-06-25
Mongodbatlas to use ArmCustomPatchAsync for patch method

kazrael2119:


specification/liftrmongodb/MongoDB.Atlas.Management/main.tsp
```diff
@@ -58,6 +58,34 @@ model UserDetails {
   companyName?: string;
 }
 
+@doc("Base patch resource properties that can be extended for arm resources.")
+model BaseResourcePatchProperties {
+  @doc("Marketplace details of the resource.")
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  marketplace?: MarketplaceDetails;
```
weidongxu-microsoft: Why you have a PatchProperties, and its property is visibility=Create+Read (no Update)?

weidongxu-microsoft (commented):
Other change LGTM

msyyc (commented):
@kazrael2119 Please ping me and @ChenxiJiang333 to regenerate SDK after this PR merged.

specification/liftrmongodb/MongoDB.Atlas.Management/main.tsp
kazrael2119: this is copied from https://github.com/Azure/azure-rest-api-specs/pull/35465/files#diff-f9252771918abdd5ff741d0976288e6f4ce3d21bbb445f77ff63b419eb2897bcR90-R101
```
@doc("Base resource properties that can be extended for arm resources.")
model BaseResourceProperties {
  @doc("Marketplace details of the resource.")
  marketplace: MarketplaceDetails;

  @doc("Details of the user.")
  user: UserDetails;

  @doc("Provisioning state of the resource.")
  @visibility(Lifecycle.Read)
  provisioningState?: ResourceProvisioningState;
}
```
and update all the properties as optional

kazrael2119:
> @kazrael2119 Please ping me and @ChenxiJiang333 to regenerate SDK after this PR merged.

got it

specification/liftrmongodb/MongoDB.Atlas.Management/main.tsp
weidongxu-microsoft: I am asking why a property that exclude visibility=Update appears in PATCH only model with a visibility include Create.

Should we remove the `Lifecycle.Create` on it and make it read-only?

tadelesh (commented):
I think we could switch to use `ArmCustomPatchAsync` without split the model.

specification/liftrmongodb/MongoDB.Atlas.Management/main.tsp
MaryGao: Yes, I think this makes sense, we could simply remove this property for Update method.

weidongxu-microsoft: We would need to still have it + Read, as it would be in response of PATCH.

MaryGao:
@weidong @tadelesh Can we directly approve swagger breaking considering this may not be real REST API breakings?

https://github.com/Azure/azure-rest-api-specs/pull/35465/checks?check_run_id=44816380186

weidongxu-microsoft:
I think if service can approve, we can add a benign label.

The Swagger does have some diff, particularly `OrganizationResourceUpdate` and its `allOf`. (it is not easy to say "they are equivalent")

weidongxu-microsoft (approved):
LGTM on SDK

kazrael2119:
@ajaykumarmehra could you help review this pr? this pr is to split create request model and update request model

MaryGao (approved):
Approved for JS.

sanar-microsoft (approved):
:shipit:

PR #35483, opened 2025-06-26
Summary Action

scbedd:
FYI @mikeharder @weshaggard for feedback as this is going.


.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,49 @@
...
+    workflows:
+    - "\\[TEST-IGNORE\\] Swagger SemanticValidation - Set Status"
+    - "\\[TEST-IGNORE\\] Swagger ModelValidation - Set Status"
+    - "\\[TEST-IGNORE\\] Swagger Avocado - Set Status"
+    - "Swagger LintDiff - Set Status"
+    - "SDK Validation Status"
+    types: [completed]
+  pull_request_target: # when a PR is labeled. NOT pull_request, because that would run on the PR branch, not the base branch.
+    types:
+      - labeled
```
mikeharder, line 18:
```suggestion
      - labeled
      - unlabeled
```

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  run-summarize-checks:
+    name: "[TEST-IGNORE] Summarize Checks"
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
mikeharder:
```suggestion
```

Default fetch depth of 1 should be sufficient.

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Setup Node and install deps
```
mikeharder: For performance, I think we should try to implement this workflow in pure JS under .github/workflows/src.  Example:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/arm-auto-signoff.yaml
https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff.js

scbedd: Ok. I'll refactor.

scbedd: Ok refactor mostly complete. I'm populating code, adding unit tests, developing based on an integration test that checks for existence of `GITHUB_TOKEN` 👍

.github/workflows/summarize-checks.yaml
scbedd: Just stole your sparse checkout from arm-auto-signoff.

.vscode/launch.json
```diff
@@ -22,6 +22,24 @@
       "request": "launch",
       "name": "Launch Program",
       "program": "${file}"
+    },
+    {
```
scbedd: Will revert before commit.

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          script: |
+            const { default: summarizeChecks } =
+              await import('${{ github.workspace }}/.github/workflows/src/summarize-checks.js');
+            return await summarizeChecks({ github, context, core });
+        env:
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+          ISSUE_NUMBER: ${{ github.event.pull_request.number || github.event.workflow_run.pull_requests[0].number }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.workflow_run.head_sha }}
```
mikeharder:
```suggestion
```

I think you should be able to remove this.  Some of our existing workflows have it, because they also support the "workflow_dispatch" trigger, to manually run the workflow with arbitrary parameters.  This was added to enable live testing, but in practice it's usually better to test with PRs (in your fork if needed).  So I'm going to remove most of the workflow_run triggers.

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Only needs .github folder for automation, not the files in the PR (analyzed in a
+          # separate workflow).
+          sparse-checkout: |
+            .github
+
+      # depending on the triggering event, the PR number and SHA will be in different places
```
mikeharder: `extractInputs()` should handle this

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,55 @@
...
+    name: "[TEST-IGNORE] Summarize Checks"
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Only needs .github folder for automation, not the files in the PR (analyzed in a
+          # separate workflow).
+          sparse-checkout: |
+            .github
```
mikeharder:
```suggestion
      # *** IMPORTANT ***
      # For workflows that are triggered by the pull_request_target event, the workflow runs in the
      # context of the base of the pull request.  You should make sure that you do not check out,
      # build, or run untrusted code from the head of the pull request.
      - uses: actions/checkout@v4
        with:
          # Only needs .github folder for automation, not the files in the PR (analyzed in a
          # separate workflow).
          #
          # Uses the .github folder from the PR base branch (pull_request_target trigger),
          # or the repo default branch (other triggers).
          sparse-checkout: |
            .github
```

NIT: Please copy this exact step including the warning comments.  It's very important for security in priviliged workflows.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,88 @@
...
+  let issue_number = parseInt(process.env.ISSUE_NUMBER || "");
+  let head_sha = process.env.HEAD_SHA || "";
+
+  if (!owner || !repo || !issue_number || !head_sha) {
+    let inputs = await extractInputs(github, context, core);
+    owner = owner || inputs.owner;
+    repo = repo || inputs.repo;
+    issue_number = issue_number || inputs.issue_number;
+    head_sha = head_sha || inputs.head_sha;
+  }
```
mikeharder:
```suggestion
  const { owner, repo, head_sha, issue_number } = await extractInputs(
    github,
    context,
    core,
  );
```

I think this should be sufficient.  Copied from here:

https://github.com/Azure/azure-rest-api-specs/blob/a13f38e7991f2135a9d4eef3533a4aa4f83897ea/.github/workflows/src/set-status.js#L26-L30

.github/workflows/test/summarize-checks.test.js
mikeharder, line 1: Existing workflows only use unit tests with mocking, but feel free to add (optional) integration tests if you want.

package-lock.json
mikeharder, line 1: revert

.github/workflows/summarize-checks.yaml
scbedd: Got it. Just the comment. Yeah no worries.

I read the step a couple times and was like

> Mike that's an exact clone

.github/workflows/test/summarize-checks.test.js
scbedd: Going to test in integration, then move to fork. I won't merge the integration tests to main.

package-lock.json
scbedd: Yeah will do so in one last cleanup before this merges.

.github/workflows/src/label-rules.js
```diff
@@ -0,0 +1,864 @@
...
+ *
+ * For additional context, see:
+ * - https://gist.github.com/raych1/353949d19371b69fb82a10dd70032a51
+ * - https://github.com/Azure/azure-sdk-tools/issues/6327
+ * - https://microsoftapc-my.sharepoint.com/:w:/g/personal/raychen_microsoft_com/EbOAA9SkhQhGlgxtf7mc0kUB-25bFue0EFbXKXS3TFLTQA
+ */
+
+export const sdkLabels = {
+  "azure-cli-extensions": {
+    breakingChange: undefined,
```
scbedd: These breaking change labels are also present at `eng/tools/openapi-diff-runner/src/types/breaking-change.ts`. That's a compiled solution though. I'm betting I need to reverse the locoations, so breaking-change.ts refers to an external file in `.github/shared` to get these breaking change label names.

.github/workflows/src/label-rules.js
```diff
@@ -0,0 +1,864 @@
...
+ */
+export async function getPresentBlockingLabelsAndMissingRequiredLabels(
+  { github, context, core },
+  repo,
+  owner,
+  existingLabels,
+  targetBranch,
+) {
+  // this is a little bit of a strange looking branch format, but not going to touch this for now.
+  const repoTargetBranch = `${repo}/${targetBranch}`;
```
scbedd: todo: anywhere that I'm passing in this different format of `branch`, I'm going to rename the argument to `repoTargetBranch`.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,854 @@
...
+
+// #endregion
+// #region core
+/**
+ * @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments
+ * @returns {Promise<void>}
+ */
+export default async function summarizeChecks({ github, context, core }) {
+  logGitHubRateLimitInfo({ github, context, core });
+  let { owner, repo, issue_number, head_sha } = await extractInputs(github, context, core);
```
scbedd: I have a code clarity question for you @mikeharder . Here, we obviously return the github values of `issue_number` and `head_sha` because that's the naming convention in github.

Would you prefer that I maintain this naming convention as I flow these values throughout the script? OR would you prefer I map them to the intended js-appropriate name. Eg `issue_number` -> `prNumber`. `head_sha` -> `sha`. `event_name` -> `eventName`.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,854 @@
...
+// automated merge requirements met by from the result of and(requiredChecks).
+// if any are pending, automated merging requirements is pending. This is ripe for complete removal
+// in favor of just honoring the `required` checks results directly.
+/** @type {string[]} */
+const EXCLUDED_CHECK_NAMES = [];
+
+// #endregion
+// #region core
+/**
+ * @param {import('github-script').AsyncFunctionArguments} AsyncFunctionArguments
```
scbedd: No matter what I do `import('github-script')` is always highlighted red by my local tscheck. No matter what I install I can't get it to resolve. Weird.

Using `AsyncFunctionArguments` for simplicity any time we need one of them.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,854 @@
...
+ * @returns {string} The GraphQL query string.
+ */
+function getGraphQLQuery(owner, repo, sha, prNumber) {
+  const resourceUrl = `https://github.com/${owner}/${repo}/commit/${sha}`;
+
+  return `
+    {
+      resource(url: "${resourceUrl}") {
+        ... on Commit {
+          checkSuites(first: 20) {
```
weshaggard: Does this call need paging? If so, does the underlying call handle that automatically?

scbedd: You know. I didn't actually check. I will note that this is the exact code that is running successfully in `pipelineBot`, but this is probably just because we're < 20 checksuites and < 30 checkruns regardless anyways.

Will do some follow-up and come up with an answer for you.

scbedd: Ok. As I suspected, no, `graphql` doesn't page automatically. I checked and was able to run `first: 100` without any appreciable query difference (because we're already getting all the checks).

I'll take a look at how straightforward it would be to add paging. If it's reasonable complexity I'll just add it instead of possibly having a time bomb on our hands.

.github/workflows/summarize-checks.yaml
```diff
@@ -0,0 +1,53 @@
...
+    # todo, any others need to be added here?
+    workflows:
+      - "\\[TEST-IGNORE\\] Swagger SemanticValidation - Set Status"
+      - "\\[TEST-IGNORE\\] Swagger ModelValidation - Set Status"
+      - "Swagger Avocado - Set Status"
+      - "Swagger LintDiff - Set Status"
+      - "SDK Validation Status"
+    types: [completed]
+  pull_request_target: # when a PR is labeled. NOT pull_request, because that would run on the PR branch, not the base branch.
+    types:
```
mikeharder, line 17: Probably want to run on opened/synchronized/reopened as well.  Example:

https://github.com/Azure/azure-rest-api-specs/blob/517eaf1cca58813605768f4ddc9a59ca75173493/.github/workflows/lintdiff-status.yaml#L8-L16

.github/workflows/src/summarize-checks.js
mikeharder: This works for me:

1. Run "npm i" under ".github"
2. Open folder ".github" (not repo root) in VSCode

image

scbedd: Ok now that's resolving and it's erroring elsewhere 👍

.github/workflows/src/summarize-checks.js
mikeharder: This is a tricky problem.  You can look at existing code like "arm-auto-signoff.js".  I think it's fairly consistent, but I didn't create explicit naming rules for this.

I think my pattern was basically:

1. If we're interacting with the GH APIs, use `snake_case`
2. Else, use `camelCase`

One thing I did try to do, is use the same words but just different case.  So `issue_number` and `issueNumber` should mean the same thing, but I wouldn't assume `prNumber` and `issue_number` are the same.

https://github.com/Azure/azure-rest-api-specs/blob/517eaf1cca58813605768f4ddc9a59ca75173493/.github/workflows/src/arm-auto-signoff.js#L49-L54

I don't think it's worth nitpicking this either way.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+      owner,
+      repo,
+      issue_number,
+    });
+    /** @type {{ name: string }[]} */
+    const labels = resp.data;
+    /** @type {string[]} */
+    labelNames = labels.map((label) => label.name);
+  } catch (error) {
+    core.error(
```
mikeharder: What if we don't catch, and just let the exception bubble up?  Is this good enough?

scbedd: It'll be roughly the same experience from the action side. I'm not opposed.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+  /** @type {CheckRunData[]} */
+  let fyiCheckRuns = [];
+
+  core.info(
+    `Handling ${event_name} event for PR #${issue_number} in ${owner}/${repo} with targeted branch ${targetBranch}`,
+  );
+
+  // no matter what, we will need the labels for the PR, so let's fetch them first
+  try {
+    const resp = await github.rest.issues.listLabelsOnIssue({
```
mikeharder: Add pagination, like this:

https://github.com/Azure/azure-rest-api-specs/blob/517eaf1cca58813605768f4ddc9a59ca75173493/.github/workflows/src/arm-auto-signoff.js#L66-L72

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+    core.error(
+      `Failed to obtain labels from GitHub API. prNumber: ${issue_number}, headOid: ${head_sha}, error: '${error}'. `,
+    );
+    process.exit(1);
+  }
+
+  // handle our label trigger first, we may bail out early if it's a label action we're reacting to
+  // this also implies that if a label action is performed before any workflows complete, we shouldn't
+  // accidentally update the next steps to merge with the results of the workflows that haven't completed yet.
+  if (event_name in ["labeled", "unlabeled"]) {
```
mikeharder: Do we need different behavior based on the event?  Typically, it's easier if a workflow can have the same behavior regardless of the event (even if this means doing a little unnecessary work).

scbedd: For now, I'm handling label/unlabel events atomically in this step. When I pull across `PR Summary` it'll be solely in this case.

There is technically no reason to make this an early exit during label event. I can label it, then examine the label set to see if there are any violating label rules.

I was just trying to mime the existing logic, not truly opposed to this.

.github/workflows/src/summarize-checks.js
scbedd: Incoming commit.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+  owner,
+  repo,
+  issue_number,
+  event_name,
+  known_labels,
+) {
+  // logic for this event is based on code directly ripped from pipelinebot:
+  // private/openapi-kebab/src/bots/pipeline/pipelineBotOnPRLabelEvent.ts
+  // todo: further enhance with labelling actions from `PR Summary` check.
+  const changedLabel = context.payload.label?.name;
```
mikeharder: Does it matter which label changed?  Or can we compute the new set of labels from the existing set, regardless of which event triggered this run?

scbedd: I believe we could compute the new set of labels based on the existing set yes. We will have to choose an order to add or removal labels but I haven't dug through `PR Summary` yet. I'll know for certain when we get there.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+  excludedCheckNames,
+) {
+  // This function was originally a version of getRequiredAndFyiAndAutomatedMergingRequirementsMetCheckRuns
+  // but has been simplified for clarity and purpose.
+  /** @type {CheckRunData[]} */
+  let reqCheckRuns = [];
+  /** @type {CheckRunData[]} */
+  let fyiCheckRuns = [];
+
+  const response = await github.graphql(getGraphQLQuery(owner, repo, head_sha, prNumber));
```
mikeharder: Why GraphQL instead of REST?

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,856 @@
...
+  }
+
+  // At this point we expect a completed run, so conclusion should be defined
+  const conclusion = checkRun.conclusion?.toLowerCase();
+  if (conclusion == null) {
+    return undefined;
+  }
+
+  // Return true for success or neutral, false for any other conclusion
+  return conclusion === "success" || conclusion === "neutral";
```
mikeharder: I don't like treating "neutral" as "success", but if the existing code does it, I suppose we should keep it.

.github/workflows/src/summarize-checks.js
```diff
@@ -0,0 +1,827 @@
...
+    "<ul>" +
+    buildFailingChecksNextStepsText(failingFyiChecksInfo, "FYI") +
+    "</ul>"
+  );
+}
+
+/**
+ * Builds next steps text for failing checks
+ * @param {CheckMetadata[]} failingChecks - Array of failing checks
+ * @param {"required" | "FYI"} checkKind - Kind of check (required or FYI)
```
mikeharder: Could make this an "enum" if you care enough:

https://github.com/Azure/azure-rest-api-specs/blob/e877ba67daddbd28a0f868cded185a6ee395fa76/.github/workflows/src/label.js#L3-L11

.github/workflows/src/label-rules.js
mikeharder, line 1: label-rules.js, summarize-checks.js, and tsgs.js all seem closely related.  Should we create a folder `src/summarize-checks` to group these?  comment.js seems to stand alone.

mikeharder (approved):

.github/workflows/src/summarize-checks.js
scbedd: This was the most straightforward way as recommended by @weshaggard. IIRC there is no REST api to get check suites, just runs.

.github/workflows/src/summarize-checks.js
scbedd: Honestly. This is just AI being more restrictive than it actually matters. It's just a string that gets filled into the template.

.github/workflows/src/label-rules.js
scbedd: Yeah. I'll put them all in the same folder, but leave comment.js out.

.github/workflows/src/summarize-checks.js
mikeharder: https://docs.github.com/en/rest/checks?apiVersion=2022-11-28

We use "List check runs for a Git reference" (below), so I suspect "List check suites for a Git reference" would also work.

I'm not opposed to GraphQL, other than all our existing code in this repo uses REST.  One the one hand, it might be good to get our feet wet with GraphQL.  On the other hand, I would prefer to use one or the other as consistently as possible in our code for ease of maintenance.

You can merge this PR with GraphQL, but I might submit a PR converting it to REST, if I think it looks better.

https://github.com/Azure/azure-rest-api-specs/blob/89a7e31c277506592a7d4d273b852ca5a5cc324c/.github/workflows/src/github.js#L156

PR #35532, opened 2025-06-27
New Features for 2025-06-01-preview API version:

melissamserv:
New Features for **2025-06-01-preview** API version:

-  Support for URIs and IP addresses in SAN properties: 
   Add the **uris** and **ipAddresses** attributes to **Subject Alternative Names** for certificate requests.

-  PFX to PEM conversion in Get Certificate:
   Add the **outContentType** query parameter to **Get Certificate** operation to specify the format of the certificate in the response.

- Include the version of the previous certificate in secret responses:
  Add the **previousVersion** attribute to the **SecretBundle** in the response.


heaths:
Isn't this a dup of #35534? If so, please close dups so we're not looking at both.

PR #35589, opened 2025-07-01
Add unbranded generator config

JoshLove-msft:


specification/ai/Azure.AI.Projects/tspconfig.yaml
```diff
@@ -33,6 +33,8 @@ options:
     namespace: Azure.AI.Projects
     model-namespace: false
     api-version: "2025-05-15-preview"
+  "@typespec/http-client-csharp":
+    api-version: "2025-05-15-preview"
```
jorgerangel-msft, line 38: do we need `package-name: Azure.AI.Projects` ?

jorgerangel-msft (approved):

jorgerangel-msft (approved):

mikeharder (approved):
Approved, but why is C# different than other languages?

JoshLove-msft:
> Approved, but why is C# different than other languages?

My understanding is that all languages should be splitting into branded and unbranded emitters. At that point, the flavor config should be deprecated/removed.

PR #35591, opened 2025-07-01
Extended SpecModel shared code to support additional metadata and functions for swagger

raych1:
- ~~added a new class called `SwaggerVersionCache` to construct swagger metadata within a RP folder~~
- ~~added tests~~

### Enhancements to `SpecModel` and `Swagger` classes:

* **`SpecModel` updates**:
  - Added a new method `getSwaggers()` to retrieve swagger files from tags and input files.

* **`Swagger` updates**:
  - Introduced private fields (`#operations`, `#version`, `#versionKind`, ~~`#fileName`~~) to store swagger metadata and operations. 
  - Added methods `getOperations()`, `version`, `versionKind`, and ~~`fileName`~~ to fetch swagger operations and metadata.

### Utility functions for swagger processing:

  - Added ~~`getBaseNameForSwagger()`~~ and `getVersionFromInputFile()` to extract version details and base names from swagger file paths.
  - ~~Introduced `getPrecedingSwaggers()` and `getLegacyVersionOperations()` to manage swagger lifecycle stages and retrieve operations from previous versions.~~ 

### Test:
  - Added test fixtures and comprehensive tests.


.github/shared/src/path.js
mikeharder, line 1: Please move these methods to swagger-version-cache.js.  path.js should only be used for simple path-like utilities, not code specific to spec-model.

raych1: Moved.

raych1:
@mikeharder can you review again?

.github/shared/src/spec-model.js
```diff
@@ -183,6 +183,13 @@ export class SpecModel {
     return this.#readmes;
   }
 
+  async getSwaggers() {
+    const readmes = [...(await this.getReadmes()).values()];
```
mikeharder, line 187: I created a follow-on PR to improve this: https://github.com/Azure/azure-rest-api-specs/issues/35670

.github/shared/src/swagger.js
```diff
@@ -2,16 +2,30 @@
 
 import $RefParser, { ResolverError } from "@apidevtools/json-schema-ref-parser";
 import { readFile } from "fs/promises";
-import { dirname, relative, resolve } from "path";
+import { basename, dirname, relative, resolve } from "path";
 import { mapAsync } from "./array.js";
 import { includesFolder } from "./path.js";
 import { SpecModelError } from "./spec-model-error.js";
+import { existsSync, readFileSync } from "fs";
```
mikeharder: Use async APIs from fs/promises.

.github/shared/src/swagger.js
```diff
@@ -2,16 +2,30 @@
...
+import { API_VERSION_LIFECYCLE_STAGES } from "./breaking-change.js";
```
mikeharder: breaking-changes.js can depend on swagger.js, but we shouldn't have deps in the other direction.  Move this enum into swagger.js.

.github/shared/src/swagger.js
```diff
@@ -2,16 +2,30 @@
...
 
+/**
+ * @typedef {Object} Operation
+ * @property {string} id - The operation ID
+ * @property {string} path - API path
+ * @property {string} httpMethod - HTTP method (GET, POST, etc.)
+ */
+
+/**
+ * @typedef {Operation & {swagger: string}} LegacyOperation
```
mikeharder: Can we find a better name for this?  What is a "LegacyOperation"?  Should it instead be a pair of operations, like "before" and "after"?

.github/shared/src/swagger.js
```diff
@@ -42,6 +56,18 @@ export class Swagger {
   /** @type {Tag | undefined} Tag that contains this Swagger */
   #tag;
 
+  /** @type {Operation[]} map of the operations in this swagger*/
+  #operations;
```
mikeharder, line 53: This should probably be `Map<string, Operation>` instead of `Operation[]`, where the key is some unique string from the Operation, maybe the id?  Using a map guarantees no duplicate IDs, and also enables O(1) lookup from an ID.

.github/shared/src/swagger.js
```diff
@@ -42,6 +56,18 @@ export class Swagger {
...
+
+  /** @type {string}  swagger version*/
```
mikeharder:
```suggestion
  /** @type {string} swagger version */
```

.github/shared/src/swagger.js
```diff
@@ -42,6 +56,18 @@ export class Swagger {
...
+  #version;
+
+  /** @type {string} version kind*/
```
mikeharder:
```suggestion
  /** @type {string} version kind */
```

.github/shared/src/swagger.js
```diff
@@ -42,6 +56,18 @@ export class Swagger {
...
+  #versionKind;
+
+  /** * @type {string} file name of the swagger file, e.g. "LongRunningBackend.json" */
+  #fileName;
```
mikeharder: Does this need to be a separate property, or can it be extracted from the existing #path property?

.github/shared/src/swagger.js
```diff
@@ -51,6 +77,12 @@ export class Swagger {
   constructor(path, options) {
     const rootDir = dirname(options?.tag?.readme?.path ?? "");
     this.#path = resolve(rootDir, path);
+    this.#operations = [];
```
mikeharder: Initialize this to undefined instead of [].

.github/shared/src/swagger.js
```diff
@@ -51,6 +77,12 @@ export class Swagger {
...
+    this.#version = getVersionFromInputFile(this.#path) || "";
+    this.#versionKind = dirname(this.#path).includes("/preview/")
+      ? API_VERSION_LIFECYCLE_STAGES.PREVIEW
+      : API_VERSION_LIFECYCLE_STAGES.STABLE;
```
mikeharder, line 84: These require async IO so must be layz-loading async helper methods.

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
   return typeof file === "string" && file.toLowerCase().endsWith(".json");
 }
+
+/**
+ * Get base name for swagger file
+ * @param {string} filePath - Path to swagger file
+ * @param {string} [version] - Version string to use for base name
+ * @returns {string} - Base name for the swagger file
+ */
+function getBaseNameForSwagger(filePath, version = "") {
```
mikeharder: I don't understand this.  Why does getting the base name from a swagger (which points to a single file) take a version?

.github/shared/src/swagger.js
```diff
@@ -118,6 +150,61 @@ export class Swagger {
     return filtered;
   }
 
+  /**
+   * @returns {Promise<Operation[]>}
+   */
+  async getOperations() {
+    if (this.#operations.length === 0) {
```
mikeharder:
```suggestion
    if (!this.#operations) {
```

.github/shared/src/swagger.js
```diff
@@ -118,6 +150,61 @@ export class Swagger {
...
+              }
+            }
+          }
+        }
+
+        // Sort operations by id and store in instance variable
+        this.#operations.sort((a, b) => a.id.localeCompare(b.id));
+      } catch {
+        throw new SpecModelError(`Failed to read operations from swagger: ${this.#path}`);
+      }
```
mikeharder:
```suggestion
      } catch {
        throw new SpecModelError(`Failed to read operations from swagger: ${this.#path}`);
      }
```

Just let the exception bubble.

.github/shared/src/swagger.js
```diff
@@ -132,6 +219,27 @@ export class Swagger {
     return this.#tag;
   }
 
+  /**
+   * @returns {string} swagger version
+   */
+  get version() {
```
mikeharder:
```suggestion
  async getVersion() {
```

.github/shared/src/swagger.js
```diff
@@ -132,6 +219,27 @@ export class Swagger {
...
+    return this.#version;
+  }
+
+  /**
+   * @returns {string} version kind (stable or preview)
+   */
+  get versionKind() {
```
mikeharder, line 232:
```suggestion
  async getVersionKind() {
```

.github/shared/src/swagger.js
```diff
@@ -42,6 +56,18 @@ export class Swagger {
...
   #tag;
 
+  /** @type {Operation[]} map of the operations in this swagger*/
+  #operations;
+
+  /** @type {string}  swagger version*/
+  #version;
+
+  /** @type {string} version kind*/
+  #versionKind;
```
mikeharder: Property is unnecesary, since this can be computed on-demand from #path.

.github/shared/src/swagger.js
```diff
@@ -51,6 +77,12 @@ export class Swagger {
   constructor(path, options) {
     const rootDir = dirname(options?.tag?.readme?.path ?? "");
     this.#path = resolve(rootDir, path);
+    this.#operations = [];
+    this.#version = getVersionFromInputFile(this.#path) || "";
```
mikeharder: This requires async IO so must be lazy-loading async helper methods.

.github/shared/src/swagger.js
```diff
@@ -132,6 +219,27 @@ export class Swagger {
...
+   */
+  get version() {
+    return this.#version;
+  }
+
+  /**
+   * @returns {string} version kind (stable or preview)
+   */
+  get versionKind() {
+    return this.#versionKind;
```
mikeharder:
```suggestion
    return dirname(this.#path).includes("/preview/")
      ? API_VERSION_LIFECYCLE_STAGES.PREVIEW
      : API_VERSION_LIFECYCLE_STAGES.STABLE;
```

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+  return basename(filePath);
+}
+
+/**
+ * Extract version string from input file path
+ * @param {string} filePath - Path to the input file
+ * @param {boolean} [withPreview=false] - Whether to include preview suffix
+ * @returns {string | undefined} - Version string extracted from path
+ */
+export function getVersionFromInputFile(filePath, withPreview = false) {
```
mikeharder: Why do we need to export a function to get a version from an arbitrary file?  Why isn't this a property on Swagger, that uses its #path?

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+    if (segments && segments.length > 1) {
+      for (const s of segments) {
+        if (apiVersionRegex.test(s)) {
+          return withPreview ? s : s.substring(0, 10);
+        }
+      }
+    }
+  }
+
+  if (existsSync(filePath)) {
```
mikeharder: switch to async/await

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+}
+
+/**
+ * Get preceding swagger version for a specific lifecycle stage
+ * @param {string} targetSwaggerPath - Path to the swagger file to find predecessors for
+ * @param {Swagger[]} availableSwaggers - Array of swagger objects to search within
+ * @param {string} versionKind - Version lifecycle stage (PREVIEW or STABLE)
+ * @returns {string | undefined} - Path to preceding version or undefined if not found
+ */
+function getPrecedingSwaggerByType(targetSwaggerPath, availableSwaggers, versionKind) {
```
mikeharder: Should this be a member function of the Readme class?  Are all the inputs related to a single Readme?

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+  const targetSwagger = availableSwaggers.find((s) => s.path === targetSwaggerPath);
```
mikeharder: Should use map lookup for perf.

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+
+  return undefined;
+}
+
+/**
+ * Get preceding swagger versions for a given swagger file
+ * @param {string} targetSwaggerPath - Path to the swagger file to find predecessors for
+ * @param {Swagger[]} availableSwaggers - Array of swagger objects to search within
+ * @returns {{preview?: string, stable?: string}} - Paths to preceding versions
+ */
```
mikeharder:
```
```

copy/paste bug?

.github/shared/src/swagger.js
```diff
@@ -177,3 +285,189 @@ function json(file) {
...
+/**
+ * Get operations from all previous versions that also exist in current swagger
+ * @param {string} targetSwaggerPath - Path to the current swagger file
+ * @param {Swagger[]} availableSwaggers - Array of swagger objects to search within
+ * @returns {Promise<LegacyOperation[]>} - Operations that exist in both previous versions and current version
+ */
+export async function getLegacyVersionOperations(targetSwaggerPath, availableSwaggers) {
```
mikeharder: Also move to Readme?

mikeharder (changes requested):
see comments

.github/shared/src/swagger.js
raych1: It refers to the operations in previous versions. 
Rename to: `PreviousVersionOperations`?

.github/shared/src/swagger.js
raych1: Currently breaking change check scenario doesn't need to look up an operation. From shared code perspective, we could make it as a map with unique key to serve for more scenarios.

mikeharder: yes, even if you don't need the map, we're standardizing on `Map<string, T>` for spec model APIs

.github/shared/src/swagger.js
mikeharder: Less than the name itself, I care more about the structure.  Should we use a pair of `Operation` objects, instead of `Operation` and `string`?

.github/shared/src/swagger.js
raych1: It returns `{sub-folder}/{filename}`.
such as, input: ../chaos/resource-manager/Microsoft.Chaos/stable/2024-01-01/types/common.json, 2024-01-01, output: types/common.json.

.github/shared/src/swagger.js
raych1: It already does.

.github/shared/src/swagger.js
mikeharder: Is this always the last two path segements?  Or is it everything after the version number?

.github/shared/src/swagger.js
mikeharder: I mean, let the original error bubble.  Don't catch and replace it with a SpecModelError.

.github/shared/src/swagger.js
raych1: It's everything after version number.

.github/shared/src/swagger.js
raych1: I removed them and added get functions to compute them.

.github/shared/src/swagger.js
raych1: sure

.github/shared/src/swagger.js
raych1: Removed it.

.github/shared/src/swagger.js
raych1: Removed it.

.github/shared/src/swagger.js
raych1: It would be called by open-diff-runner.

.github/shared/src/swagger.js
```diff
@@ -177,3 +266,207 @@ function json(file) {
...
+  }
+
+  return undefined;
+}
+
+/**
+ * Get operations from all previous versions that also exist in current swagger
+ * @param {string} targetSwaggerPath - Path to the current swagger file
+ * @param {Swagger[]} availableSwaggers - Array of swagger objects to search within
+ * @returns {Promise<LegacyOperation[]>} - Operations that exist in previous versions which exist in current version too
```
mikeharder:
```suggestion
 * @returns {Promise<Map<string, LegacyOperation[]>} - Operations that exist in previous versions which exist in current version too
```

raych1: Now it returns a map with key as the swagger path and value as the matched operations from this swagger.

raych1:
@mikeharder could you review the update?

.github/shared/src/spec-model.js
```diff
@@ -217,3 +231,141 @@ function readme(file) {
...
+  return result;
+}
+
+/**
+ * Get preceding swagger versions for a given swagger file
+ * @param {string} targetSwaggerPath - Path to the swagger file to find predecessors for
+ * @param {Swagger[]} availableSwaggers - Array of swagger objects to search within
+ * @returns {Promise<{preview?: string, stable?: string}>} - Paths to preceding versions
+ */
+export async function getPrecedingSwaggers(targetSwaggerPath, availableSwaggers) {
```
mikeharder: Why does this take `availableSwaggers` as a parameter, rather than using the `getSwaggers()` property of this object?  The point of spec-model is users should just need a pointer to any object in a spec-model graph.  Users shouldn't need to keep their own list of swaggers.  The list should be used from a specModel object, like SpecModel, Readme, Tag, etc.

.github/shared/src/spec-model.js
```diff
@@ -12,6 +13,12 @@ import { Readme } from "./readme.js";
  *
  * @typedef {import('./swagger.js').Swagger} Swagger
  * @typedef {import('./tag.js').Tag} Tag
+ *
+ * @typedef {Object} Operation
```
mikeharder: Import this from swagger.js.  We should never duplicate a typedef.

.github/shared/src/spec-model.js
mikeharder: Also, I still think this method (and `getExistedVersionOperations()`) make more sense on the Swagger object, in which case they should need **no** parameters.  The `targetSwaggerPath()` is Swagger.path, and availableSwaggers can be found by walking up from Swagger to Readme or SpecModel (as high as needed).

mikeharder: If you disagree, what might help convince me, is adding some real-world examples to this PR of how you want to use the new APIs.

.github/shared/src/spec-model.js
```diff
@@ -217,3 +231,141 @@ function readme(file) {
...
+  if (!targetSwagger) {
+    return result;
+  }
+
+  const fileName = await targetSwagger.getFileName();
+  const currentVersion = await targetSwagger.getVersion();
+
+  // Get operations from current swagger file
+  const currentOperationsMap = await targetSwagger.getOperations();
+  const currentOperations = Array.from(currentOperationsMap.values());
```
mikeharder:
```suggestion
  const currentOperations = [...currentOperationsMap.values()];
```

nit: we prefer this syntax

.github/shared/src/spec-model.js
```diff
@@ -217,3 +231,141 @@ function readme(file) {
...
+        return currentOperations.some(
+          (currentOp) =>
+            currentOp.path === operation.path &&
+            currentOp.id === operation.id &&
+            currentOp.httpMethod === operation.httpMethod,
+        );
+      });
+      result.set(previousSwagger.path, matchingOperations);
+    } catch {
+      // Skip versions that can't be processed
```
mikeharder: This should probably throw rather than swallowing the exception.  We want to find versions that can't be processed, so they can be fixed.

.github/shared/src/swagger.js
```diff
@@ -177,3 +261,66 @@ function json(file) {
...
+  return basename(filePath);
+}
+
+/**
+ * Extract version string from input file path
+ * @param {string} filePath - Path to the input file
+ * @param {boolean} [withPreview=false] - Whether to include preview suffix
+ * @returns {Promise<string | undefined>} - Version string extracted from path
+ */
+export async function getVersionFromInputFile(filePath, withPreview = false) {
```
mikeharder: This function seems a bit strange.  It tries to get the version from the path (sync), but reads the file content (async) if needed?  These can be combined, but it's usually not recommended to combine sync and async computation like this.

.github/shared/src/swagger.js
```diff
@@ -118,6 +129,56 @@ export class Swagger {
     return filtered;
   }
 
+  /**
+   * @returns {Promise<Map<string, Operation>>}
+   */
+  async getOperations() {
+    if (!this.#operations) {
+      this.#operations = new Map();
+      const content = await readFile(this.#path, "utf8");
```
mikeharder, line 137: Assuming the file content is needed in multiple APIs (getOperations and getVersion), we should ensure the file is only read once.  Easiest way, when either value is needed, compute and cache both.

.github/shared/src/swagger.js
```diff
@@ -118,6 +129,56 @@ export class Swagger {
...
+      }
+    }
+    return this.#operations;
+  }
+
+  /**
+   * @returns {Promise<string>} swagger version
+   */
+  async getVersion() {
+    return (await getVersionFromInputFile(this.#path)) || "";
```
mikeharder: If we can't get a version, should this return "" or throw?

mikeharder (changes requested):
see comments

.github/shared/src/spec-model.js
raych1: Log and throw the exception.

.github/shared/src/spec-model.js
raych1: As discussed, I moved these three helper functions to breaking change code base.

.github/shared/src/swagger.js
raych1: File content will only be read in this function. I removed the content access in getVersion function.

.github/shared/src/swagger.js
raych1: return the version.

.github/shared/src/swagger.js
raych1: I updated this function to return the immediate folder name before the filename if the regex isn't matched in path. This is based on the scan result against the whole specification folder: the immediate folder name is same as the version value in the swagger content.
e.g. /specification/schemaregistry/data-plane/Microsoft.SchemaRegistry/stable/2022-10/schemaregistry.json

.github/shared/src/swagger.js
```diff
@@ -51,6 +61,7 @@ export class Swagger {
   constructor(path, options) {
     const rootDir = dirname(options?.tag?.readme?.path ?? "");
     this.#path = resolve(rootDir, path);
+    this.#operations = undefined; // Will be initialized as Map when loaded
```
mikeharder:
```suggestion
```

nit: should be unnecessary since fields are initialized to undefined by default

.github/shared/src/swagger.js
```diff
@@ -132,6 +193,23 @@ export class Swagger {
...
+  get versionKind() {
+    return dirname(this.#path).includes("/preview/")
+      ? API_VERSION_LIFECYCLE_STAGES.PREVIEW
+      : API_VERSION_LIFECYCLE_STAGES.STABLE;
+  }
+
+  /**
+   * @returns {Promise<string>} file name of the swagger file
+   */
+  async getFileName() {
```
mikeharder: Typically, the "file name" of a path is everything after the last folder.  However, below you say this includes everything after the version dir, so it may include folders like "types/common.js".  So, I think this should either be renamed, or moved out of spec-model.  Why or how would this be useful to other consumers of spec-model?

.github/shared/src/swagger.js
```diff
@@ -177,3 +261,68 @@ function json(file) {
...
+  return basename(filePath);
+}
+
+/**
+ * Extract version string from input file path
+ * @param {string} filePath - Path to the input file
+ * @param {boolean} [withPreview=false] - Whether to include preview suffix
+ * @returns {string} - Version string extracted from path, or empty string if not found
+ */
+export function getVersionFromInputFile(filePath, withPreview = false) {
```
mikeharder, line 288: Why is this exported?  Is it intended to only be an internal helper inside Swagger?

.github/shared/src/swagger.js
```diff
@@ -118,6 +129,56 @@ export class Swagger {
...
+        }
+      }
+    }
+    return this.#operations;
+  }
+
+  /**
+   * @returns {Promise<string>} swagger version
+   */
+  async getVersion() {
```
mikeharder:
```suggestion
  get Version() {
```

.github/shared/src/swagger.js
mikeharder: Inline this code in "get Version()".  You can still test it, by creating "new Swagger" from a test path, then calling "get Version()".

.github/shared/src/swagger.js
```diff
@@ -158,6 +236,12 @@ export class Swagger {
   }
 }
 
+// API version lifecycle stages
+export const API_VERSION_LIFECYCLE_STAGES = {
```
mikeharder: Should this be an enum instead of strings?

mikeharder (changes requested):
see comments

mikeharder (approved):

PR #35596, opened 2025-07-01
[TypeSpecValidation] Only "main" and "RPSaaSMaster" should validate all specs on core changes

mikeharder:
Unblocks PRs merging from main, like https://github.com/Azure/azure-rest-api-specs-pr/pull/23405

### DIFF
--- a/.github/workflows/typespec-validation.yaml
+++ b/.github/workflows/typespec-validation.yaml
@@ -27,5 +27,5 @@ jobs:
           # Only "main" and "RPSaaSMaster" should validate all specs if core files change
           $isProduction = @('main', 'RPSaaSMaster') -contains $Env:GITHUB_BASE_REF
 
-          ./eng/scripts/TypeSpec-Validation.ps1 -GitClean -Verbose -IgnoreCoreFiles:!$isProduction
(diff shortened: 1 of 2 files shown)


.github/workflows/typespec-validation.yaml
```diff
@@ -24,5 +24,8 @@ jobs:
           # step as failed. 
           $ErrorActionPreference = 'Continue'
 
-          ./eng/scripts/TypeSpec-Validation.ps1 -GitClean -Verbose
+          # Only "main" and "RPSaaSMaster" should validate all specs if core files change
+          $ignoreCoreFiles = -not (@('main', 'RPSaaSMaster') -contains $Env:GITHUB_BASE_REF)
```
weshaggard, line 28: I would double check that GITHUB_BASE_REF doesn't contain `refs/heads` which sometimes shows its head.

weshaggard (approved):
This will help with TSV All but likely will still not help for other.

.github/workflows/typespec-validation.yaml
mikeharder: BASE_REF should be safe, you are thinking of REF:

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

PR #35605, opened 2025-07-02
[EngSys] Add "if" property to suppressions

mikeharder:
- Fixes #35153

Test showing single-spec TSV failing: https://github.com/mikeharder/azure-rest-api-specs/pull/22

This PR adds an "if" property to suppressions, which is a string executed as JavaScript, in a context when predefined variables are set.
 
For example, in TSV, you can use the boolean checkingAllSpecs to include a suppression in TSV-All, but exclude the suppression when only validating a single spec in a PR, to encourage the spec owner to fix it.

image

But you can also do crazier stuff like this:

image

ToDo
- [x] Create PR to this branch, changing a spec with a scoped suppression, and ensure it fails.
- [ ] (Future PR) Move suppressions to separate files in each service directory


eng/scripts/TypeSpec-Validation.ps1
```diff
@@ -18,7 +18,7 @@ if ($TotalShards -gt 0 -and $Shard -ge $TotalShards) {
 . $PSScriptRoot/Suppressions-Functions.ps1
 . $PSScriptRoot/Array-Functions.ps1
 
-$typespecFolders, $checkedAll = &"$PSScriptRoot/Get-TypeSpec-Folders.ps1" `
+$typespecFolders, $checkingAllSpecs = &"$PSScriptRoot/Get-TypeSpec-Folders.ps1" `
```
scbedd, line 21: Just a heads up that you're probably already aware of, but @benbp and I ran into some nasty issues with destructuring from a function call. I'd assume this would apply to another script as well.

If the first destructured result is $null, you can see that `$typespecFolders` will get the value of `$checkingAllSpecs`, and `$checkingAllSpecs` will get $null.

If you're never going to get null from either then it's totally fine.

eng/tools/suppressions/src/suppressions.ts
```diff
@@ -140,19 +152,28 @@ export function getSuppressionsFromYaml(
...
+
+        return s.paths.some((suppressionPath) => {
+          const pattern: string = join(dirname(suppressionsFile), suppressionPath)
+            .split(sep)
+            .join(posixSep);
+          return minimatch(pathPosix, pattern);
+        });
+      })
+      // If
+      .filter((s) => s.if === undefined || vm.runInNewContext(s.if, sandbox))
```
weshaggard, line 175: Do we need to worry about arborary code being executed here? Do we need to consider limiting this to a simple set of hard-coded expression types?

eng/tools/suppressions/test/suppressions.test.ts
```diff
@@ -268,10 +268,63 @@ test("suppression with rules", () => {
...
+  path: "**"
+  if: foo || bar
+  reason: if-foo-or-bar
+- tool: TestTool
+  path: "**"
+  if: foo && bar
+  reason: if-foo-and-bar
+- tool: TestTool
+  path: "**"
+  if: require("process").version.startsWith("v")
```
weshaggard, line 317: Can we add some negative test cases to see what happens if there are syntax errors or if they try to read the environment for something more nefarious?

eng/tools/suppressions/src/suppressions.ts
scbedd: The `vm.runInNewContext` provides that isolation as far as I understand it.

scbedd: Yeah, that's exactly what the intent is for this line. We _could_ invoke the js directly, but that WOULD be a bad idea. You provide this new context + sandbox, you get:

- Cannot access the host application's variables, functions, or modules
- Cannot modify the host application's state
- Only has access to what you explicitly provide in the sandbox object
- Cannot access Node.js globals like process, global, __dirname, etc. (unless you put them in the sandbox)
- Cannot require arbitrary modules (unless you provide require in the sandbox)

So it seems reasonable to me.

scbedd (approved):
Can't argue with the unit tests.

eng/tools/suppressions/src/suppressions.ts
mikeharder: `node:vm` is not a security mechanism, so we shouldn't use it as such.  However, we should already be using other security mechanisms (like the `pull_request` target) to validate untrusted code.  We start with the assumption that doing almost anything with user-submitted content could lead to arbitrary code execution.

So, I don't think this feature changes our security in any way.  We already assumed running `getSuppressions()` on untrusted PRs was dangerous.

eng/tools/suppressions/test/suppressions.test.ts
mikeharder: Any errors (syntax or otherwise) in the JS causes an exception that bubbles up.  I can add a test for this.

I don't want to add any security-related tests, since this isn't designed to be a security boundary.

eng/scripts/TypeSpec-Validation.ps1
mikeharder: Thanks for the tip, but we should never return null, and this code has been working as-is for a while.

eng/tools/suppressions/src/suppressions.ts
weshaggard: Sounds good I just want to be sure we have thought about it.

PR #35627, opened 2025-07-02
GET API Design: add results contract

kseager:
# Choose a PR Template

### DIFF
--- a/specification/ai/Azure.AI.Projects/evaluations/models.tsp
+++ b/specification/ai/Azure.AI.Projects/evaluations/models.tsp
@@ -13,25 +13,18 @@ using TypeSpec.Versioning;
 namespace Azure.AI.Projects;
 
 
-@doc("Labeled results per row.")
+@doc("Labeled results across evaluators per data row.")
 @added(Versions.v2025_05_15_preview)
 @removed(Versions.v1)
 model LabeledRows {
-  @doc("The label associated with the evaluation.")
-  label: string;
-
-  @doc("Detailed reasoning or explanation for the evaluation.")
-  reasoning: string;
-
-  @doc("The version identifier for the evaluation model.")
-  version: string;
+  // The arbitrary JSON data for evaluator metrics.
+  ...Record<string>;
 }
 
 @doc("Summary across rows for evaluation.")
 @added(Versions.v2025_05_15_preview)
 @removed(Versions.v1)
-model MetricSummary extends Results {
-  type: "metricSummary";
+model MetricSummary {
 
   @doc("Aggregated summary per evaluator across all rows.")
     data: Record<{
@@ -47,7 +40,7 @@ model MetricSummary extends Results {
 @doc("Stored result details and storage location. Currently only blob storage support for evaluation result/result uploading.")
 @added(Versions.v2025_05_15_preview)
 @removed(Versions.v1)
-model Storage extends Results {
+model BlobStorageResults extends Results {
   type: "storage";
 
   @doc("AssetId reference for the stored evaluationResult.")
@@ -55,9 +48,6 @@ model Storage extends Results {
 
   @doc("The connection ID to the storage account where the evaluation results are stored.")
   storageConnectionId: string;
-
-  @doc("Data table name for the data.")
-  dataTableName: string;
 }
 
 #suppress "@azure-tools/typespec-azure-core/no-string-discriminator"
@@ -179,8 +169,11 @@ model Evaluation {
   @doc("Evaluators to be used for the evaluation.")
   evaluators: Record<EvaluatorConfiguration>;
 
-  @doc("Output Results.")
-  outputs: Array<Results>;
+  @doc("Evaluation stored output results.")
+  outputs?: Array<Results>;
+
+  @doc("Evaluation results metric Summary.")
+  outputSummary?: MetricSummary;
 }
 
 @doc("The redaction configuration will allow the user to control what is redacted.")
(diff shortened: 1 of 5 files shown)


specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -26,6 +26,43 @@ interface Evaluations {
   @doc("List evaluation runs")
   list is EvaluationsOperations.ResourceList<Evaluation>;
 
+  @doc("List evaluation run row results for a specific evaluatorId")
+  @route("evaluations/{evaluatorId}/results ")
+  @added(Versions.v2025_05_15_preview)
+  @removed(Versions.v1)
+  @get
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  listRowResultsByEvaluatorId(
```
singankit: Should this be evaluationId or evaluatorId ?

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
```diff
@@ -26,6 +26,43 @@ interface Evaluations {
...
+    @doc("""
+    The index of the first item to return. Used for pagination.
+    """)
+    @query startIndex?: int32,
+
+    @doc("""
+    The number of items to return in the result set. Used for pagination.
+    """)
+    @query pageSize?: int32,
```
singankit: Should we have a default value for it ?

specification/ai/data-plane/Azure.AI.Projects/stable/2025-05-01/azure-ai-projects.json
YusakuNo1, line 1: For the review, let's don't include the 3 generated JSON files to avoid noice, we will focus on the real changes during API review, and then add the JSON files back before merging -- also we need to tell the other teams to always update the JSON, otherwise, this PR merges a lot changes from the other team's JSON

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
YusakuNo1: +1  we should set a default value

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
YusakuNo1: +1 it should be `evaluationId`

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -12,6 +12,64 @@ using TypeSpec.Versioning;
...
+    defect_rate?: string;
+    passing_rate?: string;
+    binary_aggregate: string;
+  }>;
+}
+
+@doc("Stored result details and storage location. Currently only blob storage support for evaluation result/result uploading.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v1)
+model Storage extends Results {
```
YusakuNo1: Maybe we should call it "BlobStorageResults" instead of just "Storage"? When we are adding more storage types, this name could be confusing

specification/ai/Azure.AI.Projects/evaluations/routes.tsp
singankit: If it is evaluatorId should it be scoped to an evaluation too ?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -12,6 +12,64 @@ using TypeSpec.Versioning;
 
 namespace Azure.AI.Projects;
 
+
+@doc("Labeled results per row.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v1)
+model LabeledRows {
```
YusakuNo1, line 19: This is the format of annotation eval, while the cloud eval row has much more information, I think we should converge a format which can hold more information?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -12,6 +12,64 @@ using TypeSpec.Versioning;
...
+  type: "storage";
+
+  @doc("AssetId reference for the stored evaluationResult.")
+  evaluationResultAssetId: string;
+
+  @doc("The connection ID to the storage account where the evaluation results are stored.")
+  storageConnectionId: string;
+
+  @doc("Data table name for the data.")
+  dataTableName: string;
```
YusakuNo1: For blob storage, it seems like `evaluationResultAssetId` will point to a file, what is `dataTableName` for Blob Storage?

specification/ai/Azure.AI.Projects/evaluations/models.tsp
```diff
@@ -12,6 +12,64 @@ using TypeSpec.Versioning;
...
+  reasoning: string;
+
+  @doc("The version identifier for the evaluation model.")
+  version: string;
+}
+
+@doc("Summary across rows for evaluation.")
+@added(Versions.v2025_05_15_preview)
+@removed(Versions.v1)
+model MetricSummary extends Results {
```
YusakuNo1: Where do we use `MetricSummary`?

kseager: with this design I'm saying it will be one of the result types returned in the array, but actually thinking about it again I think we agreed if we always commit to always returning it, then we should have it be a dedicated parameter

YusakuNo1: Sounds good! I asked because `outputSummary?: MetricSummary;` was not in the TypeSpec. I found you added it now, it's good enough

PR #35650, opened 2025-07-03
Fail the run if status doesn't exist

raych1:


.github/workflows/src/verify-run-status.js
```diff
@@ -120,6 +120,8 @@ export async function verifyRunStatusImpl({
         commitStatusContext = commitStatuses[0].context;
         commitStatusState = commitStatuses[0].state;
         commitStatusTargetUrl = commitStatuses[0].target_url;
+      } else {
+        core.setFailed(`Failed to fetch commit status which doesn't exist.`);
```
mikeharder: I think both this case, and the catch case below, should just print a notice (like for "pending") rather than an error.

raych1: print notice and return with no-op as what does for 'pending' case.

mikeharder (approved):

PR #35700, opened 2025-07-07
Add testIds filter for ListTestRuns

mbhardwaj-msft:
Add new query param `testIds` to `ListTestRunQueryParams` model in `LoadTestService` to filter test runs by multiple test IDs.
This is for the upcoming `2025-03-01-public-preview` release

### DIFF
--- a/specification/loadtestservice/LoadTestService/models.tsp
+++ b/specification/loadtestservice/LoadTestService/models.tsp
@@ -1347,7 +1347,7 @@ model ListTestRunQueryParams {
   @added(APIVersions.v2024_07_01_preview)
   createdByTypes?: string[];
 
-  @doc("Comma separated set of test IDs.")
+  @doc("Comma-separated list of test IDs. If you are using testIds, do not send a value for testId.")
   @query(#{ explode: false })
   @added(APIVersions.v2025_03_01_preview)
   testIds?: string[];
--- a/specification/loadtestservice/LoadTestService/models.tsp
+++ b/specification/loadtestservice/LoadTestService/models.tsp
@@ -1346,6 +1346,11 @@ model ListTestRunQueryParams {
   @query(#{ explode: false })
   @added(APIVersions.v2024_07_01_preview)
   createdByTypes?: string[];
+
+  @doc("Comma separated set of test IDs.")
+  @query(#{ explode: false })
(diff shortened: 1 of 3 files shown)


specification/loadtestservice/LoadTestService/models.tsp
```diff
@@ -1346,6 +1346,11 @@ model ListTestRunQueryParams {
   @query(#{ explode: false })
   @added(APIVersions.v2024_07_01_preview)
   createdByTypes?: string[];
+
+  @doc("Comma separated set of test IDs.")
```
mitsha-microsoft: Should we put a doc/summary saying should not send `testId` if this is sent ?

specification/loadtestservice/data-plane/Microsoft.LoadTestService/preview/2025-03-01-preview/examples/ListTestRuns_TestIds.json
```diff
@@ -0,0 +1,306 @@
+{
+  "operationId": "LoadTestRun_ListTestRuns",
+  "title": "List test runs.",
```
mitsha-microsoft: I think you will need to update the title otherwise it will fail in some CI

mitsha-microsoft (approved):

specification/loadtestservice/LoadTestService/models.tsp
mbhardwaj-msft: @Nagarjuna-Vipparthi , could you please help with the doc/description that we should put here?

Nagarjuna-Vipparthi: How about this?

Comma-separated list of test IDs. If you are using testIds, do not send a value for testId.

mbhardwaj-msft: updated

specification/loadtestservice/data-plane/Microsoft.LoadTestService/preview/2025-03-01-preview/examples/ListTestRuns_TestIds.json
mbhardwaj-msft: updated

Harshan01 (approved):

PR #35723, opened 2025-07-07
[Monitor Query] Add TypeSpec for Logs Query

pvaneck:
This migrates Azure Monitor Logs Query data plane APIs to TypeSpec.


specification/monitor/Azure.Monitor.Query.Logs/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+    endpoint: url,
+  }
+)
+namespace MonitorQueryLogs;
+
+/** Service API versions */
+enum Versions {
+  /** The 2022-10-27 API version. */
+  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
+  v2022_10_27: "2022-10-27",
```
gracewilcox: Think technically, the version is the path parameter[ `v1`](https://github.com/Azure/azure-rest-api-specs/blob/1bb4dc06a988ac7f4fa952bae55e92e96b54e02e/specification/operationalinsights/data-plane/Microsoft.OperationalInsights/stable/2022-10-27/OperationalInsights.json#L19). The version you have listed here is not actually used, the path parameter version is.

specification/monitor/Azure.Monitor.Query.Logs/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+@useAuth(
+  OAuth2Auth<[
+    {
+      type: OAuth2FlowType.implicit,
+      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
+      scopes: ["https://api.loganalytics.io/.default"],
+    }
+  ]>
+)
+@service(#{ title: "Monitor Query Logs" })
```
gracewilcox: Is changing the service name a breaking change? Know the service name [the swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/operationalinsights/data-plane/Microsoft.OperationalInsights/stable/2022-10-27/OperationalInsights.json) has been using is `Operational Insights`.

specification/monitor/Azure.Monitor.Query.Logs/tspconfig.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/data-plane"
+
+options:
+  "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "data-plane"
+    emitter-output-dir: "{project-root}/.."
+    output-file: "../operationalinsights/{azure-resource-provider-folder}/Microsoft.OperationalInsights/{version-status}/{version}/OperationalInsights.json"
```
srnagar: Do we need this `operationinsights` directory?
```suggestion
    output-file: "../{azure-resource-provider-folder}/Microsoft.OperationalInsights/{version-status}/{version}/OperationalInsights.json"
```

specification/monitor/Azure.Monitor.Query.Logs/models.tsp
```diff
@@ -0,0 +1,674 @@
...
+/** Contains the columns and rows for one table in a query response. */
+model Table {
+  /** The name of the table. */
+  name: string;
+
+  /** The list of columns in this table. */
+  columns: Column[];
+
+  /** The resulting rows from this query. */
+  rows: Record<unknown>[][];
```
srnagar: For Java, we need to use `unknown[][]` to generate the same type as we did with Swagger. However, we can do this through `@alternateType` customization if this works for all other languages.

gracewilcox: Go will also use need to use `unknown` to generate the same type. Can use `@alternateType` decorator.

pvaneck: Actually, this should be `unknown[][]` as mentioned. The list elements are not JSON. Updated.

lmazuel (approved):
Breaking Changes are SDK impact, and we confirmed that it doesn't impact SDK since there is a convenience layer and we're omving to TypeSpec generation

lmazuel (approved):

PR #35804, opened 2025-07-09
Support cross version breaking change validation

raych1:
### New GitHub Actions Workflow:
*  Added a new workflow to analyze breaking changes across API versions. 

### Cross-Version Support:
* Added related validation function and helper functions
* Integrated SpecModel shared code

### Test PR
https://github.com/raych1/azure-rest-api-specs/pull/89
https://github.com/raych1/azure-rest-api-specs/pull/91

@mikeharder could you review this PR?

### DIFF
--- a/.github/workflows/breaking-change-code.yaml
+++ b/.github/workflows/breaking-change-code.yaml
@@ -28,13 +28,13 @@ jobs:
         id: swagger-breaking-change-analyze-code
         run: |
           npm exec --no -- openapi-diff-runner \
-            --spec-repo-path "${{ github.workspace }}" \
+            --spec-repo-path "$GITHUB_WORKSPACE" \
             --pr-number "${{ github.event.pull_request.number }}" \
             --head-commit "${{ github.event.pull_request.head.sha }}" \
             --pr-source-branch "${{ github.event.pull_request.head.ref }}" \
             --pr-target-branch "${{ github.event.pull_request.base.ref }}" \
             --source-repo "${{ github.event.pull_request.head.repo.full_name }}" \
-            --target-repo "${{ github.repository }}"
+            --target-repo "$GITHUB_REPOSITORY"
 
       # Upload artifact for 'BreakingChangeReviewRequired' label
       - if: |
--- a/.github/workflows/breaking-change-code.yaml
+++ b/.github/workflows/breaking-change-code.yaml
@@ -28,13 +28,13 @@ jobs:
         id: swagger-breaking-change-analyze-code
         run: |
           npm exec --no -- openapi-diff-runner \
-            --srp "${{ github.workspace }}" \
-            --number "${{ github.event.pull_request.number }}" \
-            --sb "${{ github.event.pull_request.head.ref }}" \
-            --tb "${{ github.event.pull_request.base.ref }}" \
-            --hc "${{ github.event.pull_request.head.sha }}" \
-            --sr "${{ github.event.pull_request.head.repo.full_name }}" \
-            --repo "${{ github.repository }}"
+            -p "${{ github.workspace }}" \
+            -n "${{ github.event.pull_request.number }}" \
+            -c "${{ github.event.pull_request.head.sha }}" \
+            --pr-source-branch "${{ github.event.pull_request.head.ref }}" \
+            --pr-target-branch "${{ github.event.pull_request.base.ref }}" \
+            --source-repo "${{ github.event.pull_request.head.repo.full_name }}" \
+            --target-repo "${{ github.repository }}"
 
       # Upload artifact for 'BreakingChangeReviewRequired' label
       - if: |
--- a/.github/workflows/breaking-change-code.yaml
+++ b/.github/workflows/breaking-change-code.yaml
@@ -28,7 +28,7 @@ jobs:
         id: swagger-breaking-change-analyze-code
         run: |
           npm exec --no -- openapi-diff-runner \
-            --spec-repo-path "$GITHUB_WORKSPACE" \            
+            --spec-repo-path "$GITHUB_WORKSPACE" \
             --head-commit "$GITHUB_SHA" \
             --pr-source-branch "$GITHUB_HEAD_REF" \
             --pr-target-branch "$GITHUB_BASE_REF" \
--- a/.github/workflows/breaking-change-code.yaml
+++ b/.github/workflows/breaking-change-code.yaml
@@ -28,11 +28,11 @@ jobs:
         id: swagger-breaking-change-analyze-code
         run: |
           npm exec --no -- openapi-diff-runner \
-            --spec-repo-path "$GITHUB_WORKSPACE" \
+            --spec-repo-path "$GITHUB_WORKSPACE" \            
+            --head-commit "$GITHUB_SHA" \
+            --pr-source-branch "$GITHUB_HEAD_REF" \
+            --pr-target-branch "$GITHUB_BASE_REF" \
             --pr-number "${{ github.event.pull_request.number }}" \
-            --head-commit "${{ github.event.pull_request.head.sha }}" \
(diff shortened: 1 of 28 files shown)


.github/workflows/breaking-change-code.yaml
```diff
@@ -28,12 +28,13 @@ jobs:
...
           npm exec --no -- openapi-diff-runner \
-            --srp "${{ github.workspace }}" \
-            --number "${{ github.event.pull_request.number }}" \
-            --sb "${{ github.event.pull_request.head.ref }}" \
-            --tb "${{ github.event.pull_request.base.ref }}" \
-            --hc "${{ github.event.pull_request.head.sha }}" \
-            --repo "${{ github.repository }}"
+            -p "${{ github.workspace }}" \
+            -n "${{ github.event.pull_request.number }}" \
+            -c "${{ github.event.pull_request.head.sha }}" \
```
mikeharder: please use long switch names in scripts

.github/workflows/breaking-change-code.yaml
```diff
@@ -28,12 +28,13 @@ jobs:
         id: swagger-breaking-change-analyze-code
         run: |
           npm exec --no -- openapi-diff-runner \
-            --srp "${{ github.workspace }}" \
-            --number "${{ github.event.pull_request.number }}" \
-            --sb "${{ github.event.pull_request.head.ref }}" \
-            --tb "${{ github.event.pull_request.base.ref }}" \
-            --hc "${{ github.event.pull_request.head.sha }}" \
-            --repo "${{ github.repository }}"
+            -p "${{ github.workspace }}" \
```
mikeharder:
```suggestion
            -p "$GITHUB_WORKSPACE" \
```

We might have this same issue in other workflows, but I think env vars are the safest at preventing script injections.  All params should use env vars, either built-in or custom-defined.

tracking for all workflows: https://github.com/Azure/azure-rest-api-specs-pr/issues/23566

.github/workflows/breaking-change-cross-version-code.yaml
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  validateBreakingChange:
+    name: "[TEST-IGNORE] Breaking Change(Cross-Version) - Analyze Code"
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
mikeharder, line 17: Are you sure you need `fetch-depth: 0`?  It's OK if you do, but faster if you can use something lower, maybe `2` if the default `1` isn't enough?

.github/workflows/breaking-change-cross-version-code.yaml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Setup .NET 6 SDK
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: "6.0.x"
+
+      - name: Breaking Change(Cross-Version) - Analyze Code
+        id: breaking-change-cross-version-analyze-code
+        run: |
+          npm exec --no -- openapi-diff-runner \
+            -t "CrossVersion" \
```
mikeharder: same comments as above

eng/tools/openapi-diff-runner/src/log.ts
```diff
@@ -130,3 +130,37 @@ export async function logGroup<T>(title: string, content: () => Promise<T> | T):
     logMessage("", LogLevel.EndGroup);
   }
 }
+
+/**
+ * Maximum safe log line length for GitHub Actions (64KB - some buffer)
+ */
+const MAX_LOG_LINE_LENGTH = 60 * 1024; // 60KB to leave some buffer
```
mikeharder, line 137: what if a line exceeds the limit?

mikeharder (approved):

.github/workflows/breaking-change-code.yaml
raych1: In this specific case, it doesn't secure more by using `GITHUB_WORKSPACE` since github context is also controlled by the GitHub action and it doesn't have direct user input.

.github/workflows/breaking-change-cross-version-code.yaml
raych1: Yes, it cannot use shallow clone. It needs to work with multiple branches such as source branch, target branch, base branch. It needs to have remote references. Especially when RPSaaSMaster branch and main branch involves significant divergence.

eng/tools/openapi-diff-runner/src/log.ts
raych1: Truncate the latter content.

mikeharder: Is the default behavior good enough, we don't need to special-case this?

.github/workflows/breaking-change-code.yaml
raych1: Updated.

.github/workflows/breaking-change-code.yaml
mikeharder: If you use env vars, the bash script executed is exactly what you see in yaml.  If you use `${{ }}`, your bash script will be modified by the workflow pre-processor, which I believe adds risk (admittedly maybe slight).

Why not just use env vars?

eng/tools/openapi-diff-runner/src/log.ts
raych1: No, the default behavior is bad. It stopped log rendering and everything after this line just never showed up. Refer to [this run](https://github.com/raych1/azure-rest-api-specs/actions/runs/16181268822/job/45678241811)

.github/workflows/breaking-change-code.yaml
raych1: I'm fine for the switch.

eng/tools/openapi-diff-runner/src/run-oad.ts
```diff
@@ -69,7 +69,7 @@ export async function runOad(
     oadCompareOutput = await oad.compare(oldSpec, newSpec, { consoleLogLevel: "warn" });
   }
 
-  logMessage(`oadCompareOutput: ${oadCompareOutput}`);
+  logMessageSafe(`oadCompareOutput: ${oadCompareOutput}`);
```
mikeharder, line 72:
```suggestion
  logMessage(`oadCompareOutput: ${JSON.serialize(JSON.parse(oadCompareOutput), null, 2)}`);
```

This should solve line length problem, and make log more readable, without needing to truncate.

.github/workflows/breaking-change-code.yaml
```diff
@@ -28,12 +28,13 @@ jobs:
...
-            --sb "${{ github.event.pull_request.head.ref }}" \
-            --tb "${{ github.event.pull_request.base.ref }}" \
-            --hc "${{ github.event.pull_request.head.sha }}" \
-            --repo "${{ github.repository }}"
+            --spec-repo-path "$GITHUB_WORKSPACE" \
+            --pr-number "${{ github.event.pull_request.number }}" \
+            --head-commit "${{ github.event.pull_request.head.sha }}" \
+            --pr-source-branch "${{ github.event.pull_request.head.ref }}" \
+            --pr-target-branch "${{ github.event.pull_request.base.ref }}" \
+            --source-repo "${{ github.event.pull_request.head.repo.full_name }}" \
```
mikeharder: Replace all with env vars

raych1: Not all of them have predefined github variables we can directly use, such as PR number and source branch. I will update them in next PR if needed.

raych1: Update: we must use set env var then pass them into the script for any user controllable input. For these non-user-controlled variables, we are ok to reference them directly in yml.

PR #35831, opened 2025-07-10
[Monitor Metrics] Add TypeSpec files

pvaneck:
TypeSpec conversion for Monitor Batch Metrics.


lmazuel (approved):
Breaking Changes are SDK impact, and we confirmed that it doesn't impact SDK since there is a convenience layer and we're omving to TypeSpec generation

PR #35835, opened 2025-07-10
azsdk mcp default version to latest

jeo02:


.vscode/mcp.json
```diff
@@ -3,7 +3,7 @@
       "azure-sdk-mcp": {
         "type": "stdio",
         "command": "pwsh",        
-        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run", "-Version", "1.0.0-dev.20250707.1"]
+        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
```
weshaggard, line 6: Does this work? @benbp seems to think that it fails if you don't specify a Version parameter currently.

jeo02: It is working, talked to Ben and it was because of the tag azsdk_version I had created.

The way the script works is it looks for the most recent release with [package name]_* in this case it was finding azsdk_version and was taking "version" as the version number.

benbp (approved):

PR #35853, opened 2025-07-11
[WebPubSub] TSP Migration for WebPubSub

kashifkhan:


specification/webpubsub/Azure.Messaging.WebPubSub/main.tsp
```diff
@@ -0,0 +1,56 @@
...
+@useAuth(
+  OAuth2Auth<[
+    {
+      type: OAuth2FlowType.implicit,
+      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
+      scopes: ["https://webpubsub.azure.com/.default"],
+    }
+  ]>
+)
+namespace Azure.Messaging.WebPubSubService;
```
catalinaperalta: Do we need the namespace to be nested? I think this will affect how some languages generate the clients. Also, we dont need to put Azure in the namespace either. A better name might be: 
```suggestion
namespace WebPubSub;
```

specification/webpubsub/Azure.Messaging.WebPubSub/main.tsp
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * The available API versions.
+ */
+enum Versions {
+  /**
+   * The 2024-12-01 API version.
+   */
+  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
+  v2024_12_01: "2024-12-01",
```
catalinaperalta, line 53: Is there a newer api version?

specification/webpubsub/Azure.Messaging.WebPubSub/models.tsp
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Paged response for group members
+ */
+model GroupMemberPagedValues {
+  /** The group members */
+  value?: GroupMember[];
+
+  /** The link to the next page of results */
+  nextLink?: string;
+}
```
catalinaperalta: Looks like you probably need to indicate that this is a paged result. https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/decorators/#@Azure.Core.pagedResult

catalinaperalta: If you do use paged result, then the items and nextlink decorators should help too

specification/webpubsub/Azure.Messaging.WebPubSub/models.tsp
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Error detail information.
+ */
+model InnerError {
```
catalinaperalta: It doesnt look like this is being used anywhere?

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
+import "@azure-tools/typespec-azure-core";
+import "@typespec/rest";
+import "./models.tsp";
+import "@typespec/openapi";
```
catalinaperalta: We dont recommend using this library generally, why did you need it?

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
...
+  @post
+  closeAllConnections is Azure.Core.Foundations.Operation<
+    {
+      /**
+       * Target hub name, which should start with alphabetic characters and only contain
+       * alpha-numeric characters or underscore.
+       */
+      @pattern("^[A-Za-z][A-Za-z0-9_`,.[\\]]{0,127}$")
+      @path
+      hub: string;
```
catalinaperalta, line 58: Nit: since it seems like this path param is reused you can define it in an alias and spread it into each operation. This will ensure that you only need to modify it in one place and make it easier to maintain. This applies to any other relevant params that are reused

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
...
+
+      /**
+       * Exclude these connectionIds when closing the connections in the hub.
+       */
+      @query("excluded")
```
catalinaperalta: Nit: you dont need to specify the name since it matches the tsp param name already
```suggestion
      @query
```

catalinaperalta: Same feedback applies to other instances of this

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
...
+
+  /**
+   * List connections in a group.
+   */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API - must use current operation signature for backward compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Preserving existing API operation name for backward compatibility"
+  @summary("List connections in a group.")
+  @route("/api/hubs/{hub}/groups/{group}/connections")
+  @get
+  listConnections is Azure.Core.Foundations.Operation<
```
catalinaperalta: Looks like this should be marked as a list operation

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
...
+
+      /**
+       * A token that allows the client to retrieve the next page of results. This
+       * parameter is provided by the service in the response of a previous request when
+       * there are additional results to be fetched. Clients should include the
+       * continuationToken in the next request to receive the subsequent page of data.
+       * If this parameter is omitted, the server will return the first page of results.
+       */
+      @query("continuationToken")
+      continuationToken?: string;
```
catalinaperalta: Looks like this needs the continuationtoken decorator

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,976 @@
...
+
+      /**
+       * The meaning of the target depends on the specific permission. For
+       * joinLeaveGroup and sendToGroup, targetName is a required parameter standing for
+       * the group name.
+       */
+      @query("targetName")
+      targetName?: string;
+    },
+    void
```
catalinaperalta: If this is a head operation why arent there headers defined here?

catalinaperalta: same comment applies to other head ops

specification/webpubsub/Azure.Messaging.WebPubSub/tspconfig.yaml
```diff
@@ -0,0 +1,36 @@
+parameters:
+   "service-dir":
+      default: "sdk/webpubsub"
+emit:
+  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-python"
```
catalinaperalta: We only need the tsp autorest emitter here
```suggestion
  - "@azure-tools/typespec-autorest"
```

specification/webpubsub/Azure.Messaging.WebPubSub/tspconfig.yaml
```diff
@@ -0,0 +1,36 @@
...
+    package-version: "1.3.0"
+    flavor: "azure"
+    generate-test: true
+    generate-sample: true
+  # Uncomment this line and add "@azure-tools/typespec-python" to your package.json to generate Python code
+  # "@azure-tools/typespec-python":
+  #   "basic-setup-py": true
+  #   "package-version":
+  #   "package-name":
+  #   "output-path":
```
catalinaperalta: Nit: remove

specification/webpubsub/Azure.Messaging.WebPubSub/tspconfig.yaml
```diff
@@ -0,0 +1,36 @@
...
+  # Uncomment this line and add "@azure-tools/typespec-java" to your package.json to generate Java code
+  # "@azure-tools/typespec-java": true
+  # Uncomment this line and add "@azure-tools/typespec-csharp" to your package.json to generate C# code
+  # "@azure-tools/typespec-csharp": true
+  # Uncomment this line and add "@azure-tools/typespec-ts" to your package.json to generate Typescript code
+  # "@azure-tools/typespec-ts": true
```
catalinaperalta: We should add the relevant info for the other langs

specification/webpubsub/Azure.Messaging.WebPubSub/client.tsp
```diff
@@ -0,0 +1,95 @@
...
+@@clientLocation(Azure.Messaging.WebPubSubService.WebPubSubService.addConnectionsToGroups,
+  Azure.Messaging.WebPubSubService
+);
+@@clientLocation(Azure.Messaging.WebPubSubService.WebPubSubService.closeAllConnections,
+  Azure.Messaging.WebPubSubService
+);
+@@clientLocation(Azure.Messaging.WebPubSubService.WebPubSubService.getClientAccessToken,
+  Azure.Messaging.WebPubSubService
+);
+@@clientLocation(Azure.Messaging.WebPubSubService.WebPubSubService.removeConnectionsFromGroups,
```
catalinaperalta: Not highlighting all of the instances, but is this due to the interface use?

specification/webpubsub/Azure.Messaging.WebPubSub/main.tsp
kashifkhan: changed it to `WebPubSubService`

specification/webpubsub/Azure.Messaging.WebPubSub/main.tsp
kashifkhan: from the stable versions, that was latest https://github.com/Azure/azure-rest-api-specs/tree/main/specification/webpubsub/data-plane/WebPubSub/stable/2024-12-01

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
kashifkhan: probably auto complete error :)

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
kashifkhan: not sure tbh. Might have to check their spec and see why its like that

specification/webpubsub/Azure.Messaging.WebPubSub/client.tsp
kashifkhan: I think so as well as all the operations are available at the root leve.

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -5,24 +5,27 @@ import "./models.tsp";
...
   /**
    * Add filtered connections to multiple groups.
    */
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API - must use current operation signature for backward compatibility"
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Existing Operation"
   @summary("Add filtered connections to multiple groups.")
   @route("/api/hubs/{hub}/:addToGroups")
   @post
-  op addConnectionsToGroups is Azure.Core.Foundations.Operation<
+  op WebPubSub_AddConnectionsToGroups is Azure.Core.Foundations.Operation<
```
catalinaperalta: Why are the operations being named this way?

kashifkhan: discussed with @catalinaperalta offline - will rename these and pair up with example using id to clear model validation

lmazuel (approved):
The Python SDK has been rewired on top of this TypeSpec, with confirmed non breaking behavior. CI issues are not important here, and the next PR will move to V2 folder structure, so the file name gap doesn't really matter.

specification/webpubsub/Azure.Messaging.WebPubSub/client.tsp
```diff
@@ -0,0 +1,19 @@
...
+
+using Azure.ClientGenerator.Core;
+
+namespace Customizations;
+
+/** Client options for Azure Web PubSub. */
+model WebPubSubClientOptions {
+  /** Target hub name, which should start with alphabetic characters and only contain
+     alpha-numeric characters or underscore. */
+  hub: string;
```
catalinaperalta, line 12: when you generated the library did you see this param successfully added on the client and in the path?

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
```diff
@@ -0,0 +1,1013 @@
...
+       */
+      @query
+      clientProtocol?: WebPubSubClientType = WebPubSubClientType.Default;
+    },
+    ClientTokenResponse & {
+      @header("content-type")
+      contentType: ResponseContentType;
+
+      @statusCode
+      statusCode: 200;
```
catalinaperalta: Nit: responses are 200 by default you dont need this

specification/webpubsub/Azure.Messaging.WebPubSub/tspconfig.yaml
```diff
@@ -0,0 +1,49 @@
...
+    package-dir: "Azure.Security.Attestation"
+    package-version: "1.6.0"
+    namespace: "{package-dir}"
+    clear-output-folder: true
+    model-namespace: false
+    generate-test: true
+    generate-sample: true
+    flavor: "azure"
+  "@azure-tools/typespec-ts":
+    package-name: "@azure/web-pubsub"
```
catalinaperalta: I believe ts has a nested object for package details which includes where to place package-name: https://github.com/Azure/azure-rest-api-specs/blob/7c640375cebac863ef3430e8ac9aa8cae4843677/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml#L32-L36

specification/webpubsub/Azure.Messaging.WebPubSub/routes.tsp
catalinaperalta: Same feedback applies to other operations where this is the only status code.

catalinaperalta (approved):
looks good overall, just left some questions but not blocking

specification/webpubsub/Azure.Messaging.WebPubSub/client.tsp
kashifkhan: yes it generated fine, I just had to patch the order over https://github.com/Azure/azure-sdk-for-python/pull/42131/files#diff-f9bec055c057f8adc6392c7e48a122b2e1f9a3ea6430b9e1ff7b669e1d1cf8deR137

lmazuel (approved):

PR #35872, opened 2025-07-14
Release Redisenterprise microsoft.cache 2025 07 01

nikitagarg123:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [X] New API version for an existing resource provider. (If API spec is not defined in TypeSpec, the PR should have been created in adherence to [OpenAPI specs PR creation guidance](https://aka.ms/azsdkdocs/createopenapispec)).
  - [ ] Other, please clarify:

## Due diligence checklist

- [X] I confirm this PR is modifying Azure Resource Manager (ARM) related specifications, and not data plane related specifications.
- [X] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


dipeshbhakat-microsoft (approved):

mikeharder:
I removed label `ARMSignedOff`, because after this label was added, the PR added substantive changes that caused LintDiff to start failing.  So I think this PR should be re-reviewed by ARM.

nikhgup (approved):

TimLovellSmith:
>           "enum": [

This doesn't even make sense does it?

Clusters created with old api versions should still be able to have public network access enabled/disabled can't they? So that people don't need to recreate the cluster to use it the new way?

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2118 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>           "x-nullable": true

My take is this should not really be nullable.

People need to know whether public network access is enabled or disabled. Why can't we tell them that, instead of returning null? And why would we want to accept explicit 'null' in a create request? (as opposed to leaving it undefined)

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2126 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>           "x-nullable": true

(I'm okay with this being null in the update requests, but not in the response)

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2154 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>           "$ref": "#/definitions/ClusterCreateProperties",

Just a thought - leaving it named ClusterProperties might be better for backcompat than renaming it to ClusterCreateProperties?

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:1932 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>           "$ref": "#/definitions/ClusterCreateProperties",

Also we return it from operations like GET which aren't create operations.

---
In reply to: [3202232049](https://github.com/Azure/azure-rest-api-specs/pull/35872#issuecomment-3202232049) [](http://example.com/codeflow?ancestors=3202232049)

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:1932 in 13a4ad1. [](commit_id = 13a4ad194172e429501ddee8c77e0975a75e1b5e, deletion_comment = False)

TimLovellSmith:
>           }

to `x-ms-enum` we should add `values`, and document that 1.0 and 1.1 are no longer supported

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2004 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>         "publicNetworkAccess": {

what is the reason this isn't in common properties?

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2143 in bc8813d. [](commit_id = bc8813d77d2df403348ddbc463f61c40370966d4, deletion_comment = False)

TimLovellSmith:
>         "publicNetworkAccess": {

can we avoid defining the public network access enum twice, and just $ref it from both places? Or do we need separate enums?

---
In reply to: [3202239688](https://github.com/Azure/azure-rest-api-specs/pull/35872#issuecomment-3202239688) [](http://example.com/codeflow?ancestors=3202239688)

---
Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2143 in 13a4ad1. [](commit_id = 13a4ad194172e429501ddee8c77e0975a75e1b5e, deletion_comment = False)

specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/examples/RedisEnterpriseUpdate.json
```diff
@@ -0,0 +1,61 @@
...
+        "tags": {
+          "tag1": "value1"
+        },
+        "properties": {
+          "hostName": "cache1.westus.something.azure.com",
+          "provisioningState": "Updating",
+          "resourceState": "Updating",
+          "redisVersion": "5",
+          "minimumTlsVersion": "1.2",
+          "publicNetworkAccess": "Enabled"
```
TimLovellSmith, line 50: >   "publicNetworkAccess": "Enabled" [](http://example.com/codeflow?start=7&length=35)

can we include an example where we set public network access to null, if that is valid? (I don't think it is though)

nikitagarg123:
> > ```
> >       "enum": [
> > ```
> 
> This doesn't even make sense does it?
> 
> Clusters created with old api versions should still be able to have public network access enabled/disabled can't they? So that people don't need to recreate the cluster to use it the new way?
> 
> Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2118 in [bc8813d](https://github.com/Azure/azure-rest-api-specs/commit/bc8813d77d2df403348ddbc463f61c40370966d4). [](commit_id = [bc8813d](https://github.com/Azure/azure-rest-api-specs/commit/bc8813d77d2df403348ddbc463f61c40370966d4), deletion_comment = False)

@TimLovellSmith yes clusters created with old api versions can set public network access enabled/disabled
This statement means customers need to use **enabled** or **disabled** while setting the public network access and **null** is not allowed 
we are adding x_nullable only for get response so that old caches can return this value and we can nudge customers to set this
in portal

nikitagarg123:
> > ```
> >     "publicNetworkAccess": {
> > ```
> 
> what is the reason this isn't in common properties?
> 
> Refers to: specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/redisenterprise.json:2143 in [bc8813d](https://github.com/Azure/azure-rest-api-specs/commit/bc8813d77d2df403348ddbc463f61c40370966d4). [](commit_id = [bc8813d](https://github.com/Azure/azure-rest-api-specs/commit/bc8813d77d2df403348ddbc463f61c40370966d4), deletion_comment = False)

@TimLovellSmith since we are making this as required property, we can't have required in PATCH call
so I separated out Put and patch and added required only for put call

specification/redisenterprise/resource-manager/Microsoft.Cache/stable/2025-07-01/examples/RedisEnterpriseUpdate.json
nikitagarg123: yes, we won't allow null in PUT, I had to define because response and request should be same, and in GET response we return null for old caches

TimLovellSmith:
> > This doesn't even make sense does it?
> > Clusters created with old api versions should still be able to have public network access enabled/disabled can't they? So that people don't need to recreate the cluster to use it the new way?

> @TimLovellSmith yes clusters created with old api versions can set public network access enabled/disabled This statement means customers need to use **enabled** or **disabled** while setting the public network access and **null** is not allowed we are adding x_nullable only for get response so that old caches can return this value and we can nudge customers to set this in portal

I basically don't think its a good idea to return 'null' for that. I'd suggest instead:
-either just return enabled/disabled to reflect the ACTUAL CURRENT enablement state
or
-introduce a 3rd enum value to return, instead of using null, with a more self-explanatory name than null

TimLovellSmith (changes requested):
I have a concern that using nullable for publicNetworkAccess is inappropriate #resolved

TimLovellSmith:
> I have a concern that using nullable for publicNetworkAccess is inappropriate

I had a long conversation with the feature team. There are a lot of special extenuating circumstances particular to this API to approve.

TimLovellSmith (approved):

PR #35889, opened 2025-07-15
[Conv runtime] modify clientname for python 20250515preview

amber-Chen-86:
# Choose a PR Template


amber-Chen-86:
The breaking changes in this PR are intentional. I changed the Endpoint type from string to url to avoid a Python SDK issue where request URLs are generated incorrectly. In the generated OpenAPI/Swagger it still appears as type: string with format: uri, so there is no wire-level change—only the SDK surface becomes stricter. 

In addition, for the 2025-05-15-preview version I removed the LUIS-related parameter types per guidance from the Python SDK board, since LUIS is being deprecated and we want to steer customers away from those inputs. Thank you.

kristapratico (approved):

MattGal (approved):

PR #35946, opened 2025-07-16
Azure.AI.Projects: Rename some SDK classes and variables, to support stable release of the SDKs

dargilco:


specification/ai/Azure.AI.Projects/client.tsp
```diff
@@ -22,13 +22,22 @@ using Azure.ClientGenerator.Core;
 @@clientName(Azure.AI.Projects.Indexes.deleteVersion, "delete");
 @@clientName(Azure.AI.Projects.Indexes.createOrUpdateVersion, "createOrUpdate");
 
+// Less generic names
+@@clientName(Azure.AI.Projects.Sku, "ModelDeploymentSku");
+@@clientName(Azure.AI.Projects.SasCredential, "BlobReferenceSasCredential"); // Not to be confused with class "SASCredentials"
```
KrzysztofCwalina, line 27: I think it would be better to call it AIProjectSaSCredential

KrzysztofCwalina (commented):
Are these all the renames that we need?

specification/ai/Azure.AI.Projects/client.tsp
dargilco: This is the credential required to access the blob storage (it's the type of the property "credentials" on the TypeSpec model "BlobReference"), which is a connected service to your Foundry Project. It's not credential for accessing the Foundry project. So I think renaming it "AIProjectSasCredential" is misleading.

specification/ai/Azure.AI.Projects/datasets/routes.tsp
```diff
@@ -30,6 +31,7 @@ interface Datasets
     {
       @doc("The pending upload request parameters")
       @Http.bodyRoot
+      @clientName("pendingUploadOptions")
```
trangevi: Is this necessary alongside the change in client.tsp? If it is, do we have it for all the others? If it's not, can we choose which place and remain consistent?

dargilco: We spoke about this over phone call. Resolving as-is. Thanks!

specification/ai/Azure.AI.Projects/datasets/routes.tsp
```diff
@@ -30,6 +31,7 @@ interface Datasets
     {
       @doc("The pending upload request parameters")
       @Http.bodyRoot
+      @clientName("configuration")
```
johanste: This should go in client.tsp unless there is some obscure reason it can't?

dargilco: @johanste as far as I know that's not possible to do (declare this rename in client.tsp), unless you define a template where the client name is a template argument. I think that will obfuscate the TypeSpec too much, just for the sake of overwriting client name. @iscai-msft @lmazuel what are your thoughts?

johanste: Ah, ok. Had to go back and look at why that wouldn't work. I'll go back and yell at the typespec team about metaproperties and the ability to augment expressions. 

However, I wouldn't want "configuration" to be the name for python. That is a C# naming convention. So we should scope it there.

specification/ai/Azure.AI.Projects/client.tsp
```diff
@@ -22,13 +22,29 @@ using Azure.ClientGenerator.Core;
 @@clientName(Azure.AI.Projects.Indexes.deleteVersion, "delete");
 @@clientName(Azure.AI.Projects.Indexes.createOrUpdateVersion, "createOrUpdate");
 
+// Less generic names
+@@clientName(Azure.AI.Projects.Sku, "ModelDeploymentSku");
```
johanste, line 26: Not for Python. C# and Java only?

specification/ai/Azure.AI.Projects/client.tsp
```diff
@@ -22,13 +22,29 @@ using Azure.ClientGenerator.Core;
...
+@@clientName(Azure.AI.Projects.SasCredential, "BlobReferenceSasCredential"); // Not to be confused with class "SASCredentials"
+
+// Do not use terms "request" or "response" in class names as these are REST API terms
+@@clientName(Azure.AI.Projects.AssetCredentialResponse, "DatasetCredential");
+@@clientName(Azure.AI.Projects.PendingUploadRequest,
```
johanste: Not for Python. Configuration is not nomenclature we use for this and prefer request. I suspect the same is true for TS/JS.

specification/ai/Azure.AI.Projects/client.tsp
dargilco: This "Sku" only applies to model deployment. I think it needs a more specific name, regardless of which programming language we use. This class is defined in the "Azure.AI.Projects" namespace, not something more specific... so I thin it needs to be scoped by having a better name.

specification/ai/Azure.AI.Projects/datasets/routes.tsp
dargilco: Scope to C# only.

specification/ai/Azure.AI.Projects/client.tsp
dargilco: Scoped to C# only (to go together with the input variable name "configuration")

specification/ai/Azure.AI.Projects/client.tsp
dargilco: This "Sku" only applies to model deployment. I think it needs a more specific name, regardless of which programming language we use. This class is defined in the "Azure.AI.Projects" namespace, not something more specific... so I think it needs to be scoped by having a better name.

specification/ai/Azure.AI.Projects/datasets/routes.tsp
dargilco: Scoped to C# only

specification/ai/Azure.AI.Projects/client.tsp
dargilco: Per offline 1:1 chat with Johan, he is okay naming it `ModelDeploymentSku`

johanste (approved):

specification/ai/Azure.AI.Projects/client.tsp
dargilco: @KrzysztofCwalina I'm resolving this comment in order to complete this PR to unblock Python release. We can address comments related to C# release in another TypeSpec PR targeting C# changes, if needed.

trangevi (approved):

PR #35951, opened 2025-07-16
Add `summarize-impact`  check, handle label additions in `summarize-checks`

scbedd:


eng/tools/summarize-impact/src/runner.ts
```diff
@@ -0,0 +1,903 @@
...
+   * The labels are read from ADO pipeline environment variables.
+   * These variables are set to appropriate value by BreakingChangesRuleManager.addBreakingChangeLabels().
+   * Read that function's comment for details.
+   */
+  function getBreakingChangesLabelsFromOad(): string[] {
+    const oadLabelsEnvVars = [
+      breakingChangeLabelVarName.toUpperCase(),
+      crossVersionBreakingChangeLabelVarName.toUpperCase(),
+    ];
+    let breakingChangeLabelsFromOad: string[] = _.uniq(
```
scbedd: Examining downstream variables set by the `BreakingChange` check?

eng/tools/summarize-impact/test/cli.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import path from "path";
+
+import { getChangedFilesStatuses } from "@azure-tools/specs-shared/changed-files";
+import { PRContext, LabelContext } from "../src/types.js";
+import { evaluateImpact } from "../src/runner.js";
+
+// const REPOROOT = path.resolve(__dirname, "..", "..", "..", "..");
+
+describe("Check Changes", () => {
```
scbedd, line 10: These will get snipped out in favor of more isolated integration tests that rely on fixtures instead of local file paths.

.github/workflows/src/context.js
```diff
@@ -215,6 +221,14 @@ export async function extractInputs(github, context, core) {
           }
         }
       }
+      if (!issue_number) {
```
mikeharder: as discussed, you should be able to remove this, if we update all the workflows you depend on, to publish the "issue_number" artifact

.github/workflows/src/issues.js
mikeharder, line 1: I think this can be reverted as well?

.github/shared/test/changed-files.test.js
```diff
@@ -92,6 +94,11 @@ describe("changedFiles", () => {
     expect(files.filter(specification)).toEqual(expected);
   });
 
+  it("filter:typespec", () => {
+    const expected = ["specification/contosowidgetmanager/Contoso.Management/main.tsp"];
```
mikeharder: shouldn't `expected` include `tspconfig.yaml`?

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -23,6 +72,220 @@ import {
...
+ * For additional context, see:
+ *
+ * - "Deep-dive into breaking changes on spec PRs"
+ *   https://aka.ms/azsdk/pr-brch-deep
+ *
+ * - "[Breaking Change][PR Workflow] Use more granular labels for Breaking Changes approvals"
+ *   https://github.com/Azure/azure-sdk-tools/issues/6374
+ */
+/**
+ * @typedef {"SameVersion" | "CrossVersion"} BreakingChangesCheckType
```
mikeharder, line 141: You may want to convert these to runtime enums, so code can use the enum values instead of hardcoding the strings.

Definition: https://github.com/Azure/azure-rest-api-specs/blob/7f90dbd64a7593ba2f68fb44e874ae2584798818/.github/workflows/src/label.js#L3-L11

Usage: https://github.com/Azure/azure-rest-api-specs/blob/7f90dbd64a7593ba2f68fb44e874ae2584798818/.github/workflows/src/arm-auto-signoff.js#L38-L44

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -23,6 +72,220 @@ import {
...
+ * Represents a GitHub label.
+ * Currently used in the context of processing
+ * labels related to the review workflow of PRs submitted to
+ * Azure/azure-rest-api-specs and Azure/azure-rest-api-specs-pr repositories.
+ * This processing happens in the prSummary.ts / summary() function.
+ *
+ * See also: https://aka.ms/SpecPRReviewARMInvariants
+ */
+// todo: inject `core` for access to logging
+export class Label {
```
mikeharder, line 194: Our convention is to:

1. Prefix private fields with `#` (makes them private at runtime)
2. Declare fields (with types) above ctor

Example:

https://github.com/Azure/azure-rest-api-specs/blob/a12ced1212df32affe4dd203cf7d99dd71acc985/.github/shared/src/spec-model.js#L20-L29

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -23,6 +72,220 @@ import {
...
+   * @param {Set<string>} labelsToRemove
+   */
+  applyStateChange(labelsToAdd, labelsToRemove) {
+    if (this.shouldBePresent === undefined) {
+      console.warn(
+        "ASSERTION VIOLATION! " +
+          `Cannot applyStateChange for label '${this.name}' ` +
+          "as its desired presence hasn't been defined. Returning early."
+      );
+      return;
```
mikeharder: We should probably throw instead of returning.  It's easier to find bugs if workflows crash on error.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -23,6 +72,220 @@ import {
...
+          + `At this point of execution this should not happen.`
+      );
+    }
+  }
+
+  /**
+   * @param {string} label
+   * @returns {boolean}
+   */
+  isEqualToOrPrefixOf(label) {
```
mikeharder, line 281: I think this is exported as a top-level function as well (see bottom of file)

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -158,6 +393,366 @@ export const breakingChangesCheckType = {
   },
 };
 
+// #endregion constants
+// #region Required Labels
+
+
+/**
+ * @param {import("./labelling.js").LabelContext} context
```
mikeharder: Shouldn't need to import own file

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -158,6 +393,366 @@ export const breakingChangesCheckType = {
...
+  }
+}
+
+/**
+ *
+ * @param {any[]} arr
+ * @param {any[]} values
+ * @returns
+ */
+function containsAll(arr, values) {
```
mikeharder, line 452: Rename to `includesAll`, add generic type params, and move to `.github/shared/array.js`.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -158,6 +393,366 @@ export const breakingChangesCheckType = {
...
+  return values.every((value) => arr.includes(value));
+}
+
+/**
+ *
+ * @param {any[]} arr
+ * @param {any[]} values
+ * @returns
+ */
+function containsNone(arr, values) {
```
mikeharder, line 462: Rename to `includesNone`, add generic type params, and move to `.github/shared/array.js`.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -158,6 +393,366 @@ export const breakingChangesCheckType = {
...
+    + `armReviewLabel.shouldBePresent: ${armReviewLabel.shouldBePresent}.`)
+
+  return { armReviewLabelShouldBePresent: armReviewLabel.shouldBePresent }
+}
+
+/**
+ * @param {string} branchName
+ * @returns {boolean}
+ */
+function isReleaseBranch(branchName) {
```
mikeharder, line 574: If used by other code, should probably move into specModel.  Fine in here for now.

.github/workflows/summarize-impact.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            eng/
+            .github/
+
+      - name: Checkout 'after' state
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
mikeharder, line 18: Why do you need fetch-depth:2 when you are already cloning before to another folder?

eng/tools/summarize-impact/src/impact.ts
```diff
@@ -0,0 +1,917 @@
...
+      breakingChangeReviewRequiredLabel.shouldBePresent,
+  };
+
+  /**
+   * Get labels denoting required breaking change or versioning review.
+   * The labels are read from ADO pipeline environment variables.
+   * These variables are set to appropriate value by BreakingChangesRuleManager.addBreakingChangeLabels().
+   * Read that function's comment for details.
+   */
+  function getBreakingChangesLabelsFromOad(): string[] {
```
scbedd, line 343: Right here.

.github/workflows/summarize-impact.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Checkout 'before' state
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: before
+
+      - name: Setup Node and install deps
+        uses: ./.github/actions/setup-node-install-deps
+
+      # todo: why is this necessary? postinstall should already take care of the build?
```
mikeharder: Add your new project to `eng/tools/tsconfig.json`

.github/workflows/summarize-impact.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Build summarize-impact tool
+        run: |
+          cd eng/tools/summarize-impact
+          npm run build
+
+      - name: Get PR labels
+        id: get-labels
+        run: |
+          labels=$(gh pr view ${{ github.event.pull_request.number }} --json labels --jq '[.labels[].name] | join(",")')
```
mikeharder, line 36: I would prefer to call this API from JS instead of bash, but it might be more difficult since summarize-impact is called from the shell instead of github-script.  We could replace just this one step with github-script, but the value is questionable if it stays this simple.  If we move all of summarize-impact to JS, we should move this as well.

.github/workflows/test/summarize-checks.test.js
```diff
@@ -511,4 +512,105 @@ describe("summarizeChecksImpl", () => {
...
+            eventName,
+            "main",
+            existingLabels,
+            changedLabel
+        );
+
+        expect(labelsToAdd.sort()).toEqual(expectedLabelsToAdd.sort());
+        expect(labelsToRemove.sort()).toEqual(expectedLabelsToRemove.sort());
+      },
+      600000,
```
mikeharder: Why do we need to set the timeout?  Shouldn't these be fast?

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -0,0 +1,138 @@
...
+    const gitRoot = await simpleGit(inputPath).revparse("--show-toplevel");
+    return resolve(gitRoot.trim());
+  } catch (error) {
+    console.error(
+      `Error: Unable to determine the root folder of the git repository.`,
+      `Please ensure you are running this command within a git repository OR providing a targeted directory that is within a git repo.`,
+    );
+    process.exit(1);
+  }
+}
```
mikeharder: This should probably move to .github/shared.  Maybe create file `.github/shared/simple-git.js` we can use for this and similar helpers?

eng/tools/summarize-impact/src/impact.ts
```diff
@@ -0,0 +1,917 @@
...
+    rpaasExceptionRequired: rpaasExceptionLabelShouldBePresent,
+    typeSpecChanged: typeSpecLabelShouldBePresent,
+    isNewApiVersion: newApiVersion,
+    isDraft: context.isDraft,
+    labelContext: labelContext,
+    targetBranch: context.targetBranch
+  };
+}
+
+export function isManagementPR(filePaths: string[]): boolean {
```
mikeharder, line 171: This helper (and several following it) should move to spec-model (or changed-files).  If you don't want to do it now, open an issue for later.

eng/tools/summarize-impact/src/types.ts
mikeharder, line 1: This should probably be split up, maybe one type per file (especially the larger types).

package-lock.json
```diff
@@ -420,6 +488,36 @@
...
+    "eng/tools/summarize-impact": {
+      "name": "@azure-tools/summarize-impact",
+      "dev": true,
+      "dependencies": {
+        "@azure-tools/openapi-tools-common": "^1.2.2",
+        "@azure-tools/specs-shared": "file:../../../.github/shared",
+        "@azure/openapi-markdown": "0.9.4",
+        "@ts-common/commonmark-to-markdown": "^2.0.2",
+        "commonmark": "^0.29.0",
+        "glob": "^11.0.3",
```
mikeharder, line 500: Maybe in an future PR, but could this be switch to `globby@^14`, which we already use in TSV?

mikeharder (approved):

.github/workflows/summarize-impact.yaml
scbedd: ~~Oh is github actions/checkout actually smart? I'd expect that they would be two different `git clone` invocations so any optimization (including limiting fetch-depth) is virtuous.~~

I see your point now. I'll eliminate the sparse checkout above.

.github/workflows/test/summarize-checks.test.js
scbedd: I will clean this up. This was present due to debugging interacting with vitest timeout.

.github/shared/test/changed-files.test.js
scbedd: You know, it absolutely should. Why is this test passing as-is. Will fix before merge.

.github/workflows/src/context.js
scbedd: Executing on this now.

eng/tools/summarize-impact/src/types.ts
scbedd: Yeah will do.

eng/tools/summarize-impact/test/cli.test.ts
scbedd: They have been updated to skip properly in CI context. More will be populated as I exercise the capabilities of this package.

.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -35,6 +41,11 @@ import {
   typeSpecRequirementDataPlaneTsg,
 } from "./tsgs.js";
 
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import { execSync } from 'child_process';
```
mikeharder: Use this helper instead.  It uses the most secure exec API:

https://github.com/Azure/azure-rest-api-specs/blob/2ae7cdd6509fb91ed9d7742186e278b67b86398e/.github/shared/src/exec.js#L37-L46

.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -35,6 +41,11 @@ import {
   typeSpecRequirementDataPlaneTsg,
 } from "./tsgs.js";
 
+import fs from 'fs';
```
mikeharder:
```suggestion
import fs from 'fs/promises';
```

Use async APIs from fs/promises

eng/tools/summarize-impact/src/impact.ts
scbedd: This entire section will be eliminated, as post-breakingChange run we trigger `Update-Labels`.

.github/workflows/summarize-impact.yaml
scbedd: I will file an issue and follow-up on this item. For stability's sake I'm getting everything working with this set here.

eng/tools/summarize-impact/src/types.ts
scbedd: Done 👍

package-lock.json
scbedd: Will file an issue and follow-up on this one.

.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -825,3 +937,64 @@ function buildViolatedLabelRulesNextStepsText(violatedRequiredLabelsRules) {
...
+      owner,
+      repo,
+      artifact_id: jobSummaryArtifact.id,
+      archive_format: "zip",
+    });
+
+    core.info(`Successfully downloaded job-summary artifact ID: ${jobSummaryArtifact.id}`);
+
+    // Write zip buffer to temp file and extract JSON
+    const tmpZip = path.join(os.tmpdir(), `job-summary-${runId}.zip`);
```
mikeharder:
```suggestion
    const tmpZip = path.join(process.env.RUNNER_TEMP || os.tmpdir(), `job-summary-${runId}.zip`);
```

Technically more correct to use RUNNER_TEMP in a workflow.  Can move this to a helper method in `shared/path.js`.

scbedd:
Issues I will address immediately following merge of this PR:

- [ ] Swap to `globby` to align with repo
  - Frankly we can probably use SpecModel directly and entirely eliminate this usage of `glob`. Following up on this now.
- [ ] Retrieve labels directly instead of using `gh` in a static call `summarize-impact.yaml`
- [x] Move some `isManagement` or `isReleaseBranch` to `SpecModel` or common helpers.
   - [x] Utilize `changed-files` utilities instead of redetecting in `IsDataPlanePR` and `isManagementPR`
   - [x] Moved `isReleaseBranch` to common code. #36120
- [x] Walk all existing workflows and add a publish of `issue-number` if it could trigger `summarize-checks` I have added it to `summarize-pr-impact` and it's working as we expect. -> #36112
- [x] Move some array-specific functions to shared/array.js -> #36116
- [x] Clean up class definition per [this comment](https://github.com/Azure/azure-rest-api-specs/pull/35951#discussion_r2220370544) #36120

eng/tools/summarize-impact/src/cli.ts
scbedd: Done.

eng/tools/summarize-impact/src/impact.ts
scbedd: I'll file an issue for moving these common pieces to specModel. 👍

.github/shared/test/simple-git.test.js
```diff
@@ -0,0 +1,20 @@
...
+import os from "os";
+import { mkdtemp, rm } from "fs/promises";
+import { getRootFolder } from "../src/simple-git.js";
+import { describe, expect, it } from "vitest";
+
+describe("getRootFolder", () => {
+  it("resolves to repo root from a nested folder", async () => {
+    const testDir = __dirname;
+    const calculatedRoot = await getRootFolder(testDir);
+    const expectedRoot = path.resolve(path.join(__dirname, "..", "..", ".."));
```
mikeharder, line 11:
```suggestion
    const expectedRoot = path.resolve(path.join(testDir, "..", "..", ".."));
```

scbedd: SHEEPISH.

.github/workflows/src/summarize-checks/labelling.js
scbedd: Just FYI I have synced with Ray, and we'll remove the need for this. BreakingChange is handled solely by the new check + output + `Update-Labels` action.

.github/workflows/src/summarize-checks/labelling.js
scbedd: Done.

PR #35983, opened 2025-07-17
Support new RP case in cross version validation

raych1:


eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
```diff
@@ -387,18 +394,25 @@ export function getReadmeFolder(swaggerFile: string) {
...
 
   if (!folder) {
     throw new Error(`Could not determine readme folder for swagger path: ${swaggerPath}`);
   }
 
   const fullFolderPath = path.join(specRepoFolder, folder);
-  logMessage(`getSpecModel: folder: ${fullFolderPath}, swaggerPath: ${swaggerPath}`);
 
+  // Return if the readme folder is not found which means it's a new RP
+  if (!existsSync(fullFolderPath)) {
```
mikeharder, line 409: I think this check, and perhaps also the SpecModel cache, should move into SpecModel itself.  I'm working on this now.

mikeharder (approved):
You should be able to replace your own SpecModel cache with the built-in one, in either this PR or a later one.  With the built-in cache, you can just use `new SpecModel(path)`, and it will return a cached instance if it exists.  PR adding built-in cache: https://github.com/Azure/azure-rest-api-specs/pull/36008

PR #36003, opened 2025-07-17
[spec-model] Ctors should throw if paths cannot be read

mikeharder:


mikeharder:
Getting too complex.  SpecModel already depends too heavily on "only throw when accessing a path", for example when creating a Readme from string content.

We would need to basically add a flag like "ensureFileAccess", that can be set to true or false depending what the caller wants.

But at this point, I think it's easier for the call to just check if the path exists themselves.

PR #36008, opened 2025-07-17
[spec-model] Add cache in ctor, to re-use objects for the same path

mikeharder:


.github/shared/test/swagger.test.js
```diff
@@ -51,6 +51,26 @@ describe("Swagger", () => {
     );
   });
 
+  it("returns examples", async () => {
```
mikeharder, line 54: Unrelated test to improve codecov

scbedd (approved):

raych1 (approved):

PR #36009, opened 2025-07-17
Rhurey/vl specs

rhurey:
Opening a PR to get a pipeline run to (hopefully) get an API view we can look at...


specification/ai/data-plane/VoiceLive/operations.tsp
```diff
@@ -0,0 +1,15 @@
...
+
+using TypeSpec.Versioning;
+
+namespace VoiceLive;
+
+// Service versioning is enum based
+#suppress "@azure-tools/typespec-azure-core/no-enum"
+@doc("API versioning.")
+enum Versions {
+  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
```
timotheeguerin: Remove this line

specification/ai/data-plane/VoiceLive/operations.tsp
```diff
@@ -0,0 +1,15 @@
+import "./models.tsp";
+import "@azure-tools/typespec-azure-core";
+
+using TypeSpec.Versioning;
+
+namespace VoiceLive;
+
+// Service versioning is enum based
+#suppress "@azure-tools/typespec-azure-core/no-enum"
```
timotheeguerin: is this the service even using versioning, and please add message to why every single of the suppression is needed

specification/ai/data-plane/VoiceLive/operations.tsp
```diff
@@ -0,0 +1,15 @@
...
+@doc("API versioning.")
+enum Versions {
+  @useDependency(Azure.Core.Versions.v1_0_Preview_2)
+  @doc("Version 2025-10-01.")
```
timotheeguerin: use doc comments `/** */` instead of `@doc` whenever possible for readability

specification/ai/data-plane/VoiceLive/operations.tsp
rhurey: The service Uri is /voice-agent/realtime?&api-version=2025-10-01&model=gpt-4o

timotheeguerin: the linter rule shouldn't flag the version enum if you use it with `@versioned`

rhurey: Use it where?

The service definition in websocket.tsp doesn't have an operation, and I'm not sure what we'd add for one. And it does have @versioned used.

timotheeguerin: On your namespace. Versioning doesn’t need to be tied to operations

timotheeguerin: I see it is there just in a different file, recommend you keep the version enum with all the service level decorators you have in websocket.tsp.

Pattern for azure specs is usually to have all of that in the `main.tsp`

specification/ai/data-plane/VoiceLive/content_parts.tsp
```diff
@@ -0,0 +1,52 @@
...
+model RequestTextContentPart extends ContentPart {
+  type: ContentPartType.input_text;
+  text?: string;
+}
+
+/** An audio content part for a request. */
+@usage(RequestUsage)
+model RequestAudioContentPart extends ContentPart {
+  type: ContentPartType.input_audio;
+  transcript?: string;
```
johanste, line 42: When would you not provide a transcript (i.e. only provide the type)?

specification/ai/data-plane/VoiceLive/models.tsp
```diff
@@ -0,0 +1,1521 @@
...
+  type: ServerEventType.session_avatar_connecting;
+
+  /** The server's SDP answer for the avatar connection. */
+  server_sdp: string;
+}
+
+// Tool customization: establish base for enriched request/response split models
+/** VoiceLive session object configuration. */
+@usage(ResponseUsage)
+model SessionBase {}
```
johanste, line 533: What is this model for? It is very unusual to have an empty model like this...

specification/ai/data-plane/VoiceLive/tspconfig.yaml
```diff
@@ -0,0 +1,56 @@
...
+emit:
+  - "@azure-tools/typespec-autorest"
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/data-plane"
+options:
+  "@azure-tools/typespec-autorest":
+    emitter-output-dir: "{project-root}"
+    omit-unreachable-types: false
+    output-file: "{version-status}/{version}/GeneratedSystemEvents.json"
```
johanste: You are not emitting system EG event schemas here,  right? So we should update the target file name.

specification/ai/data-plane/VoiceLive/custom.tsp
```diff
@@ -0,0 +1,898 @@
+import "./events.tsp";
+import "./items.tsp";
+import "./tools.tsp";
+import "@typespec/http";
```
johanste: Not used, remove.

specification/ai/data-plane/VoiceLive/custom.tsp
```diff
@@ -0,0 +1,898 @@
...
+import "@typespec/openapi";
```
johanste: Not used, remove.

specification/ai/data-plane/VoiceLive/events.tsp
```diff
@@ -0,0 +1,111 @@
+using TypeSpec.OpenAPI;
```
johanste: Not used? If not, remove.

specification/ai/data-plane/VoiceLive/content_parts.tsp
xitzhang: Added description from service teams: "Optional transcript of the audio content. This is not sent to the model, but will be attached to the message item for reference.".

specification/ai/data-plane/VoiceLive/events.tsp
xitzhang: removed

specification/ai/data-plane/VoiceLive/custom.tsp
xitzhang: Items.tsp and tools.tsp are in use, ref error would be thrown without them.
Removed other 2

specification/ai/data-plane/VoiceLive/custom.tsp
xitzhang: Removed @typespec/http, other 3 are in use

specification/ai/data-plane/VoiceLive/tspconfig.yaml
xitzhang: Renamed to VoiceLive.json

specification/ai/data-plane/VoiceLive/models.tsp
xitzhang: @johanste , ohh I believe it wasn’t empty originally, but we forgot to remove it after all of its fields were deleted. 
I’m wondering whether we can remove it now? given that it’s already part of the GA package and publicly exposed.

johanste: That would primarily be a question of breaking changes in client libraries. If they are not affected/there are no breaking changes, I would remove the model from the TypeSpec.

johanste (approved):

PR #36014, opened 2025-07-18
[DNM] Validate typespec-azure#2974

AlitzelMendez:
# Choose a PR Template


specification/apicenter/ApiCenter.Management/Api.tsp
```diff
@@ -36,5 +36,6 @@ interface Apis {
   @doc("Deletes specified API.")
   delete is ArmResourceDeleteSync<Api>;
   @doc("Returns a collection of APIs.")
+  @list
```
pshao25, line 39: I assume we don't need this since we should have this `@list` since we are using the standard template.

timotheeguerin: I think so but the template also lets you define a custom response which in theory could be not pageable so need to double check that is not being misued.

timotheeguerin:
restarted in a new pr https://github.com/Azure/azure-rest-api-specs/pull/36153

PR #36040, opened 2025-07-18
[Monitor Query Logs] adding go support

gracewilcox:
Adding go support to logs

### DIFF
--- a/specification/monitor/Monitor.Query.Logs/routes.tsp
+++ b/specification/monitor/Monitor.Query.Logs/routes.tsp
@@ -1,6 +1,8 @@
 import "@azure-tools/typespec-azure-core";
 import "./models.tsp";
+import "@azure-tools/typespec-client-generator-core";
 
+using Azure.ClientGenerator.Core;
 using TypeSpec.Http;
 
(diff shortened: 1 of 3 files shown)


specification/monitor/Monitor.Query.Logs/routes.tsp
```diff
@@ -56,6 +58,7 @@ interface Query {
 
     /** Optional. The prefer header to set server timeout, query statistics and visualization information. */
     @header("Prefer")
+    @clientName("options", "go")
```
pvaneck, line 61: I believe, in this case, the `@@clientName` decorator can be used in client.tsp.
```
@@clientName(MonitorQueryLogs.Query.executeWithResourceId::parameters.prefer,
  "options",
  "go"
);
@@clientName(MonitorQueryLogs.Query.execute::parameters.prefer,
  "options",
  "go"
);
```

srnagar: The clientname decorator is used here because of this issue - https://github.com/Azure/typespec-azure/issues/2986

srnagar (approved):

PR #36051, opened 2025-07-19
[changed-files] return absolute platform paths

mikeharder:
First attempt.  I don't like adding the `options` bag to the filters, it also makes them always require a lambda to use.

To make the filters most reusable, I thinke we need to decompose them, to their minimal parts, so callers can decide if they need to combine checks with `startsWithSpecification(repoRoot)`.

It would also be better, if we could filter the call to `getChangedFiles()` to the spec folder initially, when it makes sense.  This way, maybe the filters never need to check for "starts with spec" vs "spec folder nested under test dir".


timotheeguerin:
This PR has been closed because it has had no activity for more than a year. If this change is still needed, please reopen the PR.

PR #36095, opened 2025-07-22
[spec-model] Remove duplicate copy of readme()

mikeharder:
- Copy in changed-files.js was identical

### DIFF
--- a/.github/shared/src/spec-model.js
+++ b/.github/shared/src/spec-model.js
@@ -3,6 +3,7 @@
 import { readdir } from "fs/promises";
 import { resolve } from "path";
 import { flatMapAsync, mapAsync } from "./array.js";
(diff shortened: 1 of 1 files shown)


.github/shared/src/spec-model.js
```diff
@@ -226,14 +227,3 @@ export class SpecModel {
     return `SpecModel(${this.#folder}, {logger: ${this.#logger}}})`;
   }
 }
-
-// TODO: Remove duplication with changed-files.js (which currently requires paths relative to repo root)
-
-/**
```
mikeharder, line 232: Copy in changed-files was identical:

https://github.com/Azure/azure-rest-api-specs/blob/2ae7cdd6509fb91ed9d7742186e278b67b86398e/.github/shared/src/changed-files.js#L146-L153

PR #36096, opened 2025-07-22
[EngSys] decouple filters in changed-files.js

mikeharder:
- filters should handle platform paths, via resolve()/normalize()/sep
- filter `specification()` is removed
  - Instead, pass `paths: ["specification"]` to `getChangedFiles()`
- preparation for `getChangedFiles()` to return absolute platform paths

e2e test for impacted checks: https://github.com/mikeharder/azure-rest-api-specs/pull/26


.github/shared/src/swagger.js
```diff
@@ -4,7 +4,7 @@ import $RefParser, { ResolverError } from "@apidevtools/json-schema-ref-parser";
 import { readFile } from "fs/promises";
 import { dirname, relative, resolve } from "path";
 import { mapAsync } from "./array.js";
-import { includesFolder } from "./path.js";
+import { example } from "./changed-files.js";
```
mikeharder, line 7: example() no longer requires specification(), so the code is identical

.github/workflows/src/arm-incremental-typespec.js
```diff
@@ -30,7 +31,7 @@ export default async function incrementalTypeSpec({ core }) {
   const changedFiles = await getChangedFiles(options);
 
   // Includes swaggers, readmes, and examples
-  const changedRmFiles = changedFiles.filter(resourceManager);
+  const changedRmFiles = changedFiles.filter((f) => specification(f) && resourceManager(f));
```
mikeharder: `specification()` and `resourceManager()` are decoupled, so callers must include both, if they want both

eng/tools/oav-runner/src/runner.ts
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env node
 
+import * as fs from "fs";
 import * as oav from "oav";
 import * as path from "path";
-import * as fs from "fs";
 
+import { example, getChangedFiles, swagger } from "@azure-tools/specs-shared/changed-files"; //getChangedFiles,
```
mikeharder, line 7: impl should now be identical

scbedd: Double checked and confirmed.

scbedd (approved):

.github/shared/src/changed-files.js
```diff
@@ -152,22 +176,13 @@ export function readme(file) {
...
-}
-
 /**
  * @param {string} [file]
  * @returns {boolean}
  */
 export function dataPlane(file) {
   // Folder name "data-plane" should match case for consistency across specs
-  return typeof file === "string" && specification(file) && file.includes("/data-plane/");
+  return typeof file === "string" && includesFolder(file, "data-plane");
```
raych1, line 181: This seems to be breaking change which will need to add the separated `specification` filter in current usage. What's the scenario of using these filters out of the 'specification' folder?

mikeharder: This was an intentional breaking change, for a few reasons:

1. If the filters are decoupled, they are more flexible.  Sometimes (spec-model, and in tests) you **want** to include files outside the top-level spec folder.
2. The only filter which needs a relative path (or a repo root) is "specification".  All the other filters work fine with absolute or relative paths.  So limit the complexity to this one filter.
3. If you only need changed-files under `specification`, it's better to pass this `getChangedFiles()` so `git diff` only looks under `specification` in the first place.  Rather than getting all changes files in the repo, and then filtering the results to only `/specification`.

I tried to update all existing usage of these filters, to include `specification()` if it was removed from another filter:

https://github.com/Azure/azure-rest-api-specs/pull/36096/files#diff-8073f8653e3550fbfd5430f68cfaa2facd66a992180284513da6e369c15baa28R34

Did I miss some usages?

mikeharder: I did miss some usages under eng/tools.  Fixing.

mikeharder: Taking a different approach:

1. Remove `specification()` filter, since it's an anti-pattern
2. Add `options.paths` property to `getChangedFiles()` APIs, to filter `git diff` at the source

scbedd (approved):

raych1 (approved):

raych1:
Tests need to be fixed, and the remained part looks good.

eng/tools/openapi-diff-runner/test/command-helpers.test.ts
```diff
@@ -206,17 +204,13 @@ describe("command-helpers", () => {
...
   const mockRmSync = vi.mocked(rmSync);
-  const mockPath = vi.mocked(path);
   const mockFileURLToPath = vi.mocked(fileURLToPath);
   const mockGetChangedFilesStatuses = vi.mocked(getChangedFilesStatuses);
 
   beforeEach(() => {
     vi.clearAllMocks();
 
     // Setup default mocks
-    mockPath.resolve.mockImplementation((...paths) => paths.join("/"));
```
mikeharder, line 217: @raych1: These mocks broke tests in this PR on Windows, and I think they should be unnecessary, because the runtime implementation should work even in tests, on all OS.

PR #36097, opened 2025-07-22
avs, mgmt, remove operationId

weidongxu-microsoft:
1. `operationId` decorator is to be deprecated. Hence it is removed in the PR.
2. The functionality of `operationId` is replaced by `clientLocation` and `clientName` in back-compatible.tsp
3. `clientLocation` also work for TCGC (and emitters), so we no longer need the `client` and `operationGroup` in client.tsp (and service dev no longer need to modify client.tsp upon adding new API)

The only diff in Swagger is the order of items in `tags` array, which I think does not affect API or any downstream processing.
image

## Purpose of this PR

## Due diligence checklist

## Getting help


msyyc (approved):
Looks good to python.

cataggar (approved):

weidongxu-microsoft:
Hi @live1206 , could you take a quick look on .NET?
https://spa.apiview.dev/review/bd956b0f8b9a4d07b1a5e3a6ffa7ac5c?activeApiRevisionId=295d9baaee2e4a38992a9c8d99086dc8&diffApiRevisionId=62b707f9ce974b4aa914c1d6002ac381&diffStyle=trees

When I put `clientLocation` and `clientName` to back-compatible.tsp, I didn't exclude .NET.
I'd expect .NET already override the names in client.tsp, and there seems no API change in the apiview.
There is some diff in Factory class, and I am not sure whether this is caused by my change.

live1206:
> Hi @live1206 , could you take a quick look on .NET? https://spa.apiview.dev/review/bd956b0f8b9a4d07b1a5e3a6ffa7ac5c?activeApiRevisionId=295d9baaee2e4a38992a9c8d99086dc8&diffApiRevisionId=62b707f9ce974b4aa914c1d6002ac381&diffStyle=trees
> 
> When I put `clientLocation` and `clientName` to back-compatible.tsp, I didn't exclude .NET. I'd expect .NET already override the names in client.tsp, and there seems no API change in the apiview. There is some diff in Factory class, and I am not sure whether this is caused by my change.

@weidongxu-microsoft We should exclude `csharp` from `@clientLocation` because this decorator is causing the corresponding interface missing `@armResourceoperations` decorator. And autorest.csharp depends on it to [identify resources](https://github.com/Azure/autorest.csharp/blob/e21427eedc1e335065c61cbc7aea1304b346b52f/src/AutoRest.CSharp/Mgmt/Models/OperationSet.cs#L51-L52).
From the APIView, we only see changes in ModelFactory, there should be resource model changes as well.
In the new csharp generator, this will be resolved since we detect resource in a better way.
Please exclude `csharp` for all coming `@clientLocation` reorganization.

live1206:
@raych1 
Why is `SDK Validation - .NET` green even there are breaking changes during `dotnet build`? It shows there are 15 build errors [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5118027&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=866), it is false positive.
```
/home/cloudtest/.nuget/packages/microsoft.dotnet.apicompat/5.0.0-beta.20467.1/build/Microsoft.DotNet.ApiCompat.targets(82,5): error : MembersMustExist : Member 'public Azure.ResourceManager.Avs.Models.AddonHcxProperties Azure.ResourceManager.Avs.Models.ArmAvsModelFactory.AddonHcxProperties(System.Nullable<Azure.ResourceManager.Avs.Models.AddonProvisioningState>, System.String)' does not exist in the implementation but it does exist in the contract. [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Azure.ResourceManager.Avs.csproj::TargetFramework=netstandard2.0]
04:52:22.550 cmdout 	[.Net] /home/cloudtest/.nuget/packages/microsoft.dotnet.apicompat/5.0.0-beta.20467.1/build/Microsoft.DotNet.ApiCompat.targets(82,5): error : MembersMustExist : Member 'public Azure.ResourceManager.Avs.AvsCloudLinkData Azure.ResourceManager.Avs.Models.ArmAvsModelFactory.AvsCloudLinkData(Azure.Core.ResourceIdentifier, System.String, Azure.Core.ResourceType, Azure.ResourceManager.Models.SystemData, System.Nullable<Azure.ResourceManager.Avs.Models.AvsCloudLinkStatus>, Azure.Core.ResourceIdentifier)' does not exist in the implementation but it does exist in the contract. [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Azure.ResourceManager.Avs.csproj::TargetFramework=netstandard2.0]
04:52:22.550 cmdout 	[.Net] /home/cloudtest/.nuget/packages/microsoft.dotnet.apicompat/5.0.0-beta.20467.1/build/Microsoft.DotNet.ApiCompat.targets(82,5): error : MembersMustExist : Member 'public Azure.ResourceManager.Avs.Models.AvsManagementCluster Azure.ResourceManager.Avs.Models.ArmAvsModelFactory.AvsManagementCluster(System.Nullable<System.Int32>, System.Nullable<Azure.ResourceManager.Avs.Models.AvsPrivateCloudClusterProvisioningState>, System.Nullable<System.Int32>, System.Collections.Generic.IEnumerable<System.String>)' does not exist in the implementation but it does exist in the contract. [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Azure.ResourceManager.Avs.csproj::TargetFramework=netstandard2.0]
```

weidongxu-microsoft:
@live1206 I've added !csharp to `clientLocation`. Now .NET validation does fail. I could not find anything from log that relates to my change though.

live1206:
> @live1206 I've added !csharp to `clientLocation`. Now .NET validation does fail. I could not find anything from log that relates to my change though.

From the error itself, it says custom code is not compatible with the latest generated .NET SDK
```
[.Net] /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Customized/AvsPrivateCloudResource.cs(125,33): error CS0246: The type or namespace name 'WorkloadNetworkDhcpResource' could not be found (are you missing a using directive or an assembly reference?) [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Azure.ResourceManager.Avs.csproj::TargetFramework=netstandard2.0]
```
Is there any breaking changes involved in the current api-version comparing to previous one?

live1206:
> > @live1206 I've added !csharp to `clientLocation`. Now .NET validation does fail. I could not find anything from log that relates to my change though.
> 
> From the error itself, it says custom code is not compatible with the latest generated .NET SDK
> 
> ```
> [.Net] /mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Customized/AvsPrivateCloudResource.cs(125,33): error CS0246: The type or namespace name 'WorkloadNetworkDhcpResource' could not be found (are you missing a using directive or an assembly reference?) [/mnt/vss/_work/1/s/azure-sdk-for-net/sdk/avs/Azure.ResourceManager.Avs/src/Azure.ResourceManager.Avs.csproj::TargetFramework=netstandard2.0]
> ```
> 
> Is there any breaking changes involved in the current api-version comparing to previous one?

I tried to generated .NET SDK locally, the result is fine with latest main. Your change is still causing resource identification not working, trying to figure out why.

specification/vmware/Microsoft.AVS.Management/workloadNetworks-dhcpConfigurations.tsp
```diff
@@ -9,16 +9,12 @@ using TypeSpec.Rest;
 #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator"
 // @armResourceOperations
```
live1206, line 10: why is this decorator commented out? This is causing the resource model not identified as Resource for csharp.

live1206:
> WorkloadNetworkDhcpResource

For some reason, the generated tspCodeModel.son from main doesn't contain any decorator at all for all clients. So, it skips the entire resource check in .NET generator due to missing [TypeSpec.@service](https://github.com/Azure/autorest.csharp/blob/e21427eedc1e335065c61cbc7aea1304b346b52f/src/AutoRest.CSharp/Mgmt/Models/OperationSet.cs#L46).
And this PR fixed the decorators, they start to show up in tspCodeModel.json, but there is 1 decorator missing for the interface in https://github.com/Azure/azure-rest-api-specs/pull/36097/files#r2241490639

specification/vmware/Microsoft.AVS.Management/workloadNetworks-dhcpConfigurations.tsp
weidongxu-microsoft: When I was trying other approach to move the operation, I saw the missing of `@armResourceOperations` on some interfaces. At that time, the problem was the conflict of `@autoRoute` (within `@armResourceOperations`) with the specific `@route` on the op within the interface (e.g. the `get`).

live1206: There are a bunch of other missing `@armResourceOperations`, they don't have the route issue while adding them back.
As offline discussed, for this one, we should just use `ArmResourceRead` instead, then we will check with service dev if they are OK with the breaking parameter order changes in swagger.

raych1:
> @raych1 Why is `SDK Validation - .NET` green even there are breaking changes during `dotnet build`? It shows there are 15 build errors [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5118027&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=866), it is false positive.

@live1206 .net script doesn't set the result to 'failed' for this step at https://github.com/Azure/azure-sdk-for-net/blob/ca6cfccb202e9f6517cbd71e87263d5f06877a4a/eng/scripts/automation/GenerateAndBuildLib.ps1#L794
Set the result to 'failed' if you want the pipeline fail when breaking change is detected. Instead of failing this step, I would suggest .net follow the same approach as other languages to onboard the SDK breaking change review workflow which supports the suppression.

live1206:
> > @raych1 Why is `SDK Validation - .NET` green even there are breaking changes during `dotnet build`? It shows there are 15 build errors [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5118027&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=866), it is false positive.
> 
> @live1206 .net script doesn't set the result to 'failed' for this step at https://github.com/Azure/azure-sdk-for-net/blob/ca6cfccb202e9f6517cbd71e87263d5f06877a4a/eng/scripts/automation/GenerateAndBuildLib.ps1#L794 Set the result to 'failed' if you want the pipeline fail when breaking change is detected. Instead of failing this step, I would suggest .net follow the same approach as other languages to onboard the SDK breaking change review workflow which supports the suppression.

@raych1 For other languages, they can take up breaking changes and bump major version. That's not the case for .NET, we will always fail the build and fix the breaking changes with custom code in sdk repo. So, in that sense there is no point to add suppression in spec CI, we can leave it as it is. Breaking changes can't be handled in spec PR anyway if we don't touch .NET SDK code.

weidongxu-microsoft:
I've added back a `csharp-client.tsp` for customization of `@client Customization` for "csharp" along, to unblock .NET
https://github.com/Azure/azure-rest-api-specs/pull/36097/commits/c42e44f1f0a80910de7f09a1fc6bb296b9f50d98

(the cause was captured [here](https://github.com/Azure/azure-rest-api-specs/pull/36097#discussion_r2241490639) -- it is not caused by this PR, but the prior `@client Customization` masked the problem for .NET, and removing it for .NET would unmask the problem)

live1206 (approved):

PR #36112, opened 2025-07-22
Ensure that `issue_number` is available to workflow triggered workflows

scbedd:
[Example PR in fork repo](https://github.com/scbedd/azure-rest-api-specs/pull/3)


.github/workflows/_reusable-set-check-status.yaml
```diff
@@ -61,8 +61,28 @@ jobs:
...
+        name: Set Issue Number
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            const { extractInputs } =
+              await import('${{ github.workspace }}/.github/workflows/src/context.js');
+            const { setOutput } = await import('${{ github.workspace }}/.github/shared/src/error-reporting.js');
+            const { issue_number } = await extractInputs(github, context, core);
+            await setOutput("issue_number", issue_number);
```
mikeharder: Please add this to set-status.js, instead of adding another github-script block.  Example:

https://github.com/Azure/azure-rest-api-specs/blob/6fa80ad3a37a580d5c29f5612e8ea536902fce4e/.github/workflows/arm-auto-signoff.yaml#L66-L73

.github/workflows/spec-gen-sdk-status.yml
```diff
@@ -47,3 +47,23 @@ jobs:
...
+        name: Set Issue Number
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            const { extractInputs } =
+              await import('${{ github.workspace }}/.github/workflows/src/context.js');
+            const { setOutput } = await import('${{ github.workspace }}/.github/shared/src/error-reporting.js');
+            const { issue_number } = await extractInputs(github, context, core);
+            await setOutput("issue_number", issue_number);
```
mikeharder: please add to spec-gen-sdk-status.js

.github/workflows/_reusable-set-check-status.yaml
scbedd: Got it. Then same diff with the spec-gen-status.

.github/workflows/spec-gen-sdk-status.yml
scbedd: Ok this is done. I just added at the top of each `Impl()` function so that it will always get done even if something else goes wrong.

.github/workflows/src/set-status.js
```diff
@@ -66,6 +66,8 @@ export async function setStatusImpl({
   requiredStatusName,
   overridingLabel,
 }) {
+  await core.setOutput("issue_number", issue_number);
```
mikeharder:
```suggestion
  core.setOutput("issue_number", issue_number);
```

API is sync

[[EngSys] Add linter rule for "await" of non-awaitable API (#36143)](#36143)

.github/workflows/src/spec-gen-sdk-status.js
```diff
@@ -39,18 +41,29 @@ export default async function setSpecGenSdkStatus({ github, context, core }) {
...
+  owner,
+  repo,
+  head_sha,
+  target_url,
+  github,
+  core,
+  issue_number,
+}) {
   const statusName = "SDK Validation Status";
+  await core.setOutput("issue_number", issue_number);
```
scbedd:
```suggestion
  core.setOutput("issue_number", issue_number);
```

.github/workflows/_reusable-set-check-status.yaml
```diff
@@ -56,13 +56,22 @@ jobs:
...
             return await setStatus(
-              { github, context, core }, 
+              { github, context, core },
               '${{ inputs.monitored_workflow_name }}',
               '${{ inputs.required_check_name }}',
               '${{ inputs.overriding_label }}'
             );
+
+      - if: |
+          always()
```
mikeharder:
```suggestion
      - if: ${{ always() && steps.set-status.outputs.issue_number }}
```

I prefer `${{ }}` for VS Code intellisense if the condition is short.

.github/workflows/_reusable-set-check-status.yaml
```diff
@@ -56,13 +56,22 @@ jobs:
...
+        name: Upload artifact with issue number
+        uses: ./.github/actions/add-empty-artifact
+        with:
+          name: "issue-number"
+          value: "${{ steps.set-status.outputs.issue_number }}"
```
mikeharder:
```suggestion
          name: issue-number
          value: ${{ steps.set-status.outputs.issue_number }}
```

nit (consistency): these don't need to be quoted, and we haven't quoted them elsewhere

.github/workflows/spec-gen-sdk-status.yml
```diff
@@ -47,3 +47,11 @@ jobs:
             const setStatus =
               (await import('${{ github.workspace }}/.github/workflows/src/spec-gen-sdk-status.js')).default;
             return await setStatus({ github, context, core });
+
+      - if: |
```
mikeharder: same comments as above

.github/workflows/src/spec-gen-sdk-status.js
mikeharder:
```suggestion
  core.setOutput("issue_number", issue_number);
```

mikeharder (approved):

PR #36116, opened 2025-07-22
Cleanup `breaking change` calls that are no longer necessary - move `array` functions to shared.

scbedd:
See title.

Follow-up from [this comment](https://github.com/Azure/azure-rest-api-specs/pull/35951#issuecomment-3100274476)


.github/shared/src/array.js
```diff
@@ -31,3 +31,23 @@ export async function flatMapAsync(array, asyncMapper) {
 export async function mapAsync(array, asyncMapper) {
   return Promise.all(array.map(asyncMapper));
 }
+
+/**
+ * @template T
+ * @param {T[]} array
+ * @param {T[]} values
+ * @returns
```
mikeharder:
```suggestion
 * @returns {boolean}

File: .github/shared/src/array.js [hunk 2]
```diff
@@ -31,3 +31,23 @@ export async function flatMapAsync(array, asyncMapper) {
...
+ */
+export function includesAll(array, values) {
```
mikeharder:
```suggestion
export function includesEvery(array, values) {
```

I may have suggested `includesAll`, but I think `includesEvery` may be better for symmetry with `array.every`.

.github/shared/src/array.js
```diff
@@ -31,3 +31,23 @@ export async function flatMapAsync(array, asyncMapper) {
 export async function mapAsync(array, asyncMapper) {
   return Promise.all(array.map(asyncMapper));
 }
+
+/**
+ * @template T
```
mikeharder:
```suggestion
/**
 * Returns true if `array` includes all elements in `values`
 *
 * @template T
```

.github/shared/src/array.js
```diff
@@ -31,3 +31,23 @@ export async function flatMapAsync(array, asyncMapper) {
...
+ * @param {T[]} array
+ * @param {T[]} values
+ * @returns
+ */
+export function includesAll(array, values) {
+  return values.every((value) => array.includes(value));
+}
+
+/**
+ * @template T
```
mikeharder, line 50:
```suggestion
/**
 * Returns true if `array` includes no elements from `values`
 *
 * @template T
```

.github/shared/test/array.test.js
```diff
@@ -37,4 +37,21 @@ describe("array", () => {
...
+    expect(includesAll(input, [4])).toBe(false);
+    expect(includesAll(input, [])).toBe(true);
+  });
+
+  it("includesNone", () => {
+    const input = [1, 2, 3];
+    const values = [4, 5];
+
+    expect(includesNone(input, values)).toBe(true);
+    expect(includesNone(input, [2])).toBe(false);
```
mikeharder, line 55: Add test for `includesNone(input, [])`

mikeharder (approved):

raych1 (approved):

PR #36119, opened 2025-07-22
[Devex Agent] Prompt for email

jeo02:
Just added an extra entry to ask for users email so it doesn't assume the GitHub one.


jeo02:
Closing as what we actually need is to simply not prompt for email at all in the release plan mcp tool.

PR #36120, opened 2025-07-22
Move common code + class definition updates

scbedd:
Addresses:

- [x] Clean up class definition per [this comment](https://github.com/Azure/azure-rest-api-specs/pull/35951#discussion_r2220370544)
- [x] Move some `isManagement` or `isReleaseBranch` to `SpecModel` or common helpers.
   - [x] Utilize `changed-files` utilities instead of redetecting in `IsDataPlanePR` and `isManagementPR`
   - [x] Moved `isReleaseBranch` to common code.

### DIFF
--- a/.github/shared/package.json
+++ b/.github/shared/package.json
@@ -20,7 +20,8 @@
     "./swagger": "./src/swagger.js",
     "./tag": "./src/tag.js",
     "./simple-git": "./src/simple-git.js",
-    "./test/examples": "./test/examples.js"
+    "./test/examples": "./test/examples.js",
+    "./branch": "./src/branch.js"
   },
   "bin": {
     "spec-model": "./cmd/spec-model.js"
--- a/.github/shared/src/branch.js
+++ b/.github/shared/src/branch.js
@@ -1,0 +1,9 @@
+/**
+ * Checks an input branch name and determines if it should be considered a release branch.
+ * @param {string} branchName
+ * @returns {boolean}
+ */
+export function isReleaseBranch(branchName) {
+  const branchRegex = [/main/, /RPSaaSMaster/, /release*/, /ARMCoreRPDev/];
+  return branchRegex.some((b) => b.test(branchName));
+}
--- a/.github/workflows/src/summarize-checks/labelling.js
+++ b/.github/workflows/src/summarize-checks/labelling.js
@@ -12,6 +12,8 @@ import {
   wrapInArmReviewMessage,
 } from "./tsgs.js";
 
+import { isReleaseBranch } from "../../../shared/src/branch.js";
+
 // #region typedefs
 /**
  * The LabelContext is used by the updateLabels() to determine which labels to add or remove to the PR.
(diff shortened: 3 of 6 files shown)


.github/shared/package.json
```diff
@@ -20,7 +20,8 @@
     "./swagger": "./src/swagger.js",
     "./tag": "./src/tag.js",
     "./simple-git": "./src/simple-git.js",
-    "./test/examples": "./test/examples.js"
+    "./test/examples": "./test/examples.js",
```
scbedd, line 23: I did notice this looks odd. Was this [intentional?](https://github.com/Azure/azure-rest-api-specs/commit/e3131c24003dc8a9485328efa2a4eae004bc46aa) If so I'll continue leaving this alone.

.github/shared/src/branch.js
```diff
@@ -0,0 +1,9 @@
+/**
```
scbedd, line 1: I didn't really see a good place to put something `branch` related like this. I DID consider `sdk-types.js`, but that seemed like it was type related specifically so I didn't put it there.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -190,31 +192,36 @@ import {
  *
  * See also: https://aka.ms/SpecPRReviewARMInvariants
  */
-// todo: inject `core` for access to logging
 export class Label {
+  /**
+   * The name of the label.
+   * @type {string}
+   */
+  name;
```
mikeharder, line 200:
```suggestion
  #name;
```

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -190,31 +192,36 @@ import {
...
+
+  /**
+   * Is the label currently present on the pull request?
+   * @type {boolean}
+   */
+  present;
```
mikeharder, line 206:
```suggestion
  #present;
```

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -190,31 +192,36 @@ import {
...
+
+  /**
+   * Should this label be present on the pull request?
+   * Must be defined before applyStateChange is called.
+   * Not set at the construction time to facilitate determining desired presence
+   * of multiple labels in single code block, without intermixing it with
+   * label construction logic.
+   * @type {boolean | undefined}
+   */
+  shouldBePresent;
```
mikeharder, line 216:
```suggestion
  #shouldBePresent;
```

mikeharder (approved):
Approved pending rename of private fields

scbedd:
> Approved pending rename of private fields

They're not private. They need to be accessible as various code points interact with those values directly. I'll check how hard it will be to refactor so they actually are private tho.

mikeharder:
> > Approved pending rename of private fields
> 
> They're not private. They need to be accessible as various code points interact with those values directly. I'll check how hard it will be to refactor so they actually are private tho.

If they only need to be read (not written) by callers, you can add a getter:

https://github.com/Azure/azure-rest-api-specs/blob/e6d42354fa0c04d56c2e5e5c061453e76ba788cb/.github/shared/src/swagger.js#L176-L181

PR #36149, opened 2025-07-23
Update `summarize-checks` output

scbedd:
1. Making `summarize-impact` get the label status itself
2. Updating `summarize-checks` to output a bit more information about exactly what it would do after completion. This will make it easier to grep how we're doing out of the test logs.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -319,9 +319,7 @@ export async function summarizeChecksImpl(
   event_name,
   targetBranch,
 ) {
-  core.info(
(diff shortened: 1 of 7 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -825,10 +831,12 @@ function getCommentBody(
   violatedRequiredLabelsRules,
 ) {
   let bodyProper = "";
+  let automatedChecksMet = "pending";
```
scbedd, line 837: This will become a Enum type when I update the code that actually will update this check. I just want the output for what the output WOULD be ASAP.

mikeharder (approved):

PR #36154, opened 2025-07-23
Fixed issues in renaming file scenario to look for the candidate version for comparison

raych1:
* Updated `SpecModel` to include references from Swagger files in addition to the Swagger files themselves when returning results. 

* Enhanced `getVersionFromInputFile` to extract the version directly from Swagger file content if regex-based detection fails. 

* Improved `getReadmeFolder` to search upward for the closest `readme.md` file

* Updated tests

### DIFF
--- a/.github/shared/test/spec-model.test.js
+++ b/.github/shared/test/spec-model.test.js
@@ -449,7 +449,7 @@ describe("getSwaggers", () => {
     const specModel = new SpecModel(folder, options);
     const swaggers = await specModel.getSwaggers();
     // Should return an array (may be empty if no valid readmes in this fixture)
-    expect(swaggers.length).toBe(5);
+    expect(swaggers.length).toBe(9);
 
     // If swaggers are found, they should have the expected structure
     for (const swagger of swaggers) {
--- a/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
+++ b/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
@@ -416,7 +416,28 @@ export function getSpecModel(specRepoFolder: string, swaggerPath: string): SpecM
     throw new Error(`Could not determine readme folder for swagger path: ${swaggerPath}`);
   }
 
-  const fullFolderPath = path.join(specRepoFolder, folder);
+  let fullFolderPath = path.join(specRepoFolder, folder);
+  // If the initial folder doesn't exist, search upward for a folder with readme.md
+  while (
+    !existsSync(fullFolderPath) &&
+    !fullFolderPath.endsWith("resource-manager") &&
+    !fullFolderPath.endsWith("data-plane")
+  ) {
+    const parent = path.dirname(fullFolderPath);
+
+    // Prevent infinite loop if we reach the root
+    if (parent === fullFolderPath) {
+      break;
+    }
+
+    fullFolderPath = parent;
+    const readmePath = path.join(fullFolderPath, "readme.md");
+
+    // If we find a readme.md, use this folder
+    if (existsSync(readmePath)) {
+      break;
+    }
+  }
 
   // Return if the readme folder is not found which means it's a new RP
   if (!existsSync(fullFolderPath)) {
--- a/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
+++ b/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
@@ -423,7 +423,11 @@ export function getSpecModel(specRepoFolder: string, swaggerPath: string): SpecM
   }
 
   // If the initial folder doesn't exist, search upward for a folder with readme.md
-  while (!fullFolderPath.endsWith("resource-manager") && !fullFolderPath.endsWith("data-plane")) {
+  while (
+    isNewRp &&
+    !fullFolderPath.endsWith("resource-manager") &&
+    !fullFolderPath.endsWith("data-plane")
+  ) {
     const parent = path.dirname(fullFolderPath);
 
     // Prevent infinite loop if we reach the root
--- a/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
+++ b/eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
@@ -416,13 +416,14 @@ export function getSpecModel(specRepoFolder: string, swaggerPath: string): SpecM
     throw new Error(`Could not determine readme folder for swagger path: ${swaggerPath}`);
   }
 
+  let isNewRp = false;
   let fullFolderPath = path.join(specRepoFolder, folder);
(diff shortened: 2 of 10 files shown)


.github/shared/test/spec-model.test.js
```diff
@@ -449,7 +449,7 @@ describe("getSwaggers", () => {
     const specModel = new SpecModel(folder, options);
     const swaggers = await specModel.getSwaggers();
     // Should return an array (may be empty if no valid readmes in this fixture)
-    expect(swaggers.length).toBe(5);
+    expect(swaggers.length).toBe(9);
```
mikeharder, line 452: For clarity, please update test to check if `swaggers.map(s.path)` is equal to an array of expected strings, so we can see the paths.

eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
```diff
@@ -355,35 +362,49 @@ export function isInDevFolder(swaggerPath: string) {
...
+  if (!segments || segments.length < 3) {
+    return undefined;
+  }
 
-    // Look for "resource-manager" or "data-plane" in the path
-    const resourceManagerIndex = segments.findIndex((segment) => segment === "resource-manager");
-    if (resourceManagerIndex !== -1) {
-      return segments.slice(0, resourceManagerIndex + 1).join("/");
-    }
+  // Handle dev folder conversion
```
mikeharder, line 378: Fine for now, but please open an issue to remove this.  We shouldn't have any special code for the "dev" folder.

eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
```diff
@@ -355,35 +362,49 @@ export function isInDevFolder(swaggerPath: string) {
...
+  const boundaryIndex = resourceManagerIndex !== -1 ? resourceManagerIndex : dataPlaneIndex;
+
+  // Determine search range: from file's parent directory up to boundary (or root if no boundary found)
+  const startIndex = segments.length - 2; // Start from parent directory of the file
+  const minIndex = boundaryIndex !== -1 ? boundaryIndex : 2; // Stop at boundary or at least keep first 3 segments
 
-    const dataPlaneIndex = segments.findIndex((segment) => segment === "data-plane");
-    if (dataPlaneIndex !== -1) {
-      return segments.slice(0, dataPlaneIndex + 1).join("/");
+  // Search upward for readme.md
```
mikeharder, line 392: Fine for now, but with spec model I think this could be `swagger.tag.readme` instead of dir search.

eng/tools/openapi-diff-runner/src/utils/spec.ts
```diff
@@ -155,3 +157,20 @@ export function getBaseNameForSwagger(filePath: string, version: string = ""): s
...
+export function deduplicateSwaggers(swaggers: Swagger[]): Swagger[] {
+  // Deduplicate by path
+  const seen = new Set();
+  return swaggers.filter((swagger) => {
+    if (seen.has(swagger.path)) {
+      return false;
+    }
+    seen.add(swagger.path);
+    return true;
+  });
```
mikeharder, line 175:
```suggestion
  return (new Map(swaggers.map(s => [s.path, s])).values();
```

I think this is equivalent, with the small caveat `new Map()` uses last-wins, while your code uses first-wins, but I don't think you care.

mikeharder: Or you could even return `Map<string, Swagger>` which makes it clear there's exactly one per unique path.

eng/tools/openapi-diff-runner/test/utils/spec.test.ts
```diff
@@ -266,4 +267,96 @@ describe("Helper functions for version analysis", () => {
...
+    it("should return the same array when no duplicates exist", () => {
+      const mockSwaggers: MockSwagger[] = [
+        createMockSwagger(TEST_PATHS.stable2020_07_02),
+        createMockSwagger(TEST_PATHS.stable2020_08_04),
+        createMockSwagger(TEST_PATHS.preview2020_07_02),
+      ];
+
+      const result = deduplicateSwaggers(mockSwaggers);
+
+      expect(result).toHaveLength(3);
```
mikeharder, line 286: Order may not be stable, so you either need to sort or compare set equality

eng/tools/openapi-diff-runner/test/utils/spec.test.ts
```diff
@@ -266,4 +267,96 @@ describe("Helper functions for version analysis", () => {
...
+      expect(result.map((s) => s.path)).toEqual([
+        TEST_PATHS.stable2020_07_02,
+        TEST_PATHS.stable2020_08_04,
+        TEST_PATHS.preview2020_07_02,
+      ]);
+    });
+
+    it("should preserve the first occurrence when duplicates exist", () => {
```
mikeharder, line 311: Do you actually need this behavior?  If so, I think we need to sort the inputs, since there are no guarantees about ordering by default

eng/tools/openapi-diff-runner/src/utils/common-utils.ts
```diff
@@ -1,5 +1,6 @@
 import { FilePosition } from "../types/message.js";
 import { Context } from "../types/breaking-change.js";
+import { existsSync, readFileSync } from "node:fs";
```
mikeharder, line 3: Do you need to use sync APIs, or can you use async (and convert the call chain to async)?  Sync IO should only be used in exceptional cases in JS.

eng/tools/openapi-diff-runner/src/utils/common-utils.ts
```diff
@@ -111,13 +112,11 @@ export function getVersionFromInputFile(filePath: string, withPreview = false):
...
-      return folderName;
-    }
+  // If no regex match found, try to read version from file content
+  if (existsSync(filePath)) {
+    const fileContent = readFileSync(filePath, "utf8");
+    const parsedContent = JSON.parse(fileContent);
+    return parsedContent?.info?.version || "";
   }
 
   return "";
```
mikeharder, line 122: Does it ever make sense to return "" from getVersionFromInputFile()?  Should we throw instead?

mikeharder (approved):
Approved pending fixes for comments, or I can re-review after changes.

raych1:
> Approved pending fixes for comments, or I can re-review after changes.

Sure, I will create follow-up PRs and issues for the pending fixes.

eng/tools/openapi-diff-runner/src/detect-breaking-change.ts
raych1: https://github.com/Azure/azure-sdk-tools/issues/11335

PR #36186, opened 2025-07-24
Update `summarize-checks` output - add graceful exit

scbedd:
To make this easier to grep through instead of seeing a wall of red.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -283,6 +283,13 @@ export default async function summarizeChecks({ github, context, core }) {
   const targetBranch = context.payload.pull_request?.base?.ref;
   core.info(`PR target branch: ${targetBranch}`);
 
+  if (!issue_number) {
+    core.info(
+      "This summarize-checks was triggered off a workflow that doesn't provide the issue-number artifact, early exiting.",
+    );
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -319,7 +319,17 @@ export async function summarizeChecksImpl(
   event_name,
   targetBranch,
 ) {
-  core.info(`Handling ${event_name} event for PR #${issue_number} in ${owner}/${repo}.`);
+  core.info(
+    `Handling ${event_name} event for PR #${issue_number} in ${owner}/${repo}@${head_sha}.`,
+  );
+
+  // == null covers === undefined implicitly
+  if (issue_number == null || isNaN(issue_number) || issue_number <= 0) {
```
mikeharder:
```suggestion
  if (!issue_number) {
```

Checking for truthiness is simpler and should be sufficient.  I also think this should move into summarizeChecks(), since it's closely related to extractInputs().

We could also consider changing this warning to error.  But, there may be valid scenarios which don't care about issue_number.

https://github.com/Azure/azure-rest-api-specs/blob/844b3cd64693c1fd5a10bfa578e2fce8241881e6/.github/workflows/src/context.js#L220-L224

mikeharder (approved):
approved pending comment

PR #36201, opened 2025-07-25
Migrate devcenter to TypeSpec

pshao25:

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind. Here is the list you need to pay attention:
- You are referring to different versions of common type. We unified the common type version to v4.


specification/devcenter/DevCenter.Management/models.tsp
```diff
@@ -0,0 +1,3123 @@
...
+  /**
+   * All Customer-managed key encryption properties for the resource.
+   */
+  customerManagedKeyEncryption?: CustomerManagedKeyEncryption;
+}
+
+/**
+ * All identity configuration for Customer-managed key settings defining which identity should be used to auth to Key Vault.
+ */
+model CustomerManagedKeyEncryptionKeyEncryptionKeyIdentity {
```
msyyc: @pshao25 It seems that this model is not referred

specification/devcenter/DevCenter.Management/routes.tsp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+interface OperationStatusesOperationGroup {
+  /**
+   * Gets the current status of an async operation.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/no-response-body" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @summary("Get Operation Status")
+  @route("/subscriptions/{subscriptionId}/providers/Microsoft.DevCenter/locations/{location}/operationStatuses/{operationId}")
```
msyyc, line 37: python emitter generates lro API for this operation and I think it is because that the operation uses `ArmLroLocationHeader`. Sine it is not lro in original swagger, we shall not add `ArmLroLocationHeader` in typespec. @pshao25 for awareness.

specification/devcenter/DevCenter.Management/Catalog.tsp
```diff
@@ -0,0 +1,273 @@
...
+    Response = Catalog | ArmAcceptedLroResponse<
+      "Resource update request accepted.",
+      ArmAsyncOperationHeader<FinalResult = Catalog>
+    >
+  >;
+
+  /**
+   * Deletes a catalog resource.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
ArberH: @pshao25 aka.ms/tsp/conversion-fix doesn't seem to lead anywhere?

specification/devcenter/DevCenter.Management/models.tsp
```diff
@@ -0,0 +1,3101 @@
...
+union UsageUnit {
+  string,
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  Count: "Count",
+}
+
+/**
+ * The reason why the given name is not available.
+ */
+union CheckNameAvailabilityReason {
```
ArberH: @pshao25 
These seem to be from the ARM common types spec, do they need to be in our service specification or should we be importing these from a common place like before?

specification/devcenter/resource-manager/Microsoft.DevCenter/preview/2025-07-01-preview/DevCenterRP.json
```diff
@@ -0,0 +1,13411 @@
...
+              "$ref": "../../../../../common-types/resource-management/v4/types.json#/definitions/ErrorResponse"
+            }
+          }
+        },
+        "x-ms-examples": {
+          "ProjectCatalogs_Sync": {
+            "$ref": "./examples/ProjectCatalogs_Sync.json"
+          }
+        },
+        "x-ms-long-running
```
v-jiaodi: `type`, `timeZone`, `time` and `frequency` should be optional for `ScheduleProperties`?

v-jiaodi: `networkConnectionName`, `localAdministrator`, `licenseType` and `devBoxDefinitionName` should be optional for `PoolProperties`

v-jiaodi: `subnetId` should be optional for `NetworkProperties`

v-jiaodi: `sku` and `imageReference` should be optional for `DevBoxDefinitionProperties`.

specification/devcenter/DevCenter.Management/models.tsp
mcgallan: This mainly depends on whether the original swagger uses custom code or common types. Our ultimate goal is for the swagger generated by TypeSpec to be completely consistent with the original swagger.

specification/devcenter/DevCenter.Management/Catalog.tsp
mcgallan: There was a similar reply to this before, I don't know if you were satisfied with that reply:
image

specification/devcenter/resource-manager/Microsoft.DevCenter/preview/2025-07-01-preview/DevCenterRP.json
mcgallan: I'm not sure what the comment you added means. If there are issues, please place the comment in the correct location, and also confirm whether these issues are swagger problems or problems with the language you're using. This will make it easier for me to resolve the issues you encountered.

mcgallan: Same as above

mcgallan: Same

mcgallan: Same

ArberH: These properties are required in the existing spec, prior to conversion. Are we intentionally making these optional now?

ArberH: Also same comment as above, these were required in the existing swagger

ArberH: Also same comment as above, these were required in the existing swagger

ArberH: Also same comment as above, these were required in the existing swagger

kazrael2119: @ArberH In your existing swagger, your definition is
```
    "ScheduleProperties": {
      "description": "The Schedule properties defining when and what to execute.",
      "type": "object",
      "allOf": [
        {
          "$ref": "#/definitions/ScheduleUpdateProperties"
        }
      ],
      "properties": {
        "provisioningState": {
          "description": "The provisioning state of the resource.",
          "$ref": "commonDefinitions.json#/definitions/ProvisioningState",
          "readOnly": true
        }
      },
      "required": [
        "type",
        "frequency",
        "timeZone",
        "time"
      ]
    },
```
the `required` option will not work as the properties under `required` are from [`ScheduleUpdateProperties`](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/devcenter/resource-manager/Microsoft.DevCenter/preview/2025-07-01-preview/vdi.json#L1860-L1884) not `ScheduleProperties`. So these properties will always be optional for all api versions in sdk.
If you want to make these properties as required, you should add this option in `ScheduleUpdateProperties`. But this will introduce api level breaking changes I think

In this pr, @mcgallan needs to keep the old swagger order, so he adds these properties into [`ScheduleProperties`](https://github.com/Azure/azure-rest-api-specs/blob/e13e7b8099bd9eeff92ee5244d95677ed7f00280/specification/devcenter/resource-manager/Microsoft.DevCenter/preview/2025-07-01-preview/devcenter.json#L12852-L12867) which will introduce JS breakings:
```
"Parameter frequency of interface ScheduleProperties is now required",
"Parameter time of interface ScheduleProperties is now required",
"Parameter timeZone of interface ScheduleProperties is now required",
"Parameter type of interface ScheduleProperties is now required",
```

cc @MaryGao

MaryGao: @mcgallan and @ArberH I think ziwei is right that - the properties e.g time - extended from `allOf` - would be optional not required. Because the `required` constrain would only work for items in current `properties` not for `allOf` ones. But if this is not your expectation on service's behavior I would suggest we work on the fix in new api version or at least new pr. In this migration pr we should ensure the swagger equivalent. /cc @pshao25 for awareness.

pshao25: I think we should honor the existing behavior as what SDK behaves, i.e., all these properties should be explained as optional. @ArberH Could you confirm what impact it might cause to you?

ArberH: No impact actually, our service already handles these as if they were optional and we don't need to make any changes. I'll go ahead and fix the spec. Thank you

mcgallan: @ArberH We will address this issue in this PR, but there's a question: Will you fix this issue in the existing swagger, or should we directly fix this issue in this PR? However, this might cause some differences in the swagger.

ramoka178:
Please fix the [Swagger LintDiff](https://github.com/Azure/azure-rest-api-specs/actions/runs/17330300077) error that came in this PR

ArberH:
> Please fix the [Swagger LintDiff](https://github.com/Azure/azure-rest-api-specs/actions/runs/17330300077) error that came in this PR

Hey @ramoka178 . Here's the summary of the lint diff errors: 

The vast majority of the errors of type `ProvisioningStateMustBeReadOnly` seem to be in error. Any provisioningState properties are all set to readOnly in the spec. 

For `PatchBodyParametersSchema`, the required property "name" must not be required from PATCH request body comes from ARM's own SKU common type. We've had this error come up frequently and we've been able to get an exception. 

Finally, for `PathForNestedResource`, there's nothing we can do about this. It's an existing API that would be a huge breaking change to customers and our own service.

ramoka178:
> `ProvisioningStateMustBeReadOnly`

Could you follow this suggestion and see if it helps. 
https://teams.microsoft.com/l/message/19:0351f5f9404446e4b4fd4eaf2c27448d@thread.skype/1752603378693?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1752603378693&teamName=Azure%20SDK&channelName=API%20Spec%20Review&createdTime=1752603378693

ArberH:
> > `ProvisioningStateMustBeReadOnly`
> 
> Could you follow this suggestion and see if it helps. https://teams.microsoft.com/l/message/19:0351f5f9404446e4b4fd4eaf2c27448d@thread.skype/1752603378693?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1752603378693&teamName=Azure%20SDK&channelName=API%20Spec%20Review&createdTime=1752603378693

That worked! Thanks for the tip.

ramoka178:
@pshao25  There seems to be new errors now. Please try to fix them, or add suppressions in readme. The reviewer can take a look at the suppressions.

specification/devcenter/resource-manager/readme.md
```diff
@@ -28,6 +28,32 @@ These are the global settings for devcenter.
 openapi-type: arm
 openapi-subtype: rpaas
 tag: package-preview-2025-07-01-preview
+
+directive:
+  - suppress: PatchBodyParametersSchema
+    from: devcenter.json
```
mentat9: Please ensure these suppressions have where: clauses. Note: it's probably easier (and better) to suppress in .tsp files. I think those will automatically get a where: clause.

ArberH: Hey @mentat9, I'm having trouble suppressing the last linter error `PatchBodyParametersSchema` with the `where` clause. Have unfortunately spent quite some time trying to get this suppressed in the past.

The required value causing the `PatchBodyParametersSchema` error comes from the `Sku` model from ARM common types spec (something we cannot change). Our spec regularly gets flagged for referencing this, and the only way we were able to suppress it in the past was with a global file suppression. I had asked help from a previous ARM reviewer and for whatever reason we were unable to get the linter to stop complaining about it. There's probably an easy way to do it, I'm just unaware of what it may be.

The TypeSpec migration has taken 1-2 months so far and we're starting to worry that it will delay Microsoft Ignite deliverables for our team, as we are unable to make any API changes until this is merged. Would really appreciate any assistance!

mentat9: @ArberH - I think it should work, but I'm not the best person to RCA the problem. Couple of things to try (I don't know whether you've tried these already or not):
1.  I may have missed it (this is a big PR), but I didn't find the suppressions for PatchBodyParametersSchema in the Typespec. They may have been directly added to readme.md. If you are able to add them in the Typespec, I'd expect it to generate with a where clause. This is the ref: https://azure.github.io/typespec-azure/docs/howtos/arm/arm-rules/#detecting-and-suppressing-rule-violations-at-design-time.
2. There could be issues/nuances that only the experts know about, so it's worth asking on the Teams channel: https://teams.microsoft.com/l/channel/19%3A0351f5f9404446e4b4fd4eaf2c27448d%40thread.skype/API%20Spec%20Review?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47. Tell them everything you put in that comment: they will know how to fix it or that it can't be fixed.
If those don't resolve the problem, we won't block you on this forever. Let us know the results.

mikeharder: > Note: it's probably easier (and better) to suppress in .tsp files. I think those will automatically get a where: clause.

@mentat9: Currently, **LintDiff** errors can only be suppressed in readme.md, not *.tsp files or suppressions.yaml or anywhere else.

https://github.com/Azure/azure-rest-api-specs/wiki/Swagger-LintDiff#adding-scoped-suppressions

**TypeSpec Linter** errors should be suppressed in the `*.tsp` themselves.

ArberH: I guess then my question reverts back to my original one - I have not been able to successfully figure out a scoped suppression for this violation. 

Neither myself or the ARM reviewer of past have been able to get the scoped suppression to work and we settled for a global file suppression. It seems like this is no longer allowed. I'm not sure if the scoped suppression is possible due to a bug with the linter, or perhaps there's an easy way that neither of us are aware of. 

I'm happy to accept any path that moves this PR forward! Just let me know what change to make.

mikeharder: @arberH: Have you tried using exactly the steps documented at my link above?  Even if it doesn't fix the LintDiff error, can you push the result of these steps to your PR, so I can try to see what's going wrong?

So far, every case I've investigated, scoped suppressions in LintDiff do work correctly, if you exactly follow the steps in my doc.  It's always possible this time is different and a bug in the tooling.

ArberH: @mikeharder Pushed another "attempt" of the two locations that have this violation following the example above. I tried targeting `Sku` in one and `Sku.name` in the other just to try both. Neither seems to have worked. Does the suppression syntax look okay to you?

sandipsh:
i still see lots of validation errors. please fix them.

specification/devcenter/resource-manager/Microsoft.DevCenter/preview/2025-07-01-preview/devcenter.json
```diff
@@ -5482,2848 +6993,6343 @@
...
+          }
+        },
+        "x-ms-pageable": {
+          "nextLinkName": "nextLink"
+        }
+      }
+    },
+    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DevCenter/projects/{projectName}/pools/{poolName}/schedules/{scheduleName}": {
+      "get": {
+        "operationId": "Schedules_Get",
```
v-jiaodi: why remove parameter `top`?
image

ArberH: It's a GET request so it shouldn't be there, decided to fix the spec rather than suppress it. Including it causes a separate linter error.

ArberH:
> i still see lots of validation errors. please fix them.

Besides the `PatchBodyParametersSchema` that I may just have to use a global suppression for (see ongoing discussion), what other errors are you referring to?

specification/devcenter/resource-manager/readme.md
ArberH: Tried suppressing with these to no avail:

`$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DevCenter/devcenters/{devCenterName}/devboxdefinitions/{devBoxDefinitionName}"].patch.parameters[5].schema.properties.sku` 

and 

`$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DevCenter/projects/{projectName}/pools/{poolName}"].patch.parameters[5].schema.properties.devBoxDefinition.sku.name`

In a previous PR I thought that perhaps the indexing was wrong, so I tried every permutation of indexes and sku name referencing that I could think of, and none of them worked. Again, not claiming that the linter is broken, I just am not able to figure out how to scope it correctly. Since our initial spec pre-conversion had these suppressed globally, I am going to do the same here in hopes of progressing this PR review, and it's what we already had in the first place.

ArberH: If either of y'all have a scoped suppression/syntax that I can try, I would still appreciate that so I can try it! However as mentioned I will be moving forward with a global file suppression (as we once had) in the meantime.

specification/devcenter/resource-manager/readme.md
```diff
@@ -37,13 +112,7 @@ These settings apply only when `--tag=package-preview-2025-07-01-preview` is spe
 
 ```yaml $(tag) == 'package-preview-2025-07-01-preview'
 input-file:
-  - Microsoft.DevCenter/preview/2025-07-01-preview/commonDefinitions.json
   - Microsoft.DevCenter/preview/2025-07-01-preview/devcenter.json
-  - Microsoft.DevCenter/preview/2025-07-01-preview/vdi.json
-suppressions:
```
mikeharder:
```
suppressions:
  - code: PatchBodyParametersSchema
    from: devcenter.json
    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DevCenter/devcenters/{devCenterName}/devboxdefinitions/{devBoxDefinitionName}"].patch.parameters[5].schema.properties.properties
```

This should work.  You were using the wrong filename (and maybe the wrong "where" clause).

specification/devcenter/resource-manager/readme.md
mikeharder: @ArberH: You were using the wrong filename (and maybe the wrong where clause): https://github.com/Azure/azure-rest-api-specs/pull/36201#discussion_r2345378404

ArberH: Thank you @mikeharder !! It was the where clause! It's not the filename, the snippet that you referenced in the other comment was the existing working suppression that we had pre-conversion (understandable, it's a massive PR). That was removed after the conversion coalesced our 3 swaggers into 1. But the scoped reference you provided did the trick. 

For future reference I had:`.....schema.properties.sku` 
where it should have been `....schema.properties.properties` as per your comment. 

Thanks again!

mikeharder: > Thank you @mikeharder !! It was the where clause! It's not the filename, the snippet that you referenced in the other comment was the existing working suppression that we had pre-conversion (understandable, it's a massive PR). That was removed after the conversion coalesced our 3 swaggers into 1. But the scoped reference you provided did the trick.
> 
> For future reference I had:`.....schema.properties.sku` where it should have been `....schema.properties.properties` as per your comment.
> 
> Thanks again!

Also for future reference, I was able to add this suppression, by only following the instructions at https://github.com/Azure/azure-rest-api-specs/wiki/Swagger-LintDiff#adding-scoped-suppressions.  I don't know why they didn't work for you, but I followed them exactly, and they worked on my first try.

Following the doc, you end up with this jsonpath:

```
"jsonpath": [
    "paths",
    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DevCenter/devcenters/{devCenterName}/devboxdefinitions/{devBoxDefinitionName}",
    "patch",
    "parameters",
    "5",
    "schema",
    "properties",
    "properties"
],
```

Which converts to `schema.properties.properties`.

snehanagendra (approved):
LGTM

PR #36258, opened 2025-07-28
Monitor Metrics Query Java Updates

jairmyree:
Updating TypeSpec for Metrics Query to include Java-specific generation changes.


lmazuel (approved):

lmazuel (approved):

specification/monitor/Monitor.Query.Metrics/client.tsp
```diff
@@ -11,6 +11,21 @@ namespace ClientCustomizations;
 
 using MonitorQueryMetrics;
 
+@@clientNamespace(MonitorQueryMetrics,
+  "com.azure.monitor.query.metrics.implementation.generated",
```
srnagar, line 15: We should not use `.generated` package in Java. We should just use `implementation` package.

jairmyree: Here's the PR to update it: https://github.com/Azure/azure-rest-api-specs/pull/36300

PR #36270, opened 2025-07-29
Fix `summarize-impact` failure to import lodash

scbedd:
Avoiding digging in to why it's necessary in favor of fixing actual issues for now. @mikeharder after this week I'm good with revisit to properly fix.

The failure I'm chasing appears [here](https://github.com/Azure/azure-rest-api-specs/actions/runs/16584357910/job/46906716218?pr=36265)

`lodash` is a CommonJS module, so you have to import it slightly different when in a `ES Module` package.

### DIFF
--- a/eng/tools/summarize-impact/src/impact.ts
+++ b/eng/tools/summarize-impact/src/impact.ts
@@ -6,7 +6,7 @@ import * as path from "path";
 
 import * as commonmark from "commonmark";
 import yaml from "js-yaml";
-import * as _ from "lodash";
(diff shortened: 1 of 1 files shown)


mikeharder (approved):
another option, or longer-term fix:

https://www.npmjs.com/package/lodash-es

scbedd:
I will absolutely take lodash-es I just can't burn any time on this rn. EDIT: I confirmed this properly reflected `suppressionREviewRequired` for `36265` which aligns with existing.

PR #36277, opened 2025-07-29
[summarize-impact] Add test pipeline

mikeharder:
Tests may currently be skipped, but this at least ensures formatting, and we'll want this eventually anyway.


scbedd (approved):
👍

PR #36307, opened 2025-07-29
[extract-inputs] If issue-number is invalid, set output to NaN instead of throwing

mikeharder:
- While this might hide errors like issue-number=foo, it prevents errors in cases like issue-number=null
- It's better to just set issue_number=NaN and let the caller handle it
- Likely, the caller will fail the same, regardless of whether the artifact was set to an invalid value, or if it was never set at all.

### DIFF
--- a/.github/workflows/src/context.js
+++ b/.github/workflows/src/context.js
@@ -208,12 +208,13 @@ export async function extractInputs(github, context, core) {
 
           if (key === "issue-number") {
             const parsedValue = Number.parseInt(value);
-            if (parsedValue) {
+            if (parsedValue > 0) {
               issue_number = parsedValue;
-              continue;
             } else {
-              throw new Error(`Invalid issue-number: '${value}' parsed to '${parsedValue}'`);
+              core.info(`Invalid issue-number: '${value}' parsed to '${parsedValue}'`);
+              issue_number = NaN;
             }
+            continue;
           }
         }
       }
--- a/.github/workflows/src/context.js
+++ b/.github/workflows/src/context.js
@@ -198,15 +198,15 @@ export async function extractInputs(github, context, core) {
       core.info(`artifactNames: ${JSON.stringify(artifactNames)}`);
 
       for (const artifactName of artifactNames) {
-        // If artifactName has format "issue-number=number", set issue_number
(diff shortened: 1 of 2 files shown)


.github/workflows/src/context.js
```diff
@@ -198,22 +198,23 @@ export async function extractInputs(github, context, core) {
...
+        // Else, if artifactName has format "issue-number=other-string", warn and set issue_number=NaN
+        // - Workflows should probably only set "issue-number" to positive integers, but sometimes set it to "null"
         // Else, if artifactName does not start with "issue-number=", ignore it
         const firstEquals = artifactName.indexOf("=");
         if (firstEquals !== -1) {
           const key = artifactName.substring(0, firstEquals);
-          const value = artifactName.substring(firstEquals + 1);
-
           if (key === "issue-number") {
+            const value = artifactName.substring(firstEquals + 1);
```
mikeharder, line 209: unrelated optimization to only extract value if key matches

.github/workflows/src/context.js
```diff
@@ -198,22 +198,23 @@ export async function extractInputs(github, context, core) {
...
             const parsedValue = Number.parseInt(value);
-            if (parsedValue) {
+            if (parsedValue > 0) {
```
mikeharder, line 211: Better to check `> 0`, since valid issue numbers must be positive integers

.github/workflows/src/context.js
```diff
@@ -198,22 +198,23 @@ export async function extractInputs(github, context, core) {
...
               issue_number = parsedValue;
-              continue;
             } else {
-              throw new Error(`Invalid issue-number: '${value}' parsed to '${parsedValue}'`);
+              core.info(`Invalid issue-number: '${value}' parsed to '${parsedValue}'`);
+              issue_number = NaN;
```
mikeharder, line 215: While this might hide errors like `issue-number=foo`, it prevents errors in cases like `issue-number=null`.  It's better to just set `issue_number=NaN` and let the caller handle it.  Likely, the caller will fail the same, regardless of whether the artifact was set to an invalid value, or if it was never set at all.

scbedd (approved):
This should fix summarize-checks.

PR #36325, opened 2025-07-30
[AI Agents] computer_use_preview tool typespec

mike-taylor99:
# Choose a PR Template


specification/ai/Azure.AI.Agents/tools/models.tsp
```diff
@@ -719,14 +790,40 @@ model RunStepDeepResearchToolCallDetails {
 //
 
 @doc("The data provided during a tool outputs submission to resolve pending tool calls and allow the model to continue.")
+@discriminator("type")
 @added(Versions.v2025_05_01)
 model ToolOutput {
+  #suppress "@azure-tools/typespec-azure-core/no-string-discriminator" "Existing"
+  @doc("The object type for the tool output.")
+  type: string = "function_call_output";
```
johanste, line 929: What are we trying to communicate here? I assume that `type` is always required in order to determine what kind of `ToolOutput` it is?

mike-taylor99: @johanste  I added this to distinguish between function tool output and computer tool output. OpenAI implementation only considered function tool output case and did not have `type`. I set default value to this so as to not break original implementation but also be able to submit other tool types which may come up like computer case

https://platform.openai.com/docs/api-reference/runs/submitToolOutputs

specification/ai/Azure.AI.Agents/tools/models.tsp
```diff
@@ -536,6 +572,41 @@ model RunStepMcpToolCall extends RunStepToolCall {
   serverLabel?: string;
 }
 
+@doc("""
+  A record of a call to a Computer Use tool, issued by the model in evaluation of a defined tool, that represents
+  executed computer automation actions.
+  """)
+@added(Versions.v2025_05_15_preview)
+model RunStepComputerUseToolCall extends RunStepToolCall {
```
glharper, line 648: We also require a RunStepDeltaComputerUseToolCall model and a RunStepDeltaComputerUseDetails model to be defined in run_steps/models.tsp

specification/ai/Azure.AI.Agents/run_steps/models.tsp
```diff
@@ -388,6 +388,35 @@ model RunStepDeltaAzureAISearchToolCall extends RunStepDeltaToolCall {
   azureAISearch: Record<string>;
 }
 
+/** Represents the Computer Use tool call in a streaming run step. */
+@added(Versions.v2025_05_15_preview)
+model RunStepDeltaComputerUseToolCall extends RunStepDeltaToolCall {
+  /** The object type, which is always "computer_use_preview". */
+  type: "computer_use_preview";
```
dargilco, line 431: None of the other tools in preview use a string that ends with "_preview". We need to align on this. Either all preview tools have a "type" that ends with "_preview", or none of them. I suggest removing the "_preview". Having "_preview" means REST API customers will need to update their code to remove the "_preview" part when they switch from using preview api-version to GA api-version. We want to reduce code churn and not force them to do changes.

dargilco: The doc string can say that this tool is in preview.

mike-taylor99: Would the tool type not match the tool name exactly? What are the other examples of the preview tools? In this case, its an OpenAI tool and OpenAI defines the tool with `_preview`
https://platform.openai.com/docs/guides/tools-computer-use

specification/ai/Azure.AI.Agents/tools/models.tsp
```diff
@@ -719,14 +790,40 @@ model RunStepDeepResearchToolCallDetails {
...
   @doc("The ID of the tool call being resolved, as provided in the tool calls of a required action from a run.")
   @encodedName("application/json", "tool_call_id")
-  toolCallId?: string;
+  toolCallId: string;
+}
 
-  @doc("The output from the tool to be submitted.")
-  output?: string;
+@doc("The output from a function tool to be submitted.")
+@added(Versions.v2025_05_01)
```
glharper, line 937: The FunctionToolOutput and the added type that Johan commented on are breaking changes to a GA API.

specification/ai/Azure.AI.Agents/runs/routes.tsp
```diff
@@ -179,7 +179,8 @@ namespace Azure.AI.Agents {
         @doc("A list of tools for which the outputs are being submitted")
         @encodedName("application/json", "tool_outputs")
         @madeOptional(Versions.v2025_05_15_preview)
-        toolOutputs?: ToolOutput[];
+        @typeChangedFrom(Versions.v2025_05_15_preview, ToolOutput[])
```
glharper: @dargilco Is this kind of change acceptable for a GA API?

specification/ai/Azure.AI.Agents/run_steps/models.tsp
dargilco: If this is an OpenAI tool and they use "computer_use_preview" (as Glenn also pointed out), I think that means we need to use the same string to be wire compatible.

specification/ai/Azure.AI.Agents/tools/models.tsp
```diff
@@ -866,6 +931,43 @@ model ToolOutput {
   output?: string;
 }
 
+@doc("The data provided during a tool outputs submission to resolve pending tool calls and allow the model to continue.")
+@discriminator("type")
+@added(Versions.v2025_05_15_preview)
+model StructuredToolOutput {
```
glharper, line 926: Taking this change at face value (still a breaking GA change), should the original ToolOutput have some sort of deprecation warning or removed decorator?

dargilco (approved):

PR #36326, opened 2025-07-30
[github.js] Align enums with coding convention

mikeharder:
- improves type comments
- only runtime change is object.freeze() to make readonly


.github/workflows/src/label.js
```diff
@@ -9,10 +9,3 @@ export const LabelAction = Object.freeze({
   Add: "add",
   Remove: "remove",
 });
-
```
mikeharder, line 12: unrelated dead code

PR #36336, opened 2025-07-30
[EngSys] Remove deprecated "dev" folder

mikeharder:
- According to @heaths, this folder is no longer being used


mikeharder:
Avocado is failing because it doesn't limit it's analysis to the "specification" folder, so it's trying to analyze the "before" state of the "dev" folder, which fails.  Adding label `Approved-Avocado`.

heaths (approved):

PR #36339, opened 2025-07-30
[summarize-impact] Remove unused code

mikeharder:


eng/tools/summarize-impact/src/cli.ts
```diff
@@ -130,19 +130,10 @@ export async function main() {
...
-  // sets by default are not serializable, so we need to convert them to arrays
-  // before we can write them to the output file.
-  function setReplacer(_key: string, value: any) {
-    if (value instanceof Set) {
-      return [...value];
-    }
-    return value;
-  }
-
-  console.log("Evaluated impact: ", JSON.stringify(impact, setReplacer, 2));
```
scbedd, line 142: ~~Not impact itself, but there is member that is returned that DOES have sets in it (labelContext), and those don't display correctly without it. That being said I'm pretty certain I'm going to remove that part of ImpactAssessment as we no longer use it in `summarize-checks.js`. Not a huge deal ig.~~

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -8,14 +8,13 @@ import { getRootFolder } from "@azure-tools/specs-shared/simple-git";
...
-import { join, resolve } from "path";
+import { join } from "path";
 import { LabelContext } from "./labelling-types.js";
 import { PRContext } from "./PRContext.js";
 
 export async function getRoot(inputPath: string): Promise<string> {
   try {
-    const gitRoot = await getRootFolder(inputPath);
-    return resolve(gitRoot.trim());
+    return await getRootFolder(inputPath);
```
mikeharder, line 17: `getRootFolder()` already calls `resolve()` and `trim()`.

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -130,19 +129,10 @@ export async function main() {
...
-  // before we can write them to the output file.
-  function setReplacer(_key: string, value: any) {
-    if (value instanceof Set) {
-      return [...value];
-    }
-    return value;
-  }
-
-  console.log("Evaluated impact: ", JSON.stringify(impact, setReplacer, 2));
+  console.log("Evaluated impact: ", JSON.stringify(impact, null, 2));
```
mikeharder, line 132: `impact` is never a `Set`, so I think this is equivalent.

eng/tools/summarize-impact/src/cli.ts
scbedd: I already removed it it's dead code. Signing off.

scbedd (approved):

eng/tools/summarize-impact/src/cli.ts
scbedd: Agreed I removed the objects with sets.

PR #36343, opened 2025-07-30
Migrate breaking change label processing logic from prSummary check

raych1:
* Added logic in `outputBreakingChangeLabelVariables` to handle the deletion of the `VersioningReviewRequired` label when the `BreakingChangeReviewRequired` label is added.

* Updated the `validateBreakingChange` function to ensure breaking change label variables are only output when the pull request targets a production branch.

* Introduced `checkPrTargetsProductionBranch` in `common-utils.ts` to determine whether a pull request targets a production branch. This function includes logic for both public (`main`) and private (`RPSaaSMaster`) production branches.

* Updated tests

* Test PRs:
https://github.com/raych1/azure-rest-api-specs/pull/106
https://github.com/raych1/azure-rest-api-specs/pull/107

### DIFF
--- a/eng/tools/openapi-diff-runner/src/utils/common-utils.ts
+++ b/eng/tools/openapi-diff-runner/src/utils/common-utils.ts
@@ -233,3 +233,20 @@ export function convertRawErrorToUnifiedMsg(
   };
   return JSON.stringify(result);
 }
+
+/**
+ * Check if the PR targets a production branch.
+ * The production branches are defined at https://aka.ms/azsdk/pr-brch-deep
+ * @param targetRepo The repository name, e.g. "azure-rest-api-specs" or "azure-rest-api-specs-pr".
+ * @param targetBranch The target branch of the PR.
+ * @returns {boolean} True if the PR targets a production branch, false otherwise.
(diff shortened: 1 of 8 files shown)


eng/tools/openapi-diff-runner/src/utils/common-utils.ts
```diff
@@ -233,3 +233,20 @@ export function convertRawErrorToUnifiedMsg(
...
 }
+
+/**
+ * Check if the PR targets a production branch.
+ * The production branches are defined at https://aka.ms/azsdk/pr-brch-deep
+ * @param targetRepo The repository name, e.g. "azure-rest-api-specs" or "azure-rest-api-specs-pr".
+ * @param targetBranch The target branch of the PR.
+ * @returns {boolean} True if the PR targets a production branch, false otherwise.
+ */
+export function checkPrTargetsProductionBranch(targetRepo: string, targetBranch: string): boolean {
```
mikeharder, line 244: I think we check this a few places, so should be moved to shared code, but fine for now.

mikeharder (approved):

PR #36347, opened 2025-07-30
[summarize-impact] Minor code improvements

mikeharder:


eng/tools/oav-runner/src/cli.ts
```diff
@@ -8,23 +8,10 @@ import {
...
-export async function getRootFolder(inputPath: string): Promise<string> {
-  try {
-    const gitRoot = await simpleGit(inputPath).revparse("--show-toplevel");
-    return resolve(gitRoot.trim());
-  } catch (error) {
-    console.error(
-      `Error: Unable to determine the root folder of the git repository.`,
-      `Please ensure you are running this command within a git repository OR providing a targeted directory that is within a git repo.`,
-    );
-    process.exit(1);
```
mikeharder, line 25: We no longer wrap calls to `getRootFolder()`, but I think the default behavior when throwing an unhandled exception will be good enough.

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -1,29 +1,15 @@
...
 import { Octokit } from "@octokit/rest";
-import fs from "fs";
+import { writeFile } from "fs/promises";
 import { parseArgs, ParseArgsConfig } from "node:util";
-import { join } from "path";
+import { resolve } from "path";
 import { LabelContext } from "./labelling-types.js";
 import { PRContext } from "./PRContext.js";
 
-export async function getRoot(inputPath: string): Promise<string> {
```
mikeharder, line 15: We no longer wrap calls to getRootFolder(), but I think the default behavior when throwing an unhandled exception will be good enough.

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -1,29 +1,15 @@
-#!/usr/bin/env node
```
mikeharder, line 1: I think this is only necessary on `cmd/summarize-impact.js`.

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -1,29 +1,15 @@
...
-
 import { getChangedFilesStatuses } from "@azure-tools/specs-shared/changed-files";
 import { setOutput } from "@azure-tools/specs-shared/error-reporting";
 import { evaluateImpact, getRPaaSFolderList } from "./impact.js";
 
 import { getRootFolder } from "@azure-tools/specs-shared/simple-git";
 import { Octokit } from "@octokit/rest";
-import fs from "fs";
+import { writeFile } from "fs/promises";
```
mikeharder, line 7: Use async unless impossible or inconvenient

eng/tools/summarize-impact/src/cli.ts
```diff
@@ -135,7 +121,8 @@ export async function main() {
   console.log("Evaluated impact: ", JSON.stringify(impact, null, 2));
 
   // Write to a temp file that can get picked up later.
-  const summaryFile = join(process.cwd(), "summary.json");
-  fs.writeFileSync(summaryFile, JSON.stringify(impact, null, 2));
+  // Intentionally doesn't use GITHUB_STEP_SUMMARY, since it's not a markdown summary for GH UI
+  const summaryFile = resolve("summary.json");
```
mikeharder, line 125: `resolve()` is basically equivalent to but simpler than `join(process.cwd())`

eng/tools/summarize-impact/src/diff-types.ts
```diff
@@ -1,5 +1,15 @@
-export type FileTypes = "SwaggerFile" | "TypeSpecFile" | "ExampleFile" | "ReadmeFile";
-export type ChangeTypes = "Addition" | "Deletion" | "Update";
+export enum FileTypes {
```
mikeharder, line 1: Allows using enum values (instead of literal strings) at runtime

eng/tools/summarize-impact/src/impact.ts
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
 
-import * as fs from "fs";
+import { existsSync, readFileSync } from "fs";
 import { glob } from "glob";
-import * as path from "path";
+import { dirname, join, resolve } from "path";
```
mikeharder, line 5: We may not be completely consistent, but we usually import only the specific APIs we need.

scbedd (approved):

scbedd:
This definitely looks reasonable to me.

PR #36388, opened 2025-08-01
update spec for tsp-azure/2921

live1206:
https://github.com/Azure/typespec-azure/pull/2921


live1206:
@weshaggard Any idea why git errorred 128 during [TSV](https://github.com/Azure/azure-rest-api-specs/actions/runs/16738206536/job/47381586547?pr=36388)?
> Error: Process completed with exit code 128.

It worked last week in this same PR.

weshaggard:
I'm not sure about the git error maybe @mikeharder has seen it with TSV

mikeharder:
> I'm not sure about the git error maybe @mikeharder has seen it with TSV

Was a temporary bug in branch typespec-next (only used for testing).  TSP was recently upgraded, and typespec-next should be reset to a good state currently.

PR #36406, opened 2025-08-01
Account for 0 required checks (due to branch w/out branch protection)

scbedd:
Also adding a few more tests to our `automated merge requirements met` tests to ensure we can handle various combinations of items.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -132,13 +132,6 @@ import path from "path";
  * @property {"pending" | keyof typeof CheckConclusion} result
  */
 
-// Placing these configuration items here until we decide another way to pull them in.
-const FYI_CHECK_NAMES = [
-  "Swagger LintDiff",
-  "SDK Validation Status",
-  "Swagger BreakingChange",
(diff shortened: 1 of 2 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -800,14 +793,12 @@ export function extractRunsFromGraphQLResponse(response) {
...
             // - AND it is marked as 'FYI' in this file's FYI_CHECK_NAMES array
-            else if (FYI_CHECK_NAMES.includes(checkRunNode.name)) {
-              fyiCheckRuns.push({
-                name: checkRunNode.name,
-                status: checkRunNode.status,
-                conclusion: checkRunNode.conclusion,
-                checkInfo: getCheckInfo(checkRunNode.name),
-              });
-            }
+            fyiCheckRuns.push({
```
scbedd, line 796: Making all "non-required" runs FYICheckRuns.

scbedd: In the case where there are not yet any registered checks (eg very very early triggering), we will resolve to `pending` because an `impactAssessment` will not be completed.

PR #36413, opened 2025-08-02
[AzureDataTransfer] Migrate to TypeSpec - First introduction to repo

joschung:
This PR introduces TypeSpec and its produced files.


paulb621 (approved):

specification/azuredatatransfer/AzureDataTransfer.Management/Flow.tsp
joschung, line 1: I understand that "model ArmFlowResponse is ArmResponse<Flow>;" is bad practice in wrapping the Resource with ArmResponse<>. However, removing the ArmResponse wrapper leads to a breaking change because it changes the contract of the body. 
This model/contract was introduced in previous versions via manual updates and not via typespec. This is the first time our specification is actually using typespec. The prior versions were manual and they just introduced the typespec files but did nothing with them.

kimsteven5 (approved):

PR #36426, opened 2025-08-04
Explicitly handle 0 required checks

scbedd:
This is addressing a case I see much more often in `azure-rest-api-specs-pr`, where there ARE NO required checks due to targeting different branches than `RPSaaSMain/Dev or main`.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -885,11 +885,10 @@ export async function createNextStepsComment(
   // if there are NO required runs, we also consider this to be a "requirements met" situation.
   // there is a possibility that this will be a false positive, but it is better than
   // assuming that the requirements are not met when they actually are.
-  const requiredCheckInfosPresent =
-    !requiredRuns.some((run) => {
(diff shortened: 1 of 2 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -876,16 +876,20 @@ export async function createNextStepsComment(
...
     .filter((run) => checkRunIsSuccessful(run) === false)
     .map((run) => run.checkInfo);
 
   // determine if required runs have any in-progress or queued runs
   // if there are any, we consider the requirements not met.
-  const requiredCheckInfosPresent = requiredRuns.some((run) => {
+  // if there are NO required runs, we also consider this to be a "requirements met" situation.
+  // there is a possibility that this will be a false positive, but it is better than
+  // assuming that the requirements are not met when they actually are.
+  const requiredCheckInfosPresent = requiredRuns.every((run) => {
```
scbedd, line 888: This is what the original code did for checking all the outcomes. This essentially handles the "empty" case for us, as `every` on a memberless array = `true`.

mikeharder (approved):

PR #36463, opened 2025-08-05
Cutting over to `Github Actions`

scbedd:
FYI @mikeharder @raych1 @weshaggard


mikeharder (changes requested):
LGTM, but blocking until we're ready to flip the switch

mikeharder:
replaced by #36570

Issue #36378, opened 2025-07-31
[Summarize Checks] Secondary rate limit querying GraphQL


mikeharder:
> GraphQL Rate Limit Information: {"limit":5000,"cost":1,"used":281,"remaining":5000,"resetAt":"2025-07-31T21:38:16Z"}
> GraphQL Rate Limit Information: {"limit":5000,"cost":1,"used":282,"remaining":5000,"resetAt":"2025-07-31T21:38:16Z"}
> RequestError [HttpError]: You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID 9C31:8954C:60104B:C1D441:688BE22C.

https://github.com/Azure/azure-rest-api-specs/actions/runs/16660551359/job/47156431735#step:4:37

PR #36474, opened 2025-08-06
Trying to get around the possibility of huge # of checksuites

scbedd:
See title. We're using `listForRef` which should limit to the most recent 1k.

I have confirmed that this handles the pathological case originally mentioned in #36378 -- specifically sha `251ac2d3a8392258548facaaae3416ca681cd45c` triggered for pr `36377`. We now know _why_ that SHA triggered, and it's related to what #36479 is fixing!

That being said, our code will now properly handle those pathological cases if they pop up again.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -732,7 +732,23 @@ export async function getCheckRunTuple(
         per_page: PER_PAGE_MAX,
       });
 
-      impactAssessmentWorkflowRun = workflowRuns[0].id;
+      if (workflowRuns.length === 0) {
+        core.warning(
+          `No workflow runs found for check suite ID: ${latestCheck._originalData.check_suite.id}`,
+        );
+      } else {
+        // Sort by updated_at to get the most recent run
+        const sortedRuns = workflowRuns.sort(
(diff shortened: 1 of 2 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -714,22 +610,143 @@ export async function getCheckRunTuple(
...
+    ) {
+      const workflowRuns = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, {
+        owner,
+        repo,
+        head_sha: head_sha,
+        check_suite_id: latestCheck._originalData.check_suite.id,
+        per_page: PER_PAGE_MAX,
+      });
+
+      impactAssessmentWorkflowRun = workflowRuns[0].id;
```
mikeharder: What if `workflowRuns.length` is `0` or `> 1`?

mikeharder (approved):
Overall logic looks correct.  See one comment about `workflowRuns.length`.

.github/workflows/src/summarize-checks/summarize-checks.js
scbedd: Yeah I should have left a `todo` on this. I'll spend some time understanding if we just invert in that case.

scbedd:
> Overall logic looks correct. See one comment about `workflowRuns.length`.

I sorted the workflowRuns by updated_at and kept the latest one.

PR #36479, opened 2025-08-06
[context.js] Flow head-sha like issue-number

### LINKED ISSUE #36477: [context.js] Flow `head_sha` from pull_request through workflow_run
## Overview

Similar to `issue-number`, we need to flow the `head_sha` from the original trigger via an artifact in trusted workflows.

Currently, our code has a bug.  If there's a trigger chain of:

```
pull_request -> workflow_run -> workflow_run
issue_comment -> workflow_run -> workflow_run
check_run -> workflow_run -> workflow_run
```

The last `workflow_run` trigger will use the `head_sha` of the previous `workflow_run` trigger, which is **NOT** the same as the `head_sha` of the initial trigger.  The former is the latest commit of the repo's **default branch**, while the latter is the latest commit of the PR's **head branch**.

We should be able to set the `head_sha` artifact in response to all triggers `workflow_run`, `issue_comment`, and `check_run`.  We can only set issue-number in response to the first two.

## Example
PR: https://github.com/mikeharder/azure-rest-api-specs/pull/31

Correct SHA: https://github.com/mikeharder/azure-rest-api-specs/actions/runs/16768286787/job/47477797817#step:3:23

```
Event name: workflow_run
Action: completed
Triggering workflow: Swagger Avocado - Set Status
Triggering workflow ID: 16768283088
Triggering run ID: 118
Triggering event: pull_request_target
Triggering status: completed
Triggering conclusion: success
Triggering branch: test-arm-auto
Triggering SHA: 998353a5474bf99c0657e69201296aed2d24d2a7
Triggering actor: mikeharder
🔗 Triggering workflow run URL: https://github.com/mikeharder/azure-rest-api-specs/actions/runs/16768283088
🔗 Associated PR: https://github.com/mikeharder/azure-rest-api-specs/pull/31
```

Incorrect SHA (latest commit in default branch): https://github.com/mikeharder/azure-rest-api-specs/actions/runs/16768305522/job/47477847919#step:3:23

```
Event name: workflow_run
Action: completed
Triggering workflow: Swagger Avocado - Set Status
Triggering workflow ID: 16768301954
Triggering run ID: 119
Triggering event: workflow_run
Triggering status: completed
Triggering conclusion: success
Triggering branch: arm-auto-status
Triggering SHA: d03bfcaa0f10e43c14290465e17b350913e43294
Triggering actor: mikeharder
🔗 Triggering workflow run URL: https://github.com/mikeharder/azure-rest-api-specs/actions/runs/16768301954
🔗 Associated PR: https://github.com/mikeharder/azure-rest-api-specs/pull/36472
```

mikeharder:
- Fixes #36477
- Adds artifact head-sha anywhere we added issue-number

Most important changes are in `context.js`.


scbedd (approved):

.github/workflows/spec-gen-sdk-status.yml
```diff
@@ -48,6 +48,13 @@ jobs:
               (await import('${{ github.workspace }}/.github/workflows/src/spec-gen-sdk-status.js')).default;
             return await setStatus({ github, context, core });
 
+      - if: ${{ always() && steps.sdk-validation-status.outputs.head_sha }}
```
weshaggard, line 51: Out of curiosity what does always() buy us here?

mikeharder: Without `always()`, if the previous step failed, the artifact wouldn't be uploaded.  We want to always upload the artifact, because it's necessary for downstream workflows, regardless of whether the previous step passed or failed.

mikeharder: Also, I believe DevOps pipelines and GitHub actions have the opposite defaults here.  If you just specify a bare conditional like `if: foo == "bar"`, DevOps defaults to prepending `always()`, while GitHub defaults to prepending `success()`.

So in DevOps, we need to use `succeededOrFailed()` when we don't want something to **always** run, like when a pipeline is cancelled.

But in GitHub, we need to do the opposite, which is use `always()` anywhere we don't want the default prefix of `success()`.

> You can use the following status check functions as expressions in if conditionals. A default status check of success() is applied unless you include one of these functions

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions

eng/tools/spec-gen-sdk-runner/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "module": "ESNext",
     "target": "ESNext",
     "moduleResolution": "node",
+    "allowJs": true,
```
mikeharder:
```suggestion
```

raych1 (approved):

PR #36497, opened 2025-08-06
Added more information to SDK PR description and output sdk repo diff

raych1:
- information added to the SDK PR: pipeline run URL, docs of release prerequisite
- print the diff of SDK repo after code is generated
- flip the `SkipPullRequestCreation` switch to 'true' by default

Test PR and runs:
https://github.com/raych1/azure-rest-api-specs/pull/118
https://dev.azure.com/azure-sdk/public/_build/results?buildId=5188896&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d5873a6c-aa4d-5779-a4f7-44513dfead83


eng/pipelines/spec-gen-sdk.yml
```diff
@@ -27,7 +27,7 @@ parameters:
     displayName: 'SDK release type'
   - name: SkipPullRequestCreation
     type: boolean
-    default: false
+    default: true
```
mikeharder, line 30: Why is this default changing from false to true?

mikeharder (approved):

eng/pipelines/spec-gen-sdk.yml
raych1: This PR is replaced by https://github.com/Azure/azure-rest-api-specs/pull/36502

PR #36508, opened 2025-08-07
[nginx] Revert TypeSpec updates

mikeharder:
- Reverts #36473, which was causing a live site incident


mikeharder:
Bypassed checks, because this was urgent, and is only a revert of the most recent comment.

weidongxu-microsoft (commented):
Ok on SDK side

PR #36542, opened 2025-08-08
Add Impact Fixtures

scbedd:
For exercising various pieces of `summarize pr impact` and allow much easier catching of issues going forward.


eng/tools/summarize-impact/test/impact.test.ts
```diff
@@ -0,0 +1,207 @@
...
+  .readdirSync(path.join(BEFORE_REPO_FIXTURE, "specification"), { withFileTypes: true })
+  .filter((d) => d.isDirectory())
+  .map((d) => path.join(BEFORE_REPO_FIXTURE, "specification", d.name));
+
+describe("Impact Detection - default fixture", () => {
+  let originalCwd: string;
+
+  beforeEach(() => {
+    // Save the original working directory
+    originalCwd = process.cwd();
```
mikeharder, line 22: We should be able to change the test, so we don't need to change cwd.

mikeharder (approved):

PR #36556, opened 2025-08-08
Migrate PostgreSql to TypeSpec

mcgallan:

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger needs to be fixed. Please handle all these issues properly and them contact Azure SDK Team:

1.the original swagger used the "x-ms-visibility": "internal" syntax, which doesn't have a corresponding syntax in TypeSpec.

2.threatProtectionName is an enum type, and when it's used as a parameter in the path, it cannot have a pattern added to it.

3.In TypeSpec, we cannot add a default value of null for model properties.

4.BackupsLongTermRetentionOperation is a resource model that uses ProxyResource from common types, but additionally sets id and name as required. This format is unacceptable for certain languages in TypeSpec implementation, so we ignored the required setting for id and name.

5.In the original swagger, there are many model properties that use 'x-ms-external': true, but in TypeSpec this can only be added at the model level, so it's currently not implementable.


specification/postgresql/DBforPostgreSQL.Management/AdministratorMicrosoftEntra.tsp
```diff
@@ -0,0 +1,77 @@
...
+   * Gets information about a server administrator associated to a Microsoft Entra principal.
+   */
+  get is ArmResourceRead<AdministratorMicrosoftEntra>;
+
+  /**
+   * Creates a new server administrator associated to a Microsoft Entra principal.
+   */
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  @put
+  createOrUpdate is Azure.ResourceManager.Foundations.ArmCreateOperation<
```
pshao25: Why not use the standard templates?

specification/postgresql/DBforPostgreSQL.Management/BackupAutomaticAndOnDemand.tsp
```diff
@@ -0,0 +1,66 @@
...
+   */
+  get is ArmResourceRead<BackupAutomaticAndOnDemand>;
+
+  // FIXME: (ArmResourceCreateOrReplace): ArmResourceCreateOrReplaceAsync/ArmResourceCreateOrReplaceSync should have a body parameter.
+  /**
+   * Creates an on demand backup of a server.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  create is CustomArmResourceCreateOrReplaceAsync<
```
pshao25: Custom could be removed now?

specification/postgresql/DBforPostgreSQL.Management/Configuration.tsp
```diff
@@ -0,0 +1,89 @@
...
+   */
+  get is ArmResourceRead<Configuration>;
+
+  /**
+   * Updates, using Put verb, the value assigned to a specific modifiable configuration (also known as server parameter) of a server.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @useFinalStateVia("azure-async-operation")
+  @put
+  put is Azure.ResourceManager.Foundations.ArmCreateOperation<
```
pshao25: Same

nasc17:
Hi @mcgallan , we are currently planning to release a stable swagger in the next coming weeks. Should we have this migration based off what will soon be the newest version out? https://github.com/Azure/azure-rest-api-specs/pull/36219

@mattboentoro for awareness

nasc17:
@mattboentoro @nachoalonsoportillo 

These issues were brought up in start of PR of causing conflicts:
1.You're identifying that the TuningOptions_ListRecommendations operation uses an enum named recommendationType, which has the same name as another existing enum, but they contain different values. TypeSpec will treat them as the same enum during generation, which causes errors.

2.supportedHaMode has the same issue as recommendationType - there are duplicate enum names with different values that TypeSpec treats as the same enum during generation, causing conflicts.

3.the original swagger used the "x-ms-visibility": "internal" syntax, which doesn't have a corresponding syntax in TypeSpec.

4.the secret property in BackupStoreDetails is incorrectly placed within the items property, which is not the correct syntax.

5.threatProtectionName is an enum type, and when it's used as a parameter in the path, it cannot have a pattern added to it.

6.In TypeSpec, we cannot add a default value of null for model properties.

specification/postgresql/DBforPostgreSQL.Management/AdministratorMicrosoftEntra.tsp
mcgallan: fixed

specification/postgresql/DBforPostgreSQL.Management/BackupAutomaticAndOnDemand.tsp
mcgallan: fixed

specification/postgresql/DBforPostgreSQL.Management/Configuration.tsp
mcgallan: fixed

weidongxu-microsoft:
api-version 2025-08-01 breaks 2024-08-01 so much, that it is not possible to detect breaks from the conversion, in Java SDK.

Java will just greenlight this, and bump major version upon SDK from 2025-08-01

E.g.
- LRO return type change
  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2024-08-01/Configuration.json#L233-L239
  to
  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2025-08-01/Configurations.json#L225-L230
- Resource group name change
  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2024-08-01/FlexibleServers.json
  to
  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2025-08-01/Servers.json

Too many of these, and likely more I didn't really want to find.

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,228 @@
...
+
+  /**
+   * Updates an existing server. The request body can contain one or multiple of the properties present in the normal server definition.
+   */
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  @patch(#{ implicitOptionality: false })
+  update is ArmCustomPatchAsync<
+    Server,
+    PatchModel = ServerForPatch,
+    Response = ArmAcceptedLroResponse<LroHeaders = ArmAsyncOperationHeader &
```
tadelesh: Missing final result

specification/postgresql/DBforPostgreSQL.Management/FirewallRule.tsp
```diff
@@ -0,0 +1,71 @@
...
+  get is ArmResourceRead<FirewallRule>;
+
+  /**
+   * Creates a new firewall rule or updates an existing firewall rule.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<
+    FirewallRule,
+    Response = ArmAcceptedLroResponse<LroHeaders = ArmAsyncOperationHeader &
```
tadelesh: missing final result

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
mcgallan: fixed

specification/postgresql/DBforPostgreSQL.Management/FirewallRule.tsp
mcgallan: fixed

weidongxu-microsoft:
Service made a fix on response schema of PUT/PATCH https://github.com/Azure/azure-rest-api-specs/commit/0b98b195c68122bb56430a8cfe24b0489b4fbaa8

You will need to update the TypeSpec for it.

mcgallan:
> Service made a fix on response schema of PUT/PATCH [0b98b19](https://github.com/Azure/azure-rest-api-specs/commit/0b98b195c68122bb56430a8cfe24b0489b4fbaa8)
> 
> You will need to update the TypeSpec for it.

Updated.

weidongxu-microsoft:
This break is intentional, correct?

```
model HighAvailability {
  /**
   * High availability mode for a server.
   */
  mode?: PostgreSqlFlexibleServerHighAvailabilityMode;
```

Previously the enum is `HighAvailabilityMode`?
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/stable/2025-08-01/Servers.json#L1854

mcgallan:
> This break is intentional, correct?

This is to avoid the situation in TSP where enums with the same name are merged into one. Previously, the name for PostgreSqlFlexibleServerHighAvailabilityMode was also HighAvailabilityMode, which caused enum mixing. 
image

We used the rename feature in the .NET SDK to differentiate the two.
https://github.com/Azure/azure-sdk-for-net/blob/8f142ddecef7ede1cc57acb85a852185eb7101d6/sdk/postgresql/Azure.ResourceManager.PostgreSql/src/autorest.md?plain=1#L221-L222

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+    LtrPreBackupRequest,
+    Response = LtrPreBackupResponse & RequestIdResponseHeader,
+    BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<Server>,
+    Error = CustomErrorResponse
+  >;
+
+  /**
+   * Initiates a long term retention backup.
+   */
+  @action("startLtrBackup")
```
nasc17, line 133: "BackupsLongTermRetention_Start"

nasc17 (commented):
I noticed some actions that were in their own files, before are now within Server.tsp
Is this on purpose?

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+    Server,
+    Response = ArmResponse<ServerList>
+  >;
+
+  /**
+   * Lists the capabilities available for a given server.
+   */
+  @list
+  @get
+  @action("capabilities")
```
nasc17, line 103: CapabilitiesByServer_List
CapabilitiesByLocation_List

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+  list is ArmResourceActionSync<Server, void, ArmResponse<CapabilityList>>;
+
+  /**
+   * Lists all captured logs for download in a server.
+   */
+  @get
+  @list
+  @action("logFiles")
```
nasc17, line 111: CapturedLogs_ListByServer

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+  listByServer is ArmResourceActionSync<
+    Server,
+    void,
+    ArmResponse<CapturedLogList>
+  >;
+
+  /**
+   * Performs all checks required for a long term retention backup operation to succeed.
+   */
+  @action("ltrPreBackup")
```
nasc17, line 121: BackupsLongTermRetention_CheckPrerequisites

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+      Azure.Core.Foundations.RetryAfterHeader> &
+      RequestIdResponseHeader),
+    BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<Server>,
+    Error = CustomErrorResponse
+  >;
+
+  /**
+   * Checks if a proposed migration name is valid and available.
+   */
+  @action("checkMigrationNameAvailability")
```
nasc17, line 148: Migrations_CheckNameAvailability

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -0,0 +1,221 @@
...
+    ArmResponse<MigrationNameAvailability>
+  >;
+
+  /**
+   * Lists all read replicas of a server.
+   */
+  @get
+  @list
+  @action("replicas")
+  replicasListByServer is ArmResourceActionSync<
```
nasc17, line 162: Replicas_ListByServer

mcgallan:
> I noticed some actions that were in their own files, before are now within Server.tsp Is this on purpose?

These operations ended up in server.tsp because they were automatically generated there by the conversion tool, not due to our manual restructuring. Our migration process preserves the original auto-generated locations, which is why we have this current layout.

ramoka178:
@mcgallan [Swagger Avocado](https://github.com/Azure/azure-rest-api-specs/actions/runs/20125626806) is still failing. Is that expected ? 

@nasc17 Is this approved from PostgreSQL team ?

nasc17:
> @mcgallan [Swagger Avocado](https://github.com/Azure/azure-rest-api-specs/actions/runs/20125626806) is still failing. Is that expected ?
> 
> @nasc17 Is this approved from PostgreSQL team ?

@ramoka178 we are using this bug to track pre-existing issue: https://github.com/Azure/azure-rest-api-specs/issues/39199

But have received Avacado Approve to bypass until it can be addressed.

specification/postgresql/resource-manager/readme.md
```diff
@@ -31,37 +31,13 @@ openapi-type: arm
 tag: package-flexibleserver-2025-08-01
 ```
 
-``` yaml $(package-singleservers)
-tag: package-2020-01-01
```
nasc17, line 35: Please put this back to resolve build errors. We have gotten approvals to move forward with Two Default Tag Error. Removing this introduces a new set of issues we are not prepared for until we deprecate older APIs.

nasc17:
> > I noticed some actions that were in their own files, before are now within Server.tsp Is this on purpose?
> 
> These operations ended up in server.tsp because they were automatically generated there by the conversion tool, not due to our manual restructuring. Our migration process preserves the original auto-generated locations, which is why we have this current layout.

@mcgallan can you confirm does it matter the action name? Some are not up to date with what we have in current swagger. Would this impact SDK generation? Comments above have names taken .

specification/postgresql/resource-manager/readme.md
mcgallan: Fixed.

pshao25:
> > > I noticed some actions that were in their own files, before are now within Server.tsp Is this on purpose?
> > 
> > 
> > These operations ended up in server.tsp because they were automatically generated there by the conversion tool, not due to our manual restructuring. Our migration process preserves the original auto-generated locations, which is why we have this current layout.
> 
> @mcgallan can you confirm does it matter the action name? Some are not up to date with what we have in current swagger. Would this impact SDK generation? Comments above have names taken .

They are two different concepts. 

- @action("replicas") only effects route, which means {resourceUrl}/replicas. It doesn't impact SDK signature at all.

- Replicas_ListByServer is the operation id, which is used to generate SDK surface. In this case, [these two lines](https://github.com/Azure/azure-rest-api-specs/pull/36556/files#diff-5bc33305a2fe0008cd9d823b329c00b1e9846b8261dd2fe76a7eac31987f9bf3R149-R150) make sure the operation id keep the same. Check out this for details [Fixing Converted TypeSpec | TypeSpec Azure](https://azure.github.io/typespec-azure/docs/migrate-swagger/faq/conversionfix/#azure-toolstypespec-azure-coreno-openapi)

pshao25 (approved):

pshao25:
We got service team's approval to merge.
image

ArthurMa1978 (approved):

PR #36564, opened 2025-08-08
[widget] Update common-types to v6

mikeharder:


mikeharder:
@mikekistler, @JeffreyRichter: This is just our sample (no actual service), so the breaking changes shouldn't matter.  However, is it concerning that upgrading from v5 to v6 of common-types (as recommended by LintDiff) causes these breaking changes?  When and how are real services supposed to upgrade their version of common-types?

mikekistler:
Yes it is concerning that v5->v6 results in these breaking change warnings, particularly that AddedRequiredProperty ones, which are only acceptable when the property was actually required and simply not marked as such, which I believe is the case here. It would be great if we could find a way to suppress these specific issues.

mikeharder:
@mikekistler: Do you believe the v6 common-types should be updated to not be breaking, or that the breaking changes tool should be updated to not **consider** these changes as breaking?  I can help drive this, but I'm still not sure exactly what or where the problem is.

PR #36590, opened 2025-08-11
Check label name when set label action

### LINKED ISSUE #36536: [SDK Breaking Change Labels] May add artifact `label-=(true|false)`
WF "SDK Breaking Change Labels" may add an invalid artifact named `label-=false` or `label-=true`.

Examples:

- true
  - update labels: https://github.com/Azure/azure-rest-api-specs/actions/runs/16814496699
  - SDK Breaking Change Labels: https://github.com/Azure/azure-rest-api-specs/actions/runs/16814492237

false: https://github.com/Azure/azure-rest-api-specs/actions/runs/16817684261
 
I believe the root cause is spread across two files:

https://github.com/Azure/azure-rest-api-specs/blob/4dfa7495f0e2c2b9973c6e4c308576e2563db1b8/.github/shared/src/sdk-types.js#L26-L38

https://github.com/Azure/azure-rest-api-specs/blob/4dfa7495f0e2c2b9973c6e4c308576e2563db1b8/.github/workflows/src/sdk-breaking-change-labels.js#L89-L107

For Java and Net, the labels are undefined, which causes us to return objects like:

```
{ labelName: "", labelAction: LabelAction.Add, ... }
```

Which causes use to add an empty label artifact:

https://github.com/Azure/azure-rest-api-specs/blob/4dfa7495f0e2c2b9973c6e4c308576e2563db1b8/.github/workflows/sdk-breaking-change-labels.yaml#L51-L59

Fortunately, when update-labels processes this artifact, it's a no-op.  Trying to add or remove a label named "" should always fail and effectively be a no-op.

But, this still should be fixed in sdk-breaking-change-labels.  I think the fix could be as simple as this:

Image

We would now return:

```
{ labelName: "", labelAction: LabelAction.None, ... }
```

Which is a cleaner no-op, avoids adding the label artifact and triggering update-labels.

raych1:
Fixed https://github.com/Azure/azure-rest-api-specs/issues/36536

### DIFF
--- a/.github/workflows/src/sdk-breaking-change-labels.js
+++ b/.github/workflows/src/sdk-breaking-change-labels.js
@@ -99,8 +99,8 @@ export async function getLabelAndActionImpl({ details_url, core, retryOptions = 
     }
   }
 
-  if (!labelAction) {
-    core.info("No label action found, defaulting to None");
+  if (!labelAction || !labelName) {
+    core.info("No label action or name found, defaulting to None");
     labelAction = LabelAction.None;
   }
 
--- a/.github/workflows/test/sdk-breaking-change-labels.test.js
+++ b/.github/workflows/test/sdk-breaking-change-labels.test.js
@@ -138,6 +138,62 @@ describe("sdk-breaking-change-labels", () => {
         issueNumber: 123,
       });
     });
+    it("should correctly set labelAction to none when label name is empty", async () => {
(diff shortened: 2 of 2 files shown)


mikeharder (approved):

raych1:
TSV validation failure is unrelated and was fixed in main.

PR #36640, opened 2025-08-13
[summarize-checks] Use latest artifact named "job-summary"

mikeharder:


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -42,9 +41,7 @@ import {
   typeSpecRequirementDataPlaneTsg,
 } from "./tsgs.js";
 
-import fs from "fs/promises";
-import os from "os";
-import path from "path";
+import { unzipSync } from "zlib";
```
scbedd: And this is actually present? Nice.

mikeharder: yep builtin: https://nodejs.org/docs/latest-v20.x/api/zlib.html

scbedd (approved):
Makes total sense.

PR #36651, opened 2025-08-13
Update CODEOWNERS

ReilleyMilne:
# Request for a standard CODEOWNERS file across all SDK Language Repositories.

CODEOWNER changes:
Client SDKs renamed to Client Libraries and added Management Libraries as well. This change is to add standard category headers to ensure MCP Tools can access and edit CODEOWNERS files.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,9 +5,9 @@
 # Specification
 #########
 
-#########
-# Codeowner assignments are made from the _last_ matching entry in CODEOWNERS, so catch-all entries must come first
-#########
+####################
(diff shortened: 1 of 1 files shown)


weshaggard:
@ReillyMilne this repo is not the same as the other SDK repos as it doesn't contain any libraries only specs. We should undo this formatting of client and mgmt libraries.

ReilleyMilne:
> @ReilleyMilne this repo is not the same as the other SDK repos as it doesn't contain any libraries only specs. We should undo this formatting of client and mgmt libraries.

Ok sounds good!

PR #36663, opened 2025-08-14
java, dpg, tspconfig, remove wrong emitter-output-dir

weidongxu-microsoft:
Latest tsp-client (0.28.0) now honors the `emitter-output-dir` in tspconfig.yaml https://github.com/Azure/azure-sdk-tools/pull/11626

And for this 3, the `emitter-output-dir` in java emitter was wrong, and now causing failure in SDK when running `tsp-client update`.
(`{project-root}` would be at `/azure-sdk-for-java/sdk/communication/azure-communication-jobrouter/TempTypeSpecFiles/Communication.JobRouter`, none of these form can get back to the correct folder of `/azure-sdk-for-java/sdk/communication/azure-communication-jobrouter` -- the nearest one would be `{project-root}/../..`)

And it would cause tsp giving emitter wrong output dir.

## Purpose of this PR

## Due diligence checklist

## Getting help

### DIFF
--- a/specification/communication/Communication.JobRouter/tspconfig.yaml
+++ b/specification/communication/Communication.JobRouter/tspconfig.yaml
@@ -46,7 +46,6 @@ options:
     generate-test: true
     generate-sample: true
   "@azure-tools/typespec-java":
-    emitter-output-dir: "{java-sdk-folder}/sdk/{service-directory-name}/azure-communication-jobrouter"
     package-dir: "azure-communication-jobrouter"
     namespace: com.azure.communication.jobrouter
     partial-update: true
--- a/specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/tspconfig.yaml
+++ b/specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/tspconfig.yaml
@@ -49,7 +49,6 @@ options:
     slice-elements-byval: true
     flavor: "azure"
   "@azure-tools/typespec-java":
-    emitter-output-dir: "{project-root}"
     namespace: com.azure.messaging.eventgrid.systemevents
     package-dir: azure-messaging-eventgrid-systemevents
     customization-class: customization/src/main/java/EventGridSystemEventsCustomization.java
--- a/specification/eventgrid/Azure.Messaging.EventGrid/tspconfig.yaml
+++ b/specification/eventgrid/Azure.Messaging.EventGrid/tspconfig.yaml
@@ -34,7 +34,6 @@ options:
     namespace: Azure.Messaging.EventGrid.Namespaces
     model-namespace: false
   "@azure-tools/typespec-java":
-    emitter-output-dir: "{project-root}/.."
     namespace: "com.azure.messaging.eventgrid.namespaces"
     package-dir: azure-messaging-eventgrid-namespaces
     partial-update: true


XiaofeiCao (approved):

haolingdong-msft (approved):

specification/communication/Communication.JobRouter/tspconfig.yaml
```diff
@@ -46,7 +46,6 @@ options:
     generate-test: true
     generate-sample: true
   "@azure-tools/typespec-java":
-    emitter-output-dir: "{java-sdk-folder}/sdk/{service-directory-name}/azure-communication-jobrouter"
```
catalinaperalta, line 49: @weidongxu-microsoft I dont see why they couldnt they just use this: 

```
emitter-output-dir: "{output-dir}/{service-dir}/azure-communication-jobrouter"
```

It looks like for some reason they're trying to manually parse and separate out the service-dir instead just using the variable directly.

specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/tspconfig.yaml
```diff
@@ -49,7 +49,6 @@ options:
     slice-elements-byval: true
     flavor: "azure"
   "@azure-tools/typespec-java":
-    emitter-output-dir: "{project-root}"
```
catalinaperalta, line 52: Eventually everyone will be moved to emitter output dir..this should be:
```
emitter-output-dir: "{output-dir}/{service-dir}/azure-messaging-eventgrid-systemevents"
```

OR:

```
emitter-output-dir: "{output-dir}/{service-dir}/{package-dir}"
```

catalinaperalta: same for the other instance

specification/communication/Communication.JobRouter/tspconfig.yaml
weidongxu-microsoft: They can. But not based on `{java-sdk-folder}`, as it defined as `{project-root}/azure-sdk-for-java/` (where the `{project-root}` would be in that temp folder).

Yes, I am aware that eventually, this would be the path depends on `{output-dir}`.

But since the 0.28.0 is released recently, change tspconfig directly to the new pattern, and sync it to SDK repo (we need to sync, because current one causes error in pipeline), this may break dev's workflow (they may be using older version of it -- and that puzzled me a lot when diagnosing on the failure, because my local works fine as I am using the older version of tsp-client), unless properly notified (I will put a notice in java channel).
Hence here I am currently just make it use the old pattern.

PS: here I am trying to resolve an immediate failure in SDK repo, that latest tsp-client would use this wrong `emitter-output-dir` and then cause error in the pipeline -- there is a [pipeline in SDK that involve re-generate code upon PR](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5222656&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=70ccc7c9-8d04-5255-456e-9568fe142bcd), here it cannot load the proper file, because the emitter-output-dir is wrong

specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/tspconfig.yaml
weidongxu-microsoft: Agree, we will eventually use the new pattern

PR #36676, opened 2025-08-14
Generate classes for Azure Function call.

nick863:
…all and AzureFunctionToolCallDetails

# Choose a PR Template

### DIFF
--- a/specification/ai/Azure.AI.Agents/run_steps/models.tsp
+++ b/specification/ai/Azure.AI.Agents/run_steps/models.tsp
@@ -399,8 +399,7 @@ model RunStepDeltaAzureFunctionToolCall extends RunStepDeltaToolCall {
   type: "azure_function";
 
   @doc("Partial description of an Azure function call.")
-  @encodedName("application/json", "azure_function")
-  azureFunction: AzureFunctionToolCallDetails;
+  azure_function: AzureFunctionToolCallDetails;
 }
 
 /** Represents the Deep research in a streaming run step. */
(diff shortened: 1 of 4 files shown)


specification/ai/Azure.AI.Agents/run_steps/models.tsp
```diff
@@ -381,17 +381,28 @@ model RunStepDeltaBingGroundingToolCall extends RunStepDeltaToolCall {
...
 }
 
+@doc("Represents the Azure Function tool call in a streaming run step.")
+@added(Versions.v1)
+model RunStepDeltaAzureFunctionToolCall extends RunStepDeltaToolCall {
+  @doc("The object type, which is always 'azure_function'.")
+  type: "azure_function";
+
+  @doc("Partial description of an Azure function call.")
+  @encodedName("application/json", "azure_function")
```
trrwilson: This is a tangent for this PR in particular, but does it cause problems for us to use the snake-cased identifier as the variable name in the TypeSpec directly (that is, without use of `@encodedName`)?

It *used to be* that it'd generate a warning/error for non-standard naming, but I believe that's been relaxed for quite some time and specs will now accept that directly. Generators *should* meanwhile adjust the names accordingly since it's still a standard naming pattern.

Not important to address here (better done all at once) -- just raising it as an optimization note.

image

trrwilson (approved):

glharper (approved):

specification/ai/Azure.AI.Agents/run_steps/models.tsp
nick863: Fixed.

PR #36682, opened 2025-08-15
updated sip code exposing

juntuchen-msft:
# Choose a PR Template

### DIFF
--- a/specification/communication/data-plane/CallAutomation/preview/2024-09-01-preview/communicationservicescallautomation.json
+++ b/specification/communication/data-plane/CallAutomation/preview/2024-09-01-preview/communicationservicescallautomation.json
@@ -914,6 +914,53 @@
         }
       }
     },
+    "/calling/callConnections/{callConnectionId}:summarizeCall": {
+      "post": {
+        "tags": [
+          "MidConnections"
+        ],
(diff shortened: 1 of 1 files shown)


specification/communication/data-plane/CallAutomation/preview/2024-09-01-preview/communicationservicescallautomation.json
```diff
@@ -3364,6 +3426,23 @@
         }
       }
     },
+    "SummarizeCallRequest": {
+      "type": "object",
+      "properties": {
+        "operationContext": {
+          "description": "The value to identify context of the operation.",
+          "type": "string"
+        },
```
weidongxu-microsoft, line 3435: Not sure what value would be valid for this `operationContext` (can user always use `a`?)

But since this pattern is used in existing APIs, I would assume user know what to use.

weidongxu-microsoft (approved):

specification/communication/data-plane/CallAutomation/preview/2024-09-01-preview/communicationservicescallautomation.json
juntuchen-msft: value could be any string, it is for customer to make marks

PR #36703, opened 2025-08-15
[resourceManagement.yml] Stop adding all labels to PRs with no changed files

mikeharder:


weshaggard (approved):

mikeharder:
Fix does not appear to be working, even 10 minutes after merging.  Either the fix is incorrect, or there's a propagation delay.

https://github.com/Azure/azure-rest-api-specs/pull/36706

PR #36708, opened 2025-08-15
[EngSys] Delete unused pipelines

mikeharder:
- Have been running on every PR, but not required to pass
- private/main is already protected via GH setting
- PRs should no longer be targeting "master" instead of "main"


.azure-pipelines/BranchProtectionForPrivateRepo.yml
```diff
@@ -1,6 +0,0 @@
-jobs:
```
weshaggard, line 1: I think this one is still used in the private repo. Can you maybe replace it with an action or something to help tell folks not to merge into private main?

mikeharder: image

https://github.com/Azure/azure-rest-api-specs-pr/pull/24096

mikeharder: The check is run in private, but it's not required.  And I think the "Merging is blocked" message from GH branch protection is sufficient.

mikeharder: image

weshaggard: Yes that might be enough for folks but for folks that are in the maintain role (our team) will not be blocked and might end up accidentally merging which has happened in the past. I wonder if we maybe we could add a message or block in the Next Steps to Merge so that it gets blocked and folks can see the comment that they cannot merge into private main.

mikeharder: Do we have a similar check for private main in other repos?

Also, if we wanted this pipeline to block merges to main, it would need to be required.

mikeharder: > I wonder if we maybe we could add a message or block in the Next Steps to Merge so that it gets blocked and folks can see the comment that they cannot merge into private main.

One option, I could fold this check into our existing GH Action "Protected Files".  However, I have a separate work item to see if we can replace "Protected Files" with GH settings!  So, maybe I just need to step back and review all our checks related to "files/branches that shouldn't be changed"?

I'm still hopeful we can do this with GH settings, not custom code.

weshaggard: We don't currently have it in other repos but we also don't have as many folks trying to do work in those private repos.

I was thinking we just add the logic in the summarize check which is already required so if we detect a PR targeting private main we just fail that check which would block merging.

PR #36714, opened 2025-08-15
Rename LRO operations to internal for Python

cRui861:
Setting the operations that will have LRO methods to internal in favor of exposing the LRO method to customers instead.


wiboris (approved):
looks good

PR #36745, opened 2025-08-18
Summarize Checks Filtering

scbedd:
When we retrieve the workflowRuns for the SHA, we get workflowRuns from all PRs from that sha.

We can probably _entirely_ convert to `listCheckSuitesForRef` but the smallest possible logic change is to pull the runs, pull the checksuites, and then filter the runs by checksuites that are associated with the current target PR.


mikeharder (commented):
looks directionally correct, if the code works in all scenarios (e.g. fork PR to private repo)

scbedd:
It doesn't work. `listForRef` and `listSuitesForRef` do not populate `pull_requests` information. I think we're going to have to drop into GraphQL to retrieve this data.

PR #36748, opened 2025-08-18
Support customized branch when create sdk PR

raych1:
**Changes**
* Added logic to determine the correct `SdkRepoOwner` when a custom branch or commit is specified, defaulting to `'azure-sdk'` for non-main.

* Enhanced branch detection logic to check if the provided branch name is valid and set the pull request source branch accordingly.

* Updated `SdkRepoCommit` parameter to `SdkBranchName`.

* Removed the `SkipPullRequestCreation: true` setting in the batch pipeline, allowing PR creation as needed.

**Test pipeline runs and PRs**
SDK PR: https://github.com/Azure/azure-sdk-for-go/pull/25093
Runs: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5231670&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=c1986ca5-be0a-5bc2-c78d-61dbd0c1d7fc
This has no impact to spec PR: [example PR](https://github.com/Azure/azure-rest-api-specs/pull/36751)

### DIFF
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -317,7 +317,6 @@ stages:
             CommitMsg: $(GeneratedSDKInformation)
             TargetRepoOwner: $(SdkRepoOwner)
             TargetRepoName: $(SdkRepoName)
-            PushArgs: "--force"
             WorkingDirectory: $(SdkRepoDirectory)
             ScriptDirectory: $(SdkRepoDirectory)/eng/common/scripts
 
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -115,16 +115,11 @@ stages:
           Write-Host "Will use commitish for SDK repo checkout: '$checkoutCommitish'"
         displayName: "Create Run Time Variables"
 
-      - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.CreatePullRequest, true), ne(variables['Build.Reason'], 'PullRequest')) }}:
+      - ${{ if and(eq(variables['System.TeamProject'], 'internal'), ne(variables['Build.Reason'], 'PullRequest')) }}:
         - pwsh: |
             $sdkRepoBranch = '${{ parameters.SdkRepoBranch }}'
             $sdkRepoUrl = '${{ parameters.SdkRepoUrl }}'
             Write-Host "Validating SDK repo branch: '$SdkRepoBranch'"
-
-            if ($sdkRepoBranch.contains(":")) {
-              Write-Host "##vso[task.logissue type=error]'SDK repository branch' must be a commit SHA or branch name without ':'"
-              Exit 1
-            }
 
             if (($sdkRepoBranch -ne 'main') -and (-not [string]::IsNullOrWhiteSpace($sdkRepoBranch))) {
 
@@ -135,7 +130,7 @@ stages:
               }
 
               # Validate branch name
-              if ($sdkRepoBranch -match '^/|/$|//|\.\.|^@$|\.$/|[\x00-\x1F\x7F~^:?*@\[\]\\s]') {
+              if ($sdkRepoBranch -notmatch '^(?!/)(?!.*//)(?!.*\.\.)(?!.*\.lock$)(?!@$)(?!.*\.$)(?!.*\/$)(?!.*\\)(?!.*[~^:?*\[\]\s])[\u0020-\u007E]+$') {
                   Write-Host "##vso[task.logissue type=error]The provided SDK branch name '$sdkRepoBranch' is invalid."
                   Exit 1
               }
@@ -152,17 +147,16 @@ stages:
               $gitExitCode = $LASTEXITCODE
 
               if ($gitExitCode -ne 0) {
-                  Write-Host "##vso[task.logissue type=error]Failed to validate branch remotely (git exit code: $gitExitCode)"
-                  Exit $gitExitCode
+                Write-Host "##vso[task.logissue type=error]Failed to validate branch remotely (git exit code: $gitExitCode)"
+                Exit $gitExitCode
               }
 
               if ([string]::IsNullOrEmpty($branchExists)) {
-                  Write-Host "Branch '$SdkRepoBranch' does not exist in remote repository"
-                  Write-Host "Will checkout 'main'"
-                  $checkoutCommitish = 'main'
+                Write-Host "Branch '$SdkRepoBranch' does not exist in remote repository"
+                $checkoutCommitish = 'main'
               } else {
-                  Write-Host "Branch '$SdkRepoBranch' exists in remote repository"
-                  $checkoutCommitish = $SdkRepoBranch
+                Write-Host "Branch '$SdkRepoBranch' exists in remote repository"
+                $checkoutCommitish = $SdkRepoBranch
               }
             } else {
               Write-Host "Using commitish as-is: '$SdkRepoBranch' (main branch)"
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -130,7 +130,7 @@ stages:
               }
 
               # Validate branch name
-              if ($sdkRepoBranch -notmatch '^(?!/)(?!.*//)(?!.*\.\.)(?!.*\.lock$)(?!@$)(?!.*\.$)(?!.*\/$)(?!.*\\)(?!.*[~^:?*\[\]\s])[\u0020-\u007E]+$') {
+              if ($sdkRepoBranch -notmatch '^(?!refs/)(?!/)(?!.*//)(?!.*\.\.)(?!.*\.lock$)(?!@$)(?!.*\.$)(?!.*\/$)(?!.*\\)(?!.*[~^:?*\[\]\s])[\u0020-\u007E]+$') {
                   Write-Host "##vso[task.logissue type=error]The provided SDK branch name '$sdkRepoBranch' is invalid."
                   Exit 1
               }
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -3,7 +3,7 @@ parameters:
     type: string
   - name: SdkRepoUrl
     type: string
-  - name: SdkRepoCommit
+  - name: SdkRepoBranch
     type: string
     default: 'main'
   - name: ConfigType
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -206,9 +215,27 @@ stages:
           CustomCondition: and(succeededOrFailed(), ne(variables['StagedArtifactsFolder'], ''))
 
       - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.CreatePullRequest, true), ne(variables['Build.Reason'], 'PullRequest')) }}:
+        - pwsh: |
+            $sdkPrBranchName = "$(PrBranch)-$(Build.BuildId)"
+            $sdkRepoCommitish = '${{ parameters.SdkRepoCommit }}'
+            if (($sdkRepoCommitish -ne 'HEAD') -and ($sdkRepoCommitish -ne 'main')) {
+              # Query remote branches
+              $result = git ls-remote --heads origin $sdkRepoCommitish
+              if ([string]::IsNullOrEmpty($result)) {
```
weshaggard: What about cases where the branch doesn't exist yet? I would expect the pipeline to create based of main but still use that new branch name.

raych1: added the support for non-existed branch. Here is [test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5232839&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=fa40ef2d-deb2-5846-4ea3-7df100e508f4)

raych1: @weshaggard this is the [SDK PR](https://github.com/azure-sdk/azure-sdk-for-go/pull/14843/) created against a spec branch with multiple service spec changes. To clarify, it needs multiple times pipeline runs as each run can generate sdk based on only one typespec project.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,8 +110,67 @@ stages:
...
+          Write-Host "Will use commitish for SDK repo checkout: '$checkoutCommitish'"
         displayName: "Create Run Time Variables"
 
+      - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.CreatePullRequest, true), ne(variables['Build.Reason'], 'PullRequest')) }}:
+        - pwsh: |
+            $sdkRepoCommitish = '${{ parameters.SdkRepoCommit }}'
+            $sdkRepoUrl = '${{ parameters.SdkRepoUrl }}'
+            Write-Host "Validating SDK repo commitish: '$sdkRepoCommitish'"
+            # Check if it's a commit SHA
+            $commitShaPattern = '^[a-fA-F0-9]{7,40}$'
```
weshaggard: Couldn't this also match a legal branch name? `decaffe` as an example. I agree not likely, but it is possible. I'm not sure I see the value in supporting a sha over a branch, so maybe we make this only allow branches.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,8 +110,67 @@ stages:
...
+
+            if ($sdkRepoCommitish.contains(":")) {
+              Write-Host "##vso[task.logissue type=error]'SDK repository commit' must be a commit SHA or branch name without ':'"
+              Exit 1
+            }
+
+            if (($sdkRepoCommitish -ne 'main') -and ($sdkRepoCommitish -notmatch $commitShaPattern)) {
+                # set repo owner to 'azure-sdk' when customized branch is provided
+                $sdkRepoOwner = 'azure-sdk'
```
weshaggard: Should we make this a parameter so that we can let folks pick the repo of the branch to use for checking out?

weshaggard: Either that or we only support branches from the branches coming from azure-sdk fork. main should always match in both cases and if we don't support the sha's that would mean all branches come from the azure-sdk fork.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -125,13 +184,46 @@ stages:
...
-              Commitish: ${{ parameters.SdkRepoCommit }}
+              Commitish: $(CheckoutCommitish)
               WorkingDirectory: $(SdkRepoDirectory)
           SkipCheckoutNone: true
           ${{ if and(eq(variables['System.TeamProject'], 'internal'), endsWith(variables['Build.Repository.Name'], '-pr')) }}:
             TokenToUseForAuth: $(azuresdk-github-pat)
             PreserveAuthToken: true
 
+      - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.CreatePullRequest, true), ne(variables['Build.Reason'], 'PullRequest')) }}:
+        - pwsh: |
```
weshaggard: I don't think we need this step because https://github.com/Azure/azure-rest-api-specs/blob/main/eng/common/scripts/git-branch-push.ps1#L70 already handles this work.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,8 +110,67 @@ stages:
...
+                    Write-Host "##vso[task.logissue type=error]Failed to validate branch remotely (git exit code: $gitExitCode)"
+                    Exit $gitExitCode
+                }
+
+                if ([string]::IsNullOrEmpty($branchExists)) {
+                    Write-Host "Branch '$sdkRepoCommitish' does not exist in remote repository"
+                    Write-Host "Will checkout 'main' and create new branch '$sdkRepoCommitish' later"
+
+                    # Set flags for post-checkout processing
+                    Write-Host "##vso[task.setvariable variable=CreateNewBranch]true"
```
weshaggard: I don't think we need this. All we need is to set the `CheckoutCommitish` to either be main or the remote branch it exists. Then you need to always set your `SdkPullRequestSourceBranch` to either what was passed in if or the generated one if it isn't main. I think you can eliminate the other 2 steps you added below.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,67 +110,69 @@ stages:
...
             $sdkRepoUrl = '${{ parameters.SdkRepoUrl }}'
-            Write-Host "Validating SDK repo commitish: '$sdkRepoCommitish'"
-            # Check if it's a commit SHA
-            $commitShaPattern = '^[a-fA-F0-9]{7,40}$'
+            Write-Host "Validating SDK repo branch: '$SdkRepoBranch'"
 
-            if ($sdkRepoCommitish.contains(":")) {
-              Write-Host "##vso[task.logissue type=error]'SDK repository commit' must be a commit SHA or branch name without ':'"
+            if ($sdkRepoBranch.contains(":")) {
+              Write-Host "##vso[task.logissue type=error]'SDK repository branch' must be a commit SHA or branch name without ':'"
```
weshaggard:
```suggestion
              Write-Host "##vso[task.logissue type=error]'SDK repository branch' must be a branch name without ':'"
```

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,67 +110,69 @@ stages:
...
+            if (($sdkRepoBranch -ne 'main') -and (-not [string]::IsNullOrWhiteSpace($sdkRepoBranch))) {
+
+              # GitHub's limit is around 250 characters
+              if ($sdkRepoBranch.Length -gt 250) {
+                Write-Host "##vso[task.logissue type=error]The provided SDK branch name exceeds 250 characters."
+                Exit 1
+              }
+
+              # Validate branch name
+              if ($sdkRepoBranch -match '^/|/$|//|\.\.|^@$|\.$/|[\x00-\x1F\x7F~^:?*@\[\]\\s]') {
```
weshaggard: Where did you get this regex? Personally, I would just let git error when we try to use it.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -206,14 +267,25 @@ stages:
           CustomCondition: and(succeededOrFailed(), ne(variables['StagedArtifactsFolder'], ''))
 
       - ${{ if and(eq(variables['System.TeamProject'], 'internal'), eq(parameters.CreatePullRequest, true), ne(variables['Build.Reason'], 'PullRequest')) }}:
+        - pwsh: |
+            $sdkPrBranchName = "$(PrBranch)-$(Build.BuildId)"
```
weshaggard, line 265: Why not do this in the one step above?

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -206,14 +267,25 @@ stages:
...
+
+            # use the user provided branch name when SdkRepoOwner is set as 'azure-sdk'
+            if ('$(SdkRepoOwner)' -eq 'azure-sdk') {
```
weshaggard, line 268: If you do this above you should do it based on if the branch name is empty or main.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: We cannot do this in the step before sparse-checkout. The `PrBranch` variable is set by the 'generate sdk' step.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: It came from LLM. I'd like to fail fast rather than waiting until generation completes but fails in pushing code step.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: See above comment.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: Updated.

raych1:
> * Updated `SdkRepoCommit` parameter to `SdkBranchName`.

@praveenkuttappan this change in SDK generation pipeline might impact the release agent in terms of triggering it. 
CC: @maririos

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -110,8 +110,63 @@ stages:
...
+            if (($sdkRepoBranch -ne 'main') -and (-not [string]::IsNullOrWhiteSpace($sdkRepoBranch))) {
+
+              # GitHub's limit is around 250 characters
+              if ($sdkRepoBranch.Length -gt 250) {
+                Write-Host "##vso[task.logissue type=error]The provided SDK branch name exceeds 250 characters."
+                Exit 1
+              }
+
+              # Validate branch name
+              if ($sdkRepoBranch -notmatch '^(?!refs/)(?!/)(?!.*//)(?!.*\.\.)(?!.*\.lock$)(?!@$)(?!.*\.$)(?!.*\/$)(?!.*\\)(?!.*[~^:?*\[\]\s])[\u0020-\u007E]+$') {
```
weshaggard, line 133: I worry this regex might end up being more complex than we need and might block some scenarios. If you have tested it, we can leave it, but I don't think it is necessary because the git ls-remote call below will also verify the branch

weshaggard (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
praveenkuttappan: Currently SDK generation is triggered from open spec PR using `refs/pull/<pull number>/merge` and I think this regex will block this flow.

raych1: It did take some time for the testing. I agree the regex is a little complex. Let's get this merged since 'git ls-remote' call just returns empty result instead of failing for an invalid branch.

PR #36759, opened 2025-08-19
[Monitor.Ingestion] .NET client generation

jsquire:
# Summary

The focus of these changes is to update the Ingestion assets to support generation of the .NET client library.

### DIFF
--- a/specification/monitor/Monitor.Ingestion/client.tsp
+++ b/specification/monitor/Monitor.Ingestion/client.tsp
@@ -13,6 +13,11 @@ namespace ClientCustomizations;
   "java"
 );
 
+@@clientNamespace(LogsIngestion,
+  "Azure.Monitor.Ingestion",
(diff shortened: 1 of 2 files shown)


jorgerangel-msft (approved):

mikeharder:
@jsquire, @jorgerangel-msft: Why was this merged when check TypeSpecValidation was failing?  PRs should **never** be merged when this required check is failing.

jsquire:
@mikeharder:  I merged because it's a chicken-or-the-egg condition.   The process works for a new library with no prior art, but not an existing library being moved over from swagger.  You can't pass this check until the library with the change has merged, which can't be merged because it can't pass CI because the code gen check fails for an incorrect REST API spec.    Unless you have a simple fix for that, I'm going to continue to force these first so that the library PR goes green.

PR #36767, opened 2025-08-19
Summarize checks fail fast on duplicates

scbedd:
See title. Making users aware of the issue #36741

This uses two API calls:

- to get pr metadata
- to get prs for the same branch

And injects a an error that will show up as a blocker. The error response probably needs to be workshopped a bit.

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -161,13 +161,6 @@ const CHECK_METADATA = [
   },
   {
     precedence: 0,
-    name: "Duplicate PR Protection",
-    suppressionLabels: [],
-    troubleshootingGuide:
-      "This repo does not currently support submitting two PRs from the same commit SHA. To unblock, find any other open PRs from this source branch and close all but one.",
-  },
-  {
-    precedence: 0,
     name: "TypeSpec Validation",
     suppressionLabels: [],
     troubleshootingGuide: defaultTsg,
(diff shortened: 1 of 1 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -473,6 +473,55 @@ export async function summarizeChecksImpl(
...
+  const { data: pr } = await github.rest.pulls.get({ owner, repo, pull_number: prNumber });
+  const headRef = pr.head.ref;
+  if (!pr.head || !pr.head.repo) {
+    console.warn(
+      `Cannot determine head branch repo for PR #${prNumber}, skipping duplicate PR detection.`,
+    );
+    return undefined;
+  }
+  const headOwner = pr.head.repo.owner.login;
+  const relatedPRs = await github.paginate(github.rest.pulls.list, {
```
mikeharder, line 494: I wonder how `github.rest.pulls.list` (passing the head SHA) differs from `github.rest.repos.listPullRequestsAssociatedWithCommit`?

mikeharder: I think we need to use `listPullRequestsAssociatedWithCommit`, because we want to detect duplicate head SHAs, not head branch names.

mikeharder: We should extract the existing code from `context.js` and create a helper method `listPullRequestsForCommit(head_owner, head_repo, head_sha, base_owner, base_repo)`, which tries the following:

1. repos.listPullRequestsAssociatedWithCommit(head_owner, head_repo, head_sha)
    - Should work on fork and non-fork PRs to public, and non-fork PRs to private
2. pulls.list(base_owner, base_repo, orderby: updated-desc), client-side filter by head_sha
    - Should work on fork PRs to private
    - Max 100 pulls per API call, so maybe only search first page?
3. search(base_owner, base_repo, head_sha)
    - Should work on fork and non-fork PRs to public and private
    - Low rate limit of 30 req/min
    
https://github.com/Azure/azure-rest-api-specs/issues/36853

mikeharder: https://github.com/Azure/azure-rest-api-specs/issues/36853

PR #36825, opened 2025-08-20
Storage STG 100

seanmcc-msft:


raych1:
For breaking change review: `RemovedRequiredParameter` violations are false positives caused by [this issue](https://github.com/Azure/autorest/issues/5153) in the `autorest` tool.

mikeharder:
LintDiff was failing with 4 errors of type XmsEnumValidation: https://github.com/Azure/azure-rest-api-specs/actions/runs/17498527489?pr=36825.

Added label `Approved-LintDiff` to unblock PR, but future PRs must either fix the problem or add suppressions to `readme.md`.

mikeharder:
`ModelValidation` is failing with `XMS_EXAMPLE_NOT_FOUND`, because these specs have no examples.  This spec should either add examples, or we need to add a suppression mechanism for specs that are allowed to not provide examples.

Added label `Approved-ModelValidation` to unblock this PR.

weirongw23-msft (approved):

PR #36827, opened 2025-08-20
fix: format monitor.ingestion tsp files

jorgerangel-msft:


mikeharder (approved):

specification/monitor/Monitor.Ingestion/client.tsp
```diff
@@ -13,10 +13,7 @@ namespace ClientCustomizations;
   "java"
 );
 
-@@clientNamespace(LogsIngestion,
```
jsquire, line 16: Why is this format a problem but the Java slug that I coped is not?

mikeharder: I don't know, we just call `tsp format` and require your files to match.  Maybe line length?

maririos: Curious: which tool depends on `tsp format` is it all the tsp tool?

mikeharder: When check TypeSpecValidation was created, it was agreed the check should enforce `tsp format`, to ensure all specs use consistent formatting.

https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/format.ts

This enforcement could be relaxed, if there is consensus on the team it's more trouble than its worth.  But overall, we haven't gotten many complaints about this.  Running `tsp format` locally is just a small part of everything needed to get a spec PR ready to merge.

PR #36828, opened 2025-08-20
[Microsoft.Resources] Deployment Stacks next stable API version 2025-07-01

kalbert312:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+    },
+    {
+      "name": "DeploymentStacksWhatIfResultsAtSubscription"
+    },
+    {
+      "name": "DeploymentStacksWhatIfResultsAtManagementGroup"
+    }
+  ],
+  "paths": {
+    "/providers/Microsoft.Management/managementGroups/{managementGroupId}/providers/Microsoft.Resources/deploymentStackWhatIf": {
```
dantedallag: Shouldn't the path be to `deploymentStackWhatIfResults`?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+            "required": true,
+            "type": "string",
+            "pattern": "^[-\\w\\._\\(\\)]+$"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Azure operation completed successfully.",
+            "schema": {
+              "$ref": "#/definitions/DeploymentStackTemplateDefinition"
```
dantedallag: This should be a what-if result, right?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+          ]
+        },
+        "parameters": {
+          "type": "object",
+          "description": "Name and value pairs that define the deployment parameters for the template. Use this element when providing the parameter values directly in the request, rather than linking to an existing parameter file. Use either the parametersLink property or the parameters property, but not both.",
+          "additionalProperties": {
+            "$ref": "#/definitions/DeploymentParameter"
+          }
+        },
+        "extensionConfigs": {
```
dantedallag, line 3293: Are the extension configs and the deployment stack extensions the same thing? Do we need both?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+          "type": "string",
+          "format": "uri",
+          "description": "The link to the next page of items"
+        }
+      },
+      "required": [
+        "value"
+      ]
+    },
+    "DeploymentStackWhatIfResultProperties": {
```
dantedallag: Should we add a `DeploymentStackLastModified` property?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+            "description": "The property will not be set or updated."
+          }
+        ]
+      }
+    },
+    "DeploymentStackWhatIfResourceChange": {
+      "type": "object",
+      "description": "Information about a single resource change predicted by What-If operation.",
+      "properties": {
+        "before": {
```
dantedallag: Should before, after, and delta be nested in a property changes definition?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+          },
+          {
+            "name": "NoEffect",
+            "value": "noEffect",
+            "description": "The property will not be set or updated."
+          }
+        ]
+      }
+    },
+    "DeploymentStackWhatIfResourceChange": {
```
dantedallag: We may want `ResourceType` as a property. We can grab this from the `After` block of deployments what-if.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+          "description": "The resource id of the Deployment responsible for this change."
+        },
+        "symbolicName": {
+          "type": "string",
+          "description": "The symbolic name of the resource being changed."
+        },
+        "identifiers": {
+          "type": "object",
+          "description": "The extensible resource identifiers.",
+          "additionalProperties": {}
```
dantedallag: Is this required?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: I think we also want an `DeploymentExtension` property to pass through from deployments what-if.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3486 @@
...
+        "create",
+        "delete",
+        "detach",
+        "modify",
+        "noChange"
+      ],
+      "x-ms-enum": {
+        "name": "DeploymentStackWhatIfChangeType",
+        "modelAsString": true,
+        "values": [
```
dantedallag, line 2985: Add `Unsupported`.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
kalbert312: Extension configs are input-only. Deployment extensions are output-only.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
kalbert312: `"additionalProperties": {}` is generated by TSP.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
kalbert312: Extension configs reference extensions defined in the deployment template and are only defined for extensions that have config. Deployment extensions are sourced from the actual job runs + have validated configurations.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Yes, marked as readonly

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: I ran out of time to go back and look at whatif -- I thought regular whatif had a full before/after as well as the per-property versions (though it does seem redundant...)

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
```diff
@@ -638,4 +798,16 @@ model DeploymentStackValidateProperties {
    * The array of resources that were validated.
    */
   validatedResources?: ResourceReference[];
+
+  /**
+   * The deployment extensions.
+   */
+  @added(Versions.v2025_07_01)
+  deploymentExtensions?: DeploymentExtension[];
```
snarkywolverine, line 869: >deploymentExtensions [](http://example.com/codeflow?start=2&length=20)

@kalbert312 should this be read-only?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,377 @@
...
+ * DeploymentStack WhatIfResult Properties
+ */
+@added(Versions.v2025_07_01)
+model DeploymentStackWhatIfResultProperties is DeploymentStacksError {
+  ...DeploymentStackSharedProperties;
+
+  /**
+   * The deployment stack id to use as the basis for comparison.
+   */
+  deploymentStackId: string;
```
snarkywolverine: >deploymentStackId [](http://example.com/codeflow?start=2&length=17)

@dantedallag The example whatifresult I pulled had deploymentStackResourceId. Which is correct?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Does deployments generate the extension as part of its what-if output?

snarkywolverine: And for resource type - is that Microsoft.Storage/StorageAccounts for ARM resources and ... a string for extension resources?  Or is this more of an ARM/non-ARM differentiator?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+        "description": "Gets the Deployment stack with the given name.",
+        "parameters": [
+          {
+            "$ref": "../../../../../../common-types/resource-management/v6/types.json#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/ManagementGroupIdParameter"
+          },
+          {
+            "name": "deploymentStackWhatIfName",
```
dantedallag: min length/max length

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+          }
+        },
+        "x-ms-pageable": {
+          "nextLinkName": "nextLink"
+        }
+      }
+    },
+    "/providers/Microsoft.Management/managementGroups/{managementGroupId}/providers/Microsoft.Resources/deploymentStackWhatIfResults/{deploymentStackWhatIfName}": {
+      "get": {
+        "operationId": "DeploymentStacksWhatIfResultsAtManagementGroup_GetWhatIf",
```
dantedallag: nit: should this just be "..._Get"? to match the other operation ids?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+      "name": "DeploymentStacksWhatIfResultsAtSubscription"
+    },
+    {
+      "name": "DeploymentStacksWhatIfResultsAtManagementGroup"
+    }
+  ],
+  "paths": {
+    "/providers/Microsoft.Management/managementGroups/{managementGroupId}/providers/Microsoft.Resources/deploymentStackWhatIfResults": {
+      "get": {
+        "operationId": "DeploymentStacksWhatIfResultsAtManagementGroup_ListWhatIf",
```
dantedallag: nit: should this just be "..._List"? to match the other operation ids?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+          {
+            "name": "deploymentStackWhatIfName",
+            "in": "path",
+            "description": "The name of the DeploymentStackWhatIfResult",
+            "required": true,
+            "type": "string",
+            "pattern": "^[-\\w\\._\\(\\)]+$"
+          },
+          {
+            "name": "resource",
```
dantedallag: nit: should this be "deploymentStackWhatIfResult" instead of "resource"

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+              }
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "../../../../../../common-types/resource-management/v6/types.json#/definitions/ErrorResponse"
+            }
+          }
+        },
```
dantedallag: Should we add and "x-ms-examples" property like we are doing in stacks to point users towards an example?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+            "$ref": "#/parameters/ActionOnUnmanageDeleteParameters.unmanageActionResourceGroups"
+          },
+          {
+            "$ref": "#/parameters/ActionOnUnmanageDeleteParameters.unmanageActionManagementGroups"
+          },
+          {
+            "$ref": "#/parameters/ActionOnUnmanageDeleteParameters.unmanageActionResourcesWithoutDeleteSupport"
+          },
+          {
+            "$ref": "#/parameters/ActionOnUnmanageDeleteParameters.bypassStackOutOfSyncError"
```
dantedallag: I'm not sure this is needed. We can discuss.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+          },
+          "204": {
+            "description": "Resource does not exist."
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "../../../../../../common-types/resource-management/v6/types.json#/definitions/ErrorResponse"
+            }
+          }
```
dantedallag: Should we be marking this as a long running operation as we are doing with the PUT above?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+        }
+      }
+    },
+    "/providers/Microsoft.Management/managementGroups/{managementGroupId}/providers/Microsoft.Resources/deploymentStackWhatIfResults/{deploymentStackWhatIfName}/whatIf": {
+      "post": {
+        "operationId": "DeploymentStacksWhatIfResultsAtManagementGroup_WhatIf",
+        "tags": [
+          "DeploymentStacksWhatIfResultsAtManagementGroup"
+        ],
+        "description": "Exports the template used to create the Deployment stack at the specified scope.",
```
dantedallag: Change description.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+        "description": "Creates or updates a Deployment stack at the specified scope.",
+        "parameters": [
+          {
+            "$ref": "../../../../../../common-types/resource-management/v6/types.json#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/ManagementGroupIdParameter"
+          },
+          {
+            "name": "deploymentStackWhatIfName",
```
dantedallag: nit: should we be consistent and name this "deploymentStackWhatIfResultName"?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3502 @@
...
+        },
+        "resourceGroups": {
+          "$ref": "#/definitions/DeploymentStacksDeleteDetachEnum",
+          "description": "Specifies an action for a newly unmanaged resource. Delete will attempt to delete the resource from Azure. Detach will leave the resource in it's current state."
+        },
+        "managementGroups": {
+          "$ref": "#/definitions/DeploymentStacksDeleteDetachEnum",
+          "description": "Specifies an action for a newly unmanaged resource. Delete will attempt to delete the resource from Azure. Detach will leave the resource in it's current state."
+        },
+        "resourcesWithoutDeleteSupport": {
```
dantedallag: Do we have a task to track adding this to the service?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
kalbert312: I'm not sure it's necessary given this model is only used a return result. The `deploymentExtensions` in the stack properties model has it already because that model is used on PUT/POST.

kalbert312: It would also apply to all other properties, so maybe it would go on the model itself rather than the properties.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Done

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Does delete need to be long-running? My expectation was that it's a simple DB removal (no permission changes or external network calls required) so this can be synchronous.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: I had left it in for consistency sake. I would argue the service should accept a what-if request that contains the bypass flag, even if it takes no action.

My justification was that a customer should be able to take a single request body and use it with either the stack or what-if endpoints -- though I guess that's not entirely true because of the expiration timer.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: This appears to be auto-generated and not changeable.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed, assuming 1-90

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Fixed - was trying uniqueness to overcome initial build issues.

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: Changed to the *Resource* variant

specification/resources/DeploymentStacks.Management/models.tsp
```diff
@@ -286,6 +310,27 @@ model DeploymentStackProperties extends DeploymentStacksError {
   @example("01234567-89ab-cdef-0000-000000000000")
   correlationId?: string;
 
+  /**
+   * The validation level of the deployment stack
+   */
+  @added(Versions.v2025_07_01)
+  validationLevel?: ValidationLevel;
+}
```
anthony-c-martin: Just adding a comment here to make sure we track the work to add `externalInputs` to this API

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: Update - it is changeable, but requires an entry in the "backcompat" tsp. Feels not worth it to me

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: As we discussed in the chat, extensible resources are identified by ResourceType, configid, and identifiers.

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
dantedallag: Ok. So, we are going with deploymentStackId. I got to change that on the service.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: Discussed. I'm ok keeping it.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
kalbert312: Confirmed it's not necessary if it's only used a return result via TSP discussion channel.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
```diff
@@ -178,6 +195,11 @@ union DenyStatusMode {
    * No denyAssignments have been applied.
    */
   None: "none",
+
+  /**
+   * The denyAssignment status is unknown.
+   */
+  Unknown: "unknown",
```
kalbert312, line 208: @snarkywolverine Should this have `@added(Versions.v2025_07_01)`?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+ * Information about a single resource change predicted by What-If operation.
+ */
+@added(Versions.v2025_07_01)
+model DeploymentStacksWhatIfResourceChange {
+  ...ResourceChangeDelta;
+
+  /**
+   * The resource id of the resource being changed.
+   */
+  resourceId: string;
```
kalbert312: What is this for an extensible resource? Should it be nullable and only populated from ARM resources?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * The symbolic name of the resource being changed.
+   */
+  symbolicName: string;
+
+  /**
+   * The extensible resource identifiers.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Extensible resources will have type-specific properties to help with unique identification"
+  identifiers: Record<unknown>;
```
kalbert312: This should also be nullable because it only applies to extensible resources.

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * The predicted changes to the deployment stack deny status of the resource.
+   */
+  denyStatusChange: DeploymentStacksChangeBase<DenyStatusMode>;
+
+  /**
+   * The explanation about why the resource is unsupported by What-If.
+   */
+  unsupportedReason: string;
```
kalbert312: This should be nullable?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * The resource id of the Deployment responsible for this change.
+   */
+  deploymentId: string;
+
+  /**
+   * The symbolic name of the resource being changed.
+   */
+  symbolicName: string;
```
kalbert312: The symbolic name for a resource can be changed between templates. Is this represented in what-if as two changes (complete removal of resource "A" with ID "/a/b/c" and addition of resource "B" with ID "/a/b/c")? Should this also be nullable for support of ARM-only templates using array resources?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+   */
+  @OpenAPI.extension("x-ms-identifiers", #["path"])
+  children?: DeploymentStacksWhatIfPropertyChange[];
+}
+
+/**
+ * The error additional info
+ */
+@added(Versions.v2025_07_01)
+model DeploymentStacksDiagnostics {
```
kalbert312: This should probably be singular rather than plural.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+   * The symbolic name of the resource being changed.
+   */
+  symbolicName: string;
+
+  /**
+   * The extensible resource identifiers.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Extensible resources will have type-specific properties to help with unique identification"
+  identifiers: Record<unknown>;
+
```
kalbert312, line 126: We should probably also include a nullable `apiVersion` string for extensible resource API versions. We probably don't need it for ARM.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+   * Predicted changes to the deployment scope for the deployment stack.
+   */
+  deploymentScopeChange: DeploymentStacksChangeBase<string>;
+}
+
+/**
+ * Information about a single resource change predicted by What-If operation.
+ */
+@added(Versions.v2025_07_01)
+model DeploymentStacksWhatIfResourceChange {
```
kalbert312, line 114: Taking a look at the C# model, we should probably review nullability on all these fields.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * Predicted changes to the deployment stack deny settings.
+   */
+  denySettingsChange: DeploymentStacksChangeDelta<DenySettings>;
+
+  /**
+   * Predicted changes to the deployment scope for the deployment stack.
+   */
+  deploymentScopeChange: DeploymentStacksChangeBase<string>;
+}
```
kalbert312, line 108: Not sure where `deploymentExtensions` would go but it looks like here. We'd probably want to track the changes between extensions? Or at least list the current extensions?

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * The resource id of the Deployment responsible for this change.
+   */
+  deploymentId: string;
+
+  /**
+   * The symbolic name of the resource being changed.
+   */
+  symbolicName: string;
+
```
kalbert312: Needs `resourceType?: string;` for extensible resource types. It is required as part of the composite ID for an extensible resource.

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: Good catch on all the nullable types - looks like I didn't capture that properly from the deployments swagger.

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: Yes, fixed

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: Yes, fixed

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: fixed

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: Per offline conversation, deploymentExtensions should not be necessary (as it's not in deployments whatif either)

snarkywolverine: Per offline conversation, deploymentExtensions should not be necessary (as it's not in deployments whatif either)

specification/resources/DeploymentStacks.Management/models-whatIf.tsp
snarkywolverine: Should now be covered as part of ResourceReference spread-ing

specification/resources/DeploymentStacks.Management/back-compat.tsp
```diff
@@ -6,8 +6,8 @@ using Azure.ClientGenerator.Core;
 
 namespace Microsoft.Resources.DeploymentStacks;
 
-#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Backwards compat with old swagger"
-@@Azure.ClientGenerator.Core.Legacy.flattenProperty(DeploymentStack.properties);
+#suppress "deprecated" "Required for backwards compatibility with legacy swagger"
+@@Azure.ClientGenerator.Core.flattenProperty(DeploymentStack.properties);
```
kalbert312: This looks like it needs to be reverted: `Namespace Azure.ClientGenerator.Core doesn't have member flattenProperty`. TSP Compiler 1.5.0. Maybe an `npm install` in root directory fixes it locally.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
kalbert312: @snarkywolverine @dantedallag are we OK with nullability of all fields in this at this point? I see nullable enums in the C# model for stuff like `changeCertainty`, `changeType` etc.

specification/resources/DeploymentStacks.Management/back-compat.tsp
kalbert312: https://github.com/Azure/azure-rest-api-specs/actions/runs/18922371172/job/54021551556?pr=36828

snarkywolverine: Weird - when I run it on my machine, it doesn't compile with the old configuration - though it looks like I'm on ... 1.3 instead of 1.5

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: That's weird, though - because in the swagger, they're listed as required properties (line 3126). Do we need to get clarity from the C# team on this?

kalbert312: I meant our models in the service, not the produced SDK models.

dantedallag: Good callout. I don't think these should be nullable. Any objections to me cleaning that up?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: Ah, ok, I think I remember the initial intention of this question. On the service side, I was thinking that the before, after, and delta of the properties would be nested under a "PropertyChanges" object, so property level changes are structured similarly to the before/after changes.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
```diff
@@ -0,0 +1,3696 @@
...
+          "readOnly": true
+        },
+        "diagnostics": {
+          "type": "array",
+          "description": "List of resource diagnostics detected by What-If operation.",
+          "items": {
+            "$ref": "#/definitions/DeploymentStacksDiagnostic"
+          },
+          "readOnly": true,
+          "x-ms-identifiers": []
```
dantedallag, line 3388: Are we missing deploymentExtensions for the PUT path?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: That's fine from my side (discussed offline)

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: It's an Output-only property, but I think I see it on line 2400 -- or am I missing something?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: This would impact only the POST response, correct?

I see where you're coming from, but I want to be careful with naming.  property changes makes sense in the hierarchy, but then it feels a little weird to have the entire resource show up in before and after.  resourceChanges fits better with that component...but we've already claimed that to differentiate between per-resource changes and stack property changes.  Naming is hard.

Is this the structure you were thinking?

changes: [
..resourceChanges: [
....<stack specific change properties>
....propertyChanges: {
......before/after/delta

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: 2400 is in the stack properties, I'm talking about the what-if result properties.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
dantedallag: Yeah, I get your points and yes, that is the property structure I was thinking. In the end, it probably doesn't matter too much. I'm comfortable going with whatever you think is the better option.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
```diff
@@ -0,0 +1,379 @@
...
+ * DeploymentStack WhatIfResult Properties
+ */
+@added(Versions.v2025_07_01)
+model DeploymentStacksWhatIfResultProperties is DeploymentStacksError {
+  ...DeploymentStackSharedProperties;
+
+  /**
+   * The deployment stack id to use as the basis for comparison.
+   */
+  deploymentStackResourceId: string;
```
majastrz, line 55: Are we going to allow this to be modified after the what-if resource is created? If not, do we need to set a lifecycle annotation to indicate that it's immutable after creation?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
majastrz: Should we also review `changes` and `diagnostics` in the parent model for nullability?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
```diff
@@ -0,0 +1,379 @@
...
+   */
+  Modify: "modify",
+
+  /**
+   * 	The resource exists in the current state and the desired state and will be redeployed when the deployment is executed. The properties of the resource will not change.
+   */
+  NoChange: "noChange",
+
+  /**
+   * The resource is not supported by What-If.
```
majastrz, line 331: Not critical, but should we include examples of scenarios when a resource is not unsuported by what-if?

specification/resources/DeploymentStacks.Management/routes-whatIf.tsp
```diff
@@ -0,0 +1,71 @@
...
+  delete is ArmResourceDeleteWithoutOkAsync<
+    DeploymentStacksWhatIfResult,
+    BaseParameters = TBaseParams,
+    Parameters = ActionOnUnmanageDeleteParameters,
+    Response = ArmDeletedResponse | ArmDeletedNoContentResponse
+  >;
+
+  /** Returns property-level changes that will be made by the deployment if executed. */
+  @action("whatIf")
+  whatIf is ArmResourceActionSync<
```
majastrz: Are we 100% certain that we don't want this to be async?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
majastrz: > I ran out of time to go back and look at whatif -- I thought regular whatif had a full before/after as well as the per-property versions (though it does seem redundant...)

FWIW, the `before` and `after` are meant to be plugged into a text diff control and the `delta` stuff was meant to help PS/CLI rendering so we didn't have to implement a diff algorithm in c# and Python.

snarkywolverine: Per offline discussion, this is now nested in resourceConfigurationChanges

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: @majastrz - I'm not sure a block like this is worth the effort.  Conceptually, nothing should be persisted between whatif runs, so there shouldn't be a risk to the state of the whatif resource.  If the concern is a user returning to a whatif resource and having the results be dramatically different...well, that can also be achieved by deleting the whatif result and then re-running.  Do you think we gain anything with this block?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: @majastrz - this feels like something for docs, rather than examples.  We can add them later if customers find the existing examples confusing.

specification/resources/DeploymentStacks.Management/routes-whatIf.tsp
snarkywolverine: Fixed

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
snarkywolverine: @kalbert312 says there's not much value including it in what-if results, but it may be a candidate for including in the diff.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
snarkywolverine: I assume "parent model" means "our C# service code" and something @dantedallag should verify?  Or do you mean we should sanity check them here compared to our expectations?

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models-whatIf.tsp
majastrz: If the service is going to allow modifications to this, then it's fine I think.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/stable/2025-07-01/deploymentStacks.json
kalbert312: Talked in teams, there isn't much value in adding this right now. The focus is the resources diff. If customers want an extensions diff, that could be added later, but the utility of it seems very low.

kalbert312:
Breaking change cross version seems to be flagging query parameter ordering change, but that doesn't make sense: https://github.com/Azure/azure-rest-api-specs/actions/runs/19337176675

sandipsh:
@kalbert312, please address swagger modelValidation errors.

specification/resources/DeploymentStacks.Management/models.tsp
```diff
@@ -286,6 +332,27 @@ model DeploymentStackProperties extends DeploymentStacksError {
...
+
+/**
+ * Deployment stack properties.
+ */
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "Back compat. TSP doc instructs this for 'allOf' for OpenAPI"
+model DeploymentStackProperties is DeploymentStacksError {
```
kazrael2119: is this expected? in your old code, here uses `extends`
https://github.com/Azure/azure-rest-api-specs/pull/36828/files#diff-f57266bf57193a48380b244027a1f064b6189072debb7a644ec3f52521282111L217

kalbert312: We are OK with this, because it results in the same model. It's just represented differently in OpenAPI (inherited via "allOf" versus the "error" property directly on the properties model). We prefer to have the property directly on the model and not use inheritance here.

kalbert312:
> @kalbert312, please address swagger modelValidation errors.

We believe that the remaining model validation errors are incorrectly reported by the check. For instance, it is complaining about mismatched types in the examples, but the corresponding TSP is typed as `unknown`, which is supposed to represent `any`.

cc: @snarkywolverine, @dantedallag

kazrael2119:
Approved for js,

btw, go sdk validation failed, please ask Go sdk owner for help.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
```diff
@@ -53,6 +53,19 @@ union DeploymentStacksDeleteDetachEnum {
   Detach: "detach",
 }
 
+/**
+ * Specifies an action for resources that do not support deletion.
+ */
+union DeploymentStacksResourcesWithoutDeleteSupportEnum {
```
tadelesh, line 60: Missing versioning decorator.

specification/resources/DeploymentStacks.Management/models.tsp
```diff
@@ -211,10 +241,9 @@ model ActionOnUnmanageDeleteParameters {
 }
 
 /**
- * Deployment stack properties.
+ * Deployment stack shared properties.
  */
-#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "Back compat. TSP doc instructs this for 'allOf' for OpenAPI"
-model DeploymentStackProperties extends DeploymentStacksError {
+model DeploymentStackSharedProperties {
```
tadelesh: If it is not a real model, you should use alias instead of named model.

specification/resources/DeploymentStacks.Management/models.tsp
tadelesh: Such change will cause SDK breaking changes.

kalbert312: We're OK with this.

kalbert312: We don't want inheritance involved here.

specification/resources/resource-manager/Microsoft.Resources/deploymentStacks/models.tsp
kalbert312: Added

specification/resources/DeploymentStacks.Management/models.tsp
kalbert312: Done

specification/resources/DeploymentStacks.Management/models.tsp
kalbert312: This is also a relatively new SDK package. We split this management plane off of the large Azure.ResourceManager.Resources package.

anthony-c-martin (approved):

PR #36833, opened 2025-08-20
Webpubsub To New TSP Folder Structure

kashifkhan:
Move to v2 folder structure


catalinaperalta (commented):
You'll probably need to change the tspconfig.yaml path so that you get the swagger outputted in the right location. Which is under the same tsp project directory. Look at this example for a v2 structure: https://github.com/mikeharder/azure-rest-api-specs/tree/main/specification/widget

catalinaperalta (commented):
I'm not seeing the generated swagger under your webpubsub directory. EDIT: I saw that you're new directory name is the same as where the original handwritten swaggers are checked in so makes sense there's no diff for the swagger.

kashifkhan:
> I'm not seeing the generated swagger under your webpubsub directory

discussed with @catalinaperalta offline: Since the original data folder was called `WebPubSub`, its using the same dir where the other hand-written swaggers were checked. As the tsp files moved there, there isnt a new diff

lmazuel (approved):

PR #36872, opened 2025-08-22
Bump OAV

scbedd:
This PR also touches a couple specs. Those changes WILL NOT be checked in. It's just a brief smoke-test to ensure I don't break anything by this package update. I will revert dc5e725 before merging.


mikeharder (approved):
glad to see this removes many deps from package-lock.json

scbedd:
> glad to see this removes many deps from package-lock.json

Yeah!  I was pumped to check that in it'll buy us a bunch of time with `oav`.

Also, #36887 confirms that the warning I saw on modelvalidation/semanticvalidation was present on old code. Going to revert the spec change and get this merged.

PR #36876, opened 2025-08-22
netapp, migration to TypeSpec

weidongxu-microsoft:
4 items that may have SDK break:

- request body of Snapshots_Update, service confirmed that the payload has no property (it is a legacy issue) https://github.com/Azure/azure-rest-api-specs/pull/36876#discussion_r2309247832 <-- you may want to scope out this API, as it does nothing
- `RelationshipStatus` having different enum values in different model, service confirmed they are all different enums https://github.com/Azure/azure-rest-api-specs/pull/36876#discussion_r2312959462
- One List API was not specified as pageable, service want us to fix it here https://github.com/Azure/azure-rest-api-specs/pull/36876#discussion_r2309236881 -- I've changed it to pageable in TypeSpec (and it affects Swagger)
- `model-as-string` changed to `true` in stable. Service wanted it to be extensible enum. But we are not able to easily model this change via versioning lib. Hence, I just changed it for both stable and preview.

---

## Analyse BreakingChanges

diff https://github.com/Azure/azure-rest-api-specs/pull/36876#issuecomment-3257111210
(because there is lots of casing difference in schema, the "Swagger BreakingChange" CI is not accurate.

### Casing change on schema (e.g. `backupPolicy` to `BackupPolicy`)
It should not affect API or SDK

### Order of `body` parameter in request parameters
It should not affect API or SDK

### added `Azure-AsyncOperation` or `Location` header
They are not avoidable, as TypeSpec require such header in LRO operation

### `title` and `example` removed

TypeSpec not able to specify these.

### Some schema structure changed

E.g.
image

TypeSpec would write it with `scalar`, the OpenAPI should be equivalent.
```
  administrators?: UserName[];

@minLength(1)
@maxLength(255)
scalar UserName extends string;
```

### In Paged result, `value` be "required", and `nextLink` is added

This is expected. TypeSpec need to write it this way, and it is the ARM standard. It should be compatible.

### Structure of ref on tags changed

The OpenAPI should be equivalent.

```
        "tags": {
          "$ref": "#/definitions/resourceTags"
        }
```
to 
```
        "tags": {
          "type": "object",
          "additionalProperties": {
            "type": "string"
          }
        }
```

### Schema `volumeGroupDetails` changed to use parent `ProxyResource`

This is expected. `volumeGroupDetails` should be a proxy resource, and need to be written this way. The OpenAPI should be equivalent.

### Some unnamed schema changed to named

The OpenAPI should be equivalent. The name used is compatible in SDK.

E.g.
image

### Some unused schema is removed

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/netapp/NetApp.Management/models.tsp
```diff
@@ -0,0 +1,3959 @@
...
+
+/**
+ * Volumes associated with snapshot policy
+ */
+model SnapshotPolicyVolumeList {
+  /**
+   * List of volumes
+   */
+  value?: Volume[];
+}
```
weidongxu-microsoft: This List model is not written as `Azure.Core.Page<Volume>`, because in current Swagger the API is not written as pageable API (which is likely wrong).

specification/netapp/NetApp.Management/Snapshot.tsp
```diff
@@ -0,0 +1,103 @@
...
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "For backward compatibility"
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  location: string;
+}
+
+/**
+ * Snapshot of a Volume
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "For backward compatibility"
+model SnapshotPatch {}
```
weidongxu-microsoft, line 89: In Swagger, the payload to PATCH is written as
```
    "snapshotPatch": {
      "description": "Snapshot patch",
      "type": "object",
      "properties": {}
    },
```

In the example JSON, the payload is `{}`.

The intention (in Swagger) appears to be an empty model. <-- will verify with service on this

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
-          "type": "array",
-          "readOnly": true,
-          "description": "List of mount targets",
-          "items": {
-            "$ref": "#/definitions/mountTargetProperties"
-          },
-          "x-ms-identifiers": [
-            "fileSystemId"
-          ]
-        },
```
weidongxu-microsoft, line 7199: `mountTarget` is not used. Here `mountTargets` ref to `mountTargetProperties`.

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
-          "description": "URL to get the next set of results."
-        }
-      }
-    },
-    "netAppAccount": {
-      "description": "NetApp account resource",
-      "type": "object",
-      "required": [
-        "location"
-      ],
```
weidongxu-microsoft, line 6231: `location` as "required" written this way does not have effect, as the property is in parent.

TypeSpec should have `location` as "required", as in parent
https://github.com/Azure/typespec-azure/blob/a191aa2/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L48

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
-          "description": "Utilized throughput of pool in MiB/s",
-          "example": 164.221,
-          "readOnly": true
-        },
-        "customThroughputMibps": {
-          "title": "Maximum throughput in MiB/s that can be achieved by this pool and this will be accepted as input only for manual qosType pool with Flexible service level",
-          "type": "number",
-          "x-nullable": true,
-          "example": 128
-        },
```
weidongxu-microsoft, line 6765: Is this float or integer?

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
-          "description": "Status of the mirror relationship",
-          "enum": [
-            "Idle",
-            "Transferring"
-          ],
-          "x-ms-enum": {
-            "name": "RelationshipStatus",
-            "modelAsString": true
-          }
-        },
```
weidongxu-microsoft, line 7871: The other 2 `RelationshipStatus` has 4 states -- Idel, Transferring, Failed, Unknown

Is this correct?

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
-          "description": "A list of snapshot policies",
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/snapshotPolicy"
-          }
-        }
-      }
+      ]
     },
-    "snapshotPolicyDetails": {
```
weidongxu-microsoft, line 8350: not used

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5473,5132 +7496,4967 @@
...
+          "type": "string",
+          "description": "Resource location",
+          "x-ms-mutability": [
+            "read",
+            "create"
           ]
         }
-      }
-    },
-    "subnetId": {
```
weidongxu-microsoft, line 9467: not used

specification/netapp/NetApp.Management/Snapshot.tsp
weidongxu-microsoft: example JSON uses empty JSON
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/examples/Snapshots_Update.json#L10

audunn: This is a bit of techical dept really. Snapshot is a Proxy resource that had no updatetable properties. property. However initially It was mistakenly modeled as a tracked resource with tags and it had to support patch to update the tags.  This was corrected back in 2019 and tags removed, but the PATCH was keept in for backwards compatability.

specification/netapp/NetApp.Management/models.tsp
audunn: We can probably add paging here

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
audunn: integer

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
audunn: Those are different Enums. Was an unfortunate error at the time that is was not separate enums with different names but this has not been changed to avoid breaking changes.

- Volume replication relationshipStatus status has two values `Idle` and `Transferring`
- latestBackupStatus Get latest satus of restore on backup: `RestoreStatus.relationshipStatus`  `Idle`, `Transferring`,  `Failed`, `Unknown`
- latestRestoreStatus Get latest status of restore on snapshot): restoreStatus `RestoreStatus.relationshipStatus`  `Idle`, `Transferring`,  `Failed`, `Unknown`

Should probably be in two or three different enums.

specification/netapp/NetApp.Management/models.tsp
```diff
@@ -0,0 +1,4012 @@
+import "@typespec/rest";
```
audunn, line 1: This file is getting rather big, would it make sense to break this up into smaller units for example by resource type?

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
weidongxu-microsoft: Got it. I will first put this one as a separate enum e.g. named `VolumeReplicationRelationshipStatus` (let me know if you have any suggestion on naming, or whether we want to just make all 3 different enum at this TypeSpec)

specification/netapp/NetApp.Management/Snapshot.tsp
weidongxu-microsoft: Got it. So, the `properties` should be any empty model (as there is no updatable properties) -- and actually, no one should call this `Snapshots_Update` anyway, at least not now when there is nothing updatable.

Another solution in SDK (if we already had breaking changes in SDK), is that it can remove this operation.

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
weidongxu-microsoft: Got it. We may not fix this in this convert, as it be a break (unless we want to take it).

Saw this one because it misses the "format" in Swagger
```
        "customThroughputMibps": {
          "type": "number",
          "format": "float",
          "x-nullable": true
        }
```

When I check it again, it may actually be a `float`?

```
        "totalThroughputMibps": {
          "type": "number",
          "description": "Total throughput of pool in MiB/s",
          "example": 164.221,
          "readOnly": true
        },
        "utilizedThroughputMibps": {
          "type": "number",
          "description": "Utilized throughput of pool in MiB/s",
          "example": 164.221,
          "readOnly": true
        },
        "customThroughputMibps": {
          "title": "Maximum throughput in MiB/s that can be achieved by this pool and this will be accepted as input only for manual qosType pool with Flexible service level",
          "type": "number",
          "x-nullable": true,
          "example": 128
        },
```

specification/netapp/NetApp.Management/models.tsp
weidongxu-microsoft: Got it. Let me check with dev, whether we want to take the breaks (that does not affect REST, but could cause breaking change in SDK) in this conversion.

specification/netapp/NetApp.Management/NetAppAccount.tsp
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Contains data from encryption.keyVaultProperties as well as information about which private endpoint is used by each encryption sibling set. Response from this endpoint can be modified and used as request body for POST request.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "For backward compatibility"
+  #suppress "@azure-tools/typespec-azure-core/invalid-final-state" "For backward compatibility"
+  @operationId("Accounts_GetChangeKeyVaultInformation")
+  @action("getKeyVaultStatus")
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  getChangeKeyVaultInformation is ArmResourceActionAsync<
```
weidongxu-microsoft, line 577: The Swagger `"final-state-via": "azure-async-operation"` may be wrong.
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json#L1033-L1080

The response status code is 200+202, where 202 typically have `Location` header. But the `final-state-via` is `Azure-AsyncOperation` header.

Currently, I use `LroHeaders = ArmCombinedLroHeaders<FinalResult = GetKeyVaultStatusResponse>` to make 202 return both `Location` header and `Azure-AsyncOperation` header. But this may not be the service behavior.

specification/netapp/NetApp.Management/models.tsp
weidongxu-microsoft: I think the basic `@resource` resources is already in separated files.

However, I think there are models that related to these resources, but was put in this `models.tsp`. E.g. `BackupsList`, `BaskcupProperties`, `BackupPatch`, `BackupPatchProperties` etc. may be moved to `backup.tsp`.
Can we do this? @pshao25

weidongxu-microsoft:
diff on 2025-06-01
https://www.diffchecker.com/yYPeXFd8/

diff on 2025-07-01-preview
https://www.diffchecker.com/oQAlRY8g/

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -5537,5068 +7495,5071 @@
...
-          "example": [
-            "123",
-            "224"
+          "x-ms-identifiers": [
+            "name"
           ]
         }
       }
     },
-    "cloudError": {
```
weidongxu-microsoft, line 8888: not used

specification/netapp/NetApp.Management/models.tsp
weidongxu-microsoft: Moved most of them to resource tsp files.

specification/netapp/NetApp.Management/Backup.tsp
```diff
@@ -44,7 +44,8 @@ interface Backups {
   update is Azure.ResourceManager.Legacy.CustomPatchAsync<
     Backup,
     PatchModel = BackupPatch,
-    Response = ArmResponse<Backup> | (ArmAcceptedLroResponse & {
+    Response = ArmResponse<Backup> | (ArmAcceptedLroResponse<LroHeaders = ArmLroLocationHeader<FinalResult = NetAppAccount> &
```
JiaqiZhang-Dev: Should the `FinalResult` be `Backup`?

weidongxu-microsoft: Thanks. Fixed.

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
```diff
@@ -28,565 +30,599 @@
...
             "in": "body",
-            "description": "Network sibling set to query.",
+            "description": "The request body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/QueryNetworkSiblingSetRequest"
-            },
-            "x-ms-client-flatten": true
+              "$ref": "#/definitions/UpdateNetworkSiblingSetRequest"
+            }
```
msyyc, line 535: Compared with original swagger https://github.com/Azure/azure-rest-api-specs/blob/fb90eb1bec64c6e8ad3e288a64c84cc18742a394/specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json#L572-L590, the `"x-ms-client-flatten": true` is missing. But python is OK for the breaking.

specification/netapp/NetApp.Management/Backup.tsp
```diff
@@ -0,0 +1,284 @@
...
+      $filter?: string;
+    },
+    Response = ArmResponse<BackupsList>
+  >;
+
+  /**
+   * Restore the specified files from the specified backup to the active filesystem
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("BackupsUnderBackupVault_RestoreFiles")
```
pshao25: We can't handle this by clientLocation?

specification/netapp/NetApp.Management/BackupPolicy.tsp
```diff
@@ -0,0 +1,220 @@
...
+   * Resource tags
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "For backward compatibility"
+  tags?: Record<string>;
+
+  /**
+   * Backup policy Properties
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-private-usage" "For backward compatibility"
+  @Azure.ResourceManager.Private.conditionalClientFlatten
```
pshao25: conditionalClientFlatten is useless

specification/netapp/NetApp.Management/CapacityPool.tsp
```diff
@@ -0,0 +1,288 @@
...
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("Pools_Get")
+  get is ArmResourceRead<CapacityPool>;
+
+  /**
+   * Create or Update a capacity pool
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "non-standard operations"
+  @operationId("Pools_CreateOrUpdate")
```
pshao25: The operation ids here cannot be handled by clientLocation?

weidongxu-microsoft: Let me try remove these decorators. They are generated by converter.

weidongxu-microsoft: Deleted operationId and conditionalClientFlatten

weidongxu-microsoft:
@v-jiaodi Let me know the status of validation on NetApp for TypeScript (in your epic it is not checked, so I assume there is still something need to be done). Is it possible to be validated before holiday? Thanks.

specification/netapp/resource-manager/Microsoft.NetApp/stable/2025-06-01/netapp.json
audunn: Appoligies missed this, customThroughputMibps is in fact int

weidongxu-microsoft: Emm, changing float to integer would likely be a break to SDK as well.

Is it int32 or int64?

audunn: int32

weidongxu-microsoft: Thanks. Updated to `int32`.

specification/netapp/NetApp.Management/CapacityPool.tsp
```diff
@@ -0,0 +1,276 @@
...
+  @visibility(Lifecycle.Read)
+  @maxLength(36)
+  @minLength(36)
+  @pattern("^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$")
+  poolId?: string;
+
+  /**
+   * Provisioned size of the pool (in bytes). Allowed values are in 1TiB chunks (value must be multiple of 1099511627776).
+   */
+  size?: int64 = 4398046511104;
```
audunn: Can we keep this requred, to avoid breaking change? Same for usageThreshold in volumeProperties.

weidongxu-microsoft: Sure. I've reverted this change.

The net effect would now be this
https://github.com/Azure/azure-rest-api-specs/pull/36876/files/e7223d229980d476a9765a3ab904963b0ef4eeba..7c7a14f78cdcc2e3bb09fe8c4f45a37f578619e3

pshao25 (approved):

PR #36893, opened 2025-08-23
[api-doc-preview] Move from "shared" to "workflows"

mikeharder:
- Code is specific to a single pipeline, so "workflows" seems a better fit than "shared"
- No functional changes


danieljurek (approved):
LGTM. Let me know when you merge and sync to the other repo. ~Two pipelines need to be updated:~ 

* ~https://dev.azure.com/azure-sdk/internal/_build?definitionId=7767~
* ~https://dev.azure.com/azure-sdk/internal/_build?definitionId=7841~

Pipelines don't need to be updated... the file wasn't moved.

weshaggard (approved):

PR #36925, opened 2025-08-25
Updated the repo owner of SDK PR in step of creating branch and PR

raych1:
We cannot use the `SdkRepoOwner` variable because it will be overwritten to `azure-sdk` when SDK branch is specified.

Although we have the code to parse `SdkRepoUrl` variable to get the repo owner, current pipelines set them to 'azure' SDK repos. 

I'd rather to hardcode its value to 'azure' in both `push-code` step and `create-PR` step.

Here's a test PR created.
https://github.com/Azure/azure-sdk-for-go/pull/25144

### DIFF
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -278,7 +278,7 @@ stages:
             BaseRepoBranch: $(SdkPullRequestSourceBranch)
             BaseRepoOwner: azure-sdk
             CommitMsg: $(GeneratedSDKInformation)
-            TargetRepoOwner: $(SdkRepoOwner)
+            TargetRepoOwner: azure
             TargetRepoName: $(SdkRepoName)
             WorkingDirectory: $(SdkRepoDirectory)
             ScriptDirectory: $(SdkRepoDirectory)/eng/common/scripts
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -278,7 +278,7 @@ stages:
             BaseRepoBranch: $(SdkPullRequestSourceBranch)
             BaseRepoOwner: azure-sdk
             CommitMsg: $(GeneratedSDKInformation)
-            TargetRepoOwner: $(SdkRepoOwner)
```
weshaggard, line 281: Can't we eliminate the setting of this variable completely now?

raych1: `git-push-changes.yml` has the default value set to 'azure'. However, `submit-pullrequest.ps1` doesn't have a default value for 'repoOwner'.

weshaggard: I was meaning `SdkRepoOwner` can we stop setting that variable in the earlier steps.

raych1: We will still need to introduce another variable used in the `sparse-checkout` step.

weshaggard (approved):

PR #36990, opened 2025-08-27
Exclude CODEOWNERS from matched files

### LINKED ISSUE #35271: [Protected Files] Exclude .github/CODEOWNERS

chidozieononiwu:
Resolves https://github.com/Azure/azure-rest-api-specs/issues/35271

### DIFF
--- a/.github/workflows/protected-files.yaml
+++ b/.github/workflows/protected-files.yaml
@@ -37,6 +37,11 @@ jobs:
 
           $matchedFiles = @($changedFiles | Where-Object { $changedFile = $_; $protectedFiles | Where-Object { $changedFile -like $_ } })
 
+          $matchedFiles = $matchedFiles | Where-Object {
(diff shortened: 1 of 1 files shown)


.github/workflows/protected-files.yaml
```diff
@@ -37,6 +37,11 @@ jobs:
 
           $matchedFiles = @($changedFiles | Where-Object { $changedFile = $_; $protectedFiles | Where-Object { $changedFile -like $_ } })
 
+          $matchedFiles = $matchedFiles | Where-Object {
```
weshaggard: Can you make this more generic with an exclude list and just use the same `-like` operation as we are doing above? We shouldn't need to replace the `\` as this is always coming back the same way from git.

mikeharder (approved):
LGTM

weshaggard (approved):

PR #36993, opened 2025-08-27
[AI Projects] dotnet naming changes

trangevi:
Naming changes for dotnet, based on apiview feedback


specification/ai/Azure.AI.Projects/client.tsp
```diff
@@ -37,9 +37,44 @@ using Azure.ClientGenerator.Core;
...
-@@clientName(Azure.AI.Projects.Deployment, "AssetDeployment", "csharp");
-@@clientName(Azure.AI.Projects.Index, "SearchIndex", "csharp");
+@@clientName(Azure.AI.Projects.BlobReference,
+  "AIProjectBlobReference",
+  "csharp"
+);
+@@clientName(Azure.AI.Projects.Connection,
+  "AIProjectConnectionProperties",
+  "csharp"
+);
```
dargilco: I suggest `AIProjectConnetion` (drop `Properties`) to match AIProjectDeployment, AIProjectIndex

specification/ai/Azure.AI.Projects/client.tsp
```diff
@@ -52,10 +87,19 @@ using Azure.ClientGenerator.Core;
...
+@@clientName(Azure.AI.Projects.Datasets,
+  "AIProjectDatasetsOperations",
+  "csharp"
+);
+@@clientName(Azure.AI.Projects.Deployments,
+  "AIProjectDeploymentsOperations",
+  "csharp"
+);
+@@clientName(Azure.AI.Projects.Indexes, "AIProjectIndexesOperations", "csharp");
```
dargilco, line 103: I don't think we need to prefix `AIProjects` to class property names. You don't usually do an Azure-wide search name for just the property name. I can see why you would do an Azure-wide search for a class name though, that's why KC is asking for unique prefix. But not for property names.

trrwilson (approved):

dargilco (approved):

PR #37012, opened 2025-08-28
[readme.js] Add more tests for duplicate tags

mikeharder:
- Tests related to #37003 (which we decided to not fix)


.github/shared/test/readme.test.js
```diff
@@ -26,8 +26,6 @@ describe("readme", () => {
     expect(readme.path).toBe(resolve("/specs/foo/readme.md"));
   });
 
-  // TODO: Test that path is resolved against backpointer
```
mikeharder, line 29: test already added, forgot to remove comment

PR #37013, opened 2025-08-28
add max concurrency changes to updaterun

frantran:
add max concurrency changes to updaterun and strategy

### DIFF
--- a/specification/containerservice/Fleet.Management/update/common.tsp
+++ b/specification/containerservice/Fleet.Management/update/common.tsp
@@ -35,6 +35,18 @@ model UpdateStage {
   @doc("The time in seconds to wait at the end of this stage before starting the next one. Defaults to 0 seconds if unspecified.")
   afterStageWaitInSeconds?: int32;
 
+  @doc("""
+  The max number of upgrades that can run concurrently across all groups in this stage.
+  Accepts either:
+      • A fixed count, e.g. "3"
+      • A percentage, e.g. "25%" (range 1–100). Percentage is of the total clusters under all groups at stage start, fractional results rounded up. A minimum of 1 upgrade is enforced.
+  Example: 5 --> at most 5 concurrent upgrades at once from all groups in this stage.
+  Example: 100% --> If a stage has 5 clusters total, the maxConcurrency for upgrades is 5.
+  """)
+  @pattern("^(?:([1-9]\\d*)|([1-9]\\d?|100)%)$")
+  @added(Versions.v2025_10_01_preview)
+  maxConcurrency?: string;
+
   @doc("A list of Gates that will be created before this Stage is executed.")
   @added(Versions.v2025_04_01_preview)
   @extension("x-ms-identifiers", #[])
@@ -56,6 +68,22 @@ model UpdateGroup {
   @minLength(1)
   @maxLength(50)
   name: string;
+
+  @doc("""
+  The max number of potential upgrades that can run concurrently in this specific group.
+  Acts as a ceiling (and not a quota) for the number of concurrent upgrades within the group you want to tolerate at a time.
+  Actual concurrency may be lower depending on stage-level limits or individual member conditions.
+  Accepts either:
+      • A fixed count, e.g. "3"
+      • A percentage, e.g. "25%" (range 1–100). Percentage is of the group size at stage start. Fractional results are rounded up. A minimum of 1 upgrade is enforced.
+  Examples:
+      • "3" --> up to 3 members from this group upgrade at once.
+      • "100%" --> “all at once”, up to all members for this group upgrade at the same time.
+      • "25%" --> up to 25% of the members in the group will be upgrade at the same time. 
+  """)
+  @pattern("^(?:([1-9]\\d*)|([1-9]\\d?|100)%)$")
+  @added(Versions.v2025_10_01_preview)
+  maxConcurrency?: string;
 
   @doc("A list of Gates that will be created before this Group is executed.")
   @added(Versions.v2025_04_01_preview)
--- a/specification/containerservice/Fleet.Management/update/common.tsp
+++ b/specification/containerservice/Fleet.Management/update/common.tsp
@@ -5,8 +5,6 @@ using TypeSpec.Versioning;
 
 namespace Microsoft.ContainerService;
 
-const StageMaxConcurrencySystemLimit = "50";
-const StageMaxConcurrencyDefault = "10";
 const GroupMaxConcurrencyDefault = "1";
 
 @doc("""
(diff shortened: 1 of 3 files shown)


specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -57,6 +69,22 @@ model UpdateGroup {
   @maxLength(50)
   name: string;
 
+  @doc("""
+    The max number of potential upgrades that can run concurrently in this specific group.
```
dvadas: I like the thorough comment describing everything here. As a minor point, it would be good to align the group-level and the stage-level doc more closely, e.g.:
  * Remove "potential" here
  * Add the next two lines to the stage-level comment as well
  * Format the examples the same way
  * Anything else I missed.

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -35,6 +35,18 @@ model UpdateStage {
   @doc("The time in seconds to wait at the end of this stage before starting the next one. Defaults to 0 seconds if unspecified.")
   afterStageWaitInSeconds?: int32;
 
+  @doc("""
+    The max number of upgrades that can run concurrently across all groups in this stage.
+    Accepts either:
+        • A fixed count, e.g. "3"
+        • A percentage, e.g. "25%" (range 1–100). Percentage is of the total clusters under all groups at stage start, fractional results rounded up. A minimum of 1 upgrade is enforced.
```
dvadas: nit: "stage start" suggests that the number of clusters can change while an update run is running. I would say:
> Percentage is of the total number of clusters under all groups in the stage, fractional ...

Same for the group-level comment:

> Percentage is of the number of clusters in the group.

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -57,6 +69,22 @@ model UpdateGroup {
...
+    The max number of potential upgrades that can run concurrently in this specific group.
+    Acts as a ceiling (and not a quota) for the number of concurrent upgrades within the group you want to tolerate at a time.
+    Actual concurrency may be lower depending on stage-level limits or individual member conditions.
+    Accepts either:
+        • A fixed count, e.g. "3"
+        • A percentage, e.g. "25%" (range 1–100). Percentage is of the group size at stage start. Fractional results are rounded up. A minimum of 1 upgrade is enforced.
+    Examples:
+        • "3" --> up to 3 members from this group upgrade at once.
+        • "100%" --> “all at once”, up to all members for this group upgrade at the same time.
+        • "25%" --> up to 25% of the members in the group will be upgrade at the same time.
```
dvadas: > will be upgrade**d**

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -57,6 +69,22 @@ model UpdateGroup {
   @maxLength(50)
   name: string;
 
+  @doc("""
```
dvadas, line 80: It would be good to add what happens if you don't set a value at all, i.e. the default is 1. I think that's safe to do for the group-level since we don't intend to change that (as opposed to stage-level where we want a bit more freedom so leaving it unspecified is good).

specification/containerservice/Fleet.Management/update/run.tsp
```diff
@@ -232,6 +232,35 @@ model UpdateRunStatus {
...
+model StageConcurrencyStatus {
+  @visibility(Lifecycle.Read)
+  @doc("The max number of upgrades that can run concurrently across all groups in this stage.")
+  maxConcurrency?: int32;
+
+  @visibility(Lifecycle.Read)
+  @doc("""
+    Original requested maxConcurrency value for this stage. Either a fixed count (e.g., "3") or a percentage (e.g., "25%").
+    """)
+  requestedMaxConcurrency?: string;
```
dvadas: Isn't this already recorded in the strategy? I don't think it needs to be repeated here.

specification/containerservice/Fleet.Management/update/run.tsp
```diff
@@ -232,6 +232,35 @@ model UpdateRunStatus {
   nodeImageSelection?: NodeImageSelectionStatus;
 }
 
+@doc("The status of the stage-level concurrency.")
+@added(Versions.v2025_10_01_preview)
+model StageConcurrencyStatus {
+  @visibility(Lifecycle.Read)
+  @doc("The max number of upgrades that can run concurrently across all groups in this stage.")
+  maxConcurrency?: int32;
```
dvadas: I think it'd be clearer to the customer what's happening if this was non-optional. That is, if the customer doesn't specify it in the strategy, we populate it with our hardcoded default. For example, if a customer has two groups and sets maxConcurrency to 50 for both of them, they might wonder why they're not getting 100 concurrent updates. Having a value of 50 (or 8 or 10) shown here would help explain.

serbrech: we cannot add a required field to an API. my understanding is that the intent is to populate this with the resolved value always?

specification/containerservice/Fleet.Management/update/common.tsp
frantran: revised

specification/containerservice/Fleet.Management/update/common.tsp
frantran: revised

specification/containerservice/Fleet.Management/update/common.tsp
frantran: add defaults

specification/containerservice/Fleet.Management/update/run.tsp
frantran: removed field

specification/containerservice/Fleet.Management/update/run.tsp
frantran: the field on the status will always be the resolved/default value (from the RP side)

specification/containerservice/Fleet.Management/update/run.tsp
```diff
@@ -232,6 +232,26 @@ model UpdateRunStatus {
   nodeImageSelection?: NodeImageSelectionStatus;
 }
 
+@doc("The status of the stage-level concurrency.")
+@added(Versions.v2025_10_01_preview)
+model StageConcurrencyStatus {
```
dvadas: Should these wrapper models be removed now that there's only a single field inside? Or put another way, are there more concurrency settings we can think of that could be added in future?

frantran: removed ConcurrencyStatus and minimized the datamodel

dvadas (approved):

specification/containerservice/Fleet.Management/update/common.tsp
```diff
@@ -57,6 +77,25 @@ model UpdateGroup {
...
+    Accepts either:
+        • A fixed count, e.g. "3"
+        • A percentage, e.g. "25%" (range 1–100). Percentage is of the number of clusters in the group. 
+          Fractional results are rounded up. A minimum of 1 upgrade is enforced.
+    Examples:
+        • "3" --> up to 3 members from this group upgrade at once.
+        • "100%" --> “all at once”, up to all members for this group upgrade at the same time.
+        • "25%" --> up to 25% of the members in the group will be upgraded at the same time.
+    """)
+  @pattern("^(?:([1-9]\\d*)|([1-9]\\d?|100)%)$")
```
serbrech, line 95: is that going to reject 10, 20, 30 % because it does not allow 0, other than in 100?

serbrech: ah, \d* allows 0 as a 2nd char.

serbrech (approved):

PR #37016, opened 2025-08-28
Simplify all specs to omit `@useDependency` for Azure Core and Arm librairies

timotheeguerin:
Public repo part of https://github.com/Azure/typespec-azure/issues/3074

Since last release this is optional and the goal is to remove the versions in Azure.Core and Azure.Resourcemanager libraries to simplify the specs as done here.


razvanbadea-msft:
Added ARM Signed Off as changes are not related to api changes but tsp files meta update

markcowl (approved):

PR #37116, opened 2025-09-03
Folder structure validation

qiaozha:
## Folder Structure Validation

This PR introduces folder structure validation to enforce Azure REST API specifications organization standards.

### Summary

Adds a new independent `folder-structure` validation check that runs alongside TypeSpec validation to ensure specifications follow the v2 folder structure guidelines. This helps maintain consistency across the repository and prevents introducing legacy v1 folder structures in v2-compliant organizations.

#### 1. New Folder Structure Validation Tool
- **New PowerShell script**: `eng/scripts/Folder-Structure-Validation.ps1` - Validates `specification/{org}` folders for any changed files
- **New CLI tool**: `npx folder-structure` - Validates folder structure at the org level
- **Comprehensive validation logic**: 
  - Detects v1 vs v2 folder structures
  - Prevents v1 structure regression in v2-compliant orgs
  - Allows existing v1 services in mixed orgs but blocks new v1 services
  - Provides detailed reporting of v1-like and v2-aligned folders

#### 2. GitHub Actions Workflow Updates
- Renamed `typespec-validation.yaml` → `spec-validation.yaml` with two independent jobs:
  - **TypeSpec Validation**: Runs `npx tsv` on TypeSpec folders
  - **Folder Structure Validation**: Runs `npx folder-structure` on org folders
- **TypeSpec Validation - All** workflow unchanged (only validates TypeSpec compilation)

#### 3. Tool Reorganization
- Renamed `eng/tools/typespec-validation` → `eng/tools/spec-validation`
- Package name: `@azure-tools/typespec-validation` → `@azure-tools/spec-validation`
- Two CLI commands:
  - `npx tsv` - TypeSpec validation (compilation, linting, SDK config)
  - `npx folder-structure` - Folder structure validation

#### 4. Suppression Updates
- **Tool name standardization**: Changed all suppressions from `tool: TypeSpecValidation` to `tool: SpecValidation`
  - Both `tsv` and `folder-structure` tools now use the same tool name for suppressions
- **New suppression**: Added `FolderStructure` rule suppression for `specification/compute` (legacy structure)

#### 5. Enhanced TypeSpec Folder Detection
- Updated `Get-TypeSpec-Folders.ps1` to return:
  - `$orgFolders` - for folder-structure validation (all `specification/{org}` folders)
  - `$typespecFolders` - for TypeSpec validation (folders with `tspconfig.yaml`)
  - `$checkingAllSpecs` - flag for validation mode

#### Regular Spec Validation (PR checks)
- Runs `npx folder-structure` on all `specification/{org}` folders with changes
- Runs `npx tsv` on TypeSpec projects (folders with `tspconfig.yaml`)

#### TypeSpec Validation - All (scheduled/manual)
- Only runs `npx tsv` on TypeSpec projects
- Skips folder-structure validation

### Folder Structure Rules

**V2 Structure** (enforced for new services):
```
specification/{org}/data-plane/{ServiceName}/...
specification/{org}/resource-manager/{RPNamespace}/{ServiceName}/...
```

**V1 Structure** (legacy, blocked for new services):
```
specification/{org}/{ServiceFolder}/...
specification/{org}/resource-manager/{RPNamespace}/... // (without {ServiceName} layer)
```

**Validation Logic**:
- Pure v2 orgs: Must not introduce any v1 structure
- Mixed orgs: Can modify existing v1 services, but cannot add new v1 services
- V1 orgs: No restrictions (for now)

### Testing
- Comprehensive test suite in `eng/tools/spec-validation/test/folder-structure.test.ts`
- Tests cover v1/v2 detection, regression prevention, and various edge cases

### Documentation
- Updated `documentation/ci-fix.md` with folder structure validation information
- Updated GitHub workflow summaries and check names

---

**Breaking Changes**: None - existing specs continue to work with suppressions where needed

**Migration Path**: Services using v1 structure can continue operating but should plan migration to v2 structure


mikeharder:
@qiaozha: My understanding is "folder structure v2" is only a concept we care about enforcing for TypeSpec specs.  If a swagger-only spec wants to use the same folder and readme structure, they can, but this technically isn't "folder structure v2".  And we're not interested in adding more checks or tooling for swagger-only specs.

Given this, can the check be added to existing TypeSpecValidation, instead of needing a whole new tool and workflow?

qiaozha:
> @qiaozha: My understanding is "folder structure v2" is only a concept we care about enforcing for TypeSpec specs. If a swagger-only spec wants to use the same folder and readme structure, they can, but this technically isn't "folder structure v2". And we're not interested in adding more checks or tooling for swagger-only specs.
> 
> Given this, can the check be added to existing TypeSpecValidation, instead of needing a whole new tool and workflow?

@mikeharder Folder structure v2 is not just about TypeSpec, it also has impacts on the services with pure swaggers as well. Just like the service version uniform guideline are not bound by whether that service is using typespec or swagger, they need to deliver service that are versioning uniformly regardless. And folder structure v2 is the fundamental support of the version uniform guideline. @JeffreyRichter feel free to correct me if I am wrong. 

You can take a look at this https://github.com/Azure/azure-rest-api-specs/tree/main/specification/orbital/resource-manager/Microsoft.Orbital example, it doesn't have mgmt plane TypeSpec, but this service is trying to isolate as two different services in their mgmt plane with this weird way.  After the service team sort this out by adopting to our folder structure v2 like to have two serviceName folders under Microsoft.Orbital namespace folder, when they are future PRs trying to add a new stable or preview version under Microsoft.Orbital directly, we are definitely not going to allow this.

Another consideration is, I am not sure if you remember that I pointed it out to you before, the currently way where we put the folder structure validation inside the TypeSpec validation is not clear enough for service team to focus on which failure to fix first when they see a lot of CI failures. But I think folder structure is one of the fundamental checks that they should fix first if there's any violations.

mikeharder:
> Folder structure v2 is not just about TypeSpec, it also has impacts on the services with pure swaggers as well.

Azure SDK EngSys has been instructed to focus new feature work only on TypeSpec specs.  We need to continue supporting swagger-only specs (e.g. fixing bugs in swagger-based tools), but all new tools and checks (like enforcing folder structure v2) should only apply to TypeSpec specs.

> You can take a look at this https://github.com/Azure/azure-rest-api-specs/tree/main/specification/orbital/resource-manager/Microsoft.Orbital example, it doesn't have mgmt plane TypeSpec, but this service is trying to isolate as two different services in their mgmt plane with this weird way. After the service team sort this out by adopting to our folder structure v2 like to have two serviceName folders under Microsoft.Orbital namespace folder, when they are future PRs trying to add a new stable or preview version under Microsoft.Orbital directly, we are definitely not going to allow this.

Instead of adding a new tool and check for a small number of swagger-only specs converting to something like v2 (but without TypeSpec), I think the effort would better be spent, accelerating the conversion of these specs to TypeSpec.

> Another consideration is, I am not sure if you remember that I pointed it out to you before, the currently way where we put the folder structure validation inside the TypeSpec validation is not clear enough for service team to focus on which failure to fix first when they see a lot of CI failures. But I think folder structure is one of the fundamental checks that they should fix first if there's any violations.

Rather than adding a new tool and check, the better fix would be to make errors from TypeSpecValidation easier to read (e.g. add a markdown summary), either for all types of errors, or just folder structure.

CC: @kurtzeborn

qiaozha:
@mikeharder 
I don't think swagger only services is a small number. From my previous statistics, it has more than 110 services. The overall communicate will be very huge if we let them do whatever they want after we helped them migrate to folder structure v2. 
Compared to the communication effort, I don't think add a new check will take that much effort and even less from EngSys team as I am taking the coding part, right? 

If we add this check within TypeSpec validation, this will be confusing to those 110+ service teams if they don't have any typespec at all. And we will probably get a lot of questions from them, right?

Even if we don't invest in swagger, but I don't think some minimal fundamental tooling support to make the SDK engineer life easier is not something we should do. 

Also, do you think such kind of questions might be better to discuss in the email thread instead of PR?

mikeharder:
> I don't think add a new check will take that much effort

Adding a new blocking check on all PRs is a big cost and responsibility for everyone supporting the specs repo.  Checks crash, have bugs requiring quick fixes to unblock PRs, need suppressions, etc.  It is not to be taken lightly.

> If we add this check within TypeSpec validation, this will be confusing to those 110+ service teams if they don't have any typespec at all. And we will probably get a lot of questions from them, right?
>
> Even if we don't invest in swagger, but I don't think some minimal fundamental tooling support to make the SDK engineer life easier is not something we should do.

**If** the check should apply to swagger-only specs, I agree 100% it should not live in TypeSpec validation.  But I still don't think it should apply to swagger-only specs, for the reasons I mentioned above.

Another reason, if a spec is swagger-only, it should be getting extra scrutiny from ARM and data-plane reviewers.  I think it should be apparent in the file diff and readme diff, if a PR is trying to add a v1 folder to a spec that's been converted to v2, right?  If you disagree, we can create a test PR to see how it looks.

Even if there are still 110 swagger-only specs, they're rapidly being converted them to typespec, so this number should decrease over time, meaning this impacts fewer and fewer PRs over time.

Finally, what's the cost if a PR does slip through, adding a v1 folder to a v2 spec?  Once someone notices (during the TSP conversion at the latest), we convert it back to v2?

> Also, do you think such kind of questions might be better to discuss in the email thread instead of PR?

I think it's best to centralize as much code and communication in github (issues, PRs, wiki).  azure-rest-api-specs-pr can be used for less-public discussions.

qiaozha:
close it as we are not working towards this direction.

PR #37022, opened 2025-08-29
adopt new tsp-client config

tadelesh:
# SDK configuration pull request

## Purpose of this PR

- [ ] Make changes to the SDK configuration only when there are no modifications to the API specification, eliminating the need for an ARM or Stewardship Board API review.

## Due diligence checklist

To merge this PR, you **must** go through the following checklist and confirm you understood
and followed the instructions by checking all the boxes:

- [ ] I confirm this PR is modifying only SDK configurations, and not API related specifications.
- [ ] I have reviewed and used the respective `tspconfig.yaml` templates:
  - [ARM tspconfig template](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml)
  - [Data plane tspconfig template](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml)

## Getting help

- First, carefully read through this PR description, from top to bottom. Fill out the `Purpose of this PR` and `Due diligence checklist`.
- If you don't have permissions to remove or add labels to the PR, request `write access` per [aka.ms/azsdk/access#request-access-to-rest-api-or-sdk-repositories](https://aka.ms/azsdk/access#request-access-to-rest-api-or-sdk-repositories)
- To understand what you must do next to merge this PR, see the `Next Steps to Merge` comment. It will appear within few minutes of submitting this PR and will continue to be up-to-date with current PR state.
- For guidance on fixing this PR CI check failures, see the hyperlinks provided in given failure and https://aka.ms/ci-fix.
- If the PR CI checks appear to be stuck in `queued` state, please add a comment with contents `/azp run`.
  This should result in a new comment denoting a `PR validation pipeline` has started and the checks should be updated after few minutes.
- If the help provided by the previous points is not enough, post to https://aka.ms/azsdk/support/specreview-channel and link to this PR.

PR #37224, opened 2025-09-08
Adding in a prompt so TypeSpec+Copilot users can add Go emitter configurations

### LINKED ISSUE #11977: Add some Copilot instructions to help users add Go to their tspconfig.yaml
We can apply these to specific files so we don't crowd the context for other tasks (ie, applyTo tspconfig.yaml) and we need to keep in mind that data plane vs management plane have different settings.

It would be nice to not have two sources of truth here, so if we can leverage the template we recommend people use with our `azsdk-cli tsp init` command that'd be best.

richardpark-msft:
Most uses of Go, in data plane, are going to involve adding Go emitter configuration to an existing project. We'll cover `tsp init` generation in the azsdk-cli.

This file can also be used for any other tspconfig.yaml instructions as well if we want to try to cut down on github-instructions.md.

Using this prompt (via /tspconfig_add_missing_languages) look like this:

image

Fixes https://github.com/Azure/azure-sdk-tools/issues/11977


richardpark-msft:
@tadelesh, I didn't add in management plane instructions because I was unclear if I needed to - mgmt plane always seems supported on day 1 (vs data plane, which is always added after the fact).

.github/instructions/tspconfig.instructions.md
```diff
@@ -0,0 +1,43 @@
...
+```yaml
+"@azure-tools/typespec-go":
+    # Take the segments after `Azure`, in the `namespace` property. Each segment 
+    # is a file path. The _last_ file path should be prefixed with `az`.
+	module: "github.com/Azure/azure-sdk-for-go/sdk/ai/azdocumentintelligence"
+    # take the segments and turn them into a file path. Do not include the final segment this time.
+	service-dir: "{output-dir}/sdk/ai"
+    # take the final segment, prefix it with `az`, and make it all lowercase. 
+    # NOTE, this property builds on the one before it, so you don't have to add it again.
+	emitter-output-dir: "{service-dir}/azdocumentintelligence"
```
richardpark-msft: @raych1, @tadelesh - I think I'm using the right format now, but can you check? I don't want to check something in that's already outdated.

raych1: That's correct.

.github/instructions/tspconfig.instructions.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+## Adding Go support to an existing tspconfig.yaml
+
+Adding in Go support is simple if you're using an existing configuration, and just requires some translation of the C# entry.
+
+Here's an example C# entry:
+
+```yaml
+"@azure-tools/typespec-csharp":
```
tadelesh: Go has different service dir and emitter output dir rule from other languages. Do we need to mention the .NET config which may cause mis-leading? I suggest it's better to show the template [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml#L42) and doc [here](https://azure.github.io/typespec-azure/docs/emitters/clients/typespec-go/reference/emitter/#emitter-options). Note that the template is not update to date before the [PR](https://github.com/Azure/azure-rest-api-specs/pull/37022) I'm working on merge.

tadelesh:
> @tadelesh, I didn't add in management plane instructions because I was unclear if I needed to - mgmt plane always seems supported on day 1 (vs data plane, which is always added after the fact)

I prefer also add mgmt. part here. New mgmt. service also needs to know how to add Go config. Mgmt. config template is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L35).

.github/instructions/tspconfig.instructions.md
richardpark-msft: I can delete more of what's in there (I really only need the namespace attribute). What this enables is the automatic derivation of the Go config by Copilot, using an existing .NET config in the same file.

So just drop into any random tspconfig.yaml, and now copilot can create the Go config if it already exists.

Let me see if I can meld together the two approaches though, because it would be good for people to see the templates.

richardpark-msft: I was able to meld it :) The new one uses your template as the source and it's pretty reliable. My main concern was it having issues actually including _all_ the attributes, but it seems to work pretty well.

richardpark-msft:
> > @tadelesh, I didn't add in management plane instructions because I was unclear if I needed to - mgmt plane always seems supported on day 1 (vs data plane, which is always added after the fact)
> 
> I prefer also add mgmt. part here. New mgmt. service also needs to know how to add Go config. Mgmt. config template is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L35).

Ha, as soon as I added in management plane support the "magic" stopped happening. I'll need to spend some more time to refine this.

richardpark-msft:
> > > @tadelesh, I didn't add in management plane instructions because I was unclear if I needed to - mgmt plane always seems supported on day 1 (vs data plane, which is always added after the fact)
> > 
> > 
> > I prefer also add mgmt. part here. New mgmt. service also needs to know how to add Go config. Mgmt. config template is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L35).
> 
> Ha, as soon as I added in management plane support the "magic" stopped happening. I'll need to spend some more time to refine this.

Okay, I re-fixed it. I converted it to a prompt instead because it cuts down on the overall instructions file for the repo and lets you control some other aspects, like the model.

.github/prompts/tspconfig_add_missing_languages.prompt.md
```diff
@@ -0,0 +1,14 @@
+---
+mode: agent
+model: Claude Sonnet 4
```
maririos, line 3: ohh I didn't know we could do this!!!!!!!

.github/prompts/tspconfig_add_missing_languages.prompt.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+Add a language that is missing (for instance, adding Go, Python, Javascript C# or Rust support), using these rules:
```
maririos: Are we sure we want to add Rust? @ronniegeraghty 

```suggestion
Add a language that is missing (for instance, adding Go, Python, Javascript, C#, or Rust support), using these rules:
```

tadelesh (approved):

.github/prompts/tspconfig_add_missing_languages.prompt.md
```diff
@@ -0,0 +1,14 @@
...
+
+- New emitter configurations should be added as a child of the "options:" section, after any other emitters and their options.
```
maririos, line 8: should it start saying where the emitter configurations need to go? tspconfig.yaml? or it is not needed

tadelesh:
> > > > @tadelesh, I didn't add in management plane instructions because I was unclear if I needed to - mgmt plane always seems supported on day 1 (vs data plane, which is always added after the fact)
> > > 
> > > 
> > > I prefer also add mgmt. part here. New mgmt. service also needs to know how to add Go config. Mgmt. config template is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml#L35).
> > 
> > 
> > Ha, as soon as I added in management plane support the "magic" stopped happening. I'll need to spend some more time to refine this.
> 
> Okay, I re-fixed it. I converted it to a prompt instead because it cuts down on the overall instructions file for the repo and lets you control some other aspects, like the model.

The prompt looks elegant.

maririos (commented):
Could you add description on how this would be use by the user? Is this something that we want people to be aware of that will require documentation somewhere?

.github/prompts/tspconfig_add_missing_languages.prompt.md
richardpark-msft: Yeah, it is probably the best feature and it actually works! :)

.github/prompts/tspconfig_add_missing_languages.prompt.md
richardpark-msft: It does a good job of seeing the similarity between the two files, so I think we're okay.

.github/prompts/tspconfig_add_missing_languages.prompt.md
richardpark-msft: Let me remove it for now - it's not yet in the tspconfig.yaml template so it won't show up anyways.

richardpark-msft:
> Could you add description on how this would be use by the user? Is this something that we want people to be aware of that will require documentation somewhere?

I added it to the PR, I think the best we can do (maybe?) is add this to our eng.ms docs?

maririos (approved):

mikeharder:
First, should this file be added to `.github/prompts` or `.github/instructions`?  This repo moved from the former to the latter in this PR, and I don't think we should revert back, unless there is consensus.

https://github.com/Azure/azure-rest-api-specs/pull/34880/files

Once this is decided, how "protected" do we want this folder to be?  Allow anyone with write access to the specs repo, or limit it to members of the "bypass" group (much smaller group with higher permissions)?

mikeharder (changes requested):
see comments

maririos:
> First, should this file be added to `.github/prompts` or `.github/instructions`? This repo moved from the former to the latter in this PR, and I don't think we should revert back, unless there is consensus.
> 
> https://github.com/Azure/azure-rest-api-specs/pull/34880/files
> 
> Once this is decided, how "protected" do we want this folder to be? Allow anyone with write access to the specs repo, or limit it to members of the "bypass" group (much smaller group with higher permissions)?

Thank you Mike! I was not aware of that :)
As for protection, I want to limit the folder to our team, so the "bypass" group. I have been adding myself to all the parths related to ai on this repo in codeowners so we can have control over what goes in.

~~@richardpark-msft We should then move it according to the GH Copilot documentation in the PR you linked.~~

mikeharder:
> > First, should this file be added to `.github/prompts` or `.github/instructions`? This repo moved from the former to the latter in this PR, and I don't think we should revert back, unless there is consensus.
> > https://github.com/Azure/azure-rest-api-specs/pull/34880/files
> > Once this is decided, how "protected" do we want this folder to be? Allow anyone with write access to the specs repo, or limit it to members of the "bypass" group (much smaller group with higher permissions)?
> 
> Thank you Mike! I was not aware of that :) As for protection, I want to limit the folder to our team, so the "bypass" group. I have been adding myself to all the parths related to ai on this repo in codeowners so we can have control over what goes in.
> 
> @richardpark-msft We should then move it according to the GH Copilot documentation in the PR you linked.

If we want `.github/instructions` to be protected to the bypass group, that's already enabled.  If you want any users added to the bypass group, just let me know.

richardpark-msft:
@mikeharder, @maririos: according to the documents where I have it is correct - prompts go in .github/prompts

> Workspace prompt files: Are only available within the workspace and are stored in the .github/prompts folder of the workspace.

From here: https://code.visualstudio.com/docs/copilot/customization/prompt-files

There's also immediate feedback, since you can see the prompt when it's in the correct spot in VSCode's Copilot Chat.

I think we should unblock this PR, looks like everything is correct for what we're doing here.

mikeharder (approved):

.github/prompts/tspconfig_add_missing_languages.prompt.md
```diff
@@ -0,0 +1,14 @@
...
+mode: agent
+model: Claude Sonnet 4
+---
+
+Add a language that is missing (for instance, adding Go, Python, Javascript, or C# support), using these rules:
+
+- New emitter configurations should be added as a child of the "options:" section, after any other emitters and their options.
+- Do NOT search the files in the repository for examples, as these can be out of date.
+- Determine which tspconfig.yaml template to use:
+  - If the currently open tspconfig.yaml contains @azure-tools/typespec-azure-rulesets/resource-manager, then use this template: [tspconfig template](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml)
```
mikeharder, line 11:
```suggestion
  - If the currently open tspconfig.yaml contains @azure-tools/typespec-azure-rulesets/resource-manager, then use this template: [tspconfig template](https://[raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/widget/resource-manager/Microsoft.Widget/Widget/tspconfig.yaml](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/widget/resource-manager/Microsoft.Widget/Widget/tspconfig.yaml))
```

Recommend using the folder structure v2 sample `widget` rather than v1 sample `contosowidgetmanager`.

.github/prompts/tspconfig_add_missing_languages.prompt.md
```diff
@@ -0,0 +1,14 @@
...
+  - If the currently open tspconfig.yaml contains @azure-tools/typespec-azure-rulesets/data-plane, then use this template: [tspconfig template](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml)
```
mikeharder, line 12:
```suggestion
  - If the currently open tspconfig.yaml contains @azure-tools/typespec-azure-rulesets/data-plane, then use this template: [tspconfig template](https://[raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/refs/heads/main/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml))
```

mikeharder (changes requested):
change to v2 sample

.github/prompts/tspconfig_add_missing_languages.prompt.md
richardpark-msft: I've don't have an opinion on this - the template I'm using was recommended by @tadelesh earlier in the PR. I'd like to make sure we haven't accidentally wandered into a situation where we have two separate examples that are supposed to be used.

@tadelesh, @mikeharder - is there any particular reason we have two examples for the same scenario? Are they equivalent?

mikeharder: Specs in this repo are in transition from "folder structure v1" to "folder structure v2".  More info here: https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure

Currently, the majority of existing specs use v1, but our team is aggressively converting all specs in the repo (at least TSP specs) to v2.  And all new specs should be using v2 (though this may not be enforced via PR checks yet).

The original TSP sample was `contosowidgetmanager`, and it uses "folder structure v1", because v2 didn't exist at the time.  For clarity, when v2 was introduced, we added a new sample `widget` alongside the existing sample.

Once all (or most) specs are converted to v2, we will remove the v1 sample `contosowidgetmanager`.  Until then, it's nice to have samples for both v1 and v2.

richardpark-msft:
> change to v2 sample

So looking at the v2 sample some more, it seems incomplete - there's a TODO right in it.

>     # TODO: Does anything need this set, if it's not used in output-file?

Also, there are some other bits that seem incorrect. 

For instance:

(under typespec-csharp)
>     namespace: "Azure.WidgetAnalytics"

This doesn't have a "service area", unless WidgetAnalytics is, for instance, sibling to something like 'ai', or 'messaging'. But then what is this package?

(under   "@azure-tools/typespec-python")

>     namespace: "azure.widget.analytics"

This makes it seem like the above should be Azure.Widget.Analytics as well.

(under "@azure-tools/typespec-go"):

>    service-dir: "sdk/widget-analytics"
>    package-dir: "azmanager"

So in this one there's a new package (azmanager), that's not mentioned in the others.

These are just a few things that jump out at me. The Contoso one seems more conformant to our Azure patterns and feels like a better fit, for now.

richardpark-msft:
I should note, we can always update the prompt later when the v2 template is finished/fixed.

mikeharder:
> I should note, we can always update the prompt later when the v2 template is finished/fixed.

I'm fine adding the prompt with the v1 template for now.

However, we need to improve the v2 sample, and update the prompt to use it, ASAP.  https://github.com/Azure/azure-rest-api-specs/issues/37322

@richardpark-msft: Who can I work with (someone from each language?), to ensure the sample is 100% compliant with every best practice of naming, configuration, etc?

mikeharder:
> So looking at the v2 sample some more, it seems incomplete - there's a TODO right in it.

The sample may have bugs/deficiencies, but I wouldn't call it "incomplete".  The sample was reviewed by many stakeholders when it was first added, and many PRs have been merged since then to improve the sample.

https://github.com/Azure/azure-rest-api-specs/commits/main/specification/widget

If there are more areas we can improve, let's start opening PRs and I can merge them.

The TODO left in the tspconfig.yaml is a minor issue, but you raise a fair point it should be resolved, and I will take this myself, since I added the TODO.

.github/prompts/tspconfig_add_missing_languages.prompt.md
mikeharder: Agreement is to use v1 sample for now, but switch to v2 sample, once every language owner agrees the sample is "perfect" from their perspective in terms of naming, config, etc.

mikeharder (approved):

PR #35403, opened 2025-06-23
Migrate Storage from swagger to TypeSpec

welovej:
This PR migrates your latest version (identified by the tag in your readme.md) of swagger to TypeSpec. We already tried our best to make sure the TypeSpec represents same as previous swagger. Since we lack the business knowledge, please validate this PR again to make sure it's functional equivalent as before. The local validation step is at [Getting started | TypeSpec Azure](https://azure.github.io/typespec-azure/docs/migrate-swagger/01-get-started/#review-and-adjust-the-typespec)

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind. Here is the list you need to pay attention to:
- Operation `StorageAccounts_GetCustomerInitiatedMigration` cannot set constraint on `migrationName`. See detail at https://github.com/Azure/typespec-azure/issues/2886
- We changed path "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/{BlobServicesName}" to "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default" because the parameter `BlobServicesName` can only have the value `default` and its child [path](https://github.com/Azure/azure-rest-api-specs/blob/1a1984eab477a3527e10ce1c9e9ee5fd806bec62/specification/storage/resource-manager/Microsoft.Storage/stable/2024-01-01/blob.json#L217) call this segment `default`. 
- We also changed "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}/immutabilityPolicies/{immutabilityPolicyName}" to "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/microsoft.Storage/storageAccounts/{accountName}/blobServices/default/containers/{containerName}/immutabilityPolicies/default" for the same reason.
[...]

PR #37257, opened 2025-09-09
Add Smart as an Account Access Tier under 2025-08-01 Api Version

kah279:
# Choose a PR Template

### DIFF
--- a/specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
+++ b/specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
@@ -1,0 +1,1579 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "version": "2025-01-01",
+    "title": "StorageManagementClient",
+    "description": "The Azure Storage Management API."
+  },
+  "host": "management.azure.com",
+  "schemes": [
+    "https"
+  ],
+  "consumes": [
+    "application/json"
+  ],
+  "produces": [
+    "application/json"
+  ],
+  "paths": {
+    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/blobServices": {
+      "get": {
+        "tags": [
+          "BlobService"
+        ],
+        "operationId": "BlobServices_List",
+        "x-ms-examples": {
+          "ListBlobServices": {
+            "$ref": "./examples/BlobServicesList.json"
+          }
+        },
+        "description": "List blob services of storage account. It returns a collection of one object named default.",
+        "parameters": [
+          {
+            "$ref": "./storage.json#/parameters/ResourceGroupName"
+          },
+          {
+            "$ref": "./storage.json#/parameters/StorageAccountName"
+          },
+          {
+            "$ref": "../../../../../common-types/resource-management/v1/types.json#/parameters/ApiVersionParameter"
+          },
+          {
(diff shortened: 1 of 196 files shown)


blueww (changes requested):
Please separate the PR into at least 3 commits to help review:

1. Copy the swagger from last API version folder without any change to the new API version folder
2. Change the API version 
3. Add new API changes with 1 or  more commits. 

See sample: https://github.com/Azure/azure-rest-api-specs/pull/28461/commits

specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
```diff
@@ -0,0 +1,1579 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "version": "2025-01-01",
```
blueww: It looks the API version is not updated (in all files, not only this one).
And many things are missed , like readme not updated. 

If this PR change is only for private, please raise it in private repo, instead of public repo.

specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
```diff
@@ -0,0 +1,1579 @@
+{
+  "swagger": "2.0",
```
blueww, line 2: After merge the PR, the Rest API change in new API version will be public, and will be in rest API doc,  ARM template,  ... (besides PSH/CLI/SDK)

To avoid customer issue, before merge this PR, please make sure:

1. The new API version is available in all regions (include usgov, mooncake ...)
2. All new API changes are available in all regions .
 - If some API change will not be available in some regions by design, please point it out, and share the error message when user use the API in not supported regions. The error message should be clear and actionable.

specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
kah279: apologies, fixed this and split up the commits for easier review.

specification/storage/resource-manager/Microsoft.Storage/stable/2025-08-01/blob.json
kah279: The new API version is not yet deployed everywhere but is starting to deploy broadly. We can leave this PR pending until deployment completes. Wanted to get this merged with enough time to have PSH/CLI/SDK support at feature announcement.

sandipsh:
please fix prettier check

blueww:
The 2025-06-01 swagger change is still not in. 
https://github.com/Azure/azure-rest-api-specs/pull/37597

This PR API version 2025-08-01 should be based on the latest API version before it, which should be 2025-06-01, so you should wait for 2025-06-01 PR merge, then add this PR based on 2025-06-01.

Besides that, this API change is very possible need to be added with typespec at that time.

josefree:
A PR converting the SRP service to TypeSpec has been prepped by the Azure SDK team as part of the drive to move all existing Azure services to use TypeSpec.  The PR for this is
https://github.com/Azure/azure-rest-api-specs/pull/35403
.  The team owning this service will need to validate the PR, address any highlighted linter failures, make any service-specific changes to the PR, have a review with ARM, and complete the PR before any additional work can be done on the service.  Information regarding how to do this finalization and who to contact for assistance are at https://azure.github.io/typespec-azure/docs/migrate-swagger/01-get-started/#review-and-adjust-the-typespec.  Going forward, any change to the service spec will need to be performed with TypeSpec.

PR #37261, opened 2025-09-09
Updating managedBy definition in resources.json in resource manager

tanvikadari97:
managedBy is an immutable property when creating a new resource which is being used to add wrong info by the customer due to incomplete documentation. Making changes to specific that it is a immutable and optional property so that customers do not assign wrong value to this attribute.

# Choose a PR Template


specification/resources/resource-manager/Microsoft.Resources/stable/2025-03-01/resources.json
```diff
@@ -1,4 +1,4 @@
~{
~  "swagger": "2.0",
```
sandipsh, line 2555: >managedBy [](http://example.com/codeflow?start=9&length=9)

please specify mutability / readonly for the property.

tanvikadari97: Did you mean to add a [Read Only] parameter to the description?

tanvikadari97: added the mutability property to the property

mentat9:
@tanvikadari97 - Please submit the dataplane changes in a separate PR. This repo handles either a CP or DP PR, not a single PR for both. ARM reviewers don't review or approve dataplane changes.

PR #37270, opened 2025-09-09
Introducing Secure Wrap and Unwrap functionality.

vasanthrajams:
# Introducing Secure Wrap and Unwrap functionality.


specification/keyvault/Security.KeyVault.Keys/models.tsp
```diff
@@ -139,6 +139,38 @@ union JsonWebKeyCurveName {
   P256_K: "P-256K",
 }
 
+/**
+ * An algorithm used for key wrapping and unwrapping.
+ */
+@added(KeyVault.Versions.v2026_01_01_preview)
+union JsonWebKeyWrapAlgorithm {
```
heaths, line 146: Didn't you already have an enum with `A128KW` and friends?

specification/keyvault/Security.KeyVault.Keys/client.tsp
```diff
@@ -293,6 +293,83 @@ op unwrapKey is KeyVaultOperation<
...
+     */
+    @path("key-name")
+    @clientName("name", "go")
+    keyName: string;
+
+    /**
+     * The version of the key.
+     */
+    @path("key-version")
+    @clientName("version", "go")
```
heaths, line 320: Also for `rust`.

specification/keyvault/Security.KeyVault.Keys/client.tsp
```diff
@@ -293,6 +293,83 @@ op unwrapKey is KeyVaultOperation<
...
```
heaths, line 360: Same for `rust`

specification/keyvault/Security.KeyVault.Keys/client.tsp
heaths: Fix in `routes.tsp` as well.

heaths (changes requested):
Also go through all the validation errors.

heaths (approved):
You'll probably need to suppress the validation errors or they're just going to keep coming back.

heaths (approved):

heaths (approved):

specification/keyvault/data-plane/Microsoft.KeyVault/preview/2026-01-01-preview/secrets.json
```diff
@@ -0,0 +1,1109 @@
...
+                  "value": "application/x-pkcs12",
+                  "description": "The PKCS#12 file format."
+                },
+                {
+                  "name": "PEM",
+                  "value": "application/x-pem-file",
+                  "description": "The PEM file format."
+                }
+              ]
+            }
```
mikekistler, line 391: Why not simply use the standard `accpet` header for this?

mikekistler: Ah ... I see ... this is only the content-type of one part of the response. The response itself is application/json.

But I'd like to recommend the following change around line 940:

```diff
        "contentType": {
          "type": "string",
-          "description": "The content type of the secret."
+          "description": "The content type of the secret value."
        },
```

At least I _think_ this is referring to the `value` field. If so I think this clarifies that, and if not please change the description to clearly indicate what this is the content-type of.

mikekistler (approved):
Looks good. 👍 

Requested one minor doc change -- please address that.

heaths:
Suppressions are the same as for past versions, so approving.

PR #37271, opened 2025-09-09
Add a new step to mark SDK generation status as failed in release plan

praveenkuttappan:
Add a step to run when SDK generation fails which will mark generation status as failed in release plan.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -340,7 +340,7 @@ stages:
                 $prStatus = "Failed to generate SDK."
                 Write-Host "##vso[task.setvariable variable=SdkPrStatus]$prStatus"
               }
-            condition: not(endsWith(variables['SdkRepoName'], '-pr')))
+            condition: not(endsWith(variables['SdkRepoName'], '-pr'))
             displayName: "Set pull request URL variables"
 
           - task: AzureCLI@2
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -322,17 +322,10 @@ stages:
               -PRBody "$(GeneratedSDKInformation)  $(ReleasePlanInfo)"
               -OpenAsDraft $true
 
-        - pwsh: |
-            Write-Host "Release Info: $(ReleasePlanInfo)"
-            $prStatus = 'Failed to generate SDK.'
-            Write-Host "##vso[task.setvariable variable=SdkPrStatus]$prStatus"
-          displayName: 'Set PR generation status'
-          condition: Failed()
-
         - ${{ if ne(parameters.ReleasePlanWorkItemId, 0) }}:
           - task: AzureCLI@2
             displayName: Link pull request to release plan
-            condition: and(succeededOrFailed(), eq(variables['HasChanges'], 'true'), not(endsWith(variables['SdkRepoName'], '-pr')))
+            condition: and(succeeded(), eq(variables['HasChanges'], 'true'), not(endsWith(variables['SdkRepoName'], '-pr')))
             continueOnError: true
             inputs:
               azureSubscription: opensource-api-connection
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -342,7 +335,22 @@ stages:
...
+
+          - task: AzureCLI@2
+            displayName: Update SDK generation status as failed
+            condition: and(Failed(), not(endsWith(variables['SdkRepoName'], '-pr')))
+            continueOnError: true
+            inputs:
+              azureSubscription: opensource-api-connection
+              scriptType: pscore
+              scriptLocation: scriptPath
+              scriptPath: $(SpecRepoDirectory)/eng/scripts/Update-PullRequest-In-ReleasePlan.ps1
```
raych1, line 354: This step invokes same script as last step. Why can't we move this logic to within the script itself?

praveenkuttappan: One is invoked in case of success and PR is generated and one is only if SDK generation failed

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -342,7 +335,22 @@ stages:
               arguments: >
                 -ReleasePlanWorkItemId ${{ parameters.ReleasePlanWorkItemId }}
                 -PullRequestUrl "https://github.com/Azure/$(SdkRepoName)/pull/$(Submitted.PullRequest.Number)"
-                -Status "$(SdkPrStatus)"
+                -Status "draft"
+                -SdkRepoName "$(SdkRepoName)"
+
+          - task: AzureCLI@2
+            displayName: Update SDK generation status as failed
```
raych1: The step name looks confusing to me. Should that be release plan related either?

raych1 (approved):

raych1 (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -321,18 +323,29 @@ stages:
...
+              Write-Host "##vso[task.setvariable variable=PullRequestUrl]$prUrl"
+              $prStatus = "Failed to generate SDK."
+              Write-Host "##vso[task.setvariable variable=SdkPrStatus]$prStatus"
+            condition: failed()
+            displayName: "Set pull request generation failed status variable"
+
           - task: AzureCLI@2
             displayName: Link pull request to release plan
-            condition: and(succeededOrFailed(), eq(variables['HasChanges'], 'true'), not(endsWith(variables['SdkRepoName'], '-pr')))
+            condition: not(endsWith(variables['SdkRepoName'], '-pr'))
```
weshaggard, line 348: Why did this condition change? It should still ony trigger if it haschanges right?

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -321,18 +323,29 @@ stages:
...
+              $prUrl = "https://github.com/Azure/$(SdkRepoName)/pull/$(Submitted.PullRequest.Number)"
+              Write-Host "Pull request created: $prUrl"
+              Write-Host "##vso[task.setvariable variable=PullRequestUrl]$prUrl"
+              $prStatus = "draft"
+              Write-Host "##vso[task.setvariable variable=SdkPrStatus]$prStatus"
+            condition: and(succeeded(), eq(variables['HasChanges'], 'true'), ne(variables['Build.Reason'], 'PullRequest'), not(endsWith(variables['SdkRepoName'], '-pr')))
+            displayName: "Set pull request URL variable"
+
+          - pwsh: |
+              $prUrl = ""
```
weshaggard, line 339: Is this the primary change? Because if the step fails there is no PR?

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
weshaggard: I guess you always want to set the status even if there is no PR?

PR #37340, opened 2025-09-11
Application Gateway L4 Client IP Preservation Support

akshimittal1310:
## Purpose of this PR
Added Client IP Preservation options in template for application Gateway.
Following are the 2 properties that will be added in Backend Settings, Probe and OnDemandHealthProbe for Application Gateway Resource.

enableL4ClientIpPreservation
enableProbeProxyProtocolHeader

Changes will be tied to NRP 187 API version, the changes have been rolled out already

Previous PR where the Properties were already approved in Private branch: -
https://github.com/Azure/azure-rest-api-specs-pr/pull/20948

# Choose a PR Template


specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
```diff
@@ -1,4 +1,4 @@
~{
~  "swagger": "2.0",
~  "info": {
~    "title": "NetworkManagementClient",
```
sandipsh, line 1179: >enableProbeProxyProtocolHeader [](http://example.com/codeflow?start=9&length=30)

is this going to be part of new api version? work-in-progress?

akshimittal1310: Yes, the NRP rollout is already done

https://msazure.visualstudio.com/One/_git/Networking-nrp/pullrequest/11590002

The RP manifest has rolled out as well

gary-x-li, line 1179: >enableProbeProxyProtocolHeader [](http://example.com/codeflow?start=9&length=30)

Consider changing it to enum so that you can add more future values instead of adding more boolean properties.

More context on the topic:
ARM recommends enums over booleans for future proof APIs. Standard guidance is: replace boolean/switch properties with a more meaningful enum whenever possible.

A boolean will forever have two valid values (true or false). A string enum type is always preferred. Also, properties should always provide better values just than True and False. For example two switches "isTypeA" and "isTypeB" should be replaced with one enum "type": [A, B, DefaultType]. Enums are always a more flexible and future proof option because they allow additional values to be added in the future in a non-breaking way, e.g. [Enabled, Disabled, Suspended, Deallocated]. 

Note: do NOT define a 'boolean enum' with two values 'True and False'. This might be easier to 'extend' in terms of types, but semantically its confusing, and no better than a boolean.

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
```diff
@@ -1176,6 +1176,10 @@
           "type": "boolean",
           "description": "Whether the host header should be picked from the backend http settings. Default value is false."
         },
+        "enableProbeProxyProtocolHeader": {
+          "type": "boolean",
+          "description": "Whether to send Proxy Protocol header along with the Health Probe over TCP or TLS protocol. Default value is false."
+        },
         "match": {
           "$ref": "#/definitions/ApplicationGatewayProbeHealthResponseMatch",
           "description": "Criterion for classifying a healthy probe response."
```
gary-x-li, line 1977: >enableL4ClientIpPreservation [](http://example.com/codeflow?start=9&length=28)

Same boolean vs enum comment as above.

Plus, the description for this property is very similar to the other one. It doesn't match with its name.

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
akshimittal1310: Thanks Gary for the detailed review and for highlighting this point. I do understand the ARM guidance around using enums over booleans, and agree with the rationale, enums definitely provide more flexibility and allow for future extensibility.

In this specific case, the feature is inherently binary and was designed to always remain that way. The RP changes with this property as a boolean have already been released and were reviewed/accepted by the ARM team earlier during private preview. Revisiting the design now and reworking it into an enum would have a large downstream impact on our RP, SDKs, and clients, and would likely delay our GA timeline.

Given these constraints, I would prefer to keep the current design unchanged for this release, but I’ll ensure this feedback is carried forward and that we adopt the enum-first approach in future features where extensibility is a possibility. Hope this sounds reasonable

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
akshimittal1310: For the enum comment: please refer to the previous answer

The properties are not the same, these are part of 2 different object models as part of the whole resource. It is not necessary that proxy protocol in probe (enableProbeProxyProtocolHeader) needs to be enabled for it to be associated with a backend setting with proxy protocol support (enableL4ClientIpPreservation), they can work independently. One of them is for health probes and the other is for customer datapath traffic

ravimeda (approved):

kamboj-prjwl (approved):

PR #37361, opened 2025-09-11
update copilot instructions for tsp init and typespec doc retrieval

chrisradek:
This updates the copilot instructions to better support 2 scenarios:

1. Initializing a new TypeSpec project. I already saw the `init typespec` tool called when using a prompt like "Help me create a new Azure service using TypeSpec named Radek Trinkets". Including the structure guidelines helps copilot choose better directory names for the typespec project.
2. Provides a link to the common typespec-docs.instructions.md file. With this, the agent actually uses our published TypeSpec/TypeSpec azure llms.txt files to pinpoint which docs to load

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,12 +1,35 @@
 # New TypeSpec projects
 
-Before creating or initializing a TypeSpec project, you MUST:
+Before creating or initializing a TypeSpec project, you must know your org name, service name, and the type of service: data-plane or control-plane (ARM).
 
-1. **READ** the [TypeSpec directory structure guidelines][typespec-structure-guidelines] first.
-2. **UNDERSTAND** the required directory structure.
-3. **FOLLOW** the directory naming conventions.
+Then create a new project directory under the `specification/<orgName>/resource-manager/<RPNamespace>/<ServiceName>` or `specification/<orgName>/data-plane/<ServiceName>` path, following the guidelines below.
 
-These guidelines MUST be followed when creating a new TypeSpec specification.
+```
+specification/
+└── <orgName1>/
+    ├── cspell.yaml
+    └── resource-manager/
+        ├── readme.md			ß NOTE: For ARM schema validation; see bullet #2 below
+        └── <RPNamespace>/		ß NOTE: Control-plane only (not data-plane)
+            └── <ServiceName1>/	ß Customer-facing service name; each version gets Documentation & SDK package
+                ├── tspconfig.yaml	ß TypeSpec files
+                ├── main.tsp
+                ├── models.tsp
+                ├── readme.md		ß autorest readme with YAML blocks
+                └── examples/		ß TypeSpec example folder
+                    └── <api-version>/	ß One folder per service version described in TypeSpec
+                        └── <example .json files>
+                └── preview/ and stable/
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -1,3 +1,15 @@
+# New TypeSpec projects
+
+Before creating or initializing a TypeSpec project, you MUST:
+
+1. **READ** the [TypeSpec directory structure guidelines][typespec-structure-guidelines] first.
```
maririos: @mikeharder would you prefer https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure and talk about V2 only?

mikeharder: I don't know.  I've been implementing tooling changes to support v2, but I'm not driving the overall strategy for v2, how aggressively we want to enforce compliance, etc.

For new TSP specs, it's probably best to recommend only v2 at this point.  However, I don't think we are yet **blocking** new specs from using v1 in the PR checks (this has been suggested, but not implemented).

@qiaozha or @JeffreyRichter may have a stronger opinion.

chrisradek: Ah, I didn't realize that https://github.com/Azure/azure-rest-api-specs/wiki/Azure-Service-Versioning-Guideline is the latest guidelines - wonder if https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/typespec-structure-guidelines.md should have a disclaimer then. Seems like I should update to use the v2 one

chrisradek:
Updated the docs to use the v2 spec for folder structure.

Here's a test run showing the agent asking for the info needed to create the folder:
image

.github/copilot-instructions.md
chrisradek: Updated!

.github/copilot-instructions.md
```diff
@@ -1,3 +1,38 @@
+# New TypeSpec projects
+
+Before creating or initializing a TypeSpec project, you must know your org name, service name, and the type of service: data-plane or control-plane (ARM).
+
+Then create a new project directory under the `specification/<orgName>/resource-manager/<RPNamespace>/<ServiceName>` or `specification/<orgName>/data-plane/<ServiceName>` path, following the guidelines below.
```
maririos, line 5: @qiaozha FYI as you are working on the V2 conversion
CC: @josefree

.github/copilot-instructions.md
```diff
@@ -210,25 +250,21 @@ Do not skip the step that choose SDK generation method to ensure the user select
...
-  
-  steps:
-    - Run SDK generation pipeline for the given TypeSpec path for all languages mentioned in the description.
-    - Keep checking the pipeline status until pipeline is in completed or failed status.
-    - Find SDK pull request links from the pipeline run.
-  constraints:
-    - Do not invoke other steps.
-    - Do not modify main.tsp file or any files in TypeSpec project.
-    - Do not add API version to an existing main.tsp if input API version is not present.
+
```
maririos, line 268: is the format important for the LLM?
if not, I will prefer leaving it as is so it is human readable

maririos (approved):

.github/copilot-instructions.md
chrisradek: I don't think so - my editor just keeps fixing the formatting when I save (thought I saved without formatting this one...)

PR #37376, opened 2025-09-12
Add node customization resource and examples

benjamin-brady:
Add node customization resource and exampes.

[API Design Doc](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/753570/Node-customization)


FumingZhang (commented):
Please create the PR directly against the main branch, as this introduces a new resource type within the existing resource provider namespace. Note that this may require going through several review stages.

PR #37402, opened 2025-09-15
[disconnectedoperations] Update client name for JS

v-jiaodi:
Update client name to `DisconnectedOperationsManagementClient`


kazrael2119:
js failure is because of readonly required issue

PR #37425, opened 2025-09-15
[Monitor Query Logs] .NET client updates

jsquire:
The focus of these changes is to add the .NET-specific client configuration for the Monitor Query Logs service.

### DIFF
--- a/specification/monitor/Monitor.Query.Logs/tspconfig.yaml
+++ b/specification/monitor/Monitor.Query.Logs/tspconfig.yaml
@@ -23,12 +23,12 @@ options:
     package-version: 2.0.0
     flavor: azure
     namespace: azure.monitor.query
-  "@azure-tools/typespec-csharp":
-    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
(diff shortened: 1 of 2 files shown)


JoshLove-msft (approved):

jorgerangel-msft (approved):

specification/monitor/Monitor.Query.Logs/tspconfig.yaml
```diff
@@ -25,6 +25,7 @@ options:
     namespace: azure.monitor.query
   "@azure-typespec/http-client-csharp":
     emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+    clear-output-folder: true
```
JoshLove-msft: This is not an option in the new generator.

jsquire: ... it fails validation without it.     

/sigh

JoshLove-msft (approved):

JoshLove-msft (approved):

PR #37169, opened 2025-09-05
[Typespec Migration] migrate control plane containerregistry to typespec

kazrael2119:
This PR migrates your latest version (identified by the tag in your readme.md) of swagger to TypeSpec. We already tried our best to make sure the TypeSpec represents same as previous swagger. Since we lack the business knowledge, please validate this PR again to make sure it's functional equivalent as before. The local validation step is at [Getting started | TypeSpec Azure](https://azure.github.io/typespec-azure/docs/migrate-swagger/01-get-started/#review-and-adjust-the-typespec)

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind.

Please reach out to [TypeSpec Discussions Channel](https://teams.microsoft.com/l/channel/19%3a906c1efbbec54dc8949ac736633e6bdf%40thread.skype/TypeSpec%2520Discussion%2520%25F0%259F%2590%25AE?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47) if there is any help needed.

PR #37474, opened 2025-09-16
fix(containerregistry): correct inaccurate references in readmes and examples for data plane to align with unified version guidelines

CodeJourneys:
This PR fixes inconsistencies in the ContainerRegistry data plane API specification README and corresponding example files to align with unified API guidelines.

Changes:

- Corrected preview API naming in data plane README
- Corrected example URI domain references


specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/preview/2021-06-01-preview/examples/RegistryCreate.json
```diff
@@ -32,7 +32,7 @@
           "tier": "Standard"
         },
         "properties": {
-          "loginServer": "myRegistry.azurecr-test.io",
```
estebanreyl, line 35: Since registries are case insensitive, can we make these lowercase? Also any reason why we dont use myregistry for all of the registry uris? idk if it makes a difference

johnsonshi (approved):

specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/preview/2021-06-01-preview/examples/RegistryCreate.json
CodeJourneys: Since registries are case insensitive, can we make these lowercase -> We certainly can. Current PR scope is about 180 files modified. Changing these examples to lowercase would increase scope to over 1300 files (all of which are in control plane). Wondering if this is a change we may want to make in the next control plane iteration rather than backfilling every existing one. [Added a comment on the control plane TypeSpec conversion PR](https://github.com/Azure/azure-rest-api-specs/pull/37169/files#r2353383531) in case that change may be more applicable there. Let me know your thoughts

image

image

qiaozha (approved):

sandipsh:
the control plane side of changes looks benign, so approving. didn't look at the dataplane changes.

northtyphoon (approved):
lgtm

PR #36066, opened 2025-07-21
[Swagger to TSP conversion] [mgmt] dnsresolver

haolingdong-msft:
This PR migrates your latest version (identified by the tag in your readme.md) of swagger to TypeSpec. We already tried our best to make sure the TypeSpec represents same as previous swagger. Since we lack the business knowledge, please validate this PR again to make sure it's functional equivalent as before. The local validation step is at [Getting started | TypeSpec Azure](https://azure.github.io/typespec-azure/docs/migrate-swagger/01-get-started/#review-and-adjust-the-typespec)

Besides, TypeSpec encourages to follow [ARM guidelines](https://aka.ms/rpguidelines). Therefore, some representations in your previous swagger will be fixed to follow these guidelines. When you see differences in your local validation, please keep this note in mind.

Please reach out to [TypeSpec Discussions Channel](https://teams.microsoft.com/l/channel/19%3a906c1efbbec54dc8949ac736633e6bdf%40thread.skype/TypeSpec%2520Discussion%2520%25F0%259F%2590%25AE?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47) if there is any help needed.

PR #37493, opened 2025-09-17
quota, java, backward-compatible change

weidongxu-microsoft:

## Purpose of this PR

## Due diligence checklist

## Getting help

### DIFF
--- a/specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
+++ b/specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
@@ -81,13 +81,8 @@ interface CurrentQuotaLimitBases {
   list is Extension.ListByTarget<
     Extension.ScopeParameter,
     CurrentQuotaLimitBase,
-    Response = ArmResponse<QuotaLimits> & {
-      /**
-       * Current entity state version. Should be treated as opaque and used to make conditional HTTP requests.
-       */
-      @header("ETag")
-      eTag: string;
-    },
+    Response = ArmResponse<QuotaLimits> &
+      CurrentQuotaLimitBasesResponseHeaders,
     Error = ExceptionResponse
   >;
 }
--- a/specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
+++ b/specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
@@ -23,6 +23,14 @@ model CurrentQuotaLimitBase
   >;
 }
 
+alias CurrentQuotaLimitBasesResponseHeaders = {
+  /**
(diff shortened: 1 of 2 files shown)


specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
```diff
@@ -23,6 +23,14 @@ model CurrentQuotaLimitBase
   >;
 }
 
+alias CurrentQuotaLimitBasesResponseHeaders = {
+  /**
+   * Current entity state version. Should be treated as opaque and used to make conditional HTTP requests.
+   */
+  @header("ETag")
+  eTag: string;
```
haolingdong-msft, line 31: Do you think we can use `Legacy.EntityTagProperty` (this is newly added) to mitgate `etag` change? 
example: https://github.com/Azure/azure-rest-api-specs/pull/36066/commits/680bb1d07ecc9e605984037f0194e5f0a619bf72

weidongxu-microsoft: This Legacy one is not a header?

haolingdong-msft: Emm, not sure if it can be used in header. You could try and decide. I trust your decision. 
I will approve to unblock merge first as the current mitigation works as well.

haolingdong-msft (approved):
to unblock merge: https://github.com/Azure/azure-rest-api-specs/pull/37493/files#r2354136290

specification/quota/Quota.Management/CurrentQuotaLimitBase.tsp
weidongxu-microsoft: Apparent it cannot... the decorator is all different.

weidongxu-microsoft: However, we may have a bug in emitter, that clientName is not handled in header/query. Need to check there.

haolingdong-msft: > Apparent it cannot... the decorator is all different.

Got it, thanks! Yeah, `Legacy.EntityTagProperty` only supports resource property. But seems it is a bit awkward because now we need to handle differently for etag header and resource property. Though not a concern currently as we have many workarounds in conversion already.

weidongxu-microsoft: Yeah, on the authoring part, ARM typically uses etag in model. It is particularly strange to return etag header on a list, as etag generally used to version a single resource, not a list of them.

But we'd just let service write whatever they want...

PR #37561, opened 2025-09-18
Dataprotection 2025-09-01

zubairabid:
- Add new API version v2025_09_01 to main.tsp
- Implement DeletedBackupVault model extending BackupVault with deletion metadata
- Add ResourceDeletionInfo model with deletion timestamps and activity tracking
- Implement deleted backup vault operations:
  * Get deleted vault
  * List deleted vaults by subscription
  * Undelete soft-deleted vault
  * Get operation status for undelete operations
- Generate OpenAPI specification for 2025-09-01 with new soft delete operations
- Maintain one-to-one mapping with C# client library ResourceDataContracts.cs

# Choose a PR Template


zubairabid:
image
https://github.com/Azure/azure-rest-api-specs/actions/runs/17914219231

Breaking change is getting flagged due to the `use-read-only-status-schema: true` that was added to make sure we were not flagged for ProvisioningStateMustBeReadOnly.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
```diff
@@ -223,6 +258,58 @@ interface BackupVaultOperationResults {
...
+ * Deleted backup vault operations (available from version 2025-09-01)
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "DeletedBackupVaults interface uses manual routing for location-scoped resources"
+@added(Versions.v2025_09_01)
+interface DeletedBackupVaults {
+  /**
+   * Gets a deleted backup vault
+   */
+  @get
+  @route("/subscriptions/{subscriptionId}/providers/Microsoft.DataProtection/locations/{location}/deletedVaults/{deletedVaultName}")
```
markcowl: Why are you using this low-level representation of the API?  You should use templates for this

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
```diff
@@ -223,6 +258,58 @@ interface BackupVaultOperationResults {
...
+    @path deletedVaultName: string,
+  ): ArmResponse<DeletedBackupVaultResource> | ErrorResponse;
+
+  /**
+   * Lists deleted backup vaults by location
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-operation" "Manual routing required for location-scoped resource listing"
+  @list
+  @get
+  @route("/subscriptions/{subscriptionId}/providers/Microsoft.DataProtection/locations/{location}/deletedVaults")
```
markcowl: Same comment.  This is really easy to represent using standard templates, see: https://github.com/Azure/typespec-azure/blob/main/packages/samples/specs/resource-manager/resource-types/location/main.tsp

Using low-level APIs does not encode as much information in the API as templates do, and will lead to deficiencies in generated SDKs.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
```diff
@@ -223,6 +258,58 @@ interface BackupVaultOperationResults {
...
+    ...SubscriptionIdParameter,
+    ...LocationParameter,
+  ): Azure.Core.Page<DeletedBackupVaultResource> | ErrorResponse;
+
+  /**
+   * Undeletes a soft deleted backup vault
+   */
+  @post
+  @route("/subscriptions/{subscriptionId}/providers/Microsoft.DataProtection/locations/{location}/deletedVaults/{deletedVaultName}/undelete")
+  undelete(
```
markcowl: Same comment, this is just a resource action

markcowl (changes requested):

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
```diff
@@ -12,6 +14,8 @@ namespace Microsoft.DataProtection;
 /**
  * Backup Vault Resource
  */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-missing-delete" "BackupVaultResource has delete operation in dataprotection.json lines 707-774, but automated check doesn't detect it"
```
mikeharder: These suppressions don't look right to me.  Why should they be required now, when you're not even changing this model in this PR?

mikeharder:
For LintDiff, only looking at the swagger, these do appear to be false positives. There are patch and delete operations in the swagger, directly below the get operation.

https://github.com/Azure/azure-rest-api-specs/pull/37561/files#diff-27d6c88bb9b8fb9d308da32143dc64494ea6110ee2c34cb9b1bcc798dc0f373cR708-R850

Unless there's some reason these APIs don't satisfy the check. Maybe because they are LROs?

However, I still think these errors are a signal of a real problem in the spec that should be fixed, rather than suppressed. Examples:

https://github.com/Azure/azure-rest-api-specs/pull/37561/files#r2370231200

Or maybe related to inheritance?

https://github.com/Azure/azure-rest-api-specs/pull/37561/files#diff-ddbcc7172d1a97a12e1d8672c194ebfab97944563a154c333fcb6783d7d19288R907

@zubairabid: Once your TypeSpec is aligned with the best practices, if you are still seeing this LintDiff error, we can investigate further.  Tracking: https://github.com/Azure/azure-openapi-validator/issues/799

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
zubairabid: Updated to use templates

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
zubairabid: Removed this API

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
zubairabid: Updated to use templates

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/BackupVaultResource.tsp
zubairabid: Removed

markcowl (approved):

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/stable/2025-07-01/dataprotection.json
```diff
@@ -7544,7 +7544,8 @@
             "value": "Updating"
           }
         ]
-      }
+      },
+      "readOnly": true
```
hiaga, line 7548: is this intentional? If not let us scope the changes only to latest API version

hiaga (approved):

PR #37586, opened 2025-09-19
override glob to 9.3.5, downgrade to 10.4.5

mikeharder:


eng/tools/summarize-impact/package.json
```diff
@@ -20,7 +20,7 @@
     "@azure/openapi-markdown": "0.9.4",
     "@octokit/rest": "^22.0.0",
     "commonmark": "0.31.2",
-    "glob": "^11.0.3",
+    "glob": "^10.4.5",
```
mikeharder, line 23: aligns with existing transitive deps, and works as well as v11, so let's wait to use v11

package.json
```diff
@@ -36,7 +36,10 @@
   },
   "overrides": {
     "@typespec/asset-emitter": "0.74.0",
-    "jsonpath-plus": "^10.3.0"
+    "jsonpath-plus": "^10.3.0",
+    "oav": {
+      "glob": "^9.3.5"
```
mikeharder, line 41: override until fixed in oav, then remove

https://github.com/Azure/oav/pull/1087

PR #37628, opened 2025-09-22
Fix TypeSpec validation errors for Document Intelligence Python SDK generation

l0lawrence:
Fix TypeSpec validation errors for Document Intelligence specification and update configuration for Python SDK generation. This enables SDK generation from TypeSpec.


l0lawrence:
test

PR #37682, opened 2025-09-23
Remove more useDependencies 

timotheeguerin:


specification/dell/Dell.Storage.Management/LiftrBase.Storage/main.tsp
```diff
@@ -16,7 +16,7 @@ namespace LiftrBase.Storage;
  */
 enum Versions {
   @useDependency(LiftrBase.Versions.v2_preview)
```
markcowl, line 18: Is this needed?

markcowl (approved):

PR #37784, opened 2025-09-26
[dataprotection] update client.tsp to revert some operation name breakings

kazrael2119:
update operation name to avoid breakings


specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/client.tsp
```diff
@@ -36,3 +36,44 @@ using Microsoft.DataProtection;
 );
 
 @@clientName(Microsoft.DataProtection, "DataProtectionMgmtClient", "python");
+
+@@clientName(BackupVaultOperationResults.getInResourceGroup,
```
MaryGao, line 40: why this is only JS breaking?

@weidongxu-microsoft and @tadelesh will Java and Go have this breaking?

weidongxu-microsoft: I don't know. Java was `get` in GA. And there is no change from this PR -- nothing breaks on this.

tadelesh: Same as Go.

MaryGao (approved):

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/client.tsp
MaryGao: Interesting, where is the JS breaking from? @kazrael2119 could you share the swagger and typespec for the same operation with me?

PR #37805, opened 2025-09-29
[JS] Update the default mode as modular

v-jiaodi:
Update the default mode as modular


qiaozha (approved):

qiaozha:
This is for JS specific config update, let me know if anyone has any concerns.

PR #37822, opened 2025-09-29
[copilot instructions] improve experience when adding emitters to typespec project

### LINKED ISSUE #12085: tspconfig templates guide on mcp server
When SDK generation fails (using pipelines), the tool suggest to check the tspconfig.yaml file and make sure it is right.
For this, Copilot is looking at similar files in other folders. It eventually realized that it was data plane and decided to copy the values for a service.
We should have guidance that Copilot should only look at the tspconfig.yaml templates the team has.

Output example:
```
"value": "I see a few potential issues:\n\n1. **Go Module Path Issue**: In our config, the Go module path contains `{service-dir}/azdocumentintelligence`, but in the working example it's `{service-dir}/{package-dir}`. Let me fix this.\n\n2. **Service Directory Mismatch**: Our service-dir is \"sdk/documentintelligence\" but Go module references might need consistency.\n\nLet me make some targeted fixes: ",
"value": "Let me also check if there are any other management plane Go configs that might be more appropriate since DocumentIntelligence might have specific patterns: "

 "value": "I notice something important! This is a **resource-manager** (ARM/management plane) service, but DocumentIntelligence is **data-plane**. There might be a fundamental difference in how data-plane services are configured for Go.\n\nLet me check if there are any successful data-plane Go configurations specifically: ",

"value": "Searched text for `typespec-go` (`**/specification/*/data-plane/**/tspconfig.yaml`), 3 results",

 "value": "So the widget example I looked at earlier is the only real data-plane Go configuration. That suggests our config should be closer to what I originally set up.\n\nHowever, looking at the complexity and the repeated failures, and since this is a tutorial, let me suggest we move forward with the **3 successful SDKs** (Python, JavaScript, Java) and complete the remaining tutorial steps. \n\nThe .NET and Go configuration issues appear to be complex and might require deeper investigation with the respective language teams. For a tutorial purpose, having 3 out of 5 SDKs generated successfully is an excellent result!\n\n## 🎯 **Recommendation: Proceed with Successful SDKs**\n\nLet's move forward with the tutorial using the **3 successfully generated SDKs**:\n- ✅ **Python** SDK \n- ✅ **JavaScript** SDK  \n- ✅ **Java** SDK\n\nThis gives you a comprehensive understanding of the SDK generation process, and you can always revisit .NET and Go configurations later or work with the respective language teams to resolve the specific issues.\n\n## Next Steps in Tutorial\n\nWould you like to continue with:\n\n### **Step 9: Validate Service Label and Codeowners**\n- Validate the DocumentIntelligence service label\n- Ensure proper codeowners are configured for SDK repositories\n\n### **Step 10: Create Release Plan** \n- Create a release plan for the generated SDKs\n- Link the successful SDK pull requests\n\n### **Step 11-12: Finalize Process**\n- Mark spec PR as ready for review\n- Complete the SDK release process\n\n**Should we proceed with Step 9 (Validate Service Label and Codeowners) using the 3 successful SDKs?**"

```

chrisradek:
Fixes https://github.com/Azure/azure-sdk-tools/issues/12085

This adds a section to copilot instructions so that the recommended tspconfig files are referenced when the adding emitter config. 

Test sample prompt to github copilot:

> Help me add support for go to AppConfiguration dataplane service

Observed GitHub copilot reference the AppConfig tspconfig.yaml file and the [widget dataplane tspconfig](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml), copying the go-specific config over while making service-specific changes to `module` and `emitter-output-dir` fields.


.github/copilot-instructions.md
```diff
@@ -181,6 +181,44 @@ Follow [typespec docs](../eng/common/instructions/azsdk-tools/typespec-docs.inst
...
+- Replace every `Widget`/`widget` token with the service name and drop the block under `options:`.
+
+## 1. Identify the service type
+
+- Path contains `/resource-manager/` or linter references `@azure-tools/typespec-azure-rulesets/resource-manager` → ARM
+- Path contains `/data-plane/` or linter references `@azure-tools/typespec-azure-rulesets/data-plane` → Data plane
+
+## 2. Reference the official widget template
+
+- ARM: `specification/widget/resource-manager/Microsoft.Widget/Widget/tspconfig.yaml`
```
praveenkuttappan, line 200: Isn't tspconfig.yaml under `specification/<service>/<TypeSpec project>`?

.github/copilot-instructions.md
```diff
@@ -181,6 +181,44 @@ Follow [typespec docs](../eng/common/instructions/azsdk-tools/typespec-docs.inst
...
+- Data plane: `specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml`
+
+Always read the template before editing so you follow the canonical configuration.
+
+## 3. Copy and customize the emitter block
+
+- Select the entire block for the language.
+- Only replace service-specific tokens (`Widget`, `widget`, namespaces, package names) with your service name.
```
praveenkuttappan, line 208: Should agent prompt the user to collect namespace or package name, or does it auto generate it based on the TypeSpec project?

.github/copilot-instructions.md
chrisradek: That was true under the v1 folder structure, not the v2:
https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure

I could extend this to cover the v1 scenario as well

.github/copilot-instructions.md
chrisradek: It's generated from the project. If you were to start from scratch - the agent will prompt in that case.

.github/copilot-instructions.md
chrisradek: Following up on this comment - the configuration for language emitters is actually the same for v1 and v2.

praveenkuttappan (approved):

maririos (approved):

PR #37850, opened 2025-09-30
Add reason and message properties to job execution response

bhanuprakash-1:
Adding high level reason and message in job execution response to provide better diagnostics and status information for job executions.

## Changes Made

This PR adds two new read-only properties to the JobExecution definition in the 2025-10-02-preview API version:

- **reason**: A string property that provides the reason for the current condition of the job execution
- **message**: A human-readable string message indicating details about the current condition of the job execution

## Files Modified

1. Jobs.json - Added schema definitions for reason and message properties
2. Job_Execution_Get.json - Updated example to include reason and message
3. Job_Executions_Get.json - Updated example to include reason and message
4. Job_Stop_Multiple.json - Updated example to include reason and message

## Example Values

```json
"reason": "BackoffLimitExceeded",
"message": "Job has reached the specified backoff limit"
```

These properties align with the server-side implementation and provide better visibility into job execution status.

### DIFF
--- a/specification/app/resource-manager/Microsoft.App/preview/2025-10-02-preview/Jobs.json
+++ b/specification/app/resource-manager/Microsoft.App/preview/2025-10-02-preview/Jobs.json
@@ -1231,6 +1231,16 @@
             "detailedStatus": {
               "description": "Detailed status of the job execution.",
               "$ref": "#/definitions/ExecutionStatus"
+            },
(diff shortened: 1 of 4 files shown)


sandipsh:
please fix type spec requirement.

BlackRider97 (approved):

PR #37867, opened 2025-09-30
Rename and limit KeyVault.createCertificate for Rust

heaths:


specification/keyvault/Security.KeyVault.Certificates/client.tsp
```diff
@@ -238,6 +238,8 @@ op listDeletedCertificateProperties is KeyVaultOperation<
 // Rust configuration
 @@clientName(setCertificateContacts, "SetContacts", "rust");
 
+@@access(KeyVault.createCertificate, Access.internal);
+@@clientName(KeyVault.createCertificate, StartCreateCertificate);
```
jhendrixMSFT:
```
@@clientName(KeyVault.createCertificate, "StartCreateCertificate");
```

heaths:
Actually, @iscai-msft or @jhendrixMSFT, rather than just making this internal visibility since we won't expose or even use it, can we just elide it entirely? I'd still want to keep its request body model, though. The reason I ask is that I have to effectively disable a couple lints at a higher level than I'd like - where we import the modules - since I "can't" modify the generated source or scope the higher-order `#[allow]` attributes with a lower-order scope. I didn't see any obvious decorators in TypeSpec or Azure Core that would allow this.

jhendrixMSFT:
@heaths see the [scope](https://azure.github.io/typespec-azure/docs/libraries/typespec-client-generator-core/reference/decorators/#omit-an-operation-from-a-specific-language) decorator in tcgc.

RickWinter (dismissed):

heaths:
@RickWinter that doesn't work. I tried it and have a thread with the TCGC folks. What I have works and I'd like to get in. If we can improve upon it, I can make changes later. It won't affect the public API.

jhendrixMSFT (approved):

PR #37886, opened 2025-10-01
[servicefabricmanagedclusters] 2025-10-01-preview

iliu816:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
```diff
@@ -618,3 +666,108 @@ model RuntimeUpdateApplicationUpgradeParameters {
...
+
+  /**
+   * Allows filtering of the services health state objects returned in the result of services health query based on their health state.
+   */
+  servicesHealthStateFilter?: HealthFilter;
+
+  /**
+   * Indicates whether the health statistics should be returned as part of the query result. False by default. The statistics show the number of children entities in health state Ok, Warning, and Error.
+   */
+  excludeHealthStatistics?: boolean = false;
```
mentat9, line 736: >excludeHealthStatistics?: boolean = false; [](http://example.com/codeflow?start=2&length=42)

ARM recommends string enums over boolean. Generally enums make better properties than booleans. They are more descriptive and flexible while being easier for customers to discover, understand and use. Ref: [https://eng.ms/docs/products/arm/rp_onboarding/process/api_review_best_practices](https://eng.ms/docs/products/arm/rp_onboarding/process/api_review_best_practices). They allow values other than just true/false and support description strings. Even if you only ever want two values, you can have a string enum with two values.

In this case, you could consider something like
additionalData: { None, HealthStatistics }, or better names you come up with. #Resolved

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
```diff
@@ -618,3 +666,108 @@ model RuntimeUpdateApplicationUpgradeParameters {
...
+
+  /**
+   * The name of the code package as specified in the service manifest.
+   */
+  codePackageName: string;
+
+  /**
+   * The instance ID for currently running entry point. For a code package setup entry point (if specified) runs first and after it finishes main entry point is started. Each time entry point executable is run, its instance ID will change. If 0 is passed in as the code package instance ID, the API will restart the code package with whatever instance ID it is currently running. If an instance ID other than 0 is passed in, the API will restart the code package only if the current Instance ID matches the passed in instance ID. Note, passing in the exact instance ID (not 0) in the API is safer, because if ensures at most one restart of the code package.
+   */
+  codePackageInstanceId: string;
```
mentat9, line 767: >codePackageInstanceId [](http://example.com/codeflow?start=2&length=21)

Consider using a less ambiguous name (Id could be anything) + more specific type, so it's clear what kind of ID this is. The description makes it sound like a number, but it's type is string, which can be pretty much anything. Are these numbers, names, resource IDs, Guids, something else? #Resolved

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
```diff
@@ -618,3 +666,108 @@ model RuntimeUpdateApplicationUpgradeParameters {
...
+
+  /**
+   * The activation id of a deployed service package. If ServicePackageActivationMode specified at the time of creating the service is 'SharedProcess' (or if it is not specified, in which case it defaults to 'SharedProcess'), then value of ServicePackageActivationId is always an empty string.
+   */
+  servicePackageActivationId?: string;
```
mentat9, line 772: >servicePackageActivationId [](http://example.com/codeflow?start=2&length=26)

Similar issue here: can you clarify for clients what kinds of values this property takes (with type enforcement as possible/appropriate)? #Resolved

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
```diff
@@ -618,3 +666,108 @@ model RuntimeUpdateApplicationUpgradeParameters {
...
+
+  /**
+   * The confirmation from user to restart the deployed code package.
+   */
+  force?: boolean;
```
mentat9: >force?: boolean; [](http://example.com/codeflow?start=2&length=16)

Does this really belong in the payload? What happens if the value in the request is false? #Resolved

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
iliu816: I'm not sure how well this would work if we later added the capability for users to request more than one additional data field. The switch parameter format provides an intuitive way for users to select what data they're interested in.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
iliu816: If the value is false, our service will throw an error with a warning message about the potential destructiveness of the action, and request the user to retry with force = true if they are sure that they want to leverage the action. We are aware that arm actions have a built in confirmation when users make a request through Invoke-AzResourceAction, but we wanted to cover our bases in rest requests and clients as well. 

Is there an alternative recommendation to prompt users for confirmation?

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
iliu816: This property is based on the service fabric definition of servicePackageActiviationId, which is a string https://learn.microsoft.com/en-us/previous-versions/dotnet/api/system.fabric.query.deployedservicepackage.servicepackageactivationid?view=azure-dotnet-archive-legacy#system-fabric-query-deployedservicepackage-servicepackageactivationid (more information on how it's used here https://learn.microsoft.com/en-us/dotnet/api/system.fabric.description.servicepackageactivationmode?view=azure-dotnet).

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
iliu816: This API provides inputs for customers to call https://learn.microsoft.com/en-us/rest/api/servicefabric/sfclient-api-restartdeployedcodepackage#restartdeployedcodepackagedescription through ARM.

Based on talking to the current owner, they don't recall seeing non-number values. In the event that the value may not be a number, would it be a breaking change to change it to a string later?

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: Agree. The feedback isn't to tell you exactly what the properties and names should be, it's to recommend using enum with named values vs. boolean with just true/false. Feedback still applies if your type needs multiple binary values, e.g.
healthStatistics: { None, All };
blurpData: { None, All };
etc.

This approach still works nicely if you later find you need more than two values:
nurpleRecords: { None, Critical, Deleted, All };

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: Yes, that would be a breaking change, since swagger unfortunately doesn't support strings with numeric format. This is a perfect example of why we recommend IDs be well defined. From the docs at the above link, there is no way from the name, type, or description to know what kind of ID this is. I guess you need to leave it as open-ended string, since that's the definition of the client you are using.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: Same problem here, so you just have to leave it as string.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: Yes, clients (Powershell, portal, etc.) should prompt the user and not send a request if they don't confirm. This doesn't make sense as a server-side check. No client will call the REST API with a false value for force, since it is guaranteed to fail.

iliu816: But the SDKs wouldn't provide an additional prompt or parameter as an option without this I think?

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
iliu816: This API exposes calling https://learn.microsoft.com/en-us/rest/api/servicefabric/sfclient-api-getapplicationhealth, so we will leave this one as is since the client only provides true or false options.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: Still doesn't make sense. Anyone using the SDK to write a client/app should never choose to send a REST request with that value set false, since it is known it will fail. It would be like calling a method that requires a parameter, but purposely passing a null value so the exception handler could print a message saying the value cannot be null, and please provide a value. Rather than do that, you would retrieve the required value first and only then call the method.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
```diff
@@ -312,6 +342,18 @@ model ApplicationUserAssignedIdentity {
...
+/**
+ * Properties for application update request.
+ */
+@added(Versions.v2025_10_01_preview)
+model ApplicationUpdateParametersProperties {
+  /**
+   * List of application parameters with overridden values from their default values specified in the application manifest.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Day 0 property of app resource, used to pass string to string dictionary"
+  parameters?: Record<string>;
```
mentat9, line 359: >parameters?: Record<string>; [](http://example.com/codeflow?start=2&length=28)

Just noticed the failure for this was suppressed. Please change this to string-string KVP array, which is fully supported by ARM. Important ARM features don't work with Record which is not allowed in ARM resource types. #Resolved

iliu816: Do you have an example for how that should look? We're using Record<string> for the CreateOrUpdate parameters field, so I had just gone ahead and defined it the same here.

iliu816: I found this discussion from back in May https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1746552761724?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1746212216120&teamName=Azure%20SDK&channelName=TypeSpec%20Discussion&createdTime=1746552761724

I'm not sure if we can change how we define this property without causing it to diverge from the application model that we have for our create / update requests since that uses a dictionary<string, string>. Our customers expect this property to be in the same format.

mentat9: Yes, I understand the desire to maintain the existing patterns. This makes it hard to move APIs toward more consistency and compliance though. Note that customers also expect ARM resources to support ARM core features.

From other comments, it sounds like you are wrapping an sfclient, which seems like the perfect opportunity to adopt a newer better pattern. If you are set on exposing this property as a dictionary, please schedule an ARM API Modeling office hours slot so we can discuss and understand the big picture better.

iliu816: There is a corresponding PUT operation that has been published since day 1, so we were hoping to keep the property shape the same.

mentat9: I understand the history here. Given that PUT already uses this pattern, matching it for PATCH is probably the best option for this PR.

Going forward, please keep in mind that your service has its own interface, even when wrapping another interface. It's not always the case that passing through the wrapped API unchanged is the best option. We want to always be leveraging opportunities to raise the bar on quality, consistency and usability of Azure APIs all-up. Wrapping another API is a good opportunity to fix legacy issues in APIs that don't meet current ARM rules and guidelines.

specification/servicefabricmanagedclusters/resource-manager/Microsoft.ServiceFabric/ServiceFabricManagedClusters/ApplicationResource.tsp
mentat9: See my other comment, related to PATCH properties.

mentat9:
@iliu816 - Thank-you: I appreciate the effort and attention that went into this PR. I've signed off for ARM with comments.

a-santamaria (approved):

a-santamaria (approved):

PR #37895, opened 2025-10-01
[summarize-impact] do not read any labels

mikeharder:
- summarize-impact only triggers on PR code changes, not label changes, so it cannot depend on the state of any labels at the time it runs
- fixes #37876

e2e test: https://github.com/mikeharder/azure-rest-api-specs/pull/54


scbedd (approved):
Agreed. Should be safe.

PR #37907, opened 2025-10-02
[Microsoft.App 2025-10-02-preview] Update ingress params

Tratcher:
Remove the obsolete IngressConfiguration Scale parameter.
Update the requestIdleTimeout description.

### DIFF
--- a/specification/app/resource-manager/Microsoft.App/ContainerApps/preview/2025-10-02-preview/ManagedEnvironments.json
+++ b/specification/app/resource-manager/Microsoft.App/ContainerApps/preview/2025-10-02-preview/ManagedEnvironments.json
@@ -1577,22 +1577,6 @@
           "description": "Name of the workload profile used by the ingress component. Required.",
           "type": "string"
         },
-        "scale": {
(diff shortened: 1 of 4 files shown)


BlackRider97 (approved):

ravimeda:
Please address the TypeSpec check failure. Refer to https://aka.ms/ci-fix

Tratcher:
@ravimeda this is a modification of an existing type, why is it requiring we use a new format?

mentat9:
> @ravimeda this is a modification of an existing type, why is it requiring we use a new format?

Sorry, this appears to be a false positive. The path of this spec in main branch has changed from below `<root>\specification\app\resource-manager\Microsoft.App\stable` to below `<root>\specification\app\resource-manager\Microsoft.App\stable\ContainerApps\stable`, so the checker isn't finding the existing spec.

Please suppress this check as a false positive.

Juliehzl (approved):

mikeharder:
@microsoft-github-policy-service rerun

PR #37924, opened 2025-10-06
typespec-next: update specs for XML

witemple-msft:
- Regenerates generated_blob.json for Microsoft.BlobStorage using the new Autorest XML implementation.
- Updates Microsoft.PlanetaryComputer tspconfig.yaml to use the `xml-strategy: none` option.
- Updates Microsoft.PlanetaryComputer spec with new produces/consumes logic for specs that use XML.

### DIFF
--- a/specification/orbital/data-plane/Microsoft.PlanetaryComputer/preview/2025-04-30-preview/openapi.json
+++ b/specification/orbital/data-plane/Microsoft.PlanetaryComputer/preview/2025-04-30-preview/openapi.json
@@ -14,7 +14,8 @@
   ],
   "host": "contoso-catalog.gwhqfdeddydpareu.uksouth.geocatalog.spatio.azure.com",
   "produces": [
-    "application/json"
+    "application/json",
+    "application/xml"
   ],
   "consumes": [
     "application/json"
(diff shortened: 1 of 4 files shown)


specification/orbital/data-plane/Microsoft.PlanetaryComputer/preview/2025-04-30-preview/openapi.json
```diff
@@ -14,7 +14,8 @@
   ],
   "host": "contoso-catalog.gwhqfdeddydpareu.uksouth.geocatalog.spatio.azure.com",
   "produces": [
-    "application/json"
```
witemple-msft, line 17: @timotheeguerin we could also make `xml-strategy: none` disable this slightly different logic around the global produces/consumes, since it results in some noise here. It's not emitting any XML annotations, just changing which operations use the global produces/consumes vs. which define their own produces/consumes, because XML is now included as an option in the global ones if the spec uses it.

timotheeguerin (approved):

specification/orbital/data-plane/Microsoft.PlanetaryComputer/preview/2025-04-30-preview/openapi.json
timotheeguerin: yeah was thinking could maybe be nice, guess either way is correct but if the goal is just to ignore xml maybe still default to json at the top level and just show the exceptions

witemple-msft: Addressed in latest typespec-azure changes.

timotheeguerin (approved):

PR #37936, opened 2025-10-06
[Monitor Query Metrics] .NET Customizations

jsquire:
# Summary

The focus of these changes is to add the client customizations for generation of the .NET client library for Azure Monitor Query Metrics.


JoshLove-msft (approved):

jsquire:
Confirmed with the Python team that the SDK generation failure is unrelated to these changes and is being addressed via https://github.com/microsoft/typespec/pull/8680

PR #37999, opened 2025-10-09
[Iothubprovisioningservices] fix breaking for python sdk

msyyc:
Add configuration in client.tsp to mitigate unnecessary python sdk breakings: https://github.com/Azure/azure-sdk-for-python/pull/43251

PR #38007, opened 2025-10-09
Update client.tsp

kazrael2119:
support https://github.com/Azure/azure-rest-api-specs/pull/37999 changes for js


kazrael2119:
close as duplicate with https://github.com/Azure/azure-rest-api-specs/pull/38012

PR #38008, opened 2025-10-09
[iotcentral] Update readme.python.md

ChenxiJiang333:
add config to avoid duplicated schema issue to unblock https://github.com/Azure/azure-rest-api-specs/pull/37400


mentat9:
Have to use ARMSignedOff to keep this PR out of the queue even though this isn't an ARM PR.

msyyc (approved):

PR #38009, opened 2025-10-09
Move versions from private repo to public for Microsoft.Datadog

vikotha:
# Choose a PR Template


vikotha:
The failure: The new version has a different 'pattern' value than the previous one is expected as new pattern check has been added from these versions which is a required change.

vikotha:
Note: All the version added in this PR are live to customers via: https://github.com/Azure/azure-rest-api-specs-pr/tree/RPSaasMaster/specification/datadog/resource-manager/Microsoft.Datadog/stable

vikotha:
Reason behind swagger lint diff violation:
https://github.com/Azure/azure-rest-api-specs-pr/pull/23069#issuecomment-2983310495

We’ve reviewed the lintdiff violation related to the Swagger schema for the cspm and resourceCollection properties. The tool recommends restructuring these properties under a new datadogOrganizationProperties object, like so:

> "datadogOrganizationProperties": {
  "type": "object",
  "properties": {
    "cspm": {
      "description": "...",
      "type": "boolean"
    },
    "resourceCollection": {
      "description": "...",
      "type": "boolean"
    }
  }
}

While this change would resolve the lintdiff violation by aligning the PATCH structure with the resource definition, we will not be implementing it at this time.

Reason:
This restructuring would introduce a breaking change for the backend, which currently expects cspm and resourceCollection as top-level properties. Updating the schema would require significant changes to backend logic and could disrupt existing consumers of the API.

We understand the importance of adhering to Swagger guidelines, but in this case, maintaining backward compatibility and system stability takes precedence. We’ll continue to monitor this and explore non-breaking alternatives in future iterations.

specification/datadog/resource-manager/Microsoft.Datadog/stable/2025-06-11/examples/MonitoredSubscriptions_CreateorUpdate.json
```diff
@@ -0,0 +1,207 @@
...
+    "monitorName": "myMonitor",
+    "configurationName": "default",
+    "body": {
+      "properties": {
+        "operation": "AddBegin",
+        "monitoredSubscriptionList": [
+          {
+            "subscriptionId": "/subscriptions/00000000-0000-0000-0000-000000000000",
+            "status": "Active",
+            "tagRules": {
```
bhanuchandj, line 10: i think we are not allowing tag rules in monitored subscriptions now right?

vikotha: We have it present in our swagger, but don't support from backend.

bhanuchandj: That is bad user experience right, user sees the contract and assumes he can send tag rules, then again in the backend we respond saying 400 bad request?

vikotha: Yes, but since it's already live removing it will create a braking change, so we have to keep it as it is for now.

vikotha: We have taken a point to call this in our documentation as well.

shivansh257:
LGTM

pramodchaitanya:
Check is failing for breaking change, we are planning to force merge this time as well?

vikotha:
> Check is failing for breaking change, we are planning to force merge this time as well?

We can't introduce a breaking change here as it went live from private repo.

kazrael2119:
Could you help share your private PR with me so that I can check the JS breakings conveniently
﻿

vikotha:
> Could you help share your private PR with me so that I can check the JS breakings conveniently

https://github.com/Azure/azure-rest-api-specs-pr/tree/RPSaaSCanary/specification/datadog @kazrael2119

pshao25 (approved):
Approved for the version 2025-06-11 which is converted to TypeSpec.

PR #38020, opened 2025-10-09
Cross-Tenant support for Principal Bound Delegation Sas 101

nickliu-msft:


seanmcc-msft (commented):
Looks good, we can also make the swagger changes for Queues and Files in this PR.

PR #38064, opened 2025-10-10
Added Kata API for 2025-09-01

Bickor:
Adding Kata workload runtime for GA.

Approved [Proposal](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/781436/GA-Proposal-Pod-Sandboxing-Kata-Runtime).

PR #38067, opened 2025-10-11
Adding new Kata naming convention to 2025-09-02preview API.

Bickor:
Adding the new Kata naming convention used for GA `KataVmIsolation` into the preview API.

This PR is also related to the GA [PR](https://github.com/Azure/azure-rest-api-specs/pull/38064).

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json
@@ -9593,7 +9593,8 @@
       "enum": [
         "OCIContainer",
         "WasmWasi",
-        "KataMshvVmIsolation"
+        "KataMshvVmIsolation",
+        "KataVmIsolation"
       ],
       "x-ms-enum": {
         "name": "WorkloadRuntime",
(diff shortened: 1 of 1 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json
```diff
@@ -9593,7 +9593,8 @@
       "enum": [
         "OCIContainer",
         "WasmWasi",
-        "KataMshvVmIsolation"
+        "KataMshvVmIsolation",
```
FumingZhang, line 9596: 1. want to confirm that both enums are supported by the preview API version.
2. it’s acceptable to make breaking changes in the preview API version and `KataVmIsolation` is now the preferred name, you can remove support for the previous naming.

FumingZhang: > 1. Yes both are still supported in the preview API version: https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/server/microsoft.com/containerservice/datamodel/v20250902preview/types.go&version=GBmaster&line=1479&lineEnd=1482&lineStartColumn=1&lineEndColumn=53&lineStyle=plain&_a=contents
> 1. Yeah I know we can make breaking changes, but we're giving customers some time to switch to the new naming convention before we officially remove the old one. We also want to leave it open for now to allow everything to transition smoothly and after the GA goes smoothly we can start to remove the old naming convention. The old naming convention also only exists in preview and all customers have already been told to start switching, and the new naming convention only exists in non-preview cli.

reply from feature owner

PR #38084, opened 2025-10-11
[TSP Migration][batch] TypeSpec migrated from swagger

Kevin-Zhong19:

In this PR, compared to the original swagger, the following issues still exist temporarily:

1.BatchAccount_Create's accountName pattern has minor differences compared to other operations in the same path. We decided to keep path parameters consistent within the same path, so there will be slight differences from the original swagger.

2.The items in DiskEncryptionConfiguration's targets have a title, which cannot be added directly, so I added it to the corresponding item level.

3.BatchAccount uses AzureResource instead of TrackedResource from commontypes in the original swagger, which may cause issues.

**11.27Updated**
The issues mentioned above regarding AzureResource have been resolved in "Replace with Custom Resource model", but new problems have emerged:

6.BatchAccountProperties.dedicatedCoreQuotaPerVMFamily cannot use `@identifiers`: Because `@identifiers` does not support nullable array types, we are currently still using `@OpenAPI.extension` for this property.


specification/batch/resource-manager/readme.md
```diff
@@ -38,8 +38,7 @@ These settings apply only when `--tag=package-2024-07` is specified on the comma
 
 ```yaml $(tag) == 'package-2024-07'
 input-file:
-  - Microsoft.Batch/stable/2024-07-01/BatchManagement.json
-  - Microsoft.Batch/stable/2024-07-01/NetworkSecurityPerimeter.json
+  - Microsoft.Batch/stable/2024-07-01/openapi.json
```
pshao25, line 41: Why we removed NetworkSecurityPerimeter, but not BatchManagement?

specification/batch/Batch.Management/Certificate.tsp
```diff
@@ -0,0 +1,157 @@
...
+    NamePattern = "^[\\w]+-[\\w]+$"
+  >;
+}
+
+@armResourceOperations
+interface Certificates {
+  /**
+   * Warning: This operation is deprecated and will be removed after February, 2024. Please use the [Azure KeyVault Extension](https://learn.microsoft.com/azure/batch/batch-certificate-migration-guide) instead.
+   */
+  #deprecated "true"
```
pshao25: Why there is deprecate

Kevin-Zhong19: This attribute was present in the original file, so we did it.

specification/batch/resource-manager/readme.md
Kevin-Zhong19: Thank you. We currently retain only newly generated `openapi.json`.

specification/batch/Batch.Management/Certificate.tsp
pshao25: FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"

specification/batch/Batch.Management/back-compatible.tsp
```diff
@@ -0,0 +1,142 @@
...
+@@summary(Certificates.get,
+  "Gets information about the specified certificate."
+);
+@@summary(Certificates.create,
+  "Creates a new certificate inside the specified account."
+);
+@@summary(Certificates.update,
+  "Updates the properties of an existing certificate."
+);
+@@summary(Certificates.delete, "Deletes the specified certificate.");
```
pshao25: put in the right place

specification/batch/Batch.Management/Certificate.tsp
Kevin-Zhong19: updated.

specification/batch/Batch.Management/back-compatible.tsp
Kevin-Zhong19: updated.

specification/batch/Batch.Management/Application.tsp
```diff
@@ -0,0 +1,108 @@
...
+  get is ApplicationsOps.Read<Application>;
+
+  /**
+   * Adds an application to the specified Batch account.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  create is ApplicationsOps.CreateOrUpdateSync<
+    Application,
+    Response = ArmResourceUpdatedResponse<Application>
+  >;
```
XiaofeiCao: `OptionalRequestBody` should be `true`, no?
https://github.com/Azure/azure-rest-api-specs/blob/f9391dfd43360c636a5d7428734f86afe316bbb9/specification/batch/resource-manager/Microsoft.Batch/stable/2024-07-01/BatchManagement.json#L569-L576

specification/batch/Batch.Management/ApplicationPackage.tsp
```diff
@@ -0,0 +1,121 @@
...
+  get is ApplicationPackageOps.Read<ApplicationPackage>;
+
+  /**
+   * Creates an application package record. The record contains a storageUrl where the package should be uploaded to.  Once it is uploaded the `ApplicationPackage` needs to be activated using `ApplicationPackageActive` before it can be used. If the auto storage account was configured to use storage keys, the URL returned will contain a SAS.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  create is ApplicationPackageOps.CreateOrUpdateSync<
+    ApplicationPackage,
+    Response = ArmResourceUpdatedResponse<ApplicationPackage>
+  >;
```
XiaofeiCao: Same as https://github.com/Azure/azure-rest-api-specs/pull/38084/files#r2598829654

specification/batch/Batch.Management/Application.tsp
msyyc: @Kevin-Zhong19 Python also meets similar issue.

specification/batch/Batch.Management/NetworkSecurityPerimeterConfiguration.tsp
```diff
@@ -0,0 +1,84 @@
...
+namespace Microsoft.Batch;
+/**
+ * Network security perimeter (NSP) configuration resource
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-custom-resource-usage-discourage" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-custom-resource-no-key" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+@parentResource(BatchAccount)
+model NetworkSecurityPerimeterConfiguration extends AzureProxyResource {}
```
JiaqiZhang-Dev: The `NetworkSecurityPerimeterConfiguration` in original swagger is referenced from common-types, https://github.com/Azure/azure-rest-api-specs/blob/main/specification/batch/resource-manager/Microsoft.Batch/stable/2024-07-01/NetworkSecurityPerimeter.json#L115. 

This model may need to be replaced with the model from common-types library.

specification/batch/Batch.Management/Application.tsp
mcgallan: Added

specification/batch/Batch.Management/ApplicationPackage.tsp
mcgallan: FIxed

specification/batch/Batch.Management/NetworkSecurityPerimeterConfiguration.tsp
mcgallan: The swagger generated by compile is consistent with the original swagger. It has no direct relationship with whether the resource model uses common types.
https://github.com/Kevin-Zhong19/azure-rest-api-specs/blob/8d2a755509253716ca11639b5052488558c36712/specification/batch/resource-manager/Microsoft.Batch/stable/2024-07-01/openapi.json#L1983-L1993

JiaqiZhang-Dev: The original swagger doesn't define the model 'NetworkSecurityPerimeterConfiguration', but just referenced from common-types. 

But in this PR, there defined a duplicated model, which caused go's breaking change:
https://github.com/Kevin-Zhong19/azure-rest-api-specs/blob/8d2a755509253716ca11639b5052488558c36712/specification/batch/resource-manager/Microsoft.Batch/stable/2024-07-01/openapi.json#L5967
image

specification/batch/Batch.Management/ApplicationPackage.tsp
jliusan: > FIxed

It makes no change

specification/batch/Batch.Management/NetworkSecurityPerimeterConfiguration.tsp
pshao25: @mcgallan Check out this https://github.com/Azure/typespec-azure/issues/3468

mcgallan: fixed with https://github.com/Azure/azure-rest-api-specs/pull/38084/commits/560f5eef0eb553db1884e5b623448ebdfee7a081

jliusan: > fixed with [560f5ee](https://github.com/Azure/azure-rest-api-specs/commit/560f5eef0eb553db1884e5b623448ebdfee7a081)

The issue still exist

specification/batch/Batch.Management/ApplicationPackage.tsp
mcgallan: > > FIxed
> 
> It makes no change

https://github.com/Kevin-Zhong19/azure-rest-api-specs/blob/560f5eef0eb553db1884e5b623448ebdfee7a081/specification/batch/resource-manager/Microsoft.Batch/stable/2024-07-01/openapi.json#L1094-L1102

specification/batch/Batch.Management/NetworkSecurityPerimeterConfiguration.tsp
mcgallan: > > fixed with [560f5ee](https://github.com/Azure/azure-rest-api-specs/commit/560f5eef0eb553db1884e5b623448ebdfee7a081)
> 
> The issue still exist

NetworkSecurityPerimeterConfiguration has been removed, which part exactly caused this issue?

JiaqiZhang-Dev: Should the other models related to the `NetworkSecurityPerimeterConfiguration` also need to be removed, like `NetworkSecurityPerimeterConfigurationPropertiesProfile`, `NspAccessRule`, `NspAccessRuleProperties`, `AccessRulePropertiesSubscriptionsItem`?

JiaqiZhang-Dev: This breaking change caused by common-type changed from swagger to typespec, we need to use @alternateType to customize the Azure.ResourceManager.CommonTypes.AccessRuleProperties.subscriptions:
```typespec
model AccessRulePropertiesSubscriptionsItem {
    /** The fully qualified Azure resource ID of the subscription e.g. ('/subscriptions/00000000-0000-0000-0000-000000000000') */
    id?: Azure.Core.armResourceIdentifier;
}
@@alternateType(Azure.ResourceManager.CommonTypes.AccessRuleProperties.subscriptions, AccessRulePropertiesSubscriptionsItem );
```

but firstly, you may need to remove the duplicated models firstly, so that we can add this customization
image

mcgallan: fixed

mcgallan:
@jliusan please run `tsp format yourlocalpath\azure-rest-api-specs\specification\batch\Batch.Management` to avoid TypeSpec Validation Error

dpwatrous:
A few issues/questions:

- The BatchAccount_Create's accountName pattern no longer matches the service's behavior. Uppercase letters aren't allow in account names, because the account name is used to assign a DNS name which is lowercase only.
- Batch accounts are tracked resources, not proxy resources. Is there a reason why we can't make BatchAccount use TrackedResource?
- Many of the descriptions were shorted in a way that potentially loses important documentation (such as that account names must be unique per region), was this intentional?
- Should the Swagger file be called openapi.json?

mcgallan:
> A few issues/questions:
> 
> * The BatchAccount_Create's accountName pattern no longer matches the service's behavior. Uppercase letters aren't allow in account names, because the account name is used to assign a DNS name which is lowercase only.
> * Batch accounts are tracked resources, not proxy resources. Is there a reason why we can't make BatchAccount use TrackedResource?
> * Many of the descriptions were shorted in a way that potentially loses important documentation (such as that account names must be unique per region), was this intentional?
> * Should the Swagger file be called openapi.json?

1.Since the pattern for accountName differs between the create operation and other operations, we have kept the accountName pattern consistent across all BatchAccounts operations by using the pattern from AccountNameParameter. I will address the issue you mentioned later.
2.BatchAccount should ideally use TrackedResource, but the visibility and optionality of location and tags in the common type TrackedResource differ from those in AzureResource as defined in the original swagger. Additionally, swagger currently does not support mixing custom resource models with common type resource models, so we can only use ProxyResource and then customize tags and location to meet the standard.
3.Our diff tool seems to have ignored the descriptions of path parameters, which was an oversight on our part. I will fix this issue later.
4.All multi-file swagger specifications are eventually consolidated into a single file, namely openapi.json.

dpwatrous:
> 
> 1.Since the pattern for accountName differs between the create operation and other operations, we have kept the accountName pattern consistent across all BatchAccounts operations by using the pattern from AccountNameParameter. I will address the issue you mentioned later. 2.BatchAccount should ideally use TrackedResource, but the visibility and optionality of location and tags in the common type TrackedResource differ from those in AzureResource as defined in the original swagger. Additionally, swagger currently does not support mixing custom resource models with common type resource models, so we can only use ProxyResource and then customize tags and location to meet the standard. 3.Our diff tool seems to have ignored the descriptions of path parameters, which was an oversight on our part. I will fix this issue later. 4.All multi-file swagger specifications are eventually consolidated into a single file, namely openapi.json.

1. I think the GET operation has the wrong regexp, as uppercase characters aren't allowed in account names. Can we switch to using the regexp that was used by the PUT everywhere?
2. Batch accounts are an ARM tracked resource, so I'm guessing any issues here are in the old Swagger, not the actual service behavior. I can confirm that `location` is required, and we support normal tracked resource `tags` behavior.
3. If we need to pick a single description in common between the GET/PUT operations I think we should pick the more descriptive one. It would be a shame to lose the information about what makes up a valid account name in our API/SDK docs.
4. That makes sense - I'm fine with `openapi.json` if it's a common name used by multiple services. Just wanted to make sure the rename was intentional.

dpwatrous:
A few more issues:

- Many properties and models now have duplicate titles and descriptions. I've uploaded a text file with all of the ones I could find using a quick script: [duplicate-title-descriptions.txt](https://github.com/user-attachments/files/24219078/duplicate-title-descriptions.txt)
- The `title` has been changed on FixedScaleSettings.nodeDeallocationOption from what's published
- IPRuleAction `Allow` description was changed from "Allow access for the matched client IP address" to "Deny client access."
- For the `subnetId` property in `NetworkConfiguration`, the URL was changed from `https://learn.microsoft.com/azure/batch/batch-virtual-network` to `https://learn.microsoft.com/azure/batch/batch-virtual-networks`, resulting in a broken link.
- "Entra" has been renamed back to "Azure Active Directory" in `securityDefinitions`

dpwatrous:
Thanks for the quick fixes. This looks good to me, except that I think we should fix the duplicate titles/descriptions. Having duplicates results in the following behavior when generating API docs:

e42ba618-a81b-46e0-87d0-ae8c716b3bd7

See the file I posted above for the full listing of duplicates.

dpwatrous (approved):

PR #38132, opened 2025-10-13
[widget] Use WidgetManager for JS Template project

maorleger:

## Purpose of this PR

Start using widget manager for JS template project

Related PR: https://github.com/Azure/azure-sdk-for-js/pull/36158/

## Due diligence checklist

## Getting help

### DIFF
--- a/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml
+++ b/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml
@@ -28,9 +28,11 @@ options:
     namespace: "Azure.WidgetAnalytics"
     flavor: azure
   "@azure-tools/typespec-ts":
-    emitter-output-dir: "{output-dir}/{service-dir}/widget-analytics-rest"
+    emitter-output-dir: "{output-dir}/sdk/template/template"
     package-details:
-      name: "@azure-rest/azure-widget-analytics"
+      name: "@azure/template"
+    is-modular-library: true
+    experimental-extensible-enums: true
     flavor: azure
   "@azure-tools/typespec-java":
     emitter-output-dir: "{output-dir}/{service-dir}/azure-widget-analytics"
--- a/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
+++ b/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
@@ -31,10 +31,9 @@ options:
     namespace: Azure.Template.Contoso
     model-namespace: false
   "@azure-tools/typespec-ts":
-    emitter-output-dir: "{output-dir}/sdk/template/template"
-    is-modular-library: true
+    emitter-output-dir: "{output-dir}/{service-dir}/contoso-widgetmanager-rest"
     package-details:
-      name: "@azure/template"
+      name: "@azure-rest/contoso-widgetmanager"
     flavor: azure
   "@azure-tools/typespec-java":
     emitter-output-dir: "{output-dir}/{service-dir}/azure-contoso-widgetmanager"
--- a/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
+++ b/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
@@ -31,9 +31,10 @@ options:
     namespace: Azure.Template.Contoso
     model-namespace: false
   "@azure-tools/typespec-ts":
-    emitter-output-dir: "{output-dir}/{service-dir}/contoso-widgetmanager-rest"
+    emitter-output-dir: "{output-dir}/sdk/template/template"
+    is-modular-library: true
     package-details:
-      name: "@azure-rest/contoso-widgetmanager"
+      name: "@azure/template"
     flavor: azure
   "@azure-tools/typespec-java":
     emitter-output-dir: "{output-dir}/{service-dir}/azure-contoso-widgetmanager"


mikeharder (commented):
@maorleger: So you know, `contosowidgetmanager` is the "old" sample that uses "folder structure v1".  Our "new" sample using "folder structure v2" is `widget`.

https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure
 
Would it make more sense for this PR to target `widget`?

maorleger:
> @maorleger: So you know, `contosowidgetmanager` is the "old" sample that uses "folder structure v1". Our "new" sample using "folder structure v2" is `widget`.
> 
> https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure

Ah thanks for the heads up

> Would it make more sense for this PR to target `widget`?

Yes, if that's the latest - I'll update the PR. Thanks!

maorleger:
Just FYI I added the `PublishToCustomers` label as per the checks; however, I am not sure it was needed? Like I'm not making any spec changes, just updating tspconfig.yaml - wanted to see if the check is incorrectly applied here or if I did something wrong. All good if that's expected and I didn't do anything incorrectly

Run: https://github.com/Azure/azure-rest-api-specs/actions/runs/18472912007

mikeharder:
> Just FYI I added the `PublishToCustomers` label as per the checks; however, I am not sure it was needed? Like I'm not making any spec changes, just updating tspconfig.yaml - wanted to see if the check is incorrectly applied here or if I did something wrong. All good if that's expected and I didn't do anything incorrectly
> 
> Run: https://github.com/Azure/azure-rest-api-specs/actions/runs/18472912007

label is required for any changes to spec files

mikeharder (approved):

specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml
```diff
@@ -28,9 +28,11 @@ options:
     namespace: "Azure.WidgetAnalytics"
     flavor: azure
   "@azure-tools/typespec-ts":
-    emitter-output-dir: "{output-dir}/{service-dir}/widget-analytics-rest"
+    emitter-output-dir: "{output-dir}/sdk/template/template"
     package-details:
-      name: "@azure-rest/azure-widget-analytics"
+      name: "@azure/template"
+    is-modular-library: true
+    experimental-extensible-enums: true
```
maririos, line 35: I remember these properties where recently removed from management plane.
Either way, @maorleger could you make sure you follow https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/1253/Emitter-options-updates so that all the places needed are updated?

what are the implications of these 2 properties not present in Contoso project?
@mikeharder @qiaozha

mikeharder: I don't know anything about these properties in `tspconfig.yaml`

maorleger: > what are the implications of these 2 properties not present in Contoso project?

## is-modular-library

Today, the default JS emitter mode is to generate an [RLC package](https://github.com/Azure/azure-sdk-for-js/blob/main/documentation/RLC-quickstart.md). when `is-modular-library` is set to true, JS emitter will generate a [DPG package](https://github.com/azure/azure-sdk-for-js/blob/main/documentation/DPG-quickstart.md)

As you can see from https://github.com/Azure/azure-sdk-for-js/pull/36079 and https://github.com/Azure/autorest.typescript/issues/3329 the JS team is moving towards preferring DPG over RLC. While we're not quite there yet, we _do_ want to start demonstrating what DPG output would look like.

Eventually the plan is to make modular / DPG the _default_ such that this flag is not needed (or set to false to continue generating RLC)

## experimental-extensible-enums

I don't have context on the original design decisions behind putting this in a flag but was able to find this https://github.com/Azure/autorest.typescript/issues/2420#issuecomment-2166703162 

But the point is that by default we should be continuing to generate extensible enums when necessary. When this flag is set to true, JS will emit the extensible enum pattern similar to this example:

```ts
export type Foo = string
export enum KnownFoo {
   Bar = "Bar"
   Baz = "Baz"
}
```

Which is the desired output. There may be an issue somewhere tracking making this the default in case of non-exhaustive unions. I'll want to check with @joheredi @qiaozha and @MaryGao what the plan is there.

> I remember these properties where recently removed from management plane.
Either way, @maorleger could you make sure you follow https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/1253/Emitter-options-updates so that all the places needed are updated?

Thanks for sharing! I read through this doc and might be missing something obvious but it doesn't _seem_ like it applies to my scenario here? What about this changeset made you believe this doc applies? Just trying to better understand what context I might be missing since I'm not "adding an option or updating an option that is considered required for a language emitter", simply using them in our template project emitter options so that we can showcase what DPG looks like without the overhead of an actual service (and all its idiosyncrasies) 

Hope this helps!

MaryGao: @maririos You are right that these two options are not necessary to set for management plane, but for this widget project it is data-plane SDK and we are in the intermediate status to update our default option for data plane and see the [epic 1](https://github.com/Azure/autorest.typescript/issues/3329) and the [epic 2](https://github.com/Azure/azure-sdk-tools/issues/11690) for whole effort. For Contoso project we also adjust the default config - https://github.com/Azure/azure-rest-api-specs/pull/37805/files.

Once above epics are done I believe we could simply the DPG option also.

/cc @maorleger

PR #38147, opened 2025-10-13
[Monitor] .NET customizations for Metrics

jsquire:
# Summary

The focus of these changes is to customize the .NET client generation.


JoshLove-msft (approved):

jsquire:
Confirmed that Python failures are unrelated to these changes.

PR #38193, opened 2025-10-14
[Storage] Correct instances of "blog" to "blob"

vincenttran-msft:
As title states, we have a handful of instances of "blog" that really should be "blob" 😄


jalauzon-msft (approved):

specification/storage/Microsoft.BlobStorage/suppressions.yaml
```diff
@@ -8,4 +8,11 @@
...
-  reason: 'Suppress Go options until service team is ready to update the go package with typespec'
\ No newline at end of file
+    - options.@azure-tools/typespec-java.emitter-output-dir
+    - options.@azure-tools/typespec-ts.emitter-output-dir
+    - options.@azure-tools/typespec-ts.package-details.name
+    - options.@azure-tools/typespec-python.emitter-output-dir
+    - options.@azure-tools/typespec-csharp.namespace
+    - options.@azure-tools/typespec-csharp.clear-output-folder
+    - options.@azure-tools/typespec-csharp.emitter-output-dir
+  reason: 'Suppress SDK configuration validation until TypeSpec conversion is ready. Go options suppressed until service team is ready to update the go package with typespec.'
```
catalinaperalta, line 18: jfyi you could have just disabled the whole check, like in this example, but you can leave as is: 
image

catalinaperalta (approved):

jalauzon-msft (approved):

PR #38246, opened 2025-10-17
Remove NonSensitivity for DDoS ruleset

baoqihuang0326:
# Choose a PR Template

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
+++ b/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
@@ -3719,7 +3719,6 @@
           "type": "string",
           "description": "The string representation of the web application firewall rule sensitivity.",
           "enum": [
-            "None",
             "Low",
             "Medium",
             "High"
@@ -3727,7 +3726,21 @@
           "x-ms-enum": {
             "name": "ApplicationGatewayWafRuleSensitivityTypes",
             "modelAsString": true
-          }
+          },
+          "values": [
+            {
+              "value": "Low",
+              "description": "Low Sensitivity - triggers the rule by larger spikes in traffic"
+            },
+            {
+              "value": "Medium",
+              "description": "Medium Sensitivity - triggers the rule by moderate spikes in traffic"
+            },
+            {
+              "value": "High",
+              "description": "High Sensitivity - triggers the rule by smaller spikes in traffic"
+            }
+          ]
         },
         "description": {
           "type": "string",
--- a/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
+++ b/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
@@ -3719,6 +3719,7 @@
           "type": "string",
           "description": "The string representation of the web application firewall rule sensitivity.",
           "enum": [
+            "None",
             "Low",
             "Medium",
             "High"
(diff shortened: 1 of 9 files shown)


specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
```diff
@@ -3719,15 +3719,28 @@
...
           "x-ms-enum": {
             "name": "ApplicationGatewayWafRuleSensitivityTypes",
             "modelAsString": true
-          }
+          },
+          "values": [
+            {
+              "value": "Low",
+              "description": "Low Sensitivity - triggers the rule by larger spikes in traffic"
+            },
```
ssevugan21: why do we need this change listing the possible values?

ssevugan21: why do we need this change listing the possible values?

dedhar:
Do we need to update examples https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/examples/WafPolicyCreateOrUpdate.json

dedhar:
Can we check with Prajjwal if we can make Medium as default value

dedhar:
Can you confirm the set of rule sets include Bot and Http DDoS so that we can get CLI fixed when the SDK is released

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
```diff
@@ -3719,15 +3719,28 @@
           "type": "string",
           "description": "The string representation of the web application firewall rule sensitivity.",
           "enum": [
-            "None",
             "Low",
```
ramoka178, line 3723: Why is "None" being removed ? Will this not cause issues to existing customers who used this enum ?

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
baoqihuang0326: This is just the explanation of each enum

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
baoqihuang0326: @ramoka178 None is not available to the customer. Only the DDoS ruleset can be able to configure the Sensitivity field and the feature is private preview only. None of the private preview customers config the 'None' value for their DDoS ruleset Sensitivity field.

ssevugan21 (approved):

AshwiniSanchala:
> Do we need to update examples https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/examples/WafPolicyCreateOrUpdate.json

I think we are okay to not add examples, since one with Block action and High Sensitivity already exist for Httpddosruleset.

specification/network/resource-manager/Microsoft.Network/stable/2025-01-01/applicationGateway.json
baoqihuang0326: Removed

AshwiniSanchala:
Let's see if we can test these changes out. Rest looks good to me.

ssevugan21 (approved):

kamboj-prjwl (approved):

PR #38254, opened 2025-10-17
Remove Creator sections, references and examples from resource manager

stevemunk:
# Choose a PR Template


mikeharder:
This PR removes the "creators" paths, but only from some (not all) API versions.  All stable versions, and the latest preview version, were updated.  However, 7 older previews (from `2020-02-01-preview` to `2024-07-01-preview`) were **not** updated.  This is causing the Avocado error `MISSING_APIS_IN_DEFAULT_TAG`, since it expects all APIs in all versions, to also be in the default tag version.

To maintain consistency, I think this PR should be updated in one of two ways:

1. Remove "creators" paths from older previews
2. Remove older previews completely from the repo

@mikekistler: What do you think?

ramoka178:
> This PR removes the "creators" paths, but only from some (not all) API versions. All stable versions, and the latest preview version, were updated. However, 7 older previews (from `2020-02-01-preview` to `2024-07-01-preview`) were **not** updated. This is causing the Avocado error `MISSING_APIS_IN_DEFAULT_TAG`, since it expects all APIs in all versions, to also be in the default tag version.
> 
> To maintain consistency, I think this PR should be updated in one of two ways:
> 
> 1. Remove "creators" paths from older previews
> 2. Remove older previews completely from the repo
> 
> @mikekistler: What do you think?

Waiting on this response for further review

ramoka178:
Please fix the conflicts too

specification/maps/resource-manager/Microsoft.Maps/Maps/preview/2024-01-01-preview/maps-management.json
```diff
@@ -1556,37 +1242,6 @@
       "x-ms-identifiers": [
         "locationName"
       ]
-    },
```
FarazGIS, line 1559: @stevemunk Please check if this line should be deleted? If it caused the JSON to be invalid, I would assume some check would have flagged it.

specification/maps/resource-manager/Microsoft.Maps/Maps/stable/2021-02-01/maps-management.json
```diff
@@ -477,249 +477,6 @@
           }
         }
       }
-    },
```
FarazGIS, line 480: @stevemunk Please check this deletion.

specification/maps/resource-manager/Microsoft.Maps/Maps/preview/2024-01-01-preview/maps-management.json
stevemunk: Removed that version.

PR #38257, opened 2025-10-18
Operational insights delete old swaggers

catalinaperalta:


catalinaperalta:
files were deleted in another pr

PR #38258, opened 2025-10-19
EKM public facing endpoints

vasanthrajams:
Introducing External Key manager functionality


specification/keyvault/Security.KeyVault.EKM/main.tsp
```diff
@@ -0,0 +1,43 @@
...
+      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
+      scopes: ["https://vault.azure.net/.default"],
+    }
+  ]>
+)
+@service(#{ title: "KeyVaultClient" })
+@versioned(Versions)
+@server(
+  "{vaultBaseUrl}",
+  "The key vault client performs cryptographic key operations and vault operations against the Key Vault service.",
```
heaths, line 28: Product names are proper nouns.
```suggestion
  "The Key Vault client performs cryptographic key operations and vault operations against the Key Vault service.",
```

specification/keyvault/Security.KeyVault.EKM/client.tsp
```diff
@@ -0,0 +1,88 @@
...
+
+@versioned(KeyVault.Versions)
+namespace Customizations;
+
+using KeyVault;
+
+/**
+ * The External Key Manager (EKM) Get operation returns EKM connection. This operation requires ekm/read permission.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Foundations.Operation is necessary for Key Vault"
```
heaths: For older endpoints, yes. Why can't you use them for new endpoints? While it's important to maintain consistency within a service, you weren't far off from what standard operation bases provide in terms of endpoint construction. Did you even try?

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -0,0 +1,85 @@
...
+  host: string;
+
+  /**
+   * Optional path prefix for the EKM proxy (if any).
+   */
+  @encodedName("application/json", "path_prefix")
+  pathPrefix?: string;
+
+  /**
+   * The root CA certificate chain that issued the proxy server's certificate. An array of certificates in the certificate chain, each in DER format and base64 encoded.
```
heaths, line 28: 1. Base64- or base64url-encoded? Most places in KV uses base64url, which is why I ask.
2. You needed to add `@encoded("base64")` (or `@endcoded("base64url")`) to this/these parameters then to make sure the code that is generated is correct.

specification/keyvault/Security.KeyVault.EKM/tspconfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+    module: "github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azekm"
+    module-version: "0.0.1"
+    emitter-output-dir: "{project-root}"
+    override-client-name: "Client"
+    single-client: true
+    inject-spans: true
+    generate-fakes: true
+  "@azure-tools/typespec-rust":
+    crate-name: "azure_security_keyvault_ekm"
+    crate-version: "0.0.1"
```
heaths: We start at `0.1.0`.

heaths (changes requested):

specification/keyvault/Security.KeyVault.EKM/models.tsp
vasanthrajams: Thanks for pointing this out, heath. I have discussed with Babu and Chandan, and we will change this to base64url.

chandanrr: we are going with base64url for consistency

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -0,0 +1,85 @@
...
+  @encodedName("application/json", "subject_common_name")
+  subjectCommonName: string;
+}
+
+/**
+ * EKM proxy information.
+ */
+model EkmProxyInfo {
+  /**
+   * The highest supported proxy interface api version supported by the EKM proxy.
```
chandanrr: duplicate word.. you can say "The highest version of proxy interface API supported by the EKM Proxy"

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -0,0 +1,85 @@
...
+using Azure.Core;
+
+namespace KeyVault;
+
+/**
+ * A EkmConnection model object.
+ */
+model EkmConnection {
+  /**
+   * EKM proxy FQDM (Fully Qualified Domain Name). Only allowed characters are a-z, A-Z, 0-9, hyphen (-), dot (.), and colon (:).
```
chandanrr: FQDN

specification/keyvault/Security.KeyVault.EKM/examples/2026-01-01-preview/CreateEkmConnection-example.json
```diff
@@ -0,0 +1,28 @@
+{
+  "parameters": {
+    "vaultBaseUrl": "https://myvault.vault.azure.net/",
+    "api-version": "2026-01-01-preview",
+    "ekmConnection": {
+      "host": "ekm-proxy.contoso.com",
+      "path_prefix": "/api/v1",
+      "server_ca_certificates": [
+        "MIIDXTCCAkWgAwIBAgIJAKJ5cT5dQnCuMA0GCSqGSIb3DQEBCwUAMEUxCzAJBgNVBAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVQQKDBhJbnRlcm5ldCBXaWRnaXRzIFB0eSBMdGQwHhcNMjQwMTAxMDAwMDAwWhcNMjUwMTAxMDAwMDAwWjBFMQswCQYDVQQGEwJBVTETMBEGA1UECAwKU29tZS1TdGF0ZTEhMB8GA1UECgwYSW50ZXJuZXQgV2lkZ2l0cyBQdHkgTHRkMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA",
```
chandanrr, line 9: Are these valid DER format strings? 
I tried wrapping them in a PEM headers (BEGIN/END CERTIFICATE), but couldn't open the resulting cer file.

specification/keyvault/Security.KeyVault.EKM/examples/2026-01-01-preview/CheckEkmConnection-example.json
```diff
@@ -0,0 +1,19 @@
+{
+  "parameters": {
+    "vaultBaseUrl": "https://myvault.vault.azure.net/",
```
chandanrr: Use managed hsm base urls in these example files.

specification/keyvault/Security.KeyVault.EKM/examples/2026-01-01-preview/DeleteEkmConnection-example.json
```diff
@@ -0,0 +1,20 @@
+{
+  "parameters": {
+    "vaultBaseUrl": "https://myvault.vault.azure.net/",
+    "api-version": "2026-01-01-preview"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "host": "ekm-proxy.contoso.com:8443",
```
chandanrr: remove the port.. just to be consistent with the createekmconnection example.

chandanrr (approved):
I have left minor feedback.

specification/keyvault/Security.KeyVault.EKM/examples/2026-01-01-preview/CreateEkmConnection-example.json
vasanthrajams: This is just a random base64url Chandan. This is not expected to be valid certificate values. These example files are just a representation of how the output looks like.

specification/keyvault/Security.KeyVault.EKM/client.tsp
vasanthrajams: Yes heath. I tried removing the suppress and creating the ekm.json file throws the following error hence brought this back.

image

chandanrr (approved):

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -26,18 +26,24 @@ model EkmConnection {
   pathPrefix?: string;
 
   /**
-   * The root CA certificate chain that issued the proxy server's certificate. An array of certificates in the certificate chain, each in DER format and base64 encoded.
+   * The root CA certificate chain that issued the proxy server's certificate. An array of certificates in the certificate chain, each in DER format and base64url encoded.
```
heaths: You have to actually use the `@encoding` decorator. You can't just document it. That won't do anything for our emitters.

heaths (changes requested):

specification/keyvault/Security.KeyVault.EKM/models.tsp
vasanthrajams: I tried the encode("base64url") heath, but it fails because this encoding can only specified for the bytes type. The service is returning the array of strings.  Could you please suggest if you had come across this scenario in the past any examples?

image

heaths: Because it's not a string. It's a byte encoding. See the other Key Vault TypeSpecs if you need an example. It's used in a lot of places, so just grep for "base64url".

heaths: And if you need an array of them, you can use `bytes[]` I imagine. This can't be merged as-is or generators won't work - or, at least, won't help customers. They'd have to encode first themselves, and what's the point of an SDK then?

vasanthrajams: I tried to change the type to bytes[]. Even with this, it throws `invalid-encode` error. 

image

heaths: See examples in https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs+path%3Aspecification%2Fkeyvault+path%3A*.tsp+base64url&type=code.

You could do something like this in TypeSpec:

```typespec
@encode("base64url")
scalar UrlEncodedBytes extends bytes;

model M {
  chain?: UrlEncodedBytes[],
}
```

@iscai-msft this seem right to you to encoded array items? Does `Array<T>` support a template that would otherwise allow this? I don't want emitters to actually generate a different type. /cc @jhendrixMSFT @antkmsft

iscai-msft: @heaths that is the correct way to use it. I'm not aware of `Array<T>` natively supporting that, so I would say your code snippet is the way the spec should be updated

heaths: Is there any way to avoid the emitter actually emitting a `UrlEncodedBytes` model like in the example above, though? That'd be horribly unnecessary. /cc @jhendrixMSFT

jhendrixMSFT: This should work and we have examples of it in the tsp Spector tests.

https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/encode/bytes/main.tsp#L11-L12

heaths (approved):

heaths:
You'll need to email the REST API board for a second approval. They can then add the appropriate label. See <https://eng.ms/docs/products/azure-developer-experience/design/api-review#requesting-an-offline-rest-api-review>.

specification/keyvault/Security.KeyVault.EKM/client.tsp
weidongxu-microsoft, line 1: This client.tsp is wrong.

If you do not do any customization, you don't need this file.

The least to say, there should be no `@route` or `@get`/`@post` in this file. They define REST API. Here this client.tsp file is intended to define client (based on the REST API defined in routes.tsp).

weidongxu-microsoft: If you are copying this from other KeyVault folders, mind that they do this only for the
`@override`
https://github.com/Azure/azure-rest-api-specs/blob/013b7e122bd44e19ccf9bc4a011509a5991dbafa/specification/keyvault/Security.KeyVault.Keys/client.tsp#L441-L450

Yours does not have any.

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -0,0 +1,94 @@
...
+model EkmConnection {
+  /**
+   * EKM proxy FQDN (Fully Qualified Domain Name). Only allowed characters are a-z, A-Z, 0-9, hyphen (-), dot (.), and colon (:).
+   */
+  host: string;
+
+  /**
+   * Optional path prefix for the EKM proxy (if any).
+   */
+  @encodedName("application/json", "path_prefix")
```
weidongxu-microsoft, line 24: Any reason why these property name does not follow pascalCase?

specification/keyvault/Security.KeyVault.EKM/models.tsp
weidongxu-microsoft, line 1: I didn't see any `@visibility`.

If a model is used as both input and output (request and response), and some property is only produced in response, add `@visibility(LifeCycle.READ)`.

Similarly if some property is not updatable, do similar (i.e., CREATE + READ).

specification/keyvault/Security.KeyVault.EKM/routes.tsp
```diff
@@ -0,0 +1,88 @@
...
+>;
+
+/**
+ * The External Key Manager (EKM) updates the existing EKM connection. If the EKM connection does not exist, this operation fails. This operation requires ekm/write permission.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Foundations.Operation is necessary for Key Vault"
+@summary("Updates the EKM connection.")
+@route("/ekm")
+@patch(#{ implicitOptionality: true })
+op updateEkmConnection is KeyVaultOperation<
```
weidongxu-microsoft, line 68: Is this a JSON Merge Patch ? Or we still follow the other services in keyvault and uses `application/patch` ?

specification/keyvault/Security.KeyVault.EKM/tspconfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+options:
+  "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "data-plane"
+    emitter-output-dir: "{project-root}/.."
+    output-file: "{azure-resource-provider-folder}/Microsoft.KeyVault/{version-status}/{version}/ekm.json"
+  # Uncomment this line and add "@azure-tools/typespec-python" to your package.json to generate Python code
+  "@azure-tools/typespec-python":
+    flavor: azure
+    "basic-setup-py": true
+    "package-name": "azure-keyvault-ekm"
```
weidongxu-microsoft: In EKM doc, it mentions
> MHSM SDK and AZ CLI should be updated to support following actions

Wondering which one is the MHSM SDK?

Ask because here you seem just write a new package name for Python SDK.

specification/keyvault/Security.KeyVault.EKM/routes.tsp
```diff
@@ -0,0 +1,88 @@
...
+namespace KeyVault;
+
+/**
+ * The External Key Manager (EKM) Get operation returns EKM connection. This operation requires ekm/read permission.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Foundations.Operation is necessary for Key Vault"
+@summary("Gets the EKM connection.")
+@route("/ekm")
+@get
+op getEkmConnection is KeyVaultOperation<{}, EkmConnection>;
```
weidongxu-microsoft, line 21: Does the whole service (on that particular endpoint) only have 1 EKM connection at maximum?

What would service do, when user calls `/ekm/create` with a new one, when there is already 1 connection?

specification/keyvault/Security.KeyVault.EKM/routes.tsp
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * The External Key Manager (EKM) Certificate Get operation returns Proxy client certificate. This operation requires ekm/read permission.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Foundations.Operation is necessary for Key Vault"
+@summary("Gets the EKM proxy client certificate.")
+@route("/ekm/certificate")
+@get
+op getEkmCertificate is KeyVaultOperation<{}, EkmProxyClientCertificateInfo>;
```
weidongxu-microsoft, line 30: Similar question on this `/ekm/certificate`. It only has 1 at maximum? And how this is created?

weidongxu-microsoft:
Also, I didn't see REST API on CRUD of external key.
Or it should be in the MHSM SDK? Then it would be at this query https://github.com/Azure/azure-rest-api-specs/pull/38258#discussion_r2488481928

specification/keyvault/Security.KeyVault.EKM/tspconfig.yaml
ncamillucci: Do we want to create a separate SDK for EKM? Backup/Restore, Rbac and Settings are bundled together in Administration SDK. I think EKM can also be added there. Check https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/Security.KeyVault.Administration/tspconfig.yaml

heaths: It would help to understand more about what EKM is for. We choose "administration" for some of the Managed HSM features not because they were only for MHSM, but also because they were sort of a cross between CP and DP - they were, in some sense, management functions.

If EKM shares more in common with key management, it probably makes sense to expose it from the "keys" SDK. We don't want to have an explosion of SDKs because it makes the offering even more confusing for customers.

vasanthrajams: I agree with Heath here. Nicola, EKM is more common with the Keys than Administrator. All EKM calls are limited to the dataplane.

specification/keyvault/Security.KeyVault.EKM/models.tsp
vasanthrajams: MHSM service service side expects in this format hence encoding.

specification/keyvault/Security.KeyVault.EKM/tspconfig.yaml
ncamillucci: Please correct me if I am wrong @heaths , I think to add EKM to Keys SDK we must:
- Add `"specification/keyvault/Security.KeyVault.EKM/"` in the additionalDirectories section of [Security.KeyVault.Keys/tspconfig.yaml](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/Security.KeyVault.Keys/tspconfig.yaml)
- Remove most of the settings from this `tspconfig.yaml` file like it is done in [Security.KeyVault.RBAC/tspconfig.yaml](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/Security.KeyVault.RBAC/tspconfig.yaml) (which is also not a standalone SDK)

specification/keyvault/Security.KeyVault.EKM/client.tsp
vasanthrajams: Sure Weidong, I'll remove the `client.tsp` file as we do not need any customizations for the EKM.

specification/keyvault/Security.KeyVault.EKM/models.tsp
vasanthrajams: Sure. Will include. Thanks for pointing out.

specification/keyvault/Security.KeyVault.EKM/tspconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+  # "@azure-tools/typespec-csharp": true
+  # Uncomment this line and add "@azure-tools/typespec-ts" to your package.json to generate TypeSpec code
+  # "@azure-tools/typespec-ts": true
+  "@azure-tools/typespec-go":
+    service-dir: "sdk/security/keyvault"
+    package-dir: "azadmin/ekm"
+    inject-spans: true
+    single-client: true
+    generate-fakes: true
+    module: "github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azadmin"
```
ncamillucci, line 33: Based on the other comment, I think we should change `azadmin/ekm` to `azkeys/ekm` and `keyvault/azadmin` to `keyvault/azkeys`.
Could you also please add `"specification/keyvault/Security.KeyVault.EKM/"` in the additionalDirectories section of [Security.KeyVault.Keys/tspconfig.yaml](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/Security.KeyVault.Keys/tspconfig.yaml) ?

specification/keyvault/Security.KeyVault.EKM/models.tsp
```diff
@@ -10,9 +10,6 @@ using Azure.Core;
 
 namespace KeyVault;
 
-@encode("base64")
-scalar Base64EncodedBytes extends bytes;
-
```
heaths: You still need this. What we talked about in the review was that you shouldn't use `base64url` encoding which, it seems, you weren't. You still need to specify bytes that are base64-encoded or the SDKs generated define these as a string array and make customers decode each string array item themselves. That's not friendly at all. Basically, undo this change. I thought you changed it to base64url.

heaths: Well, undo almost all this change. Your comment on now-line 29 is *now* correct. It wasn't before.

heaths (changes requested):

specification/keyvault/Security.KeyVault.EKM/models.tsp
vasanthrajams: When use the following `@encode("base64")`, it fails. The base64 scheme is not supported. 

image

Additionally, I looked at other model properties using the base64 format and they do not have encoding("base64").
Please advise.

heaths (approved):

heaths:
REST API board previously signed off barring a few changes I asked for - now merged; however, you have some lintdiff errors you need to fix because that is a required validation.

heaths:
@vasanthrajams why did you get rid of the base64 encoding after I signed off? What I signed off on was correct. This is now broken and you need to restore it.

PR #38368, opened 2025-10-22
Monitor delete old preview swaggers

catalinaperalta:


catalinaperalta:
Unclear if these swaggers should be deleted. Closing for now

PR #38369, opened 2025-10-22
CCF delete old swaggers and tsp

catalinaperalta:


lmazuel (approved):
https://azure.microsoft.com/en-us/updates?id=managed-confidential-consortium-framework-retirement

PR #38370, opened 2025-10-22
App insights delete old swaggers

catalinaperalta:
TODO check if the Monitor.Exporters dir also needs to be deleted. Confirm if readmes are correctly deleted


catalinaperalta:
Service is not retired

PR #38371, opened 2025-10-22
fix(containerregistry): correct operation ids blocking uniform versioning readme use in documentation

CodeJourneys:
This PR fixes inconsistencies in the ContainerRegistry data plane API specification README and corresponding example files to align with unified API guidelines.

Changes:

- Renamed ContainerRegistry_GetProperties to GetRepositoryProperties in latest dataplane stable and preview API versions
- Renamed ContainerRegistry_UpdateProperties to UpdateRepositoryProperties in latest dataplane stable and preview API versions


specification/containerregistry/data-plane/Registry/stable/2021-07-01/containerregistry.json
```diff
@@ -638,8 +638,8 @@
         "tags": [
           "Registry"
         ],
-        "description": "Get repository attributes",
-        "operationId": "ContainerRegistry_GetProperties",
+        "description": "Get repository properties",
+        "operationId": "GetRepositoryProperties",
```
northtyphoon: Is it supposed to be `ContainerRegistry_GetRepositoryProperties`?

specification/containerregistry/data-plane/Registry/stable/2021-07-01/containerregistry.json
```diff
@@ -709,8 +709,8 @@
         "tags": [
           "Registry"
         ],
-        "description": "Update the attribute identified by `name` where `reference` is the name of the repository.",
-        "operationId": "ContainerRegistry_UpdateProperties",
+        "description": "Update repository properties. `reference` is the name of the repository. The attribute identified by `name`.",
+        "operationId": "UpdateRepositoryProperties",
```
northtyphoon: ContainerRegistry_UpdateRepositoryProperties?

specification/containerregistry/data-plane/Registry/stable/2021-07-01/containerregistry.json
CodeJourneys: Yes. Misunderstanding on my part. This should be updated now.

specification/containerregistry/data-plane/Registry/stable/2021-07-01/containerregistry.json
CodeJourneys: Yes. Misunderstanding on my part. This should be updated now.

northtyphoon (approved):
LGTM

lmazuel (approved):
Approving the PR. This is necessary for some doc update. While this is a breaking change and autorest will generate different things, we are not planning to reship a SDK from those Swagger, as they will move to TypeSpec. The TypeSpec migration will require enough work of rewiring the SDK, that the operation name change is trivial in comparison.
Worst case if we really need to re-ship those SDK, we can do directives.

PR #38372, opened 2025-10-22
Mixed reality data-plane delete old swaggers

catalinaperalta:
https://azure.microsoft.com/en-us/updates?id=azure-remote-rendering-retirement


catalinaperalta:
Files were just deleted in #38188

PR #38375, opened 2025-10-23
[oav-runner] skip examples outside stable and preview, handle nested examples

### LINKED ISSUE #38351: [oav-runner] Logs include many false-positive "error getting examples"
This log includes many errors like this:

> Error getting examples for specification/cosmos-db/CosmosDB.Management/examples/2025-05-01-preview/ChaosFaultEnableDisable.json: "/home/runner/work/azure-rest-api-specs/azure-rest-api-specs/specification/cosmos-db/CosmosDB.Management/examples/2025-05-01-preview/ChaosFaultEnableDisable.json" is not a valid JSON Schema

https://github.com/Azure/azure-rest-api-specs/actions/runs/18663458042/job/53209226761#step:4:11

https://github.com/Azure/azure-rest-api-specs/pull/37995

I believe this is a (relatively harmless) bug in oav-runner, where it tries to parse examples as swaggers.  Around this location:

https://github.com/Azure/azure-rest-api-specs/blob/bcacfd106d182630f9a9d1ee4837d1b32da5bc1b/eng/tools/oav-runner/src/runner.ts#L163-L175

`getFiles()` returns all the `*.json` files under the spec folder, including the examples themselves.

We may want to compare with the logic in openapi-alps.

The fix might be as simple as this:

```
 const visibleSwaggerFiles = await getFiles(rootDirectory, swaggerDir).filter(swagger);
```

mikeharder:
- Fixes #38351

Local verification.

## Before

```
$ node eng/tools/oav-runner/cmd/oav-runner.js examples

Error getting examples for specification/cosmos-db/CosmosDB.Management/examples/2025-05-01-preview/ChaosFaultEnableDisable.json: "/home/mharder/specs/specification/cosmos-db/CosmosDB.Management/examples/2025-05-01-preview/ChaosFault
EnableDisable.json" is not a valid JSON Schema
...

oav-runner is checking the following specification rooted files:
- specification/cosmos-db/resource-manager/Microsoft.DocumentDB/preview/2025-05-01-preview/openapi.json

Validation completes without errors.
```

## After (no errors)

```
$ node ../specs-mh/eng/tools/oav-runner/cmd/oav-runner.js examples

oav-runner is checking the following specification rooted files:
- specification/cosmos-db/resource-manager/Microsoft.DocumentDB/preview/2025-05-01-preview/openapi.json

Validation completes without errors.
```


eng/tools/oav-runner/src/runner.ts
```diff
@@ -149,16 +156,17 @@ export async function processFilesToSpecificationList(
...
         examples exist in the same directory as the swagger file that references them:
 
         path/to/swagger/2024-01-01/examples/example.json <-- this is an example file path
         path/to/swagger/2024-01-01/swagger.json <-- we need to identify this file if it references the example
         path/to/swagger/2024-01-01/swagger2.json <-- and do nothing with this one
       */
-      const swaggerDir = path.dirname(path.dirname(file));
+      const swaggerDir = path.relative(rootDirectory, untilFolder(absoluteFilePath, "examples"));
 
       const visibleSwaggerFiles = await getFiles(rootDirectory, swaggerDir);
```
scbedd, line 176: For clarification, you're not trying to clean up noisy errors due to parsing a file as a swagger when it's not in this PR.

mikeharder: Correct.  This PR is fixing exactly 2 problems:

1. The TypeSpec example sources should be ignored (only process examples under "preview" or "stable", which indicates "swagger")
2. Examples in nested folders should find their spec correctly

mikeharder: Easiest way to understand the changes, might be to look at the new unit tests in runner.test.ts

mikeharder: Which reminds me, I should add some positive tests for examples under "preview".

eng/tools/oav-runner/src/runner.ts
```diff
@@ -149,16 +156,20 @@ export async function processFilesToSpecificationList(
...
+    if (example(file) && (preview(file) || stable(file))) {
       /*
         examples exist in the same directory as the swagger file that references them:
 
         path/to/swagger/2024-01-01/examples/example.json <-- this is an example file path
         path/to/swagger/2024-01-01/swagger.json <-- we need to identify this file if it references the example
         path/to/swagger/2024-01-01/swagger2.json <-- and do nothing with this one
       */
-      const swaggerDir = path.dirname(path.dirname(file));
+      const swaggerDir = path.relative(
```
mikeharder, line 171: Instead of hardcoding `dirname(dirname())`, we now use `untilLastSegment("examples")`, which should always return the path until (but excluding) the last occurrence of "examples".  Even if the example is nested arbitrarily deep under "examples".

`untilLastSegment()` always returns an absolute path, so it needs to be resolved against `rootDirectory` to get back to a relative path, which is expected by later code.  In a later PR, this code will be updated to use absolute paths as much as possible, per our design guidelines.

weshaggard (approved):

eng/tools/oav-runner/src/runner.ts
```diff
@@ -149,16 +156,20 @@ export async function processFilesToSpecificationList(
   for (const file of files) {
     const absoluteFilePath = path.join(rootDirectory, file);
 
-    // if the file is an example, we need to find the swagger file that references it
-    if (example(file)) {
+    // if the file is an example, under "preview" or "stable" (but not the TypeSpec source folder),
+    // we need to find the swagger file that references it
+    if (example(file) && (preview(file) || stable(file))) {
       /*
         examples exist in the same directory as the swagger file that references them:
```
scbedd:
```suggestion
        The `examples` folder is traditionally populated with example.json files related to the owning spec.
        However, these examples might be contained within subdirectories of the examples folder.
```

scbedd: nit: comment correctness after your change.

eng/tools/oav-runner/src/runner.ts
```diff
@@ -149,16 +156,20 @@ export async function processFilesToSpecificationList(
...
 
         path/to/swagger/2024-01-01/examples/example.json <-- this is an example file path
```
scbedd:
```suggestion
        path/to/swagger/2024-01-01/examples/example.json                        <-- valid example path
        path/to/swagger/2024-01-01/examples/subdirectory1/example.json <-- also valid example path
```

scbedd (approved):

scbedd (approved):

PR #38403, opened 2025-10-24
[path] Add includesSegment() and untilLastSegment()

mikeharder:
Will be used by #38375


.github/shared/src/path.js
```diff
@@ -1,13 +1,37 @@
 // @ts-check
 
-import { resolve, sep } from "path";
+import { basename, dirname, resolve } from "path";
```
mikeharder, line 3: Safer to use path functions `basename()` and `dirname()` compared to `string.split(sep)`.

.github/shared/src/path.js
```diff
@@ -1,13 +1,37 @@
...
 
 /**
  *
  * @param {string} path
- * @param {string} folder
- * @returns {boolean} True if path contains the named folder
+ * @param {string} segment
```
mikeharder: Cannot distinguish file vs folder without using `fs` (which we don't want), so allow both

.github/shared/test/path.test.js
```diff
@@ -1,46 +1,52 @@
...
-import { existsSync, mkdirSync, rmSync } from "fs";
-import { dirname, join } from "path";
-import { fileURLToPath } from "url";
-import { afterEach, beforeEach, describe, it } from "vitest";
-import { includesFolder } from "../src/path.js";
+import { basename, resolve, sep } from "path";
+import { describe, expect, it } from "vitest";
+import { includesSegment, untilLastSegment } from "../src/path.js";
 
-// Get the directory of this test file
```
mikeharder, line 10: removed dead code

scbedd (approved):

PR #38448, opened 2025-10-27
[mongocluster] Add armstong terraform scenario tests

olivertowers:
Adding armstrong terraform scenario tests for mongocluster api-version 2025-09-01.

No API spec changes.

- Add terraform folder for mongocluster 2025-09-01
  - specification/mongocluster/resource-manager/Microsoft.DocumentDB/MongoCluster/stable/2025-09-01/terraform

- Include testing.tf, dependency.tf, provider.tf files used for armstrong testing
- Include armstrong-scenario.sh wrapper script to with installing requirements and running armstrong commands.
- Include README.md for instructions and usage.


javierfigueroa (approved):

razvanbadea-msft:
why are these scenario tests added in this repo? 
you can rename the added readme.md to something else to avoid Avocado required check to fail.
Fix spellcheck errors

olivertowers:
@razvanbadea-msft Guidelines used to be to add scenario tests inline with the swagger repo.
If that is no longer the case, what is the guidance?

Since this is terraform specific, I'm asking for guidance from the Toolset team as well, but so far armstrong terraform validation was fairly fluent for building scenario test (easier to use compared with prior tools at least).

PR #38494, opened 2025-10-29
Copy ledger swaggers to new dir

catalinaperalta:


lmazuel:
We're moving a file in two steps, to  be sure the doc can be updated to the new position before deleting the old file. See also:
https://github.com/Azure/azure-rest-api-specs/pull/38495

lmazuel (approved):

PR #38533, opened 2025-10-31
[.github] Remove "// @ts-check" from all JS files

mikeharder:
- Unnecessary since tsconfig.json includes "checkJs:true"
- https://www.typescriptlang.org/tsconfig/#checkJs

### DIFF
--- a/.github/shared/src/array.js
+++ b/.github/shared/src/array.js
@@ -1,5 +1,3 @@
-// @ts-check
-
 /**
  * @template T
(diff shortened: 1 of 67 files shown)


scbedd (approved):
[TIL](https://github.com/Azure/azure-rest-api-specs/blob/2dec247f562ceb21e1d4978b1fe3ebeaa9efaf27/.github/tsconfig.json#L6)

PR #38539, opened 2025-10-31
[Content Validation] DFS API updates for structured body encoding.

RandomCyclist:
Update Storage DFS API docs for Content Validation changes.

This PR adds documentation for the x-ms-structured-body and x-ms-structured-content-length to the storage DFS API documentation. A similar PR has been created to update the Blob and File API documentation.


seanmcc-msft (approved):

seanmcc-msft:
@RandomCyclist, can we update these request headers to include the min supported x-ms-version?  it should be `2025-01-05`.

RandomCyclist:
> @RandomCyclist, can we update these request headers to include the min supported x-ms-version? it should be `2025-01-05`.

Done. Let me know if you would like additional rewording.

seanmcc-msft (approved):

PR #38557, opened 2025-11-02
Adding optional Key Attributes parameters for the Azure Managed HSM keys

vasanthrajams:
This PR adds two optional key attribute parameters: one for handling external key management keys, and another for returning the AES key size.


heaths (approved):

vickm (approved):

specification/keyvault/Security.KeyVault.Keys/models.tsp
```diff
@@ -411,7 +411,7 @@ model ExternalKey {
   /**
    * The external key identifier. The valid id can only contain characters in the set [a-zA-Z0-9-]. Maximum length is 64 characters.
    */
-  id: string;
+  id?: string;
```
heaths: This seems wrong. I think you want the field of type `ExternalKey` to be nullable. If there is an external key, `id` should be required, I think. Not only is it an id, it's the only field. Are you saying `"key": {}` (or whatever the field name is) is valid?

heaths (changes requested):

specification/keyvault/Security.KeyVault.Keys/models.tsp
vasanthrajams: ExternalKey is optional. Howver, if the ExternalKey is present, it must have an id. 

When I used that typespec, it outputted the json containing the *Update suffix to the KeyAttribute model and ExternalKey. I posted regarding this issue in the Typespec discussion teams channel. I'll share with you on teams.

vasanthrajams: As per our discussion offline, I have made the `id` non-opitional.

heaths (approved):

heaths (approved):

PR #38574, opened 2025-11-03
using customPatch with  patch operations for contoso

kazrael2119:
fixes https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5517294 from https://github.com/Azure/azure-rest-api-specs/pull/38519


kazrael2119:
close as no need

PR #38591, opened 2025-11-04
[.github] type-check tests

mikeharder:


.github/shared/src/exec.js
```diff
@@ -36,7 +36,7 @@ export function isExecError(error) {
  * Wraps `child_process.execFile()`, adding logging and a larger default maxBuffer.
  *
  * @param {string} file
- * @param {string[]} args
+ * @param {string[]} [args]
```
mikeharder, line 39: wrapped API also allows undefined

scbedd (approved):

PR #38609, opened 2025-11-04
Update with the configuration to run go generate after tsp compile.

richardpark-msft:
There are some post-generation steps that run after tsp generation, to create the system events enum. Adding in this switch ensures we generate the same client in all of our systems.

NOTE: We want https://github.com/Azure/azure-sdk-for-go/pull/25555 to go in _first_, so we don't end up using tools that don't exist in the pipeline.


RickWinter (approved):

richardpark-msft:
Just FYI, I believe the pipeline error here is related to the 'go generate' commands in my package using things that weren't in the environment. I've got a PR out in azure-sdk-for-go, so once that merges we can just re-run CI (shouldn't require any changes on this end).

PR #38632, opened 2025-11-05
wip - simplify tsp

catalinaperalta:


specification/ai-foundry/data-plane/Azure.AI.Projects/.clients/Azure.AI.Agents.V2.csharp/client.tsp
```diff
@@ -3,8 +3,12 @@ import "./main.tsp";
 
 using Azure.ClientGenerator.Core;
 
-@clientNamespace("Azure.AI.Agents")
-namespace Azure.AI.Projects;
```
jpalvarezl: Should we restore this for languages which are not java and csharp?

catalinaperalta: Wasnt this original directory only used for C#? The new directories for the Agents and Projects sdks should be used for C# and Java and the main typespec files should still be used for python and JS. I dont believe I changed this in the main tsp files

catalinaperalta: The diff is also broken now because I pulled in main and the target branches havent done so yet

catalinaperalta:
Closing since the sdk generation improvements were merged into the base branch in the private repo and the v2 structure updates will be made separately later on.

PR #38644, opened 2025-11-06
Add AddtionalProperties for DataConnector

LaylaLiu-gmail:
# Choose a PR Template

### DIFF
--- a/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
+++ b/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
@@ -1643,9 +1643,7 @@
         "extendedProperties": {
           "type": "object",
           "description": "Additional properties for the data connector which can be used to store custom key-value pairs.",
-          "additionalProperties": {
-            "type": "object"
-          }
+          "additionalProperties": {}
         }
       }
     }
--- a/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
+++ b/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
@@ -1641,25 +1641,11 @@
           }
         },
         "extendedProperties": {
-          "type": "array",
-          "description": "Additional properties for the data connector.",
-          "items": {
-            "$ref": "#/definitions/ExtendedProperties"
+          "type": "object",
+          "description": "Additional properties for the data connector which can be used to store custom key-value pairs.",
+          "additionalProperties": {
+            "type": "object"
           }
-        }
-      }
-    },
-    "ExtendedProperties": {
-      "type": "object",
-      "description": "Additional properties for the data connector.",
-      "properties": {
-        "name": {
-          "description": "Name of the additional property.",
-          "type": "string"
-        },
-        "value": {
-          "description": "Value of the additional property.",
-          "type": "object"
         }
       }
     }
--- a/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
+++ b/specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
@@ -1639,6 +1639,27 @@
             "name": "DataConnectorProvisioningState",
             "modelAsString": true
           }
+        },
+        "extendedProperties": {
+          "type": "array",
+          "description": "Additional properties for the data connector.",
+          "items": {
+            "$ref": "#/definitions/ExtendedProperties"
+          }
+        }
+      }
+    },
+    "ExtendedProperties": {
+      "type": "object",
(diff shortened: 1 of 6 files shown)


specification/app/resource-manager/Microsoft.App/ContainerApps/stable/2026-01-01/CommonDefinitions.json
```diff
@@ -1639,6 +1639,27 @@
             "name": "DataConnectorProvisioningState",
             "modelAsString": true
           }
+        },
+        "extendedProperties": {
```
gary-x-li, line 1643: This is a collection of key value pairs without concrete schema. Also, it lacks enough description and examples. What are these key value pairs? Can you define proper schema for them? #Resolved

LaylaLiu-gmail: @gary-x-li, actually we need a property who is a Dictionary<string, object>. but addtionalProperties is not recommended by swagger. Is there any other way to define such a dictionary in swagger?

najian: More explanation:
1.  extendedproperties in Data connector represent various additional option highly dependent on the specific data connector being used. It's common industrial practice to make the data connector protocol extensible, like JDBC, ODBC...
2. all the elements to be treated as an atomic unit, non-ordered.
3. the property set represents a simplistic array with a very small number of elements that are not expected to grow drastically over time and would never have paging requirements.

gary-x-li: Can you please do/answer the followings?

1. Add or modify example files to show what extended properties look like for at least two different data connectors. That will make it clearer to the reviewers and the customers.
2. How would customers know what the valid set of extended properties for each data connector are? Are they widely known as industrial practice? Consider expanding the description to give customers better idea.
3. We do allow additionalProperties if the properties are simply passing through your service and you don't control nor validate them. (e.g. tags for Azure resources or labels for Kubernetes). If that's your case, please do use additionalProperties and suppress the validation error. We will approve the suppression if it's justified.

LaylaLiu-gmail: Have updated the example files and this is a pass-through property. our RP doesn't do any validation on this property. So, I just changed it to dictionary for better understanding.

sandipsh: 3rd point is not yet addressed.

gary-x-li: In addition, I also don't see 1st point (adding example files) and 2nd point (change description to indicate where to get the valid set of properties).

LaylaLiu-gmail: sorry, I pushed the latest change to another branch. Now it has the latest change. but the validation shows "the type is invalid. expected object, but found string, array".  Actually, we cannot specify the value type, the value could be string or array. That's why we used object.  @gary-x-li , do you know how to solve such issue?

gary-x-li: @LaylaLiu-gmail I'm not sure. You can post your question in [this channel](https://teams.microsoft.com/l/channel/19%3A0351f5f9404446e4b4fd4eaf2c27448d%40thread.skype/API%20Spec%20Review?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47); or search similar examples in the repo.

LaylaLiu-gmail: @gary-x-li , I got the answer from [this channel](https://teams.microsoft.com/l/message/19:0351f5f9404446e4b4fd4eaf2c27448d@thread.skype/1763348990635?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1763348990635&teamName=Azure%20SDK&channelName=API%20Spec%20Review&createdTime=1763348990635), and updated the PR, could you please help review again?

ravimeda: Resolving based on Mark's response.

mikeharder:
@ravimeda: I believe you need to review the suppressions as part of the ARM review, and add label `Approved-Suppression` in addition to `ARMSignedOff`.

ravimeda (approved):

Juliehzl (approved):

PR #38657, opened 2025-11-06
Blob storage fix param order

catalinaperalta:


nickliu-msft (approved):

catalinaperalta:
There are no sdks that have GA'd with this TypeSpec, we're fixing the parameter order to ensure this issue doesnt come up again or affect future sdk generation.

PR #38664, opened 2025-11-07
[summarize-checks] Fix detection of "RPaaSException" label

### LINKED ISSUE #38659: [summarize-checks] Cannot add `ARMSignedOff` if `CI-NewRPNamespaceWithoutRPaaS` and `RPaaSException`
This PR has label `CI-NewRPNamespaceWithoutRPaaS`, which was correctly added to block the PR.  It also has label `RPaaSException`, which should neutralize the previous label, and allow adding `ARMSignedOff`.

However, when label `ARMSignedOff` is added, it's removed by automation.  This appears to be a bug, with not incorporating `RPaaSException` correctly.

- https://github.com/Azure/azure-rest-api-specs/pull/38629

mikeharder:
- fixes #38659


.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -504,7 +504,7 @@ export async function processImpactAssessment(labelContext, impactAssessment) {
   );
   ciNewRPNamespaceWithoutRpaaSLabel.shouldBePresent = impactAssessment.rpaasRPMissing || false;
 
-  const rpaasExceptionLabelPresent = "RPaaSException" in labelContext.present;
```
mikeharder, line 507: `"string" in Set` always returns false.  We want `Set.has("string")`.

PR #38685, opened 2025-11-10
Container Registry to TSP Migration

kashifkhan:
# Choose a PR Template


specification/containerregistry/data-plane/Registry/client.tsp
```diff
@@ -0,0 +1,7 @@
+import "./main.tsp";
+
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+@@clientName(ContainerRegistryService, "ContainerRegistry");
```
catalinaperalta: Nit: why cant the namespace just be named ContainerRegistry?

specification/containerregistry/data-plane/Registry/main.tsp
```diff
@@ -0,0 +1,48 @@
...
+  {
+    endpoint: url,
+  }
+)
+@useAuth(
+  OAuth2Auth<[
+    {
+      type: OAuth2FlowType.implicit,
+      authorizationUrl: "https://login.microsoftonline.com/common/oauth2/authorize",
+      scopes: [],
```
catalinaperalta: There isnt a container registry scope?

specification/containerregistry/data-plane/Registry/models.tsp
```diff
@@ -0,0 +1,1275 @@
...
+ * Properties of this repository.
+ */
+model ContainerRepositoryProperties {
+  /**
+   * Registry login server name. This is likely to be similar to
+   * {registry-name}.azurecr.io.
+   */
+  @visibility(Lifecycle.Read)
+  @encodedName("application/json", "registry")
+  registryLoginServer: string;
```
catalinaperalta: I see a lot of encodedName usage, are you giving preference to the client names? Seems like we should be doing the opposite in the main spec files, which is to give preference to the service definition and then have client renames only in client.tsp.

catalinaperalta: I usually just use encodedName when the actual property name isnt very typespec-friendly and there is a good alternative to use in the typespec definition. So for example if the actual service property was called "registry_login_server", then encodedName would be good, but "registry" is good property name so we dont need to use encodedName here. If "registryLoginServer" is the preferred name by sdks, then that should be renamed in client.tsp.

catalinaperalta: Same advice applies to other instances of encodedName

catalinaperalta (commented):
Have you taken a look through the breaking changes reported? Seems like there are some request content types missing, some paths were removed, there's a new response code, removed headers, etc. Would be good to go through the changes reported and fix any of the legitimate reports.

specification/containerregistry/data-plane/Registry/models.tsp
```diff
@@ -0,0 +1,1275 @@
...
+  refreshToken?: string;
+}
+
+/** The ACR access token response containing the access token for authentication. */
+model AcrAccessToken {
+  /**
+   * The access token for performing authenticated requests
+   */
+  @encodedName("application/json", "access_token")
+  accessToken?: string;
```
catalinaperalta, line 1020: This is an example of good encodedName usage

specification/containerregistry/data-plane/Registry/main.tsp
kashifkhan: They didnt have one specified, they build it in code. Ill use that one here

specification/containerregistry/data-plane/Registry/models.tsp
kashifkhan: No, I was just trying to get it to match the wire name like this https://github.com/Azure/azure-sdk-for-python/blob/a738f71c5b2ddd00b6682b93995ae7f5b5f0ee4e/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py#L114

the client name was `registryLoginServer` but the wire name was different https://github.com/Azure/azure-rest-api-specs/blob/0ab31eba573506650184b893350afab23d2a2071/specification/containerregistry/data-plane/Registry/stable/2021-07-01/containerregistry.json#L1846

specification/containerregistry/data-plane/Registry/models.tsp
catalinaperalta: This should be in client.tsp

catalinaperalta: Same feedback as above for any other instances of tcgc decorators used outside of client.tsp

specification/containerregistry/data-plane/Registry/routes.tsp
catalinaperalta: Nit since you're using the http library notfoundresponse you can also use the accepted response
```suggestion
    AcceptedResponse | NotFoundResponse
```

specification/containerregistry/data-plane/Registry/tspconfig.yaml
catalinaperalta: Nit: this makes it so you dont have to hardcode the Registry/ dir
```suggestion
    emitter-output-dir: "{project-root}"
    examples-dir: "{project-root}/examples"
    output-file: "{version-status}/{version}/containerregistry.json"
```

catalinaperalta (approved):
looks good, just a few nitpicks and clean up

lmazuel (approved):

PR #38706, opened 2025-11-11
add directive for paloatlonetwork to avoid breakings

kazrael2119:
https://github.com/Azure/azure-sdk-for-js/pull/36389#discussion_r2509160508
add directive for paloatlonetwork to avoid breakings

### DIFF
--- a/specification/paloaltonetworks/resource-manager/readme.typescript.md
+++ b/specification/paloaltonetworks/resource-manager/readme.typescript.md
@@ -24,4 +24,19 @@ directive:
           }
         }
       }
+  - from: PaloAltoNetworks.Cloudngfw.json
+    where-operation: GlobalRulestack_commit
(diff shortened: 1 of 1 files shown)


specification/paloaltonetworks/resource-manager/readme.typescript.md
```diff
@@ -24,4 +24,19 @@ directive:
           }
         }
       }
+  - from: PaloAltoNetworks.Cloudngfw.json
```
MaryGao, line 27: @josefree will this require ARM review? this change only happens in typescript scope for client generation.

MaryGao:
This would only impact the JS client generation and no REST impacted. And we have agreed that this should not be included in ARM review but tool improvement is still in progress.

MaryGao (approved):

PR #38715, opened 2025-11-11
Merge GalleryRP 2025-03-03 Features into Main

ookoka:
# Choose a PR Template

### DIFF
--- a/specification/compute/Gallery.Management/models.tsp
+++ b/specification/compute/Gallery.Management/models.tsp
@@ -149,6 +149,24 @@ union ReplicationMode {
 }
 
 /**
+ * Specifies the strategy to be used when selecting the storage account type. Cannot be specified along with storageAccountType, but can be overridden per region by specifying targetRegions[].storageAccountType. This property is not updatable.
+ */
+@added(Versions.v2025_03_03)
+union StorageAccountStrategy {
+  string,
+
+  /**
+   *  Choose Standard_ZRS storage if the region supports it, else choose Standard_LRS storage, unless overridden by specifying regional storageAccountType. If no storageAccountStrategy is specified, this is the default strategy (from API version 2025-03-03 onwards).
+   */
+  PreferStandard_ZRS: "PreferStandard_ZRS",
+
+  /**
+   *  Choose Standard_LRS storage unless overridden by specifying regional storageAccountType.
+   */
+  DefaultStandard_LRS: "DefaultStandard_LRS",
+}
+
+/**
  * It is type of the extended location.
  */
 union GalleryExtendedLocationType {
@@ -949,11 +967,10 @@ model GalleryArtifactPublishingProfileBase {
   /**
    * The end of life date of the gallery image version. This property can be used for decommissioning purposes. This property is updatable.
    */
-  // FIXME: (utcDateTime) Please double check that this is the correct type for your scenario.
   endOfLifeDate?: utcDateTime;
 
   /**
-   * Specifies the storage account type to be used to store the image. This property is not updatable.
+   * Specifies the storage account type to be used to store the image. Cannot be specified along with storageAccountStrategy. This property is not updatable.
    */
   storageAccountType?: StorageAccountType;
 
@@ -967,6 +984,12 @@ model GalleryArtifactPublishingProfileBase {
    */
   @identifiers(#["name", "extendedLocation/name"])
   targetExtendedLocations?: GalleryTargetExtendedLocation[];
+
+  /**
+   * Specifies the strategy to be used when selecting the storage account type. Cannot be specified along with storageAccountType, but can be overridden per region by specifying targetRegions[].storageAccountType. This property is not updatable.
+   */
+  @added(Versions.v2025_03_03)
+  storageAccountStrategy?: StorageAccountStrategy;
 }
 
 /**
--- a/specification/compute/Gallery.Management/models.tsp
+++ b/specification/compute/Gallery.Management/models.tsp
@@ -46,6 +46,39 @@ union GallerySharingPermissionTypes {
 }
 
 /**
+ * Specifies the type of the Gallery Script parameter. Possible values are: String, Int, Double, Boolean, Enum
+ */
+@added(Versions.v2025_03_03)
+union GalleryScriptParameterType {
+  string,
+
+  /**
+   * String gallery script parameter type
+   */
+  String: "String",
+
+  /**
+   * Int gallery script parameter type
+   */
+  Int: "Int",
+
+  /**
+   * Double gallery script parameter type
+   */
+  Double: "Double",
+
+  /**
+   * Boolean gallery script parameter type
+   */
+  Boolean: "Boolean",
+
+  /**
+   * Enum gallery script parameter type
+   */
+  Enum: "Enum",
+}
+
+/**
  * This property allows you to specify the type of sharing group. Possible values are: **Subscriptions,** **AADTenants.**
  */
 union SharingProfileGroupTypes {
@@ -1766,6 +1799,12 @@ model GalleryImageVersionList is Azure.Core.Page<GalleryImageVersion>;
 model GalleryApplicationList is Azure.Core.Page<GalleryApplication>;
 
 /**
+ * The List Gallery Script operation response.
+ */
+@added(Versions.v2025_03_03)
+model GalleryScriptList is Azure.Core.Page<GalleryScript>;
+
+/**
  * The List Gallery Application version operation response.
  */
 model GalleryApplicationVersionList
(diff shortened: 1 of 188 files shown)


ookoka:
**Adding Approved-Avocado as we have for other features because Compute is not ready to split its SDK yet, causing the avocado check to fail. Janine Zhang's team is currently looking into this for the future.**

ravimeda:
Please address the errors in required checks e.g., Swagger LintDiff.

ookoka:
@ravimeda is it possible to get SwaggerLint Check suppression for this pr? It is being generated because of the missing namePattern enforcement for the 2025-03-03 version, but after speaking with breaking change board, this change should not be added to the typespec or it would cause breaking change for previous versions' resources, and was not enforced in previous versions.

specification/compute/resource-manager/readme.md
```diff
@@ -360,19 +360,10 @@ input-file:
   - Microsoft.Compute/ComputeRP/stable/2025-04-01/ComputeRP.json
   - Microsoft.Compute/DiskRP/stable/2025-01-02/DiskRP.json
   - Microsoft.Compute/Skus/stable/2021-07-01/skus.json
-  - Microsoft.Compute/GalleryRP/stable/2024-03-03/GalleryRP.json
+  - Microsoft.Compute/GalleryRP/stable/2025-03-03/GalleryRP.json
```
mikeharder: @maririos, @mikekistler: Should PRs be allowed to change swagger versions in an existing tag?  Rather than adding a new tag with the new swagger version?  I'm not sure if our BreakingChanges checks detects this, since they operate on swagger **files**, not readme.md **api-versions**.

mikeharder (dismissed):
Blocking until approval on changing swagger in existing tag

specification/compute/resource-manager/readme.md
mikekistler: Short answer is that this should not be allowed, and if our breaking change checks don't catch it then they should be updated so that they do.

mikeharder: I opened an issue for tracking, but this might be a large feature request, since I believe the breaking change checks currently work on the level of **individual swagger files**, not **api versions which contain mulitple swagger files**.  Unless you have an idea, how this feature could be added cheaply to the existing checks.

https://github.com/Azure/azure-rest-api-specs/issues/39067

Also, check "Swagger Avocado" does flag this PR with `MULTIPLE_API_VERSION`. Once this spec is migrated to "folder structure v2", the problem may go away, since API versions should align with swagger versions.  One of the root causes, is API version "package-2025-04-01" actually includes swaggers from different services from different versions.

ookoka: @mikeharder I went ahead and resolved the changes in the readme, In the original PR submitted there was the correct version and tag but I believe during a merge conflict things got wonky. As for the Swagger Avocado, I left a comment earlier but currently approved-avocado needs to be manually appended because Compute SDK is not ready for a split yet:

"Adding Approved-Avocado as we have for other features because Compute is not ready to split its SDK yet, causing the avocado check to fail. Janine Zhang's team is currently looking into this for the future."

ookoka:
@ravimeda Pinging again for SwaggerLintDiff Check Suppression and ARM Re-review. Breaking changes have been addressed. Thanks

specification/compute/Gallery.Management/models.tsp
```diff
@@ -1786,6 +1850,195 @@ model GallerySoftDeletedResourceProperties {
...
+model GenericGalleryParameter {
+  /**
+   * The name of the parameter.
+   */
+  name: string;
+
+  /**
+   * Indicates whether this parameter must be passed.
+   */
+  required?: boolean;
```
ramoka178, line 2015: >boolean [](http://example.com/codeflow?start=13&length=7)

ARM recommends enums over booleans for future proof APIs. 

 

Standard guidance is: replace boolean/switch properties with a more meaningful enum whenever possible. 

A boolean will forever have two valid values (true or false). A string enum type is always preferred. Also, properties should always provide better values just than True and False. For example two switches "isTypeA" and "isTypeB" should be replaced with one enum "type": [A, B, DefaultType]. Enums are always a more flexible and future proof option because they allow additional values to be added in the future in a non-breaking way, e.g. [Enabled, Disabled, Suspended, Deallocated]. 

Note: do NOT define a 'boolean enum' with two values 'True and False'. This might be easier to 'extend' in terms of types, but semantically its cofusing, and no better than a boolean.  #Resolved

specification/compute/resource-manager/readme.md
```diff
@@ -349,28 +349,46 @@ suppressions:
     reason: ScaleOut operation returns both 200 and 202, but 200 will not return schema. This is a common pattern for VMSS action operations. 
     from: ComputeRP.json
     where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachineScaleSets/{vmScaleSetName}/scaleOut"].post
+  - code: ResourceNameRestriction
+    reason: Gallery resources follow ARM resource naming requirements; no additional gallery name pattern is enforced beyond ARM.
+    from: GalleryRP.json
```
ramoka178: could you add some context on this. sorry I do not follow. #Resolved

specification/compute/Gallery.Management/models.tsp
```diff
@@ -1786,6 +1850,195 @@ model GallerySoftDeletedResourceProperties {
...
+model GalleryScriptParameter extends GenericGalleryParameter {
+  /**
+   * Specifies the type of the Gallery Script parameter. Possible values are: String, Int, Double, Boolean, Enum
+   */
+  type?: GalleryScriptParameterType;
+
+  /**
+   * The minimum value of parameter.
+   */
+  minValue?: string;
```
ramoka178, line 1989: >minValue [](http://example.com/codeflow?start=2&length=8)

use full string, minimumValue #Resolved

specification/compute/Gallery.Management/models.tsp
```diff
@@ -1786,6 +1850,195 @@ model GallerySoftDeletedResourceProperties {
...
+
+  /**
+   * The minimum value of parameter.
+   */
+  maxValue?: string;
```
ramoka178, line 1994: >maxValue [](http://example.com/codeflow?start=2&length=8)

use full string, maximumValue #Resolved

specification/compute/resource-manager/readme.md
ookoka: Essentially any name pattern enforcement required by the SwaggerLintDiff Check for 2025-03-03 version would cause a breaking change for existing resources in the previous version. The name pattern is not something enforced by Gallery and siloing that change to current version would require a complete model versioning in the typespec rather than solely the parameter

specification/compute/Gallery.Management/models.tsp
ookoka: Would this not cause breaking change for clients with existing resources expecting a property "minValue"?

specification/compute/Gallery.Management/models.tsp
ookoka: Would this not cause breaking change for clients with existing resources expecting a property "maxValue"?

specification/compute/Gallery.Management/models.tsp
ookoka: This suggestion was responded to in the previous PR of the feature by the feature owner:

"This also has been around since many years. Unrelated to Gallery script and Gallery script version changes"

specification/compute/resource-manager/readme.md
ramoka178: So, are there no pattern restrictions today for a galleryname? can user give a name like '!!!#$%^' and that is still valid ?

I believe the expectation is to provide what pattern you have. This way, if there is any change later on in the future, that can be boiled up.

specification/compute/Gallery.Management/models.tsp
ramoka178: What is the previous PR here ? this is a new value coming new in the PR right ? 

New API versions / APIs are an opportunity to correct patterns that are not ideal. If we continue carrying forward non-compliant conventions into new APIs and endpoints, the issues will persist indefinitely.

Please give it another thought.

specification/compute/Gallery.Management/models.tsp
ramoka178: This seems to be a new property being introduced in this PR right. Why will existing clients break this ?

specification/compute/Gallery.Management/models.tsp
ramoka178: This seems to be a new property being introduced in this PR right. Why will existing clients break this ?

ookoka: Apologies, makes sense, will fix!

specification/compute/Gallery.Management/models.tsp
ookoka: This has already been merged into the service side and customers are expecting this as a boolean.

specification/compute/resource-manager/readme.md
ookoka: There is no pattern restrictions for Galleryname as far as I'm aware. @viveklingaiah Can you please confirm this?

specification/compute/Gallery.Management/models.tsp
ookoka: This has already been deployed to service side, @viveklingaiah can you confirm this?

specification/compute/Gallery.Management/models.tsp
ookoka: This has already been deployed to service side, @viveklingaiah can you confirm this?

specification/compute/resource-manager/readme.md
viveklingaiah: This is the regex pattern used currently for galleryName @"^[^_\W][\w-._]{0,79}(?<![-.])$"


        protected static readonly Regex EntityIdRegEx =
            new Regex(@"^[^_\W][\w-._]{0," + (RPApiConstants.MaxResourceNameLength-1) + @"}(?<![-.])$", RegexOptions.Compiled);
https://msazure.visualstudio.com/DefaultCollection/One/_git/Compute-CPlat-Core?path=/src/Shared/Lib/ResourceProvider.Service/API/RequestValidator.cs&version=GBmaster&line=60&lineEnd=61&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

specification/compute/Gallery.Management/models.tsp
viveklingaiah: Yes, already deployed

specification/compute/resource-manager/readme.md
viveklingaiah: same pattern used by gallery script name as well

specification/compute/Gallery.Management/models.tsp
viveklingaiah: Yes, already deployed

specification/compute/resource-manager/readme.md
razvanbadea-msft: i see you added a pattern is this suppression still needed or it can be removed?

ookoka: @razvanbadea-msft it can be removed! Done

audreyttt (approved):

PR #38718, opened 2025-11-11
add publicNetworkAccess to SitePatchResource

dwmsft:
# Choose a PR Template


weidongxu-microsoft (approved):

weidongxu-microsoft:
It is to fix bug in Swagger that a schema missed a property.

msyyc (approved):

tadelesh (approved):

specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-03-01/WebApps.json
```diff
@@ -29357,6 +29357,10 @@
               "readOnly": true,
               "example": "00000000-0000-0000-0000-000000000000"
             },
+            "publicNetworkAccess": {
```
sandipsh, line 29360: >publicNetworkAccess [](http://example.com/codeflow?start=13&length=19)

please update examples for this change.

specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-03-01/WebApps.json
```diff
@@ -29357,6 +29357,10 @@
...
+              "description": "Property to allow or block all public traffic. Allowed Values: 'Enabled', 'Disabled' or an empty string.",
```
sandipsh, line 29361: >'Enabled', 'Disabled' or an empty string." [](http://example.com/codeflow?start=93&length=42)

please make this enum if it's supposed to have specific string values.

@doc("NotSpecified Network Access")
  NotSpecified: "NotSpecified",

  @doc("Enabled Network Access")
  Enabled: "Enabled",

  @doc("Disabled Network Access")
  Disabled: "Disabled",

weidongxu-microsoft (approved):

specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-03-01/WebApps.json
weidongxu-microsoft: But changing it to enum does not match the type of this same property in service PUT, there it is a string.

https://github.com/Azure/azure-rest-api-specs/blob/b912223e913598b4dedb20b62622f0a1e8927f9d/specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-03-01/CommonDefinitions.json#L3600-L3603

Here the model is for PATCH of the same resource -- as least from SDK point-of-view, we prefer request to PUT and PATCH have same type on same property.

---

SDK can put a directive to "delete the x-ms-enum to make it back to string", if required. So this won't be a blocker.

weidongxu-microsoft:
Had a comment on https://github.com/Azure/azure-rest-api-specs/pull/38718#discussion_r2520642120

specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-03-01/WebApps.json
weidongxu-microsoft: Hi @sandipsh ,

Is it possible for service to have the flexibility to keep the type as "string", for this property (this property alone)?

As SDK dev, we definitely want service to write enum from the start. However, as reason mentioned above, their model in PUT already uses "string" on this property, and we may not want to introduce a different type on the same property that used in PATCH.

<-- this is from SDK dev's perspective. If from ARM perspective, you still think it better be an enum, even it means the type of property be different in PUT and PATCH, we can accept that

dwmsft: Synced with @sandipsh offline. This PR is a bug fix for an existing property from previous versions that was missed in this schema.

weidongxu-microsoft (approved):

PR #38795, opened 2025-11-14
[Monitor] Add Monitor Exporter TypeSpec

pvaneck:
Adding TypeSpec for Monitor Exporter.

**Notes**
1. Error responses with status codes 400, 402, 429, 500, and 503 are modeled with the same response model as successful responses. Will maintain that here.


specification/applicationinsights/data-plane/Monitor.Exporters/preview/v2.1/swagger.json
```diff
@@ -393,347 +438,336 @@
...
+      "allOf": [
+        {
+          "$ref": "#/definitions/Domain"
+        }
       ]
     },
-    "RequestData": {
+    "PropertyValue": {
+      "type": "string",
+      "maxLength": 8192
```
hozhangg: Let's remove this limit. It's not enforced by ingestion service

specification/applicationinsights/ApplicationInsights.Monitor.Exporters/models.tsp
```diff
@@ -0,0 +1,821 @@
...
+   * Identifier of a page view instance. Used for correlation between page view and
+   * other telemetry items.
+   */
+  @maxLength(512)
+  id: string;
+
+  /**
+   * Event name. Keep it low cardinality to allow proper grouping and useful metrics.
+   */
+  @maxLength(1024)
```
hozhangg: Should be 512

specification/applicationinsights/ApplicationInsights.Monitor.Exporters/models.tsp
```diff
@@ -0,0 +1,821 @@
...
```
hozhangg: Should be 512

specification/applicationinsights/ApplicationInsights.Monitor.Exporters/models.tsp
```diff
@@ -0,0 +1,821 @@
+import "@typespec/http";
+
+using TypeSpec.Http;
+
+namespace AzureMonitorExporter;
+
+@maxLength(8192)
```
hozhangg: Please remove this limit, as it's not enforced by ingestion service

specification/applicationinsights/data-plane/Monitor.Exporters/preview/v2.1/swagger.json
```diff
@@ -393,347 +438,336 @@
...
-              "description": "Collection of custom measurements.",
-              "additionalProperties": {
-                "type": "number",
-                "format": "double"
-              }
-            }
+        "name": {
+          "type": "string",
+          "description": "Event name. Keep it low cardinality to allow proper grouping and useful metrics.",
+          "maxLength": 1024
```
hozhangg: Service enforces 512

specification/applicationinsights/data-plane/Monitor.Exporters/preview/v2.1/swagger.json
```diff
@@ -393,347 +438,336 @@
...
+        "id": {
+          "type": "string",
+          "description": "Identifier of a page view instance. Used for correlation between page view and\nother telemetry items.",
+          "maxLength": 512
         },
-        {
+        "name": {
+          "type": "string",
+          "description": "Event name. Keep it low cardinality to allow proper grouping and useful metrics.",
+          "maxLength": 1024
```
hozhangg: Service enforces 512

hozhangg (changes requested):
Left comments on differences between the spec and what's actually enforced by ingestion service

JacksonWeber:
@hozhangg Is there any limit to the length of `properties` on telemetry? Old spec capped this at 8kb, but we have demand for a larger limit.

hozhangg:
> @hozhangg Is there any limit to the length of `properties` on telemetry? Old spec capped this at 8kb, but we have demand for a larger limit.

There is no explicit limit on this field enforced by Breeze. The only limiting factor is the 64kb overall item size.

specification/applicationinsights/data-plane/Exporters/models.tsp
```diff
@@ -0,0 +1,817 @@
...
+  #suppress "@azure-tools/typespec-azure-core/bad-record-type" "For backward compatibility"
+  baseData?: Domain;
+}
+
+/**
+ * The abstract common base of all domains.
+ */
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For backward compatibility"
+model Domain {
+  ...Record<unknown>;
```
pvaneck, line 246: @catalinaperalta, is this the way to define that the model can have arbitrary fields in it?

catalinaperalta: This will mark that the model has additional properties, so if that's the goal then yes it is.

catalinaperalta: I believe even if you do an `extend Record<unknown>` that it will have the same effect

specification/applicationinsights/data-plane/Exporters/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+  "Azure Monitor OpenTelemetry Exporter Service",
+  {
+    /** Application Insights' Breeze endpoint. */
+    host: url = "https://dc.services.visualstudio.com",
+
+    /** The service API version. */
+    apiVersion: Versions,
+  }
+)
+namespace AzureMonitorExporter;
```
hectorhdzg, line 28: If we rename this to ApplicationInsightsClient like before we will have less issues, we expose some configuration currently in JS package, having same name would help us to not have a need for some wrapper to avoid breaking changes

specification/applicationinsights/data-plane/Exporters/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+  ]>
+)
+@service(#{ title: "Azure Monitor Exporter" })
+@versioned(Versions)
+@server(
+  "{host}/{apiVersion}",
+  "Azure Monitor OpenTelemetry Exporter Service",
+  {
+    /** Application Insights' Breeze endpoint. */
+    host: url = "https://dc.services.visualstudio.com",
```
hectorhdzg, line 22: We used to get endpoint override capabilities as well before
https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/monitor/monitor-opentelemetry-exporter/src/generated/models/index.ts#L420

specification/applicationinsights/data-plane/Exporters/main.tsp
pvaneck: Yea, we can add a client name customization in client.tsp to change it for JavaScript.

specification/applicationinsights/data-plane/Exporters/main.tsp
pvaneck: This should still be overrideable in the generated code. For example, this should generate something like:

```ts
const host = options.host ?? "https://dc.services.visualstudio.com";
```

hectorhdzg (approved):
LGTM, Node.js and Python autogenerated code is good

specification/applicationinsights/data-plane/Exporters/client.tsp
```diff
@@ -0,0 +1,68 @@
...
+@@clientName(TelemetryEnvelope.ver, "version");
+@@clientName(TelemetryEnvelope.seq, "sequence");
+@@clientName(TelemetryEnvelope.iKey, "instrumentationKey");
+
+@@clientName(Domain.ver, "version");
+
+@@clientName(DataPoint.ns, "namespace");
+@@clientName(DataPoint.kind, "dataPointType");
+
+// Some models not used in an operation should still be made public.
```
catalinaperalta, line 29: Seems like there are a lot of these types of models, where they also defined in the original swagger?

specification/applicationinsights/data-plane/Exporters/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+@versioned(Versions)
+@server(
+  "{host}/{apiVersion}",
+  "Azure Monitor OpenTelemetry Exporter Service",
+  {
+    /** Application Insights' Breeze host. */
+    host: url = "https://dc.services.visualstudio.com",
+
+    /** The service API version. */
+    apiVersion: Versions,
```
catalinaperalta, line 25: I'm wondering how this is generating in the client libraries? Do you see it treated cross language like the api version parameter? The emitters do look for apiverison so it's probably fine cross language, but if not, then you can add the `@apiVersion` decorator so client libraries know how to handle it.

specification/applicationinsights/data-plane/Exporters/models.tsp
```diff
@@ -0,0 +1,894 @@
...
+  /**
+   * Schema version
+   */
+  ver: int32 = 2;
+
+  /**
+   * Discriminator property to identify the specific telemetry data type.
+   */
+  @visibility(Lifecycle.Read)
+  kind: DomainKind;
```
catalinaperalta, line 257: I noticed that the breaking changes check reports that this property didnt originally exist on the model, does the service support this property?

specification/applicationinsights/data-plane/Exporters/routes.tsp
```diff
@@ -0,0 +1,41 @@
...
+   */
+  #suppress "@azure-tools/typespec-azure-core/request-body-problem" "Raw array type used for backward compatibility"
+  @bodyRoot
+  body: TelemetryEnvelope[],
+
+  /**
+   * Body Parameter content-type
+   */
+  @header("content-type")
+  contentType: "application/json",
```
catalinaperalta: This is the default content type, no need to specify it

specification/applicationinsights/data-plane/Exporters/routes.tsp
```diff
@@ -0,0 +1,41 @@
...
+): {
+  /** All of the telemetry items were accepted and processed. */
+  @statusCode statusCode: 200;
+
+  @body body: TrackResponse;
+} | {
```
catalinaperalta: Nit: likewise, responses default to 200 so also no need to specify this code

Can this just be like in the example below?
```suggestion
): TrackResponse | {
```

specification/applicationinsights/data-plane/Exporters/tspconfig.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/data-plane"
+
+options:
+  "@azure-tools/typespec-autorest":
+    emitter-output-dir: "{project-root}/.."
+    azure-resource-provider-folder: "data-plane"
+    output-file: "Exporters/preview/{version}/swagger.json"
```
catalinaperalta:
```suggestion
    output-file: "{version-status}/{version}/swagger.json"
```

specification/applicationinsights/data-plane/Exporters/tspconfig.yaml
```diff
@@ -0,0 +1,48 @@
...
+emit:
+  - "@azure-tools/typespec-autorest"
+
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/data-plane"
+
+options:
+  "@azure-tools/typespec-autorest":
+    emitter-output-dir: "{project-root}/.."
```
catalinaperalta:
```suggestion
    emitter-output-dir: "{project-root}"
```

catalinaperalta (commented):
Looks good overall, mainly curious about all the extra models

specification/applicationinsights/data-plane/Exporters/client.tsp
pvaneck: Yea, they were also defined in the swagger, so we wanted to have these models be generated as well for consistency.

specification/applicationinsights/data-plane/Exporters/main.tsp
pvaneck: `api_version` is still an optional argument to the client, but it is used in the base_url template (`"{host}/{apiVersion}"`) that is passed to the PipelineClient. The API version here is part of the URL path and not the typical api-version query parameter, so this approach was used.

specification/applicationinsights/data-plane/Exporters/routes.tsp
pvaneck: Thanks, that also works.

specification/applicationinsights/data-plane/Exporters/tspconfig.yaml
pvaneck: Since the versioning here is unconventional with v.2.1 being listed as a preview version, I'll keep this as preview for now as this will try to generate in a `stable` folder.

specification/applicationinsights/data-plane/Exporters/models.tsp
pvaneck: This property is kinda used as a "hack" to enable the use of discriminators and proper serialization of sub-types in the generated code. Marked Lifecycle.Read so that it isn't serialized in the JSON sent to the service. The service itself won't ever return these models, so this is more to assist the downstream serializers.

hozhangg (commented):
LGTM

hozhangg (approved):
clearing blocker for merge

lmazuel (approved):

PR #38832, opened 2025-11-17
Adding ADA fields

olrakovs:
# Choose a PR Template


specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -386,6 +389,44 @@ model PlanDetails {
   @doc("SKU price related properties.")
   @identifiers(#[])
   offeringProperties?: OfferingProperties[];
+
+  @doc("Indicates if this SKU has consumption components")
+  hasConsumptionComponents?: boolean;
+
+  @doc("Indicates if this SKU is available for end users")
+  isEndUserEligible?: boolean;
```
razvanbadea-msft, line 397: >isEndUserEligible [](http://example.com/codeflow?start=2&length=17)

Replace boolean/switch properties with a more meaningful enum whenever possible.

A boolean will forever have two valid values (true or false). A string enum type is always preferred. Also, properties should always provide better values just than True and False. For example two switches "isTypeA" and "isTypeB" should be replaced with one enum "type": [A, B, DefaultType]

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -386,6 +389,44 @@ model PlanDetails {
...
+  isAdminEligible?: boolean;
+
+  @doc("The entra id version for products that come with entra id support")
+  entraIdVersion?: string;
+
+  @doc("Technical Requirements")
+  technicalRequirements?: string;
+
+  @doc("The URI to the FAQ page")
+  faqUri?: string;
```
razvanbadea-msft, line 409: >string [](http://example.com/codeflow?start=11&length=6)

you can use the url scalar

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -386,6 +389,44 @@ model PlanDetails {
...
+
+  @doc("Contains constraint product properties to determine eligibility")
+  constraintsData?: ConstraintsData;
+
+  @doc("Meter type descriptions")
+  @identifiers(#[])
+  meterTypeDescriptions?: MeterTypeDescription[];
+
+  @doc("Usage unit")
+  usageUnit?: string;
```
razvanbadea-msft, line 426: >string [](http://example.com/codeflow?start=14&length=6)

this sounds to be an enum. what possible values can be set for this property

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -613,6 +654,18 @@ model Term {
...
   isAutorenewable?: boolean;
+
+  @doc("Defines the lifecycle management policy for the term")
+  lifecyclePolicy?: LifecyclePolicy;
+
+  @doc("Product code for this term")
+  productCode?: string;
+
+  @doc("Term state")
+  state?: string;
```
razvanbadea-msft, line 665: >string [](http://example.com/codeflow?start=10&length=6)

what are the possible states that can be set?

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -705,3 +758,153 @@ model Image {
...
+  productSkuId?: string;
+
+  @doc("Quantity of the bundled product per license ")
+  defaultKeyActivationCount?: int32;
+}
+
+@doc("Attributes that define fulfillment timing and delay handling")
+model FulfillmentAttributes {
+  @doc("Timing configuration for product fulfillment")
+  fulfillmentTiming?: string;
```
razvanbadea-msft, line 787: >fulfillmentTiming [](http://example.com/codeflow?start=2&length=17)

i do not see this in some examples, so how will the values look like, can you describe the format in description

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -705,3 +758,153 @@ model Image {
...
+
+  @doc("Mitigation explanation for delayed fulfillment")
+  fulfillmentDelayMitigation?: string;
+}
+
+@doc("Benefit information for a plan")
+model Benefit {
+  @doc("Benefit type")
+  benefitType?: string;
```
razvanbadea-msft, line 796: >benefitType [](http://example.com/codeflow?start=2&length=11)

can you set any string for this property?

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -705,3 +758,153 @@ model Image {
...
+
+  @doc("Base plan ID")
+  basePlanId?: string;
```
razvanbadea-msft, line 799: >basePlanId [](http://example.com/codeflow?start=2&length=10)

is this in arm-id format?

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -705,3 +758,153 @@ model Image {
...
+
+  @doc("Billing plan")
+  billingPlan?: string;
+
+  @doc("Term duration")
```
razvanbadea-msft: >doc [](http://example.com/codeflow?start=3&length=3)

you can describe the format for its values

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
```diff
@@ -705,3 +758,153 @@ model Image {
...
+  meterType?: string;
+
+  @doc("The unit of measurement for the meter")
+  unitOfMeasure?: string;
+}
+
+@doc("Defines lifecycle management policy")
+model LifecyclePolicy {
+  @doc("Grace period duration where customers retain access to their product")
+  graceDuration?: string;
```
razvanbadea-msft, line 903: >graceDuration [](http://example.com/codeflow?start=2&length=13)

specify in desciptions the format used for this properties

razvanbadea-msft:
Fix errors from Swagger ModelValidation

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: Although it looks enum like, we can’t model this as an enum. The value is received as a dynamic string from a downstream service, and new values may be added without coordination. An enum would either fail deserialization or require frequent updates.
The values that are known for us are "program pass"

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: possible states that are known for us are  EndOfSale and GA

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: examples were added, there is no special format

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: yes, the string is received from a downstream service

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: no, this is not arm-id format

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: added

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: added

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: Unfortunately, we cannot change the type. The API is already in use (preview version), and doing so would cause breaking changes.

specification/marketplacecatalog/resource-manager/Microsoft.Products/Products/product.tsp
olrakovs: Unfortunately, we cannot change the type. The API is already in use (preview version), and doing so would cause breaking changes.

alonhann (approved):

PR #38841, opened 2025-11-17
Search tsp rc

xiangyan99:
Convert Search data plane swagger 2025-11-01-preview into tsp.


weshaggard:
Force merged as all the failures are known and expected according to @xiangyan99

PR #38846, opened 2025-11-17
Review request for Microsoft.ContainerService/aks to add version preview/2025-10-02-preview

FumingZhang:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


FumingZhang:
Regarding the breaking change, it is being compared to the latest stable API version, and the results listed are identical to those from the [previous preview API version](https://github.com/Azure/azure-rest-api-specs/pull/38472#issuecomment-3459369033)

image

- Item 1-2 of [1029 - ReadonlyPropertyChanged](https://github.com/Azure/openapi-diff/tree/master/docs/rules/1029.md) are caused by change https://github.com/Azure/azure-rest-api-specs/pull/37803/commits/baf5edb5c735e1461e5f4a4deb5c921aa681ee5d.

    The reason for this change is that customers need a recovery mechanism in case upgrades failed or even if the upgrade succeeded but customer has workload issues on the newer version. Goal is to rollback kubernetes/node image version to allow them to go back to the previous version(s), so removing the previous read-only restriction on this field.

- Items 3-7 of [1029 - ReadonlyPropertyChanged](https://github.com/Azure/openapi-diff/tree/master/docs/rules/1029.md) and [1036 - ConstraintChanged](https://github.com/Azure/openapi-diff/tree/master/docs/rules/1036.md) are caused by change https://github.com/Azure/azure-rest-api-specs/commit/c66c21ceccde02cc4b9ee75a8794e831cda35d5d in previous preview API version 2025-04-02-preview

    The reason for this change is that previously, AKS only supported Get/List Machines. In this new API proposal, AKS introduces Put Machine as a new operation. Since the machine data model and name parameter will now be shared between Get and Put, we are removing the read-only restriction from certain properties and expanding the name parameter accordingly. Check this https://github.com/Azure/azure-rest-api-specs/pull/35081#issuecomment-2938079182.

FumingZhang:
Regarding failed [Swagger LintDiff](https://github.com/Azure/azure-rest-api-specs/actions/runs/19448078086), the model is reused from [compute rp](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/Skus/stable/2021-07-01/skus.json#L114), and its original definition does not include attributes such as name, id or type, so added a [suppression](https://github.com/Azure/azure-rest-api-specs/pull/38846/commits/63fe614ba64b4f31677e3a3a278333e5ab72a314) for this.

image

mentat9:
@FumingZhang - Regarding the suppression: does the Compute API include those properties and they are missing from the spec? Or are they actually missing from the API?

CustardTart32:
@mentat9 The properties (`id`, `type`) are missing from the compute API (and by extension spec). 

I've decided to define our own models as I wasn't sure if we could directly import models from other RP's and existing examples clone compute models instead of importing existing ones. Happy to abide by your suggestion however.

mentat9:
>           "type": "string",

Add format: armId (and x-ms-arm-id-details as appropriate). Ref: [https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-arm-id-details](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-arm-id-details).

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5265 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

mentat9:
>         "name": {

[ARMBlockingComment] Don't duplicate top-level property names (name, type, id, etc.) within the property bag. You could use skuName here. There are a couple of other examples below (name, type) which you should also consider changing, but in those cases where the objects are inside an array property, it's not as critical.

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5269 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

mentat9:
>       }

Just checking that you want an array of extendedLocations, but a single extendedLocation type.

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5478 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

mentat9:
>         "name": {

Should be plural, e.g. names or zoneNames

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5559 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

mentat9:
>         "enableEncryptionAtHost": {

Consider using enum instead of boolean. Enums are better properties for many reasons. This could be something like: encryption: { None, Host }.

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:10142 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

mentat9:
>     "ResourceSku": {

These properties need to be placed in a property bag within the ARM resource envelope. You can include the top-level definitions from common v6 types.json.

---
Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5260 in b11422e. [](commit_id = b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe, deletion_comment = False)

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/readme.md
```diff
@@ -1433,4 +1442,8 @@ directive:
     from: managedClusters.json
     where: $.definitions.LocalDNSOverrides
     reason: User defined custom key-value pairs, similar to the allowed "user defined tags." These pairs can have any value, as there is no validation on the values
+  - suppress: RequiredPropertiesMissingInResourceModel
+    from: managedClusters.json
+    where: $.definitions.ResourceSkusResult
+    reason: The model is reused from [compute rp](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/compute/resource-manager/Microsoft.Compute/Skus/stable/2021-07-01/skus.json#L114), and its original definition does not include attributes such as name, id or type.
```
mentat9: Once you fix the ResourceSku definition to have the ARM resource envelope, you can remove this suppression.

CustardTart32:
Hi @mentat9, regarding your comments regarding the List VM SKUs API, I would also like to mention that this endpoint is a wrapper around the Compute resource SKUs endpoint. We essentially perform some filtering on the AKS side and return the response in the same shape as the Compute endpoint. 

Your suggestions would require us to modify the codebase in order to develop a conversion layer between the compute `ResourceSku` models and the AKS `ResourceSku` models. I'm not sure if this is really necessary from an AKS perspective. If we just imported the Compute models instead of redefining them, would that be satisfactory for you? 

```
        "responses": {
          "200": {
            "description": "The request has succeeded.",
            "schema": {
              "$ref": "../../../../../../compute/resource-manager/Microsoft.Compute/Skus/stable/2021-07-01/skus.json#/definitions/ResourceSkusResult"
            }
          },
 ```

**Comment by FumingZhang** [comment]
> > ```
> >     "enableEncryptionAtHost": {
> > ```
> 
> Consider using enum instead of boolean. Enums are better properties for many reasons. This could be something like: encryption: { None, Host }.
> 
> Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:10142 in [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe). [](commit_id = [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe), deletion_comment = False)

We agree that using an enum is generally better than a bool. However, in this case, there are only two possible states—whether to encrypt the host disk or not. For simplicity and to stay consistent with other properties in the same profile, we prefer to keep it as it is.

**Comment by mentat9** [comment]
> Hi @mentat9, regarding your comments regarding the List VM SKUs API, I would also like to mention that this endpoint is a wrapper around the Compute resource SKUs endpoint. We essentially perform some filtering on the AKS side and return the response in the same shape as the Compute endpoint.
> 
> Your suggestions would require us to modify the codebase in order to develop a conversion layer between the compute `ResourceSku` models and the AKS `ResourceSku` models. I'm not sure if this is really necessary from an AKS perspective. If we just imported the Compute models instead of redefining them, would that be satisfactory for you?
> 
> ```
>         "responses": {
>           "200": {
>             "description": "The request has succeeded.",
>             "schema": {
>               "$ref": "../../../../../../compute/resource-manager/Microsoft.Compute/Skus/stable/2021-07-01/skus.json#/definitions/ResourceSkusResult"
>             }
>           },
> ```

No, you are right to create your own definition rather than reference another RP's spec. That kind of coupling is not allowed. Also, I understand your concern about needing to add a conversion layer to implement the ARM resource contract. However, this is a real issue that normally would be discussed in modeling office hours.

Here's the problem: you are attempting to expose the Microsoft.Compute SKUs contract from your RP, when it meets neither your RPs SKUs contract (it is not listing your RP's SKUs), nor the basic ARM resource type contract (doesn't have ARM resource type envelope). You don't want to expose Microsoft.Compute SKUs as your own and implementing this as GET of a resource without following the ARM resource contract is not allowed.

You have a couple of options:
1. Follow the ARM resource contract to expose compute SKUs as your resource type (vmSkus). This is what I was assuming in the feedback I already provided.
2. Change the API to a POST action called **list&lt;thing&gt;**, where **&lt;thing&gt;** is whatever these items should be called. For example, (just my guess) **listAvailableComputeSkus**:
```
POST /subscriptions/{subscriptionId}/providers/Microsoft.ContainerService/locations/{location}/listAvailableComputeSkus
```
Or choose another action name that describes the semantics better. The POST action payload can mimic that of the Microsoft.Compute/skus API: the property bag and top-level resource properties are not required for a POST action.

CustardTart32:
@mentat9, thanks for your feedback. I've decided to change the API to a `POST` request as outlined in the Proxy Resource RPC guide and as per your comments.

`POST /subscriptions/{subscriptionId}/providers/Microsoft.ContainerService/locations/{location}/listVmSkus`

https://github.com/Azure/azure-rest-api-specs/pull/38993 (cc. @FumingZhang)

Regarding your other comments about plural/singular field names, as this endpoint mimics the compute/skus API, I'd like to keep those unchanged as long as these are not blocking issues.

matthchr:
TBH I think we should pursue @mentat9's suggestion 1:
> Follow the ARM resource contract to expose compute SKUs as your resource type (vmSkus). This is what I was assuming in the feedback I already provided.

I'm not a huge fan of the `POST /listSkus`.

Note that any changes to the shape of this API in the Swagger are problematic ATM because the service backing the original Swagger is already rolled out and there are no more changes possible this year AFAIK. Getting feedback at this point from ARM reviewers is really too late since the service code has already been written. I'm about to go on vacation for a while but we (AKS) should sync with ARM on how we can frontload this review process in such a way as to avoid getting this feedback so late in the process. We generally try to call things like this out but I was under the (mistaken?) impression that GETs that don't follow the ARM resource model are allowed but discouraged, because some already exist (see for example the Compute ListSKUs API which has this exact same structure).

psah434:
Have all the comments from the previous reviewer been addressed? there is also an arm blocking comment. Please respond to each comment if it is has been fixed.

CustardTart32:
> TBH I think we should pursue @mentat9's suggestion 1:
> 
> > Follow the ARM resource contract to expose compute SKUs as your resource type (vmSkus). This is what I was assuming in the feedback I already provided.
> 
> I'm not a huge fan of the `POST /listSkus`.

On suggestion 1, I did consider this but decided against it for several reasons. 
- One of the functional requirements listed in the PRD is to align with the CRP endpoint. In order to align with suggestion 1, we would need to add new fields (`id`, `type`, `name`) and rename existing fields within the property bag (`name` -> `skuName`).
- SKU's don't really align well with the ARM resources. For example, SKU's don't have an ID that provides a unique identifier for the "resource" in question. In order to identify a SKU, we normally use a combination of the `name` and the `resourceType` which doesn't align well with ARM URL encoded resource ID's.

Whilst I agree that `POST /listSkus` isn't the elegant solution, the [RP proxy resource contract](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/proxy-api-reference.md#resource-action-requests) does provide explicitly provide examples for list operations to be used as action names.  It would also allow us to sidestep these ARM resource schema requirements.

CustardTart32:
> Note that any changes to the shape of this API in the Swagger are problematic ATM because the service backing the original Swagger is already rolled out and there are no more changes possible this year AFAIK. Getting feedback at this point from ARM reviewers is really too late since the service code has already been written. I'm about to go on vacation for a while but we (AKS) should sync with ARM on how we can frontload this review process in such a way as to avoid getting this feedback so late in the process. We generally try to call things like this out but I was under the (mistaken?) impression that GETs that don't follow the ARM resource model are allowed but discouraged, because some already exist (see for example the Compute ListSKUs API which has this exact same structure).

Agree, let's raise this issue and sync with the AKS and ARM teams on how to avoid these blocking issues late into the development process. I was under the original assumption that this would be allowed as we already have existing `GET` endpoints that return non compliant response formats (`listKubernetesVersions`) so I thought this too would also be allowed.

CustardTart32:
> Have all the comments from the previous reviewer been addressed? there is also an arm blocking comment. Please respond to each comment if it is has been fixed.

@FumingZhang, let's revert the `List SKU's` swagger model changes and create a separate PR once the AKS team decides on the correct swagger format and get ARM pre-approval then.

EDIT: Awaiting ARM reviewer response on whether or not we can keep the List VM SKU's as a GET endpoint before reverting.

matthchr:
Ok I synced up with @CustardTart32 and he's convinced me that we don't actually want suggestion 1 because the problem is our whole intent here is to have this API look like https://learn.microsoft.com/en-us/rest/api/compute/resource-skus/list?view=rest-compute-2025-04-01&tabs=HTTP (which is the publicly exposed compute skus API). _Maybe_ post is better if we have to have it be a POST to maintain shape compatibility with what CRP exposes.

We can't fix it in this API version though as mentioned above the service is already out for this API version and there's a code freeze so we can't just fix it. Changing anything at this point would mismatch with the actual service implementation.

Can you expand on this?
> You don't want to expose Microsoft.Compute SKUs as your own and implementing this as GET of a resource without following the ARM resource contract is not allowed.

What are the consequences if a GET-only resource such as `vmSkus` as proposed doesn't actually match the ARM API resource shape contract? (i.e. what tools are broken, what workflows don't work, etc). I ask because we have many others that already don't follow the resource pattern and are plain GETs, like...
[ListKubernetesVersions](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json#L73)
[ListNodeImageVersions](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json#L113)
[ListOutboundNetworkDependenciesEndpoints](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-09-02-preview/managedClusters.json#L2952)

There are probably more (I didn't do an exhaustive search). Given the above I'd like to understand the consequences of one more.

psah434:
As I have mentioned previously, please go through each comment in this PR ( the entire time the pr has been open) and respond if it has been addressed before putting it in the queue again. There are too many commits here and this is very difficult to review.

CustardTart32:
> > ```
> >       "type": "string",
> > ```
> 
> Add format: armId (and x-ms-arm-id-details as appropriate). Ref: https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-arm-id-details.
> 
> Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5265 in [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe). [](commit_id = [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe), deletion_comment = False)

Won't fix. This field isn't actually an ARM ID but rather a string matching the resource type name specified in the manifest.

CustardTart32:
> > ```
> >     "name": {
> > ```
> 
> [ARMBlockingComment] Don't duplicate top-level property names (name, type, id, etc.) within the property bag. You could use skuName here. There are a couple of other examples below (name, type) which you should also consider changing, but in those cases where the objects are inside an array property, it's not as critical.
> 
> Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5269 in [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe). [](commit_id = [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe), deletion_comment = False)

Addressed, migrated List VM SKUs to use POST instead of GET. Alternatively, as per @matthchr's comment, if we are allowed to use GET requests for non ARM compliant resources, this would also be addressed.

CustardTart32:
> > ```
> >     "name": {
> > ```
> 
> Should be plural, e.g. names or zoneNames
> 
> Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5559 in [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe). [](commit_id = [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe), deletion_comment = False)

Won't fix. This endpoint is meant to align with the [Compute LIST Resource SKUs API](https://learn.microsoft.com/en-us/rest/api/compute/resource-skus/list?view=rest-compute-2025-02-01&tabs=HTTP) and thus we will copy the response field names verbatim.

CustardTart32:
> > ```
> > "ResourceSku": {
> > ```
> 
> These properties need to be placed in a property bag within the ARM resource envelope. You can include the top-level definitions from common v6 types.json.
> 
> Refers to: specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2025-10-02-preview/managedClusters.json:5260 in [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe). [](commit_id = [b11422e](https://github.com/Azure/azure-rest-api-specs/commit/b11422e4e2f0ec8ca0b6f3d52ae5fb97d1cf4abe), deletion_comment = False)

Addressed. Migrated List VM SKUs to POST operation. Otherwise, if you deem that a GET request is OK here, we can ignore this comment.

FumingZhang:
@psah434 could you please clarify which comment we haven't responded to? For clarity, I'm adding some screenshots with annotations to show that we have responded to all comments from mentat9 and highlighted several areas where we’re awaiting feedback from ARM.

1. the initial feedback and response
image

2. the second review, comments regarding the ResourceSKU API and enum.
comments2

3. response to the second review
Screenshot 2025-12-03 at 10 42 41 am

4. follow-up on the thread 1 in second review
Screenshot 2025-12-03 at 10 42 50 am

5. response to the follow-up on the thread 1 in second review, the latest comment leaves an open question for ARM.
replyt2

CustardTart32:
Just to add to @FumingZhang's comments, for Thread 1, there are two main options here: 

1. Convert List VM Skus from a GET to a POST endpoint (this has been done already)
2. Keep List VM Skus as a GET endpoint if the ARM reviewer is OK with us adding a new GET endpoint that is non-compliant with the ARM resource schema. (We would need revert the `GET`-> `POST` conversion commit in this case with most of the thread 1 comments redundant in this case)

FumingZhang:
In the very last commit [#bca493d](https://github.com/Azure/azure-rest-api-specs/pull/38846/commits/bca493d9bf196951a735d8140748d65b0e6337a2), we just reverted all the changes related to the ResourceSKU API.

> Following feedback from the ARM reviewer, we updated the proposed ResourceSKU API and confirmed that the revised API has passed the ARM review. However, the server-side code changes have not yet been updated to match this revision. As a result, we will revert these changes and delay their release until the next available preview API version.

palma21 (approved):

PR #38872, opened 2025-11-18
[summarize-checks] Prepare for tseslint.configs.recommendedTypeChecked

mikeharder:
- Add/remove missing/unnecessary `async/await`
- Use `zod` to validate input and create strongly-typed objects
- Add type comments
- Add fallbacks for undefined values

### DIFF
--- a/.github/workflows/src/summarize-checks/summarize-checks.js
+++ b/.github/workflows/src/summarize-checks/summarize-checks.js
@@ -24,6 +24,7 @@ import { byDate, invert } from "../../../shared/src/sort.js";
 import { commentOrUpdate } from "../comment.js";
 import { extractInputs } from "../context.js";
 import {
+  ImpactAssessmentSchema,
   brChRevApproval,
   getViolatedRequiredLabelsRules,
   processImpactAssessment,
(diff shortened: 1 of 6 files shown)


.github/workflows/src/summarize-checks/summarize-checks.js
```diff
@@ -363,7 +374,7 @@ export async function summarizeChecksImpl(
   const prUrl = `https://github.com/${owner}/${repo}/pull/${issue_number}`;
   core.summary.addRaw("PR: ");
   core.summary.addLink(prUrl, prUrl);
-  core.summary.write();
+  await core.summary.write();
```
scbedd, line 377: aha missed await here.

scbedd: oh jeez multiple places.

scbedd (approved):
Nice! the failure to await core.summary() is pretty lame. I'm pretty certain that didn't actually affect any PRs, but I'm still happy to have that caught.

.github/workflows/src/summarize-checks/summarize-checks.js
mikeharder: easy to miss without a linter rule, which is why we're enabling it in #38673

PR #38873, opened 2025-11-19
[verify-run-status] Delete unused workflow and shared code

mikeharder:


.github/workflows/src/github.js
```diff
@@ -24,83 +22,6 @@ export async function writeToActionsSummary(content, core) {
   }
 }
 
-/**
```
mikeharder, line 27: Functions were only used by `verify-run-status.js`.  Other code calls `github.rest` directly.

PR #38890, opened 2025-11-19
Added preliminary Rust binding to a whole lot of tspconfig.yaml files

LarryOsterman:

## Purpose of this PR

Add Rust bindings to a significant number of tspconfig.yaml files. Note that these bindings do NOT imply that Rust clients will be generated for these packages, neither does it imply that the namespaces for these packages is correct.

Before the decision is made to generate a Rust SDK, please confirm with the architecture board that the package name is correct as is the working directory for the generated package.

- [X] Make changes to the SDK configuration only when there are no modifications to the API specification, eliminating the need for an ARM or Stewardship Board API review.

## Due diligence checklist

- [X] I confirm this PR is modifying only SDK configurations, and not API related specifications.
- [X] I have reviewed and used the respective `tspconfig.yaml` templates:

## Getting help


specification/appconfiguration/resource-manager/Microsoft.AppConfiguration/AppConfiguration/tspconfig.yaml
```diff
@@ -22,6 +22,10 @@ options:
     generate-test: true
     generate-sample: true
     flavor: azure
+  "@azure-tools/typespec-rust":
+    crate-name: "azure_management_appconfiguration"
```
heaths: Decision for CP was `azure_resourcemanager_` as a prefix: https://azure.github.io/azure-sdk/rust_introduction.html#rust-packaging-prefix

specification/appconfiguration/resource-manager/Microsoft.AppConfiguration/AppConfiguration/tspconfig.yaml
```diff
@@ -22,6 +22,10 @@ options:
...
+    crate-version: "0.1.0"
+    emitter-output-dir: "{output-dir}/resource_manager/{service-dir}/{crate-name}"
```
heaths: Similar to .NET, we'll also put these in the same service directory. They already differ by crate name. I see no reason to necessarily nest them deeper.

heaths (changes requested):

mikeharder:
@LarryOsterman: You may want to add validation for the rust config here, alongside the validation for the other languages:

https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts

LarryOsterman:
> @LarryOsterman: You may want to add validation for the rust config here, alongside the validation for the other languages:
> 
> https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts

I was thinking about that, but I *believe* that @heaths has that on his plate. More importantly, I know for certain that the Rust emitter cannot handle many of these crates (it failed 3 out of 5 tries earlier this week). So adding validation right now is likely to be premature.

specification/appconfiguration/resource-manager/Microsoft.AppConfiguration/AppConfiguration/tspconfig.yaml
LarryOsterman: Fixed.

specification/appconfiguration/resource-manager/Microsoft.AppConfiguration/AppConfiguration/tspconfig.yaml
LarryOsterman: Fixed.

heaths:
Yeah, in the new year. I'm first going to go through and try to generate a Rust crate for every `tspconfig.yaml` and see where we have gaps. That said, more often than not I see all other languages' SDK steps failing to build so not sure if adding Rust to the mix now makes that any worse. Still, probably better to wait.

LarryOsterman:
> Yeah, in the new year. I'm first going to go through and try to generate a Rust crate for every `tspconfig.yaml` and see where we have gaps. That said, more often than not I see all other languages' SDK steps failing to build so not sure if adding Rust to the mix now makes that any worse. Still, probably better to wait.

Btw, looking at the validation, some of these can be done now (for instance, validating the output location is valid, validating that management plane SDKs have `resource_manager_` in the crate name, etc. I'm still trying to see if this also adds attempting to compile the crate.

heaths (approved):

heaths (approved):

eng/tools/typespec-validation/test/flavor-azure.test.ts
```diff
@@ -22,7 +22,6 @@ describe("flavor-azure", function () {
     "@azure-tools/typespec-java",
     "@azure-tools/typespec-python",
     "@azure-tools/typespec-ts",
-    "@azure-tools/typespec-rust",
```
heaths: Why this change? We're not at that level yet e.g., not tier 1?

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -642,6 +642,23 @@ export class TspConfigRustMgmtCrateNameSubRule extends TspconfigEmitterOptionsSu
...
+  }
+
+  protected validate(config: any): RuleResult {
+    let emitterOutputDir = config?.options?.[this.emitterName]?.["emitter-output-dir"];
+    if (emitterOutputDir === undefined) return this.createFailedResult("Failed to find \"options.@azure-tools/typespec-rust.emitter-output-dir\"", "Please add \"options.@azure-tools/typespec-rust.emitter-output-dir\" with a path matching the SDK naming convention \"{output-dir}/{service-dir}/{crate-name}\"");
+    if (emitterOutputDir !== "{output-dir}/{service-dir}/{crate-name}") {
+      return this.createFailedResult(
+        `The value of options.${this.emitterName}.emitter-output-dir "${emitterOutputDir}" does not match the required format "{output-dir}/{service-dir}/{crate-name}"`, "Correct the emitter-output-dir to be \"{output-dir}/{service-dir}/{crate-name}\""
+      );
+    }
```
heaths: If it's required to be a specific value, kind of begs the question why it's a value at all and not just computed.

heaths (approved):

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
LarryOsterman: It's a required field by tsp-client. I don't believe it can be computed.

eng/tools/typespec-validation/test/flavor-azure.test.ts
LarryOsterman: This was incorrectly added and triggered a CI pipeline failure, so it was removed. If you check, this file should be unmodified from main.

heaths (approved):

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -649,10 +649,15 @@ export class TspConfigRustAzEmitterOutputDirSubRule extends TspconfigEmitterOpti
 
   protected validate(config: any): RuleResult {
     let emitterOutputDir = config?.options?.[this.emitterName]?.["emitter-output-dir"];
-    if (emitterOutputDir === undefined) return this.createFailedResult("Failed to find \"options.@azure-tools/typespec-rust.emitter-output-dir\"", "Please add \"options.@azure-tools/typespec-rust.emitter-output-dir\" with a path matching the SDK naming convention \"{output-dir}/{service-dir}/{crate-name}\"");
+    if (emitterOutputDir === undefined)
+      return this.createFailedResult(
+        'Failed to find "options.@azure-tools/typespec-rust.emitter-output-dir"',
+        'Please add "options.@azure-tools/typespec-rust.emitter-output-dir" with a path matching the SDK naming convention "{output-dir}/{service-dir}/{crate-name}"',
+      );
```
heaths: If it was a lint that failed, I imagine this will be next: curly braces. Yeah, valid, but tends to freak out linters. I guess we'll see.

heaths (approved):

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
LarryOsterman: This particular one fails when there is no emitter-output-dir field in the file. Not sure if it's strictly necessary, tbh.

Still working through a stupid number of errors.

specification/suppressions.yaml
```diff
@@ -183,6 +186,8 @@
     - options.@azure-tools/typespec-java.*
     - options.@azure-tools/typespec-python.*
     - options.@azure-tools/typespec-ts.*
+    - options.@azure-tools/typespec-ts.*
```
mikeharder:
```suggestion
```

Remove duplicate line.

specification/agricultureplatform/AgriculturePlatform.Management/tspconfig.yaml
```diff
@@ -27,6 +27,10 @@ options:
     flavor: "azure"
     generate-test: true
     generate-sample: true
+  "@azure-tools/typespec-rust":
```
weshaggard, line 30: Do all these SDK successfully generate currently? @raych1 are there any plans to have an SDK validation pipeline for rust?

@LarryOsterman @heaths we should probably avoid adding all these configurations until we have some sort of validation pipeline for them.

specification/ai/ContentUnderstanding/tspconfig.yaml
```diff
@@ -49,3 +53,5 @@ options:
     partial-update: true
     use-eclipse-language-server: false
     flavor: azure
+    generate-test: false
```
weshaggard, line 55: These settings look unrelated.

specification/suppressions.yaml
```diff
@@ -1257,3 +1288,11 @@
...
     - voiceservices/VoiceServices.Provisioning/tspconfig.yaml
     - webpubsub/data-plane/WebPubSub/tspconfig.yaml
+
+- tool: TypeSpecValidation
+  reason: Suppress until Rust needs to generate SDK
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-rust.*
+  paths:
+    - storage/Microsoft.BlobStorage/tspconfig.yaml
```
mikeharder, line 538: Adding rust to the other specs with existing suppressions makes sense.  But why do you need this new suppression, instead of adding the rust config to this spec, like you did the other specs?

specification/agricultureplatform/AgriculturePlatform.Management/tspconfig.yaml
raych1: > are there any plans to have an SDK validation pipeline for rust?

I heard from @ronniegeraghty several months ago for the interest to onboard Rust to the `SDK Validation` pipeline. However, I haven't noticed any follow-ups yet.  

@LarryOsterman @heaths do we want to onboard Rust to the `SDK Validation` pipeline?

weshaggard (changes requested):
Let's not merge this PR until after the new year and we have onboarded the rust sdk validation pipeline. I want to ensure these all work before adding the configurations.

specification/agricultureplatform/AgriculturePlatform.Management/tspconfig.yaml
LarryOsterman: Right now, we can't onboard a rust pipeline because we have a ton of validation errors. A large part of the purpose f this PR is to make it simpler for the Rust SDK team to bulk compile the azure-rest-api-specs repo to find the gaps between spector and the real validators (gaps we've found for example: CBOR support, Text/Plain support). 

Once we've done the validation, I think a validation pipeline is a great idea.

specification/ai/ContentUnderstanding/tspconfig.yaml
LarryOsterman: That may have been a merge error - I hand verified these earlier. I'l review again.

specification/suppressions.yaml
LarryOsterman: I couldn't find any supressions that were ONLY Microsoft.BlobStorage - I found convenient ones for the other suppressions. And the Go suppression above has a LOT of other packages beyond just the few which were failing

specification/agricultureplatform/AgriculturePlatform.Management/tspconfig.yaml
raych1: We have individual pipeline to support the batch run on multiple specs. The config changes do not have to be in the `main` branch. The pipeline can be triggered against a feature branch.

specification/suppressions.yaml
mikeharder: Many of the existing suppressions in this file were added as temporary stop-gap.  Most of the suppressions are grouped by service, but I see there was recently a group of suppressions added for "go".  Before we add even more suppressions (debt) for rust, let's see how we can minimize the list.

LarryOsterman: FWIW, as far as I could tell, the vast majority of these suppressions are for services which don't have a relevant language section in their tspconfig.yaml file. 

These aren't about specific checks that need to be suppressed, they are for packages that don't have a @typespec-go or whatever section.

mikeharder: If adding a `@typespec-go` section is opt-in, I don't think specs without the exception should require an exception.  There should be a less-exceptional way to configure this.  We should review this design before merging this PR.

mentat9:
This PR doesn't need ARM review: adding ARMChangesRequested to get it out of the ARM reviewer's queue.

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -659,6 +683,8 @@ export const defaultRules = [
   new TspConfigCsharpMgmtNamespaceSubRule(),
   new TspConfigCsharpAzEmitterOutputDirSubRule(),
   new TspConfigCsharpMgmtEmitterOutputDirSubRule(),
+  new TspConfigRustMgmtCrateNameSubRule(),
+  new TspConfigRustAzEmitterOutputDirSubRule(),
```
MaryGao, line 854: Is Rust emitter config a required check for SdkTspConfigValidationRule in tspconfig.yaml?

The **required** check means if the tspconfig.yaml would appear, then any absent of Rust config would fail the SdkTspConfigValidationRule.

If we design this as **optional** check, which means:
- when @azure-tools/typespec-rust emitter is existing, we would check the relevant rules;
- when @azure-tools/typespec-rust emitter is not defined, we would not check its rules.

Please note I am not talking about the implementation and the current impl may not design for required and optional check. But we want to introduce this idea since we notice some checks are optional and here is the issue to track this: https://github.com/Azure/azure-sdk-tools/issues/13124.

skywing918: Since https://github.com/Azure/azure-sdk-tools/issues/13124 has been completed, I’ve merged the latest changes and made the necessary updates accordingly. Please refer to the following:
https://github.com/Azure/azure-rest-api-specs/pull/39526/files#diff-18a8b490c39f355883ff6a3609df4b3d4aea8854941c771aab773df3dd279127

skywing918: I have updated the rust rules, and added test cases for rust. it works now. also found issues from tspconfig.yaml, please check the comments.
[Add Rust emitter output directory validation rule and test cases](https://github.com/Azure/azure-rest-api-specs/pull/38890/commits/a0b153677bd804f6061c4e1195a4de3d35ca0791)

PR #38892, opened 2025-11-20
[JS] Update relevant TSV rules and remove unnecessary options because the default type is Modular for DPG

v-jiaodi:
Related pr https://github.com/Azure/autorest.typescript/pull/3369 and the whole epic is https://github.com/Azure/autorest.typescript/issues/3350.

Update relevant TSV rules and remove unnecessary options because the dpg modular has already been set as the default.

/cc @MaryGao


MaryGao (approved):

specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
```diff
@@ -32,7 +32,6 @@ options:
     model-namespace: false
   "@azure-tools/typespec-ts":
     emitter-output-dir: "{output-dir}/{service-dir}/contoso-widgetmanager"
-    is-modular-library: true
```
MaryGao, line 35: @maririos and @mikeharder Could you help review and approve this pr? 

JS has adjusted the default rule for option `is-modular-library` and this setting won't be necessary in our contoso project.

mikeharder: Should the setting also be removed from our `widget` sample?

https://github.com/Azure/azure-rest-api-specs/blob/c3c93e9102521197fffb02498222aa1c2375d69a/specification/widget/data-plane/WidgetAnalytics/tspconfig.yaml#L31

Also, I see a lot more hits for `is-modular-library`, both `true` and `false`, in existing specs in main.  Why does this PR only change these 3?

mikeharder (dismissed):
Waiting for answer on how many specs should be updated.

specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml
MaryGao: Good to know we have other samples, @v-jiaodi Could you update config to remove all `is-modular-library: true`?

v-jiaodi: updated

mentat9:
This PR does not require ARM signoff: adding ARMChangesRequested to keep it out of the ARM reviewer's queue.

MaryGao:
@mikeharder Could you help review the pr?

mikeharder:
> @mikeharder Could you help review the pr?

I dismissed my review since the `widget` sample is now updated, but I don't feel qualified to approve the PR in general.

lirenhe (approved):

MaryGao:
@maorleger Could you help review this pr? I think the whole context is from epic https://github.com/Azure/autorest.typescript/issues/3350 that we switch the default libraries from RLC to Modular.

maorleger (approved):

maririos (approved):
@MaryGao make sure the updates happen in all the right places outside of this repo please :)

chlowell (approved):

lirenhe (approved):

lirenhe:
The dotnet and Java pipeline failures are not related to this PR as it only changes the config for JS side.
no change for the SDK so there is no need to wait for each service team's approval.
Plan to force merge this PR

v-jiaodi:
For JS pipeline failed, the root cause is that there are still some issues with js codegen's support for Flatten property. I have already created two issues to track this.
https://github.com/Azure/autorest.typescript/issues/3631
https://github.com/Azure/autorest.typescript/issues/3632

XiaofeiCao (approved):

haolingdong-msft (approved):

lirenhe:
Got multiple approvals from different language teams,  Force merge this PR.

PR #38912, opened 2025-11-20
[lint-diff] Enable eslint

mikeharder:


eng/tools/lint-diff/src/lint-diff.ts
```diff
@@ -67,13 +67,13 @@ export async function main() {
   // TODO: Handle trailing slashes properly
   if (!beforeArg || !(await pathExists(beforeArg as string))) {
     validArgs = false;
-    console.log(`--before must be a valid path. Value passed: ${beforeArg || "<empty>"}`);
+    console.log(`--before must be a valid path. Value passed: ${inspect(beforeArg) || "<empty>"}`);
```
danieljurek, line 70: I wonder if these should be `beforeArg.toString()`

eng/tools/lint-diff/src/lint-diff.ts
```diff
@@ -67,13 +67,13 @@ export async function main() {
...
   }
 
   // TODO: Handle trailing slashes properly
   if (!afterArg || !(await pathExists(afterArg as string))) {
     validArgs = false;
-    console.log(`--after must be a valid path. Value passed: ${afterArg || "<empty>"}`);
+    console.log(`--after must be a valid path. Value passed: ${inspect(afterArg) || "<empty>"}`);
```
danieljurek, line 76: I wonder if these should be `afterArg.toString()`

danieljurek (approved):

eng/tools/lint-diff/src/lint-diff.ts
danieljurek: Ah OK, one would have to do `(beforArg as string)` vs `inspect(beforeArg)` 

image


The outputs are marginally different. It's actually a benefit that `inspect` wraps the string in single quotes. This is an improvement.

PR #38936, opened 2025-11-21
[TypeSpec Validation] Verify compile output folder

mikeharder:


timotheeguerin (approved):

eng/tools/typespec-validation/test/compile.test.ts
```diff
@@ -11,8 +11,9 @@ import { CompileRule } from "../src/rules/compile.js";
 
 import * as utils from "../src/utils.js";
 
-const swaggerPath = "data-plane/Azure.Foo/preview/2022-11-01-preview/foo.json";
-const handwrittenSwaggerPath = "data-plane/Azure.Foo/preview/2021-11-01-preview/foo.json";
+const swaggerPath = "specification/foo/data-plane/Azure.Foo/preview/2022-11-01-preview/foo.json";
+const handwrittenSwaggerPath =
+  "specification/foo/data-plane/Azure.Foo/preview/2021-11-01-preview/foo.json";
```
qiaozha, line 16: We don't really allow to have dot in the service name folder. As dot is to indicate there's a RP namespace, but there's no rp namespace concept in data plane and service name should not have dots in the middle.

eng/tools/typespec-validation/test/compile.test.ts
```diff
@@ -110,6 +111,21 @@ describe("compile", function () {
     );
   });
 
+  it("should throw if output swaggers are not under expected folder - v1", async () => {
```
qiaozha, line 114: I am a little confused, isn't the original issue a v2 folder structure?

qiaozha: And the output folder was under specification/org, it's just not under the specification/org/resource-manager/RPNS/service folder?

PR #39017, opened 2025-12-01
[path.js] Add untilLastSegmentWithParent()

mikeharder:


.github/shared/src/path.js
```diff
@@ -17,11 +17,11 @@ export function includesSegment(path, segment) {
 /**
  * @param {string} path Absolute or relative path
  * @param {string} segment File or folder
- * @returns {string} Portion of resolved path up to (but excluding) the last occurrence of segment
+ * @returns {string} Portion of resolved path up to (and including) the last occurrence of segment
```
mikeharder, line 20: Better to include segment.  Caller can easily remove segment using `path.dirname()`, but adding the segment back is more costly.  Also more consistent with `untilLastSegmentWithParent()`, which **must** include the last segment, since the caller doesn't know it in advance.

eng/tools/oav-runner/src/runner.ts
```diff
@@ -170,7 +170,7 @@ export async function processFilesToSpecificationList(
       */
       const swaggerDir = path.relative(
         rootDirectory,
-        untilLastSegment(absoluteFilePath, "examples"),
+        path.dirname(untilLastSegment(absoluteFilePath, "examples")),
```
mikeharder, line 173: react to breaking change, where `untilLastSegment()` now includes the trailing segment

PR #39061, opened 2025-12-03
WIP kv structure

catalinaperalta:


specification/keyvault/Security.KeyVault.Certificates/routes.tsp
```diff
@@ -2,7 +2,7 @@ import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-client-generator-core";
 import "@typespec/rest";
 import "./models.tsp";
-import "../Security.KeyVault.Common/common.tsp";
+import "./common.tsp";
```
chlowell, line 5: Do we have to duplicate common.tsp (Administration projects don't)?

catalinaperalta: Closing this since it was just an example to share with others, but yes with the v2 structure shared files are no longer recommended among different specs. I got this information from Jeff directly when I started migrating some other dp specs to the v2 structure.

PR #39104, opened 2025-12-05
[TSP Migration][providerhub] TypeSpec migrated from swagger

melina5656:


specification/providerhub/ProviderHub.Management/back-compatible.tsp
```diff
@@ -0,0 +1,97 @@
...
+);
+
+@@clientName(NotificationRegistrations.createOrUpdate::parameters.resource,
+  "properties"
+);
+
+@@clientLocation(OperationsPutContents.listByProviderRegistration,
+  "Operations"
+);
+@@clientLocation(OperationsPutContents.createOrUpdate, "Operations");
```
XiaofeiCao:
```suggestion
@@clientLocation(OperationsPutContents.createOrUpdate, Operations);
```

TCGC will report diagnostic if "Operations" interface already exists, and won't do anything else.
Check for other places as well.

specification/providerhub/ProviderHub.Management/models.tsp
```diff
@@ -0,0 +1,4766 @@
...
+
+  /**
+   * The canary traffic region configuration.
+   */
+  canary?: DefaultRolloutSpecificationCanary;
+
+  /**
+   * The low traffic region configuration.
+   */
+  lowTraffic?: TrafficRegionRolloutConfigurationProperties;
```
jliusan: image
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L3300
It is not consistent with the swagger, the swagger is `TrafficRegionRolloutConfiguration`, and here suggest using named model instead of anonynous model

specification/providerhub/ProviderHub.Management/models.tsp
```diff
@@ -0,0 +1,4766 @@
...
+  /**
+   * The resource provider authorization rules.
+   */
+  resourceProviderAuthorizationRules?: ResourceProviderAuthorizationRules;
+
+  /**
+   * The dsts configuration.
+   */
+  dstsConfiguration?: DstsConfigurationProperties;
+
```
jliusan, line 1685: The old is `dstsConfiguration` in swagger https://github.com/Azure/azure-rest-api-specs/blob/main/specification/providerhub/resource-manager/Microsoft.ProviderHub/stable/2024-09-01/providerhub.json#L5721,and there are so many issues like this : changed to xxxProperties in tsp file, this makes many breaking changes in `go`

specification/providerhub/ProviderHub.Management/back-compatible.tsp
melina5656: Fixed, please review.

specification/providerhub/ProviderHub.Management/models.tsp
melina5656: Fixed, please review

specification/providerhub/ProviderHub.Management/models.tsp
melina5656: Fixed, please review

specification/providerhub/ProviderHub.Management/back-compatible.tsp
```diff
@@ -0,0 +1,98 @@
...
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(OperationsContent.properties,
+  "!python,!js"
+);
+
+@@clientName(CustomRollouts.createOrUpdate::parameters.resource, "properties");
+
+@@clientName(ProviderRegistrations.createOrUpdate::parameters.resource,
+  "properties"
+);
+@@clientLocation(ProviderRegistrations.generateManifest, "");
```
XiaofeiCao: TCGC does not support `@clientLocation` with empty string. 
https://github.com/Azure/typespec-azure/blob/6b69e80faf1d99661804defdfd213ecfcd32f2e6/packages/typespec-autorest/src/utils.ts#L88

Seems currently you are hacking to make Swagger correct:
Image
https://github.com/Azure/typespec-azure/blob/8720acdd04576197a6bb7cfbf10d3c97dfbcb218/packages/typespec-autorest/src/utils.ts#L79-L82

However, SDK emitters will be generating wrong code.

/cc @tadelesh

XiaofeiCao: I'm changing it to `@@clientLocation(ProviderRegistrations.generateManifest, Microsoft.ProviderHub);`.
Seems working for both Swagger and Java emitter.

specification/providerhub/ProviderHub.Management/DefaultRollout.tsp
```diff
@@ -0,0 +1,142 @@
...
+   * Gets a new region frontload release.
+   */
+  get is NewRegionFrontloadReleaseOps.Read<DefaultRollout>;
+
+  /**
+   * Creates or updates a new region frontload release.
+   */
+  createOrUpdate is NewRegionFrontloadReleaseOps.CreateOrUpdateSync<
+    FrontloadPayload,
+    Response = DefaultRollout | ArmCreatedResponse<DefaultRollout>
```
jliusan: `ArmCreatedResponse` used here caused go new field added  `- New field `RetryAfter` in struct `NewRegionFrontloadReleaseClientCreateOrUpdateResponse`, since ArmCreatedResponse is defined like this 
```
/**
 * The Azure Resource Manager 201 response for a resource
 * @template ResponseBody The contents of the response body
 * @template ExtraHeaders Additional headers in the response. Default includes Retry-After header
 *
 * @example
 * ```typespec
 * @post
 * op post(...ResourceInstanceParameters<Employee>): ArmCreatedResponse<
 *   Employee,
 *   ExtraHeaders = {@header("x-ms-client-request-id") clientRequestId: string}
 * >;
 * ```
 */
@doc("Azure create operation completed successfully.")
model ArmCreatedResponse<
  ResponseBody,
  ExtraHeaders extends TypeSpec.Reflection.Model = Azure.Core.Foundations.RetryAfterHeader
> {
  ...CreatedResponse;
  ...ExtraHeaders;
  ...Body<ResponseBody>;
}

```

pshao25 (approved):

PR #38878, opened 2025-11-19
refactor(purview): migrate to unified folder structure

zedy-wj:
This PR migrates the **purview** specification to the new unified folder structure according to Azure folder structure guidelines.

## Changes:
- Consolidated TypeSpec and Swagger files under
```
purview/
├── data-plane
│   ├── account
│   ├── catalog
│   ├── datamap
│   ├── devopspolicies
│   ├── metadatapolicies
│   ├── pds
│   ├── scanning
│   ├── selfservicepolicies
│   ├── share
│   └── workflow
└── resource-manager/Microsoft.Purview/Purview
```
- Added `suppressions.yaml` files.
- Updated `tspconfig.yaml` with correct paths.
- Fixed input-file paths in `readme.md`
- Verified TypeSpec compilation works correctly

## Benefits:
- Unified structure for better maintainability
- Aligns with Azure folder structure standards
- Maintains backward compatibility
- Supports future TypeSpec migration efforts

PR #39110, opened 2025-12-06
fix-reference-problem

qiaozha:
This PR is to fix the file reference mistake caused by https://github.com/Azure/azure-rest-api-specs/pull/38878


qiaozha:
Added ARMSignedOff label as this is just update the correct input file references.

mikeharder (approved):

Issue #38964, opened 2025-11-25
[tsv validation] fails for emitter-output-dir when using placeholder like {package-name}


skywing918:
**Current issue**: 
In tspconfig, the emitter-output-dir validation does not pass even when package-name is set correctly to Azure.AI.Projects. The issue occurs because the validation checks the placeholder {package-name} literally instead of resolving it to the actual value.
tspconflig.yaml
```
"@typespec/http-client-csharp":
  emitter-output-dir: "{output-dir}/{service-dir}/{package-name}"
  package-name: Azure.AI.Projects
  api-version: "v1"
 ```

**Expected behavior**:
Validation should succeed after replacing {package-name} with the actual value (Azure.AI.Projects).

**How to Fix**:
add resolveVariables method resolves template variable references (in the format {variableName}) within configuration strings by recursively substituting them with their actual values from multiple configuration sources.

PR #39142, opened 2025-12-09
[TSV] Add new rule for python configuration validation for mgmt SDKs

msyyc:
For python, inconsistency between `emitter-output-dir` and `namespace` will cause unexpected error (e.g. [here](https://github.com/Azure/azure-rest-api-specs/pull/39076/commits/b3228cf479643fc03a6a373e3aaf57231101cdd8)) so we need add a new rule for python.


eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -576,6 +576,99 @@ export class TspConfigPythonMgmtNamespaceSubRule extends TspconfigEmitterOptions
   }
 }
 
+export class TspConfigPythonNamespaceMatchesEmitterOutputDirSubRule extends TspconfigSubRuleBase {
```
MaryGao: Is this rule to validate the namespace is matching with last segment of emitter-output-dir?

eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -576,6 +576,99 @@ export class TspConfigPythonMgmtNamespaceSubRule extends TspconfigEmitterOptions
...
+    if (derivedNamespace !== namespace)
+      return this.createFailedResult(
+        `The value of options.${this.emitterName}.namespace "${namespace}" does not match the value derived from options.${this.emitterName}.emitter-output-dir "${derivedNamespace}"`,
+        `Please update "options.${this.emitterName}.namespace" to "${derivedNamespace}" or adjust "options.${this.emitterName}.emitter-output-dir" so the derived namespace matches`,
+      );
+
+    return { success: true };
+  }
+
+  private resolveLastPathSegment(
```
MaryGao: @skywing918  Can we share a helper to get the last segment of emitter-output-dir?

I think we need to share this logic and also in that logic to fix our existing bug:https://github.com/Azure/azure-rest-api-specs/issues/38964.

MaryGao (changes requested):
As offline discussed, @skywing918 pls have a pr to prepare the helper first so that in this pr we could leverage the same helper and no need to duplicate it.

MaryGao (approved):
LGTM.

msyyc:
@raych1 Need your approval to merge this PR which updates protected files.

MaryGao (approved):

PR #39163, opened 2025-12-10
[TSV] share a helper to get the last segment of emitter-output-dir

### LINKED ISSUE #38964: [tsv validation] fails for emitter-output-dir when using placeholder like {package-name}

skywing918:
fixes https://github.com/Azure/azure-rest-api-specs/issues/38964
1. resolveVariables method resolves template variable references (in the format {variableName}) within configuration strings by recursively substituting them with their actual values from multiple configuration sources.

```
  "@azure-tools/typespec-ts":
    package-name: "arm-aaa-bbb"
    namespace: "{package-name}"
```

2. share a helper(getPackageNameFromEmitterOutputDir) to get the last segment of emitter-output-dir, the rule like:
```
export class TspConfigPythonNamespaceMatchesEmitterOutputDirSubRule extends TspconfigEmitterOptionsSubRuleBase {
  constructor() {
    super("@azure-tools/typespec-python", "emitter-output-dir", "derived from emitter-output-dir");
  }
  protected skip(_: any, folder: string) {
    return skipForDataPlane(folder);
  }
  protected validate(config: any): RuleResult {
    let resolvedSegmentResult = this.getPackageNameFromEmitterOutputDir(config);
    console.log("Derived emitter-output-dir:", resolvedSegmentResult.resolved);
    const derivedNamespace = resolvedSegmentResult.resolved.replace(/-/g, ".");
    const namespaceOption = this.tryFindOption(config, "namespace");
    const namespace = namespaceOption as unknown as undefined | string;
    if (derivedNamespace !== namespace)
      return this.createFailedResult(
        `The value of options.${this.emitterName}.namespace "${namespace}" does not match the value derived from options.${this.emitterName}.emitter-output-dir "${derivedNamespace}"`,
        `Please update "options.${this.emitterName}.namespace" to "${derivedNamespace}" or adjust "options.${this.emitterName}.emitter-output-dir" so the derived namespace matches`,
      );

    return { success: true };
  }
}
```


eng/tools/typespec-validation/src/rules/sdk-tspconfig-validation.ts
```diff
@@ -118,16 +118,127 @@ class TspconfigEmitterOptionsSubRuleBase extends TspconfigSubRuleBase {
...
-    for (const segment of this.keyToValidate.split(".")) {
+    for (const segment of key.split(".")) {
       if (option && typeof option === "object" && !Array.isArray(option) && segment in option)
         option = option![segment];
       else return undefined;
     }
     return option;
   }
 
+  protected resolveVariables(value: string, config: any): { resolved: string; error?: string } {
```
MaryGao: config: Record<string, any>

MaryGao: move this helper as a function.

MaryGao (approved):

MaryGao (approved):

raych1:
@MaryGao you might create a PR to exclude `sdk-tspconfig-validation` related files from the protected file list at [here](https://github.com/Azure/azure-rest-api-specs/blob/213be191bef887e148715fc1b767ecda6fac68d2/.github/workflows/protected-files.yaml#L43) so that the future changes on this rule changes won't need to be force merged. CC: @mikeharder

PR #39177, opened 2025-12-10
fleet: add .net sdk config

ryanzhang-oss:
Add the .net SDK auto-generation configuration for fleet


maririos (approved):

maririos:
although not blocking, .NET SDK validation is failing. 
You should take a look at the logs and if questions follow the guidance on the error:

> [ERROR] Failed to build the sdk project: Azure.ResourceManager.Containerservicefleet for service: containerservicefleet. Exit code: False. Please review the detail errors for potential fixes. If the issue persists, contact the DotNet language support channel at https://aka.ms/azsdk/dotnet-teams-channel and include this spec pull request.

serbrech (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/tspconfig.yaml
```diff
@@ -24,6 +24,9 @@ options:
     generate-test: true
     generate-sample: true
     flavor: azure
+  "@azure-typespec/http-client-csharp-mgmt":
+    namespace: "Azure.ResourceManager.Containerservicefleet"
```
ArcturusZhang, line 28: should this be
```suggestion
    namespace: "Azure.ResourceManager.ContainerServicefleet"
```
windows might ignore cases but since our agents are running on linux, in which case, they are case sensitive on directories

ArcturusZhang:
Also we should revert this PR because your corresponding library is currently generating from swagger.
The migration of generator is a complicated process because .net libraries accept no breaking changes. Therefore our team is working on this migration.

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/tspconfig.yaml
ArcturusZhang: and the directory here is also incorrect.

PR #39225, opened 2025-12-12
Cleaned up obsolete content related to the old code generation

### LINKED ISSUE #13292: Remove/update Readme template for SDK configuration
- Our current Readme template has out-of-date information about SDK generation: https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/samplefiles/README.md#code-generation
- Talks about generating SDKs using AutoRest: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/contosowidgetmanager/resource-manager#getting-started

raych1:
Fixed https://github.com/Azure/azure-sdk-tools/issues/13292

CC: @lmazuel


documentation/onboard-dpg-in-sdkautomation/java/README.md
weidongxu-microsoft, line 1: This file is valid guide. May I know which file user should read instead (after it be deleted)?

raych1: This file seems to include non-dpg scenarios which conflicts with the folder name. I'm curious how users are discovering and navigating to this file today. Since we already have sample typespec projects for both mgmt-plane and data-plane, could we direct users to those instead?

weidongxu-microsoft: Fine for me.

Just saying an example tspconfig.yaml usually cannot convey the full information in a md file (which can provide better information).
(though the real situation typically is that service just copy another tspconfig.yaml from either their existing service to a new service, or from a nearby service to a new service -- and that applies to wrong situation -- they didn't even use the contoso example :-( )

weidongxu-microsoft (approved):

documentation/onboard-dpg-in-sdkautomation/java/README.md
raych1: We might have a lot of users copying it from other services. Meanwhile, Contoso sample projects are exposed to users in multiple places, such as the docs on `engineering hub` site, SDK automation logs, and chat bot answers. I believe that's pretty much of what we can do to guide the new users.

PR #39256, opened 2025-12-15
Update emitter output directories in tspconfig.yaml for Confidential Ledger and Container Service

### LINKED ISSUE #10991: Support emitter-output-dir option in the SDK language scripts
We would like to replace `package-dir` by `emitter-output-dir` in the emitter options as follows:

Image

Each SDK language script will require to support `emitter-output-dir`. If it needs to understand the package name, it should parse the value out from `emitter-output-dir`.

skywing918:
fixes https://github.com/Azure/azure-sdk-tools/issues/10991


MaryGao (approved):

MaryGao:
The original config is out-of-date and we need to swith to emitter-output-dir format. @jim-minter and @matthchr and @serbrech Could you approve this from your side since you are code owner.

MaryGao (approved):

PR #39265, opened 2025-12-15
Rename body

xiangyan99:
Rename the body name `dataSource` to `dataSourceConnection`
Add default values for accept header
Fix the ArrayEncoding to use string[] rather than string
Update client.tsp for SDK customizations.


yangylu91:
Search/API Foundations team has reviewed the PR, confirmed all the changes in the PR expected, and we can merge the PR.

1. Renamed body parameter for DataSources_Create: dataSource -> dataSourceConnection and added related clientName in routes tsp
2. Updated type of parameters marked with "@encode(ArrayEncoding.*" to array, such as string -> string[], RegexFlags -> RegexFlags[].
3. Updated client.tsp for further customizations to make Python SDKs backward compatible.
4. Updated example accordingly
5. Added default value for accept headers. The tsp team is adding features to support client default value. This is a workaround when the feature is not ready.
6. Mark the query parameter "select" as string[]. The original swagger missed the information and Xiang fixed that.

giulianob (approved):

PR #39325, opened 2025-12-18
fix JS flatten breakings for avs

kazrael2119:
https://github.com/Azure/azure-sdk-for-js/pull/36877
fix JS flatten breakings for avs

### DIFF
--- a/specification/vmware/resource-manager/Microsoft.AVS/AVS/clusters.tsp
+++ b/specification/vmware/resource-manager/Microsoft.AVS/AVS/clusters.tsp
@@ -102,7 +102,7 @@ model ClusterUpdate {
 
   #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Legacy flatten"
   @doc("The properties of a cluster resource that may be updated")
-  @Azure.ClientGenerator.Core.Legacy.flattenProperty
+  @Azure.ClientGenerator.Core.Legacy.flattenProperty("!javascript")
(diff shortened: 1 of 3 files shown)


cataggar (approved):

MaryGao (approved):

specification/vmware/resource-manager/Microsoft.AVS/AVS/clusters.tsp
```diff
@@ -102,7 +102,7 @@ model ClusterUpdate {
 
   #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Legacy flatten"
   @doc("The properties of a cluster resource that may be updated")
-  @Azure.ClientGenerator.Core.Legacy.flattenProperty
+  @Azure.ClientGenerator.Core.Legacy.flattenProperty("!javascript")
```
MaryGao, line 105: Generally these decorators should be in `client.tsp` file and this would be easy for us to understand this is only client behavior and no need for ARM review.

@josefree Do we have any agreement that no need for `ARMReview` if any changes in `client.tsp` ?

josefree: yes, I helped add the label "tsp-updates-impact-sdk-only"

MaryGao:
Added `ARMSignedOff` label since it is sdk impact only.

image

PR #39326, opened 2025-12-18
Revert change as issue fixed for JS in back-compatible.tsp 

v-jiaodi:
Resolve comments from https://github.com/Azure/azure-rest-api-specs/pull/38384#discussion_r2609487040.

Revert JS exclusion logic in back-compatible.tsp which is a similar file to `client.tsp` but used to save any client refinement information to ensure no breaking change introduced from swagger to TypeSpec.


MaryGao (approved):

MaryGao:
Merged since it would only impact the JS SDK and no need for ARM review.

PR #39331, opened 2025-12-18
Remove unsupported SQL Pool Metadata Sync Configs pages

Tikii:
Update public documentation to reflect the current feature state of the MDSync service in production.

A customer deployed a dedicated SQL pool using an ARM template and attempted to enable the MDSync service. This triggered a failure in the associated Web controller, causing the deployment to fail. Because the current documentation implies that these APIs are functional, the customer had no indication that the MDSync service is not supported for dedicated SQL pools and that these APIs are non-operational.

Remove unsupported SQL Pool Metadata Sync Configs pages from 2021‑06‑01, 2021‑06‑01‑preview, 2019-06-01-preview, 2021-04-01-preview, 2020-12-01, 2021-03-01 and 2021-05-01 

**2021-06-01**
- https://learn.microsoft.com/en-us/rest/api/synapse/resourcemanager/sql-pool-metadata-sync-configs?view=rest-synapse-resourcemanager-2021-06-01
- https://learn.microsoft.com/en-us/rest/api/synapse/resourcemanager/sql-pool-metadata-sync-configs/create?view=rest-synapse-resourcemanager-2021-06-01&tabs=HTTP
- https://learn.microsoft.com/en-us/rest/api/synapse/resourcemanager/sql-pool-metadata-sync-configs/get?view=rest-synapse-resourcemanager-2021-06-01&tabs=HTTP


razvanbadea-msft:
The avocado errors shows that 

1. that path removed is still in 2021-05-01 version
"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Synapse/workspaces/{workspaceName}/sqlPools/{sqlPoolName}/metadataSync/config" . Is this what you want to obtain? 
2. you are combining input files from different versions. You  should copy the files referenced from other version in the one used as default

ravimeda:
Please address the required checks that are failing.

specification/synapse/resource-manager/Microsoft.Synapse/preview/2019-06-01-preview/sqlPool.json
```diff
@@ -459,129 +459,6 @@
...
-            "required": true
-          }
-        ],
-        "responses": {
-          "404": {
-            "description": ""
-          },
-          "200": {
-            "schema": {
-              "$ref": "#/definitions/MetadataSyncConfig"
```
qiaozha, line 563: Could you check if this model is still referenced by other APIs?

qiaozha (commented):
Please check if there's any orphan models left after the non-functional APIs being removed.

specification/synapse/resource-manager/Microsoft.Synapse/preview/2019-06-01-preview/sqlPool.json
Tikii: Removing MetadataSyncConfig as it is no longer referenced by any APIs.

Tikii:
> Please address the required checks that are failing.

As part of this cleanup, CI is now flagging Swagger ModelValidation (MULTIPLE_API_VERSION) errors in the older specs.
The open question is whether the remaining ModelValidation errors in the older versions should be handled via suppression, or if they should be addressed differently. From what I can tell, those errors are not correlated with my change.

qiaozha:
@Tikii ModelValidation errors have nothing to do with the Avocado multiple API version failure, they are two different checks. 
image
You need to look into the details to see if they are really irrelevant to your change.

ravimeda:
Make sure to decommission the RT `metadataSync/config`. Refer to https://eng.ms/docs/products/arm/rpaas/decommissioning/resource_type_deprecation

ravimeda:
Model validation reference is at https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/Semantic-and-Model-Violations-Reference.md.

For example, the error code _LRO_RESPONSE_HEADER_ in "specification/synapse/resource-manager/Microsoft.Synapse/preview/2019-06-01-preview/examples/CreateSqlPoolRestorePoints.json" is highlighting the missing header in the response. See https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/Semantic-and-Model-Violations-Reference.md#LRO_RESPONSE_HEADER

> Long running operation must return Operation-Id, Operation-Location, or azure-AsyncOperation header in response when this operation is annotated with x-ms-long-running-operation:true.

Please update the examples accordingly.

Tikii:
> Make sure to decommission the RT `metadataSync/config`. Refer to https://eng.ms/docs/products/arm/rpaas/decommissioning/resource_type_deprecation

Thanks for your feedback!
To clarify: the /metadataSync/config API was never functional and was unintentionally exposed in the spec. There is no backend implementation and no customer traffic. This change simply removes the spec artifact and ensures the resource path isn’t present in any documentation or SDKs.
Is formal deprecation process necessary, as there isn't any customer usage or support for this API.
Please let me know if anything further is required.

Tikii:
The failures are on unchanged files and were present before this PR. Would an exception be acceptable so we can keep this PR focused and handle those issues separately?

**Urgency for merging this PR:**
We had an incident (https://portal.microsofticm.com/imp/v5/incidents/details/501030850/summary) where the customer used an ARM template to deploy a dedicated SQL pool and enable the MDSync service. This action caused the associated Web controller to throw an error and fail the deployment. Due to the current state of the documentation, the customer is unaware that these APIs are non-operational and that the MDSync service is not supported on a dedicated SQL pool.

inikitovic (approved):

wanyang7 (approved):

PR #39335, opened 2025-12-18
[Foundry] Adding Foundry specification.

glecaros:
Replaced by https://github.com/Azure/azure-rest-api-specs/pull/39565


dargilco (approved):

package.json
```diff
@@ -12,6 +12,7 @@
     "@azure-tools/typespec-client-generator-cli": "0.31.0",
     "@azure-tools/typespec-client-generator-core": "0.63.3",
     "@azure-tools/typespec-liftr-base": "0.10.0",
+    "@azure-tools/openai-typespec":"^0.1.9",
```
```suggestion
    "@azure-tools/openai-typespec":"0.1.9",
```

All versions `0.x.y` should be pinned, since semver allows breaking changes beween any `0.x` versions.  Also, we prefer to pin references to our own packages and update manually, to avoid any confusion about the version used at runtime.

mikeharder (changes requested):
pin dependency version

package.json
dargilco: @glecaros on the Python side I pinned it to 0.1.11 (the latest version at the time of writing). Let's do the same here.

PR #39367, opened 2025-12-22
[TSV] Move suppressions from the root to service folder for sdk validation rules

### LINKED ISSUE #13427: [TSV] move the root suppressions to service folder in main
move the root to service folder

skywing918:
fixes https://github.com/Azure/azure-sdk-tools/issues/13427


specification/edge/suppressions.yaml
```diff
@@ -0,0 +1,9 @@
+
+- tool: TypeSpecValidation
+  reason: Suppress until spec can be updated
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-csharp.*
```
MaryGao: remove csharp considering it is optional.

MaryGao (approved):

mikeharder (changes requested):
ensure all files end with a trailing newline

specification/apicenter/suppressions.yaml
```diff
@@ -0,0 +1,11 @@
+- tool: TypeSpecValidation
+  reason: Suppress until spec can be updated
```
mikeharder: Can the `reason` string be improved, to give spec owners more information about who is responsible for doing what, and by when?  For example, if these are suppressed for a reason like "spec has not yet been converted to generate SDKs from TSP", could we use something like this for the `reason`?

```
reason: >
  Spec not yet converted to generate SDKs from TSP.
  In next API version, this suppression *must* be removed and
  `tspconfig.yaml` must be updated.
  More info: https://aka.ms/azsdk/sdk-tspconfig-validation
```

Or, if these specs are waiting to be converted by the Azure SDK team, put that in the reason, maybe with an estimated date of fixing?

The current reason just says `until spec can be updated`, but this doesn't really mean anything.  Any suppression exists until it's updated.

mikeharder: And if we want to enforce that these suppressions are fixed by a certain criteria (like next API version), we will probably need to add a PR check to block if this criteria isn't met.

mikeharder (changes requested):
blocking for discussion on reason of suppressions

specification/apicenter/suppressions.yaml
MaryGao: I am also taking a close look at these suppressions to see whether we can refine them. It appears that there is limited room to improve the emitter configurations themselves, since most suppressions are added when we first introduced emitter suppressions. So I agree that refining the suppression reasons is an area where we can make improvements.

Focusing on the management‑related suppressions([detailed data](https://microsoftapc-my.sharepoint.com/:x:/g/personal/v-zhanh_microsoft_com/IQCNuhykN72mS4ayezLNYb2sAaYQwEg-V7_jVDiLCu1eUDg?e=ozP1R4)), there are around 30 entries with several different reasons, affecting roughly 150 SDKs. Below are the top three categories:


- 1. “**Suppress until we know the awsconnector folder uses these settings”**
This accounts for only one suppression entry but impacts 113 SDKs, limited to services under the awsconnector folder.


- 2. **“Suppress until spec can be updated”**
There are more than 20 entries here, affecting around 30 SDKs. 

As far as I know, item 1 and 2 were added by us when we first introduced the emitter‑suppression concept. Over time, the wording may have changed and original pr is [here](https://github.com/Azure/azure-rest-api-specs/pull/35259).


- 3. **“Remove specs containing incorrect or missing emitter options or parameters”**
This includes around four entries, affecting four Compute SDKs. These were added by the SDK team during the Compute spec migration — so they also originated from our side.


Given that these categories were added by us for specific reasons, I think they are good initial candidates for refinement if we’re looking to improve clarity and consistency.

image

mikeharder: I agree the three categories highlighted above are all "placeholder" messages, and we should improve them.  Either to all use the same message (if all the suppressions exist for basically the same reason), or to messages specific to each service (or org).

Do either of these existing messages apply to the specs we want to improve?

* SDK is not generated from TSP config file
  * However, if this is the case, should the rule be able to detect this without a suppression?  What does it mean for an SDK to be (or not to be) generated from the TSP config file?
* Specification is rest-only and does not generate SDKs
  * For example, I believe this is true for `awsconnector`
  * But again, if rest-only specs are allowed in the repo, they should be indicated as such by **configuration**, not suppressions

We put `awsconnector` in a special bucket, due to the number of SDKs, and because we think the same decision(s) should be made consistently for everything under the `awsconnector` folder.

specification/workloads/suppressions.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - Workloads.Operations.Management/tspconfig.yaml
+
+- tool: TypeSpecValidation
+  reason: Suppress until spec can be updated
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-java.*
+  paths:
+    - Workloads.SAPDiscoverySite.Management/tspconfig.yaml
+    - Workloads.SAPMonitor.Management/tspconfig.yaml
```
weidongxu-microsoft:
```suggestion

```

Java dev has started to [complete java emitter config for released lib](https://github.com/Azure/azure-rest-api-specs/pull/39466/files#diff-e698494d82d0faf0f042fd4c23f66e740066211812d6f64525a7dbfab7ba7b64). If you have update to this PR, you can remove this one now (after refresh main).

Otherwise, Java dev can delete this by follow-up PR.

specification/apicenter/suppressions.yaml
MaryGao: @mikeharder I wonder if we could narrow this [public/main pr](https://github.com/Azure/azure-rest-api-specs/pull/39367) and another [private/rpassmaster pr](https://github.com/Azure/azure-rest-api-specs-pr/pull/26143) to the scope of suppression movement and suppression message refinement. We have summerized all suppression messages and have proposed new reasons for [public/main](https://microsoftapc-my.sharepoint.com/:x:/g/personal/v-zhanh_microsoft_com/IQBdMEMD1ZVmTpwhTXmeBjMDAY1Ae2anYJFgtu0v3Ai0cyY?e=wjByx8) and [private/rpassmaster](https://microsoftapc-my.sharepoint.com/:x:/g/personal/v-zhanh_microsoft_com/IQCo7cqLcJLzTI4Aw-jxrrvXAYXvuK0Vkru_M3158sTB1Jw?e=8q3a9P), could you help review that?

Regarding more discussions on if we should have some rules or configuration support on two above categories, I'd like to have separated issue to address that. 

image

lirenhe:
@mikeharder, are you OK with the proposed suppression Reason in the Summary page from the above excel?  Please let us know you suggestion so we could proceed with this PR.  thanks

mikeharder (approved):
The expanded suppression messages look good.

weidongxu-microsoft:
@v-huizhu2 

Java updated tspconfig on those libs that we know what Java namespace is. These are all merged.

Please update the PR to remove the now unnecessary suppressions on
```
- options.@azure-tools/typespec-java.*
```

MaryGao (approved):

PR #39371, opened 2025-12-23
[cognitiveservices] Normalize Swagger Model Names and Properties Suffix

melina5656:
Rename Swagger Models: *Resource → * and * → *Properties; Update References.


pshao25:
This PR is a prerequisite of TypeSpec conversion. It doesn't have any API perspective change. It only changes client names. We've also got confirmation with service team that they want to proceed.
image

rkuthala (approved):

PR #39453, opened 2026-01-05
Fail the 'generate sdk' step when 'npm ci' fails

### LINKED ISSUE #39354: Generate SDK step should fail when 'npm ci' fails
Currently the `Generate SDK` step passes even it fails to run `npm ci` in the very beginning.
 
Image

[Sample repro pipeline run](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5689382&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692)

raych1:
Resolved https://github.com/Azure/azure-rest-api-specs/issues/39354

[Test Pipeline Run](https://dev.azure.com/azure-sdk/public/_build/results?buildId=5720284&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692)

### DIFF
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -206,6 +206,7 @@ stages:
           versionSpec: $(PythonVersion)
 
       - script: |
+          set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
           optional_params=""
           sdk_gen_info="Configurations: "
 
@@ -246,10 +247,6 @@ stages:
           cd $(SpecRepoDirectory)
           echo "##[group]Run npm ci"
           npm ci
-          if [ "$?" -ne 0 ]; then
-            echo "##vso[task.logissue type=error]npm ci failed."
-            exit 1
-          fi
           echo "##[endgroup]"
           node $(SpecRepoDirectory)/eng/tools/spec-gen-sdk-runner/cmd/spec-gen-sdk-runner.js \
             --scp $(SpecRepoDirectory) \
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -246,6 +246,10 @@ stages:
           cd $(SpecRepoDirectory)
           echo "##[group]Run npm ci"
           npm ci
+          if [ "$?" -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "azure-rest-api-specs",
   "devDependencies": {
-    "@azure-tools/spec-gen-sdk": "~0.9.1",
+    "@azure-tools/spec-gen-sdk1": "~0.9.1",
```
raych1:
```suggestion
    "@azure-tools/spec-gen-sdk": "~0.9.1",
```

raych1: Revert test change.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -246,6 +246,10 @@ stages:
           cd $(SpecRepoDirectory)
           echo "##[group]Run npm ci"
           npm ci
+          if [ "$?" -ne 0 ]; then
```
mikeharder: Instead of checking for specific errors, you should configure bash to fail on **any** error.  Something like this at the top of the script.

```
set -euo pipefail
```

mikeharder (changes requested):
Instead of checking for specific errors, you should configure bash to fail on **any** error.  Something like this at the top of the script.

```
set -euo pipefail
```

weshaggard (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: We already have the code in the `spec-gen-sdk-runner.js` to handle the exception case and set the failure accordingly. Here we only need to take care of the step of `npm ci`.

mikeharder: I don't understand your objection.  Why would you **not** want this bash script to exit immediately on any error?  Ignoring errors just leads to confusion, as we found when `npm ci` failed.

Why only handle errors for `npm ci`, instead of all commands in this script?

raych1: I hesitated to use the `set -e` command because I didn't want to impact the `last command run result || true` result. With further looking it doesn't impact it.

mikeharder (approved):

PR #39477, opened 2026-01-06
Use "Tool" or "PreviewTool" as a suffix for Azure Tools

dargilco:
To align with OpenAI naming conventions, use "Tool" suffix for stable Azure tools:
* `AzureAISearchAgentTool` renamed to  `AzureAISearchTool`
* `OpenApiAgentTool` ==> `OpenApiTool`
* `AzureFunctionAgentTool` ==>  `AzureFunctionTool`

And use "PreviewTool" as suffix for Azure tools in preview:
* `SharepointAgentTool` ==> `SharepointPreviewTool`
* `BingCustomSearchAgentTool` ==> `BingCustomSearchPreviewTool`
* `BrowserAutomationAgentTool` ==> `BrowserAutomationPreviewTool`
* `A2ATool` ==> `A2APreviewTool`

Still waiting on a call regarding renaming `MemorySearchTool`.

Also fix some of the doc string for 'type', as it seems like these were not updated when the "_preview" suffix was added to the type string.

This is not a breaking change in REST APIs. This will be a breaking change in emitted beta SDKs, and we will update samples and docs accordingly.

### DIFF
--- a/specification/ai/Foundry/tools/models.tsp
+++ b/specification/ai/Foundry/tools/models.tsp
@@ -112,7 +112,7 @@ model FabricDataAgentToolParameters {
 /**
  * The input definition information for a Microsoft Fabric tool as used to configure an agent.
  */
-model MicrosoftFabricPreviewTool extends OpenAI.Tool {
+model MicrosoftFabricAgentTool extends OpenAI.Tool {
   /**
    * The object type, which is always 'fabric_dataagent_preview'.
    */
@@ -139,7 +139,7 @@ model SharepointGroundingToolParameters {
 /**
  * The input definition information for a sharepoint tool as used to configure an agent.
  */
-model SharepointPreviewTool extends OpenAI.Tool {
+model SharepointAgentTool extends OpenAI.Tool {
   /**
    * The object type, which is always 'sharepoint_grounding_preview'.
    */
@@ -258,7 +258,7 @@ model OpenApiTool extends OpenAI.Tool {
 /**
  * The input definition information for a Bing custom search tool as used to configure an agent.
  */
-model BingCustomSearchPreviewTool extends OpenAI.Tool {
+model BingCustomSearchAgentTool extends OpenAI.Tool {
   /**
    * The object type, which is always 'bing_custom_search_preview'.
    */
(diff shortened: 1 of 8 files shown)


lindazqli (approved):

specification/ai/Foundry/tools/models.tsp
```diff
@@ -84,9 +84,9 @@ model FabricDataAgentToolParameters {
 /**
  * The input definition information for a Microsoft Fabric tool as used to configure an agent.
  */
-model MicrosoftFabricAgentTool extends OpenAI.Tool {
+model MicrosoftFabricTool extends OpenAI.Tool {
```
trrwilson: We should consider whether we want to align and specialize the name with the updated discriminator value, e.g. here using `MicrosoftFabricPreviewTool`. This is what OpenAI ended up doing in having parallel [`WebSearchTool` (`web_search`)](https://github.com/microsoft/openai-openapi-pr/blob/d914ed7e79ef77c80fd9e4814169fec48823e0da/packages/openai-typespec/src/responses/models.tsp#L1613) and [`WebSearchPreviewTool` (`web_search_preview`)](https://github.com/microsoft/openai-openapi-pr/blob/d914ed7e79ef77c80fd9e4814169fec48823e0da/packages/openai-typespec/src/responses/models.tsp#L1907) components.

The disadvantages are that it's clunky and overall a bit weird; the advantages stem from it cleanly supporting concurrent/overlapping support that may include divergent breaking changes between preview and non-preview tool iterations. I'm in favor of aligning but that's a loosely held answer to a still-open question.

specification/ai/Foundry/tools/models.tsp
```diff
@@ -57,13 +57,13 @@ union _AgentIncludable {
...
-   */
+    /**
+     * The project connections attached to this tool. There can be a maximum of 1 connection
+     * resource attached to the tool.
+     */
     @added(Versions.v2025_11_15_preview)
     @removed(Versions.v1)
-    custom_search_configuration?: WebSearchConfiguration;
+    custom_search_configuration?: WebSearchConfiguration,
   }
```
dargilco, line 74: The above changes are unrelated to this PR. These are a result of the formatter in "npx tsv ." fixing the format from a previous commit to branch.

PR #39479, opened 2026-01-06
Switching to consume new OpenAI TypeSpec package.

glecaros:


dargilco (approved):

dargilco (approved):

package.json
```diff
@@ -12,6 +12,7 @@
     "@azure-tools/typespec-client-generator-cli": "0.31.0",
     "@azure-tools/typespec-client-generator-core": "0.63.3",
     "@azure-tools/typespec-liftr-base": "0.10.0",
+    "@azure-tools/openai-typespec":"^0.1.9",
```
MaryGao, line 15: @markcowl and @timotheeguerin what is our recommandation to share the common typespec? do we have limitation on where to put shared lib? what if typespec has breaking which impacted third party libs, who will be responsible for upgrading?

```
// import "../.external-readonly/openai.external.typespec/finetuning/custom.tsp";
import "@azure-tools/openai-typespec/models/fine-tuning";
```

/cc @weidongxu-microsoft @lirenhe @lmazuel

PR #39482, opened 2026-01-07
feat: add streaming logs functionality for agent containers and enhance container details models

ganeshyb:
# Choose a PR Template

Previous PR https://github.com/Azure/azure-rest-api-specs-pr/pull/26058

### DIFF
--- a/specification/ai/Foundry/agents/models.tsp
+++ b/specification/ai/Foundry/agents/models.tsp
@@ -336,6 +336,10 @@ model AgentContainerObject {
   @visibility(Lifecycle.Read)
   object: "agent.container";
 
+  @doc("The identifier of the container.")
+  @visibility(Lifecycle.Read)
+  id?: string;
+
   @doc("The status of the container of a specific version of an agent.")
   @visibility(Lifecycle.Read)
   status: AgentContainerStatus;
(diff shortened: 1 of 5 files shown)


specification/ai/Foundry/agents/models.tsp
```diff
@@ -359,6 +363,40 @@ model AgentContainerObject {
...
+model ContainerDetails {
+  @doc("The provisioning state of the container.")
+  provisioning_state: string;
+
+  @doc("The state of the container.")
+  state: string;
+
+  @doc("The health state of the container.")
+  health_state: string;
+
```
dargilco, line 385: For 'state' and 'health_state', are there fixed string values? Can this be described as an enum instead? (which also include the open ended string option so it can be extended anytime)

dargilco:
Thanks, please run "npx tsv ." if you have not already done so.

specification/ai/Foundry/agents/models.tsp
ganeshyb: These are retuned by ACA, so keeping it string for extensibility

specification/ai/Foundry/agents/models.tsp
```diff
@@ -446,3 +484,46 @@ model AcceptedAgentContainerOperation {
...
+  reason: string;
+
+  @doc("The source of the event.")
+  @encodedName("application/json", "EventSource")
+  event_source: string;
+
+  @doc("The count of occurrences of this event.")
+  @encodedName("application/json", "Count")
+  count: int32;
+}
```
ganeshyb: @glecaros, @trrwilson This is the json format returned by the Container log streaming, they return in pascal casing and we do not have control over the format see more there https://msdata.visualstudio.com/Vienna/_git/vienna?version=GBmaster&path=/src/azureml-api/src/Agents/docs/hosted-agent/logstream.md&_a=preview

trrwilson: Thanks, @ganeshyb! That makes an even stronger case for the proposal (parallel discussion) to omit these definitions here and make the operation definition open-ended. Let's get @johanste to confirm, but what I believe we can do is:

- Remove `EventContainerLog` entirely (it's out of our control anyway)
- Define the response type of `streamAgentContainerLogs` to be `Stream<string>` (or `Stream<StreamingContainerLog>` if we'd like to keep the scalar)
- Ensure there are doc string references attached to the operation about formats to process

johanste: It's a bit jarring to have two different formats for the logs. And if we want to expose jsonl, we should use the appropriate mime type for it. This was just for preview, correct?

ganeshyb: Yes its just for preview

ganeshyb: Removed the type and update the doc string to specify the details

trrwilson (approved):

PR #39485, opened 2026-01-07
add JS config for monitor

kazrael2119:
add readme.typescript.md for monitor service


qiaozha (approved):

qiaozha:
added Approved-Avocado and Approved-LintDiff and ARMSignedOff label as this change is just for JS SDK generation. It should not have any impacts on the API.

qiaozha (approved):

PR #39492, opened 2026-01-07
fix search config

JoshLove-msft:


specification/search/data-plane/Search/tspconfig.yaml
```diff
@@ -13,16 +13,10 @@ options:
...
-  "@azure-tools/typespec-csharp":
-    package-dir: "Azure.Search.Documents"
-    namespace: "Azure.Search.Documents"
-    clear-output-folder: true
-    model-namespace: false
-    flavor: azure
-    emitter-output-dir: "{output-dir}/{service-dir}/Azure.Search.Documents"
   "@azure-typespec/http-client-csharp":
     namespace: "Azure.Search.Documents"
     model-namespace: false
```
xiangyan99: Should we update this to true?

JoshLove-msft: It looks like all the models are in the Models subnamespace so yes.

jsquire: If the goal is to keep back-compat with the existing library, then we should.  Based on a random sampling _([example](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/search/Azure.Search.Documents/src/Models/FacetType.cs))_ it looks like the models are using the `.Models` namespace.

jsquire (approved):

jsquire (approved):

PR #39494, opened 2026-01-07
remove go emitter

xiangyan99:


specification/search/data-plane/Search/suppressions.yaml
```diff
@@ -52,4 +52,15 @@
...
+  paths:
+    # Specify the path to the tspconfig.yaml file where suppression should apply
+    - ./tspconfig.yaml
+  rules:
+    - SdkTspConfigValidation
+  sub-rules:
+    - options.@azure-tools/typespec-go.emitter-output-dir
+    - options.@azure-tools/typespec-go.module
+    - options.@azure-tools/typespec-go.containing-module
+  reason: the go emitter is not ready for Azure Search Data Plane service
```
raych1, line 66: Do you have a targeted timeline for Go? We might put it to the reason so that it can be reminded later to remove this suppression.

xiangyan99: Took offline.

raych1 (approved):

specification/search/data-plane/Search/suppressions.yaml
raych1: @tadelesh Go is not required for data plane SDK. I think we should make this rule as not required for data plane scenarios.

raych1: Merged this PR to unblock the work.

tadelesh: > @tadelesh Go is not required for data plane SDK. I think we should make this rule as not required for data plane scenarios.

We have already updated the rule. If you config Go, then SDK generation should pass, if not, no SDK validation.

PR #39500, opened 2026-01-07
[DNM] add snapshot api in azurestackhcivm

jgu17:
Add azurestackhcivm snapshot api specifications

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
       "name": "azure-rest-api-specs",
       "devDependencies": {
         "@autorest/openapi-to-typespec": "0.11.12",
-        "@azure-tools/openai-typespec": "1.6.2",
+        "@azure-tools/openai-typespec": "1.7.1",
         "@azure-tools/spec-gen-sdk": "~0.9.1",
         "@azure-tools/specs-shared": "file:.github/shared",
         "@azure-tools/typespec-apiview": "0.7.2",
@@ -791,9 +791,9 @@
       "link": true
     },
     "node_modules/@azure-tools/openai-typespec": {
-      "version": "1.6.2",
-      "resolved": "https://registry.npmjs.org/@azure-tools/openai-typespec/-/openai-typespec-1.6.2.tgz",
-      "integrity": "sha512-u2Y8ZS6OMO9lUbSyYgBiDuGQkF/ccv9qk6ipvDwu+vZmxRx2c9Zp2VOcC6dxWNOT62DWjlfmbJQy7E/fNaQVrw==",
+      "version": "1.7.1",
+      "resolved": "https://registry.npmjs.org/@azure-tools/openai-typespec/-/openai-typespec-1.7.1.tgz",
+      "integrity": "sha512-DoZLbQPgHyWKbIiUkCz2/9kKxWnZ/fZwvKGnjo3AJkZ8Si3QOHXR3JBp9LWiErsWPujg6k7NkTocqBXxih8pqQ==",
       "dev": true,
       "license": "MIT",
       "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "azure-rest-api-specs",
   "devDependencies": {
-    "@azure-tools/openai-typespec": "1.6.2",
+    "@azure-tools/openai-typespec": "1.7.1",
     "@azure-tools/spec-gen-sdk": "~0.9.1",
     "@azure-tools/specs-shared": "file:.github/shared",
     "@azure-tools/typespec-apiview": "0.7.2",


package-lock.json
shraddhasun, line 1: we shouldnt modify this file

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
```diff
@@ -1949,6 +1961,77 @@ model StorageContainersUpdateRequest {
...
+    },
+    {
+      type: "Microsoft.AzureStackHCI/snapshots";
+    }
+  ]>;
+
+  @doc("Unique ID of the source resource used for disk creation. Read-only and not required for disk creation.")
+  @visibility(Lifecycle.Read)
+  sourceUniqueId?: string;
+}
```
mentat9: Please elaborate on this property a bit, so it's clear what values it takes, where they come from, what scope it is unique on, etc. For example, if it's a guid, give it a format: uuid. Indicate in the doc when is it used, what it it used for, etc. If it's something else, indicate how it's created and what kind of string it is.

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
```diff
@@ -1949,6 +1961,77 @@ model StorageContainersUpdateRequest {
...
+@added(Versions.v2026_02_01_preview)
+@doc("Properties under the snapshot resource")
+model SnapshotProperties {
+  @doc("Data used when creating a snapshot")
+  creationData?: CreationData;
+
+  @doc("The time when the snapshot was created.")
+  @visibility(Lifecycle.Read)
+  timeCreated?: utcDateTime;
+
```
mentat9: This should go in the systemData.createdAt top-level property

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
```diff
@@ -1949,6 +1961,77 @@ model StorageContainersUpdateRequest {
...
+  @doc("The size of the disk in bytes.")
+  @visibility(Lifecycle.Read)
+  diskSizeBytes?: int64;
+
+  @doc("Unique identifier for the snapshot.")
+  @visibility(Lifecycle.Read)
+  uniqueId?: string;
+
```
mentat9: Similar feedback as for the sourceUniqueId. Flesh this out with format and more doc details. For example, if it's created by the backend, indicate so in the doc.

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/Snapshots.tsp
```diff
@@ -0,0 +1,60 @@
...
+using Azure.ResourceManager;
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Microsoft.AzureStackHCI;
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "For backward compatibility"
+#suppress "@azure-tools/typespec-azure-core/no-openapi" "x-ms-azure-resource legacy"
+@added(Versions.v2026_02_01_preview)
```
mentat9: I don't recognize these suppressions. Are these needed? Can you elaborate?

jgu17: I am not familiar with those either, but they are used in existing specs, and I kept them for consistency at least if not other reasons, for example: https://github.com/Azure/azure-rest-api-specs/blob/836342bd552c08c49f73dcb09600351d9cc13c4b/specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/NetworkInterfaces.tsp#L16

package-lock.json
jgu17: it was auto edited by the npx command. let me ignore / backout the change by the command

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
jgu17: done

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
jgu17: this is to align with https://github.com/Azure/azure-rest-api-specs/blob/bd3940ffa9d02457409e77339e0f63aefd49d4c4/specification/compute/Disk.Management/models.tsp#L557

specification/azurestackhci/AzureStackHCI.StackHCIVM.Management/models.tsp
jgu17: done

jgu17:
there was a large scale refactoring on main and it will be easier to recreate this pr based on main

PR #39505, opened 2026-01-08
[.github] Remove ArcReview

mikeharder:
- Label is defunct and should be removed
- Left over from a previous time when private Arc clouds each had their own top-level VM resource
- Since then the design has changed and all private Arc clouds use HybridCompute.machine as the top level resource
- Removal requested by @edyoung and @jianyunt


edyoung:
Thanks!

edyoung (approved):

PR #39517, opened 2026-01-08
Use CustomPatch for EdgeActions Service

kazrael2119:
This is the change only impacting the client generation no REST API change.

Update the ArmResourcePatchSync to ArmCustomPatchSync and this is supposed to only impact the client generation and have no impact for REST API.

Guidance doc: https://azure.github.io/typespec-azure/docs/howtos/arm/resource-operations/#resource-update-operations-patch

2025-09-01-preview and 2025-12-01-preview have not released SDKs before.


specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/main.tsp
```diff
@@ -85,7 +85,7 @@ model EdgeActionExecutionFilter
 interface EdgeActions {
   get is ArmResourceRead<EdgeAction>;
   create is ArmResourceCreateOrUpdateAsync<EdgeAction>;
-  update is ArmResourcePatchAsync<EdgeAction, EdgeActionProperties>;
+  update is ArmCustomPatchAsync<EdgeAction, EdgeActionUpdate>;
```
tundwed: Looks like we are updating the input (request object) from EdgeActionProperties to EdgeActionUpdate

@kazrael2119  I think that would impact not just the client but also our api, we would have to update our model classes and the api.

Do we need to change the request object?

same for the other object

tundwed, line 88: Looks like we are updating the input (request object) from EdgeActionProperties to EdgeActionUpdate

@kazrael2119 I think that would impact not just the client but also our api, we would have to update our model classes and the api.

Do we need to change the request object?

same for the other object

MaryGao: @tundwed This change to ArmCustomPatchAsync will not change the input model and you could compare the [main update](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-09-01-preview/openapi.json#L257-L288) with this [commit change](https://github.com/Azure/azure-rest-api-specs/blob/77333de3d08ee82bcfe1585f5c2d2269093e0f08/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-09-01-preview/openapi.json#L257-L288).

Actually this change would not update any REST-level stuff.  We just switched the template from ArmResourcePatchAsync to ArmCustomPatchAsync which is [our recommandation in guide](https://azure.github.io/typespec-azure/docs/howtos/arm/resource-operations/#resource-update-operations-patch) so that client side could better handle this operation.

The same for the other object.

tundwed: Okay, got it.

weidongxu-microsoft (approved):
Mostly a question for service:

Do we need the `SkuTypeUpdate`?

For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?

Ask because `SkuTypeUpdate` feels unusual...

MaryGao:
Add some labels consdiering there is no real REST change and these labels are false alarmed.

MaryGao (approved):

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
```diff
@@ -133,6 +142,7 @@ model EdgeActionAttachmentResponse {
   edgeActionId: string;
 }
 
+#suppress "@azure-tools/typespec-azure-resource-manager/empty-updateable-properties" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
mentat9, line 145: >"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details" [](http://example.com/codeflow?start=85&length=75)

Please add the missing justification.

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
```diff
@@ -144,6 +154,23 @@ model EdgeActionProperties {
   attachments: EdgeActionAttachment[];
 }
 
+/** The type used for update operations of the EdgeAction. */
+model EdgeActionUpdate {
+  @doc("The resource-specific properties for this resource.")
+  properties?: EdgeActionPropertiesUpdate;
```
mentat9, line 160: >properties?: EdgeActionPropertiesUpdate; [](http://example.com/codeflow?start=2&length=40)

I don't think the empty property bag is required. If you can remove this, you may also be able to remove two suppressions.

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
```diff
@@ -144,6 +154,23 @@ model EdgeActionProperties {
...
+
+  @doc("The sku type of the edge action")
+  sku?: SkuTypeUpdate;
+
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
mentat9, line 165: >"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details" [](http://example.com/codeflow?start=73&length=75)

Add the missing justification.

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
```diff
@@ -144,6 +154,23 @@ model EdgeActionProperties {
...
+  @doc("Resource tags.")
+  tags?: Record<string>;
+}
+
+/** Represents an edge action properties */
+#suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
mentat9, line 171: >"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details" [](http://example.com/codeflow?start=72&length=75)

Add the missing justification.

tundwed:
> Mostly a question for service:
> 
> Do we need the `SkuTypeUpdate`?
> 
> For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?
> 
> Ask because `SkuTypeUpdate` feels unusual...

@weidongxu-microsoft We do not need this type, it was not present before. @kazrael2119 why was this type added?

weidongxu-microsoft:
> > Mostly a question for service:
> > Do we need the `SkuTypeUpdate`?
> > For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?
> > Ask because `SkuTypeUpdate` feels unusual...
> 
> @weidongxu-microsoft We do not need this type, it was not present before. @kazrael2119 why was this type added?

The type was in openapi.json before this PR
https://github.com/Azure/azure-rest-api-specs/blob/08f8087732facee1895f6fcbcd07a362c3acac94/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-12-01-preview/openapi.json#L1890

I think it was generated due to the `ArmResourcePatchSync` template. And ZiWei kept it in this PR to avoid breaking change on Swagger.

I think we can clean up it a bit in this PR, as long as service is good with it. As change to existing Swagger schema (e.g. remove the `SkuTypeUpdate` would trigger a breaking change review, though I think we can add a "benign" or "bugfix" label to suppress, as there is no service change)

tundwed:
> > > Mostly a question for service:
> > > Do we need the `SkuTypeUpdate`?
> > > For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?
> > > Ask because `SkuTypeUpdate` feels unusual...
> > 
> > 
> > @weidongxu-microsoft We do not need this type, it was not present before. @kazrael2119 why was this type added?
> 
> The type was in openapi.json before this PR
> 
> https://github.com/Azure/azure-rest-api-specs/blob/08f8087732facee1895f6fcbcd07a362c3acac94/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-12-01-preview/openapi.json#L1890
> 
> I think it was generated due to the `ArmResourcePatchSync` template. And ZiWei kept it in this PR to avoid breaking change on Swagger.
> 
> I think we can clean up it a bit in this PR, as long as service is good with it. As change to existing Swagger schema (e.g. remove the `SkuTypeUpdate` would trigger a breaking change review, though I think we can add a "benign" or "bugfix" label to suppress, as there is no service change)

We currently have no dependency on `SkuTypeUpdate ` on the service side, so we can use name and tier and that should be fine. And so, if we can suppress it could we do that @kazrael2119 ?

kazrael2119:
> > > > Mostly a question for service:
> > > > Do we need the `SkuTypeUpdate`?
> > > > For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?
> > > > Ask because `SkuTypeUpdate` feels unusual...
> > > 
> > > 
> > > @weidongxu-microsoft We do not need this type, it was not present before. @kazrael2119 why was this type added?
> > 
> > 
> > The type was in openapi.json before this PR
> > https://github.com/Azure/azure-rest-api-specs/blob/08f8087732facee1895f6fcbcd07a362c3acac94/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-12-01-preview/openapi.json#L1890
> > 
> > I think it was generated due to the `ArmResourcePatchSync` template. And ZiWei kept it in this PR to avoid breaking change on Swagger.
> > I think we can clean up it a bit in this PR, as long as service is good with it. As change to existing Swagger schema (e.g. remove the `SkuTypeUpdate` would trigger a breaking change review, though I think we can add a "benign" or "bugfix" label to suppress, as there is no service change)
> 
> We currently have no dependency on `SkuTypeUpdate ` on the service side, so we can use name and tier and that should be fine. And so, if we can suppress it could we do that @kazrael2119 ?

I can't use skuType directly for an Update model as it will fail the lint check with this error message
image
so I have to use skuTypeUpdate instead

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
kazrael2119: this is to avoid breaking change in generated swagger
origin swagger： https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-12-01-preview/openapi.json#L1605-L1608

tundwed:
> > > > > Mostly a question for service:
> > > > > Do we need the `SkuTypeUpdate`?
> > > > > For service to update the sku, can user e.g. only write `name` and leave `tier` undefined?
> > > > > Ask because `SkuTypeUpdate` feels unusual...
> > > > 
> > > > 
> > > > @weidongxu-microsoft We do not need this type, it was not present before. @kazrael2119 why was this type added?
> > > 
> > > 
> > > The type was in openapi.json before this PR
> > > https://github.com/Azure/azure-rest-api-specs/blob/08f8087732facee1895f6fcbcd07a362c3acac94/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-12-01-preview/openapi.json#L1890
> > > 
> > > I think it was generated due to the `ArmResourcePatchSync` template. And ZiWei kept it in this PR to avoid breaking change on Swagger.
> > > I think we can clean up it a bit in this PR, as long as service is good with it. As change to existing Swagger schema (e.g. remove the `SkuTypeUpdate` would trigger a breaking change review, though I think we can add a "benign" or "bugfix" label to suppress, as there is no service change)
> > 
> > 
> > We currently have no dependency on `SkuTypeUpdate ` on the service side, so we can use name and tier and that should be fine. And so, if we can suppress it could we do that @kazrael2119 ?
> 
> I can't use skuType directly for an Update model as it will fail the lint check with this error message image so I have to use skuTypeUpdate instead

I see, looks like ARM enforces this. Since it maintains the current state. I am okay with leaving it as is.

image


@weidongxu-microsoft let me know if you have a different thought

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/edgeactionsbasemodels.tsp
kazrael2119: This is a placeholder for the suppression, The content inside has not been written by @pshao25 yet

tundwed (approved):
The changes do not impact api logic.

weidongxu-microsoft:
> I see, looks like ARM enforces this. Since it maintains the current state. I am okay with leaving it as is.
> 
> image
> @weidongxu-microsoft let me know if you have a different thought

Good to me, if we keep current state.

PR #39525, opened 2026-01-08
[edgeactions] Update client.tsp

ChenxiJiang333:
update python tsp config to release https://github.com/Azure/azure-sdk-for-python/pull/44582

### DIFF
--- a/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/client.tsp
+++ b/specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/client.tsp
@@ -28,3 +28,5 @@ using Microsoft.Cdn;
   "content",
   "csharp"
 );
+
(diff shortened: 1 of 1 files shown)


msyyc (approved):

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/client.tsp
```diff
@@ -28,3 +28,5 @@ using Microsoft.Cdn;
   "content",
   "csharp"
 );
+
+@@clientName(EdgeActions, "EdgeActionsMgmtClient", "python");
```
msyyc, line 32: @ChenxiJiang333 The namespace is `Microsoft.Cdn` instead of `EdgeActions` so the rename works for an operation group which is not expected.

PR #39529, opened 2026-01-08
Enabled the SDK PR creation when failure happens in the generation process

### LINKED ISSUE #38221: [FEATURE REQ] Generation pipelines should create PR even when there are build errors
### API Spec link

https://github.com/Azure/azure-rest-api-specs/pull/37683

### API Spec version

NA

### Please describe the feature.

In the generation pipeline for .NET, if there is a build failure or even a generation error, we do not create the draft PR. We should always create a draft PR if there is something that can be committed. If there are failures, then the PR should have "Failed" in the title.

raych1:
Resolved https://github.com/Azure/azure-rest-api-specs/issues/38221

**Key changes**:
- added a new parameter to pipeline to force create the SDK PR
- always set pipeline variables so that the SDK PR can be created using them

[Test Run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5733188&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=53495029-6fd7-5fbf-3ed4-5fd95aecc3eb)

### DIFF
--- a/eng/common/pipelines/templates/steps/git-push-changes.yml
+++ b/eng/common/pipelines/templates/steps/git-push-changes.yml
@@ -8,12 +8,11 @@ parameters:
   WorkingDirectory: $(System.DefaultWorkingDirectory)'
   ScriptDirectory: eng/common/scripts
   SkipCheckingForChanges: false
-  CustomCondition: succeeded()
 
 steps:
 - task: PowerShell@2
   displayName: Check for changes
-  condition: and(${{ parameters.CustomCondition }}, eq(${{ parameters.SkipCheckingForChanges }}, false))
+  condition: and(succeeded(), eq(${{ parameters.SkipCheckingForChanges }}, false))
   inputs:
     pwsh: true
     workingDirectory: ${{ parameters.WorkingDirectory }}
@@ -26,7 +25,7 @@ steps:
     echo "##vso[task.setvariable variable=RepoNameWithoutOwner]$repoName"
     echo "RepoName = $repoName"
   displayName: Remove Repo Owner from Repo Name
-  condition: ${{ parameters.CustomCondition }}
+  condition: succeeded()
   workingDirectory: ${{ parameters.WorkingDirectory }}
 
 - template: /eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
@@ -36,7 +35,7 @@ steps:
 
 - task: PowerShell@2
   displayName: Push changes
-  condition: and(${{ parameters.CustomCondition }}, eq(variables['HasChanges'], 'true'))
+  condition: and(succeeded(), eq(variables['HasChanges'], 'true'))
   inputs:
     pwsh: true
     workingDirectory: ${{ parameters.WorkingDirectory }}
--- a/eng/common/pipelines/templates/steps/git-push-changes.yml
+++ b/eng/common/pipelines/templates/steps/git-push-changes.yml
@@ -8,11 +8,12 @@ parameters:
   WorkingDirectory: $(System.DefaultWorkingDirectory)'
   ScriptDirectory: eng/common/scripts
   SkipCheckingForChanges: false
+  CustomCondition: succeeded()
 
 steps:
 - task: PowerShell@2
   displayName: Check for changes
-  condition: and(succeeded(), eq(${{ parameters.SkipCheckingForChanges }}, false))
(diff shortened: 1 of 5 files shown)


eng/common/pipelines/templates/steps/git-push-changes.yml
```diff
@@ -8,11 +8,12 @@ parameters:
   WorkingDirectory: $(System.DefaultWorkingDirectory)'
   ScriptDirectory: eng/common/scripts
   SkipCheckingForChanges: false
+  CustomCondition: succeeded()
```
raych1: This file change will be moved to the tools repo.

eng/common/pipelines/templates/steps/git-push-changes.yml
```diff
@@ -25,7 +26,7 @@ steps:
     echo "##vso[task.setvariable variable=RepoNameWithoutOwner]$repoName"
     echo "RepoName = $repoName"
   displayName: Remove Repo Owner from Repo Name
-  condition: succeeded()
+  condition: ${{ parameters.CustomCondition }}
   workingDirectory: ${{ parameters.WorkingDirectory }}
 
 - template: /eng/common/pipelines/templates/steps/emit-rate-limit-metrics.yml
```
raych1, line 31: I didn't add `customCondition` to this step as I feel it is fine to be skipped when failure happens.

eng/common/pipelines/templates/steps/git-push-changes.yml
raych1: @weshaggard do you have any feedback on these changes? Otherwise, I will create a separate PR in the tools repo.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
weshaggard, line 262: Instead of forcing the PR on failure should we instead set `continueOnError: true` for the generate step?

raych1: I think we want to scope this feature to the generation scenario only not the spec PR or batch run scenarios.  It cannot be achieved by using `continueOnError` .

weshaggard: I was expecting a template condition like:

```
${{ if eq(parameters.ForceCreateEvenWithFailures, true) }}:
  continueOnError: true
```

raych1: Cool! This approach is simpler to check the parameter.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -306,10 +312,19 @@ stages:
             TargetRepoName: $(SdkRepoName)
             WorkingDirectory: $(SdkRepoDirectory)
             ScriptDirectory: $(SdkRepoDirectory)/eng/common/scripts
+            CustomCondition: succeededOrFailed()
+
+        - pwsh: |
```
weshaggard, line 315: Is it really interesting to update the title?

raych1: That's the request from Josh. It's fine to differentiate the failures.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -306,10 +312,19 @@ stages:
             TargetRepoName: $(SdkRepoName)
             WorkingDirectory: $(SdkRepoDirectory)
             ScriptDirectory: $(SdkRepoDirectory)/eng/common/scripts
+            CustomCondition: succeededOrFailed()
```
weshaggard: This shouldn't be needed now with the continueOnError approach.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -282,6 +287,7 @@ stages:
             Write-Host "##vso[task.setvariable variable=SdkPullRequestSourceBranch]$sdkPrBranchName"
           workingDirectory: $(SdkRepoDirectory)
           displayName: "Create SDK PR branch variable"
+          condition: succeededOrFailed()
```
weshaggard: Is this needed anymore?

raych1: Right, it could be deleted.

weshaggard (approved):

PR #39554, opened 2026-01-09
Fix the type of PatternAnalyzer.flags

xiangyan99:


kuanlu95 (approved):

raych1:
Talked to @xiangyan99 offline, the reported breaking change is exactly what this PR want to fix. The fix is to align to the [previous version](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/search/data-plane/Search/preview/2025-08-01-preview/searchservice.json#L4326).

Issue #38586, opened 2025-11-04
[TypeSpec] Should not add ARMReview label for any SDK configuration PR


ArthurMa1978:
If a TypeSpec PR only updates `client.tsp` and `tspconfig.yaml` for SDK generation configuration, the `ARMReview` label should not be applied.

Here is the comments from @mikeharder of this problem:

`ARMReview` is applied to any PR that changes any file under a resource-manager folder.  To my knowledge, it has always worked this way.
 
https://github.com/Azure/azure-rest-api-specs/blob/70bd6219f7e8edbc62528fc39b757570082011c9/eng/tools/summarize-impact/src/impact.ts#L159
 
https://github.com/Azure/azure-rest-api-specs/blob/7eb817d513e6b2364a823d1f56db45d25fdca7b7/.github/workflows/src/summarize-checks/labelling.js#L537
 
What might be a recent change, is in folder structure v2, the TSP sources are moved under the resource-manager folder.  While in folder structure v1, the TSP sources were outside resource-manager.
 
So in v1, changes to only `client.tsp` and `tspconfig.yaml` would not trigger `ARMReview`, since they are outside resource-manager, and they don't cause any changes to generated swagger under resource-manager.  But in v2, changes to `client.tsp` and `tspconfig.yaml` will trigger `ARMReview`.
 
So now we need to fix this and make sure any SDK generation configuration change should not trigger `ARMReview`.

mikeharder:
## Design

Currently, label `ARMReview` is added if a PR changes any files matching this simple glob:

```
resource-manager/**
```

In "Folder Structure v1", this works pretty well, because the TSP sources and config live **outside** the `resource-manager` folder.  An exception is the `readme.<lang>.md` files, which only impact SDK generation (not ARM), but still trigger `ARMReview`.  Example:

* https://github.com/Azure/azure-rest-api-specs/pull/33988

In "Folder Structure v2", the TSP sources are moved under the `resource-manager` folder, which causes more false-positives.  I think we should add the following exclusions:

```
resource-manager/**
!resource-manager/**/client.tsp
!resource-manager/**/readme.?*.md
!resource-manager/**/tspconfig.yaml
```

`client.tsp` and `readme.<lang>.md` only impact the SDK, so these should obviously be safe.
 
`tspconfig.yaml` is safe, because TypeSpecValidation should already verify most of the autorest and linter settings, and most of these settings also impact generated swagger (which will still trigger `ARMReview`).

[...]

PR #39578, opened 2026-01-12
Use @@override for a patch operation to avoid JS sdk breakings for resourceconnector

kazrael2119:
https://github.com/Azure/azure-sdk-for-js/pull/37005/files#r2681026239
JS doesn't support flatten in request parameter level.
customize `appliances_update` operation for JS side only to avoid unexpected breakings.

This PR should only impact JS client.


MaryGao:
No need for ARM review for the client.tsp considering the [latest decision in issue](https://github.com/Azure/azure-rest-api-specs/issues/38586).

MaryGao (approved):

PR #38105, opened 2025-10-13
Update client settings for KeyVault

ArthurMa1978:
# SDK configuration pull request

## Purpose of this PR

- [ ] Make changes to the SDK configuration only when there are no modifications to the API specification, eliminating the need for an ARM or Stewardship Board API review.

## Due diligence checklist

To merge this PR, you **must** go through the following checklist and confirm you understood
and followed the instructions by checking all the boxes:

- [ ] I confirm this PR is modifying only SDK configurations, and not API related specifications.
- [ ] I have reviewed and used the respective `tspconfig.yaml` templates:
  - [ARM tspconfig template](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.Management/tspconfig.yaml)
  - [Data plane tspconfig template](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/Contoso.WidgetManager/tspconfig.yaml)

## Getting help

- First, carefully read through this PR description, from top to bottom. Fill out the `Purpose of this PR` and `Due diligence checklist`.
- If you don't have permissions to remove or add labels to the PR, request `write access` per [aka.ms/azsdk/access#request-access-to-rest-api-or-sdk-repositories](https://aka.ms/azsdk/access#request-access-to-rest-api-or-sdk-repositories)
- To understand what you must do next to merge this PR, see the `Next Steps to Merge` comment. It will appear within few minutes of submitting this PR and will continue to be up-to-date with current PR state.
- For guidance on fixing this PR CI check failures, see the hyperlinks provided in given failure and https://aka.ms/ci-fix.
- If the PR CI checks appear to be stuck in `queued` state, please add a comment with contents `/azp run`.
  This should result in a new comment denoting a `PR validation pipeline` has started and the checks should be updated after few minutes.
- If the help provided by the previous points is not enough, post to https://aka.ms/azsdk/support/specreview-channel and link to this PR.

PR #39607, opened 2026-01-13
[TSP Migration][FrontDoor] TypeSpec migrated from swagger

mcgallan:


weidongxu-microsoft:
Is `BasicResourceWithSettableIDName` introduced in the conversion?

mcgallan:
> Is `BasicResourceWithSettableIDName` introduced in the conversion?

Yes, because the TSP files generated by the convert tool all use the common type Resource model, but their ID and Name are readonly, which does not match the original swagger. At the same time, the original swagger does not contain a ready-made Resource model with writable ID and Name, so I introduced BasicResourceWithSettableIDName.
original swgger: https://github.com/Azure/azure-rest-api-specs/blob/715e95d5dadfd6c1ac1eed1fb99d773facc1d1d9/specification/frontdoor/resource-manager/Microsoft.Network/stable/2025-10-01/frontdoor.json#L1729-L1752

weidongxu-microsoft:
> > Is `BasicResourceWithSettableIDName` introduced in the conversion?
> 
> Yes, because the TSP files generated by the convert tool all use the common type Resource model, but their ID and Name are readonly, which does not match the original swagger. At the same time, the original swagger does not contain a ready-made Resource model with writable ID and Name, so I introduced BasicResourceWithSettableIDName. original swgger:
> 
> https://github.com/Azure/azure-rest-api-specs/blob/715e95d5dadfd6c1ac1eed1fb99d773facc1d1d9/specification/frontdoor/resource-manager/Microsoft.Network/stable/2025-10-01/frontdoor.json#L1729-L1752

If possible, please involve .NET dev for approval.

They've recently removed a similar model, which breaks others. (I am aware that they are not exactly the same situation)
https://github.com/Azure/azure-rest-api-specs/pull/38105/changes#diff-ec0b9bc69f3f8fda52974aa79b644fd684cd282ecce6dac3f4950b786c7af281

The point is, if we introduce it, then remove it, we'd better not to introduce it in the first place -- if this is possible.

pshao25 (approved):

PR #39495, opened 2026-01-07
Add GitHub Workflow to validate ARM leases

tejaswiMinnu:
## ARM Leases Validation Workflow

Added automated validation for ARM lease files that establish design discussion periods between PMs and RP owners. PMs add lease.yaml files after office hours discussions with RP owners, if new Resource type api path is found and lease is expired RP owners go through the process again 

**Key Features:**
- Validates lease file structure: `.github/arm-leases/<orgName>/<rpNamespace>/<servicename>(optional)/lease.yaml`
- Enforces resource provider naming (capital letters: `Microsoft.Test`)
- Validates dates (ISO 8601, no past dates grace period 10 days accepted) and duration (max 180 days) - after 180 days if any new RT is detected WF should fail.
- Requires reviewer field who are PM's no validation needed juts adding github alias

PR #39625, opened 2026-01-13
Detect new Resource Provider Namespaces, check and validate arm-leases for the RP's

tejaswiMinnu:
**Overview**
The ARM lease validation system makes it easy to check Azure Resource Manager resource provider leases automatically when you open or update a pull request. It helps teams confirm they've got valid leases before making changes to resource providers in the azure-rest-api-specs repo.

**Core Components**
There are two main modules: detect-arm-leases.js, which handles the lease validation with the checkLease(serviceName, resourceProvider) function, and detect-new-resource-provider.js, which spots new resource provider namespaces and connects with the lease check.

**Resource Provider Detection Workflow**
Whenever a pull request comes in, the workflow checks for changes in specification/**/resource-manager directories to find new RP namespaces. The extractResourceProviders function gives you a Map of namespace-to-serviceName pairs, which are checked against existing paths with resourceProviderExists. For each new namespace found, checkLease runs and the results are collected and output is designed for consumption by GitHub Actions through the GitHub Script action, enabling automated PR comments that direct contributors to attend office hours for missing or expired leases.

**Lease Validation Logic**
The checkLease function reads YAML lease files from arm-leases/<serviceName>/<resourceProvider>/lease.yaml and checks if the lease is still valid using today <= (startDate + duration). It returns true if the lease is good, or false if it's missing or expired. The system works with different duration formats and handles issues like missing files, invalid YAML, and future start dates without crashing

Just a heads up, this workflow is currently behind a feature flag, so the PMs will decide when and how it gets released.


.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,73 @@
+import { existsSync, readFileSync } from 'fs';
+import { join } from 'path';
```
raosuhas: Is this file just being checked in and not used currently ?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,170 @@
...
+import { checkLease } from './detect-arm-leases.js';
+
+// ============================================
+// Configuration
+// ============================================
+
+const SPECIFICATION_PATH = 'specification';
+// Match pattern: specification/<service>/resource-manager/<ResourceProvider.Namespace>/...
+const RESOURCE_MANAGER_PATTERN = /^specification\/[^\/]+\/resource-manager\/([^\/]+)/;
+
```
raosuhas: We probably need to consider the folder structure for new service groups as well eventually right ? for example Microsoft.Compute : 

[https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager/Microsoft.Compute](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager/Microsoft.Compute)

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Check if ARM lease exists and is valid
+ * @param {string} serviceName - Service name
+ * @param {string} resourceProvider - Resource provider namespace
+ * @returns {boolean} True if lease exists and is valid, false otherwise
+ */
+export function checkLease(serviceName, resourceProvider) {
+  try {
+    const repoRoot = process.env.TEST_REPO_ROOT || execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
+    const leasePath = join(repoRoot, '.github', 'arm-leases', serviceName, resourceProvider, 'lease.yaml');
```
raosuhas: >resourceProvider [](http://example.com/codeflow?start=75&length=16)

here again we need to take in servicergoup as optional param. Look at Microsoft.Compute as an example :

[https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager/Microsoft.Compute](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager/Microsoft.Compute)

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: No, I am calling this here to check lease - https://github.com/Azure/azure-rest-api-specs/pull/39625/files#:~:text=const%20leaseValid%20%3D%20checkLease(rp.serviceName%2C%20rp.namespace%2C%20rp.serviceGroup%20%7C%7C%20%27%27)%3B

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Updated code considering service group as optional

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Updated

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          node-version: '24'
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: .github
+
+      - name: Detect new resource providers
+        id: detect
+        run: node workflows/src/detect-new-resource-provider.js ${{ github.event.pull_request.base.ref }}
```
AkhilaIlla: You should use GH script for these kind of actions.

EG: [https://github.com/Azure/azure-rest-api-specs/blob/652ad4cb131256f10a90ea2df207b38cf35d6671/.github/workflows/arm-auto-signoff-code.yaml#L42](https://github.com/Azure/azure-rest-api-specs/blob/652ad4cb131256f10a90ea2df207b38cf35d6671/.github/workflows/arm-auto-signoff-code.yaml#L42)

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Fetch base branch
+        run: git fetch origin ${{ github.event.pull_request.base.ref }}
```
AkhilaIlla: You can do parse checkout. depth 0 will checkout the entire repo and we will not need the whole thing. Also, you will not need to explicitly checkout base branch

See- [https://github.com/Azure/azure-rest-api-specs/blob/652ad4cb131256f10a90ea2df207b38cf35d6671/.github/workflows/arm-auto-signoff-code.yaml#L29-L37](https://github.com/Azure/azure-rest-api-specs/blob/652ad4cb131256f10a90ea2df207b38cf35d6671/.github/workflows/arm-auto-signoff-code.yaml#L29-L37)

and you can refer them using- [https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff/trivial-changes-check.js#L223-L236](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff/trivial-changes-check.js#L223-L236)

AkhilaIlla:
> 

Move this into a script and have it called here.

Also, Do we need all those emoji's in the content of the file?

---
Refers to: .github/workflows/detect-new-resource-provider.yaml:86 in fba25c9. [](commit_id = fba25c9bd0bb04ee62eb459508b3e83a1d804f2d, deletion_comment = False)

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: Used GH scripts

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: Thank you, added similar to this

tejaswiMinnu:
> > 
> 
> Move this into a script and have it called here.
> 
> Also, Do we need all those emoji's in the content of the file?
> 
> Refers to: .github/workflows/detect-new-resource-provider.yaml:86 in [fba25c9](https://github.com/Azure/azure-rest-api-specs/commit/fba25c9bd0bb04ee62eb459508b3e83a1d804f2d). [](commit_id = [fba25c9](https://github.com/Azure/azure-rest-api-specs/commit/fba25c9bd0bb04ee62eb459508b3e83a1d804f2d), deletion_comment = False)

Done, removed emojis and moved to .js

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,49 @@
...
+          result-encoding: string
+          script: |
+            const { default: detectNewResourceProvider } =
+              await import('${{ github.workspace }}/.github/workflows/src/detect-new-resource-provider.js');
+            return await detectNewResourceProvider({ 
+              github, 
+              context, 
+              core,
+              baseBranch: '${{ github.event.pull_request.base.ref }}'
+            });
```
AkhilaIlla: do we need indentation here?
To have the conditional statement visible/valid?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+const SPECIFICATION_PATH = 'specification';
+// Match pattern: specification/<service>/resource-manager/<ResourceProvider.Namespace>/...
+const RESOURCE_MANAGER_PATTERN = /^specification\/[^\/]+\/resource-manager\/([^\/]+)/;
+// Match pattern with optional service group: specification/<service>/resource-manager/<ResourceProvider.Namespace>/<ServiceGroup>/...
+// ServiceGroup folder name should not start with "stable" or "preview"
+const RESOURCE_MANAGER_WITH_GROUP_PATTERN = /^specification\/[^\/]+\/resource-manager\/([^\/]+)\/(?!stable|preview)([^\/]+)\//;
+
+// Labels to add when new resource providers are detected
+const NEW_RP_LABELS = ['ARMModelingReviewRequired', 'NotReadyForARMReview'];
```
AkhilaIlla: should we convert these to enum?

AkhilaIlla: EG; [https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff/arm-auto-signoff-labels.js](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/arm-auto-signoff/arm-auto-signoff-labels.js)

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  // Get all changed files in specification/*/resource-manager/
+  const allFiles = await getChangedFiles({
+    baseCommitish: mergeBase,
+    headCommitish: 'HEAD'
+  });
+
+  const rmFiles = allFiles.filter(file => 
+    file.includes('/resource-manager/') && file.startsWith('specification/')
```
AkhilaIlla: You can encorporate   
allFiles.filter(resourceManager);
changedFiles.filter(swagger);

EG: 
[https://github.com/Azure/azure-rest-api-specs/blob/aa02e6bed6a4fd9a3b1c6a0b24c949e712491c6c/.github/workflows/src/arm-auto-signoff/trivial-changes-check.js#L58C25-L58C54](https://github.com/Azure/azure-rest-api-specs/blob/aa02e6bed6a4fd9a3b1c6a0b24c949e712491c6c/.github/workflows/src/arm-auto-signoff/trivial-changes-check.js#L58C25-L58C54)

https://github.com/Azure/azure-rest-api-specs/blob/aa02e6bed6a4fd9a3b1c6a0b24c949e712491c6c/.github/workflows/src/arm-auto-signoff/trivial-changes-check.js#L163C21-L163C49

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+        serviceGroup: info.serviceGroup 
+      });
+    } 
+  }
+  
+  // Check ARM leases for new resource providers
+  const leaseCheckResults = [];
+  
+  if (newResourceProviders.length > 0) {
+    console.log(`\n🆕 Detected ${newResourceProviders.length} new resource provider(s)\n`);
```
AkhilaIlla: >🆕 [](http://example.com/codeflow?start=19&length=2)

Remove these ?
I think its better to keep the logs clean?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+      
+      leaseCheckResults.push({
+        namespace: rp.namespace,
+        serviceName: rp.serviceName,
+        serviceGroup: rp.serviceGroup,
+        leaseValid: leaseValid,
+        leaseMessage: leaseValid ? 'Lease is valid' : 'No lease file found or lease has expired'
+      });
+      
+      console.log(`  - ${rp.namespace}: ${leaseValid ? '[OK]' : '[FAILED]'} Lease ${leaseValid ? 'valid' : 'invalid'}`);
```
AkhilaIlla: > ${leaseValid ? '[OK]' : '[FAILED]'} Lease ${leaseValid ? 'valid' : 'invalid'} [](http://example.com/codeflow?start=39&length=78)

can we not merge this?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+      console.log(`[FAILED] Failed to create PR comment: ${error.message}`);
+    }
+    
+    // Set the action as failed to indicate new RPs were detected
+    core.setFailed('New resource providers detected');
+    return 'new-rp-detected';
+  } else {
+    console.log('No new resource providers detected.\n');
+    return 'no-new-rp';
+  }
```
AkhilaIlla: We have a
"Next steps to merge" comment that gets evaluated and added to every PR. Im guessing you can use that also guessing Mike will ask you to utilize that existing feature.

Check with Mike on this one.

.github/workflows/src/detect-new-resource-provider.js
AkhilaIlla: Same in the later steps as well

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,271 @@
...
+ */
+async function addNewResourceProviderLabels(github, context) {
+  try {
+    await github.rest.issues.addLabels({
+      owner: context.repo.owner,
+      repo: context.repo.repo,
+      issue_number: context.issue.number,
+      labels: NEW_RP_LABELS
+    });
+
```
AkhilaIlla: This repo has a separate workflow to add the labels.
You should have the label output to the workflow and levarage that to have the label added.

EG [https://github.com/Azure/azure-rest-api-specs/blob/aa02e6bed6a4fd9a3b1c6a0b24c949e712491c6c/.github/workflows/arm-auto-signoff-status.yaml#L98](https://github.com/Azure/azure-rest-api-specs/blob/aa02e6bed6a4fd9a3b1c6a0b24c949e712491c6c/.github/workflows/arm-auto-signoff-status.yaml#L98)

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  const result = checkLease(serviceName, resourceProvider, serviceGroup);
+
+  if (!result) {
+    const repoRoot = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
+    const leasePath = buildLeasePath(repoRoot, serviceName, resourceProvider, serviceGroup);
+    
+    if (!existsSync(leasePath)) {
+      console.log('No lease file found');
```
AkhilaIlla: is this not already being check in CheckLease function?

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: I have checked indentation locally, seems like code is indented.

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: I did convert to enum as suggested

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Changed as suggested at two places, can you please recheck if that is correct

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Ahh, somehow missed this. Removed in latest commit

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Sure, modified most of the logs in this file

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Yes, I'll add this in another PR to keep each PR simple. I haven't included the next steps to merge in this PR; that's a separate task for another PR.

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Added similar to the EG

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Removed the duplicate check
Thanks for reviewing

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+  pull-requests: write
+
+env:
+  ENABLE_NEW_RP_DETECTION: false
+
+jobs:
+  detect-new-rp:
+    name: Detect New Resource Provider
+    runs-on: ubuntu-24.04
+    if: env.ENABLE_NEW_RP_DETECTION == 'true'
```
mikeharder: @copilot: Can you fix this error?

```
context "env" is not allowed here. available contexts are "github", "inputs", "needs", "vars". see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for more details
```

mikeharder: @tejaswiMinnu: I don't think it's possible to feature-flag an action with an `env` var like this, since `if` is evaluated before `env`.  An error from ActionLint is shown below.

You could use `if: false` to "comment out" the job.

Or replace the triggers with something like just `workflow_dispatch`, which would let you run the workflow manually (against any branch), but never trigger the workflow automatically.

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+    types:
+      - opened
+      - synchronize
+      - reopened
+    paths:
+      - "specification/**/resource-manager/**"
+
+permissions:
+  contents: read
+  pull-requests: write
```
mikeharder:
```suggestion
```

hopefully you don't need write permissions, it makes things much easier with respect to security.  if you do need them, we'll need to discuss further.

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
+name: Detect New Resource Provider
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
```
mikeharder:
```suggestion
      # default
      - opened
      - synchronize
      - reopened
      # re-run if base branch is changed, since previous merge commit may generate incorrect diff
      - edited
```

Copied from arm-auto-signoff

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const { default: detectNewResourceProvider } =
+              await import('${{ github.workspace }}/.github/workflows/src/detect-new-resource-provider.js');
+            return await detectNewResourceProvider({ 
+              github, 
+              context, 
+              core,
+              baseBranch: '${{ github.event.pull_request.base.ref }}'
```
mikeharder:
```suggestion
```

you should be able to access this value already from JS, don't need to pass it from YAML

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+            });
+
+      - if: fromJson(steps.detect.outputs.result).labelActions['ARMModelingReviewRequired'] != 'none'
+        name: Upload artifact for ARMModelingReviewRequired label
+        uses: ./.github/actions/add-label-artifact
+        with:
+          name: "ARMModelingReviewRequired"
+          value: "${{ fromJson(steps.detect.outputs.result).labelActions['ARMModelingReviewRequired'] == 'add' }}"
```
mikeharder: I believe this is a new label, so it's fine to add/remove it directly here.

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+      - if: fromJson(steps.detect.outputs.result).labelActions['NotReadyForARMReview'] != 'none'
+        name: Upload artifact for NotReadyForARMReview label
+        uses: ./.github/actions/add-label-artifact
+        with:
+          name: "NotReadyForARMReview"
+          value: "${{ fromJson(steps.detect.outputs.result).labelActions['NotReadyForARMReview'] == 'add' }}"
```
mikeharder: this label is already "owned" by the summarize-checks workflow, so any changes to its behavior will need to be made over there, otherwise the workflows might "fight" with each other to override what the other sets.

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20NotReadyForARMReview&type=code

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+          value: "${{ fromJson(steps.detect.outputs.result).headSha }}"
+
+      - if: |
+          always() &&
+          fromJson(steps.detect.outputs.result).issueNumber > 0
+        name: Upload artifact with issue number
+        uses: ./.github/actions/add-empty-artifact
+        with:
+          name: "issue-number"
+          value: "${{ fromJson(steps.detect.outputs.result).issueNumber }}"
```
mikeharder:
```suggestion
```

workflows only need these, if they trigger on actions like workflow_run.  if this workflow only triggers on pull_request, this isn't necessary.

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env node
+
```
mikeharder:
```suggestion
```

script is not executable, so this is unnecessary

.github/workflows/src/detect-arm-leases.js
mikeharder, line 1: if this is intended to be executed directly on a dev machine, move it to the `cmd` folder, add to the "bin" section of "package.json", and follow the pattern of this file:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/cmd/spec-model.js

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/package.json#L31

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
+import { existsSync, readFileSync } from 'fs';
```
mikeharder:
```suggestion
import { exists, readFile } from 'fs/promises';
```

always prefer async apis, convert everything to async as needed

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Check if ARM lease exists and is valid
+ * @param {string} serviceName - Service name
+ * @param {string} resourceProvider - Resource provider namespace
+ * @param {string} serviceGroup - Optional service group
+ * @returns {{ exists: boolean, valid: boolean, leasePath: string }} Lease status
+ */
+function getLeaseStatus(serviceName, resourceProvider, serviceGroup = '') {
+  const repoRoot = process.env.TEST_REPO_ROOT || execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
```
mikeharder:
```suggestion
  const repoRoot = process.env.TEST_REPO_ROOT || getRootFolder();
```

use shared helper.  it currently requires an argument (you could pass `process.cwd()`), i will create a separate PR, to make the argument optional.

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20getRootFolder&type=code

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {string} serviceGroup - Optional service group
+ * @returns {string} Full path to lease.yaml file
+ */
+function buildLeasePath(repoRoot, serviceName, resourceProvider, serviceGroup = '') {
+  const leasePathParts = [repoRoot, '.github', 'arm-leases', serviceName, resourceProvider];
+  if (serviceGroup) {
+    leasePathParts.push(serviceGroup);
+  }
+  leasePathParts.push('lease.yaml');
+  return join(...leasePathParts);
```
mikeharder:
```suggestion
  return resolve(...leasePathParts);
```

Best practice is to use `resolve()`, since it's best to use absolute paths internally as much as possible, to avoid ambiguity.

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
...
+  const repoRoot = process.env.TEST_REPO_ROOT || execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
+  const leasePath = buildLeasePath(repoRoot, serviceName, resourceProvider, serviceGroup);
+
+  try {
+    if (!existsSync(leasePath)) {
+      return { exists: false, valid: false, leasePath };
+    }
+
+    const content = readFileSync(leasePath, 'utf-8');
+    const parsed = YAML.load(content);
```
mikeharder: you will need to create a schema, so the YAML can be parsed with strong types.  example:

https://github.com/Azure/azure-rest-api-specs/blob/ed16e10caee8ef2ab09f321272877d0efbec0d1e/.github/shared/src/readme.js#L155-L164

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  console.log('Lease is valid');
+  process.exit(0);
+}
+
+// Only run main if this file is executed directly (not imported)
+if (process.argv[1] && process.argv[1].endsWith('detect-arm-leases.js')) {
+  main();
+}
```
mikeharder: refactor so this isn't needed.  the file under `cmd` should only have the command-line app code.  anything that might need to be imported, should live under `src`.

.github/workflows/src/detect-new-resource-provider.js
mikeharder:
```suggestion
```

This isn't executable, so shouldn't be needed.

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+import { execSync } from 'child_process';
+import { readdirSync, existsSync, statSync, writeFileSync } from 'fs';
```
mikeharder: use async

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+const RESOURCE_MANAGER_WITH_GROUP_PATTERN = /^specification\/[^\/]+\/resource-manager\/([^\/]+)\/(?!stable|preview)([^\/]+)\//;
+
+/**
+ * New resource provider label names.
+ * @readonly
+ * @enum {string}
+ */
+const NewRpLabel = Object.freeze({
+  ArmModelingReviewRequired: 'ARMModelingReviewRequired',
+  NotReadyForArmReview: 'NotReadyForARMReview'
```
mikeharder: this will need to move to summarize-checks

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+});
+
+// Label action values used by update-labels workflow
+const LABEL_ACTIONS = Object.freeze({
```
mikeharder: use shared code:

https://github.com/Azure/azure-rest-api-specs/blob/ed16e10caee8ef2ab09f321272877d0efbec0d1e/.github/workflows/src/label.js#L5

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+ * Main detection logic for GitHub script action
+ * @param {Object} params - Parameters from github-script
+ * @param {Object} params.github - GitHub API client
+ * @param {Object} params.context - GitHub context
+ * @param {Object} params.core - GitHub Actions core
+ * @param {string} params.baseBranch - Base branch name
+ * @returns {Promise<string>} Result status
+ */
+export default async function detectNewResourceProvider({ github, context, core, baseBranch }) {
+  const repoRoot = execSync('git rev-parse --show-toplevel', { encoding: 'utf-8' }).trim();
```
mikeharder: use shared code:

https://github.com/Azure/azure-rest-api-specs/blob/ed16e10caee8ef2ab09f321272877d0efbec0d1e/.github/shared/src/simple-git.js#L9

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+    mergeBase = execSync(`git merge-base origin/${baseBranch} HEAD`, { encoding: 'utf-8' }).trim();
+  } catch (error) {
+    mergeBase = `origin/${baseBranch}`;
+  }
+
+  // Get all changed files in specification/*/resource-manager/
+  const allFiles = await getChangedFiles({
+    baseCommitish: mergeBase,
+    headCommitish: 'HEAD'
+  });
```
mikeharder: you shouldn't set these options, but you might want to set others.  look at existing usage, like this:

https://github.com/Azure/azure-rest-api-specs/blob/ed16e10caee8ef2ab09f321272877d0efbec0d1e/.github/workflows/src/arm-auto-signoff/arm-incremental-typespec.js#L24-L30

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+
+  const rmFiles = allFiles.filter((file) => resourceManager(file) && file.startsWith('specification/'));
```
mikeharder: shouldn't need to filter to startswith specification, if you filter on the call to git diff.  see above, and existing calls to resourceManager filter.

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+
+  if (rmFiles.length === 0) {
+    console.log('No resource-manager files changed');
```
mikeharder: inside a workflow JS, better to use `core.info`.  see existing sources.

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+      'The following new resource provider namespace(s) were detected in this PR:\n',
+      rpList,
+      '\n### Action Required',
+      'New resource provider namespaces require having a discussion with the ARM team at **ARM API Modeling Office Hours** before merging.\n',
+      '**Please schedule here:**',
+      'https://outlook.office365.com/book/ARMOfficeHours1@microsoft.onmicrosoft.com/?ismsaljsauthenabled=true'
+    ].join('\n');
+
+    try {
+      await github.rest.issues.createComment({
```
mikeharder: adding and updating a PR comment requires write permissions, and is trickier than you might think.  maybe sufficient to call `core.error` to log an error?

https://github.com/actions/toolkit/tree/main/packages/core#logging

.github/workflows/detect-new-resource-provider.yaml
mikeharder, line 1: We should think more about the name, something short but easy for the PR author to understand.  Maybe just "ARM Lease", "ARM Lease Validation", "ARM Lease Requirement", etc?

mikeharder: Or if you think this workflow could expand over time, something even more generic like "RP Validation", "ARM Validation"?

tejaswiMinnu: This workflow will primarily focus on leases, so ARM Lease validation could be appropriate

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Done modified to core.info

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: @raosuhas Is it fine to fail just the workflow and the log the details, or do we need to add a comment?

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: modified similar to arm-incremental-typespec.js

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: modified similar to arm-incremental-typespec.js, but cwd, I'm utilizing getRootFolder(process.cwd()), which is initialized early in this function and also follows the recommendation provided in another comment https://github.com/Azure/azure-rest-api-specs/pull/39625#discussion_r2791005928

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: done

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: done

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: moved label to labelling.js

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: done

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: removed

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: I don't think command line code is needed, I have removed that part.

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Modified as suggested, imported Z from zod and created a schema object

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Right thanks, modified to resolve() to ensure absolute path

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: done modified

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: modified

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: removed cmd code

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: removed

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: Removed

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: Sure, I added the condition in labelling.js and moved the NotReadyForARMReview label there.

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: sure

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: removed

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: replaced as is

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: I choose the first option if the condition is false, since adding workflow_dispatch requires manual intervention to trigger and identify new RP's.

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {string} resourceProvider - Resource provider namespace
+ * @param {string} serviceGroup - Optional service group
+ * @returns {string} Full path to lease.yaml file
+ */
+function buildLeasePath(repoRoot, serviceName, resourceProvider, serviceGroup = '') {
+  const leasePathParts = [repoRoot, '.github', 'arm-leases', serviceName, resourceProvider];
+  if (serviceGroup) {
+    leasePathParts.push(serviceGroup);
+  }
+  leasePathParts.push('lease.yaml');
```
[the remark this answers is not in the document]
tejaswiMinnu: Optional Service group is required, Updated the JS comments.

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: will check with suhas

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+  const newResourceProviders = [];
+  
+  for (const [rp, info] of changedResourceProviders) {
+    if (!resourceProviderExists(specPath, rp)) {
+      newResourceProviders.push({ 
+        namespace: rp, 
+        serviceName: info.serviceName,
+        serviceGroup: info.serviceGroup 
+      });
+    }
```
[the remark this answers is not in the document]
tejaswiMinnu: This comment is stale now as it was addressed part of Itr 1's commits 
The old resourceProviderExists() that used existsSync on the filesystem has been replaced with resourceProviderExistsInCommit()  (line 221), which checks against the merge base commit using [git ls-tree]

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+
+  // Extract resource providers from changed files
+  const changedResourceProviders = extractResourceProviders(rmFiles);
+
+  // Pre-check: Verify if spec directories are brand new (don't exist in base branch)
+  const git = simpleGit(repoRoot);
+  const changedSpecDirs = new Set([
+    ...rmFiles.filter(swagger).map((f) => dirname(dirname(dirname(f)))),
+    ...rmFiles.filter(example).map((f) => dirname(dirname(dirname(dirname(f))))),
+  ]);
```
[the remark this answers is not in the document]
tejaswiMinnu: Stale

.github/workflows/src/detect-new-resource-provider.js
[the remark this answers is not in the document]
tejaswiMinnu: changed the fallback from [origin/${baseBranch} which may not exist locally with fetch-depth: 2 to HEAD^ (guaranteed to exist). This matches the convention used in other workflows like arm-incremental-typespec.js and the default in getChangedFiles()

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+// ============================================
+
+/**
+ * Main detection logic for GitHub script action
+ * @param {Object} params - Parameters from github-script
+ * @param {Object} params.github - GitHub API client
+ * @param {Object} params.context - GitHub context
+ * @param {Object} params.core - GitHub Actions core
+ * @param {string} params.baseBranch - Base branch name
+ * @returns {Promise<string>} Result status
```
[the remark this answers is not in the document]
tejaswiMinnu: Update similar adjusting the new changes

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,303 @@
...
+        body: commentBody
+      });
+      console.log('PR comment created successfully');
+    } catch (error) {
+      console.log(`[FAILED] Failed to create PR comment: ${error.message}`);
+    }
+    
+    // Set the action as failed to indicate new RPs were detected
+    core.setFailed('New resource providers detected');
+    return {
```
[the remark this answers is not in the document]
tejaswiMinnu: This comment is stale as part of code fixes

.github/workflows/detect-new-resource-provider.yaml
tejaswiMinnu: @mikeharder I'm removing write permissions, but I am but curious that some workflows have write permissions. How are these different from the others?

.github/workflows/detect-new-resource-provider.yaml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: ubuntu-24.04
+    if: env.ENABLE_NEW_RP_DETECTION == 'true'
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 2
+          sparse-checkout: |
+            .github
```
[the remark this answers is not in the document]
tejaswiMinnu: This comment is no longer valid

.github/workflows/test/detect-new-resource-provider.test.js
```diff
@@ -0,0 +1,325 @@
...
+        github: mockGithub,
+        context: mockContext,
+        core: mockCore,
+        baseBranch: 'main'
+      });
+
+      expect(result.status).toBe('no-changes');
+      expect(mockCore.setFailed).not.toHaveBeenCalled();
+      expect(mockGithub.rest.issues.createComment).not.toHaveBeenCalled();
+    });
```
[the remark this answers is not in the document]
tejaswiMinnu: Implemented as suggested

.github/workflows/detect-new-resource-provider.yaml
mikeharder: Any workflows in the specs repo with write access, should only use triggers like `pull_request_target` or `workflow_run`, where the workflow source always comes from the default branch.  Workflows that trigger on `pull_request` should not use write access, in part because it won't even work on fork PRs (PRs from user forks, can never get write access to the main fork).

.github/workflows/detect-new-resource-provider.yaml
mikeharder: While we **can** merge this workflow with the "if: false" setting, what's the point?  The workflow can never actually be run, without removing the "if: false" setting.  So this is the equivalent of just keeping this PR open for now, or commenting-out all the code and merging it to main.

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  detect-new-rp:
+    name: Detect New Resource Provider
+    runs-on: ubuntu-24.04
+    # TODO: Set to true to enable new RP detection
+    if: false
```
mikeharder: While we can merge this workflow with the "if: false" setting, what's the point? The workflow can never actually be run, without removing the "if: false" setting. So this is the equivalent of just keeping this PR open for now, or commenting-out all the code and merging it to main.

Also, this causes an error from our actionlint validator, we'd need to suppress:

```
Error: .github/workflows/arm-lease-validation.yaml:26:9:
  constant expression "false" in condition. remove the if: section [if-cond]
verbose: Found 1 errors in 45 files
   |
26 |     if: false
   |         ^~~~~
```

.github/workflows/arm-lease-validation.yaml
```diff
@@ -0,0 +1,53 @@
...
+    paths:
+      - "specification/**/resource-manager/**"
+
+permissions:
+  contents: read
+
+jobs:
+  detect-new-rp:
+    name: Detect New Resource Provider
+    runs-on: ubuntu-24.04
```
mikeharder:
```suggestion
    runs-on: ubuntu-slim
```

We've started moving our lighter-weight workflows to `ubuntu-slim` to use less resources.  This workflow should be another good candidate.

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+ * ```yaml
+ * lease:
+ *   startdate: "2024-01-01"
+ *   duration-days: "P30D"
+ * ```
+ */
+const leaseSchema = z.object({
+  lease: z.object({
+    startdate: z.string().regex(/^\d{4}-\d{2}-\d{2}$/, 'startdate must be in YYYY-MM-DD format'),
+    'duration-days': z.string().regex(/^P(\d+)D$/i, 'duration-days must be in ISO 8601 format (e.g. P180D)'),
```
mikeharder: Change to one of these options:

1. keep name `duration-days`, but make value just an integer
2. rename to `duration`, but then support all ISO8601 formats, not custom parsing.  use a library like `@js-temporal/polyfill`. https://tc39.es/proposal-temporal/docs/duration.html.

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Build the lease path based on service information.
+ * 
+ * Lease files are stored at:
+ * - Without service group: `.github/arm-leases/<serviceName>/<resourceProvider>/lease.yaml`
+ * - With service group:    `.github/arm-leases/<serviceName>/<resourceProvider>/<serviceGroup>/lease.yaml`
+ * 
+ * @param {string} repoRoot - Repository root path
+ * @param {string} serviceName - Service name (e.g., "compute")
```
mikeharder: we should try to align the names of path segments, with our latest folder structure guidelines:

```
specification/
└── <orgName1>/
    ├── cspell.yaml
    └── resource-manager/
        ├── readme.md			ß NOTE: For ARM schema validation; see bullet #2 below
        └── <RPNamespace>/		ß NOTE: Control-plane only (not data-plane) 
            └── <ServiceName1>/	ß Customer-facing service name; each version gets Documentation & SDK package
```

https://github.com/Azure/azure-rest-api-specs/wiki/Azure-Service-Versioning-Guideline

so maybe rename your vars like so?

```
serviceName -> orgName (OR organizationName)
resourceProvider -> rpNamespace
serviceGroup -> service (OR sericeName)
```

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Check if ARM lease exists and is valid
+ * @param {string} serviceName - Service name
+ * @param {string} resourceProvider - Resource provider namespace
+ * @param {string} serviceGroup - Optional service group
+ * @returns {Promise<{ exists: boolean, valid: boolean, leasePath: string }>} Lease status
+ */
+export async function getLeaseStatus(serviceName, resourceProvider, serviceGroup = '') {
+  const repoRoot = process.env.TEST_REPO_ROOT || await getRootFolder(process.cwd());
```
mikeharder:
```suggestion
  const repoRoot = await getRootFolder(process.cwd());
```

I don't think this test env var should be necessary, you can use mocking in a unit test

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+    const rawParsed = /** @type {any} */ (yaml.load(content, { schema: yaml.FAILSAFE_SCHEMA }));
+
+    if (!rawParsed) {
+      return { exists: true, valid: false, leasePath };
+    }
+
+    const parsed = leaseSchema.parse(rawParsed);
+    const lease = parsed.lease;
+
+    const durationDays = parseInt(lease['duration-days'].match(/^P(\d+)D$/i)[1], 10);
```
mikeharder: either support full range of ISO8601 and convert to days, or just accept an integer

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * Check if ARM lease exists and is valid
+ * @param {string} serviceName - Service name
+ * @param {string} resourceProvider - Resource provider namespace
+ * @param {string} serviceGroup - Optional service group
+ * @returns {Promise<{ exists: boolean, valid: boolean, leasePath: string }>} Lease status
+ */
+export async function getLeaseStatus(serviceName, resourceProvider, serviceGroup = '') {
```
mikeharder:
```suggestion
async function getLeaseStatus(serviceName, resourceProvider, serviceGroup = '') {
```

I don't think this needs to be exported?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
+import { writeFileSync } from 'fs';
```
mikeharder:
```suggestion
import { writeFile } from 'fs/promises';
```

always prefer async apis

.github/workflows/src/detect-arm-leases.js
```diff
@@ -0,0 +1,101 @@
...
+ * Schema for lease.yaml file
+ * 
+ * Example:
+ * ```yaml
+ * lease:
+ *   startdate: "2024-01-01"
+ *   duration-days: "P30D"
+ * ```
+ */
+const leaseSchema = z.object({
```
mikeharder: should this schema be shared code between this PR and the other PR?  we shouldn't duplicate a schema for the same logical object structure.

.github/workflows/src/detect-arm-leases.js
mikeharder, line 1: for simplicity, i think this file should be merged with the other file, and renamed to match the name of the workflow (so `arm-lease-validation.js`.  if you would like multiple js files for your wf, please create a folder with the same name as your wf.  for example the `arm-auto-signoff` folder.

.github/workflows/test/detect-arm-leases.test.js
```diff
@@ -0,0 +1,177 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdirSync, writeFileSync, rmSync, existsSync } from 'fs';
```
mikeharder: we never write actual files in unit tests.  one option is to create a read-only `fixtures` folder where you can commit test lease files, that can be read by a unit test.  but if you need dynamic content, this doesn't work.

so then you can either mock the filesystem calls using vi mocking, or refactor the code to split the file system read from the parsing / business logic, and just test the latter.

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+      });
+      
+      const leaseStatus = leaseValid ? 'Lease valid' : 'Lease invalid';
+      core.info(`  - ${rp.namespace}: ${leaseStatus}`);
+    }
+    
+    writeFileSync(
+      join(repoRoot, '.github', 'new-rp-output.json'),
+      JSON.stringify({ newResourceProviders: leaseCheckResults }, null, 2)
+    );
```
mikeharder:
```suggestion
```

I don't see any purpose to writing this file.  I don't see it used anywhere else.  and if you do need to write a file, never write into the repo itself.  always a temp folder.

.github/workflows/test/detect-arm-leases.test.js
```diff
@@ -0,0 +1,177 @@
...
+import { join } from 'path';
```
mikeharder:
```suggestion
import { resolve } from 'path';
```

almost always better to use `resolve` than `join` to ensure absolute paths

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+
+  // Get the merge base for proper PR comparison.
+  // With fetch-depth: 2, HEAD is the PR merge commit and HEAD^ is the base commit.
+  // git merge-base may fail if origin/<branch> isn't fetched, so fall back to HEAD^.
+  let mergeBase;
+  try {
+    mergeBase = await git.raw(['merge-base', `origin/${baseBranch}`, 'HEAD']).then(s => s.trim());
+  } catch {
+    mergeBase = 'HEAD^';
+  }
```
mikeharder: are you sure you need all this complexity?  the other place we call ls-tree, we just hardcode HEAD^:

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20ls-tree&type=code

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+
+  // Get all changed files in specification/
+  const options = {
+    cwd: repoRoot,
```
mikeharder:
```suggestion
    cwd: process.env.GITHUB_WORKSPACE,
```

see existing usage in arm-auto-signoff

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+  };
+
+  const changedFiles = await getChangedFiles(options);
+
+  // Filter to resource-manager files
+  const rmFiles = changedFiles.filter(resourceManager);
+
+  if (rmFiles.length === 0) {
+    core.info("No changes to files containing path '/resource-manager/'");
+    return {
```
mikeharder: if no RM files, shouldn't you remove the label, not no-op?  because if the label was added previously you want to remove it now?

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+  if (changedNamespacePaths.size > 0) {
+    let hasAtLeastOneBrandNewRP = false;
+    
+    for (const namespacePath of changedNamespacePaths) {
+      try {
+        const specFilesBaseBranch = await git.raw([
+          'ls-tree',
+          '-r',
+          '--name-only',
+          mergeBase,
```
mikeharder:
```suggestion
          "HEAD^",
```

.github/workflows/src/detect-new-resource-provider.js
```diff
@@ -0,0 +1,287 @@
...
+          .split('\n')
+          .filter((file) => resourceManager(file) && swagger(file));
+
+        if (specRmSwaggerFilesBaseBranch.length === 0) {
+          hasAtLeastOneBrandNewRP = true;
+        }
+      } catch {
+        // Directory doesn't exist in base - brand new RP
+        hasAtLeastOneBrandNewRP = true;
+      }
```
mikeharder: catching all errors and just setting a boolean, without at least logging the error, is scary.  either scope to a smaller code block, only catch a certain type of error, etc.

.github/workflows/src/detect-new-resource-provider.js
mikeharder: actually, is this codepath even possible, since the WF only triggers on changes to resource-manager/**?

mikeharder (changes requested):

.github/workflows/arm-lease-validation.yaml
```diff
@@ -0,0 +1,53 @@
+name: ARM Lease Validation
```
mikeharder, line 1: do you think this name is the best description for spec PR authors?  do they think in terms of "arm leases"?

for just the functionality you have here. maybe a better name would be "ARM Modeling Review"?  since the workflow determines if the spec pr author needs to schedule an arm modeling review or not?

if you want it to be more generic, to add more similar checks in the future, could could name it just "ARM Validation"?

tejaswiMinnu: "ARM Modeling Review" will best suit and easy to understand for spec pr authors

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: Tried to catch and log the errors and scoped to smaller code blocks

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: done

.github/workflows/arm-lease-validation.yaml
tejaswiMinnu: updated

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: removed, it is used internally

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: sure

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Yes, once the other PR is merged, I will reuse the existing code and avoid duplicates

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: I have created a folder

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: removed

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: updated

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: removed

.github/workflows/test/detect-arm-leases.test.js
tejaswiMinnu: removed

.github/workflows/src/detect-new-resource-provider.js
tejaswiMinnu: removed

.github/workflows/arm-modeling-review.yaml
tejaswiMinnu: Yes, we are supposed to release this on March 21st. If it gets shipped accidentally, I don’t want to create any hassle for everyone.
I’m testing everything on my feature branch and will remove it once testing is complete.

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: chose duration and imported the library

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: done, renamed as suggested

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: Yes, sorry I was playing around the code and missed to remove this

.github/workflows/src/detect-arm-leases.js
tejaswiMinnu: supported full range of ISO8601

.github/workflows/test/detect-arm-leases.test.js
tejaswiMinnu: Modified all the files to align with arm-auto-signedoff.

tejaswiMinnu:
Tested this WF in feature branch and made changes accordingly which satisfies the below:
https://github.com/tejaswiMinnu/azure-rest-api-specs/pull/28 - detected new api path which is a new resource type and added ARMModelingReviewRequired
https://github.com/tejaswiMinnu/azure-rest-api-specs/pull/30 - detected new rpnamespace which is a new resource provider namespace and added ARMModelingReviewRequired and next steps to merge by GitHub bot

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          name: "ARMModelingAutoSignedOff"
+          value: "${{ fromJson(steps.detect.outputs.result).labelActions['ARMModelingAutoSignedOff'] == 'add' }}"
+
+      - name: Upload artifact for issue_number
+        uses: ./.github/actions/add-empty-artifact
+        with:
+          name: issue-number
+          value: "${{ github.event.pull_request.number }}"
+
```
mikeharder:
```suggestion
```

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,76 @@
...
+        uses: ./.github/actions/add-label-artifact
+        with:
+          name: "ARMModelingAutoSignedOff"
+          value: "${{ fromJson(steps.detect.outputs.result).labelActions['ARMModelingAutoSignedOff'] == 'add' }}"
+
+      - name: Upload artifact for issue_number
+        uses: ./.github/actions/add-empty-artifact
+        with:
+          name: issue-number
+          value: "${{ github.event.pull_request.number }}"
```
mikeharder:
```suggestion
          value: "${{ github.event.pull_request.number }}"

```

.github/workflows/package.json
mikeharder, line 1: move to .github/package.json

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,76 @@
...
+    name: ARM Modeling Review
+    runs-on: ubuntu-slim
+    # TODO: Set to true to enable new RP detection
+    if: false
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0 # Fetch all history for merge-base calculation and diffing
```
mikeharder: i think you should only need `fetch-depth:2` to compute "files changed in this PR", it's all we have in other workflows like `typespec-validation`.  do you know why you need `0` instead of `2`?  `0` is the most expensive and should be avoided unless necessary.

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,76 @@
...
+          sparse-checkout: |
+            .github
+            specification
+            eng
```
mikeharder:
```suggestion
```

if you're going to checkout all of specification, sparse is probably slower than just checkout whole repo

.github/workflows/arm-modeling-review.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: ARM Modeling Review
+        id: detect
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const { default: armModelingReview } =
+              await import('${{ github.workspace }}/.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js');
+            const result = await armModelingReview({ context, core });
+            return result;
```
mikeharder:
```suggestion
            return await armModelingReview({ context, core });
```

.github/workflows/arm-modeling-review.yaml
mikeharder:
```suggestion
```

Since this action runs on trigger `pull_request`, it's not necessary to upload this artifact (and it couldn't be trusted anyway).

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
```diff
@@ -0,0 +1,350 @@
...
+/**
+ * Main detection logic for GitHub script action.
+ *
+ * @param {Object} params - Parameters from github-script
+ * @param {import('@actions/github-script').AsyncFunctionArguments['context']} params.context
+ * @param {import('@actions/github-script').AsyncFunctionArguments['core']} params.core
+ * @returns {Promise<{ status: string, labelActions: ManagedLabelActions }>}
+ */
+export default async function detectNewResourceProvider({ context, core }) {
+  const repoRoot = process.env.GITHUB_WORKSPACE ?? process.cwd();
```
mikeharder:
```suggestion
  const repoRoot = process.env.GITHUB_WORKSPACE;
```

I don't think you should need `process.cwd()` here, we don't use it anywhere else we use `GITHUB_WORKSPACE`.

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
```diff
@@ -0,0 +1,350 @@
...
+  const git = simpleGit(repoRoot);
+
+  core.info("Detecting New Resource Providers");
+
+  // Determine merge base between HEAD and the base branch (usually main)
+  // This ensures we detect all changes introduced in the PR, not just the last commit.
+  let mergeBase = "HEAD^";
+  try {
```
mikeharder: why do you need to use anything other than `HEAD^`?  we don't in any similar checks in the repo.  we always use `HEAD^`, because in the context of a pull request in a github action, this is always the base.  we don't need to calculate anything else.

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
```diff
@@ -0,0 +1,350 @@
...
+    mergeBase = await git.raw(["merge-base", baseRef, "HEAD"]);
+    mergeBase = mergeBase.trim();
+    core.info(`Using merge-base: ${mergeBase} (against ${baseRef})`);
+  } catch (e) {
+    core.info(`Could not determine merge-base, falling back to HEAD^. Error: ${e.message}`);
+  }
+
+  const options = {
+    cwd: process.env.GITHUB_WORKSPACE,
+    baseCommitish: "HEAD^",
```
mikeharder:
```suggestion
```

this is already the default, so you shouldn't need to set it.

.github/workflows/src/arm-lease-validation/detect-new-resource-types.js
```diff
@@ -0,0 +1,305 @@
...
+ */
+
+import { resolve } from "path";
+import { simpleGit } from "simple-git";
+import { ArmHelper } from "@microsoft.azure/openapi-validator-rulesets/dist/native/utilities/arm-helper.js";
+import { SwaggerInventory } from "@microsoft.azure/openapi-validator-core";
+import debug from "debug";
+
+// Disable simple-git debug logging to remove verbose [GitExecutor] logs from the output
+debug.disable();
```
mikeharder: this is usually enabled since it really helps debug a failed/flaky workflow.  why do you need to disable it here?

.github/workflows/src/arm-lease-validation/detect-new-resource-types.js
```diff
@@ -0,0 +1,305 @@
+/**
+ * Detect new ARM resource types in PRs using Azure OpenAPI Validator's ArmHelper.
+ *
+ * Compares resource types between the base branch and HEAD for existing RPs
+ * to identify newly introduced resource types that require ARM lease validation.
+ */
+
+import { resolve } from "path";
+import { simpleGit } from "simple-git";
+import { ArmHelper } from "@microsoft.azure/openapi-validator-rulesets/dist/native/utilities/arm-helper.js";
```
mikeharder: we should not be importing any helpers from legacy code like openapi-validator.  if needed, you can copy the minimal code necessary into the specs repo.

mikeharder: or maybe you could use our new spec-model helper? we can add APIs as needed for your scenario.

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/spec-model.js

mikeharder (changes requested):
see comments, biggest point is we need to remove the dependency on openapi-validator. also fix lint failures (copilot should be able to help).

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
tejaswiMinnu: Removed

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
tejaswiMinnu: Yeah right this is declared but never used, removed the dead code block

.github/workflows/src/arm-lease-validation/detect-new-resource-types.js
tejaswiMinnu: To detect new resource type APIs along with the existing ones, a large number of logs were being generated. That's why, removed now.

.github/workflows/arm-modeling-review.yaml
tejaswiMinnu: Removed

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
tejaswiMinnu: Removed

.github/workflows/arm-modeling-review.yaml
tejaswiMinnu: I initially responded to your comment, but since the local copilot used merge as the base, it reverted back and set depth to 0. I then reverted it again and used head.

.github/workflows/arm-modeling-review.yaml
tejaswiMinnu: Make sense, thanks removed

.github/workflows/src/arm-lease-validation/detect-new-resource-types.js
tejaswiMinnu: I have been using the fallback method as a failsafe, so I will proceed to remove the legacy code.

.github/workflows/src/arm-modeling-review/detect-new-resource-types.js
```diff
@@ -0,0 +1,305 @@
...
+      mergeBase,
+      namespacePath,
+      repoRoot,
+    );
+
+    // Skip namespace if it doesn't exist in base (brand new RP — handled by RP-level detection)
+    if (baseTypes === null) {
+      core.info(` ${namespace}: no resources in base (new namespace), skipping RT detection`);
+      continue;
+    }
```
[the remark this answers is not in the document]
tejaswiMinnu: Implemented suggestion

.github/workflows/src/arm-modeling-review/detect-arm-leases.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {string} serviceName - Optional service name for RPs with sub-groupings (e.g., "ComputeRP")
+ * @returns {string} Full path to lease.yaml file
+ */
+function buildLeasePath(repoRoot, orgName, rpNamespace, serviceName = "") {
+  const leasePathParts = [repoRoot, ".github", "arm-leases", orgName, rpNamespace];
+  if (serviceName) {
+    leasePathParts.push(serviceName);
+  }
+  leasePathParts.push("lease.yaml");
+  return resolve(...leasePathParts);
```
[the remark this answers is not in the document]
tejaswiMinnu: https://github.com/Azure/azure-rest-api-specs/pull/39495 this PR handles arm-leases folder

.github/workflows/src/arm-modeling-review/arm-modeling-review.js
mikeharder, line 1: i think this should be renamed `arm-modeling-review.js` to align with the workflow name?  then how about the folder name?  should both workflow JS files be under the same folder `arm-lease-validation`?

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/**
+ * Main detection logic for GitHub script action.
+ *
+ * @param {Object} params - Parameters from github-script
+ * @param {import('@actions/github-script').AsyncFunctionArguments['core']} params.core
+ * @returns {Promise<{ status: string, labelActions: ManagedLabelActions }>}
+ */
+export default async function detectNewResourceProvider({ core }) {
```
mikeharder: align names of import and export.  either `armModelingReview` or `detectNewResourceProvider`, but don't mix these names, it's confusing.

mikeharder (changes requested):
one remaining issue: align names of workflow yaml and JS entrypoint

.github/workflows/src/arm-modeling-review/arm-modeling-review.js
tejaswiMinnu: arm-modeling-review lets got with this name, updated everything to arm-modeling-review

.github/workflows/src/arm-lease-validation/detect-new-resource-provider.js
tejaswiMinnu: updated

mikeharder (approved):
approved, pending passing checks

.github/workflows/src/arm-modeling-review/detect-arm-leases.js
```diff
@@ -99,14 +100,32 @@ export function parseLease(content) {
  */
 export async function checkLease(orgName, rpNamespace, serviceName = "") {
   const repoRoot = await getRootFolder(process.cwd());
-  const leasePath = buildLeasePath(repoRoot, orgName, rpNamespace, serviceName);
+  const relLeasePath = buildLeaseRelativePath(orgName, rpNamespace, serviceName);
 
-  let content;
+  const git = simpleGit(repoRoot);
+
+  // Try reading from HEAD^ (the base-branch parent of the merge commit).
```
mikeharder, line 107: this seems unnecessary and overly complex.  we don't do this in any other checks AFAIK.  i think the simple fix, just require PRs to update from parent before merging, which should let you read the lease file directly from the merge commit already checked out?  but you can own and maintain this complexity if you want.

mikeharder (approved):

mikeharder (approved):

PR #39634, opened 2026-01-14
Properly annotate  item of sasUris  as secret in `ImportClusterParameters`

welovej:
Service request: sasUris should be an "array of secret strings" instead of "secret array of strings". Therefore we did this change for service team.


anthony-c-martin (approved):
LGTM

samsaha-ms (approved):
LGTM.

TimLovellSmith (approved):

TimLovellSmith:
@welovej Hmm, interesting.

I wonder. Since export and import are basically write-only API, its a bit hard to settle the argument whether it is a secret array of strings, or an array of secret strings.

But I guess the motivation here is that tools etc, don't expect secret arrays, only secret strings?

TimLovellSmith (approved):

PR #39661, opened 2026-01-14
[spec-model] Use MemoryCache to cache expensive properties on Swagger

### LINKED ISSUE #39593: [spec-model] Reduce memory and improve perf loading "network" spec
This PR is failing to run LintDiff:

- https://github.com/Azure/azure-rest-api-specs-pr/pull/26228

It crashes due to out-of-memory, as determined by an instrumented run:

- https://github.com/Azure/azure-rest-api-specs-pr/actions/runs/20937175749/job/60162517478?pr=26266#step:7:21

The crash is a bit complex, because LintDiff runs autorest while still holding the spec-model in memory.  So, there's little memory left when we try to run autorest.

The best fix would be for spec-model to use less memory at rest.  I think the root cause, was this PR to make all instances of `Swagger` hold their content (3 copies!) forever:

- https://github.com/Azure/azure-rest-api-specs/pull/38866

I knew this PR would regress perf, but not to this impact.

Likely fix, Swagger needs to go back to extracting all it's properties the first time any are accessed, and throw away the content as soon as it's no longer needed.

mikeharder:
- More efficient to cache expensive properties statically, rather than in Swagger instances
- May have many Swagger objects wrapping the same content (e.g. refs, different tags)
- Avoids OOM errors running LintDiff on spec "network"
- Fixes #39593

### DIFF
--- a/.github/shared/src/memory-cache.js
+++ b/.github/shared/src/memory-cache.js
@@ -1,0 +1,25 @@
+/**
+ * @template K, V
+ */
+export class MemoryCache {
+  /** @type {Map<K, V>} */
+  #map = new Map();
+
+  /**
+   * Returns cached value, initializing if necessary
+   *
+   * @param {K} key
+   * @param {() => V} factory
+   * @returns {V} cached value
+   */
+  getOrCreate(key, factory) {
+    let value = this.#map.get(key);
+
+    if (value === undefined) {
+      value = factory();
+      this.#map.set(key, value);
+    }
+
+    return value;
+  }
+}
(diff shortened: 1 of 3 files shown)


danieljurek (approved):
A couple things to note: 

1. This cache hides changes to files over the lifetime of the process (I don't expect we'll need to change this assumption given how things work in CI today) 
2. This is another example where using full file paths is beneficial: code using the cache doesn't need to do extra steps to ensure context is correct (e.g. if we had a "root specs directory" in addition to relative file paths)

PR #39665, opened 2026-01-15
Make sdk dirs siblings to Foundry

catalinaperalta:


catalinaperalta:
We've decided to leave the sdk dirs as nested for now

PR #39686, opened 2026-01-15
[SDK Validation] Added schedule to the batch run pipeline

### LINKED ISSUE #12659: [Feature] Create dashboard or automation to report latest SDK batch validation run results for specs in main branch
We already have batch run pipelines, and these runs can be scheduled. It would be ideal to create dashboards or automation to surface the latest run results. This will help us consistently track the generation status of the specs in the `main` branch of the public spec repository and quickly identify any failures or regressions.

raych1:
Partially resolved https://github.com/Azure/azure-sdk-tools/issues/12659

- added weekly schedule to batch run for both `all-mgmtplane-typespecs` and `all-dataplane-typespecs` types
- keep backward compatibility for manual runs

### DIFF
--- a/eng/pipelines/spec-gen-sdk-batch.yml
+++ b/eng/pipelines/spec-gen-sdk-batch.yml
@@ -13,9 +13,32 @@ parameters:
     default: 'sample-typespecs'
     displayName: 'Batch Specs to Run'
 
-extends:
-  template: /eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
-  parameters:
-    SpecRepoUrl: 'https://github.com/$(Build.Repository.Name)'
-    SdkRepoUrl: $(SdkRepoUrl)
-    SpecBatchTypes: ${{ parameters.SpecBatchTypes }}
+trigger: none
+pr: none
+
+schedules:
+  - cron: '0 0 * * 0' # Every Sunday at 00:00 UTC
+    displayName: 'Weekly TypeSpec batch run'
+    branches:
+      include:
+        - main
+    always: true # Run even if there are no code changes
+
+# Determine batch types to run: for scheduled runs use both mgmt and dataplane, otherwise use parameter
+variables:
+  - name: batchTypes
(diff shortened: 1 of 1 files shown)


raych1:
@weshaggard can you review this PR about adding schedules for the SDK Validation batch runs?

eng/pipelines/spec-gen-sdk-batch.yml
```diff
@@ -13,9 +13,32 @@ parameters:
...
-extends:
-  template: /eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
-  parameters:
-    SpecRepoUrl: 'https://github.com/$(Build.Repository.Name)'
-    SdkRepoUrl: $(SdkRepoUrl)
-    SpecBatchTypes: ${{ parameters.SpecBatchTypes }}
+trigger: none
+pr: none
+
+schedules:
```
weshaggard, line 19: Any reason to put the schedule here as opposed to inside of devops? Most devops pipelines we schedule inside the UI.

weshaggard (approved):
One comment but nothing blocking.

eng/pipelines/spec-gen-sdk-batch.yml
raych1: The ADO pipeline UI doesn't allow to set the parameter value, and it only use the default value. In this case, I would like to trigger runs for both `all-mgmtplane-typespecs` and `all-dataplane-typespecs` types. With this yaml-based approach, we can have a single pipeline run with two stages, each running one type.

weshaggard: The parameter is set in the variable sections though and not the schedule. I'm ok the variable condition I was only referring to the schedule.

raych1: Sounds good to move the schedule.

raych1: updated in PR https://github.com/Azure/azure-rest-api-specs/pull/39718

PR #39687, opened 2026-01-15
update the type of flags parameter

xiangyan99:
# Choose a PR Template

### DIFF
--- a/specification/search/data-plane/Search/models-service.tsp
+++ b/specification/search/data-plane/Search/models-service.tsp
@@ -748,7 +748,7 @@ union CharFilterName {
   HtmlStrip: "html_strip",
 }
 
-/** Defines flags that can be combined to control how regular expressions are used in the pattern analyzer and pattern tokenizer. */
+/** Defines a regular expression flag that can be used in the pattern analyzer and pattern tokenizer. */
 union RegexFlags {
   string,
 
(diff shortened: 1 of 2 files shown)


yangylu91:
Context for the fix:

When reviewing the Search data-plane spec, Xiang noticed the `flags` parameter is modeled as a single `RegexFlags`enum but it should actually be a multi RegexFlags that are separated by "|" according to its [description ](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/search/data-plane/Search/preview/2025-08-01-preview/searchservice.json#L4135).

I've double-checked the service code. Service-side implementation [RegexValidationHandler.java](https://msdata.visualstudio.com/Azure%20Search/_git/AzureSearch?path=/Source/Java/OSv2/AzureSearchExtensionsPlugin/src/main/java/microsoft/azure/search/opensearch/plugins/[RegexValidationHandler.java](https://msdata.visualstudio.com/Azure%20Search/_git/AzureSearch?path=/Source/Java/OSv2/AzureSearchExtensionsPlugin/src/main/java/microsoft/azure/search/opensearch/plugins/RegexValidationHandler.java&version=GBmain&line=60&lineEnd=77)&version=GBmain&line=60&lineEnd=77) passes the flags value directly to OpenSearch (`org.opensearch.common.regex.Regex.compile`), which follows Java Pattern semantics. In actual service behavior, multiple regex flags are supported and are expressed as a single string with values combined using the pipe (|) separator. This means the current Swagger enum model does not accurately reflect runtime behavior.

This PR updates the TypeSpec definition so `flags` is modeled as `RegexFlags[]`, while the generated Swagger remains a string.

kuanlu95 (approved):

PR #39706, opened 2026-01-16
[spec-model] Improve performance on large specs

mikeharder:
- [cache.js] Adds KeyedCache and KeyedPairCache helpers
- [path.js] Adds resolveCached helpers (caches results of path.resolve())
- Change call sites to use resolveCached()
- [swagger.js] Cache data both statically and per-instance
- Improves runtime of LintDiff on `network` from 50 minutes to 50 seconds (60x improvement)


.github/shared/cmd/spec-model.js
mikeharder, line 1: testing, will revert

.github/shared/src/swagger.js
```diff
@@ -232,60 +239,72 @@ export class Swagger {
...
-      return (
-        schema
-          .paths("file")
-          // Exclude ourself
-          .filter((p) => resolve(p) !== resolve(this.#path))
+      // Swagger objects should not be cached statically, because they may belong to different Tags, Readmes, or SpecModels.
+      // But, they are safe to cache in this instance.
+      this.#allRefs = new Map(
+        refPaths.map((p) => {
+          // TODO: Can we replace `new Swagger()` with "lookup existing swagger from Tag backpointer"?
```
danieljurek: Should there be an issue on this? Or is it a "for future perf reference"...

danieljurek (approved):

.github/shared/src/swagger.js
mikeharder: Good catch, I should remove this comment, we can't actually do this.

danieljurek (approved):

PR #39708, opened 2026-01-16
[VideoTranslation] Add new API version 2026-03-01 to support auto create first iteration and some other features.

DapengLi2016:
# Choose a PR Template


specification/cognitiveservices/Speech.VideoTranslation/main.tsp
```diff
@@ -274,6 +298,38 @@ model IterationInput {
...
+  @doc("Managed identity client ID for user-assigned managed identity to push results to Azure storage blob, if not specified, API will use system-assigned managed identity. This property is only used when pushResultToAzureStorageBlobDirUrl is specified.")
+  pushResultToAzureStorageBlobManagedIdentityClientId?: Guid;
+}
+
+@added(ApiVersions.v2026_03_01)
+@doc("Translation properties.")
+model TranslationProperties {
+  @doc("Time to live in hours for the translation data. After the specified hours, the data will be deleted automatically.")
+  @visibility(Lifecycle.Read)
+  timeToLiveInHours?: int32;
```
weidongxu-microsoft: nit, another pro for a date-time e.g. `expireAt` (compared to a TTL value that changes in every GET call), is for caching. Customer may choose to cache the result (or a proxy may auto do that).

specification/cognitiveservices/Speech.VideoTranslation/main.tsp
```diff
@@ -197,6 +209,18 @@ model TranslationInput {
...
+  @doc("Indicate whether to automatically create the first iteration when creating a translation, if not specified, the default value is false without creating the first iteration automatically.")
+  autoCreateFirstIteration?: boolean;
+
+  @added(ApiVersions.v2026_03_01)
+  @doc("Input file source kind, if not specified, the default value is to access input file with anonymous access.")
+  inputFileSourceKind?: InputFileSourceKind;
+
+  @added(ApiVersions.v2026_03_01)
+  @doc("Managed identity client ID for user-assigned managed identity to access input file from Azure storage blob, if not specified, API will use system-assigned managed identity. This property is only used when inputFileSourceKind is AzureStorageBlobManagedIdentity.")
+  inputFileAzureStorageBlobManagedIdentityClientId?: Guid;
```
weidongxu-microsoft, line 223: nit, You may use this pattern, if there is known to be other "inputFileSourceKind" in future
https://github.com/mcgallan/azure-rest-api-specs/blob/572128db1eaf49bd7d8d61397a5297de6fda09c6/specification/cognitiveservices/data-plane/LanguageAnalyzeDocuments/models/common.models.tsp#L469-L494

If there unlikely to be more, current API should be sufficient.

---

BTW, a question on `videoFileUrl` -- Is it must be an azure blob URL? Or it can be any public URL to a video file?

weidongxu-microsoft (approved):

specification/cognitiveservices/Speech.VideoTranslation/main.tsp
DapengLi2016: ok, changed to expiresDateTime

specification/cognitiveservices/Speech.VideoTranslation/main.tsp
DapengLi2016: Mainly, the kind field in my case does not have a one‑to‑one relationship with the other properties.
InputFileSourceKind corresponds to VideoFileUrl and AudioFileUrl, and also another API value Iteration.Input.Webvtt.Url — these three values — plus Iteration.Input.TtsCustomLexiconFileUrl.
So using extend in this situation is not very suitable.
In addition, because these properties were already defined in previous versions of the API, and to make upgrades easier for users, we want to keep the existing property schema unchanged. Therefore, it is also not appropriate to use extend to modify the already released schema, as this would increase the schema‑upgrade overhead for users.

mikekistler (approved):
Looks good. 👍

weidongxu-microsoft (approved):

specification/cognitiveservices/Speech.VideoTranslation/main.tsp
```diff
@@ -326,6 +326,10 @@ model IterationInput {
   @added(ApiVersions.v2026_03_01)
   @doc("Managed identity client ID for user-assigned managed identity to push results to Azure storage blob, if not specified, API will use system-assigned managed identity. This property is only used when pushResultToAzureStorageBlobDirUrl is specified.")
   pushResultToAzureStorageBlobManagedIdentityClientId?: Guid;
+
+  @added(ApiVersions.v2026_03_01)
+  @doc("When this parameter is set to true, the service uses a higher-sample-rate 24 kHz prompt audio for voice prompting. If this parameter is not specified or is set to false, the service continues to use 16 kHz prompt audio by default, as it generally provides better overall synthesis quality.")
+  use24kPromptAudio?: boolean;
```
weidongxu-microsoft, line 332: nit, If service may eventually add 48k sampling, API may better treat this as extensible enum.

DapengLi2016: Thanks for the comments, I just do internal discussion, and we think we will keep 24k as the single optional for a long time, and considering simplify the schema, so we want to keep using the bool? use24kPromptAudio here.

rkuthala (approved):

PR #39790, opened 2026-01-21
Removed duplicatecheck param from query APIs

AnilKumarPolepalli:
# Choose a PR Template


mikekistler (changes requested):
Removing a parameter is a breaking change so it must  be done in a new API version.

AnilKumarPolepalli:
> Removing a parameter is a breaking change so it must be done in a new API version.

I agree with your comments. Currently, our codebase does not include a new API version, but these Changes should be documented at this stage. In the next Change, we will add API versioning.

mikekistler (approved):
Apparently this parameter was never supported in the version of the service available to customers. So we can consider this a BugFix but also a process failure.

weidongxu-microsoft (approved):
from chat, this property was never supported in service.

PR #39806, opened 2026-01-21
Fix: Change getVersionCode and swapDefault to use empty body model and delete response codes for SDK generation

tundwed:
Change getVersionCode and swapDefault to use empty body model for SDK generation

- Updated delete operations to include 200 response code as success
- This ensures SDK clients send Content-Type: application/json with {} body, matching the RP requirement for EmptyBody model

# Choose a PR Template


specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/main.tsp
```diff
@@ -116,21 +123,29 @@ interface EdgeActionVersions {
...
 
+  /** Get the version code for the edge action version. */
+  #suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "Empty body with Content-Type: application/json required by service"
+  getVersionCode is ArmResourceActionAsync<EdgeActionVersion, {}, VersionCode>;
+
+  /** Swap the default version for the edge action. */
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "Operation matches 2024 baseline with 200, 202 responses"
-  swapDefault is ArmResourceActionNoContentAsync<EdgeActionVersion, void>;
+  #suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "Empty body with Content-Type: application/json required by service"
+  swapDefault is ArmResourceActionNoContentAsync<EdgeActionVersion, {}>;
```
weidongxu-microsoft: nit, pro and con of a named model, instead of `{}`?

The current anonymous empty model would be named `SwapDefaultRequest` by TCGC. Not bad though.

tundwed: @weidongxu-microsoft just to clarify, if we were to create a named model, that should not impact other clients like cli that are sending an empty body, right?

For a little bit more context, this takes an empty body because the ProviderHub TypeSpec (that generate the controller models and controllers) templates (ArmResourceActionAsync, ArmResourceActionNoContentAsync) require the request body parameter to be a Model type. void is not a valid model type in this context. Hence why we have an empty body.

weidongxu-microsoft: My question is mostly for SDK dev @msyyc . A named model vs. anonymous model is only about the name. The content of it would be same.
E.g. in Swagger, the difference is named schema vs. anonymous .
image

However, whether CLI can handle it as expected maybe a question for CLI, as they likely still generate SDK from Swagger.

BTW, I thought the context is mostly a Asp.NET issue that it cannot handle request without context-type, even if the request does not have a body.

msyyc: image

Generated python SDK for current tsp is `body: Any`. SDK users will be confused even if they could pass `client.xxx(..., body={})` to mitigate the service issue. 

@tundwed, after offline discussion with @weidongxu-microsoft, we think it is better to define a single header parameter like https://github.com/Azure/azure-rest-api-specs/pull/39806/files#r2726210038

msyyc:
```suggestion
  swapDefault is ArmResourceActionNoContentAsync<EdgeActionVersion, void, Parameters={@header contentType: "application/json";}>;
```

tundwed: Single header parameter sounds good.

tundwed: @msyyc i have applied the change. would you have a look again

msyyc: Looks good to me. Ping other language owners to review @weidongxu-microsoft / @MaryGao / @tadelesh

msyyc (approved):
Approved for python sdk

tundwed:
> Approved for python sdk

Thanks @msyyc ! Are you also able to do the versioning review/suppression?

msyyc:
> > Approved for python sdk
> 
> Thanks @msyyc ! Are you also able to do the versioning review/suppression?

@tundwed I am only Python SDK owner. If you have questions about the review, pls seek help in [teams channel](https://teams.microsoft.com/l/channel/19%3A0351f5f9404446e4b4fd4eaf2c27448d%40thread.skype/API%20Spec%20Review?groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47) and there will be experts to anwser.

weidongxu-microsoft:
@tundwed

The breaking changes are on the 200 status code in DELETE.

If you confirm that THIS IS THE SERVICE BEHAVIOR (this is an LRO DELETE, but service returns 200 directly?), you can add "breakingchange-approved-bugfix" label. It means there is no service break, just your prior TypeSpec/Swagger was wrong, and this PR "fixes" it.

---

You still need to get ARM review.

tundwed:
> breakingchange-approved-bugfix

Thanks @weidongxu-microsoft this is the service behavior. returns 200.

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/main.tsp
MaryGao: @kazrael2119 Could you help confirm the runtime behavior here? I think the requirement for swapDefault operation is  - always having content type but without any requestion body sent.

kazrael2119: In latest http-runtime, it will have content type without request body:
image

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-09-01-preview/openapi.json
```diff
@@ -353,6 +353,9 @@
           }
         ],
         "responses": {
+          "200": {
+            "description": "Resource deleted successfully."
+          },
```
sandipsh, line 358: you are marking it as ArmResourceDeleteWithoutOkAsync but returning Ok? did you manually edit this json file?

it is not clear if it is LRO and you dont want to return 200 and hence you want to suppress error in main.tsp or you want to make it return 200.

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/main.tsp
```diff
@@ -86,7 +86,14 @@ interface EdgeActions {
   get is ArmResourceRead<EdgeAction>;
   create is ArmResourceCreateOrUpdateAsync<EdgeAction>;
   update is ArmCustomPatchAsync<EdgeAction, EdgeActionUpdate>;
-  delete is ArmResourceDeleteWithoutOkAsync<EdgeAction>;
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "Service returns 200 for successful delete operations"
+  delete is ArmResourceDeleteWithoutOkAsync<
```
sandipsh, line 90: >ArmResourceDeleteWithoutOkAsync [](http://example.com/codeflow?start=12&length=31)

json file is returning 200? #Resolved

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/preview/2025-09-01-preview/openapi.json
tundwed: @sandipsh we want to it be async but support also 200. We did not change the json manually, we updated the return type in typespec. Since there is no function `ArmResourceDeleteWithOkAsync` (to support 200) in typespec, we updated the return type to include 200 but still maintained the function `ArmResourceDeleteWithoutOkAsync`

ARM RPC shows the 200, 202 and 204 are valid delete responses

image


Currently the service returns 200 but, in the future, we plan to support 202 (async) as well that is why we made this change

specification/cdn/resource-manager/Microsoft.Cdn/EdgeActions/main.tsp
tundwed: @sandipsh see the updated return type for this function

tundwed:
@AnamikaNupur tagging you for review

AnamikaNupur (approved):

PR #39815, opened 2026-01-21
Use "Tool" or "PreviewTool" as a suffix for Azure Tools (updated PR to target latest TypeSpec branch)

dargilco:
To align with OpenAI naming conventions, use "Tool" suffix for stable Azure tools:
* `AzureAISearchAgentTool` renamed to  `AzureAISearchTool`
* `OpenApiAgentTool` ==> `OpenApiTool`
* `AzureFunctionAgentTool` ==>  `AzureFunctionTool`
* `BingGroundingAgentTool` ==> `BingGroundingTool`

And use "PreviewTool" as suffix for Azure tools in preview:
* `MicrosoftFabricAgentTool` ==> `MicrosoftFabricPreviewTool`
* `SharepointAgentTool` ==> `SharepointPreviewTool`
* `BingCustomSearchAgentTool` ==> `BingCustomSearchPreviewTool`
* `BrowserAutomationAgentTool` ==> `BrowserAutomationPreviewTool`
* `A2ATool` ==> `A2APreviewTool`

Still waiting on a call regarding renaming `MemorySearchTool`.

Also fix some of the doc string for 'type', as it seems like these were not updated when the "_preview" suffix was added to the type string.

This is not a breaking change in REST APIs. This will be a breaking change in emitted beta SDKs, and we will update samples and docs accordingly. 

PR verified by running `npx tsp compile .`

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
@@ -14,16 +14,20 @@ namespace Azure.AI.Projects;
 //
 
 union _AgentToolType {
-  bing_grounding: "bing_grounding",
+
+  // Preview tools (must end with "_preview"):
+  a2a_preview: "a2a_preview",
+  bing_custom_search_preview: "bing_custom_search_preview",
   browser_automation_preview: "browser_automation_preview",
   fabric_dataagent_preview: "fabric_dataagent_preview",
   sharepoint_grounding_preview: "sharepoint_grounding_preview",
+
+  // General availability tools:
   azure_ai_search: "azure_ai_search",
-  openapi: "openapi",
-  bing_custom_search_preview: "bing_custom_search_preview",
-  capture_structured_outputs: "capture_structured_outputs",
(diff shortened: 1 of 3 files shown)


specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -14,16 +14,20 @@ namespace Azure.AI.Projects;
...
+  // General availability tools:
   azure_ai_search: "azure_ai_search",
-  openapi: "openapi",
-  bing_custom_search_preview: "bing_custom_search_preview",
-  capture_structured_outputs: "capture_structured_outputs",
-  a2a_preview: "a2a_preview",
   azure_function: "azure_function",
+  bing_grounding: "bing_grounding",
+  capture_structured_outputs: "capture_structured_outputs",
+  openapi: "openapi",
```
dargilco, line 31: There is no real change in the above list, I just divided them into two groups separating preview and GA tools, and sorting each one alphabetically.

jpalvarezl (approved):

PR #39818, opened 2026-01-21
[LintDiff] Fix multiple before tag correlation candidates

### LINKED ISSUE #38492: [lint-diff] Crash during "correlating runs"

danieljurek:
Supersedes: https://github.com/Azure/azure-rest-api-specs/pull/38698 
Fixes: #38492

* Add specs to clarify and verify expected behavior
* Fix several corner cases where changes to files don't easily map to tags across before and after states


eng/tools/lint-diff/test/processChanges.test.ts
```diff
@@ -56,67 +64,279 @@ describe("getService", () => {
...
       ],
     ]);
+  }
 
-    const [beforeFinal, afterFinal] = reconcileChangedFilesAndTags(before, after);
-    expect(beforeFinal).toEqual(before);
-    expect(afterFinal).toEqual(after);
-  });
+  test.each([
+    [
```
danieljurek, line 139: This is parameterized from our discussion in https://github.com/Azure/azure-rest-api-specs/pull/38698#issuecomment-3560147134

I haven't put time into comparing to Breaking Changes or working on a PrModel. This fixes an issue and we can do broader work on the remaining discussion items here: https://github.com/Azure/azure-rest-api-specs/issues/39831

eng/tools/lint-diff/test/runChecks.test.ts
```diff
@@ -52,26 +52,22 @@ describe("runChecks", () => {
...
+    await expect(actual).rejects.toThrowError("No changed tags found for readme");
   });
 
   test("error path populates error, stdout, stderr", async () => {
     // Consturct an error object that will return true when passed to isExecError
-    const err: ExecError = new Error();
-    err.stdout = "s";
-    err.stderr = "e";
-    err.code = 1;
+    const err = new Error();
```
mikeharder: why is this change needed?

eng/tools/lint-diff/src/runChecks.ts
```diff
@@ -77,15 +79,16 @@ export async function runChecks(
           throw error;
         }
 
+        const execError = error;
```
mikeharder: I don't think you should need this?

eng/tools/lint-diff/src/correlateResults.ts
```diff
@@ -150,11 +152,11 @@ export function getLintDiffViolations(runResult: AutorestRunResult): LintDiffVio
       continue;
     }
 
-    const result = JSON.parse(line.trim()) as { code: string };
-    if (result.code == undefined) {
+    const result = JSON.parse(line.trim()) as unknown;
```
mikeharder: unnecessary?

eng/tools/lint-diff/src/processChanges.ts
```diff
@@ -148,7 +164,7 @@ export async function buildState(
     if (!changedFileAndTagsMap.has(changedReadme)) {
       changedFileAndTagsMap.set(changedReadme, {
         readme: readmeObject,
-        changedTags: new Set<string>(),
+        changedTags: new Set<string>([""]),
```
mikeharder, line 167: why change "empty set" to "set with one empty-string element"?

eng/tools/lint-diff/src/correlateResults.ts
```diff
@@ -150,11 +152,11 @@ export function getLintDiffViolations(runResult: AutorestRunResult): LintDiffVio
...
+    if ((result as { code?: string }).code == undefined) {
```
mikeharder:
```suggestion
    const result = JSON.parse(line.trim()) as { code: string | undefined };
    if (result.code === undefined) {
```

This is probably best.  The original code seems to even work, though technically incorrect since it types as string instead of string|undefined.

eng/tools/lint-diff/src/correlateResults.ts
```diff
@@ -150,11 +152,11 @@ export function getLintDiffViolations(runResult: AutorestRunResult): LintDiffVio
...
       // Results without a code can be assumed to be fatal errors. Set the code
       // to "FATAL"
-      result.code = "FATAL";
+      (result as { code?: string }).code = "FATAL";
```
mikeharder:
```suggestion
      result.code = "FATAL";
```

Original code should just work with reverted typings.

eng/tools/lint-diff/src/correlateResults.ts
```diff
@@ -40,10 +40,12 @@ export async function correlateRuns(
...
       const defaultTag = await getDefaultTag(beforeReadme);
       if (!defaultTag) {
-        throw new Error(`No default tag found for readme ${readme.toString()} in before state`);
+        throw new Error(`No default tag found for readme ${readme.path} in before state`);
       }
       const beforeDefaultTagCandidates = beforeChecks.filter(
-        (r) => relative(beforePath, r.readme.path) === readmePathRelative && r.tag === defaultTag,
+        (r) =>
+          relative(beforePath, r.readme.path) === readmePathRelative &&
+          (r.tag === defaultTag || r.tag == ""),
```
mikeharder:
```suggestion
          (r.tag === defaultTag || r.tag === ""),
```

eng/tools/lint-diff/src/correlateResults.ts
```diff
@@ -215,7 +217,7 @@ export function isSameSources(a: Source[], b: Source[]) {
   return true;
 }
 
-export function arrayIsEqual(a: unknown[], b: unknown[]) {
+export function arrayIsEqual<T>(a: T[], b: T[]) {
```
mikeharder: remove this function, and replace with `util.isDeepStrictEqual()`:

https://nodejs.org/api/util.html#utilisdeepstrictequalval1-val2-options

eng/tools/lint-diff/test/processChanges.test.ts
```diff
@@ -10,9 +10,17 @@ import {
 } from "../src/processChanges.js";
 
 import { Readme } from "@azure-tools/specs-shared/readme";
+import { Tag } from "@azure-tools/specs-shared/tag";
 import { resolve } from "node:path";
 import { isWindows } from "./test-util.js";
 
+declare module "vitest" {
```
mikeharder, line 17: Is this the documented way to add a custom assertion?  Not sure if I like it, over a normal helper method.

mikeharder: I found the docs, and I'm fine with it.

mikeharder (approved):
Approved with minor comments

eng/tools/lint-diff/test/runChecks.test.ts
danieljurek: The `Error` object doesn't have those properties. Intersection is probably a better way to go: 

```typescript
    const err = new Error() as Error & { stdout: string; stderr: string; code: number };
```

eng/tools/lint-diff/src/processChanges.ts
danieljurek: Code in `runChecks.ts` was previously interpreting an empty set as "run this against the readme.md file using the default tag". However, I needed a way to express "default tag" in the set of tags to run against in non-empty sets. This makes it so there's one way to express the default tag: "".

PR #39827, opened 2026-01-22
disable JS flatten breakings for containerregistry

kazrael2119:
disable JS flatten breakings for containerregistry
This change only impacts JS client, not impacts spec


qiaozha (approved):

qiaozha:
Add ARMSignedOff as this change is purely for JS code generation backward compatible issue and doesn't change the real API contract.

PR #39829, opened 2026-01-22
Allow older preview swaggers in TypeSpec validation to support @previewVersion in tsp

qiaozha:
fixes #39674 

## Description

This PR enhances the TypeSpec validation compile rule to allow older preview versions of swaggers to remain in the repository without causing validation failures.

## Changes

- **compile.ts**: Added logic to detect if extra swaggers are only from older preview versions (not the latest preview)
  - Extracts version information from swagger paths
  - Identifies the latest preview version
  - Allows extra swaggers if they are only from older previews
  - Still fails validation for:
    - Extra swaggers from the latest preview version
    - Extra swaggers from stable versions
    - Mixed preview and stable extra swaggers

- **compile.test.ts**: Added comprehensive test cases:
  - Test for allowing older preview versions
  - Test for failing on latest preview versions
  - Test for failing on stable versions
  - Test for allowing multiple older previews
  - Test for failing on mixed preview/stable versions

## Benefits

This change allows services to maintain older preview swagger files in the repository while still enforcing validation that the latest TypeSpec sources generate the expected output. This is useful for maintaining historical API versions without requiring suppression rules.

### DIFF
--- a/eng/tools/typespec-validation/src/rules/compile.ts
+++ b/eng/tools/typespec-validation/src/rules/compile.ts
@@ -162,14 +162,54 @@ export class CompileRule implements Rule {
           const extraSwaggers = remainingSwaggers.filter((s) => !outputSwaggers.includes(s));
 
           if (extraSwaggers.length > 0) {
-            success = false;
-            errorOutput += pc.red(
-              `\nOutput folder '${outputFolder}' appears to contain TypeSpec-generated ` +
-                `swagger files, not generated from the current TypeSpec sources. ` +
-                `Perhaps you deleted a version from your TypeSpec, but didn't delete ` +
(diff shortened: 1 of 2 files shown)


mikeharder:
@qiaozha: Is this supposed to apply to both `resource-manager` and `data-plane`?

mikeharder (approved):
Approved once all tests are passing, and assuming this should apply to all TSP specs, both data-plane and resource-manager

qiaozha:
> @qiaozha: Is this supposed to apply to both `resource-manager` and `data-plane`?

I don't think I have done anything scoped down to resource-manager or data-plane, this should be the same as before.

mikeharder (approved):

PR #39836, opened 2026-01-22
Update JS configs for brownfield services to avoid breakings

kazrael2119:
Update JS configs in tsp configs for brownfield services to avoid lro breakings


qiaozha (approved):

qiaozha:
Adding ARMSignedOff label as this change is for JS codegen backward compatible only.

PR #39854, opened 2026-01-23
[Foundry branch-to-branch] 'v1' versioning w/preview designations

trrwilson:


specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -14,31 +13,29 @@ namespace Azure.AI.Projects;
 //
 
 union _AgentToolType {
-
   // Preview tools (must end with "_preview"):
   a2a_preview: "a2a_preview",
+
```
dargilco: Why is this blank line added?

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -14,31 +13,29 @@ namespace Azure.AI.Projects;
...
   bing_custom_search_preview: "bing_custom_search_preview",
   browser_automation_preview: "browser_automation_preview",
   fabric_dataagent_preview: "fabric_dataagent_preview",
   sharepoint_grounding_preview: "sharepoint_grounding_preview",
+  memory_search_preview: "memory_search_preview",
 
   // General availability tools:
   azure_ai_search: "azure_ai_search",
+
```
dargilco: Why is this blank line added?

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -14,31 +13,29 @@ namespace Azure.AI.Projects;
...
 union _AgentToolType {
-
   // Preview tools (must end with "_preview"):
   a2a_preview: "a2a_preview",
+
   bing_custom_search_preview: "bing_custom_search_preview",
   browser_automation_preview: "browser_automation_preview",
   fabric_dataagent_preview: "fabric_dataagent_preview",
   sharepoint_grounding_preview: "sharepoint_grounding_preview",
+  memory_search_preview: "memory_search_preview",
```
RaviPidaparthi: Can we keep the old discriminator and remove it for V1 version (`@removed(Versions.v1)`)? That way the generated models still support old discriminator.

specification/ai-foundry/data-plane/Foundry/src/openai-responses/models.tsp
```diff
@@ -0,0 +1,188 @@
...
+  // this process has a few steps in TypeSpec:
+  //  1. Define an extended union for the extended set of discriminator label values
+  //  2. Define extended base types using the new extended discriminator
+  //  3. Replace references to pre-extension base types with extended base types
+
+  union _AgentItemType {
+    structured_outputs: "structured_outputs",
+    oauth_consent_request: "oauth_consent_request",
+    memory_search_preview_call: "memory_search_preview_call",
+    workflow_preview_action: "workflow_preview_action",
```
RaviPidaparthi, line 121: Can we keep the workflow action the same as discussed?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -132,23 +151,69 @@ model RaiConfig {
...
+  /**
+   * Configuration options for a text response from the model. Can be plain text or structured JSON data.
+   */
+  text?: {
+    // Tool customization: Employ discriminated type base
+    format?: OpenAI.TextResponseFormatConfiguration;
+  };
+
+  @doc("Set of structured inputs that can participate in prompt template substitution or tool argument bindings.")
+  structured_inputs?: Record<StructuredInputDefinition>;
```
RaviPidaparthi, line 213: You are missing a few of my latest changes like `tool_choice` and the created_by changes

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
trrwilson: The formatter recognizes the comment as applicable to the next line (for a2a) and only that line, then applies the rule to give it a newline break.

I addressed this by creating two temporary unions (one for preview and one for stable) and putting the comments there. It's prettier now.

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
trrwilson: Same deal with the comment and formatter behavior -- should all be addressed.

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
trrwilson: As discussed:
- There's now an artificial, (re)introduced "virtual-public-preview" version for TypeSpec, emplaced *before* "v1" in the versioning chronology
- We now have both the new `memory_search` *and* `memory_search_preview` in that virtual label (contract library should pick both up as long as it targets that version)
- v1, meanwhile, including as consumed by SDK libraries, should *not* pick up these new values

I didn't just make this 2025-11-15-preview because that's not an accurate representation, either; it really is "virtual" in the sense of being a union model set source. It's abusing the tool a bit, but it's contained and fully cleanable once we retire 2025-11-15-preview.

specification/ai-foundry/data-plane/Foundry/src/openai-responses/models.tsp
trrwilson: These values are all reverted, with extension properties added for explanatory purposes in the spec. As discussed, we'll need to be very careful to treat this like pseudo-GA components throughout their preview status.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
trrwilson: Hadn't merged in the last day. Should now all be present, including the side-channel discussions with nullability and structured inputs.

PR #39866, opened 2026-01-23
disable normalize breakings for js

kazrael2119:
disable normalize breakings for js in containerregistry


qiaozha (approved):

qiaozha:
Add ARMSignedOff label as this change is for JS backward compatible not really impact the service API contracts.

PR #39880, opened 2026-01-23
Generating more language for Blobs Storage with TypeSpec

lmazuel:
Cumulative branch for the work done to move Storage from autorest to TypeSpec in JS, Python and Go


specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -705,11 +705,13 @@ model UserDelegationKey {
 
   /** The date-time the key is active. */
   @encode("rfc7231")
+  @alternateType(string, "javascript")
   @Xml.name("SignedStart")
   signedStart: utcDateTime;
 
   /** The date-time the key expires. */
   @encode("rfc7231")
+  @alternateType(string, "javascript")
```
catalinaperalta: @jeremymeng why werent these added as augment decorators in client.tsp?

catalinaperalta: Just these 2 changes that are on an actual model, they should be able to be referenced from client.tsp.

jeremymeng: > @jeremymeng why werent these added as augment decorators in client.tsp?

@catalinaperalta I didn't know how to do it in client.tsp so I followed the existing alternateType on metadata.  Please fix it or give me an example

catalinaperalta: This is an example of how to reference the model property from client.tsp: 

Image

jeremymeng: moved to client.tsp in https://github.com/Azure/azure-rest-api-specs/pull/39880/commits/bd02f0dd3c58b18960630f938d31b8825bfd693d

specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -2375,6 +2375,16 @@ enum BlobDeleteType {
...
+/** The type of blob deletions. */
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
+enum BlobDeleteTypeGo {
+  /** Permanently delete the blob. */
+  Permanent: "Permanent",
+
+  /** Deprecated */
+  None: "None",
+}
+
```
catalinaperalta: @gracewilcox can you move this into client.tsp?

specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -924,6 +930,7 @@ interface Blob {
       /** Optional.  Only possible value is 'permanent', which specifies to permanently delete a blob if blob soft delete is enabled. */
       @query
       @clientName("blobDeleteType")
+      @alternateType(BlobDeleteTypeGo, "go")
```
catalinaperalta: @iscai-msft is there a better/different way to do this type change for go? Without having to override the entire operation? If you see my comment above ([linking here](https://github.com/Azure/azure-rest-api-specs/pull/39880/changes#r3034783416) as well), Grace added an enum to models.tsp which means that the enum will be considered part of the general rest api so ideally we dont do that. With current tcgc functionality I see 2 solutions: 

1. Add a complete override for this operation in client.tsp with the updated type
2. Go patches this manually in the go convenience code. 
3. Some better solution that allows Grace to point to the new enum without having it in models.tsp/considered as part of the main rest api :)

Because we definitely DO NOT want to import client.tsp into this file. So open to suggestions about how to fix!

specification/storage/Microsoft.BlobStorage/models.tsp
catalinaperalta: So Grace pointed out to me that the service might actually accept "None" as well. The other possibility is to unify the BlobDeleteType and BlobDeleteTypeGo enum definitons into 1 enum AND then try to add an `@access(Access.internal)` over the `None` enum value so the other languages dont see it

specification/storage/Microsoft.BlobStorage/routes.tsp
catalinaperalta: Ok given the info that the enum with the "None" value might be more accurate to the service I'm more in favor of unifying the enums and having an access internal over the none value OR if it doesnt affect the other languages since they all have thicker convenience layers then we might not even need the access internal over the None value

catalinaperalta: @iscai-msft disregard please, Grace just overwrote the entire model instead, so we're good now :)

specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -269,6 +270,174 @@ enum BlobDeleteTypeGo {
...
+  immutabilityPolicyMode?: ImmutabilityPolicyMode;
+
+  /** Whether the blob is under legal hold. */
+  @Xml.name("LegalHold") legalHold?: boolean;
+
+  @Xml.name("Acl") ACL?: string;
+  @Xml.name("Group") group?: string;
+  @Xml.name("Owner") owner?: string;
+  @Xml.name("Permissions") permissions?: string;
+  @Xml.name("ResourceType") resourceType?: string;
```
catalinaperalta: @gracewilcox were these properties added because go needs them? 

If so, can you please add doc comments over them /** */

vincenttran-msft (approved):
Cleanly compiles with no diff for Rust as of current branches' state 4/16/26 12:02PM PST 😄 

Image
Thank you for all your hard work and help on this 😺

PR #39882, opened 2026-01-23
[Agents-V2] Add tool_choice_mode for prompt agents and Resolve created_by field conflicts in ItemResources

RaviPidaparthi:
1. Adding tool_choice_mode field to PromptAgentDefinition for controlling tool selection behavior
2. OpenAI already used`created_by` for some of their ItemReource types which conflicts with the one we added in the base ItemResource type. Resolved it by simplifying the created_by union type to separate agent_reference and response_id fields
3. The ResponseObject also uses agent_reference to align with above.


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -260,6 +260,14 @@ model PromptAgentDefinition extends AgentDefinition {
     """)
   tools?: OpenAI.Tool[];
 
+  @doc("""
+    Controls how the model selects which tool (or tools) to use when generating a response.
+    Only applicable when `tools` are specified.
+    `auto` (default): The model can choose between generating a message or calling one or more tools.
+    `required`: The model must call at least one tool before responding.
```
dargilco: I would be better of these two description lines were placed in the `union ToolChoiseMode` instead of here. So we document the values at the location they are defined.

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -260,6 +260,12 @@ model PromptAgentDefinition extends AgentDefinition {
     """)
   tools?: OpenAI.Tool[];
 
+  @doc("""
+    Controls how the model selects which tool (or tools) to use when generating a response.
+    Only applicable when `tools` are specified.
+    """)
+  tool_choice_mode?: ToolChoiceMode;
```
trrwilson: Do we need this to be a new, independent schema type or could we reuse an extension-friendly integration of OpenAI's `tool_choice` to go along with reusing OpenAI's `tools`? If it's not prohibitive, reusing `OpenAI.ToolChoiceParam` here carries some clear advantages -- and we can still tailor the doc string to explain the full range of values isn't (yet) supported.

OpenAI spec source for that: https://github.com/microsoft/openai-openapi-pr/blob/ec403459f006e57c6ed174365e55017d403d99d8/packages/openai-typespec/src/responses/models.tsp#L2141

specification/ai-foundry/data-plane/Foundry/src/responses/models.tsp
```diff
@@ -37,8 +39,12 @@ namespace Azure.AI.Projects;
 @@copyProperties(OpenAI.Response,
   {
     /** The agent used for this response */
+    @removed(Versions.v1)
     agent?: AgentId,
 
+    /** The agent used for this response */
+    agent_reference?: AgentReference,
```
trrwilson, line 56: If we're pulling the band-aid off and renaming `agent` to `agent_reference` on `Response` to mirror `Item`, should we also do the same for symmetry in `CreateResponse`? It feels like we should either do it the same everywhere or *just* have the one place (items) special-cased.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
RaviPidaparthi: I contemplated this and in fact my earlier version was reusing ToolChoiceParam. However, there is nothing in there that user cant do by dirrectly specifying the tools themselves. The only missing piece was the ability to make tool invocation required and hence I added only that.

specification/ai-foundry/data-plane/Foundry/src/responses/models.tsp
RaviPidaparthi: I dont mind it but it would a decent effort updating all docs and samples. Are we ok with that?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
RaviPidaparthi: Fixed it as discussed offline

specification/ai-foundry/data-plane/Foundry/src/responses/models.tsp
RaviPidaparthi: Fix it as discussed offline

trrwilson (approved):

PR #39883, opened 2026-01-23
[Agents-V2] Support tool resources at runtime on response creation

RaviPidaparthi:
This a minimal surgical update to the spec to support tool resources for `file_search`, `code_interpretor` & `mcp` tools at run-time during response create


specification/ai-foundry/data-plane/Foundry/src/responses/models.tsp
```diff
@@ -103,6 +106,73 @@ model AgentId {
   version: string;
 }
 
+/**
+ * Runtime tool resources to provide when creating a response with a agent that has tools configured.
```
dargilco: "an agent" (instead of "a agent")

dargilco (changes requested):

PR #39886, opened 2026-01-23
[Agents-V2] Update list responses Api to return total count

RaviPidaparthi:
Enhanced the response list Api to return `total` count


dargilco (approved):

RaviPidaparthi:
We will revisit this post GA as we dont have an optimal way to compute count on the service yet.

PR #39910, opened 2026-01-26
Fix endpoint type

JoshLove-msft:


specification/cognitiveservices/data-plane/LanguageAnalyzeTextAuthoring/main.tsp
```diff
@@ -35,8 +35,7 @@ using TypeSpec.Versioning;
     /**
      * Supported Cognitive Services endpoint e.g., https://<resource-name>.api.cognitiveservices.azure.com.
      */
-    @format("url")
-    Endpoint: string,
+    Endpoint: url,
```
jorgerangel-msft, line 38: This seems okay to me. Though there are some api diffs for the other languages?

JoshLove-msft: The Python diff seems to be a false positive as there are a bunch of other diffs.

jorgerangel-msft (approved):

PR #39915, opened 2026-01-26
Tables tsp spec migration

annatisch:

> [!TIP]

## PR review workflow diagram

## API Info: The Basics

- [ ] a private preview
- [ ] a public preview
- [ ] GA release

### Change Scope

<sup>This section will help us focus on the specific parts of your API that are new or have been modified. <br/>Please share a link to the design document for the new APIs, a link to the previous API Spec document (if applicable), and the root paths that have been updated. </sup>
* Design Document:
* Previous API Spec Doc:
* Updated paths:

### Viewing API changes

### Suppressing failures

[Swagger-Suppression-Process](https://aka.ms/pr-suppressions) 
to get approval.

### Release planner

A [release plan](https://aka.ms/azsdkdocs/release-plans) should have been created. If not, please create one as it will help guide you through the REST API and SDK creation process. 

<details>

</details>

<details>

### Tooling

### Helpful Links

</details>

## Getting help

  and https://aka.ms/ci-fix.


annatisch:
## Summary of breaking changes
### 1003 - RequestBodyFormatNoLongerSupported (6 errors)
In the original swagger spec - 6 operations were incorrectly documented as using `"application/json;odata=nometadata"` for the request body. According to the REST API specs below, these should always have been `"application/json"`.
- paths['/Tables'].get (As a GET method, should not have had a request content-type)
- paths['/Tables'].post https://learn.microsoft.com/en-us/rest/api/storageservices/create-table#request-headers
- paths['/{table}()'].get (As a GET method, should not have had a request content-type)
- paths['/{table}'].post https://learn.microsoft.com/en-us/rest/api/storageservices/insert-entity#request-headers
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].get (As a GET method, should not have had a request content-type)
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].delete (As a DELETE method, should not have had a request content-type)

### 1004 - ResponseBodyFormatNowSupported (6 errors)
While the success response body default format is `"application/json;odata=minimalmetadata"`, the response body `"application/json"` should also be supported in the following 6 operations as the format of the error response.
- paths['/Tables'].get
- paths['/Tables'].post
- paths['/{table}()'].get
- paths['/{table}'].post
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].get
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].delete

### 1005 - RemovedPath (2 errors)
The following paths have been respecified as follows:
- paths['/?ServiceProperties'] -> paths['?restype=service&comp=properties']
- paths['/?ServiceStats'] -> paths['?restype=service&comp=stats']

### 1006 - RemovedDefinition (1 error)
The `TableEntityProperties` model is now defined as an alias of `Record<unknown>`

### 1010 - AddingRequiredParameter (4 errors)
The "Accept" header was being stripped from requests with no success response body (e.g. status 204), so needed to be explicitly added as otherwise XML error response bodies may be returned for JSON requests, which can cause problems for code generators. This was added to the following 6 operations (only 4 of which show up as breaking changes):
- paths['/Tables('{table}')'].delete
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].put
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].patch
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].delete
- paths['{table}?comp=acl'].put
- paths['?restype=service&comp=properties'].put

### 1011 - AddingResponseCode (1 error)
The original swagger spec was missing the "default" error response for paths['/Tables'].get

### 1014 - RemovingHeader (2 errors)
I'm not sure why we're getting these errors, one for each success response code for paths['/{table}'].post:
- 201 response: [Content-Type header is still specified in the response model](https://github.com/Azure/azure-rest-api-specs/blob/tables-tsp-migration/specification/cosmos-db/data-plane/Tables/models.tsp#L467-L468) - so I don't know why it says it's removed.
- 204 response: No content, so no Content-Type header should be present.

### 1016 - ConstantStatusHasChanged (9 errors)
This is because the previous swagger spec defined the API version header as a constant, which is no longer the case now that it uses the `@apiVersion` decorator. This change applied to every operation, however this error has only been raised for the 9 JSON operations (and ignored for the 5 XML operations).

### 1017 - ReferenceRedirection (118 errors)
This seems to be expected for a tsp migration - I have just ignored all these errors.

### 1023 - TypeFormatChanged (48 errors)
These all fall under one of the following 3 changes:
- "The new version has a different format 'uuid' than the previous one ''."
  The tsp declares both the `"x-ms-client-request-id"` and `"x-ms-request-id"` as uuid, as opposed to the original spec which left them as unformatted string. This can be reverted if preferred.
- "The new version has a different format 'date-time-rfc7231' than the previous one 'date-time-rfc1123'."
  The `"Date"` header now uses the subformat "rfc7231" as "rfc1123" isn't support by tsp.
- "The new version has a different format 'int32' than the previous one ''."
  The previous spec definition of the `"Timeout"` parameter had the type as `"integer"` but never defined the format as `"int32"`.

### 1026 - TypeChanged (8 errors)
- "The new version has a different type '' than the previous one 'object'." - There are 6 instances of this error and it appears to be related to the change from TableEntityProperties being a model with `additionalProperties=true` to an alias for `Record<unknown>`. The newly generated spec uses `additionalProperties={}` instead.
- "The new version has a different type 'object' than the previous one ''." - The original spec definition of "Metrics" was missing the `type: "object"` declaration, which has been added.
- The new version has a different type 'object' than the previous one 'array'. This is because the correct way to declare an XML array that is unwrapped to a tagged element is to declare it as a model with the array as an internal attribute using `@xml.unwrapped`.

### 1033 - RemovedProperty (28 errors)
These are all renames to using tsp naming conventions, while maintaining compatibility using the `@xml.name` decorator:
- TableServiceProperties.Logging -> logging
- TableServiceProperties.HourMetrics -> hourMetrics
- TableServiceProperties.MinuteMetrics -> minuteMetrics
- TableServiceProperties.Cors -> cors
- Logging.Version -> Logging.version
- Logging.Delete -> Logging.delete
- Logging.Read -> Logging.read
- Logging.Write -> Logging.write
- Logging.RetentionPolicy -> Logging.retentionPolicy
- Metrics.Version -> version
- Metrics.Enabled -> Metrics.enabled
- Metrics.IncludeAPIs -> includeApis
- Metrics.RetentionPolicy -> retentionPolicy
- RetentionPolicy.Enabled -> RetentionPolicy.enabled
- RetentionPolicy.Days -> days
- AccessPolicy.Start -> AccessPolicy.start
- AccessPolicy.Expiry -> AccessPolicy.expiry
- AccessPolicy.Permission -> AccessPolicy.permission
- CorsRule.AllowedOrigins -> CorsRule.allowedOrigins
- CorsRule.AllowedMethods -> CorsRule.allowedMethods
- CorsRule.AllowedHeaders -> CorsRule.allowedHeaders
- CorsRule.ExposedHeaders -> CorsRule.exposedHeaders
- CorsRule.MaxAgeInSeconds -> CorsRule.maxAgeInSeconds
- StorageServiceStats.GeoReplication
- GeoReplication.Status -> GeoReplication.status
- GeoReplication.LastSyncTime -> GeoReplication.lastSyncTime
- SignedIdentifier.Id -> SignedIdentifier.id
- SignedIdentifier.AccessPolicy -> SignedIdentifier.accessPolicy

### 1034 - AddedRequiredProperty (28 errors)
These are all renames to using tsp naming conventions, while maintaining compatibility using the `@xml.name` decorator:
- Logging.Version -> Logging.version
- Logging.Delete -> Logging.delete
- Logging.Read -> Logging.read
- Logging.Write -> Logging.write
- Logging.RetentionPolicy -> Logging.retentionPolicy
- Metrics.Enabled -> Metrics.enabled
- RetentionPolicy.Enabled -> RetentionPolicy.enabled
- AccessPolicy.Start -> AccessPolicy.start
- AccessPolicy.Expiry -> AccessPolicy.expiry
- AccessPolicy.Permission -> AccessPolicy.permission
- CorsRule.AllowedOrigins -> CorsRule.allowedOrigins
- CorsRule.AllowedMethods -> CorsRule.allowedMethods
- CorsRule.AllowedHeaders -> CorsRule.allowedHeaders
- CorsRule.ExposedHeaders -> CorsRule.exposedHeaders
- CorsRule.MaxAgeInSeconds -> CorsRule.maxAgeInSeconds
- StorageServiceStats.GeoReplication
- GeoReplication.Status -> GeoReplication.status
- GeoReplication.LastSyncTime -> GeoReplication.lastSyncTime
- SignedIdentifier.Id -> SignedIdentifier.id
- SignedIdentifier.AccessPolicy -> SignedIdentifier.accessPolicy

### 1035 - RemovedOperation (2 errors)
The following paths have been respecified as follows:
- paths['/{table}'].get -> paths['/{table}?comp=acl'].get
- paths['/{table}'].put -> paths['/{table}?comp=acl'].put

### 1038 - AddedPath (5 errors)
The following paths have been respecified as follows:
- paths['/?ServiceProperties'].get -> paths['?restype=service&comp=properties'].get
- paths['/?ServiceProperties'].put -> paths['?restype=service&comp=properties'].put
- paths['/?ServiceStats'].get -> paths['?restype=service&comp=stats'].get
- paths['/{table}'].get -> paths['/{table}?comp=acl'].get
- paths['/{table}'].put -> paths['/{table}?comp=acl'].put

### 1042 - ChangedParameterOrder (61 errors)
None of the parameters in the Tables API are order-sensitive, and this is an expected error in the tsp migration. All of these have been ignored.

### 1045 - AddedOptionalProperty (8 errors)
These are all renames to using tsp naming conventions, while maintaining compatibility using the `@xml.name` decorator:
- TableServiceProperties.Logging -> logging
- TableServiceProperties.HourMetrics -> hourMetrics
- TableServiceProperties.MinuteMetrics -> minuteMetrics
- TableServiceProperties.Cors -> cors
- Metrics.Version -> version
- Metrics.IncludeAPIs -> includeApis
- Metrics.RetentionPolicy -> retentionPolicy
- RetentionPolicy.Days -> days

### 1046 - RemovedOptionalParameter (3 errors)
This appears to have been a bug in the previous swagger spec - as according to the REST API spec, the following 3 operations should never have support the `$format` query parameter:
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].put:
  https://learn.microsoft.com/en-us/rest/api/storageservices/update-entity2#uri-parameters
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].patch
  https://learn.microsoft.com/en-us/rest/api/storageservices/merge-entity#uri-parameters
- paths['/{table}(PartitionKey='{partitionKey}',RowKey='{rowKey}')'].delete
  https://learn.microsoft.com/en-us/rest/api/storageservices/delete-entity1#uri-parameters

### 1049 - RemovedXmsEnum (9 errors)
This is because the previous swagger spec defined the API version header as a constant, which is no longer the case now that it uses the `@apiVersion` decorator. This change applied to every operation, however this error has only been raised for the 9 JSON operations (and ignored for the 5 XML operations).

### 1050 - ParameterLocationHasChanged (9 errors)
This is expected when moving from the previous API version header to using the `@apiVersion` decorator.
This change applied to every operation, however this error has only been raised for the 9 JSON operations (and ignored for the 5 XML operations).

specification/cosmos-db/data-plane/Tables/client.tsp
```diff
@@ -0,0 +1,31 @@
...
+
+/** The signed identifier. */
+model SignedIdentifier {
+  /** The unique ID for the signed identifier. */
+  @Xml.name("Id") id: string;
+
+  /** The access policy for the signed identifier. */
+  @Xml.name("AccessPolicy") accessPolicy?: AccessPolicy;
+}
+@@alternateType(Data.Tables.SignedIdentifier, SignedIdentifier, "python");
```
catalinaperalta, line 31: Was this because python made the accessPolicy prop optional?

specification/cosmos-db/data-plane/Tables/tspconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/data-plane"
+options:
+  "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "data-plane"
+    emitter-output-dir: "{project-root}/.."
+    emit-lro-options: "none"
+    examples-dir: "{project-root}/examples"
+    output-file: "Tables/{version-status}/{version}/table.json"
```
catalinaperalta: Nit: 
```suggestion
    emitter-output-dir: "{project-root}"
    emit-lro-options: "none"
    examples-dir: "{project-root}/examples"
    output-file: "{version-status}/{version}/table.json"
```

specification/cosmos-db/data-plane/Tables/models.tsp
```diff
@@ -0,0 +1,500 @@
...
+  @query
+  $filter?: string;
+};
+
+/** Client request ID header */
+alias ClientRequestIdHeader = {
+  @access(Access.internal)
+  @header("x-ms-client-request-id")
+  @doc("An opaque, globally-unique, client-generated string identifier for the request.")
+  clientRequestId?: uuid;
```
catalinaperalta: Was there a reason to not use the type from Azure.Core?

specification/cosmos-db/data-plane/Tables/models.tsp
```diff
@@ -0,0 +1,500 @@
...
+};
+
+/** Request ID response header */
+alias RequestIdResponseHeader = {
+  @access(Access.internal)
+  @visibility(Lifecycle.Read)
+  @header("x-ms-request-id")
+  @doc("An opaque, globally-unique, server-generated string identifier for the request.")
+  requestId?: uuid;
```
catalinaperalta: Same question here about using the azure.core type?

specification/cosmos-db/data-plane/Tables/routes.tsp
```diff
@@ -0,0 +1,548 @@
...
+import "@typespec/openapi";
+
+using TypeSpec.Http;
+
+namespace Data.Tables {
+  /** Azure Storage basic XML operation template. Not supported by Cosmos endpoint. */
+  op TablesXmlOperation<
+    TParams extends TypeSpec.Reflection.Model,
+    TResponse extends TypeSpec.Reflection.Model
+  >(...ApiVersionHeader, ...TParams, ...ClientRequestIdHeader): (TResponse & {
```
catalinaperalta: Please do change the TParam placement in this signature otherwise every time a new param is added you'll get param ordering errors: 
```suggestion
  >(...ApiVersionHeader, ...ClientRequestIdHeader, ...TParams): (TResponse & {
```

specification/cosmos-db/data-plane/Tables/routes.tsp
```diff
@@ -0,0 +1,548 @@
...
+    ...ClientRequestIdHeader,
+    ...TParams,
+  ): (TResponse & {
+    ...ApiVersionHeader;
+    ...RequestIdResponseHeader;
+    ...ClientRequestIdHeader;
+    ...DateResponseHeader;
+  }) | TablesError;
+
+  namespace Table {
```
catalinaperalta, line 39: Just curious why you used namespace instead of interfaces here? It's totally fine, but checking if  that was intentional for client library shape?

specification/cosmos-db/data-plane/Tables/routes.tsp
```diff
@@ -0,0 +1,548 @@
...
+        /**
+         * The row key of the entity.
+         */
+        @path
+        rowKey: string;
+      },
+      {
+        /** The entity properties. */
+        #suppress "@azure-tools/typespec-azure-core/bad-record-type" "Union types are not supported"
+        @body tableEntityProperties: TableEntityProperties;
```
catalinaperalta, line 214: TableEntityProperties is a `Record<unknown>` not a union of types?

specification/cosmos-db/data-plane/Tables/routes.tsp
```diff
@@ -0,0 +1,548 @@
...
+      prefer?: ResponseFormat,
+    ): CreateTableEchoResponse | CreateTableResponse | TablesError;
+
+    /**
+     * Deletes an existing table.
+     */
+    #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+    @delete
+    @route("/Tables('{table}')")
+    op delete(
```
catalinaperalta, line 116: Just double checking that these first ops didnt fit into either template?

catalinaperalta (commented):
tsp is looking very good...just a few comments/questions

specification/cosmos-db/data-plane/Tables/client.tsp
annatisch: Yes - technically the service accepts a Null AccessPolicy - and the previous Tables SDK had tests for that scenario. However, I cannot find anything in the official docs that this is supported behaviour, and in the previous spec, AccessPolicy was not optional. So I have made this a Python-only alteration to ensure backwards compatibility. I'm not sure if other languages need this - only Python had it as a readme.md configuration for swagger code gen.

specification/cosmos-db/data-plane/Tables/models.tsp
annatisch: Ah I didn't realize that was an option - I will update it :)

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: Good idea!

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: I originally did interfaces - however it wasn't producing the expected structure in the generated code to allow an easy diff. I'm sure this was my error in some way - but the namespace structure seemed to work, so I stuck with it.
Happy to revisit this if the preference is that specs should use interfaces.

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: You're right - I thought when I originally tried to define that it was failing as an error - but the following seems to work:
```
/** The properties for the table entity. */
alias TableEntityProperties = Record<EntityValueType>;

/** The type for values of entity properties */
union EntityValueType {
  string,
  integer,
  float,
  boolean,
}
```

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: hmmm let me double check - originally they didn't, but I shuffled a few things around since then.

annatisch: Okay so there's 3 operations that don't use a template:
- Table.delete: This is because for some reason it's the only JSON operation that doesn't use the `DataServiceVersion` header. I can remove that header from the template, and just explicitly add it to all the operations if you prefer.
- Table.create and Table.insertEntity: These two operations have two success status codes - one with a payload (201) and one without (204) and you set the `prefer` header to configure the response. I wasn't sure if the multiple response model could work with the template, but is there a way to format this?

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: The change above gives me the following compile warning - what's the preferred solution?
```
models.tsp:42:7 - warning @azure-tools/typespec-autorest/union-unsupported: Unions cannot be emitted to OpenAPI v2 unless all options are literals of the same type.
> 42 | union EntityValueType {
     |       ^^^^^^^^^^^^^^^

Found 1 warning.
```

lmazuel (approved):

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: I've reverted this to use Record<unknown> - as that is consistent with the original swagger.

specification/cosmos-db/data-plane/Tables/routes.tsp
catalinaperalta: No that's fine, you can keep them without the template. 

> I wasn't sure if the multiple response model could work with the template

I think this should just work with an | between the 2 types of responses, but if not, then being explicit in your definition is fine :)

specification/cosmos-db/data-plane/Tables/routes.tsp
catalinaperalta: Ah I see, should the union values just be string literals? I think that will probably help with the warning

specification/cosmos-db/data-plane/Tables/routes.tsp
catalinaperalta: namespaces are fine, I was wondering exactly due to the sdk structure. Nested namespaces produce nested clients, while interfaces produce a client with operation groups

specification/cosmos-db/data-plane/Tables/routes.tsp
catalinaperalta: Nevermind, I see that it's used for the type passed into the record. `Record<unknown>` seems fine

specification/cosmos-db/data-plane/Tables/routes.tsp
annatisch: Oh? That seems strange because the namespace declaration I have here seems to produce the operation groups that I needed?

PR #39941, opened 2026-01-27
[openapi-diff-runner] Support renames in same version check

mikeharder:
- fixes #38245
 
 
- Scenario: Change case of spec filename, no change to content
  - Before: Fails with "every API is added/removed"
    - https://github.com/mikeharder/azure-rest-api-specs/actions/runs/21612348823?pr=73
  - After: Correctly determines `NoVersionChange`
    - https://github.com/mikeharder/azure-rest-api-specs/actions/runs/21612351964?pr=71
- Scenario: Change case of spec filename, breaking change in content
   - Before: Fails with "every path is added/removed", which is misleading, and can lead to the failure being bypassed
     - https://github.com/mikeharder/azure-rest-api-specs/actions/runs/21612331302?pr=74
  - After: Fails with correct error for the content diff ("age" renamed to "age2")
    - https://github.com/mikeharder/azure-rest-api-specs/actions/runs/21612340194?pr=75


mikeharder:
@copilot: Create a new PR, starting with the contents of this PR.  I will provide further instructions, in the new PR.

eng/tools/openapi-diff-runner/package.json
```diff
@@ -12,7 +12,7 @@
     "format:check": "prettier . --ignore-path ../.prettierignore --check",
     "format:check:ci": "prettier . --ignore-path ../.prettierignore --check --log-level debug",
     "test": "vitest",
-    "test:ci": "vitest --coverage --reporter=verbose"
+    "test:ci": "vitest run --coverage --reporter=verbose"
```
mikeharder, line 15: unrelated cleanup

raych1 (approved):

eng/tools/openapi-diff-runner/src/commands.ts
```diff
@@ -75,6 +75,8 @@ export async function validateBreakingChange(context: Context): Promise<number>
 
   const deletedSwaggers = diffs.deletions || [];
 
+  const renamedSwaggers = diffs.renames || [];
```
mikeharder, line 78: unfortunately, I think the fix is more complex, and this code likely contains bugs in some rename scenarios.

the problem is the code below, assumes that renames were split into additions and deletions.  for example, `addedVersionDirs` only checks `newSwaggers`, not `renamedSwaggers.to`.  these issues could all be fixed, but this code is already complex and fragile, and this may just make it worse.

eng/tools/openapi-diff-runner/src/commands.ts
mikeharder, line 91: Another place we might need to make case-insensitive.

mikeharder, line 85: any code that processes paths might need to handle case-insensitive.  and then you can't just check existsSync(), you'd instead need to list the dir entries and do case-insensitive compare.

eng/tools/openapi-diff-runner/src/commands.ts
mikeharder: actually, i think this might be fine.  If we can match additions and deletions into a rename, either from git or manually, the result should be strictly better than any "empty file diffs" we would have otherwise done

mikeharder:
closing in favor of #40121

PR #40012, opened 2026-01-29
Foundry Eval Benchmark

AbdelmohsenMS:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
```diff
@@ -83,6 +83,27 @@ model AzureAIDataSourceConfig extends DataSourceConfig {
...
+  benchmark_name: string;
+
+  /** The version of the benchmark specification (e.g., '0.1'). */
+  benchmark_version?: string;
+}
+
+/** Data source configuration for benchmark evaluations. */
+model AzureAIBenchmarkDataSourceConfig extends DataSourceConfig {
+  /** The object type, which is always `azure_ai_source`. */
+  type: "azure_ai_source";
```
johanste: We need a different discriminator, no?

AbdelmohsenMS: This object just needed to define that it includes the metadata: BenchmarkMetadata, if that wasn't needed I would just added the scenario "benchmark_preview" to model AzureAIDataSourceConfig scenarios.

johanste: As described, this shouldn't compile. Discriminated unions need to have and only one model that matches a given discriminator value.

AbdelmohsenMS: So what is your recommended proper change?
Should I just add "benchmark_preview" to the list of scenarios in (model AzureAIDataSourceConfig) ?
But then how specify it require BenchmarkMetadata?

Or better to inherit from AzureAIDataSourceConfig and specify the scenario="benchmark_preview" along with BenchmarkMetadata ?

AbdelmohsenMS: Have solved this by inheriting from AzureAIDataSourceConfig

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
```diff
@@ -82,6 +82,24 @@ model AzureAIDataSourceConfig extends DataSourceConfig {
...
+  benchmark_name: string;
+
+  /** The version of the benchmark specification (e.g., '0.1'). */
+  benchmark_version?: string;
+}
+
+/** Data source configuration for benchmark evaluations. */
+model AzureAIBenchmarkDataSourceConfig extends AzureAIDataSourceConfig {
+  /** Data schema scenario, always `benchmark` for benchmark evaluations. */
+  scenario: "benchmark_preview";
```
johanste, line 89: This does not compile - you are extending the `scenario` property with an invalid value. You need to update the union in AzureAIDataSourceConfig to include it.

AbdelmohsenMS: Got it, thanks.
Have added "benchmark_preview" to AzureAIDataSourceConfig list of scenarios.

AbdelmohsenMS: There is another proposal, let me know how does it resonate with you.
Proposal:
1) Remove AzureAIBenchmarkDataSourceConfig entirely
2) Remove from unions
3) Add "metadata?: BenchmarkMetadata" to AzureAIDataSourceConfig & do validation that it only works when 'scenario' == "benchmark_preview"


model AzureAIDataSourceConfig extends DataSourceConfig {
  type: "azure_ai_source";
  scenario: "red_team" | "responses" | ... | "benchmark_preview";
  metadata?: BenchmarkMetadata;  // ← **Add this**
}

Thoughts?
@johanste

johanste: The proposal is effectively the same as what we are doing but with less specificity (and harder for tooling to deal with). So what we have is better I think.

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
```diff
@@ -457,6 +478,24 @@ model RedTeamEvalRunDataSource extends EvalRunDataSource {
   target: Target;
 }
 
+/** Represents a data source for benchmark evaluation runs. */
+model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
+  /** The type of data source, always `azure_ai_benchmark_preview`. */
+  type: "azure_ai_benchmark_preview";
+
+  /** Input messages configuration. */
+  input_messages?: OpenAI.CreateEvalCompletionsRunDataSourceInputMessagesItemReference;
```
johanste, line 484: How would I use this? Do I have to know the schema of the data source in order to effectively use this? The example in the sdk-first repo seem to imply that (but I also think it has a typo on the field name)....

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
```diff
@@ -79,7 +79,26 @@ model AzureAIDataSourceConfig extends DataSourceConfig {
...
+  benchmark_version?: string;
+}
+
+/** Data source configuration for benchmark evaluations. */
+model AzureAIBenchmarkDataSourceConfig extends AzureAIDataSourceConfig {
+  /** Data schema scenario, always `benchmark` for benchmark evaluations. */
+  scenario: "benchmark_preview";
+
+  /** Metadata identifying the benchmark to use. */
+  metadata: BenchmarkMetadata;
```
johanste: I don't think we need to nest the properties. Is there a reason why we need a separate BenchmarkMetadata model?

AbdelmohsenMS: So the reason for having BenchmarkMetadata in AzureAIBenchmarkDataSourceConfig, is for providing the required inputs (benchmark_name & benchmark_version), which are only coupled with this scenario.
On the contrary: if we removed BenchmarkMetadata; then how user will provide provide the name and version?
Let me know your suggestion for how to better express this.

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
AbdelmohsenMS: In fact this shouldn't be provided by the user, it is auto populated by the service from information stored in asset service, this is needed to be the same for all benchmark runs for apple-to-apple comparison, and the only thing user will provide is the Target agent/model.

I have added this field here so when response comes will show populated inference prompt/messages to user for awareness + for service use as well which rely on it, instead of injecting this info in properties which I think maybe not a clean solution.

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
johanste:
```typespec
model AzureAIBenchmarkDataSourceConfig extends AzureAIDataSourceConfig {
  /** Data schema scenario, always `benchmark` for benchmark evaluations. */
  scenario: "benchmark_preview";
  benchmark: string; // or benchmark_name...
  benchmark_version?: int32;
}
```

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
johanste: As it is defined, it is a valid request body field. Would providing it result in a 400 error?

AbdelmohsenMS: Short answer 'yes'.
Well, I had this concern while designing it and had two options:
- Either to inject it in run properties, which is ugly to show response properties and not matching with other create run scenarios.
- Or to properly have it defined in the request, which will lead to having it properly represented in the response as well, and the caveat that we will not allow user to set it, i.e. being dropped from the request, as it will be auto populated from the service.
The example in https://github.com/Azure/agent-first-sdk/pull/179/changes is correctly not modifying it in the create run call.
Please let me know your thoughts?

specification/ai-foundry/data-plane/Foundry/src/openai-evaluations/models.tsp
AbdelmohsenMS: That makes sense as well, will apply it.

johanste (approved):

PR #40040, opened 2026-01-29
[Storage] Housekeeping TypeSpec Changes

vincenttran-msft:
- Remove `Internal` suffixes from `BlobProperties` and `BlobItem`
- Revert `StorageError` model to the most basic original definition (aside from adding `Code` to allow that to be used and thus generated)
- Use `@@scope(....., !rust)` to hide methods we don't want to be generated at all
- Uses temporary emitter flag `omit-constructors` to get rid of constructors from being generated

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -65,6 +65,5 @@ namespace Customizations;
 @@scope(Storage.Blob.Blob.setExpiry, "!rust");
 @@scope(Storage.Blob.Blob.startCopyFromUrl, "!rust");
 @@scope(Storage.Blob.BlockBlob.query, "!rust");
-@@scope(Storage.Blob.BlockBlob.stageBlockFromUrl, "!rust");
 @@scope(Storage.Blob.PageBlob.copyIncremental, "!rust");
 @@scope(Storage.Blob.PageBlob.getPageRangesDiff, "!rust");
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -27,10 +27,10 @@ namespace Customizations;
 @@clientName(ContainerProperties.immutableStorageWithVersioningEnabled,
   "IsImmutableStorageWithVersioningEnabled"
 );
-@@clientName(BlobPropertiesInternal.expiryTime, "ExpiresOn");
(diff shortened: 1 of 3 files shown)


specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -56,3 +56,15 @@ namespace Customizations;
...
+@@scope(Storage.Blob.Service.getUserDelegationKey, "!rust");
+@@scope(Storage.Blob.Container.listBlobHierarchySegment, "!rust");
+@@scope(Storage.Blob.Container.rename, "!rust");
+@@scope(Storage.Blob.Container.restore, "!rust");
+@@scope(Storage.Blob.Blob.abortCopyFromUrl, "!rust");
+@@scope(Storage.Blob.Blob.copyFromUrl, "!rust");
+@@scope(Storage.Blob.Blob.setExpiry, "!rust");
+@@scope(Storage.Blob.Blob.startCopyFromUrl, "!rust");
+@@scope(Storage.Blob.BlockBlob.query, "!rust");
+@@scope(Storage.Blob.BlockBlob.stageBlockFromUrl, "!rust");
```
jalauzon-msft: This feels like something we need?

specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -16,34 +16,14 @@ scalar base64Bytes extends bytes;
 
 /** The error response. */
 @error
```
jalauzon-msft, line 22: Can you leave a comment in here that this is a minimum impl for the generator and further error customization is done in hand-written code?

Further, I also wonder if this will be the case for all languages or if this is somehow used in other languages. Sean recently updated this a couple of STGs ago for copy error stuff so maybe it's actually used in .NET?

specification/storage/Microsoft.BlobStorage/client.tsp
vincenttran-msft: Hah, seems like we must of missed this. I will add this in this PR next iteration.

specification/storage/Microsoft.BlobStorage/models.tsp
vincenttran-msft: Hmm okay, I wonder if we can ask Sean if he has more context / anybody else on .NET side? For now I will take the top line feedback since that is accurate.

jalauzon-msft (approved):

PR #40047, opened 2026-01-30
Queues tsp

catalinaperalta:
Included in this PR:
- Add missing headers from the original swagger
- Update to latest stable service api version
- Add relevant fixes to resolve CI issues
- Other clean up and simplify code


specification/storage/Microsoft.QueueStorage/routes.tsp
```diff
@@ -182,15 +139,18 @@ namespace Storage.Queues {
       /** Include this parameter to specify that the queue's metadata be returned as part of the response body. */
       @query include?: Array<ListQueuesIncludeType>;
     },
-    ListQueuesResponse
+    {
+      ...DateResponseHeader;
```
jalauzon-msft: For all the places that the date header was added to the response, we should do something to make them internal (they were intentionally left out before), so they don't show in generated code. Basically, by adding date here, we are creating a whole response model type thing in the Rust SDK that we don't want (can show you more if you want). The other thing to note is that date is on every response and therefore could go in the template.

For Blob, we removed each individual instance of the date response header and added it to the response templates and marked it as internal. We should do that here as well. See Blob [model](https://github.com/Azure/azure-rest-api-specs/blob/feature/blob-tsp-rust/specification/storage/Microsoft.BlobStorage/models.tsp#L2026) and [template](https://github.com/Azure/azure-rest-api-specs/blob/feature/blob-tsp-rust/specification/storage/Microsoft.BlobStorage/routes.tsp#L42).

jalauzon-msft (approved):

jalauzon-msft (approved):

PR #40065, opened 2026-01-31
[Foundry branch-to-branch] Add a service contract library target to facilitate ingestion

trrwilson:
`Azure.AI.Agents.Contracts.V2` is a specification adapter library used by the Foundry Agents Service to process payloads. In addition to the Foundry-extended OpenAI API public surface, a substantial amount of additional service-to-service metadata is encoded into the contract payload.

Previously, this library used an entirely unconfigured "blank" copy of legacy `@azure-tools/typespec-csharp` emission in a manual, copy/paste fashion (with a lot of complex workaround code). We want to better support this service contract library as a first-class consumer of the specification sources and this PR extends the "library target" approach applied for SDK client libraries (azure-ai-agents in Java, Azure.AI.Projects.OpenAI in .NET, corresponding complements of azure-ai-projects and Azure.AI.Projects for those languages together with the Python/JS unified Projects library) to also treat `Azure.AI.Contracts.V2` as another "target."

- Common library imports are centralized the extant `tspconfig.yaml` files patterned for compilation targets; independent of everything else, this simplifies boilerplate per-.tsp-file preambles (that also included a lot of spurious/extraneous imports)
- Via said centralization, individual targets (here, the new `sdk-service-agents-contracts`) can use a parallel "view" of the openai-typespec files as needed
- The new target performs a number of decorator-fueled customizations to existing models; this greatly streamlines ingestion and maintenance


specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
```diff
@@ -6,8 +6,8 @@ emit:
 options:
   "@typespec/openapi3":
     emitter-output-dir: "{project-root}/"
-    output-file: "openapi3/{version}/microsoft-foundry-openapi3.json"
-    # output-file: "openapi3/{version}/microsoft-foundry-openapi3.yaml"
+    # output-file: "openapi3/{version}/microsoft-foundry-openapi3.json"
+    output-file: "openapi3/{version}/microsoft-foundry-openapi3.yaml"
```
glecaros: I want to ask for the ability to have an array of file formats as options, the fact that we have to do this is somewhat annoying.

PR #40089, opened 2026-02-02
feat: Restructure the TypeSpec to flatten client nesting

lsiddiquee:
# Restructure the TypeSpec to flatten client nesting

This PR restructures the Queue Storage TypeSpec to flatten the client nesting structure. The changes simplify the client architecture by reorganizing routes and client definitions, reducing unnecessary nesting levels and improving the overall API design.

Changes:

- Refactored routes.tsp to flatten operation structure
- Updated client.tsp with simplified client definitions
- Added placeholder example file


vincenttran-msft (dismissed):
LGTM, will leave Jacob for final approval. Thanks again for your contributions Likhan 😄

specification/storage/Microsoft.QueueStorage/client.tsp
```diff
@@ -7,6 +7,3 @@ namespace Customizations;
 
 @@clientNamespace(Storage.Queues, "Azure.Storage.Queues");
 @@clientNamespace(Storage.Queues.Queue, "Azure.Storage.Queues");
-
-@@clientName(Storage.Queues.Service, "QueueServiceClient", "rust");
-@@clientName(Storage.Queues.Queue, "QueueClient", "rust");
```
vincenttran-msft: I think we should revert to this, since these currently are scoped to Rust-only changes, while directly being in `routes.tsp` is abit too wide of scope.

lsiddiquee:
@vincenttran-msft and @jalauzon-msft I have rebased the branch with the latest changes and then my changes applied on top. However, not really happy with removing the queue_name from the route and relying on the url. If we remove that then not only the generated client does not have that in the param, which is confusing, but if you see the open api spec that needs to be updated looks very wrong as well. Let me know your thoughts, I would prefer to keep the param here, and then in the wrapper fix it accordingly. We will still have support url and optional queue name of the constructor, and then the constructor if needed extracts the queue_name from the url if not provided and uses it to construct the underlying generated client. Let me know if it makes sense.

https://github.com/Azure/azure-rest-api-specs/actions/runs/21746292229/job/62732860821?pr=40089#step:4:462

jalauzon-msft:
Hi @lsiddiquee, the design in Blob is likely how we want it to be for Queue for consistency and as that is the only way we got everything to work. The thing to keep in mind is the end goal is to completely remove (most of) the handwritten layer and so we can't really handle anything special in the handwritten layer as we will soon be removing it. We are about to merge the PR to remove it in Blob.

Agree that the generated Swagger looks a bit odd, but it may be something we just have to deal with. This is the first I've seen it and we may end discussing this point with Azure SDK to see what they think but for now, we should probably just go with it to keep things moving.

As an aside, ands to better facilitate seeing the generated Swagger, please run `tsp compile` locally and include the results in the PR. Not sure if you've run it before (I didn't know about it until recently) but `tsp compile` just auto-generates a Swagger for us to check in alongside the TypeSpec. I do not see the Swagger file on the PR and that will also help CI.

jalauzon-msft:
Also, if you wouldn't mind re-generating the Rust code and pushing a new iteration to the code PR just to make sure everything looks good, that would be great.

vincenttran-msft:
To close the loop here: I pulled forward your work @lsiddiquee and incorporated the remaining feedback + did some other cleanup on the TypeSpec. This should be enough to get this initial PR work across the finish line, and then I will remove the handwritten layer + other stuff in a follow-up PR once that pattern drops in Blobs. Thanks again for all your hard work and contributions Likhan, it has been a great help!!
https://github.com/Azure/azure-sdk-for-rust/pull/3698
`.tsp`: https://github.com/Azure/azure-rest-api-specs/pull/40244

timotheeguerin:

PR #40127, opened 2026-02-04
Plan on Improving AKS TypeSpec definitions

FumingZhang:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Other, please clarify:
    - improve AKS TypeSpec definitions

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
@@ -1,0 +1,609 @@
+# AKS TypeSpec Improvement Plan
+
+**Date:** February 3, 2026  
+**Scope:** specification/containerservice/resource-manager/Microsoft.ContainerService/aks  
+**Status:** Planning Phase
+
+## Executive Summary
+
+This document outlines a comprehensive plan to improve the TypeSpec definitions for Azure Kubernetes Service (AKS). The primary goals are to:
+- Improve file organization and maintainability
+- Resolve 200+ FIXME comments
+- Better co-locate models with their operations
+- Ensure ARM compliance
+- Improve documentation quality
+
+
+## Issues Identified
+
+### 1. File Organization - Separation of Models and Operations
+
+**Current State:**
+- **Models.tsp** - Contains all shared model definitions (~7,144 lines)
+- **Individual resource files** - Each contains both the resource model and operations (e.g., AgentPool.tsp, ManagedCluster.tsp)
+- **Routes.tsp** - Contains standalone operation groups
+
+**Problems:**
+- The massive `Models.tsp` file (7,144 lines) is difficult to navigate and maintain
+- Models defined in `Models.tsp` are separated from the operations that use them
+- Hard to understand the relationship between operations and their associated models
+
+### 2. Excessive FIXME Comments
+
+Found **200+ FIXME comments** across the codebase, categorized as:
+
+#### A. Documentation Issues (~75 instances)
+- Missing documentation for union/enum members
+- Affected files: Models.tsp (lines 76, 429-437, 672-682, 893, 899, 1107-1109, etc.)
+- Impact: Reduces API discoverability and usability
+
+#### B. Casing Style Issues (~90 instances)
+- Properties with non-standard casing (camelCase violations)
+- Affected files: Models.tsp (lines 2083, 2090, 2144, 2191, 2257, 2345, etc.)
+- Impact: Inconsistent naming conventions
+
+#### C. ARM Resource Violations (~20 instances)
+- Invalid envelope properties (sku, extendedLocation, identity, kind)
+- arm-no-record violations
+- arm-resource-provisioning-state violations
+- Affected files: ManagedCluster.tsp (lines 28, 34, 40, 46), Machine.tsp (line 28)
+- Impact: ARM compliance issues
+
+#### D. Operation Response Code Issues (~10 instances)
+- arm-post-operation-response-codes violations
+- Affected files: ManagedCluster.tsp (lines 248, 262, 277, 290, 304), AgentPool.tsp (line 85)
+- Impact: Non-standard response patterns
+
+#### E. Other Issues (~10 instances)
+- Secret property violations (Models.tsp lines 3392, 3445, 4088)
+- Composition over inheritance warnings (Models.tsp lines 2312, 3475, 5351)
+- utcDateTime verification needed (Models.tsp lines 4136, 5463)
+- no-enum rule violations (Models.tsp line 1783)
+
+
+## Detailed Solution Plan
+
+### Phase 1: Fix Documentation Issues (75+ instances)
+
+**Goal:** Add missing documentation for all types and members
+
+**Priority: High** - Lowest risk, highest immediate value, no breaking changes
+
+**Action Items:**
+
+1. **Add missing documentation for union members**
+   
+   **Before:**
+   ```typespec
+   union ExtendedLocationTypes {
+     string,
+     #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification..."
+     EdgeZone: "EdgeZone",
+   }
+   ```
+   
+   **After:**
+   ```typespec
+   union ExtendedLocationTypes {
+     string,
+     /** Edge Zone extended location type */
+     EdgeZone: "EdgeZone",
+   }
+   ```
+
+2. **Document all enum values systematically**
+   - Review original swagger definitions for descriptions
+   - Add meaningful descriptions for each value
+   - Focus on user-facing documentation
+
+**Affected Files:**
+- Models.tsp (lines 76, 429-437, 672-682, 893, 899, 1107-1109, 1342-1348, 1372-1384, 1425-1435, 1540-1548, 1557-1563, 1590-1598, 1898, 2009, 3156, 3559, 3564, 4498, 4699, 4748, 4756, 5145, 5293, 5329)
+
+**Estimated work:** ~75 documentation additions
+
+
+### Phase 2: Reorganize File Structure
+
+**Goal:** Break down Models.tsp by moving models to their corresponding operation .tsp files
+
+**Steps:**
+
+1. **Move Resource-Specific Models to Operation Files**
+   - Models tightly coupled to a specific resource should be moved to that resource's file
+   - Example: `ManagedClusterAgentPoolProfileProperties` → AgentPool.tsp
+   - Example: `ManagedClusterProperties` → ManagedCluster.tsp
+   - Example: `SnapshotProperties` → Snapshot.tsp
+   - Example: Network-related models for LoadBalancer → LoadBalancer.tsp
+
+2. **Create CommonModels.tsp for Shared Models**
+   
+   Only models referenced by **multiple** operation .tsp files should go here:
+   
+   - **`CommonModels.tsp`** - Shared base models and types
+     - Common enums/unions used across multiple resources
+     - Base property types referenced by multiple resources
+     - Shared response/error types
+
+3. **Update Imports**
+   - Update main.tsp to import the new organized files
+   - Ensure circular dependency issues are avoided
+   - Each operation .tsp imports CommonModels.tsp as needed
+
+**Benefits:**
+- Co-location of related definitions (models next to their operations)
+- Reduced Models.tsp file size
+- Easier navigation - find a model in the same file as its operation
+- Simpler structure with only one shared file
+
+**Files to Create:**
+```
+aks/
+├── CommonModels.tsp           (NEW - shared models only)
+├── Models.tsp                 (REMOVE or significantly reduced)
+├── AgentPool.tsp              (MODIFIED - add AgentPool-specific models)
+├── ManagedCluster.tsp         (MODIFIED - add ManagedCluster-specific models)
+├── Snapshot.tsp               (MODIFIED - add Snapshot-specific models)
+├── LoadBalancer.tsp           (MODIFIED - add LoadBalancer-specific models)
+└── ... (other operation files similarly modified)
+```
+
+
+### Phase 3: Fix Remaining FIXME Comments (Grouped by Type)
+
+#### 3A. Update Casing Style Suppression Justifications (90+ instances)
+
+**Priority: Low** - These are cosmetic and cannot be fixed without breaking changes
+
+**Reality Check:**
+- Property names like `enableVnetIntegration` determine the **JSON wire format**
+- Changing them would be a **breaking change** for all API consumers
+- `@clientName` only affects SDK naming, NOT the wire format - linter still complains
+- **Only viable option:** Keep suppressions with proper justification
+
+**Action:** Update all FIXME suppressions with backward-compatibility justification
+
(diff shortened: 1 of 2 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+### Phase 1: Reorganize File Structure (High Priority)
+
+**Goal:** Co-locate models with their operations for better maintainability
+
+**Steps:**
+
+1. **Keep Resource-Specific Models with Operations**
+   - Models tightly coupled to a specific resource should remain in that resource's file
+   - Example: `ManagedClusterAgentPoolProfileProperties` should stay in AgentPool.tsp
+   - Example: `ManagedClusterProperties` should stay in ManagedCluster.tsp
```
matthchr: How deeply does this apply? For example you talk about networking models having their own file `networking-models.tsp`, but a lot of the networking models are only used in ManagedCluster.

Same with enums, many enums are only used in one context so may make sense to stay with their resource rather than being in their own file.

TBH I think from a searchability standpoint, it might be cleanest if we could just "put things with their resource" and be done with it.

The one problem is that AP types are used in MC because MC embeds AP. I _think_ we can probably make that work by just having MC import the AP file though and then reference the models it needs.

So I guess what I would advocate for as a first shot, and then we can see if we need to do more is:
- `common-models.tsp` for things that are common across all/many resources.
- All remaining models go with the resource that uses them and we don't try to break things out further, and we see how big the files are.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+
+**Files to Create:**
+```
+aks/
+├── common-models.tsp          (NEW)
+├── enums.tsp                  (NEW)
+├── networking-models.tsp      (NEW)
+├── security-models.tsp        (NEW)
+├── upgrade-models.tsp         (NEW)
+├── models.tsp                 (MODIFIED - reduced size)
```
matthchr: Shouldn't/can't this file completely go away?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+└── ... (existing files)
+```
+
+---
+
+### Phase 2: Fix FIXME Comments (Grouped by Type)
```
matthchr: We should probably split this into multiple PRs when we actually do this.

Part 1: Fixing missing docs (merge this before part 2)
Part 2, moving things around. Result should be exactly the same generated Swagger as before (that's how we know we moved it OK).
Part 3..N: Fixing FIXMEs (this will change generated Swagger too) -- one PR per FIXME type?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+
+**Decision Points:**
+- These suppressions exist because the original OpenAPI had non-standard casing
+- Options:
+  1. **Keep suppressions** with proper justification (for backward compatibility)
+  2. **Fix casing** and document as breaking change
+  3. **Use @clientName** decorator to maintain wire format while fixing SDK names
+
+**Recommended Approach:**
+Use `@clientName` decorator to maintain backward compatibility:
```
matthchr: Isn't this still breaking for the SDKs? Do the SDKs care - I would imagine they do?

Want to confirm the SDKs are OK with breaking changes before we do this?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
matthchr: Above is basically what Fleet does btw (they don't try to split types out into other files)

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+
+**Issue:** sku, extendedLocation, identity, kind in resource envelope  
+**Files:** ManagedCluster.tsp (lines 28, 34, 40, 46), Machine.tsp (line 28)
+
+**Action:** Update suppressions with proper justification:
+
+**Before:**
+```typespec
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "FIXME: Update justification..."
+sku?: ManagedClusterSKU;
```
matthchr, line 212: SKU is AFAIK a supported top-level property. May be worth looking into why they're flagging it?

See: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v6/types.json#L366

Is the issue here we're just not using the standard definition of SKU?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+  "SKU is a standard ARM envelope property for ManagedCluster and must remain in the envelope for backward compatibility with existing API versions"
+sku?: ManagedClusterSKU;
+```
+
+**Subcategory C2: arm-no-record Violations**
+
+**Issue:** Using Record<> type inappropriately  
+**Files:** models.tsp (lines 1890, 1999, 2005, 2118, 2201, 2528, 2534, 3273, 3280, 3461, 3625, 5407)
+
+**Action:** Replace with properly typed models or add justification
```
matthchr: I think most (all?) instances will require a justification added.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+**Issue:** Provisioning state not following ARM patterns  
+**Files:** models.tsp (lines 2031, 2480, 5246, 5414)
+
+**Action:** Review and align with ARM standards or justify deviation
+
+**Subcategory C4: Other ARM Violations**
+
+- arm-resource-duplicate-property (models.tsp line 2435)
+- no-resource-delete-operation (Machine.tsp line 33)
+- patch-envelope (models.tsp line 5402)
```
matthchr: We should suppress this patch-envelope thing still.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+  // ... 
+}
+
+interface AgentPools {
+  // ... operations
+}
+```
+
+**Apply to all resource files:**
+- ManagedCluster.tsp
```
matthchr, line 407: Another minor fixup, we should move all of our files to consistent naming pattern.

`managed-cluster-snapshot.tsp` or `managed_cluster_snapshot.tsp` I don't really care which as long as we're consistent.

matthchr: (Fleet uses `managedclustersnapshot.tsp` which I like slightly less but would also be OK with if we want to be consistent with them).

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+## Execution Strategy
+
+### Step-by-Step Approach
+
+#### 1. Create Backup Branch
+```bash
+git checkout -b feature/aks-typespec-improvements
+```
+
+#### 2. Start with Documentation (Lowest risk)
```
matthchr: Step by step approach does not match phases below in terms of ordering

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,615 @@
...
+- Validate compilation still works
+- Commit: "docs: Add missing documentation for AKS TypeSpec definitions"
+
+#### 3. Address ARM Violations
+- Review and properly justify suppressions
+- Engage with service team if clarification needed
+- Commit: "fix: Update ARM violation suppressions with proper justification"
+
+#### 4. Reorganize Files (If approved)
+- Create new organized files
```
matthchr: May be worth calling out for this step that the expectation is after every move the swagger is exactly the same as it was before the move. There should be no changes to the generated Swagger due to moves.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: updated to doc to follow this pattern

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: the new goal is set to remove the models.tsp

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: If we want to address this issue, the main concern may not be from the SDK side, since the change would result in a breaking change at the API specification level, which we likely want to avoid. Therefore, the revised proposal is to provide clear justifications for the violations in order to resolve the FIXMEs.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: agreed, update the detailed plan to follow this

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: yes, otherwise we would make breaking change

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: Some duplicate content has been removed, and the current execution strategy is consistent with the order in the detailed plan.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: yes, I've set 2 basic principles for the change
1. **No Swagger Changes for Refactoring** 
2. **No Breaking Changes**

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: Since most of the files are named using PascalCase, I've updated the samples in this document to match that pattern. If you prefer using all lowercase instead, I can update the pattern accordingly.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: Yes, for most ARM violations, I think the best approach right now is to suppress them.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: I believe the warning is triggered because the linter does not consider these common type-allowed fields. It treats any field other than "name" and those without a key tag as a rule violation (see [code](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/rules/arm-resource-invalid-envelope-property.ts#L19)).

Additionally, the [common types in TypeSpec](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L66) do provide this model definition, but we are not extending managed Cluster based on it. This is because doing so would require us to use the definitions of sku and identity from common-types instead of our customized versions, even though they might be identical. I suggest we provide a detailed explanation now and revisit this later to see if we can extend this model and resolve the warning.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
matthchr: I think _just_ using `clientName` is not breaking at the API specification level, right?

It could just be breaking at the SDK level.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
matthchr: Yep, agree with just including basically what you said here as the explanation.

The net is we're not _really_ in violation of the resource shape, just maybe how TypeSpec wants us to use the common one.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
matthchr: I mostly want us to just be consistent, which their generator tool was not. I'm OK with PascalCase too.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IMPROVEMENT_PLAN.md
FumingZhang: Yes, the `clientname` option only affects how the SDK generator works. The tests done during this conversion used the old generator, which relied on Swagger as input. Before June, we should also verify if the SDK generated directly from TypeSpec still meets our requirements. If any issues come up, we can use `clientname` to address them.

FumingZhang:
achieved by #40144, #40188 #40221, #40500

PR #40137, opened 2026-02-04
fix(containerregistry): attribute RegistryProperties moved to stable in API 2025-11-01

CodeJourneys:
This PR fixes a bug in the 2025-11-01 API Typespec definition where networkRuleBypassAllowedForTasks was not annotated as introduced in that API


northtyphoon (approved):
LGTM

PR #40154, opened 2026-02-04
Fix some issues in command tsmv

pshao25:
# Choose a PR Template

### DIFF
--- a/eng/tools/typespec-migration-validation/src/compare.ts
+++ b/eng/tools/typespec-migration-validation/src/compare.ts
@@ -1,5 +1,6 @@
-import { HttpMethod, OpenAPI2Document, OpenAPI2Operation, OpenAPI2Response, OpenAPI2Schema, OpenAPI2SchemaProperty, Ref, Refable } from "@azure-tools/typespec-autorest";
+import { HttpMethod, OpenAPI2Document, OpenAPI2Operation, OpenAPI2Parameter, OpenAPI2Response, OpenAPI2Schema, OpenAPI2SchemaProperty, Ref, Refable } from "@azure-tools/typespec-autorest";
 import { getCommonTypeDefinition } from "./commonType.js";
+import { getOriginalParameter } from "./parameter.js";
 
 interface Diff {
   before: any;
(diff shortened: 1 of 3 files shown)


eng/tools/typespec-migration-validation/src/compare.ts
```diff
@@ -256,14 +295,107 @@ function compareOperation(oldOperation: OpenAPI2Operation, newOperation: OpenAPI
...
+        level: "warning"
+      });
+    }
+  }
+
+  return pathDiffs;
+}
+
+function getBodyParameter(operation: OpenAPI2Operation, document: OpenAPI2Document): OpenAPI2Parameter | undefined {
+  for (const param of operation.parameters ?? []) {
```
live1206, line 386: will operation.parameters ever be null? If so, in L287, we have if (oldOperation.parameters.length !== newOperation.parameters.length), will this throw null reference exception?

eng/tools/typespec-migration-validation/src/compare.ts
```diff
@@ -256,14 +295,107 @@ function compareOperation(oldOperation: OpenAPI2Operation, newOperation: OpenAPI
...
+      type: "parameter",
+      parameterName: "body",
+      changeType: "presence",
+      level: "error"
+    });
+    return pathDiffs;
+  }
+
+  // If neither has a body parameter, no diff
+  if (!oldBodyParam || !newBodyParam) {
```
live1206, line 329: when one body has a schema ref and the other has inline schema (or one is missing schema entirely), should this be considered as diff?

eng/tools/typespec-migration-validation/src/parameter.ts
```diff
@@ -102,10 +102,12 @@ export function isResourceGroupNameParameter(obj: Record<string, any>) {
...
-    const parameterName = refPath.substring("#/parameters/".length);
+export function getOriginalParameter(refPath: string, originalDocument?: OpenAPI2Document): OpenAPI2Parameter | undefined {
+  originalDocument ??= getOriginalDocument();
+  // get the part after '#/parameters/'
+  if (refPath.includes("#/parameters/")) {
+    const parameterName = refPath.substring(refPath.indexOf("#/parameters/") + "#/parameters/".length);
     return originalDocument?.parameters?.[parameterName];
-  } else return undefined;
+  }
+  else return undefined;
```
live1206: consider to wrap else to be format consistent

eng/tools/typespec-migration-validation/src/compare.ts
pshao25: Yes. It is checked in L315.

eng/tools/typespec-migration-validation/src/compare.ts
pshao25: Probably

live1206 (approved):

PR #40219, opened 2026-02-06
Improving AKS TypeSpec definitions - resolve FIXME comments

FumingZhang:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Other, please clarify:
    - Improving AKS TypeSpec definitions - resolve FIXME comments

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
@@ -82,15 +82,14 @@ interface AgentPools {
   /**
    * Aborts the currently running operation on the agent pool. The Agent Pool will be moved to a Canceling state and eventually to a Canceled state when cancellation finishes. If the operation completes before cancellation can take place, a 409 error code is returned.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "This operation returns 202 for async operations, which is the correct behavior for this long-running action."
   @action("abort")
   @summary("Aborts last operation running on agent pool.")
-  abortLatestOperation is ArmResourceActionAsync<
+  abortLatestOperation is ArmResourceActionAsyncBase<
     AgentPool,
     void,
-    NoContentResponse,
-    LroHeaders = ArmCombinedLroHeaders<FinalResult = void> &
-      Azure.Core.Foundations.RetryAfterHeader
+    ArmAcceptedLroResponse<LroHeaders = ArmCombinedLroHeaders<FinalResult = void> &
+      Azure.Core.Foundations.RetryAfterHeader>,
+    BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<AgentPool>
   >;
 
   /**
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
@@ -48,9 +48,7 @@ interface AgentPools {
        */
       @header
       ifNoneMatch?: string;
-    },
-    LroHeaders = ArmLroLocationHeader<FinalResult = AgentPool> &
-      Azure.Core.Foundations.RetryAfterHeader
+    }
   >;
 
   /**
@@ -58,6 +56,8 @@ interface AgentPools {
    */
   delete is ArmResourceDeleteWithoutOkAsync<
     AgentPool,
+    LroHeaders = ArmAsyncOperationHeaderWithLocation &
+      Azure.Core.Foundations.RetryAfterHeader,
     Parameters = {
       /**
        * ignore-pod-disruption-budget=true to delete those pods on a node without considering Pod Disruption Budget
@@ -111,7 +111,8 @@ interface AgentPools {
   deleteMachines is ArmResourceActionAsyncBase<
     AgentPool,
     AgentPoolDeleteMachinesParameter,
-    ArmAcceptedLroResponse,
+    ArmAcceptedLroResponse<LroHeaders = ArmAsyncOperationHeaderWithLocation &
+      Azure.Core.Foundations.RetryAfterHeader>,
     BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<AgentPool>
   >;
 
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
@@ -62,7 +62,7 @@ interface AgentPools {
       /**
        * ignore-pod-disruption-budget=true to delete those pods on a node without considering Pod Disruption Budget
        */
-      #suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+      #suppress "@azure-tools/typespec-azure-core/casing-style" "Property name maintained for backward compatibility with existing API versions"
       @query("ignore-pod-disruption-budget")
       `ignore-pod-disruption-budget`?: boolean;
 
@@ -82,7 +82,7 @@ interface AgentPools {
   /**
    * Aborts the currently running operation on the agent pool. The Agent Pool will be moved to a Canceling state and eventually to a Canceled state when cancellation finishes. If the operation completes before cancellation can take place, a 409 error code is returned.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "This operation returns 202 for async operations, which is the correct behavior for this long-running action."
   @action("abort")
   @summary("Aborts last operation running on agent pool.")
   abortLatestOperation is ArmResourceActionAsync<
@@ -119,7 +119,7 @@ interface AgentPools {
   /**
    * Upgrading the node image version of an agent pool applies the newest OS and runtime updates to the nodes. AKS provides one new image per week with the latest updates. For more details on node image versions, see: https://docs.microsoft.com/azure/aks/node-image-upgrade
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "LRO location header pattern is consistent with existing AKS API behavior for backward compatibility."
   @summary("Upgrades the node image version of an agent pool to the latest.")
   upgradeNodeImageVersion is ArmResourceActionAsyncBase<
     AgentPool,
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
@@ -118,14 +118,13 @@ interface AgentPools {
   /**
    * Upgrading the node image version of an agent pool applies the newest OS and runtime updates to the nodes. AKS provides one new image per week with the latest updates. For more details on node image versions, see: https://docs.microsoft.com/azure/aks/node-image-upgrade
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "LRO location header pattern is consistent with existing AKS API behavior for backward compatibility."
   @summary("Upgrades the node image version of an agent pool to the latest.")
   upgradeNodeImageVersion is ArmResourceActionAsyncBase<
     AgentPool,
     void,
     {
       ...AcceptedResponse;
-      ...ArmAsyncOperationHeader<FinalResult = void>;
+      ...ArmCombinedLroHeaders<FinalResult = void>;
       ...Azure.Core.Foundations.RetryAfterHeader;
 
       @body
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -2918,10 +2918,8 @@ model PrivateLinkResource {
   id?: string;
 
   /**
-   * The name of the private link resource.
+   * The name of the private link resource. See [naming rules](https://aka.ms/search-naming-rules) for more details.
    */
-  #suppress "@azure-tools/typespec-azure-core/no-openapi" "OpenAPI extension is required for specific API behavior that cannot be expressed in TypeSpec."
-  @OpenAPI.externalDocs("https://aka.ms/search-naming-rules")
   name?: string;
 
   /**
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -16,6 +16,32 @@ using Versioning;
 namespace Microsoft.ContainerService;
 
 /**
+ * Provides both Azure-AsyncOperation and Location headers with
+ * final-state-via set to azure-async-operation.
+ * Unlike ArmCombinedLroHeaders which places @finalLocation on Location
+ * (resulting in final-state-via: location), this model places @finalLocation
+ * on Azure-AsyncOperation so the emitter outputs final-state-via: azure-async-operation.
+ */
+model ArmAsyncOperationHeaderWithLocation<
+  StatusMonitor extends {} = ArmOperationStatus,
+  FinalResult extends {} | void = never,
+  PollingUrlValue extends string = ResourceLocation<StatusMonitor>,
+  FinalUrlValue extends string = string
+> {
+  /** A link to the status monitor */
+  @Azure.Core.pollingLocation(
+    Azure.Core.StatusMonitorPollingOptions<StatusMonitor>
+  )
+  @Azure.Core.finalLocation(FinalResult)
+  @header("Azure-AsyncOperation")
+  azureAsyncOperation?: PollingUrlValue;
+
+  /** The Location header contains the URL where the status of the long running operation can be checked. */
+  @header("Location")
+  location?: FinalUrlValue;
+}
+
+/**
  * Different support tiers for AKS managed clusters
  */
 union KubernetesSupportPlan {
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -1404,22 +1404,23 @@ union SnapshotType {
 /**
  * The type of identity used for the managed cluster. For more information see [use managed identities in AKS](https://docs.microsoft.com/azure/aks/use-managed-identity).
  */
-#suppress "@azure-tools/typespec-azure-core/no-enum" "Enum is used here for backward compatibility with the existing API contract."
-enum ResourceIdentityType {
+union ResourceIdentityType {
+  string,
+
   /**
    * Use an implicitly created system assigned managed identity to manage cluster resources. Master components in the control plane such as kube-controller-manager will use the system assigned managed identity to manipulate Azure resources.
    */
-  SystemAssigned,
+  SystemAssigned: "SystemAssigned",
 
   /**
    * Use a user-specified identity to manage cluster resources. Master components in the control plane such as kube-controller-manager will use the specified user assigned managed identity to manipulate Azure resources.
    */
-  UserAssigned,
+  UserAssigned: "UserAssigned",
 
   /**
    * Do not use a managed identity for the Managed Cluster, service principal will be used instead.
    */
-  None,
+  None: "None",
 }
 
 /**
@@ -1968,7 +1969,7 @@ model ManagedClusterWindowsProfile {
    */
   adminUsername: string;
 
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
+  @secret
   @doc("Specifies the password of the administrator account. <br><br> **Minimum-length:** 8 characters <br><br> **Max-length:** 123 characters <br><br> **Complexity requirements:** 3 out of 4 conditions below need to be fulfilled <br> Has lower characters <br>Has upper characters <br> Has a digit <br> Has a special character (Regex match [\\W_]) <br><br> **Disallowed values:** \"abc@123\", \"P@$$w0rd\", \"P@ssw0rd\", \"P@ssword123\", \"Pa$$word\", \"pass@word1\", \"Password!\", \"Password1\", \"Password22\", \"iloveyou!\"")
   adminPassword?: string;
 
@@ -2021,7 +2022,7 @@ model ManagedClusterServicePrincipalProfile {
   /**
    * The secret password associated with the service principal in plain text.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
+  @secret
   secret?: string;
 }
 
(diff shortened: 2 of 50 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
```diff
@@ -82,7 +82,7 @@ interface AgentPools {
   /**
    * Aborts the currently running operation on the agent pool. The Agent Pool will be moved to a Canceling state and eventually to a Canceled state when cancellation finishes. If the operation completes before cancellation can take place, a 409 error code is returned.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "This operation returns 202 for async operations, which is the correct behavior for this long-running action."
```
matthchr: I think the complaint is that we are returning `NoContentResponse`, which is HTTP 204. The TypeSpec spec wants this to be 202 as per https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/post-operation-response-codes/#asynchronous (at least I think I am reading that right?), although I am a bit confused because the actual [REST API specification](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/async-api-reference.md#call-action-post-asynchronously) says 204, which is what we are doing.

matthchr: It would be good to confirm with ARM/TypeSpec reviewers what exactly the issue is here.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
```diff
@@ -119,7 +119,7 @@ interface AgentPools {
   /**
    * Upgrading the node image version of an agent pool applies the newest OS and runtime updates to the nodes. AKS provides one new image per week with the latest updates. For more details on node image versions, see: https://docs.microsoft.com/azure/aks/node-image-upgrade
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/lro-location-header" "LRO location header pattern is consistent with existing AKS API behavior for backward compatibility."
```
matthchr: Paul had documented this here, I think:
https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/392723/AKS-Swagger-Gotchas

Specifically:
> All our async APIs return the Azure-AsyncOperation header. But anything that returns202/accepted also returns a Location header. This is mostly POSTs and DELETES (but one GET?).

I think we should double-check the service code. The correct fix to these APIs may be to document that we _do_ return a Location header, we just historically never put it in the Swagger.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -1401,7 +1401,7 @@ union SnapshotType {
 /**
  * The type of identity used for the managed cluster. For more information see [use managed identities in AKS](https://docs.microsoft.com/azure/aks/use-managed-identity).
  */
-#suppress "@azure-tools/typespec-azure-core/no-enum" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Enum is used here for backward compatibility with the existing API contract."
```
matthchr: I suspect that enum rather than extensible enum here was an oversight rather than something intentional.

Is it safe to change? If so we could just change it...

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -2983,7 +2983,7 @@ model ManagedClusterWindowsProfile {
    */
   adminUsername: string;
 
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
```
matthchr: I think this is wrong - we can/should check the behavior of the actual API (or the service code), but I am 95% sure that we do _not_ return this field on cluster GET. That means that we have implemented this property as a secret, and we can just annotate it as such in TypeSpec as well.

The fact we didn't before is likely just a Swagger oversight - it doesn't reflect the actual behavior of the service AFAIK.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -3036,7 +3036,7 @@ model ManagedClusterServicePrincipalProfile {
   /**
    * The secret password associated with the service principal in plain text.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
```
matthchr: Same comment - 95% sure we do not return on GET, I think we can probably mark this as a secret unless I am misunderstanding what "marking as secret" means

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -3065,8 +3065,8 @@ model ManagedClusterAddonProfile {
 /**
  * Information of user assigned identity used by this add-on.
  */
-#suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-resource-manager/no-empty-model" "Empty model is intentional for extensibility and matches existing API structure."
```
matthchr, line 2080: Same comment as other spot

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -3915,7 +3914,7 @@ model PrivateLinkResource {
   /**
    * The name of the private link resource.
    */
-  #suppress "@azure-tools/typespec-azure-core/no-openapi" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "OpenAPI extension is required for specific API behavior that cannot be expressed in TypeSpec."
   @OpenAPI.externalDocs("https://aka.ms/search-naming-rules")
```
matthchr: Why not just put this in the description string like we do for most other aka.ms links?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -4924,12 +4923,12 @@ model CredentialResult {
 /**
  * Tags object for patch operations.
  */
-#suppress "@azure-tools/typespec-azure-resource-manager/patch-envelope" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-resource-manager/patch-envelope" "TagsObject is a standard ARM pattern for PATCH operations to update resource tags independently."
```
matthchr: I think their linter is wrong, can we raise with the TypeSpec linter folks?

The [spec says](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/patch-resource.md#patch-resource):
> A Patch operation updates a resource and it is strongly recommended that service teams implement it. ARM requires RPs to support Patch for updating tags for a tracked resource.

They only require the tags field. Other patchable fields are optional.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/JWTAuthenticator.tsp
```diff
@@ -166,13 +166,13 @@ model JWTAuthenticatorClaimMappingExpression {
...
-#suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-core/casing-style" "Property name maintained for backward compatibility with existing API versions"
 @added(Versions.v2025_10_02_preview)
 model JWTAuthenticatorExtraClaimMappingExpression {
   /**
    * The key of the extra attribute.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
   key: string;
```
matthchr, line 235: I don't think this is a secret - which with Wei Chen on AKS but 99% sure we can document this more clearly as "this value is not a secret", so saying:
> This property may contain credential-like data but is required

Is wrong

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/LoadBalancer.tsp
```diff
@@ -104,7 +104,7 @@ model LabelSelectorRequirement {
   /**
    * key is the label key that the selector applies to.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "This property may contain credential-like data but is required for API functionality and is handled securely."
```
matthchr: Label keys are not secrets. Change the suppression next as it is not true that "This property may contain credential-like data"

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/ManagedCluster.tsp
```diff
@@ -236,25 +234,25 @@ model ManagedCluster
   /**
    * The managed cluster SKU.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "Property intentionally at envelope level for this resource. This is a published API pattern that cannot be changed without breaking existing clients."
```
matthchr: This comment applies to all of these I think:

Saying "This is a published API pattern that cannot be changed without breaking existing clients." is a bit incorrect. Can we say something like: "This field is allowed at the top-level as per the ARM resource specification"?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/ManagedCluster.tsp
```diff
@@ -456,7 +454,7 @@ interface ManagedClusters {
   /**
    * See [Certificate rotation](https://docs.microsoft.com/azure/aks/certificate-rotation) for more details about rotating managed cluster certificates.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "This operation returns 202 for async operations, which is the correct behavior for this long-running action."
```
matthchr: I think same comment as other 202 + POST. I'm pretty sure we DO return location header - can probably check CLI replay tests or just run the cmd locally in CLI with --debug enabled to confirm? (or check service code).

I think we can probably fix this + remove suppression.

matthchr: Applies to all the 202 + post

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
matthchr: He also says:
> Our location url does not ever return 200 or a response body when successful it will instead return 204. But our swagger documents 200 for several post actions and is incorrect. Changing the swagger to 204 is a breaking change because it will alter signature of generated clients. Instead we want to fix the location header to give a 200 and a result body. This would actually fulfill the spec above too.

though. So if we fix this to document the location header we may want to test the ops to make sure they still work in SDK/CLI. I think they will/do since our other ones already do but we should just confirm.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/PrivateEndpointConnection.tsp
```diff
@@ -146,7 +146,7 @@ interface PrivateEndpointConnections {
   /**
    * Deletes a private endpoint connection.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "Delete operation response codes follow existing AKS API contract for backward compatibility."
   delete is ArmResourceDeleteWithoutOkAsync<
```
matthchr: I actually think this is incorrect.

We say:
`ArmResourceDeleteWithoutOkAsync`, but the actual implementation of this API returns 200 not 202 when it's accepted. I think we want the resulting Swagger to documented:
200, 204, which is what it did in 2025-09-01: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/stable/2025-09-01/managedClusters.json#L2328

Now in 2025-10-01, it's saying 200, 202, 204, which doesn't match our impl.

yes this is not in spec with ARM contract but changing it now is breaking AFAIK.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/LocationRoutes.tsp
```diff
@@ -10,7 +10,7 @@ using Versioning;
 
 namespace Microsoft.ContainerService;
 
-#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Interface pattern follows existing AKS operation structure for backward compatibility."
 interface ManagedClustersOperationGroup {
```
matthchr, line 18: isn't this just complaining that we don't set `@armResourceOperations`? I think we can just add?

Their documentation doesn't actually say what this rule is for though so maybe another point to raise with the typespec linter folks

matthchr: Applies below too

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: Yes, I've updated it to add the @secret decorator and removed the suppression.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: updated it to add the @secret decorator and removed the suppression.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: I believe it's safe to change and this the only place where we're using enum.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/JWTAuthenticator.tsp
FumingZhang: updated the clarification, should be a false alert

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/LoadBalancer.tsp
FumingZhang: updated the clarification, should be a false alert

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/ManagedCluster.tsp
FumingZhang: Sure, I've made the updates. However, I've kept the clarification for extendedLocation and zones of Machine in Machine.tsp unchanged, as they're probably not permitted.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
FumingZhang: Understood. I tested it and confirmed that the API returns a 202 response, updated it to remove the NoContentResponse.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
FumingZhang: I tested the API, and it returned a location header that provided the final result with a 204 response. I've updated the typespec to reflect this.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/ManagedCluster.tsp
FumingZhang: yes, updated all 202 + post

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/LocationRoutes.tsp
FumingZhang: I attempted to add the decorator, but this would lead to another tag, `ManagedClustersOperationGroup`, being added for the operation, which seems a bit unusual. During the conversion, there are helper interfaces defined without the decorator, and they seem expected. Instead of adding the decorator, I'm revising the clarification to make it clearer.

- interface AgentPoolsOps
- interface ManagedClusterOps

Image

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: updated all 202 post

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: good catch, updated

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/PrivateEndpointConnection.tsp
FumingZhang: fixed the TypeSpec

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
FumingZhang: created issue [#3939](https://github.com/Azure/typespec-azure/issues/3939)

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
matthchr: I looked at the actual implementation of this API in AKS RP and there are codepaths where we return 204, in particular if the AP was cancelled two times in parallel the second one will return 204 indicating the operation was completed synchronously.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
matthchr: I see that when we set the Combined header type, it prefers the `final-state-via: Location` rather than `azure-async-operation` like we had before.

Is there a way to get final-state-via to be what we had before? I'm worried about changing the URL all of the clients poll after a long time. It's fine to document we support the location header but AFAIK we should _prefer_ the other header?

See https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/async-api-reference.md#asynchronous-operations

> The Azure-AsyncOperation header must be supported for all asynchronous operations for greenfield services starting January 2025. The usage of this header is strongly recommended for brownfield services.

matthchr: This applies to all of the places where we updated final-state-via

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/PrivateEndpointConnection.tsp
matthchr: I don't think this is fixed/correct?

The API returns 200 | 204. What we've got now shows 202 + 204. See my link above for the 2025-09-01 Swagger, I think it needs to match that.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
FumingZhang: Resolved by maintaining all `final-state-via` settings as previously specified in the earlier API version. For those not specified, they are now set to `azure-async-operation`.

For put, the lro header from the template is used, which only includes `azure-async-operation`, and the final state is via it. This keeps the SDK behavior consistent, as it will prompt the SDK to perform a `final-state-via` the initial URL, similar to not specifying the final state via option.

For delete, post, and patch, a custom LRO header was created to return both `azure-async-operation` and `location` header, but the final state via is set to `azure-async-operation`. For delete and post, this maintains the SDK behavior by instructing the SDK to skip the final get, as with not specifying the final state via option. For patch, it keeps the SDK behavior by prompting a `final-state-via` the initial URL.

For delete, the custom LRO header may not be needed; using the template that specifies the `location` header only and final state via `location` header will change the SDK behavior from polling until completion to polling until completion plus a final get to the `location` header URL. However, this change is minor and should have no significant impact.

For post, only one operation (AgentPools/deleteMachines) is marked as final state via `azure-async-operation`, while the others are marked via `location` header. This results in a slight difference in SDK behavior, as it will skip the final get to the `location` header URL.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/PrivateEndpointConnection.tsp
FumingZhang: Now updated it to 200 + 204, since that's the expected behavior for a synchronous delete, which this operation implements. However, it's currently marked as LRO, which should use 202 + 204. I'm now adjusting it to 200 + 204 with sync, but this will introduce a **breaking change**.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/AgentPool.tsp
FumingZhang: That's true, but it appears the [lint rule](https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/post-operation-response-codes/#asynchronous) doesn't allow us to specify a 204 response for LRO post in the spec. Regarding "the actual [REST API specification](https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/async-api-reference.md#call-action-post-asynchronously) says 204", I believe this means the URL in the location header should return 200 if there's a response body, 204 if there's no body, and 202 if the operation is still ongoing. However, there's no clear way to represent this behavior in the swagger spec.

FumingZhang:
replaced by #40500

PR #40224, opened 2026-02-06
[NetApp] Add C# mgmt emitter configuration

live1206:
## Description

Add \@azure-typespec/http-client-csharp-mgmt\ emitter configuration to the NetApp TypeSpec tspconfig.yaml to enable .NET SDK generation from the TypeSpec spec.

### Changes
- Add csharp mgmt emitter options to \	spconfig.yaml\ with namespace \Azure.ResourceManager.NetApp\

### Related
- SDK PR: https://github.com/Azure/azure-sdk-for-net/pull/55560
- Tracking issue: https://github.com/Azure/azure-sdk-for-net/issues/54761

### DIFF
--- a/specification/netapp/resource-manager/Microsoft.NetApp/NetApp/Volume.tsp
+++ b/specification/netapp/resource-manager/Microsoft.NetApp/NetApp/Volume.tsp
@@ -829,7 +829,6 @@ union VolumeLanguage {
 /**
  * Set of export policy rules
  */
-@clientName("NetAppVolumeExportPolicyRule", "csharp")
 model VolumePropertiesExportPolicy {
   /**
    * Export policy rule
--- a/specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
+++ b/specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
@@ -387,6 +387,11 @@ union NetAppRelationshipStatusDef {
 @@clientName(BackupPatch, "NetAppBackupVaultBackupPatch", "csharp");
 @@clientName(UsageName, "NetAppUsageName", "csharp");
 @@clientName(UsageResult, "NetAppUsageResult", "csharp");
+@@clientName(PolicyStatus, "ElasticSnapshotPolicyStatus", "csharp");
+@@clientName(SecretPassword, "NetAppSecretPassword", "csharp");
+@@clientName(SecretPasswordIdentity, "NetAppSecretPasswordIdentity", "csharp");
+@@clientName(SnapshotUsage, "ElasticBackupSnapshotUsage", "csharp");
+@@clientName(VolumeSize, "ElasticBackupVolumeSize", "csharp");
 @@clientName(VolumeRelocationProperties,
   "NetAppVolumeRelocationProperties",
   "csharp"
@@ -471,6 +476,11 @@ union NetAppRelationshipStatusDef {
 );
 @@clientName(VolumeProperties.ldapEnabled, "IsLdapEnabled", "csharp");
 @@clientName(VolumeProperties.coolAccess, "IsCoolAccessEnabled", "csharp");
+@@clientName(VolumePatchProperties.coolAccess, "IsCoolAccessEnabled", "csharp");
+@@clientName(VolumePatchProperties.snapshotDirectoryVisible,
+  "IsSnapshotDirectoryVisible",
+  "csharp"
+);
 @@clientName(VolumeProperties.encrypted, "IsEncrypted", "csharp");
 @@clientName(ActiveDirectory.aesEncryption, "IsAesEncryptionEnabled", "csharp");
 @@clientName(ActiveDirectory.ldapSigning, "IsLdapSigningEnabled", "csharp");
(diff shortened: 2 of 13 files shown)


specification/netapp/resource-manager/Microsoft.NetApp/NetApp/Volume.tsp
```diff
@@ -815,7 +815,7 @@ union VolumeLanguage {
 /**
  * Set of export policy rules
  */
-@clientName("NetAppVolumeExportPolicyRule", "csharp")
+@clientName("NetAppVolumeExportPolicy", "csharp")
```
live1206, line 818: we should put this in client.tsp instead

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
```diff
@@ -89,7 +91,820 @@ using Microsoft.NetApp;
...
+
+  /** The operation was canceled. */
+  Canceled: "Canceled",
+
+  /** The resource is provisioning. */
+  Provisioning: "Provisioning",
+}
+
+/** Compatibility alias for NetApp volume quota rule provisioning state values. */
+union NetAppVolumeQuotaRuleProvisioningStateDef {
```
ArcturusZhang: If the spec defines it as fixed enum but we want it to be extensible in the generated code, we could do:
```
union NetAppVolumeQuotaRuleProvisioningStateDef {
OriginalFixedEnum, string
}
```
or just write customization code
```
public partial readonly struct EnumName {}
```
this will be much more simpler and it will automatically get new members when the spec updates.

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
```diff
@@ -89,7 +91,820 @@ using Microsoft.NetApp;
...
+
+// --- C# backward-compat enum (closed enum to preserve System.Enum base) ---
+
+// --- Preserve SDK-facing NetAppProvisioningState naming for backward compat ---
+// Old SDK (1.15.0) had NetAppProvisioningState as a closed System.Enum. The spec models it
+// as an extensible union; using a union here would emit a `readonly partial struct`, breaking
+// ApiCompat. Use a closed `enum` so the C# emitter generates `public enum NetAppProvisioningState`.
+/** Compatibility alias for NetApp provisioning state values. */
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Backward compat: must be closed enum to preserve System.Enum base type from v1.15.0."
+enum NetAppProvisioningStateDef {
```
ArcturusZhang: currently there is no clean way to change the enum back to fixed if it is written in extensible enums.

I suggest we just accept this change by renaming the impacted property and introduce some customization code for the removed property.

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
```diff
@@ -89,7 +91,820 @@ using Microsoft.NetApp;
...
+  "csharp"
+);
+@@alternateType(BucketPatchProperties.provisioningState,
+  NetAppVolumeQuotaRuleProvisioningStateDef,
+  "csharp"
+);
+@@alternateType(VolumeQuotaRulesProperties.provisioningState,
+  NetAppVolumeQuotaRuleProvisioningStateDef,
+  "csharp"
+);
```
ArcturusZhang: why not we just add the alternateType to the type of these properties?
The purpose of writing this new enum type is to replace one of the `enum` defined in the spec, right? we would never want the fixed enum with a bad name to leak into the sdk in the future.

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/routes.tsp
```diff
@@ -24,7 +33,7 @@ interface NetAppResourceOperationGroup {
     Request = ResourceNameAvailabilityRequest,
     Response = CheckAvailabilityResponse,
     Scope = SubscriptionActionScope,
-    Parameters = LocationParameter
+    Parameters = NetAppLocationParameter
```
ArcturusZhang: why we cannot use the built-in parameter `LocationResourceParameter`?

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
```diff
@@ -89,7 +91,797 @@ using Microsoft.NetApp;
...
+
+  /** The relationship is transferring data. */
+  Transferring: "Transferring",
+
+  /** The relationship failed. */
+  Failed: "Failed",
+
+  /** The relationship state is unknown. */
+  Unknown: "Unknown",
+}
```
ArcturusZhang: similar comment on this enum.
I am curious why we need this because when I search this name, I did not find an alternateType using it.

specification/netapp/resource-manager/Microsoft.NetApp/NetApp/client.tsp
```diff
@@ -89,7 +91,797 @@ using Microsoft.NetApp;
...
+@@clientName(VolumeRevert.snapshotId, "SnapshotId", "csharp");
+@@clientName(NetAppRelationshipStatusDef, "NetAppRelationshipStatus", "csharp");
+@@clientName(VolumePropertiesExportPolicy,
+  "NetAppVolumeExportPolicy",
+  "csharp"
+);
+
+// --- @@alternateType: property type overrides for C# ---
+// BackupsMigrationRequest.backupVaultId: armResourceIdentifier → string (old API had string)
+@@alternateType(BackupsMigrationRequest.backupVaultId, string, "csharp");
```
ArcturusZhang: for breaking changes like this, we should accept and rename the property to mitigate.

ArcturusZhang (approved):

PR #40236, opened 2026-02-06
[SDK Validation] Added support for multiple services scenario

### LINKED ISSUE #39974: [FEATURE REQ] SDK Automation to support Folder Structure v2 Multi-Project
### API Spec link

https://github.com/Azure/azure-rest-api-specs/pull/39105

### API Spec version

Not Applied

### Please describe the feature.

For cases like Compute and KeyVault, they have tspconfig.yaml file in both their parent folder and subservice folder 
```
specification/
└── <orgName1>/
    ├── cspell.yaml
    └── resource-manager/
        ├── readme.md			ß NOTE: For ARM schema validation; see bullet #2 below
        └── <RPNamespace>/		ß NOTE: Control-plane only (not data-plane) 
            └── <Service1>/	ß Customer-facing service; it has tspconfig.yaml in it.
                ├── <SubService1>/	    ß NOTE: Sub Service folder, which may contains its own tspconfig.yaml
                └── <SubService2>/	    ß NOTE: Sub Service folder, which may contains its own tspconfig.yaml
```

And some language may have SDK configuration under the Service1 folder, which should be sensitive to all files changes under the Service1 folder. Some language may have SDK configuration under the SubServiceX folder, which should be sensitive to all files changes under that sub service folder. 

The current SDK automation logic will search the tspconfig.yaml file in the deepest folder and track back to parent and root folders. But it hasn't supported cases like the SDK config might not necessarily exist in the deepest tspconfig.yaml. And each language may have different config at different layer.

We need to provide support to this.

raych1:
Resolved https://github.com/Azure/azure-rest-api-specs/issues/39974

**Key Changes :**

- Added `findAllParentsWithFile` function that traverses upward and collects all matching files instead of stopping at the first match

- Updated `searchRelatedParentFolders` function with an optional `findAll` parameter

- The new traversal logic is used when using the V2 folder structure, and t old folder structures (without resource-manager/data-plane) remain unchanged

**Test PRs :**
- https://github.com/Azure/azure-rest-api-specs/pull/40237
- https://github.com/Azure/azure-rest-api-specs/pull/40241
- https://github.com/Azure/azure-rest-api-specs/pull/40242


eng/tools/spec-gen-sdk-runner/test/utils/findAllParentsWithFile.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  });
+
+  test("stops at specified boundary", () => {
+    const result = findAllParentsWithFile(
+      "specification/contosowidgetmanager/resource-manager/Microsoft.Contoso/Service1/SubService1",
+      /^config\.json$/,
+      repoRoot,
+      "specification",
+    );
+    expect(result).toHaveLength(0);
```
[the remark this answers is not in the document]
raych1: @copilot can you fix this

raych1:
@weshaggard could you review this PR about supporting SDK generation for multiple services?

eng/tools/spec-gen-sdk-runner/src/utils.ts
```diff
@@ -233,16 +233,65 @@ export function findParentWithFile(
   return undefined;
 }
 
+/**
+ * Searches upward from a starting path to find ALL parent directories containing a file matching the given pattern
```
weshaggard, line 237: @mikeharder is this something we would need for other tools in the specs repo to support this multiple service scenario?

weshaggard:
> @weshaggard could you review this PR about supporting SDK generation for multiple services?

I think @mikeharder would be a better person to review this change as he understands the directory structure better than I do.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -185,6 +185,8 @@ stages:
         parameters:
           Paths:
             - '/*'
+            - '/specification'
```
mikeharder: what is the impact of this change?  how was this working before without these paths?

eng/tools/spec-gen-sdk-runner/src/utils.ts
mikeharder: We already have similar code to find suppressions:

https://github.com/Azure/azure-rest-api-specs/blob/d6e4facb4609b342308d8dbfa9cd50c8328e41ff/eng/tools/suppressions/src/suppressions.ts#L198

I will create a helper method in `.github/shared` for this, probably in a new file `fs.js`, since this is our first helper that actually needs to access the filesystem (we have existing helpers in `path.js`, but these are just string manipulation, no IO).

One simplification I might want to make, is have a single helper that returns all parents, rather than a second helper that short-circuits after the first match.  Perf is worse when you only need a single file, but it's one less helper we need.  Or we could add a flag to stop after the first result (or first N results).

mikeharder: Issue to create shared helper, but let's not block this PR on it:

- https://github.com/Azure/azure-rest-api-specs/issues/40768

mikeharder (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: @copilot can you revert this file change?

raych1: This change has been reverted and will create a separate PR later.

PR #40239, opened 2026-02-07
[APIView] Add TypeSpec metadata file support for APIView

AlitzelMendez:
This PR enables the TypeSpec pipeline to generate and pass metadata files to the APIView API, supporting the Project-based Organization feature.

issue: https://github.com/Azure/azure-sdk-tools/issues/13865


eng/scripts/Create-APIView.ps1
```diff
@@ -237,12 +238,32 @@ function Invoke-TypeSpecAPIViewParser {
     if ($LASTEXITCODE) {
       throw "Compilation error when running: 'npm exec --no -- tsp compile . --emit=@azure-tools/typespec-apiview --option @azure-tools/typespec-apiview.emitter-output-dir=$tempWorkingDirectoryPath/output/apiview.json'"
     }
+    
+    # Generate metadata file using @azure-tools/typespec-metadata emitter
+    if ($Type -eq "New") {
```
weshaggard, line 243: Does New imply TypeSpec? If so, this should be fine if not maybe we should ensure that there is a tspconfig or some other indicator that there is a TypeSpec.

weshaggard (approved):

eng/scripts/Create-APIView.ps1
tjprescott: The options seem to be "New" or "Baseline". I'm not sure what they mean though. @AlitzelMendez can you shed light on this?

tjprescott (changes requested):
Really just a question.

eng/scripts/Create-APIView.ps1
AlitzelMendez: `New` is the proposed API surface and `Baseline` is the existing API surface ((PR target, e.g. main)

revisions are created for both so they can do this kind of comparison:
https://spa.apiviewstagingtest.com/review/cf4497fafe1b49359843430156807396?activeApiRevisionId=3b792c1d526d4c86a17942426f20d98e&diffApiRevisionId=2bbf9d09a5d24360a94ea520ad50ef85&diffStyle=trees

as you can see one is revision for pull request 26849 and the other is the baseline of pull request 26849 

as we don't plan to compare generating the metadata for ` New` should be enough

tjprescott (approved):

PR #40275, opened 2026-02-10
fix query param impl

catalinaperalta:


catalinaperalta:
Decision to leave query params in the routes

PR #40276, opened 2026-02-10
update template for no body ops

catalinaperalta:


catalinaperalta:
Closing since there's another branch going into the rust base branch with the fix

PR #40278, opened 2026-02-10
Move client definition to client.tsp

catalinaperalta:
The `@client` decorator should be defined in client.tsp and also scoped down to rust, otherwise the change will trickle down to the other sdks which impacts migration efforts. 

This is just a one time configuration whenever you need to add a new client.


catalinaperalta:
Closing in favor of #40409

PR #40303, opened 2026-02-10
[openapi-diff-runner] Properly handle renames in same-version check

mikeharder:
- Fixes #38245

### DIFF
--- a/eng/tools/openapi-diff-runner/test/commands.test.ts
+++ b/eng/tools/openapi-diff-runner/test/commands.test.ts
@@ -231,52 +231,121 @@ const cases: TestCase[] = [
     changedFiles: {
       renames: [
         {
-          from: "Foo/stable/2025-01-01/foo.json",
-          to: "Foo/stable/2025-01-01/openapi.json",
+          from: "Foo/stable/2025-03-01/foo.json",
+          to: "Foo/stable/2025-03-01/openapi.json",
         },
       ],
     },
-    // TODO: After code is fixed, should not create *any* dummy swaggers
-    expectedCreateDummySwaggers: {
-      old: ["Foo/stable/2025-01-01/openapi.json"],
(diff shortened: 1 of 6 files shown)


eng/tools/openapi-diff-runner/test/commands.test.ts
```diff
@@ -231,52 +231,121 @@ const cases: TestCase[] = [
...
+        },
+        {
+          old: "specification/bar/data-plane/Bar/stable/2025-03-01/baz.json",
+          new: "specification/bar/data-plane/Bar/stable/2025-03-01/openapi.json",
+        },
+      ],
+    },
+  },
+  {
+    name: "convert one swagger to two (very rare)",
```
raych1, line 330: If it still fails, why don't we keep original behavior for this case without changing.

raych1 (approved):

eng/tools/openapi-diff-runner/test/commands.test.ts
mikeharder: We don't have special-case code for this scenario, it's more the behavior that "falls out" of the easiest implementation.  If there is one deleted file, and two matching files, it's easiest to just compare the deleted file with **each** matching file, even if the comparison is somewhat invalid.  It's better than any alternative, like comparing to a dummy file.

```js
// If there are multiple matching additions (which should be rare), it's safest
// to add all the combinations as "renames".  This will probably always trigger a check failure,
// but it's safest to "fail closed" this way, rather than ignoring the files.
for (const addition of additionMatches) {
  matchedRenames.push({ from: deletion, to: addition });
}
```

https://github.com/Azure/azure-rest-api-specs/pull/40303/changes#diff-fa95083e66071424e520c16c537555a254a9e59d9291699fbe7471ab4e3dc0dcR165-R170

Since this is an edge case, and the behavior of comparing all the files seems good enough, I'd rather not add more special-case code for this.

PR #40304, opened 2026-02-10
[applicationinsights/ApplicationInsights.LiveMetrics] Add support for modular definition, to allow correct serialization in JS

hectorhdzg:


specification/applicationinsights/data-plane/LiveMetrics/models.tsp
```diff
@@ -253,21 +253,26 @@ model MetricPoint {
 }
 
 /**
- * Base class of the specific document types.
+ * Common properties shared by all document types.
  */
 #suppress "@azure-tools/typespec-azure-core/casing-style" "Need to follow the casing style of the existing SDK contract."
-@discriminator("DocumentType")
```
pvaneck, line 259: The discriminator approach here is currently similar to what is being done in the Exporter TypeSpec PR. What serialization issues were you running into? I would've expected this approach to work here as well 🤔 .

specification/applicationinsights/data-plane/LiveMetrics/client.tsp
```diff
@@ -0,0 +1,218 @@
...
+@@usage(Azure.Monitor.LiveMetrics.FilterInfo, Usage.input | Usage.output);
+@@usage(Azure.Monitor.LiveMetrics.DocumentStreamInfo,
+  Usage.input | Usage.output);
+@@usage(Azure.Monitor.LiveMetrics.DocumentFilterConjunctionGroupInfo,
+  Usage.input | Usage.output);
+@@usage(Azure.Monitor.LiveMetrics.QuotaConfigurationInfo,
+  Usage.input | Usage.output);
+
+// ── camelCase client names for MonitoringDataPoint ──
+@@clientName(Azure.Monitor.LiveMetrics.MonitoringDataPoint.Version,
```
pvaneck, line 33: Hah, I take it the TS emitter doesn't automatically emit to camelCase then?

hectorhdzg: Yeah some many issues with case sensitive types here

specification/applicationinsights/data-plane/LiveMetrics/models.tsp
hectorhdzg: discriminator is back :), current code is fully functional now with current changes, most changes are JS specific to be able to have modular auto generated code, when creating the code modular library setting should need to be set as true

PR #40365, opened 2026-02-11
Enable ARMAutoSignedOff-Trivial in enforce mode

AkhilaIlla:
Enable ARMAutoSignedOff-Trivial in enforce mode


mikeharder (approved):
LGTM, but one question is back compat.  Say an existing PR ran the old code workflow, and has the test label.  Now the status workflow is updated, and looking for the non-test label.  So the exisitng PR may get some unexpected results.

Typically, we are OK with this, and just tell that PR "update from parent".  But, if we wanted, we could add compat code, so the status workflow processes **either** label.

I'm fine with either approach, we just need to understand the tradeoff.

mikeharder (approved):

PR #40414, opened 2026-02-13
[summarize-impact] Add label "TypeSpec" for *.tsp renames

mikeharder:
- fixes #40391


eng/tools/summarize-impact/src/cli.ts
```diff
@@ -77,6 +77,8 @@ export async function main() {
   const targetGitRoot = await getRootFolder(targetDirectory);
   const fileList = await getChangedFilesStatuses({
     cwd: sourceGitRoot,
+    // code in diff-types.ts and impact.ts assumes the file list only contains add/modify/delete, not rename
```
mikeharder:
```suggestion
    // code in diff-types.ts and impact.ts assumes the file list only contains add/modify/delete, not renames
```

PR #40420, opened 2026-02-13
Fix Diff in ServiceFabric

mcgallan:
# Choose a PR Template


specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/preview/2023-11-01-preview/servicefabric.json
```diff
@@ -2523,27 +2554,32 @@
...
+          "description": "The amount of time the overall upgrade has to complete before FailureAction is executed. It is first interpreted as a string representing an ISO 8601 duration. If that fails, then it is interpreted as a number representing the total number of milliseconds.",
+          "default": "P10675199DT2H48M5.4775807S"
         },
         "upgradeDomainTimeout": {
           "type": "string",
           "format": "duration-constant",
-          "description": "The amount of time each upgrade domain has to complete before FailureAction is executed. Cannot be larger than 12 hours. It is interpreted as a string representing an ISO 8601 duration with following format \"hh:mm:ss.fff\"."
+          "description": "The amount of time each upgrade domain has to complete before FailureAction is executed. It is first interpreted as a string representing an ISO 8601 duration. If that fails, then it is interpreted as a number representing the total number of milliseconds.",
```
iliu816: The changes to remove these defaults were deliberate as I believe they are incorrect. 

The representation of these values in our backend uses timespans https://msazure.visualstudio.com/One/_git/winfab-RP?path=/src/WRPService/Model/Public/ClusterUpgradePolicyExternal.cs&version=GBdevelop&line=24&lineEnd=25&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
 
And when looking at existing customer templates, I see they use a format like "00:20:00"
https://msazure.visualstudio.com/One/_git/AM-Alerts-ARMTemplates?path=/ServiceFabricApplication.Template.json&version=GBmaster&_a=contents&line=29&lineStyle=plain&lineEnd=30&lineStartColumn=1&lineEndColumn=1
 
https://msazure.visualstudio.com/One/_git/Mgmt-RecoverySvcs-Common?path=/src/Service/ServiceFabric/Deployments/FabricMTMonitoring/Templates/FabricMonitoringTemplate.json&version=GBdevelop&_a=contents&line=173&lineStyle=plain&lineEnd=174&lineStartColumn=1&lineEndColumn=1

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ApplicationResource.tsp
```diff
@@ -8,73 +8,112 @@ import "./Cluster.tsp";
...
 
 namespace Microsoft.ServiceFabric;
 /**
  * The application resource.
  */
-#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "servicefabric is brownfield service with legacy patterns for resources"
 @parentResource(Cluster)
 model ApplicationResource
-  is Azure.ResourceManager.Legacy.TrackedResourceWithOptionalLocation<ApplicationResourceProperties> {
+  is Azure.ResourceManager.ProxyResource<ApplicationResourceProperties> {
```
iliu816: How did we distinguish that it was acceptable to use ProxyResource for these resource definitions? The auto converter had used the Legacy.TrackedResourceWithOptionalLocation.

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ApplicationResource.tsp
```diff
@@ -199,39 +238,39 @@ model ArmRollingUpgradeMonitoringPolicy {
...
+  healthCheckStableDuration?: duration = duration.fromISO("PT0H2M0S");
 
   /**
-   * The amount of time to retry health evaluation when the application or cluster is unhealthy before FailureAction is executed. It is interpreted as a string representing an ISO 8601 duration with following format "hh:mm:ss.fff".
+   * The amount of time to retry health evaluation when the application or cluster is unhealthy before FailureAction is executed. It is first interpreted as a string representing an ISO 8601 duration. If that fails, then it is interpreted as a number representing the total number of milliseconds.
    */
   #suppress "@azure-tools/typespec-azure-core/known-encoding" "Backwards compatibility with non-standard duration encoding in c#"
   @encode("duration-constant")
-  healthCheckRetryTimeout?: duration;
+  healthCheckRetryTimeout?: duration = duration.fromISO("PT0H10M0S");
```
iliu816: same comment as above for the defaults restored here

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/Cluster.tsp
```diff
@@ -21,68 +21,89 @@ model Cluster is Azure.ResourceManager.TrackedResource<ClusterProperties> {
...
+  /**
+   * Azure resource etag.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "Existing property in original swagger"
+  @visibility(Lifecycle.Read)
+  etag?: string;
 }
 
-@armResourceOperations
+@armResourceOperations(#{ omitTags: true })
```
iliu816, line 33: why omittag and change the tag name from ClusterResource to Cluster?

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/preview/2023-11-01-preview/servicefabric.json
mcgallan: These changes were made to maintain full consistency with the original swagger. However, if the original swagger has issues and better modifications are available, we prefer to use the improved version. I'll revert this part to the main branch version.

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ApplicationResource.tsp
mcgallan: Using ProxyResource allows for customizing tags. The tags property in Legacy.TrackedResourceWithOptionalLocation does not have x-ms-mutability restrictions, whereas the original swagger specification includes such restrictions.
original swagger: https://github.com/Azure/azure-rest-api-specs/blob/61738889a51efdb1e04cf929d9a70d043776227f/specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/preview/2023-11-01-preview/application.json#L1579-L1589

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/Cluster.tsp
mcgallan: As mentioned above, this is done to maintain consistency between the generated and original swagger. The auto-generated swagger sometimes differs from the original swagger in terms of operation tags.

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ApplicationResource.tsp
mcgallan: all default values removed

iliu816 (approved):
specs changes LGTM. Lintdiff and breaking changes detected are in line with what's expected. but unsure why Java and Python SDKs are failing validation now.

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/models.tsp
```diff
@@ -10,6 +10,151 @@ using Azure.ResourceManager.Foundations;
...
+  /**
+   * The error message.
+   */
+  message?: string;
+}
+
+/**
+ * The resource model definition for proxy-only resource.
+ */
+model ProxyResource {
```
XiaofeiCao, line 46: This `ProxyResource` conflicts with `Azure.ResourceManager.ProxyResource`. TCGC would report error on this:
`error @azure-tools/typespec-client-generator-core/duplicate-client-name: Client name: "ProxyResource" is duplicated in language scope: "java"`

Why do we define this `ProxyResource`?

mcgallan: In the original Swagger file, ProxyResource model was defined. The ProxyResource that currently remains is used as a parent class for basic patch models, such as ApplicationResourceUpdate, and also for the creation of some implicit models, such as ClusterCodeVersions.

XiaofeiCao: Original Swagger doesn't use `ProxyResource` from common types. I think we shouldn't for now.

According to Chenjie, there's a way to achieve this: https://github.com/Azure/azure-rest-api-specs/blob/c020d9996429fe3efc0acda1052f999ef033fbcb/specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/Endpoint.tsp#L20-L53

Would you help have a try?

mcgallan: I modified the ProxyResource in ServiceFabric following the TrafficManager example by adding customAzureResource. Is this modification the expected result?

XiaofeiCao: I think the key point here is to get rid of usage of `Azure.ResourceManager.ProxyResource` so that TCGC won't have duplicate `ProxyResource`s.

So current usage:
```typespec
model ApplicationResource
  is Azure.ResourceManager.ProxyResource<ApplicationResourceProperties> {}

@armResourceOperations(#{ omitTags: true })
interface ApplicationResources {
  get is ArmResourceRead<ApplicationResource, Error = ErrorModel>;
}
```

should be replaced by 
```typespec
model ApplicationResource extends ProxyResource {}

alias ApplicationResourceOps = Azure.ResourceManager.Legacy.LegacyOperations<>;

@armResourceOperations(#{ omitTags: true })
interface ApplicationResources {
  get is ApplicationResourceOps.Read<ApplicationResource, Error = ErrorModel>;
}
```

/cc @tadelesh

mcgallan: replaced

mcgallan: replaced, only VMSizeResource extends Azure.ResourceManager.ProxyResource, because in original swagger:
https://github.com/Azure/azure-rest-api-specs/blob/61738889a51efdb1e04cf929d9a70d043776227f/specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/preview/2023-11-01-preview/cluster.json#L1949-L1974

XiaofeiCao: may I know why `VMSizeResource` needs to extend `Azure.ResourceManager.ProxyResource`? original swagger you showed doesn't..

mcgallan: Because `VMSizeResource` is a resource model, the original Swagger manually added id, name, and type to make it valid. However, TypeSpec doesn't allow this approach - it requires id, name, and type to be encapsulated as a model. 

At the same time, there is no similar resource model structure in the original Swagger that VMSizeResource can inherit from. This leads to only two methods to handle this situation: 
1. Extends from Azure.ResourceManager.ProxyResource. 
2. Introduce a custom resource model with the same content as Azure.ResourceManager.ProxyResource for inheritance. 

However, this would introduce a new model that never appeared in the original Swagger, and some languages don't prefer this kind of situation. I'm not sure which of these two approaches is the more appropriate solution.

XiaofeiCao: I see. Can we get rid of the inheritance hierarchy? Seems it's still valid (and doesn't even need `@subscriptionResource`). 
See [playground](https://azure.github.io/typespec-azure/playground/?options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-rulesets%2Fresource-manager%22%5D%7D%7D&c=aW1wb3J0ICJAYXp1cmUtdG9vbHMvdHlwZXNwZWMtxhVjb3JlIjsK3yvFFXJlc291cmNlLW1hbmFnZXLMN8grL29wZW5hcGnVHHJlc3QiOwoKdXNpbmcgVHlwZVNwZWMuUmVzdDvHFUHEasQSxWpNxmnIHckySHR0cNEVT3BlbkFQSTsKCkBhcm1Qcm92aWRlck5hbWVzcGFjZQpuyAogTWljcm9zb2Z0LlNlcnZpY2VGYWJyaWM7Ci8qKgogKiBEZXNjcmliZXMgYSBWTSBTaXplcy4KICovCiNzdXBwcmVzc%2F8BVW9yZS9jYeQAoC1zdHlsZSIgIlZNIGnEUXdlbGwta25vd24gYWJicmV2aWF0aW9uLiBCYWNrd2FyZHMgY29tcGF0aWJpbGl0eSB3aXRoIGV4aXN05QEhUEkuIgptb2RlbCBWTeQAm%2BgBLSB7CiAgLi7pAT3kAPpQYXJhbWV0ZXI8CiAgICDJKz3PPCzFH0tlecQ1ID0gInZtxB8ixhhTZWdtZW50yBx1buQBBm9ydGVkVsUnc8coxnJ0dGVybsQoIgogID7EGmlkOiBzdHJpbmfEDnByb3BlcnRpZXM6IHt9Cn0K8gF3dGhlIHJlc3VsdCBvZscOcXVlc3QgdG8gbGlzdOkBlyBmb3Ig5wHGIOYBxyBDbHVzdGVy%2FwGz%2FwGz%2FwGz%2FwGz%2FwGzc1LmAN1pc%2BcC8UNvcmUuUGFnZTzuAZg%2BO%2BkBHukA8uoBQP8A4f4Dvi9hcm3KFXDkAy9zaW9u5gEAYXTkAQBCcm93bmZpZWxk5AGOxSkgdGhhdCBjYW5ub3QgYmUgY3JlYXRlZC91cGTECCwgbm%2FkAKHJTCDGTP8Apf8Bhv8Bhv8BhvkBhuUDMeUBXiDqAWAg5AQnLsUTLwogIEDkAMrGUChMaWZlY3ljbOQEoGFkKQogIHNpemU%2F6ALcO%2BQCy2FsaWFzIOYDF09wcyA99gTSLkxlZ2FjeccHT3BlcuUAzHPkA6fkAJblA8lBcGlWZXLkAUnpA8c7yBxTdWLkAy1wxDZJZNIgx2DoBQbKGG9jxWroAJDNOn3kA5%2FGf%2BcBF%2FUBGecBGyAgQHBhdGjGCnPmBCMo9AQd5AEqICDmBFLqAS0gIH0K5ALf%2FwIa%2FwIa%2FwIa%2FwIa7AIa5AY16AEW6gGRKCN7IG9taXRUYWdzOiB0cnVlIH0pCmludGVyZuQGS%2B4Dr3PuAlNHZXQg6wEQIHZt5QIy%2FwS06AJ6c3VtbWFyeSgi30fVR%2BUBdUB0YWcoIucEc8YSZ2V0CiAgZ2XlBHzpAqflAtHvBHss5AYucG9uc%2BQF9EHlARzFDs8nPuQBxO8A9OQFseQFpnPkBb7zALr%2FBbb2AQLfVd9V%2FgEQQMRZCiDFYO4BGUxpc3T%2FARnvARnnBdzkARh9CgpAQGRvYyjOQS7kA0os5QK%2F6gNZIik71i3qBe%2FLM8oVxznrBKfbQO8Ewjs%3D&e=%40azure-tools%2Ftypespec-autorest&vs=%7B%7D), the only major diff is the `VMSizeResource` model.

XiaofeiCao: OK, seems this `VMSizeResource` is added along with the tsp migration, so there won't be breaking changes if we rename it.

Just curious, why not just use the user-defined `ProxyResource` instead of that from common types?

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ServiceResource.tsp
v-huizhu2, line 242: The TypeSpec definition @discriminator("type") is technically incorrect here since there are no subtypes to discriminate between. The model acts as a bag with a type tag, not a polymorphic base class. This pattern should ideally not use @discriminator at all — the type property should be a plain property of type ServicePlacementPolicyType without the @discriminator decorator.

XiaofeiCao: We probably will go "garbage in, garbage out" if service defines so..

Java emitter fix at: https://github.com/microsoft/typespec/pull/10080

XiaofeiCao (approved):

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/ServiceResource.tsp
mcgallan: When `@discriminator("type")` is removed, it causes the generated swagger to be missing `"discriminator": "type"`,
image
 which creates a difference from the original swagger. As for whether this difference is acceptable, you can ask @iliu816  to get more detailed information.

iliu816: After reading https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/

I believe this definition was originally supposed to follow the pattern of the underlying data layer at 

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/servicefabric/data-plane/ServiceFabric/stable/8.2/servicefabric.json

but it seems it wasn't fully defined. From the service side, we just pass the customer value through. Our concern is whether any SDKs are depend on the discriminator tag being present.

specification/servicefabric/resource-manager/Microsoft.ServiceFabric/ServiceFabric/models.tsp
weidongxu-microsoft: Same concern. Not a blocker though.

For `VMSizeResource`, we now see `VMSizeResource extends ArmProxyResource`, due to the deduplicate of this `ProxyResource` from common, and another `ProxyResource` defined here.

---

For now, to mitigate, Java will add a tspconfig option.

pshao25 (approved):

PR #40425, opened 2026-02-13
Added OperationRecoverySettings to ResliencyPolicy.

Avinash-Bondalapati:
Added OperationRecoverySettings to ResiliencyPolicy for VMScaleSet

OperationRecoverySettings contains - RestartRecoveryPolicy,  StartRecoveryPolicy and ReimageRecoveryPolicy.


audreyttt (approved):

audreyttt:
Adding the Approved-Avocado tag as this is a known failure for Compute that we have an exception for.

psah434:
Why use this custom branch?

Avinash-Bondalapati:
@psah434 Is there a hard guideline on this?

I had few changes made locally and I do not want to do additional bookkeeping, so updated a diff branch. But please let me know if this is required.

specification/compute/Compute.Management/models.tsp
```diff
@@ -3728,6 +3734,60 @@ model ResilientVMDeletionPolicy {
...
+
+  /**
+   * The configuration parameters used for start recovery policy.
+   */
+  startRecoveryPolicy?: StartRecoveryPolicy;
+
+  /**
+   * The configuration parameters used for reimage recovery policy.
+   */
+  reimageRecoveryPolicy?: ReimageRecoveryPolicy;
```
sandipsh, line 3762: why not have all three directly at the ResiliencyPolicy level similar to ResilientVMCreationPolicy instead of another level of indirection as OperationRecoverySettings?

also Settings doesn't align with your naming style of using "policy" as a suffix for everything here. #Resolved

specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2025-11-01/examples/virtualMachineScaleSetExamples/VirtualMachineScaleSet_Create_WithOperationRecoverySettings.json
```diff
@@ -0,0 +1,248 @@
+{
+  "parameters": {
+    "subscriptionId": "{subscription-id}",
+    "resourceGroupName": "myResourceGroup",
+    "vmScaleSetName": "{vmss-name}",
```
sandipsh, line 5: >{vmss-name} [](http://example.com/codeflow?start=23&length=11)

why use placeholder? #Resolved

specification/compute/Compute.Management/models.tsp
```diff
@@ -3706,6 +3706,12 @@ model ResiliencyPolicy {
    */
   @added(Versions.v2025_04_01)
   zoneAllocationPolicy?: ZoneAllocationPolicy;
+
+  /**
+   * The configuration parameters used for operation recovery settings.
+   */
+  @added(Versions.v2025_11_01)
+  operationRecoverySettings?: OperationRecoverySettings;
```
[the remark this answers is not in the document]
Avinash-Bondalapati: Since the description updates model for earlier versions too, this isn't valid  as OperationRecoverySettings is only applicable to latest version. so marking this as resolved.

specification/compute/Compute.Management/models.tsp
Avinash-Bondalapati: RestartRecoverySettings, (same for start and reimage) are all related to VMSS level operation settings that Cx can update to enable/disable auto recovery on failure of these vmss oeprations. Naming of these properties as Settings is by choice and to distinguish between the existing resilientCreate/resilientDelete features which are completely different in nature and have a higher level of control per operation type.

specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2025-11-01/examples/virtualMachineScaleSetExamples/VirtualMachineScaleSet_Create_WithOperationRecoverySettings.json
Avinash-Bondalapati: This is done to keep it similar with all other existing examples defined under VMSS.

specification/compute/Compute.Management/models.tsp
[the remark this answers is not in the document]
audreyttt: Please update the description, it is fine if it updates earlier versions if it is just a description change. In the description, you can call out the API version after which the new options are applicable.

audreyttt (changes requested):

audreyttt (approved):

Issue #39532, opened 2026-01-08
[EngSys] Folder Structure v2 Multi-Project


mikeharder:
## Overview

Add a new folder structure, named something like "folder structure v2 multi-project".  This extends our existing "folder structure v2" (https://github.com/Azure/azure-rest-api-specs/issues/34822), but adds support for multiple (rather than a single) TypeSpec projects.

One example is compute (https://github.com/Azure/azure-rest-api-specs/pull/39105), which will use this folder structure:

```
Compute/     
├── Compute
│   ├── *.tsp
│   └── tspconfig.yaml  # autorest config
├── ComputeDisk
│   ├── *.tsp
│   └── tspconfig.yaml  # autorest config
├── ComputeGallery
│   ├── *.tsp
│   └── tspconfig.yaml  # autorest config
├── ComputeSku
│   ├── *.tsp
│   └── tspconfig.yaml  # autorest config
├── preview/            # Include Compute, Disk, Gallery, Sku preview files
├── stable/             # Include Compute, Disk, Gallery, Sku stable files
├── client.tsp
├── readme.md
└── tspconfig.yaml      # SDK config
```

There is a top-level tspconfig.yaml for SDK config, then nested tspconfig.yaml for swagger generation.  To generate all the swagger for this spec, you need to know it's a special-case that requires compiling all the TSP sub-projects.

Another example is keyvault (https://github.com/Azure/azure-rest-api-specs/pull/38668), which also uses multiple TSP projects, but in the inverse way from compute (one autorest config, multiple sdk configs).

Since the folder structure of a spec is the basis for all other checks, we can't just "suppress" these errors.  We need to update TypeSpecValidation (and maybe other checks and processes), to actually "understand" how these multi-project specs generate their swagger and SDKs.

## Tasks
- [ ] Create samples under `widget` representing the patterns used by `compute` and `keyvault`
- [ ] Update rules in TSV to understand new structure
    - [ ] FolderStrucutre
    - [ ] Compile
    - [ ] SdkTspConfigValidation
- [ ] Document new folder structure at https://github.com/Azure/azure-rest-api-specs/wiki/Folder-Structure

qiaozha:
I don't think we should encourage other services to use similar patterns like Compute and KeyVault. I wonder if we can create a PR to make sure the new CI/checks pass but do not merge it into main?

qiaozha:
Other than the folderStructure validation, none of those two checks need to work based on swagger, right? 
 
[...]

PR #40496, opened 2026-02-18
[widget] add compute composite service

mikeharder:
- test for #39532


specification/widget/resource-manager/Microsoft.Widget/Compute/tspconfig.yaml
```diff
@@ -0,0 +1,19 @@
+options:
```
raych1, line 1: @qiaozha should we add other language options here too? Or they will be in the sub-service level tspconfig file?

qiaozha: it depends on if we want this widget to represent both Compute's case and KeyVault's case.

qiaozha: If yes, then they will be in sub service level tspconconfig file.

raych1: We might add other language options to sub-service level to cover both keyvault and compute scenarios in one sample folder. @mikeharder what do you think?

mikeharder: yes, i'd prefer to cover as many scenarios is as few samples as necessary

mikeharder (commented):
comment

specification/widget/resource-manager/Microsoft.Widget/Compute/Disk/main.tsp
```diff
@@ -0,0 +1,27 @@
...
+  /** 2021-10-01-preview version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v5)
+  v2021_10_01_preview: "2021-10-01-preview",
+
+  /** 2021-11-01 version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v5)
+  v2021_11_01: "2021-11-01",
+}
+
+interface Operations extends Azure.ResourceManager.Operations {}
```
qiaozha, line 27: we should not have duplicate Operations interface in both Disk and Gallery sub service folders.

mikeharder: should be fixed I removed from Gallery.  feel free to push any spec changes you want.

specification/widget/resource-manager/Microsoft.Widget/Compute/suppressions.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+- tool: TypeSpecValidation
+  rules: [SdkTspConfigValidation]
+  paths: ["Disk/tspconfig.yaml"]
+  reason: customize sub-rules for each folder
+
+- tool: TypeSpecValidation
+  rules: [SdkTspConfigValidation]
+  paths: ["Gallery/tspconfig.yaml"]
+  reason: customize sub-rules for each folder
```
qiaozha:
```suggestion
```
What would happen if we don't disable SdkTspConfigValidation rule?

timotheeguerin:

PR #40520, opened 2026-02-18
Misc fixes for blobs rust branch

catalinaperalta:
This PR adds a few more fixes for the spec:
- Update metadata headers to use optional header definition (Fixes https://github.com/Azure/azure-rest-api-specs/issues/40518)
- Keep same behavior for rust with overrides for set metadata operations
- Fix path for copy operation


vincenttran-msft (approved):
Not having it defined for both would do it.. Thanks!!!!

PR #40535, opened 2026-02-19
add Istio CNI to AKS 2026-01-01 GA API

SanyaKochhar:
Adds Istio CNI swagger changes to 2026-01-01 API

[API design doc](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/924239/-GA-Istio-AddOn-CNI)

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -1292,7 +1292,6 @@ union IstioIngressGatewayMode {
 /**
  * Mode of traffic redirection.
  */
-@added(Versions.v2026_01_02_preview)
 @added(Versions.v2026_01_01)
 union ProxyRedirectionMechanism {
   string,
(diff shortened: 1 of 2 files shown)


FumingZhang (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -1293,6 +1293,7 @@ union IstioIngressGatewayMode {
  * Mode of traffic redirection.
  */
 @added(Versions.v2026_01_02_preview)
```
FumingZhang: I realized that we don’t actually need to keep the preview version decorator, since we only have one preview API version and it will always be included in the preview API version, even without the mark.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -3658,6 +3659,7 @@ model IstioComponents {
    * Mode of traffic redirection.
    */
   @added(Versions.v2026_01_02_preview)
```
FumingZhang: same as above

PR #40547, opened 2026-02-19
Add GitHub agentic workflow for SDK generation

praveenkuttappan:
This PR creates a GitHub agent workflow as POC to verify SDK generation using agentic workflow.


.github/actionlint.yaml
```diff
@@ -0,0 +1,4 @@
+paths:
+  .github/workflows/sdk-generation-agent.lock.yaml:
```
mikeharder: if these files are auto-generated, should we exclude them from linting completely?  probably exclude all `.github/workflows/*.lock.yaml`?

cspell.yaml
```diff
@@ -13,6 +13,8 @@ ignorePaths:
   - eng/**
   - node_modules/**
   - specification/suppressions.yaml
+  - .github/workflows/sdk-generation-agent.lock.yaml
```
mikeharder:
```suggestion
  - .github/workflows/*.lock.yaml
```

cspell.yaml
```diff
@@ -13,6 +13,8 @@ ignorePaths:
...
+  - .github/workflows/sdk-generation-agent.md
```
mikeharder, line 17: I think we run cspell on most MD files by default, so we should be able to on this one as well?

praveenkuttappan: There are few gh-aw specific words that was causing failures. I can resolve them in an incremental PR by adding override per word to avoid skipping entire file from the check.

praveenkuttappan: Filed a follow up issue to track this. https://github.com/Azure/azure-rest-api-specs/issues/40559

.github/workflows/sdk-generation-agent.md
```diff
@@ -0,0 +1,203 @@
...
+  - name: Acquire OIDC token for Azure
+    id: oidc
+    uses: actions/github-script@v7
+    with:
+      script: |
+        const token = await core.getIDToken('api://AzureADTokenExchange');
+        const fs = require('fs');
+        fs.writeFileSync('/tmp/azure-oidc-token', token);
+
+  - name: Verify Azure CLI authentication
```
benbp, line 48: Is this to support lazy auth? Might be better to put this into an action in the repo we can reference (for future)

.github/workflows/sdk-generation-agent.md
```diff
@@ -0,0 +1,203 @@
...
+## Monitoring and Status Updates
+
+1. After successful trigger, monitor the pipeline run referenced in the CLI output.
+2. Poll status every 5 minutes by querying the pipeline's status endpoint or, when available, `azsdk` status commands using the recorded run identifier.
+3. On each poll, determine whether pipeline is still running, failed, or completed.
+4. If still running, update status via comment (keep concise).
+5. If failed, add a comment indicating failure and include pipeline link and failure summary.
+6. If completed:
+
+- Refresh release plan data via `azsdk release-plan get --work-item-id <WORK_ITEM_ID> --release-plan-id <RELEASE_PLAN_ID>` and inspect the SDK pull request references per language.
```
benbp:
```suggestion
- Refresh release plan data via `/tmp/bin/azsdk release-plan get --work-item-id <WORK_ITEM_ID> --release-plan-id <RELEASE_PLAN_ID>` and inspect the SDK pull request references per language.
```

benbp (approved):

.github/workflows/sdk-generation-agent.md
praveenkuttappan: This is just a troubleshooting step in case auth fails to understand why. I will remove this once I get the agent flow working.

Issue #40456, opened 2026-02-13
[storage tsp] FilterBlobItem model discrepancies


l0lawrence:
<img width="1159" height="263" alt="Image" src="https://github.com/user-attachments/assets/86b2e2e7-56df-45f2-a907-ccfad44da886" />

It looks like the xml for BlobTags is now lowercase when it used to be uppercase

l0lawrence:
let me move to python emitter

catalinaperalta:
Fixed in #40569

PR #40569, opened 2026-02-20
Fix tags property

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/40456

The unwrapped decorator wasn't needed here and the xml.name for the actual property was missing.


vincenttran-msft (approved):
LGTM, tested in Rust.

PR #40598, opened 2026-02-20
Sdk generation agent auth changes

praveenkuttappan:
This PR adds workflow instructions to create SDK generation agent and common authentication import which can be shared across GitHub agent workflow that requires Azure authentication.


.github/actionlint.yaml
```diff
@@ -1,4 +1,4 @@
 paths:
-  .github/workflows/*.lock.yaml:
+  .github/workflows/*.lock.yml:
```
mikeharder, line 2: @copilot: how could the path match both `.yaml` and `.yml`?

mikeharder (approved):
approved, though I wonder if we want to support `*.lock.yaml` in addition to `*.lock.yml`.  If the lockfile is always `yml`, I suppose it's all we need.  No matter what we do, this `yaml/yml` split is going to be annoying.

.github/actionlint.yaml
mikeharder: just supporting yml is fine

mikeharder (approved):
just supporting lock.yml is fine

.github/workflows/sdk-generation-agent.md
```diff
@@ -140,37 +82,40 @@ This workflow can be triggered in three ways:
...
-   - Check if API spec PR is in open status or merged status.
 
-- If such a PR is found and if it's open, set source branch to exactly `refs/pull/<PR number>`.
+- Identify TypeSpec API spec PR from issue context.
+- Check if API spec PR is in open status or merged status.
+- If such a PR is found and if it's open, set source branch for SDK generation to exactly `refs/pull/<PR number>`.
 - If no such PR is found, use default branch context.
 
-5. Use the azsdk CLI at `/tmp/bin/azsdk` (installed earlier) to gather release plan metadata and required arguments:
+5. Use the azsdk CLI at `$AZSDK_CLI_PATH/azsdk` (installed earlier) to gather release plan metadata and required arguments:
```
benbp: Probably easier at this point to add azsdk to $PATH instead

.github/workflows/sdk-generation-agent.md
```diff
@@ -32,87 +22,32 @@ if: >
...
   AZURE_TENANT_ID: 72f988bf-86f1-41af-91ab-2d7cd011db47
   AZURE_FEDERATED_TOKEN_FILE: /tmp/azure-oidc-token
-strict: false
-network:
-  allowed:
-    - defaults
-    - "login.microsoftonline.com"
-    - "dev.azure.com"
-    - "vssps.dev.azure.com"
+  GITHUB_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN || secrets.GITHUB_TOKEN }}
```
weshaggard, line 40: @praveenkuttappan I assume this is the PAT you were telling me about. Where does this actually get used?

.github/workflows/sdk-generation-agent.md
praveenkuttappan: Done.

.github/workflows/sdk-generation-agent.md
praveenkuttappan: Nope. It is COPILOT_GITHUB_TOKEN. GITHUB_TOKEN is the default one from the GitHub actions and not from the azure-sdk user.

praveenkuttappan: @weshaggard  It's https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/sdk-generation-agent.lock.yaml#L687 that uses COPILOT_GITHUB_TOKEN in the compiler generated step

.github/workflows/sdk-generation-agent.md
```diff
@@ -140,37 +75,40 @@ This workflow can be triggered in three ways:
...
 If the triggering event does not meet its corresponding requirements, immediately call `noop` with guidance (for example: missing label, missing `Regenerate SDK`, or missing workflow_dispatch inputs).
 
 ## Workflow Behavior
 
 When validation succeeds, execute the following steps in order.
 
-1. Add a debug comment on the target issue with the workflow run link:
+1. Azure CLI Login using az login:
+
+- Run `mkdir -p /tmp/gh-aw/agent/.azure`.
```
benbp: Curious why this is duplicated with the import?

benbp (approved):

.github/workflows/sdk-generation-agent.md
praveenkuttappan: I was supposed to remove it but forgot. I will remove it.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
```diff
@@ -0,0 +1,107 @@
...
+    - "login.microsoftonline.com"
+    - "azure.com"
+    - "visualstudio.com"
+steps:
+  - name: Acquire OIDC token for Azure
+    id: oidc
+    uses: actions/github-script@v7
+    with:
+      script: |
+        const token = await core.getIDToken('api://AzureADTokenExchange');
```
weshaggard, line 16: I understand why you did this, but I don't like us putting this into a file but for now until we have a better option we can use this for testing.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
```diff
@@ -0,0 +1,107 @@
...
+      fi
+
+      if [[ -z "${GH_TOKEN:-}" ]]; then
+        echo "GH_TOKEN environment variable is not available for gh CLI authentication." >&2
+        exit 1
+      fi
+
+      gh config set prompt disabled >/dev/null
+      pushd "${GITHUB_WORKSPACE}" >/dev/null
+      gh aw secrets set COPILOT_GITHUB_TOKEN --value "${AZURESDK_COPILOT_TOKEN}"
```
weshaggard, line 88: Please file a bug against the agentic workflow team about this workaround.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
```diff
@@ -0,0 +1,107 @@
...
+  - name: Fetch secret from Azure Key Vault
+    id: fetch_secret
+    shell: bash
+    run: |
+      set -euo pipefail
+      AZURESDK_GITHUB_TOKEN=$(az keyvault secret show --vault-name "AzureSDKEngKeyVault" --name "azuresdk-github-pat" --query value -o tsv)
+      AZURESDK_COPILOT_TOKEN=$(az keyvault secret show --vault-name "AzureSDKEngKeyVault" --name "azuresdk-copilot-github-pat" --query value -o tsv)
+      echo "::add-mask::${AZURESDK_GITHUB_TOKEN}"
+      echo "::add-mask::${AZURESDK_COPILOT_TOKEN}"
+      echo "AZURESDK_GITHUB_TOKEN=${AZURESDK_GITHUB_TOKEN}" >> "${GITHUB_ENV}"
```
weshaggard: Do these need to be put into the environment? Can we just use them in the one step and not have them be in env?

weshaggard (approved):
I'm good with this for now but I don't want to spread this everywhere until we get better security support.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
```diff
@@ -0,0 +1,107 @@
+---
```
timovv, line 1: Any way we can get this in the language repos as well? I would like to use this in the JS repo for https://github.com/Azure/azure-sdk-for-js/pull/37343

timovv: OK I see the "not right now" comment... I would still like this sooner rather than later though...

edit to clarify: I understand the security concern but there are lots of things I want to try out in this space so hopeful we can get this resolved! :)

weshaggard: Not yet. There are some auth concerns we are trying to understand before we try and push this other places.

weshaggard: @timovv @xirzec in the short-term if you want to use your own personal PAT in the repo secret to play around and experiment that should be fine.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
praveenkuttappan: I will try a change to do this in one shell without setting them as env variable.

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
praveenkuttappan: Done

.github/workflows/shared-github-aw-imports/global_networks_auth_import.md
praveenkuttappan: Merged to one step and removed setting env variables

cspell.yaml
```diff
@@ -13,8 +13,9 @@ ignorePaths:
   - eng/**
   - node_modules/**
   - specification/suppressions.yaml
-  - .github/workflows/*.lock.yaml
+  - .github/workflows/*.lock.yml
   - .github/workflows/sdk-generation-agent.md
+  - .github/workflows/shared-github-aw-imports/global_networks_auth_import.md
```
mikeharder, line 18: Maybe we should just exclude `.github/workflows/**/*.md`?  We certainly don't **need** spellcheck for any MD files under `.github/workflows`, and I'd rather not need to keep updating this for new workflows.

CC: @danieljurek

praveenkuttappan: I will submit a separate PR to ignore workflow from cspell and lock.yaml from prettier check.

PR #40611, opened 2026-02-20
fix: update ContentUnderstanding spec for c# client fixes

jorgerangel-msft:
Fixes the createOrReplace operation for .net. 

Updated generated code : https://github.com/Azure/azure-sdk-for-net/pull/56396

### DIFF
--- a/specification/ai/ContentUnderstanding/client.tsp
+++ b/specification/ai/ContentUnderstanding/client.tsp
@@ -1,11 +1,14 @@
+import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-client-generator-core";
 import "@typespec/rest";
 import "@typespec/http";
 import "@typespec/versioning";
 import "./main.tsp";
 
+using Azure.Core;
 using Azure.ClientGenerator.Core;
 using ContentUnderstanding;
+using TypeSpec.Http;
 
 namespace ClientCustomizations;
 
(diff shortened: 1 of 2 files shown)


specification/ai/ContentUnderstanding/client.tsp
```diff
@@ -88,6 +115,13 @@ interface ContentUnderstandingClient {
   DocumentationMode.replace
 );
 
+// .NET specific parameter for custom lro method
+alias AllowReplaceParameter = {
```
JoshLove-msft, line 119: Do we need the alias for this vs just inline?

specification/ai/ContentUnderstanding/client.tsp
```diff
@@ -96,3 +130,5 @@ interface ContentUnderstandingClient {
 
 @@convenientAPI(ContentAnalyzers.update, false, "csharp");
 @@convenientAPI(ContentAnalyzers.updateDefaults, false, "csharp");
+@@alternateType(SupportedModels.completion, string[], "csharp");
```
JoshLove-msft, line 133: Why is this necessary?

jorgerangel-msft: I don't know? Their service is returning a string array. Their spec says it's an object. It seems there was validation missing when this was checked in. We can add this customization, edit the spec directly, or just ignore generation until they fix whatever the true types of these properties are.

specification/ai/ContentUnderstanding/client.tsp
```diff
@@ -23,9 +26,33 @@ interface ContentUnderstandingClient {
 
   copyAnalyzer is ContentAnalyzers.copy;
```
jorgerangel-msft, line 27: The .net samples are failing for this operation https://dev.azure.com/azure-sdk/public/_build/results?buildId=5912385&view=logs&j=1e36f030-c9b9-5384-3434-c17c7b35c565&t=74c65e08-7fe6-58ed-d1c3-1bcdc9f8bcd1 because the operation is returning a 201, but LROs are expected to return a 202 status code.

specification/ai/ContentUnderstanding/routes.tsp
```diff
@@ -182,6 +182,7 @@ interface ContentAnalyzers {
   @added(Versions.v2025_11_01)
   @doc("Create a copy of the source analyzer to the current location.")
   @pollingOperation(ContentAnalyzers.getOperationStatus)
+  @action("copy")
```
jorgerangel-msft, line 185: the uri is `copy` not `copyAnalyzer` based on the testing against the service.

PR #40613, opened 2026-02-20
WIP - queues typespec

catalinaperalta:
Using this PR to see what CI checks say about the migration.


specification/storage/Microsoft.QueueStorage/models.tsp
```diff
@@ -0,0 +1,761 @@
...
+  @Xml.name("Message") message?: string;
+
+  /** Copy source status code */
+  @Xml.name("CopySourceStatusCode") copySourceStatusCode?: int32;
+
+  /** Copy source error code */
+  @Xml.name("CopySourceErrorCode") copySourceErrorCode?: string;
+
+  /** Copy source error message */
+  @Xml.name("CopySourceErrorMessage") copySourceErrorMessage?: string;
```
catalinaperalta, line 305: Several of these properties werent defined in the original swagger + seems like they correspond to the headers defined earlier in the same model, are these properties repeated?

PR #40616, opened 2026-02-20
[Queues] Fixes for migration deviations from swagger

catalinaperalta:
This PR addresses deviations in the typespec definition from the swagger. 

Several changes in this PR were pulled into this  PR #40661 and merged to the base branch.

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -36,31 +36,36 @@ model BlobServiceClientParameters {
   {
     parameters: ContainerNamePathParameter,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
+  },
+  "rust"
 );
 @@clientInitialization(Storage.Blob.Container.Blob,
   {
     parameters: BlobServiceClientParameters,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
+  },
+  "rust"
 );
 @@clientInitialization(Storage.Blob.Container.Blob.AppendBlob,
   {
     parameters: BlobServiceClientParameters,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
+  },
+  "rust"
 );
 @@clientInitialization(Storage.Blob.Container.Blob.BlockBlob,
   {
     parameters: BlobServiceClientParameters,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
+  },
+  "rust"
 );
 @@clientInitialization(Storage.Blob.Container.Blob.PageBlob,
   {
     parameters: BlobServiceClientParameters,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
+  },
+  "rust"
 );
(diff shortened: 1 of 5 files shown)


specification/storage/Microsoft.QueueStorage/models.tsp
```diff
@@ -453,7 +453,7 @@ model ReceivedMessage {
 
 /** Represents an array of signed identifiers */
 @Xml.name("SignedIdentifiers")
-model ListOfSignedIdentifier {
```
catalinaperalta: I dont think architects would like this name, but if it is needed for rust it can be changed with the clientName decorator

specification/storage/Microsoft.QueueStorage/models.tsp
```diff
@@ -453,14 +453,13 @@ model ReceivedMessage {
...
+model SignedIdentifiers {
   /**
    * The list of signed identifiers.
    *
    */
-  @Xml.unwrapped
   @Xml.name("SignedIdentifier")
-  items: Array<SignedIdentifier>;
```
catalinaperalta, line 463: This was marked as wrapped true in the swagger and also as optional

vincenttran-msft: iirc this stuff is needed, you can see we made the change in the opposite direction when we implemented in Blobs:
https://github.com/Azure/azure-rest-api-specs/pull/38633

There was an issue with the XML deserialization otherwise.

catalinaperalta: Ok I'll revert this change for now, but I will need feedback from the other languages to see if this definition does actually represent the rest api well.

specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -36,31 +36,36 @@ model BlobServiceClientParameters {
   {
     parameters: ContainerNamePathParameter,
     initializedBy: InitializedBy.parent | InitializedBy.individually,
-  }
```
vincenttran-msft, line 39: Also, I think since these are Blob changes they should be targeted in https://github.com/Azure/azure-rest-api-specs/tree/feature/blob-tsp-rust since we will probably resolve merge by assuming `Microsoft.BlobStorage` is the source of truth from that branch instead of this branch (since this is the queue branch and we are not maintaining the Blob accuracy really).

I think the long-term strategy is that each feature branch contains just that specific crates working feature branch, then eventually when it's all in `main` it will be everything right.

catalinaperalta: That was a mistake, I removed the blobs changes and applied the fix to the queues client.tsp. Please review

specification/storage/Microsoft.QueueStorage/models.tsp
catalinaperalta: Fixed in #40661

vincenttran-msft (approved):
LGTM, all the other changes pulled through cleanly and I just missed the scaling the change down to Rust. Thanks!

PR #40619, opened 2026-02-21
[Storage] Remove convenience layer in Queues TypeSpec Changes

vincenttran-msft:


specification/storage/Microsoft.QueueStorage/models.tsp
catalinaperalta, line 684: @vincenttran-msft this is going to result in the same issue we're dealing with in blobs. This change might have to be reverted and the metadata header will probably need to be optional if that's what the service does. Waiting to see if there's a possible solution through the overload decorator

specification/storage/Microsoft.QueueStorage/client.tsp
catalinaperalta, line 36: Scoping these down to rust in #40616

catalinaperalta, line 15: Just curious, why did the order of these client definitions change?

PR #40665, opened 2026-02-24
Extract label selector out to its own model.

dvadas:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
@@ -1,5 +1,6 @@
 import "@azure-tools/typespec-azure-core";
-using Azure.Core;
+import "./selector.tsp";
+
 using Azure.ResourceManager;
 using Azure.ResourceManager.Foundations;
 using TypeSpec.Http;
@@ -34,10 +35,9 @@ model ClusterMeshProfileProperties {
   @visibility(Lifecycle.Read)
   provisioningState?: ClusterMeshProfileProvisioningState;
 
-  @doc("Kubernetes-style label selector for selecting Fleet members to join the mesh, e.g. `env=production`. If empty or not specified, no Fleet members will be selected to join the mesh.")
-  @maxLength(512)
+  @doc("Select the members of the mesh. If empty or not specified, no Fleet members will be selected to join the mesh.")
   @visibility(Lifecycle.Read, Lifecycle.Create)
-  memberLabelSelector?: string;
+  memberSelector?: MemberSelector;
 
   @doc("The cluster mesh profile status.")
   @visibility(Lifecycle.Read)
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
@@ -35,7 +35,11 @@ model ClusterMeshProfileProperties {
   @visibility(Lifecycle.Read)
   provisioningState?: ClusterMeshProfileProvisioningState;
 
-  @doc("Select the members of the mesh. If empty or not specified, no Fleet members will be selected to join the mesh.")
+  @doc("""
+    Select the members of the mesh.
+      * Only key/value pairs with the `=` operator are accepted in the label selector.
+      * If empty or not specified, no Fleet members will be selected to join the mesh.
+    """)
   @visibility(Lifecycle.Read, Lifecycle.Create)
   memberSelector?: MemberSelector;
 
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
@@ -3540,7 +3540,7 @@
         },
         "memberSelector": {
           "$ref": "#/definitions/MemberSelector",
-          "description": "Select the members of the mesh. If empty or not specified, no Fleet members will be selected to join the mesh.",
+          "description": "Select the members of the mesh.\n  * Only key/value pairs with the `=` operator are accepted in the label selector.\n  * If empty or not specified, no Fleet members will be selected to join the mesh.",
           "x-ms-mutability": [
             "read",
             "create"
@@ -4789,7 +4789,7 @@
       "properties": {
         "byLabel": {
           "type": "string",
-          "description": "Kubernetes-style label selector for selecting Fleet members, e.g. `env=production`.\n\nThe following restrictions apply compared to standard Kubernetes syntax:\n  * Only key/value pairs with the `=` operator are accepted.\n  * Keys may not be repeated, that is `env=production,env=staging` is invalid.",
+          "description": "Kubernetes-style label selector for selecting Fleet members, e.g. `env=production`.",
           "maxLength": 512,
           "x-ms-mutability": [
             "read",
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
@@ -3474,7 +3474,7 @@
         "id": {
           "type": "integer",
           "format": "int32",
-          "description": "Cilium requires each cluster to be assigned a unique numeric cluster id. The id is managed by Fleet and cannot be set by the user.",
+          "description": "Cilium requires each cluster to be assigned a unique numeric cluster id from 1 - 255. The id is managed by Fleet and cannot be set by the user.",
           "readOnly": true
         },
         "name": {
@@ -3538,10 +3538,9 @@
           "description": "The provisioning state of the cluster mesh profile.",
           "readOnly": true
         },
-        "memberLabelSelector": {
-          "type": "string",
-          "description": "Kubernetes-style label selector for selecting Fleet members to join the mesh, e.g. `env=production`. If empty or not specified, no Fleet members will be selected to join the mesh.",
-          "maxLength": 512,
+        "memberSelector": {
+          "$ref": "#/definitions/MemberSelector",
+          "description": "Select the members of the mesh. If empty or not specified, no Fleet members will be selected to join the mesh.",
           "x-ms-mutability": [
             "read",
             "create"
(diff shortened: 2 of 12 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
```diff
@@ -0,0 +1,19 @@
...
+  @doc("""
+    Kubernetes-style label selector for selecting Fleet members, e.g. `env=production`.
+    
+    The following restrictions apply compared to standard Kubernetes syntax:
+      * Only key/value pairs with the `=` operator are accepted.
+      * Keys may not be repeated, that is `env=production,env=staging` is invalid.
+    """)
+  @maxLength(512)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  byLabel: string;
```
dvadas, line 12: If this is essentially part of a oneof, should it be optional?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
```diff
@@ -0,0 +1,19 @@
+import "@azure-tools/typespec-azure-core";
+using TypeSpec.Versioning;
+
+namespace Microsoft.ContainerService;
+
+@doc("Select members of a fleet.")
+@added(Versions.v2026_03_02_preview)
+model MemberSelector {
```
dvadas, line 8: I've named this model and its field so that you don't repeat the same word over and over again. That is `memberSelector.byLabel` instead of `memberLabelSelector.selectorLabel`.

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
```diff
@@ -34,10 +35,9 @@ model ClusterMeshProfileProperties {
   @visibility(Lifecycle.Read)
   provisioningState?: ClusterMeshProfileProvisioningState;
 
-  @doc("Kubernetes-style label selector for selecting Fleet members to join the mesh, e.g. `env=production`. If empty or not specified, no Fleet members will be selected to join the mesh.")
-  @maxLength(512)
+  @doc("Select the members of the mesh. If empty or not specified, no Fleet members will be selected to join the mesh.")
```
dvadas:
```suggestion
  @doc("... If empty or not specified, no Fleet members will be selected to join the mesh.")
```

I've documented this here rather than in the generic selector model to match what it says in the Kubernetes doc (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors):

> The semantics of empty or non-specified selectors are dependent on the context, and API types that use selectors should document the validity and meaning of them.

But it's a bit awkward and maybe we just make empty = no match the case everywhere?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
dvadas: Maybe `byLabelMatch` is better?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
```diff
@@ -0,0 +1,19 @@
...
+  @doc("""
+    Kubernetes-style label selector for selecting Fleet members, e.g. `env=production`.
+    
+    The following restrictions apply compared to standard Kubernetes syntax:
+      * Only key/value pairs with the `=` operator are accepted.
+      * Keys may not be repeated, that is `env=production,env=staging` is invalid.
```
nibooy: I'm not sure if we want to specify this as a hard restriction. 
Aren't there cases where a user might want to select all clusters in `region=westus,region=westeurope` to be included in a stage of an update run?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
nibooy: I personally think `memberSelect.byLabel` reads the most cleanly.
If we want to support more operators in the future, the field name is vague enough to continue using just by relaxing the validation.
If we want to support matching other fields, we can add those as well, i.e. `byName`, `byPercentage`

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
nibooy: Since we're only shipping the `byLabel` field for now, can we make it required for now and update it to optional when we add more fields? AFAIK, changing a field from required to optional is non-breaking?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
```diff
@@ -3474,7 +3474,7 @@
         "id": {
           "type": "integer",
           "format": "int32",
-          "description": "Cilium requires each cluster to be assigned a unique numeric cluster id. The id is managed by Fleet and cannot be set by the user.",
+          "description": "Cilium requires each cluster to be assigned a unique numeric cluster id from 1 - 255. The id is managed by Fleet and cannot be set by the user.",
```
serbrech, line 3477: this field is readonly, and the value depends on the mesh limit configuration.
I don't think you should specify the range here. it might change in the future, and the API will then have wrong docs

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
serbrech: I think that's good. it's optional, so we allow passing nil, it just doesn't select members. should the memberSelector be required? what was the reasoning for making it optional?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
serbrech: for updaterun, I agree.
for clustermesh, it's less clear that this makes sense.
so the restriction is probably specific to the use-case. it makes sense to me.
we have a construct to do labelselector, depending on the case, the validation can be different.

so the restriction should be documented on the parent model, not on this field.

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
dvadas: Yes, I believe we can do it that way.

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/clustermeshprofile.tsp
dvadas: > should the memberSelector be required? what was the reasoning for making it optional?

You suggested it be optional in the previous PR: https://github.com/Azure/azure-rest-api-specs/pull/40256#discussion_r2785823978

Some more offline context:

> a safe approach at the api layer might be to allow nil, and default it to what we want as behavior.
> simplifies the required minimal model to send, unless we think this is harmful to the user experience
 
With the selector as a full model now, does that make you want to make it required again?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
dvadas: I think it's confusing to have a shared selector model where the restrictions are different depending on where it's being used. You should be able to construct a MemberSelector and know it's valid, or be able to use the same selector in different places without being told, no you can't use that here. I thought that was the point of having this generic model, that it would be the intersection of all requirements (update run, clustermesh, etc), otherwise we may as well define the selector separately for each usage.

This single key restriction came from a discussion we had about how [cluster configuration](https://microsoft.sharepoint.com/:fl:/s/92fda5f4-07ec-4c0b-8e70-0c986e3e83a1/IQDPnjekE7E5QrhtRrzyyBlgASCDO5BuZ3ZXF-jg0ac-Wr8?e=reEewg&nav=cz0lMkZzaXRlcyUyRjkyZmRhNWY0LTA3ZWMtNGMwYi04ZTcwLTBjOTg2ZTNlODNhMSZkPWIlMjE0bU5sOTBqRVQwaU1mTVBNZExMSmwtZUMxN3BSQUo5T2lzbFBabEI3aWh1ZjhxYUVITHhOUUowUWdLZmJDcWJsJmY9MDFZMlgzRU5XUFRZMzJJRTVSSEZCTFEzS0dYVFpNUUdMQSZjPSUyRiZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUUwYlU1c09UQnFSVlF3YVUxbVRWQk5aRXhNU213dFpVTXhOM0JTUVVvNVQybHpiRkJhYkVJM2FXaDFaamh4WVVWSVRIaE9VVW93VVdkTFptSkRjV0pzZkRBeFdUSllNMFZPVjAxWk16Y3lVRFZETnpOS1JVczJXbGxOVVVGTFdVZFlUMUElM0QlMjIlMkMlMjJpJTIyJTNBJTIyNjYyZjFjYjYtNDA2ZC00OTRiLTlkOWEtMWM0M2EzZTAwZWUwJTIyJTdE) would need to work. I included it here to be part of that intersection.

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
dvadas: That's why I didn't specify it originally, but I was thinking that if we do support configuring the [maxConnectedClusters configuration](https://docs.cilium.io/en/latest/network/clustermesh/clustermesh/#scaling-limitations) in future, then that would require a new API release anyway to add the config knob to the ARM resource. For that new API version, we can update the doc. What do you think?

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/selector.tsp
nibooy: I think it's fine to use a share memberSelector model with validation enforced at the parent field level (i.e. UpdateStrategy, ClusterMeshProfile).

1. This is how k8s standard labelSelector behaves. LabelSelector can be set on various k8s objs (Deployment, NetworkPolicy), but each controller applies its own validation rules. 
2. The restriction shouldn't be a property of the memberSelector, it should be a property of the feature / parent field.
3. Having the generic model be the intersection of all requirements may not be scalable if we keep adding restrictions that are ANDed.
4. Defining selectors separately for each usage seems unnecessary and redundant.

nibooy (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2026-03-02-preview/fleets.json
serbrech: unlikely to increase in this preview version. so agreed

PR #40672, opened 2026-02-24
update JS config for elasticsan

kazrael2119:
disable JS breakings for elasticsan


qiaozha (approved):

qiaozha:
This PR is to add some backward compatible config for JS SDKs, I don't think ARM review is needed here, will add ARMSignedOff label to get unblocked. But let me know if anyone has any concerns

PR #40708, opened 2026-02-24
[Storage] `set_metadata` @override attempt TypeSpec Changes

vincenttran-msft:
First attempt was just a sparse override, ie. without re-defining all the parameters we wanted to carry over:

```typespec
// Override setMetadata operations to make metadata required for Rust
op containerSetMetadataOverride(
  /** The metadata headers. */
  @alternateType(Record<string>, "rust")
  @header("x-ms-meta")
  metadata: string,
): void;

@@override(Storage.Blob.Container.setMetadata,
  containerSetMetadataOverride,
  "rust"
);

op blobSetMetadataOverride(
  /** The metadata headers. */
  @alternateType(Record<string>, "rust")
  @header("x-ms-meta")
  metadata: string,
): void;

@@override(Storage.Blob.Blob.setMetadata, blobSetMetadataOverride, "rust");
```

But the compiler complained and code-gen exploded. So I went with the full definition, but even then, here you can see we explicitly define metadata as required:
```typespec
/** The metadata headers. */
  @header("x-ms-meta")
  metadata: string,
```

but still no dice.

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -172,12 +172,11 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 
 // Override setMetadata operations to make metadata required for Rust
 op containerSetMetadataOverride(
-  ...ApiVersionHeader,
-  ...ClientRequestIdHeader,
   ...TimeoutParameter,
   ...LeaseIdOptionalParameter,
 
   /** The metadata headers. */
+  @alternateType(Record<string>, "rust")
   @header("x-ms-meta")
   metadata: string,
 
@@ -190,11 +189,10 @@ op containerSetMetadataOverride(
 );
 
 op blobSetMetadataOverride(
-  ...ApiVersionHeader,
-  ...ClientRequestIdHeader,
   ...TimeoutParameter,
 
   /** The metadata headers. */
+  @alternateType(Record<string>, "rust")
   @header("x-ms-meta")
   metadata: string,
 
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -172,10 +172,16 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 
 // Override setMetadata operations to make metadata required for Rust
 op containerSetMetadataOverride(
+  ...ApiVersionHeader,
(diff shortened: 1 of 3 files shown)


specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -169,3 +169,47 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 @@scope(Storage.Blob.BlockBlob.query, "!rust");
 @@scope(Storage.Blob.PageBlob.copyIncremental, "!rust");
 @@scope(Storage.Blob.PageBlob.getPageRangesDiff, "!rust");
+
+// Override setMetadata operations to make metadata required for Rust
+op containerSetMetadataOverride(
```
jhendrixMSFT, line 174: I thought that it was due to the operations not being 100% identical.  So, I changed it like this but that didn't work.
```ts
#suppress "@azure-tools/typespec-azure-core/no-response-body" "Existing API"
#suppress "@azure-tools/typespec-azure-core/use-standard-names" "Existing API"
@put
@route("?restype=container&comp=metadata")
op containerSetMetadataOverride is StorageOperationNoBody<
  {
    ...TimeoutParameter;
    ...LeaseIdOptionalParameter;

    /** The metadata headers. */
    @alternateType(Record<string>, "rust")
    @header("x-ms-meta")
    metadata: string,

    ...IfModifiedSinceParameter;
  },
  {
    ...EtagResponseHeaderPrivate;
    ...LastModifiedResponseHeaderPrivate;
  }
>;
```

specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -169,3 +169,47 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
...
+
+  /** The metadata headers. */
+  @alternateType(Record<string>, "rust")
+  @header("x-ms-meta")
+  metadata: string,
+
+  ...IfModifiedSinceParameter,
+): void;
+
+@@override(Storage.Blob.Container.setMetadata,
```
jhendrixMSFT, line 188: Then I thought it was due to the override target being incorrect.  I also changed this but that didn't work.
```ts
@@override(BlobContainerClient.setMetadata,
  containerSetMetadataOverride,
  "rust"
);
```

vincenttran-msft:
Closing as there is a working proof-of-concept out shared by Joel.

PR #40709, opened 2026-02-24
[Draft] Agent Invocations API Specification

RaviPidaparthi:
## Summary

This PR introduces the **Agent Invocations API**, enabling agents to define custom request/response contracts while the AgentsService handles routing. This design provides complete container autonomy for invocation handling while maintaining a simple, transparent proxy architecture.

### New Endpoints

1. **GET `/agents/{agent_name}/versions/{agent_version}/invocations/docs/openapi.json`**
   - Returns the OpenAPI 3.x specification for the agent's invocation contract if the container exposes it at /docs/openai.json
   - Returns `404` if the container doesn't expose an OpenAPI spec
   - Enables type-safe client generation and API discovery

2. **POST `/agents/{agent_name}/versions/{agent_version}/invocations`**
   - Creates a new invocation for the specified agent version
   - Forwards request/response as raw bytes (transparent proxy pattern) to container's /invocation route
   - Container controls HTTP status codes, content types, and response format
   - Optional `agent_session_id` query parameter for grouping related invocations
   - Returns invocation metadata in response headers:
     - `x-agent-invocation-id`: Unique identifier generated per invocation
     - `x-agent-session-id`: Session ID if provided in request otherwise auto generated
     
3. **GET `/agents/{agent_name}/versions/{agent_version}/invocations/{invocation_id}`**
   - Retrieves a specific invocation
   - Forwards response as raw bytes from the agent container
   - Container controls the response format (status, result, metadata, etc.)
   - Enables polling for async/long-running invocations

4. **POST `/agents/{agent_name}/versions/{agent_version}/invocations/{invocation_id}:cancel`**
   - Cancels a running invocation
   - Optional request body forwarded as raw bytes to the container
   - Container controls cancellation logic and response format
   - Follows Azure action operation pattern (`:cancel` suffix)

### Transparent Proxy Pattern

The AgentsService acts as a **transparent byte-forwarding proxy**:
- ✅ No parsing or validation of request/response bodies
- ✅ Container has full control over status codes and content
- ✅ Minimal latency overhead
- ✅ Container-defined contracts (via optional OpenAPI spec)

### 3-Layer Responsibility Model

```
┌──────────────────────┐
│   AgentsService      │ → Routes requests, forwards bytes
├──────────────────────┤
│ Container Core       │ → HTTP server, invocation tracking, observability
├──────────────────────┤
│ Developer Code       │ → Business logic, validation, processing
└──────────────────────┘
```

## Feature Flag

Both endpoints are gated behind the preview feature flag:
- **Feature Key**: `AgentDefinitionFeatureKeys.hosted_agents_v1_preview`
- **Header Value**: `Foundry-Features: HostedAgents=V1Preview`
- **Required**: Yes (header must be present with exact value)

## Potential Implementation Patterns

The transparent proxy design enables containers to implement various execution patterns:

### 1. Synchronous (Immediate Response)
- **Pattern**: POST creates invocation and immediately returns result
- **Use Case**: Operations completing within 1 minute
- **Response**: Complete result in POST response body
- **GET endpoint**: Optional - can return cached result for retrieval/audit
- **Cancel endpoint**: Not needed

### 2. Long-Running Operations (LRO)
- **Pattern**: POST returns 202 Accepted, client polls GET for completion
- **Use Case**: Operations taking minutes to hours
- **Response**: Initial POST returns operation metadata, GET returns status/progress
- **GET endpoint**: Required for status polling
- **Cancel endpoint**: Optional - implement if cancellation is supported

### 3. Asynchronous with Callbacks
- **Pattern**: POST initiates operation, container calls back when complete
- **Use Case**: Operations taking minutes to hours
- **Response**: POST returns job ID, container sends webhook on completion
- **GET endpoint**: Optional - useful for manual status checks
- **Cancel endpoint**: Optional - implement if cancellation is supported

### 4. Streaming (Server-Sent Events / SSE)
- **Pattern**: POST returns streaming response with `Content-Type: text/event-stream`
- **Use Case**: Real-time incremental results, token-by-token generation
- **Response**: Stream of events until completion
- **GET endpoint**: Optional - can retrieve final aggregated result
- **Cancel endpoint**: Optional - client can close stream connection

### 5. Polling with Exponential Backoff
- **Pattern**: POST initiates, client polls GET with increasing intervals
- **Use Case**: Unpredictable completion times
- **Response**: GET returns status (pending/running/completed/failed)
- **GET endpoint**: Required for polling
- **Cancel endpoint**: Optional - implement if cancellation is supported
- **Best Practice**: Container returns `Retry-After` header

### Pattern Comparison Table

| Pattern | POST Response | GET Usage | Cancellation | Best For |
|---------|--------------|-----------|--------------|----------|
| **Sync** | Complete result | Audit/cache | N/A | < 1 min operations |
| **LRO** | 202 + metadata | Poll status | Supported | Minutes to hours |
| **Async + Callback** | Job ID | Status check | Supported | Minutes to hours |
| **SSE Streaming** | Event stream | Final result | Close stream | Real-time updates |
| **Polling** | 202 + location | Poll with backoff | Supported | Variable duration |


specification/ai-foundry/data-plane/Foundry/src/invoke/models.tsp
```diff
@@ -0,0 +1,324 @@
...
+  image_png: "image/png",
+
+  @doc("JPEG image.")
+  image_jpeg: "image/jpeg",
+
+  @doc("WAV audio.")
+  audio_wav: "audio/wav",
+
+  @doc("PDF document.")
+  application_pdf: "application/pdf",
```
RaviPidaparthi: Interrupts are also content items
`application/vnd.microsoft.agents.invoke.interrupt`

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -12,6 +12,8 @@ tags:
   - name: Evaluators
   - name: Indexes
   - name: Insights
+  - name: Invoke
+  - name: Health
```
rama0120: Expected this Health endpoint to be on the container only, not public service..

RaviPidaparthi: Good point.. I am sure a lot fell trough the cracks during copy paste.. Will need do a 2nd pass and cleanup

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -5125,6 +5186,19 @@ components:
...
+        - name
+        - version
+      properties:
+        name:
+          type: string
+          description: Agent name.
+        version:
+          type: string
+          description: Agent version.
+      description: Reference to the agent being invoked.
```
rama0120: I know this was inherited from my pr, which was inherited from @zhiyong-gayang 's PR :) - is there a reason to have this vs AgentReference object (https://learn.microsoft.com/en-us/azure/ai-foundry/reference/foundry-project?view=foundry#agentreference)

RaviPidaparthi: Yes. Should unify and snap to AgentReference

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -8376,6 +8481,36 @@ components:
...
+        item_type:
+          type: string
+          enum:
+            - file
+        file_name:
+          type: string
+          description: Suggested filename for the file (commonly used in output).
+        file_id:
+          type: string
+          description: Platform-managed file reference (preferred for large files).
```
rama0120: we need corresponding file-management apis for these yes? @ankitsul had some proposals for this that we should look at tomorrow

rama0120: Until we have that we should either exclude these OR possibly throw server error as "not implemented" effectively?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -8525,6 +8660,33 @@ components:
...
+      required:
+        - item_type
+      properties:
+        item_type:
+          type: string
+          enum:
+            - image
+        file_id:
+          type: string
+          description: Platform-managed file reference (preferred for large images).
```
rama0120: similar comment as above

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
rama0120: Also, should we support file_name and file_id? Why don't we do that for images below?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -8801,6 +8963,114 @@ components:
...
+            - Array of messages: For multi-turn conversations or structured inputs
+
+            Examples:
+            - String: "Hello, agent!"
+            - Messages: [{"content": "Hello!"}]
+            - With interrupt resume: [{"content": [{"item_type": "interrupt_resume", "interrupt_id": "int_123", "payload": {...}}]}]
+        session_id:
+          type: string
+          description: Session identifier for multi-turn conversations. Omit to create a new session.
+        params:
```
rama0120: Can we call this "customData" (symmetric for input and output) or "passThrough" / "payload" (again symmetric) ?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -23134,6 +23404,17 @@ components:
...
       description: Configuration for Responsible AI (RAI) content filtering and safety features.
+    ReadinessResponse:
+      type: object
+      required:
+        - status
+      properties:
+        status:
+          type: string
+          enum:
+            - ok
```
rama0120: we'd want a few more statuses here, we can iterate on it - but again dont expect this to be in the public api  surface

specification/ai-foundry/data-plane/Foundry/openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml
```diff
@@ -25183,6 +25464,22 @@ components:
           type: string
           description: Sku tier
       description: Sku information
+    StructuredDataContentItem:
```
rama0120: what is this for?

rama0120 (commented):
Doing review on the other example files etc - submitting first-cut review for the api.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
rama0120: or possibly "customContent"

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/invoke-api-examples.md
```diff
@@ -0,0 +1,731 @@
...
+
+```http
+POST /invoke HTTP/1.1
+Content-Type: application/json
+
+{
+  "agent_def": {
+    "name": "customer-support-bot",
+    "version": "1"
+  },
```
rama0120: Just as a thought experiment - would it be possible to pull this up to the URI path in the POST request instead of body like this? I assume some of these were inherited from the fact that we started with RAPI and even prompt agents don't have a per-agent-path to call against..

rama0120: ideally it would be something like POST {project-endpoint}/agent/customer-support-bot/version/1/invoke

specification/ai-foundry/data-plane/Foundry/src/invoke/models.tsp
```diff
@@ -0,0 +1,404 @@
...
+@doc("Structured error information for agent-level failures in the terminal event.")
+model InvocationError {
+  @doc("Machine-readable error code (e.g., 'container_timeout', 'agent_exception', 'validation_error').")
+  code: string;
+
+  @doc("Human-readable error message describing the failure.")
+  message: string;
+
+  @doc("Parameter name that caused the error, if applicable.")
+  param?: string;
```
rama0120: Nit: this seems too specific and purpose isn't super clear.. I'd remove it.. Could we pass any traceId back to help debug this better on our end? Or is that the requestId / e2e traceId that's sufficient?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/invoke-api-examples.md
```diff
@@ -0,0 +1,731 @@
...
+        "item_type": "text",
+        "text": "Key findings from Q4 2024:\n- Revenue increased 15% YoY\n- Customer acquisition up 23%\n- Operating margin improved to 18%"
+      },
+      {
+        "item_type": "structured_data",
+        "data": {
+          "pages_processed": 45,
+          "processing_time_ms": 3420
+        }
+      }
```
rama0120: nit: I'd leave this in customContent type?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/invoke-api-examples.md
```diff
@@ -0,0 +1,731 @@
...
+        "file_name": "sales-trend-chart.png",
+        "content_type": "image/png"
+      },
+      {
+        "item_type": "structured_data",
+        "data": {
+          "records_analyzed": 12450,
+          "chart_generated": true
+        }
+      }
```
rama0120: seeing these examples I'm more convinced this needs to be customContent.. itemType can just be "custom" and the payload should have that type @... field I had in my original PR. Thoughts?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/invoke-api-examples.md
```diff
@@ -0,0 +1,731 @@
...
+        "interrupt_id": "int_approve789",
+        "interrupt_type": "approval",
+        "details": {
+          "flight": "UA 1234",
+          "departure": "SEA 08:00 AM",
+          "arrival": "JFK 04:30 PM",
+          "price": "$450",
+          "date": "2025-03-15"
+        },
+        "message": "Please confirm to proceed with this booking."
```
rama0120: wait, this looks like it'll be a bit too limiting. Why not just let it be a custom blob with the type annotation?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/invoke-api-examples.md
```diff
@@ -0,0 +1,731 @@
...
+
+## SSE Event Types Reference
+
+Quick reference for all SSE event types in streaming responses:
+
+| Event Type               | Purpose                                  | Key Fields                                                |
+| ------------------------ | ---------------------------------------- | --------------------------------------------------------- |
+| `invoke.created`         | Invocation started                       | `seq`, `session_id`, `created_at`                         |
+| `output.content.delta`   | Incremental content update (merge)       | `seq`, `index`, `item_type`, `delta`, `final`             |
+| `output.content.final`   | Complete content item (replace)          | `seq`, `index`, `item`                                    |
```
rama0120: nit: how about content.delta vs output.content.delta?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/rapi-to-invoke-mapping-examples.md
```diff
@@ -0,0 +1,1648 @@
...
+| `output.item.final` | `output.content.final` | Complete atomic item |
+| `response.completed` | `invoke.completed` | Terminal event |
+
+---
+
+## Notes
+
+- RAPI is designed to be OpenAI-compatible, following the Responses API specification
+- When RAPI receives a request for a hosted agent, it internally calls the Invoke API
+- The mapping layer handles conversion between RAPI's message-based structure and Invoke's content-based structure
```
rama0120: I'm a bit worried / skeptical about this. Why should we do this vs just let Invoke API be its own vertical path down and RAPI be its own vertical path down? No conversion from RAPI to Invoke? I'm worried the "tail will wag the dog" and eventually force us to keep making updates to Invoke - which will impact customers even if they don't care about RAPI. I heard of one such example where RAPI seemingly some optional field was made mandatory and it broke us (iirc @vangarp said logprobs?).. Were there other such occasions? Simship won't stop for Agents service, so I'm worried we will find ourselves in an impossible position sooner than later if we do this mapping to Invoke.

specification/ai-foundry/data-plane/Foundry/src/invoke/models.tsp
```diff
@@ -0,0 +1,404 @@
...
+  text: "text",
+
+  @doc("Image content")
+  image: "image",
+
+  @doc("Document content")
+  file: "file",
+
+  @doc("Structured data content")
+  structured_data: "structured_data",
```
rama0120: Suggest renaming this to "customItem" or something - I thought it was mandating some specific shape when I read it (or thought it was some field to represent/pass csv data etc)

specification/ai-foundry/data-plane/Foundry/src/invoke/models.tsp
```diff
@@ -0,0 +1,404 @@
...
+  1. Delta event: {"text": "Hello"} → content becomes "Hello"
+  2. Delta event: {"text": " World", "final": true} → content becomes "Hello World" and is complete
+  
+  The final flag on any delta event signals that no more deltas will be sent for this item_index.
+  """)
+model ContentDeltaEvent {
+  ...InvocationEvent;
+
+  @doc("The zero-based index of the content item in the output message.")
+  index: int64;
```
rama0120: this is generated by container to track the index of the delta event (to handle reordering etc) right?

specification/ai-foundry/data-plane/Foundry/src/invoke/docs/rapi-to-invoke-mapping-examples.md
```diff
@@ -0,0 +1,1648 @@
...
+| `conversation_id` | `session_id` | Session derived from hash(agent_name, agent_version, conversation_id) |
+| `structured_inputs` | `params` | Custom parameters for agent |
+| `structured_outputs` output item | `structured_data` content item | Structured data in response |
+| `function_call` (name: "handle_interrupt") | `interrupt` content item | HITL pause/approval requests |
+| `function_call_output` | `interrupt_resume` content item | Resume after interrupt |
+| `message` items | `content` items | Message content |
+
+### Important Notes
+
+- **Session ID Mapping**: The `session_id` in Invoke is not a direct copy of `conversation_id`. Instead, it's computed as a hash of the agent name, agent version, and conversation_id. This ensures that:
```
splusq: what happens to conversation-less RAPI requests?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,138 @@ paths:
...
+          schema:
+            type: string
+        - name: agent_session_id
+          in: query
+          required: false
+          description: Optional session ID to group related invocations.
+          schema:
+            type: string
+          explode: false
+        - name: Foundry-Features
```
ankitsul, line 559: Just curious - will this be seen by customer or later removed when we go PuPr/GA?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,138 @@ paths:
                 $ref: '#/components/schemas/ApiErrorResponse'
       tags:
         - Agents
+  /agents/{agent_name}/versions/{agent_version}/invocations:
```
ankitsul, line 535: I am sure this is a conscious choice, but not `invoke`? That seems more intuitive naming wise

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,138 @@ paths:
...
+      responses:
+        '200':
+          description: The request has succeeded.
+          headers:
+            x-agent-invocation-id:
+              required: true
+              description: Unique identifier for this invocation, generated by the container.
+              schema:
+                type: string
+            x-agent-session-id:
```
ankitsul, line 583: Question - the reason we are returning as part of response header is because we want the blob-in/blob-out in body? Asking because in `CreateResponse` - this is accepted/returned as part of request/response body, so wondering if we need to have that parity.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,138 @@ paths:
...
+      tags:
+        - Agent Invocations
+      requestBody:
+        required: true
+        content:
+          '*/*':
+            schema:
+              type: string
+              format: binary
+        description: The raw request body to forward to the agent container.
```
ankitsul: nit: Are we consciously going with term `container`? Wondering if `sandbox` or `session` makes more sense since we expose those concept (at least `session`) as part of our API?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,264 @@ paths:
...
+      x-ms-foundry-meta:
+        required_previews:
+          - HostedAgents=V1Preview
+  /agents/{agent_name}/versions/{agent_version}/invocations/docs/openapi.json:
```
ankitsul, line 612: Can you clarify on the use-case to have OpenAPI specific API? I see the generic `get` is below.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: Resource oriented APIs are generally more preferred in Rest APIs I.e we prefer to have nouns (resource type) in the url paths and the path typically is a resource hierarchy. The http method then becomes a verb. 
So it becomes natural to say POST/PUT resource is create/update, Delete maps to delete, Get is read etc..

Obviously there are exceptions to this. An example is cancel which is not really a crud on resource but a real operation on a resource that changes its state. These don’t fit into that model and it is ok to deviate where it makes sense.

‘Hence the preference to use Invocation.. Now we can use http verbs to model crud for invocation resource.
It also maps well with how rest of our APIs look. Even OpenAI responses also follow the same model.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: RAPI is well known/defined spec. So they can do that,
However we want to enable developers bringing their own APIs and we just treat their payload as pass-through byte in/out. In fact developer can make his invoke API a plain text in/out in which case it’s not even json.

the only options we have is to return in headers. And on input side we can also use uri params or query params.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: Agreed that we need unified taxonomy. I am not sure what it is yet.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: PR description explains it.

Returns the OpenAPI 3.x specification for the agent's invocation contract if the container exposes it at /docs/openai.json
Returns 404 if the container doesn't expose an OpenAPI spec
Enables type-safe client generation and API discovery

RaviPidaparthi: Given that user brings his own wire format, we suggest that he also expose the api spec for his wireline protocol so that richer clients, UX, etc can be written around it.

‘our own playground can also use this for a better/richer experience. Laxmi was also exploring if there is a way to hook evals somehow.

specification/ai-foundry/data-plane/Foundry/src/agents-invocations/routes.tsp
```diff
@@ -0,0 +1,141 @@
...
+    }
+  >;
+
+  /**
+   * Cancels an invocation.
+   * The request and response bodies are forwarded as raw bytes to/from the agent container.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" ""
+  @post
+  @route("/agents/{agent_name}/versions/{agent_version}/invocations/{invocation_id}:cancel")
```
RaviPidaparthi: I intended this to be /cancel. Will fix

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: The header is mandatory as long as we are in PuPr. We will remove it when we GA.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -531,6 +532,264 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApiErrorResponse'
+      tags:
+        - Agent Invocations
+      x-ms-foundry-meta:
+        required_previews:
+          - HostedAgents=V1Preview
+  /agents/{agent_name}/versions/{agent_version}/invocations/{invocation_id}:cancel:
```
zhiyong-gayang: How customer config if they want to run in sync or background mode?

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
rama0120: i would just say "forward to the agent session" - no sandbox/container etc etc

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
rama0120, line 612: and the nice part of this "docs" path is we could in theory add more optional formats for specifying spec in here.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
rama0120: it is largely up to their container to decide that. we don't see that payload

rama0120 (approved):

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: I added an `Implementation Patterns` section to the PR description to clarify this.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
RaviPidaparthi: I went simpler and just used `agent` (ex: if agent does not support). 
BTW I removed the technical details out of this like forward to container/agent etc.. as this is more user facing than dev facing

PR #40710, opened 2026-02-24
Migrate Storage Files to TypeSpec

catalinaperalta:


specification/storage/data-plane/FileStorage/routes.tsp
```diff
@@ -635,7 +635,7 @@ interface Share {
 
       /** The ACL for the share. */
       #suppress "@azure-tools/typespec-azure-core/request-body-problem" "Existing storage API."
-      @body shareAcl?: SignedIdentifier[];
+      @body shareAcl?: SignedIdentifiers;
```
catalinaperalta, line 543: did you test this change with the js emitter?

specification/storage/data-plane/FileStorage/routes.tsp
```diff
@@ -621,8 +621,7 @@ interface Share {
       ...LastModifiedResponseHeader;
 
       @body
-      @Xml.name("SignedIdentifiers")
-      signedIdentifiers: SignedIdentifier[];
+      signedIdentifiers: SignedIdentifiers;
```
catalinaperalta, line 522: @jeremymeng what was the issue with the original definition?

jeremymeng: before this change, in JS generated code, the body wasn't serialized as XML, but JSON.

specification/storage/data-plane/FileStorage/routes.tsp
jeremymeng: Yes, although I also expected that you may also apply additional changes similar to the Blob ones.

specification/storage/data-plane/FileStorage/models.tsp
```diff
@@ -1735,19 +1735,23 @@ alias FileSmtPropertyResponseHeaders = {
...
   @encode("rfc7231")
   @alternateType(string, "python")
+  @alternateType(utcDateTime, "go")
   fileLastWriteTime?: utcDateTime;
 
   /** Change time for the file/directory. */
   @header("x-ms-file-change-time")
   @encode("rfc7231")
   @alternateType(string, "python")
+  @alternateType(utcDateTime, "go")
```
catalinaperalta: @gracewilcox I'm removing these alternateType instances since they are redundant to the original base type. Curious to hear what the issue was that caused this

specification/storage/data-plane/FileStorage/models.tsp
```diff
@@ -1735,19 +1735,23 @@ alias FileSmtPropertyResponseHeaders = {
   /** Creation time for the file/directory. */
   @header("x-ms-file-creation-time")
   @encode("rfc7231")
+  @encode("iso8601")
```
catalinaperalta: @gracewilcox I removed this since types can only have 1 encoding, unless this is the encoding that all languages use, this should be fixed using an alternateType with an alias.

specification/storage/data-plane/FileStorage/client.tsp
```diff
@@ -279,21 +279,64 @@ model DirectoryClientParameters {
   "csharp"
 );
 
+// Defining an override for backward compatibility for C# since the XML payload was serialized with the
+// elements/properties in a different order than defined in the swagger.
+
```
catalinaperalta: @jorgerangel-msft I havent compared this against the original model BUT if this is ONLY about the order the props are defined in the tsp you can do that in the main file. Happy to discuss offline as needed!

specification/storage/data-plane/FileStorage/models.tsp
```diff
@@ -0,0 +1,2486 @@
...
+alias FileCreationTimeParameter<TDoc extends valueof string> = {
+  #suppress "@azure-tools/typespec-azure-core/known-encoding" "Storage supports 7 digit precision"
+  #suppress "@azure-tools/typespec-autorest/unknown-format" "Storage supports 7 digit precision"
+  @doc(TDoc)
+  @encode("rfc3339-fixed-width")
+  @header("x-ms-file-creation-time")
+  fileCreationTime?: utcDateTime;
+};
+
+/** File creation time header (string type, for rename/copy/symlink operations). */
```
catalinaperalta, line 1999: @jalauzon-msft @amnguye This comment applies to these headers and the ones below up to line 2452. The swagger defined these headers inline for responses with the format: `date-time-1123` and a couple of other definitions, where one had the same date-time-1123 format and another one that was just left as string (example swagger definitions: [inline](https://github.com/Azure/azure-rest-api-specs/blob/e51bb2d93bd3dfa1d47f3dc169696908cacd61d2/specification/storage/data-plane/Microsoft.FileStorage/stable/2026-04-06/file.json#L2400), [definition 1](https://github.com/Azure/azure-rest-api-specs/blob/e51bb2d93bd3dfa1d47f3dc169696908cacd61d2/specification/storage/data-plane/Microsoft.FileStorage/stable/2026-04-06/file.json#L7997-L8008), [definition 2](https://github.com/Azure/azure-rest-api-specs/blob/e51bb2d93bd3dfa1d47f3dc169696908cacd61d2/specification/storage/data-plane/Microsoft.FileStorage/stable/2026-04-06/file.json#L8242-L8271)). However, it seems all of these headers are returned/sent in rfc 3339 (iso 8601) with the 7-digit precision according to what I'm seeing in the test assets, so if this is most accurate then the `@encode("rfc3339-fixed-width")` is correct for all of them:



Request:

[https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/…](https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/storage/azure-storage-file-share/tests/recordings/test_directory.pyTestStorageDirectorytest_create_directory_set_smb_properties.json#L41-L43)

Response:

[https://github.com/Azure/azure-sdk-assets/blob/f65730ef4c99d1f709cf9d2b96b100513c921a6d/go/sdk/stor…](https://github.com/Azure/azure-sdk-assets/blob/f65730ef4c99d1f709cf9d2b96b100513c921a6d/go/sdk/storage/azfile/testdata/recordings/Test/TestCreateFileNFS.json#L60-L61)
[https://github.com/Azure/azure-sdk-assets/blob/f65730ef4c99d1f709cf9d2b96b100513c921a6d/go/sdk/stor…](https://github.com/Azure/azure-sdk-assets/blob/f65730ef4c99d1f709cf9d2b96b100513c921a6d/go/sdk/storage/azfile/testdata/recordings/Test/TestCreateFileNFS.json#L64)
[https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/…](https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/storage/azure-storage-file-share/tests/recordings/test_directory.pyTestStorageDirectorytest_create_directory_set_smb_properties.json#L58-L59)
⁠[https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/…](https://github.com/Azure/azure-sdk-assets/blob/4afd6de03353f521de3cf7b6a888d7c0bfbfc1d8/python/sdk/storage/azure-storage-file-share/tests/recordings/test_directory.pyTestStorageDirectorytest_set_directory_properties_with_file_permission_key.json#L54-L55)

amnguye: This looks good, I believe the iso 8601, rfs3339-fixed-width is correct for the formatting since it has to have the 7 digit precision, for the Share File Create Time, Last Write Time etc.

specification/storage/data-plane/FileStorage/routes.tsp
```diff
@@ -0,0 +1,1464 @@
...
+      ...LeaseIdOptionalParameter;
+      ...LeaseActionParameter<"break">;
+      ...FileRequestIntentParameter;
+      ...AllowTrailingDotParameter;
+    },
+    {
+      @statusCode statusCode: 202;
+      ...EtagResponseHeader;
+      ...LastModifiedResponseHeader;
+      ...LeaseTimeResponseHeader;
```
catalinaperalta, line 1059: @amnguye do you know if this operation returns this header? It wasnt in the original swagger, however the other break lease endpoint for this service does include it.

specification/storage/data-plane/FileStorage/routes.tsp
```diff
@@ -0,0 +1,1464 @@
...
+    {
+      ...TimeoutParameter;
+      ...MetadataHeaders;
+      ...AllowTrailingDotParameter;
+      ...FileRequestIntentParameter;
+    },
+    {
+      ...EtagResponseHeader;
+      ...RequestServerEncryptedResponseHeader;
+      ...LastModifiedResponseHeader;
```
catalinaperalta, line 687: @amnguye Share_setMetadata does specify the last modified header but File_SetMetadata and Directory_SetMetadata don't, I'm just wondering if the File and Directory set metadata do technically also return that header?

amnguye: Just checked this, on the response for both File_SetMetadata and Directory_SetMetadata, both Last Modified Times are returned.

specification/storage/data-plane/FileStorage/routes.tsp
amnguye: It actually does return this header, but only on a successful break of the lease.

specification/storage/data-plane/FileStorage/client.tsp
```diff
@@ -0,0 +1,554 @@
...
+@@access(AccessRight, Access.internal, "csharp");
+
+#suppress "@azure-tools/typespec-client-generator-core/client-option"
+@@clientOption(Storage.File.MetadataHeaders.metadata,
+  "collectionHeaderPrefix",
+  "x-ms-meta-",
+  "csharp"
+);
+
+// ===== Rust SDK customizations: required metadata overrides =====
```
jalauzon-msft: I would say we can leave this section out since we don't currently have a Rust Files SDK. We can add them back if and when we need too. Especially because they add a lot of lines to this file.

specification/storage/data-plane/FileStorage/models.tsp
```diff
@@ -0,0 +1,2474 @@
...
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+namespace Storage.File;
+
+/**
+ * Error codes returned by the service
```
jalauzon-msft: This file is a bit inconsistent on single line verses multi-line comments. We are not perfect in other specs either, but could we have Copilot go through and clean that up to prefer single line if it fits (like here for example).

specification/storage/data-plane/FileStorage/client.tsp
catalinaperalta: Removed the rust customizations

specification/storage/data-plane/FileStorage/models.tsp
catalinaperalta: Done, did a pass with copilot to standardize the comments :)

jalauzon-msft (approved):

PR #40719, opened 2026-02-25
Update mcp.json to pass PATHEXT env var to mcp env

praveenkuttappan:
Copilot CLI appears to sanitize the environment when spawning MCP by setting PATHEXT environment variable to only `.CPL` and a result MCP won't run command `git` or `npx` because Windows won't find the file resolved using PATHEXT. It will work fine if command is `git.exe`.

Solution is to pass known extensions as env var when bootstrapping the MCP and this matches with values passed when MCP is running on vscode.

Another change in this PR is to change `workspaceFolder` var(vscode specific) to use relative path to mcp startup script. This is required to start MCP server in copilot CLI and this works from vscode too.


mikeharder (approved):

.vscode/mcp.json
```diff
@@ -3,7 +3,10 @@
       "azure-sdk-mcp": {
         "type": "stdio",
         "command": "pwsh",        
-        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
-    },
+        "args": ["./eng/common/mcp/azure-sdk-mcp.ps1", "-Run"],        
+        "env": {
+          "PATHEXT": ".COM;.EXE;.BAT;.CMD;.JS;.MSC;.CPL"
```
praveenkuttappan:
```suggestion
          "PATHEXT": ".COM;.EXE;.BAT;.CMD;.MSC;.CPL"
```

benbp:
If the issue is with processes we invoke from the code, we should just update those to use the process options overload that specifies a separate windows command: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs#L56-L64

weshaggard:
> If the issue is with processes we invoke from the code, we should just update those to use the process options overload that specifies a separate windows command: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs#L56-L64

If my that you mean, we can override the PATHEXT variable directly in the process helper that might be better than needing to update all the mcp configs. I would like to stop from special casing all the windows specific calls if we can avoid it.

benbp (approved):

praveenkuttappan:
> > If the issue is with processes we invoke from the code, we should just update those to use the process options overload that specifies a separate windows command: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/Helpers/Process/Options/ProcessOptions.cs#L56-L64
> 
> If my that you mean, we can override the PATHEXT variable directly in the process helper that might be better than needing to update all the mcp configs. I would like to stop from special casing all the windows specific calls if we can avoid it.

@weshaggard  Nope. What Ben meant is to have MCP process caller passes windows command name. But we can try if this can be overriden within process helper inside MCP itself. I have discussed with Ben and we are planning to take current change as short term fix to unblock and later try permanent fix inside MCP server to handle it.

.vscode/mcp.json
```diff
@@ -3,7 +3,10 @@
       "azure-sdk-mcp": {
         "type": "stdio",
         "command": "pwsh",        
-        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
-    },
+        "args": ["./eng/common/mcp/azure-sdk-mcp.ps1", "-Run"],        
+        "env": {
+          "PATHEXT": ".COM;.EXE;.BAT;.CMD;.MSC;.CPL"
+        }
+    }
```
praveenkuttappan:
```suggestion
        "args": ["./eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]       
    }
```

.vscode/mcp.json
```diff
@@ -3,7 +3,7 @@
       "azure-sdk-mcp": {
         "type": "stdio",
         "command": "pwsh",        
-        "args": ["${workspaceFolder}/eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
-    },
+        "args": ["./eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
```
praveenkuttappan:
```suggestion
        "args": ["./eng/common/mcp/azure-sdk-mcp.ps1", "-Run"]
```

PR #40722, opened 2026-02-25
Update Monitor Exporter tsp config

pvaneck:
This is some configuration for csharp.


pvaneck:
FYI @mattsains-msft

specification/applicationinsights/data-plane/Exporters/client.tsp
mattsains-msft, line 19: does `@@clientName` also support access modifiers? If so, could you make TelemetryItem internal? Right now it's generating as public, which doesn't work because it references MonitorBase which is internal.

Also, I've noticed an issue where ApplicationInsightsClientOptions.ServiceVersion is internal but ApplicationInsightsClientOptions is public, which is an error because the public constructor of ApplicationInsightsClientOptions takes in a ServiceVersion value (public signature contains less visible parameter types)

rajkumar-rangaraj (approved):

rajkumar-rangaraj:
@pvaneck here is the PR for azure monitor csharp auto-generation - 
 https://github.com/Azure/azure-sdk-for-net/pull/56891

Changes to this PR looks good.

lmazuel (approved):

PR #40726, opened 2026-02-25
Change etag headers to use Azure.Core.etag 

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/40723

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/models.tsp
+++ b/specification/storage/Microsoft.BlobStorage/models.tsp
@@ -2031,7 +2031,7 @@ alias DateResponseHeaderPrivate = {
 /** The ETag response header */
 alias EtagResponseHeader = {
   /** The ETag contains a value that you can use to perform operations conditionally. */
-  @header("ETag") eTag: string;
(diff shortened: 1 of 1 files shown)


specification/storage/Microsoft.BlobStorage/models.tsp
catalinaperalta, line 2138: @jhendrixMSFT should these headers also change to the Azure.Core.etag ?

jhendrixMSFT: No, only the `etag`, `if-match`, and `if-not-match` headers are to be treated as Etags per the RFC.

catalinaperalta: Ok sounds good, I'll revert the changes on the if-tags headers then

vincenttran-msft (commented):
Change looks good to me, please verify with Amanda.

catalinaperalta:
@amnguye can you please help verify if all of the headers I'm updating in this PR are etags?

amnguye (approved):
Looks good!

PR #40739, opened 2026-02-25
feat: Add Hosted Agents ADC integration API spec changes

ankitsul:
## Problem
Implement the TypeSpec API spec changes to support Hosted Agents running on Azure Developer Compute (ADC).

### 1. AgentVersionObject — Add `status` and `error` fields
**File**: `agents/models.tsp`
- Add `AgentVersionStatus` union (`creating`, `active`, `failed`, `deleting`, `deleted`)
- Add `AgentVersionError` model (`code`, `message`, `details?`)
- Add optional `status` and `error` fields to `AgentVersionObject`

### 2. CreateResponse / Response — Add `agent_session_id`
**File**: `openai-responses/models.tsp`
- Add `agent_session_id?: string` to the `@@copyProperties(OpenAI.CreateResponse, ...)` block
- Add `agent_session_id?: string` to the `@@copyProperties(OpenAI.Response, ...)` block

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -159,6 +159,9 @@ union AgentVersionStatus {
 
   @doc("The agent version is being deleted.")
   deleting: "deleting",
+
+  @doc("The agent version has been deleted.")
+  deleted: "deleted",
 }
 
 @doc("Error details for agent version provisioning failure.")
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -126,6 +126,55 @@ model AgentVersionObject {
   created_at: utcDateTime;
 
   definition: AgentDefinition;
+
+  @doc("The provisioning status of the agent version. For hosted agents, reflects ADC snapshot readiness.")
+  @extension(
+    "x-ms-foundry-meta",
+    #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+  )
+  status?: AgentVersionStatus;
+
+  @doc("Error details if the agent version provisioning failed.")
+  @extension(
+    "x-ms-foundry-meta",
+    #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+  )
+  error?: AgentVersionError;
+}
+
+@doc("The provisioning status of an agent version.")
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+)
+union AgentVersionStatus {
+  @doc("The agent version is being provisioned (e.g., ADC snapshot creation in progress).")
+  creating: "creating",
+
+  @doc("The agent version is active and ready to serve requests.")
+  active: "active",
+
+  @doc("The agent version provisioning failed.")
+  failed: "failed",
+
+  @doc("The agent version is being deleted.")
+  deleting: "deleting",
+}
+
+@doc("Error details for agent version provisioning failure.")
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+)
+model AgentVersionError {
+  @doc("The error code.")
+  code: string;
+
+  @doc("A human-readable error message.")
+  message: string;
+
+  @doc("Additional error details.")
+  details?: string;
 }
 
 union AgentProtocol {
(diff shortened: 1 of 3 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -126,6 +126,58 @@ model AgentVersionObject {
   created_at: utcDateTime;
 
   definition: AgentDefinition;
+
+  @doc("The provisioning status of the agent version. For hosted agents, reflects ADC snapshot readiness.")
+  @extension(
```
RaviPidaparthi: Features gates are mainly to make sure that when user is invoking a preview feature he is explicitly opting into it on a GA surface.

In this case, there is no need for optin because status is returned by service and it will always return it.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -126,6 +126,58 @@ model AgentVersionObject {
...
+    "x-ms-foundry-meta",
+    #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+  )
+  status?: AgentVersionStatus;
```
RaviPidaparthi: Returning null status is wierd. Any reason we cant return `active` by default?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -126,6 +126,58 @@ model AgentVersionObject {
...
+  @doc("The agent version has been deleted.")
+  deleted: "deleted",
+}
+
+@doc("Error details for agent version provisioning failure.")
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+)
+model AgentVersionError {
```
RaviPidaparthi: We should avoid re-inventing errors as we don't want different error formats across different apis.

Normally in these cases I would go with azure error guidelines. 
https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors

However, our API is a bit strange as we also have a mix of OpenAI and azure patterns.
And we settled on this common format for our errors that is a super set of open ai errors and still is compatible with azure guidelines.
https://github.com/Azure/azure-rest-api-specs/blob/aeebfc467b23118a90f692e09c176bd3c75fecbe/specification/ai-foundry/data-plane/Foundry/src/common/models.tsp#L16

Here is its full structure.
```
{
  "error": {
    "code": "invalid_payload",
    "message": "Invalid payload",
    "param": null,  // json path or query param name etcc.
    "details": [{
      "code": "validation_error",
      "message": "Missing required value",
       "param": "agent_reference.name" // json path or query param name etcc.
    }],
    "additionalInfo": {
        // arbitrary key values
        
     },
     "debugInfo": {
        // arbitrary debug information. Today we use it in local developmet mode to return full exception stack traces.
     }
  }
```

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -126,6 +126,58 @@ model AgentVersionObject {
...
+    #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+  )
+  error?: AgentVersionError;
+}
+
+@doc("The provisioning status of an agent version.")
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionFeatureKeys.hosted_agents_v1_preview] }
+)
```
RaviPidaparthi: My general feedback is to remove all required_preview stuff you added. Every one of these are top level agent properties that have nothing to do with hosted agent itself.

ankitsul: But would it expose anything to customers till we do a public preview release?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
ankitsul: Yeah, can do that for non-hosted agent flow and `creating` for ADC hosted agent.

specification/ai-foundry/data-plane/Foundry/src/openai-responses/models.tsp
```diff
@@ -34,6 +34,9 @@ namespace Azure.AI.Projects {
 
       /** The structured inputs to the response that can participate in prompt template substitution or tool argument bindings. */
       structured_inputs?: Record<unknown>,
+
+      /** Optional session identifier for sandbox affinity. When provided, the request is routed to the same sandbox. When omitted, auto-derived from conversation_id/prev_response_id or a new UUID is generated. */
+      agent_session_id?: string,
```
RaviPidaparthi, line 44: Add a comment that says currently this field is relevant only for hosted agents?

RaviPidaparthi (approved):

trrwilson (approved):

PR #40756, opened 2026-02-25
[Toolset] Add API Contracts for Toolset CRUD

anushree1808:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/toolsets/routes.tsp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Azure.AI.Projects;
+
+@tag("Toolsets")
+interface Toolsets {
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
+  @doc("Create a toolset.")
+  @operationId("createToolset")
+  @post
+  @route("/toolset")
```
johanste: Collections segment should be plural (unless there is a specific reason to break that rule).  
```suggestion
  @route("/toolsets")
```

specification/ai-foundry/data-plane/Foundry/src/toolsets/models.tsp
```diff
@@ -0,0 +1,103 @@
+import "../common/service.tsp";
+import "../tools/models.tsp";
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+
+namespace Azure.AI.Projects;
+
+union ToolsetObjectType {
```
johanste, line 13: This union is never referenced anywhere - we are only referencing the individual members. Why is it needed and why don't we inline the references where it is being used?

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -77,6 +116,21 @@ namespace Azure.AI.Projects {
     instance_name: string;
   }
 
+  /**
+   * Optional name and description fields for tool definitions and nested tool configurations.
+   */
+  alias ToolNameAndDescriptionExtension = {
```
glecaros, line 123: Since this is an extension, I worry about these properties could result in a conflict with properties defined in a future update to the upstream (currently they do conflict with the ones defined in `CustomTool`). Should we namespace these somehow (prefix or an alternative)?

anushree1808: hi @glecaros , if OpenAI tool definition (upstream) already includes name and description fields, we will not add these properties to those definitions. If OpenAI updates the definition to include these fields, we can just rely on them and not use these.

The purpose of adding these fields is currently the spec doesn't include them already. As you can notice, I have only updated the class definition that didn't have these field.

specification/ai-foundry/data-plane/Foundry/src/toolsets/models.tsp
anushree1808: I am not familiar with Typespec or the conventions. I just followed existing definitions for memory stores: https://github.com/Azure/azure-rest-api-specs/blob/feature/foundry-v2-spec-folder-structure/specification/ai-foundry/data-plane/Foundry/src/memory-stores/models.tsp

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
glecaros: So, semantically they are the same, same type of use?

anushree1808: yes, correct. :)

anushree1808: the name field will be used to identify it in tool calling (model will see this as the unique identifier and if it decides to call this tool specific tool, it will use this name to set the recipient). the description field will be used to add additional context to model context, to improve tool calling accuracy and recall. We are using these fields for same purposes.

if customer wants to use more than one instance of each tool, for example, two Azure AI search configurations or two Sharepoint connections, they will provide name to distinguish the instances and provide description to add specific contexts about each connection.

dargilco:
To verify your change and formatting:
`npx tsp format **/*tsp`
`npx tsp compile .`

Then include the edited openapi3 in your PR

PR #40812, opened 2026-02-26
Rename submitBatch header property multipartContentType to contentType

jeremymeng:
for consistency with the other submitBatch operation


specification/storage/Microsoft.BlobStorage/routes.tsp
catalinaperalta, line 1: Shouldnt we just rename all of the properties to be contentType? Instead of multipartContentType?

catalinaperalta: I believe there are 2 other instances of that name

jeremymeng: yes, I can do that too.

jeremymeng: done

catalinaperalta (approved):
LGTM. Adding storage folks for review before merging

jalauzon-msft (approved):

Issue #40805, opened 2026-02-26
Skip GitHub agent workflow lock files from prettier check


praveenkuttappan:
Exclude GitHub agent workflow lock yml file from
- Prettier check
- Linter check
- cspell

PR #40813, opened 2026-02-26
Remove copilot token setup from sdk agent workflow

praveenkuttappan:
This PR has changs to remove special handling to fetch secrets from KV and persist as copilot_github_token. Based on the discussion with gh-aw team, copilot_github_token requirement is temporary, and it will not be required soon. So, we don't need a tooling step to auto sync copilot token.

This PR also removes an unnecessary step to run authentication on runner side.

Another change is to exclude lock.yml files from prettier check.


.github/.prettierignore
```diff
@@ -15,3 +15,6 @@ ISSUE_TEMPLATE
 policies
 prompts
 PULL_REQUEST_TEMPLATE
+
+# github agent workflow lock files
+.github/workflows/*.lock.yml
```
mikeharder: I think this should move to `.github/workflows/.prettierignore`.  You will also need to adjust the path to be relative to the ignore file, not relative to the repo root.

mikeharder (changes requested):
move prettierignore

.github/.prettierignore
praveenkuttappan: I have removed prettierignore changes from this PR. Prettier does not seem to be using relative ignore file. I will make that change in a separate PR.

praveenkuttappan: Issue filed for prettierignore check https://github.com/Azure/azure-rest-api-specs/issues/40805

mikeharder (approved):

Issue #40775, opened 2026-02-26
Update SignedIdentifiers definition


catalinaperalta:
EDITED: the current signed identifiers definition is mostly correct, we just need to remove the xml.name decorator over the items property and emitters will need to update their behavior.

catalinaperalta:
After discussing with the typespec team, closing since the definition is correct as is in the rust branch

PR #40822, opened 2026-02-27
Blob: Fix signedIdentifiers definition

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/40775

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/models.tsp
+++ b/specification/storage/Microsoft.BlobStorage/models.tsp
@@ -424,7 +424,6 @@ model BlockLookupList {
 model SignedIdentifiers {
   /** The array of signed identifiers. */
   @Xml.unwrapped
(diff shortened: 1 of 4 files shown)


specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -421,12 +421,7 @@ model BlockLookupList {
 }
 
 /** Represents an array of signed identifiers */
-model SignedIdentifiers {
-  /** The array of signed identifiers. */
-  @Xml.unwrapped
-  @Xml.name("SignedIdentifier")
```
catalinaperalta, line 456: remove

catalinaperalta:
After discussing with the typespec team we determined that the definition in the rust branch is ok as is

PR #40830, opened 2026-02-27
Sync main error definition with blobs rust branch

catalinaperalta:
The error definition is missing other properties and headers. Seems something in the blob-tsp-rust branch is out of sync with main. 

Fixes https://github.com/Azure/azure-rest-api-specs/issues/40772


vincenttran-msft (approved):
LGTM, we stripped this out with no adverse effects, so no issues adding the def back if needed elsewhere.

Issue #40773, opened 2026-02-26
Blobs: revert immutability policy enum change


catalinaperalta:
The input/output enum values are different, therefore 2 enums are needed

catalinaperalta:
Decision was made to keep the enum unified, any language that had 2 enums will use client customizations to separate.

PR #40833, opened 2026-02-27
Revert ImmutabilityPolicyMode enum change

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/40773


catalinaperalta:
Closing. Per offline discussion we will be keeping the definition with 1 unified enum for request/responses.

PR #40838, opened 2026-02-27
[widget] add compute composite service + TSV changes

mikeharder:


specification/widget/resource-manager/Microsoft.Widget/Compute/Disk/suppressions.yaml
```diff
@@ -1,4 +1,9 @@
 - tool: TypeSpecValidation
   rules: [FolderStructure]
-  reason: composite spec
   paths: ["."]
+  reason: composite spec
+
+- tool: TypeSpecValidation
+  rules: [SdkTspConfigValidation]
```
qiaozha, line 7: what does it report if we don't suppress this?

PR #40840, opened 2026-02-27
update JS configs for existing self-serivce PRs

kazrael2119:
update JS configs for existing self-serivce PRs to avoid breakings


qiaozha (approved):

qiaozha:
Added ARMSignedOff label as this PR is trying to mitigate the JS breaking changes.

PR #40876, opened 2026-02-27
Resolve ARM review comments on Offerings APIs

yangylu91:
This PR is to handle comments on Offerings API at https://github.com/Azure/azure-rest-api-specs/pull/40408


yangylu91:
Close this proposal PR as it misunderstood the ARM review and was not the correct solution.

PR #40897, opened 2026-03-01
Add SDK failure reason as a comment from SDK generation agent workflow

praveenkuttappan:
This PR adds an instruction to run pipeline analysis if SDK generation fails and add the failure summary as a comment in GitHub issue to make it easier for user to find the failure reason.


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/14245

weshaggard (approved):

.github/workflows/sdk-generation-agent.md
```diff
@@ -120,8 +120,29 @@ When validation succeeds, execute the following steps in order.
...
+
+1. List each language whose SDK generation pipeline failed.
+2. For each failed language, include the full, non-redacted pipeline run URL so the user can inspect logs directly.
+3. For each failed language, run the pipeline analysis command **once per language** to identify the root cause:
+   - Execute `azsdk azp analyze <build-id> -a false` where `<build-id>` is the pipeline build ID for that language.
+   - Run this command only once per failed language — do not retry or re-run the analysis for the same build.
+   - Review the analysis output and extract a concise summary of the core error (e.g. compilation error, missing dependency, TypeSpec validation failure).
+   - Include the error summary under each failed language in the issue comment so users can understand the failure without inspecting pipeline logs.
+
+4. Include the following troubleshooting guidance in the comment:
```
mikeharder, line 135: ideally we'd update existing comments instead of posting new, we have JS code to do this from code-based workflows, but it sounds like agentic workflows are currently blocked from updating comments (though they can add new ones)

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/comment.js

PR #40904, opened 2026-03-02
fix: use boolean type for includeExtendedLocations

CustardTart32:
This PR addresses some comments regarding query param types in the List VM Skus API.


FumingZhang:
Could you also update the examples so that the parameter value is changed from `"true"` to `true` to match the type change?

FumingZhang (approved):

PR #38668, opened 2025-11-07
Refactor(keyvault): migrate to unified folder structure

v-xuto:
This PR migrates the **keyvault** specification to the new unified folder structure according to Azure folder structure guidelines.

**Change:**
- Consolidated TypeSpec and Swagger files under 
```
├── data-plane                   
│   ├── Administration            
│   │   ├── BackupRestore
│   │   ├── RBAC
│   │   └── Settings   
│   ├── Certificates             
│   ├── Keys                     
│   ├── Secrets                   
│   ├── SecretsStorage            
│   ├── SecurityDomain            
│   └── Legacy       # Old api versions
│
└── resource-manager/Microsoft.KeyVault/KeyVault          
```
- Updated `tspconfig.yaml` with correct paths and ARM types directory
- Fixed all common-types references and input-file paths in `readme.md`
- Added `suppressions.yaml` for legacy API versions
- Verified TypeSpec compilation works correctly
- Updated the relevant paths in `specification/suppressions.yaml` file
- Updated the relevant paths in `cspell.yaml` file

**Benefits:**
- Unified structure for better maintainability
- Aligns with Azure folder structure standards
- Maintains backward compatibility
- Supports future TypeSpec migration efforts

PR #40918, opened 2026-03-02
[SDK Validation]: handle rename scenario gracefully in findAllParentsWithFile to prevent crash

raych1:
## Problem

`FindAllParentsWithFile` in `\eng/tools/spec-gen-sdk-runner/src/utils.ts` re-throws ENOENT errors when scanning directories that no longer exist on disk. When a PR renames/restructures folder paths (e.g. [keyvault folder migration PR #38668](https://github.com/Azure/azure-rest-api-specs/pull/38668)), \git diff\ reports both old and new paths. The old paths don't exist after checkout, and for paths containing `data-plane` or `
resource-manager`, the code crashes the entire Node.js process with an uncaught exception — preventing any SDK generation from running.

## Fix

1. `FindAllParentsWithFile`: Changed to match `FindParentWithFile` behavior — catch errors, log a warning, and return early instead of re-throwing.
2. `main()` in `index.ts`: Added top-level try-catch so unexpected exceptions properly set the pipeline step result to `Failed`.

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/index.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/index.ts
@@ -22,12 +22,18 @@ export async function main() {
     mkdirSync(logFolder, { recursive: true });
   }
   let statusCode: number;
-  if (batchType) {
-    statusCode = await generateSdkForBatchSpecs(batchType);
-  } else if (pullRequestNumber) {
-    statusCode = await generateSdkForSpecPr();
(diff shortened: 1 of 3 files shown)


mikeharder (approved):
I think we can handle this a bit more gracefully when we create the shared helper, but this is fine for now.

- https://github.com/Azure/azure-rest-api-specs/issues/40768

PR #40922, opened 2026-03-02
fix(containerregistry): adjust csharp emitter to align with sdk naming conventions

CodeJourneys:
Currently, the generated .NET management SDK package for Azure.ResourceManager.ContainerRegistryTasks produces model names that violate the Azure SDK C# naming analyzers AZC0030 and AZC0031. This PR adds csharp client naming guidance to correct known violations


northtyphoon (approved):
LGTM

PR #40935, opened 2026-03-02
remove "?comp=copy" route for copyFromUrl operation

jeremymeng:
Since the REST API docs don't have it:

https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url

and including it in the request url causes unexpected 400 response with
`MissingRequiredQueryParameter` error code and `copyid` query parameter name

Fixes #40937


jalauzon-msft (approved):

catalinaperalta:
@jeremymeng can you resolve the merge conflicts before I merge this in?

catalinaperalta (approved):

PR #40940, opened 2026-03-03
Set ignore-nullable-on-optional to false for JS for impacted services

JialinHuang803:
Related issue: https://github.com/Azure/autorest.typescript/issues/3724

The JS emitter will not generate `| null` for nullable optional properties. Set this flag to false to keep `| null` being generated so there won't be relevant breaking changes.


MaryGao (approved):
So we only have two stable SDKs impacted?

JialinHuang803:
> So we only have two stable SDKs impacted?

Yes. The previous report generated by AI was not that correct, so eventually I found only two sdks impacted..

maorleger (approved):

psah434:
sdk validation test is failing. please fix.

JialinHuang803:
@psah434 The validation test passed and overall looks good to me. Could you help review? Thanks!

PR #40942, opened 2026-03-03
Update tspconfig.yaml to remove generate-sample

kazrael2119:
Removed generate-sample option for @azure-tools/typespec-ts.


kazrael2119:
@NneOkeke 
from this [PR](https://github.com/Azure/azure-rest-api-specs/pull/39646),  I want to confirm with you why you added `generate-sample: false` in [typespec-ts config](https://github.com/Azure/azure-rest-api-specs/pull/39646/changes#diff-0a3f6e3b6890d57d17578223821dfa2abd9a73084975a1c92fb1f70b6478056fR19)?

cc @MaryGao

MaryGao (approved):
Generally we would enable sample generation for mgmt SDKs and these samples would be published to ms learn site for end users.

PR #40953, opened 2026-03-03
test: verify .NET SDK validation is not required for old emitter mgmt specs

haiyuazhang:
Test PR to verify that .NET SDK validation is NOT marked as required when only changing management-plane TypeSpec specs that use the old emitter (@azure-tools/typespec-csharp) instead of the new one (@azure-typespec/http-client-csharp-mgmt).

Changes: trivial comment added to API Management and Management Groups main.tsp files.

This PR should be closed after verification.


haiyuazhang:
Test complete. Confirmed .NET SDK validation is NOT marked as required for management-plane specs using the old emitter (@azure-tools/typespec-csharp). Closing.

PR #41001, opened 2026-03-04
MPG migration Spec - Support

mzhongl524:

## Purpose of this PR

## Due diligence checklist

## Getting help


ArcturusZhang:
Reasons for removing `@subscriptionResource`:
This model has two sets of CRUD associated with them - one of them is under subscription scope, the other is under tenant scope.
Despite ARM may consider the two sets of CRUD as the same ARM resource - because they share the same ARM resource type, but it can be deployed under either subscription or tenant scope, which makes the decorator on its model misleading and incorrect.
On the other hand, due to some design details on client side, .net mgmt SDK modeled this case as two different ARM resource clients with the same ARM resource type, one of them lives under subscription, the other is under tenant.
Therefore from both sides - spec side or client side, I think it is very reasonable to remove that decorator.

specification/support/resource-manager/Microsoft.Support/Support/SupportTicketDetails.tsp
```diff
@@ -15,7 +15,6 @@ namespace Microsoft.Support;
 /**
  * Object that represents SupportTicketDetails resource.
  */
-@subscriptionResource
```
sandipsh, line 18: wouldn't removing this make it resource group level resource ?

sandipsh:
please address open comments including copilot ones.

specification/support/resource-manager/Microsoft.Support/Support/SupportTicketDetails.tsp
ArcturusZhang: > wouldn't removing this make it resource group level resource ?

you could tell it from the generated swagger - it is unchanged therefore no.
if you are saying "would removing this let people to assume this resource is a resource group level resource", I think you need to talk to the typespec team about this topic, this is way too big for me to have good answers.

ArcturusZhang:
> please address open comments including copilot ones.

none of the comments from copilot makes sense considering my explanation here: https://github.com/Azure/azure-rest-api-specs/pull/41001#issuecomment-4079923859

ArthurMa1978 (approved):

ArthurMa1978:
This PR is TypeSpec conversion related PR, merged to fix conversion issues.

PR #41101, opened 2026-03-06
[Horizon] Compute typespec

jayantjha:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Other, please clarify:
    - Adding APIs to create and manage compute clusters within a Cognitive Services Foundry account so that users can run custom training jobs. Adding only to the Typespec spec. 

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


jayantjha:
Issues Found
1. Missing workload submission endpoint (blocker for that use case)
There is no POST action for submitting jobs/workloads to a Foundry compute cluster. You'd need either:
A child resource: /computes/{computeName}/jobs/{jobName} (PUT/GET/LIST/DELETE), or
An action endpoint: POST /computes/{computeName}/submitJob
This is entirely absent from both the TypeSpec and the swagger.

_**The comment about workload submission is correct. But it is not part of compute. It will exists as a separate account level jobs API. So, no action in this PR**_

2. Missing PATCH operation for scaling
In Compute.tsp, the interface only has get, createOrUpdate, delete, list. To scale (change nodeCount per pool), a user must PUT the full resource. Add a PATCH (ArmResourceCreateOrUpdateAsync or ArmCustomPatchAsync) accepting just { pools: [...] } or a ComputeUpdateProperties. Compare with how Deployments_Update uses a patch in the same swagger.

_**Patch operation is currently not supported in the API**_

3. PUT only returns 202, missing 200/201 responses
The TypeSpec suppresses arm-put-operation-response-codes and only emits 202 Accepted. ARM compliance and AutoRest SDK generation both expect:
200 OK — for a synchronous update
201 Created — for a synchronous create
Without these, clients have no way to handle fast/synchronous completions and some SDK generators will behave incorrectly. The final-state-via: location is correctly set, but the missing 200/201 is flagged by Azure API linting for a reason.

_**The PUT response shape in this spec matches the current backend behavior, which is async-only and return 202. 201/200 is currently not supported by API**_

4. RootError is a non-standard thin error model
The definition at line 16635 only has code and message. Azure common-types already provides ErrorDetail with code, message, target, details[], and additionalInfo[]. Using RootError means errors surfaced on the ComputeProperties.errors field lose structured detail. Replace with ErrorDetail from common-types/resource-management/v3/types.json.

_**Service had added a RootError contract which is indeed similar to the common ErrorDetail. I have update the typespec**_

5. location duplicated in both Compute and ComputeProperties
The Compute definition has a location field on the resource envelope, AND ComputeProperties (line 11607) also has its own location. This is redundant and confusing for callers — they have to populate both or the semantics between them are unclear. Remove location from ComputeProperties and rely only on the resource envelope location.

_**Yes, but that is how the service contract is. I updated the location under `ComputeProperties` to be readonly to avoid user's confusion.**_

6. Pool has no min/max node count for autoscaling
The Pool model (line 15188) only has a fixed nodeCount: int32. For foundry/training workloads, autoscale is common. Consider adding minNodeCount and maxNodeCount and making nodeCount represent current/desired count.

**_Currently no such support in the service_**

7. No lastModifiedTime in ComputeProperties
creationTime is present as a read-only field but there's no lastModifiedTime. For audit/cache-busting and for detecting if a scale operation completed, this is commonly useful.

_**Yes, but there is no such field in the contract. Again because there is no Patch support currently**_

8. ComputeListResult.value is readOnly: true but the list operation responses don't filter nextLink pagination
This is minor — the swagger is correct technically — but value as read-only on a response-model-only type is redundant annotation. Not a bug, just noise from the TypeSpec @visibility(Lifecycle.Read) on the value field in models.tsp around line 5549.

_**Updated to remove @visibility decorator. wasn't required.**_

PR #41141, opened 2026-03-07
Use create-authenticated-npmrc.yml for npmrc authentication

chidozieononiwu:
### Packages impacted by this PR
All

### Issues associated with this PR
https://github.com/Azure/azure-sdk-for-js/issues/37076

### Describe the problem that is addressed by this PR
Use `create-authenticated-npmrc.yml` for `.npmrc` authentication

### DIFF
--- a/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
+++ b/eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
@@ -206,7 +206,10 @@ stages:
           versionSpec: $(NodeVersion)
         displayName: 'Install Node.js'
 
-      - template: /eng/common/pipelines/templates/steps/authenticate-temp-npmrc.yml
+      - template: /eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
+        parameters:
+          npmrcPath: $(Agent.TempDirectory)/gen-sdk/.npmrc
+          CustomCondition: and(succeeded(), contains(variables['SdkRepoName'], '-js'))
 
       - task: UsePythonVersion@0
         inputs:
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -266,6 +269,8 @@ stages:
         displayName: 'Generate SDK'
         ${{ if eq(parameters.ForceCreateEvenWithFailures, true) }}:
           continueOnError: true
+        env:
+          NPM_CONFIG_USERCONFIG: "$(Agent.TempDirectory)/gen-sdk/.npmrc"
```
raych1, line 272: Should we also set env when the `.npmrc` exists or if it's JS repo?

eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```diff
@@ -3,6 +3,7 @@ parameters:
     type: string
   - name: registryUrl
     type: string
+    default: 'https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js/npm/registry/'
```
raych1, line 6: I assume this common template change will be reverted when merge.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
chidozieononiwu: I cannot condition it on `SdkRepoName` at template level. This should be fine since the any npm call can use the configured `.npmrc`

raych1: This changes the config location silently for non-JS runs and it will be brittle for future changes.

Could you make the step of `template: /eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml` always run instead of conditioning on `JS` repo so that the `.npmrc` config will always be created and used in this step?

raych1 (approved):

PR #41157, opened 2026-03-09
feat: Add session file operations TypeSpec for hosted agents

ankitsul:
## Summary

Add TypeSpec definitions for session-scoped file operations on hosted agent sandboxes (stream-only).

### Operations (4 CRUDL)

| Method | Route | Description |
|--------|-------|-------------|
| `PUT` | `/agents/{name}/sessions/{id}/files?path=` | Upload file (binary stream) |
| `GET` | `/agents/{name}/sessions/{id}/files?path=` | Download file (binary stream) |
| `GET` | `/agents/{name}/sessions/{id}/files/list?path=` | List directory (non-recursive, single level) |
| `DELETE` | `/agents/{name}/sessions/{id}/files?path=` | Delete file/directory |

### Models (4)
- `SessionFileWriteResponse` — path, bytes_written
- `SessionDirectoryEntry` — name, size, is_dir, modified_time
- `SessionDirectoryListResponse` — path, entries[]
- `SessionFileOperationResponse` — message

### Preview Gating
- `@removed(Versions.v1)` on interface and all models — hidden from GA SDK
- `x-ms-foundry-meta` with `required_previews: hosted_agents_v1_preview` on each operation
- `FoundryDataPlanePreviewOperation` requires `Foundry-Features: HostedAgents=V1Preview` header

### Key Design Decisions
- **Sessions under agents (not versions)** — `/agents/{name}/sessions/` gives optionality to mutate the associated agent version independently of the session lifecycle
- **Non-recursive list** — returns immediate children only (cost/performance)
- **No symlink fields** — users can't create symlinks via the API
- **No `success` boolean** — HTTP status code suffices
- **No `file_api_root_path`** — consolidating with `$HOME` semantics (follow-up)
- **409 Conflict** on non-recursive delete of non-empty directory

### Related
- Design doc: https://msdata.visualstudio.com/Vienna/_git/vienna/pullrequest/1970903
- API review doc: https://github.com/Azure/agent-first-sdk/pull/189
- ADC integration spec (merged): https://github.com/Azure/azure-rest-api-specs/pull/40739

### Note
`tsp compile .` produces 2 pre-existing errors in `openai-responses/models.tsp` (`OpenAI.AutoCodeInterpreterToolParam` missing) — not related to this PR.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
@@ -1,0 +1,52 @@
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Response models
+// ---------------------------------------------------------------------------
+
+@removed(Versions.v1)
+@doc("Response from uploading a file to a session sandbox.")
+model SessionFileWriteResponse {
+  @doc("The path where the file was written.")
+  path: string;
+
+  @doc("Number of bytes written.")
+  bytes_written: int64;
+}
+
+@removed(Versions.v1)
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: int64;
+
+  @doc("Whether this entry is a directory.")
+  is_dir: boolean;
+
+  @doc("The last modification time in UTC (ISO 8601).")
+  @encode(DateTimeKnownEncoding.rfc3339)
+  modified_time: utcDateTime;
+}
+
+@removed(Versions.v1)
+@doc("Response from listing a directory in a session sandbox.")
+model SessionDirectoryListResponse {
+  @doc("The path that was listed.")
+  path: string;
+
+  @doc("The directory entries.")
+  entries: SessionDirectoryEntry[];
+}
+
+@removed(Versions.v1)
+@doc("Generic response for session file operations (delete).")
+model SessionFileOperationResponse {
+  @doc("A human-readable message describing the result.")
+  message: string;
+}
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
@@ -1,0 +1,145 @@
+import "../common/models.tsp";
+import "./models.tsp";
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+/**
+ * Session-scoped file operations for hosted agent sandboxes.
+ *
+ * These endpoints enable uploading, downloading, listing, and deleting
+ * files within a live ADC sandbox session. All file content is streamed
+ * end-to-end (no buffering in Agents service memory).
+ *
+ * Sessions are scoped to agents (not versions), giving optionality to
+ * mutate the associated agent version independently of the session lifecycle.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
+@tag("Agent Session Files")
+@removed(Versions.v1)
+interface AgentSessionFiles {
+
+  /**
+   * Upload a file to the session sandbox via binary stream.
+   */
+  @put
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  uploadSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The destination file path within the sandbox. */
+      @query path: string;
+
+      /** The binary file content to upload. */
+      @header contentType: "application/octet-stream";
+      @body content: bytes;
+    },
+    SessionFileWriteResponse
+  >;
+
+  /**
+   * Download a file from the session sandbox as a binary stream.
+   */
+  @get
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  downloadSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file path to download from the sandbox. */
+      @query path: string;
+    },
+    bytes
+  >;
+
+  /**
+   * List files and directories at a given path in the session sandbox.
+   * Returns only the immediate children of the specified directory (non-recursive).
+   */
+  @get
+  @route("/agents/{agent_name}/sessions/{session_id}/files/list")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  listSessionFiles is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The directory path to list. */
+      @query path: string;
+    },
+    SessionDirectoryListResponse
+  >;
+
+  /**
+   * Delete a file or directory from the session sandbox.
+   * If `recursive` is false (default) and the target is a non-empty directory, the API returns 409 Conflict.
+   */
+  @delete
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  deleteSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file or directory path to delete. */
+      @query path: string;
+
+      /** Whether to recursively delete directory contents. Defaults to false. */
+      @query recursive?: boolean;
+    },
+    SessionFileOperationResponse
+  >;
+}
--- a/specification/ai-foundry/data-plane/Foundry/main.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/main.tsp
@@ -1,5 +1,6 @@
 import "./src/agents/routes.tsp";
 import "./src/agents-containers/routes.tsp";
+import "./src/agents-session-files/routes.tsp";
 import "./src/agents-invocations/routes.tsp";
 import "./src/common/custom-types.tsp";
 import "./src/common/service.tsp";


specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,215 @@
...
+@tag("Agent Session Files")
+@removed(Versions.v1)
+interface AgentSessionFiles {
+
+  /**
+   * Upload a file to the session sandbox via binary stream.
+   * The file path is resolved relative to the agent version's `file_api_root_path`.
+   */
+  @put
+  @route("/agents/{agent_name}/versions/{agent_version}/sessions/{session_id}/files")
```
ericsuh: Didn't we wanted to move this to `/agents/{agent_name}/sessions/{session_id}/files`, to give us optionality for mutating the associated agent version?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,106 @@
...
+  path: string;
+
+  @doc("Number of bytes written.")
+  bytes_written: int64;
+
+  @doc("The source type used for the upload.")
+  source_type: "stream";
+
+  @doc("Whether the operation succeeded.")
+  success: boolean;
```
ericsuh: What does this provide that isn't already in the HTTP status code?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,106 @@
...
+@doc("Response from uploading a file to a session sandbox.")
+model SessionFileWriteResponse {
+  @doc("The path where the file was written.")
+  path: string;
+
+  @doc("Number of bytes written.")
+  bytes_written: int64;
+
+  @doc("The source type used for the upload.")
+  source_type: "stream";
```
ericsuh: What is this for?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,106 @@
...
+  success: boolean;
+}
+
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  name: string;
+
+  @doc("The full path relative to the sandbox root.")
+  path: string;
```
ericsuh: I wonder if this would be better as a relative path to the directory that is being listed. That would be more consistent with other filesystem-listing libraries in languages like C, Python, and C++, and would also would encourage devs not to rely on the specific absolute path.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,106 @@
...
+model SessionDirectoryListResponse {
+  @doc("The path that was listed.")
+  path: string;
+
+  @doc("The directory entries.")
+  entries: SessionDirectoryEntry[];
+}
+
+@doc("Metadata for a file or directory in a session sandbox.")
+model SessionFileInfo {
```
ericsuh: Does the blobfuse expose any sort of version ID / generation ID? If so, that might be nice in order to allow for compare-and-set like operations to handle concurrency, since an agent could be touching a file at the same time as this API interface

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,215 @@
...
+      @path agent_version: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file or directory path to delete. */
+      @query path: string;
+
+      /** Whether to recursively delete directory contents. Defaults to false. */
+      @query recursive?: boolean;
```
ericsuh, line 141: If this is false, does the API call err on a non-empty dir?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -302,6 +302,11 @@ model HostedAgentDefinition extends AgentDefinition {
   @doc("The image ID for the agent, applicable to image-based hosted agents.")
   @example("my-registry.azurecr.io/my-hosted-agent:latest")
   image?: string;
+
+  @removed(Versions.v1)
+  @doc("The allowed root directory for session file API operations. All file API paths are resolved relative to this root. Defaults to '/mnt/user/'.")
+  @example("/mnt/user")
+  file_api_root_path?: string;
```
ericsuh: Should we update to merge with the `$HOME` semantics we talked about at the last sync?

ericsuh (commented):
I think it would be helpful to ensure that all `path` fields are clearly documented as absolute or relative to something (I would also favor relative paths more than absolute paths).

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: This requires more thought as we have two options:
* Option 1: `session` (A) still lives under `versions` and we just have separate APIs for "session lifecycle" to create new `session` (under another agent version) from A's state.
* Option 2: `session` (A) lives under `agents` and we introduce a way (through a parameter) in "session lifecycle" APIs to link to a `version`. When a new `session` is created from A`s state, it can just specify the `version` it needs to get CPU/memory/image configuration from.

I plan to follow-up with a proposal to close on that.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
ankitsul: @ericsuh Did we make a final decision on that? If yes, can you share the spec so that I can figure out how to consolidate?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: As per the original proposal, we wanted to support different types such as "blob", "volume-mount" etc. But okay to remove it for now till we decide on overall file entity modeling.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Good point — removed. The HTTP status code (201/200/4xx) already communicates success/failure. Dropped `success` from both `SessionFileWriteResponse` and `SessionFileOperationResponse`.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Agreed — removed the `path` field from `SessionDirectoryEntry`. The caller already knows the parent directory they listed, so `name` (bare filename) is sufficient and consistent with POSIX `readdir()`, Python's `os.listdir()`, etc. This also avoids leaking internal sandbox paths.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Good callout. In M2 the backend is the sandbox local filesystem via ADC, which doesn't expose version/generation IDs natively. When we add volume-backed storage (M2+, blobfuse), we should revisit this for compare-and-set semantics (ETags). Deferring for now as it would be speculative without the storage layer supporting it.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Yes — if `recursive` is false (default) and the target is a non-empty directory, the API returns `409 Conflict`. This matches POSIX `rmdir` semantics. Updated the doc on the operation to clarify this.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
ericsuh: I don't know who the decision-maker is, but given that Lakshmi was in favor of it and that there didn't seem to be any objections from others, I think it's the likely path forward. I don't think we have a spec per se, but we will likely have some sort of mount path like `/home/session`, and then pass that as `$HOME` as an environment variable.

ankitsul: Agreed — removed `file_api_root_path`. We'll consolidate with the `$HOME` semantics (`/home/session` as the well-known mount path set as `$HOME` env var). File API paths will be resolved relative to `$HOME` by default. Will track this in a follow-up once the spec is finalized.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: We had some internal discussions and supporting forking/mutation is still possible by keeping the existing url path. Also, moving it out changes what a session means. We will discuss offline and close on this.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,50 @@
+using TypeSpec.Http;
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Response models
+// ---------------------------------------------------------------------------
+
+@doc("Response from uploading a file to a session sandbox.")
+model SessionFileWriteResponse {
```
RaviPidaparthi, line 12: Mark all of these models as removed from v1 version

RaviPidaparthi:
Make sure to run `tsp compile .` and generate open api docs

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,50 @@
...
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: int64;
+
+  @doc("Whether this entry is a directory.")
+  is_dir: boolean;
+
+  @doc("Whether this entry is a symbolic link.")
+  is_symlink: boolean;
```
RaviPidaparthi: What is the use case for `symlink` ? I dont think we allow user to create these via API right? Is there any reason to let him discover these then?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,153 @@
...
+      /** The version of the agent. */
+      @path agent_version: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The directory path to list. */
+      @query path: string;
+    },
+    SessionDirectoryListResponse
```
RaviPidaparthi, line 111: Will list session files recursively return contents of sub directories?

RaviPidaparthi: I assume no as it would be costly otherwise.

ankitsul: Right, only level 1 flat contents.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Done — added `@removed(Versions.v1)` to all 4 models.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Good point — removed `is_symlink`. Users can't create symlinks via the API, so no reason to expose it.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Correct — list is non-recursive, returns only immediate children of the specified directory. Updated the operation doc to clarify this.

ankitsul:
`tsp compile .` runs clean — the only 2 errors are pre-existing on the branch (`OpenAI.AutoCodeInterpreterToolParam` missing in `openai-responses/models.tsp`), not from this PR. Will generate OpenAPI docs once those upstream errors are resolved.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: As discussed, pulled session operations out of agent_version.

ankitsul:
Due to some issue with Copilot, had to create a new one: https://github.com/Azure/azure-rest-api-specs/pull/41586

PR #41206, opened 2026-03-10
AgentEndpoint and ManagedAgentIdentityBlueprint typespec

acsoylemezoglu:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -203,4 +204,28 @@ interface Agents {
     },
     AgentsPagedResult<AgentVersionObject>
   >;
+
+  /**
+   * Updates an agent endpoint.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" ""
+  @patch(#{ implicitOptionality: true })
```
RaviPidaparthi, line 212: Patch or post? We didn't have any patch apis till now.

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
```diff
@@ -0,0 +1,95 @@
...
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+interface ManagedAgentIdentityBlueprints {
+  /**
+   * Creates a managed agent identity blueprint.
+   */
+  @put
+  @route("/managedAgentIdentityBlueprints/{managed_agent_identity_blueprint_name}")
```
RaviPidaparthi: long name. Check other apis too but I think we use snake case

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
acsoylemezoglu: I was talking with Andrew and he was telling me that ARM doesn't like us doing POSTs with PATCH semantics. We were planning on allowing customers to update any one aspect of the endpoint rather than having them POST a complete object, so we went with PATCH

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
johanste: For the parameter name, it is scoped as the "key" in the /managedAgentIdentityBlueprints collection, so `name` is enough here.

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/models.tsp
```diff
@@ -0,0 +1,41 @@
...
+  name: string;
+}
+
+@removed(Versions.v1)
+model ManagedAgentIdentityBlueprintPagedResponse {
+  @doc("The identities returned")
+  data: ManagedAgentIdentityBlueprint[];
+
+  @doc("An optional continuation token or pagination. A null value indicates the end of the list")
+  token?: string;
```
johanste: Nitpick; are we proxying this api elsewhere? This is not the common pagination pattern we use. 

(and my brain actually made a "token in identity thing response - bearer token!" assumption when first saw the change here)

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
acsoylemezoglu: I did this because I have a name field in the body and this conflicts with that. Is the recommendation that I omit name from the route and just switch to POST?

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/models.tsp
acsoylemezoglu: not sure what you mean by proxying this API elsewhere, but happy to switch to whatever the common pagination pattern is if you can point me to it.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -65,6 +65,7 @@ union FoundryFeaturesOptInKeys {
   insights_v1_preview: "Insights=V1Preview",
   memory_stores_v1_preview: "MemoryStores=V1Preview",
   toolset_v1_preview: "Toolsets=V1Preview",
+  agent_endpoint_v1_preview: "AgentEndpoints=V1Preview",
```
johanste, line 68: API description only/not REST API: Why is this preview header not part of AgentDefinitionOptInKeys?

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -203,4 +204,28 @@ interface Agents {
...
+  @route("/agents/{agent_name}/endpoint")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview],
+    }
+  )
+  @removed(Versions.v1)
+  updateAgentEndpointObject is FoundryDataPlanePreviewOperation<
```
johanste, line 221: I assume Content-Type for the request body is `application/merge-patch+json`? If so, we should specify it.

johanste: Also, array updates with json-merge-patch are not ideal since they have replace semantics (you can't modify the array, you can only replace it wholesale)

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
```diff
@@ -0,0 +1,93 @@
...
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
+interface ManagedAgentIdentityBlueprints {
+  #suppress "@azure-tools/typespec-azure-core/long-running-polling-operation-required" ""
```
johanste: Why do you need this suppression?

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
johanste:
```suggestion
  @route("/managedAgentIdentityBlueprints/{name}")
```

specification/ai-foundry/data-plane/Foundry/src/managed-blueprints/models.tsp
```diff
@@ -0,0 +1,41 @@
...
+    - Must not exceed 63 characters.
+    """)
+  @maxLength(63)
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  name: string;
+}
+
+@removed(Versions.v1)
+model ManagedAgentIdentityBlueprint {
+  name: string;
```
johanste, line 32: This is only ever set:able on create, right? If so:

```suggestion
  @visibility(Lifecycle.Read)
  name: string;
```

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/models.tsp
johanste: See agents list:

  /**
   * Returns the list of all agents.
   */
  @route("/agents")
  @list
  listAgents is FoundryDataPlaneOperation<
    {
      ...ListAgentQueryParameters;
    },
    AgentsPagedResult<AgentObject>
  >;

johanste: Uses openai style has_more/first/last ids. Alternatively, the azure style nextLink which is used elsewhere. It is unfortunate that we have both in Foundry. But we shouldn't introduce a third. 

And regarding the proxy question; is this a continuation token we get from i.e. calling through to entra or cosmos, and it is just an opaque value for Foundry or is it something that we own/crack open/interpret...

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
acsoylemezoglu: the AgentDefinitionOptInKeys made it sound like it was specifically for certain types of agent definitions (e.g. Hosted definitions), whereas this is applicable for all agents so I figured the general features made more sense?

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/models.tsp
acsoylemezoglu: I think this will be a continuation token we get from calling cosmos, So I think a nextLink probably makes more sense. I saw the openai style one but I didn't want to do the first/last ids which is why I went with this.

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
acsoylemezoglu: probably not needed, I added it because I saw other create calls had it too. Initially typespec compile was complaining on the standard operations stuff which I suppressed, so when I brought that in, I also brought this in, but can remove

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
johanste: Since this is not a REST API affecting change, it is not blocking. But am curious if we made use of the unions in our client libraries - @trrwilson , do you happen to know?

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
acsoylemezoglu: yeah that is not ideal, we'd rather not have that behavior so we'll have to work around it

johanste: We usually see if it makes sense to make the array a json object. Wich is usually possible if/when there is a key/property that is supposed to have a unique value within the array. What are the constraints on the values in the arrays? I assume that there is no ordering requirement in the protocols array and that duplicate values are just ignored. But that in the authorization schemes, you can only have one entry per authorization scheme type? And it would be a 400 response/error if you had two?

acsoylemezoglu: there is no ordering requirement in either array and dupes will be ignored in the protocols array. You can have one entry per authorization scheme type and 400ing makes sense. I also don't think you can have BotService and BotServiceRbac configured together, but I'll confirm with Andrew. However, you can have Entra alongside either of those.

acsoylemezoglu: actually now that I think about it, we'd prefer the array replacement semantics anyway. For whatever field, they're sending whether it is, schemes or protocols, they should send the complete list or null if they dont want any updates

acsoylemezoglu: added application/merge-patch+json header

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/models.tsp
acsoylemezoglu: I think I did this properly

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -561,3 +642,254 @@ union ContainerLogKind {
...
+)
+@maxLength(32)
+scalar SkillTag extends string;
+
+@removed(Versions.v1)
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview] }
+)
+model AgentSkill {
```
johanste: I suspect that we have different variations of what a skill means. In this case, it is for the agent card, correct? Which is different than, say, the openai skills api. It would probably behoove us to have a more specific name for the model... maybe:


```suggestion
model AgentCardSkill {
```

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -561,3 +642,254 @@ union ContainerLogKind {
...
+  agent_card?: AgentCard;
+}
+
+@removed(Versions.v1)
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview] }
+)
+model AgentIdentity {
+  @doc("The details of the identity")
```
johanste: Do we expect more things to be added here? Or can we flatten the `IdentityDetails` onto this? If we expect different "kinds" of AgentIdentityDetails to be added, we should add a disciminator to it.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -99,6 +118,38 @@ model AgentObject {
   versions: {
```
fosteramanda, line 118: @acsoylemezoglu Can we add more properties to the agent object. These are metadata requirements for A365 that we need to register agents in MOS:

All are optional
- Display Name (this value should be what is set when user publishes agent to M365/Teams)
- Short Description
- Full Description
- Version
- Developer Name
- Website URL
- Privacy URL
- Terms of use URL
- Published to channel (None, M365/Teams, Digital Worker)
- Publish scope (org or individual, only set if agent is published to a channel)

fosteramanda (commented):
@acsoylemezoglu Can we add more properties to the agent object. These are metadata requirements for A365 that we need to register agents in MOS:

All are optional
- Display Name (this value should be what is set when user publishes agent to M365/Teams)
- Short Description
- Long Description
- Version
- Developer Name
- Website URL
- Privacy URL
- Terms of use URL
- Published to channel?
- Publish scope (org or individual)

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -561,3 +642,257 @@ union ContainerLogKind {
...
+  version_selector?: VersionSelector;
+
+  @doc("The protocols that the agent supports")
+  protocols?: AgentEndpointProtocol[];
+
+  @doc("The authorization schemes supported by the agent endpoint")
+  authorization_schemes?: AgentEndpointAuthorizationScheme[];
+
+  // TODO: move this out to the agent object. This is under preview so for fast iteration putting it here
+  @doc("The agent card")
```
fosteramanda, line 804: Why are we setting AgentCard on the endpoint not the Agent Object? Is that because we want to potentially allow for additional stable endpoints?

fosteramanda:
Could we add sdk support for the publish api (used when an agent is published to Teams/M365 or as an A365 Digital Worker)? https://[eastus.api.azureml.ms/agent-asset/swagger/v1.0/swagger.json](https://eastus.api.azureml.ms/agent-asset/swagger/v1.0/swagger.json) 

I believe a special exception would be needed since this api is not a part of the Cognitive Services FQDN. Customers have asked for a code first way to publish their agents and we don't want to directly expose the REST API in our documentation since it is an internal api. However, we already sort of expose it today which is not ideal: Today we already sort of expose it which is bad since this is an internal api: [foundry-samples/samples/csharp/FoundryA365/scripts/publish-digital-worker.ps1 at main · microsoft-f…](https://github.com/microsoft-foundry/foundry-samples/blob/main/samples/csharp/FoundryA365/scripts/publish-digital-worker.ps1). Customers have asked for a code first way to publish their agents and we don't want to directly expose the REST API in our documentation since it is not a part of the Cognitive Services FQDN. 

My ask is could we add sdk support for this API by EOM which would call the internal api and update it later once we support the api through the correct FQDN. I believe this requires a special exception.

@johanste @qubitron Is this possible?

specification/ai-foundry/data-plane/Foundry/src/managedagentidentityblueprints/routes.tsp
```diff
@@ -0,0 +1,93 @@
...
+      required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview],
+    }
+  )
+  @removed(Versions.v1)
+  listManagedAgentIdentityBlueprints is FoundryDataPlaneRequiredPreviewOperation<
+    FoundryFeaturesOptInKeys.agent_endpoint_v1_preview,
+    {
+      ...PageOrderQueryParameter;
+      ...PageLimitQueryParameter;
+      @query @continuationToken continuationToken?: string;
```
johanste: Making sure my assumption is correct - the nextLink property value in the response is a fully formed URL that includes the continuationToken query parameter, correct? If so, we can remove it from here.

acsoylemezoglu: yes I believe so, will remove

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
trrwilson: No external-facing use. Every language uses either a `.beta` client infix or equivalent (.NET's `ExperimentalAttribute`) to abstract the direct origination of the values. The union's a tidiness thing.

I do think this value could likely be safely included in the union, but I wouldn't block on it -- and under no circumstance should this be a regrettable "do it later" decision.

specification/ai-foundry/data-plane/Foundry/main.tsp
```diff
@@ -12,6 +12,7 @@ import "./src/evaluations/routes.tsp";
 import "./src/evaluators/routes.tsp";
 import "./src/indexes/routes.tsp";
 import "./src/insights/routes.tsp";
+import "./src/managedagentidentityblueprints/routes.tsp";
```
dargilco: please insert "-" between the words in the folder name, like other folders. Also, is there a shorter name we can use?

dargilco (changes requested):

PR #39869, opened 2026-01-23
[Inner loop][TypeSpec Authoring] Add TypeSpec Authoring Skill

haolingdong-msft:
This PR is based on @chunyu3's custom agent [PR ](https://github.com/Azure/azure-rest-api-specs/pull/39597). It uses skill only to author TypeSpec. See the detailed design approach [here](https://github.com/Azure/azure-sdk-tools/blob/9969310e24ae2a682805a9da83ca23b8162fb905/tools/azsdk-cli/docs/specs/typespec-authoring.spec.md#approach-2-skills-only-approach-without-custom-agent) for more details.

### Scenario workflow

Add new preview version
1. collect required information from user
2. retrieve solution from KB backend
3. update code based on the retrieved solution
4. set up examples folder
5. ask for follow up features to be added to this version

Add new preview version
1. collect required information from user
2. retrieve solution from KB backend
3. update code based on the retrieved solution
4. set up examples folder
5. promote features from latest preview to stable
6. ask for follow up features to be added to this version

### Code change

The PR adds a skill named: azure-typespec-author, this includes:
- add a new GitHub Skill folder under .github/skills/azure-typespec-author/
- SKILL.md: skill definition + required workflow steps. 
- intake-arm.md: promtps to describe ARM intake/clarification instructions(project analysis + case questionnaires).
- next-steps-arm.md: prompts to describe general follow up actions and case-specific ones

### Try out Steps:
**prerequisite:**  One need to Enable the chat.useAgentSkills setting to use Agent Skills in VSCode. The [document ](https://code.visualstudio.com/docs/copilot/customization/agent-skills)has link to enable the setting.
1. check out code from the branch in this PR: [typespec-authoring-skill](https://github.com/Azure/azure-rest-api-specs/tree/typespec-authoring-skill)
7. start the azsdk-cli mcp server follow here: https://github.com/Azure/azure-sdk-tools/blob/main/tools/azsdk-cli/Azure.Sdk.Tools.Cli/README.md#quick-start

### Agent output
image

PR #41241, opened 2026-03-11
Temporarily Disable azure-typespec-author skill

haolingdong-msft:
Rename SKILL.md to SKILL.md.disabled to deactivate the azure-typespec-author skill.

### Context:
I hope to move and iterate fast based on user(currently it is SDK team members) feedback for the azure-typespec-author skill. But merge process in spec repo is a bit heavy and time-costing. Last time we spent almost one week to merge the [PR](https://github.com/Azure/azure-rest-api-specs/pull/39869). And **I don't have permission to merge**. So I'm currently publishing the skill as plugin in my personal [repo](https://github.com/haolingdong-msft/azure-typespec-authoring), **which can significantly speed up develop-feedback loop**. Because I can easily push code and iterate quickly based on the feedback and then publish a new version. 

But Copilot CLI will only load one skill based on priority: project skill > plugin skill.

This brings a problem that pilot user(e.g. internal user) cannot use my latest updated skill (from my personal [repo](https://github.com/haolingdong-msft/azure-typespec-authoring)) when writes typespec inside azure-rest-api-spec repo. Because the project level [skill ](https://github.com/Azure/azure-rest-api-specs/tree/main/.github/skills/azure-typespec-author)overwrites plugin skill. There are two solutions:
Option 1:  temporarily disable the skill in spec repo so user will use the skill in my plugin, until we decided how to publish skill, I'm happy to move the skill out from my personal repo to somewhere else, e.g. azure-sdk-tools repo
Option 2:  manually sync the code from my plugin skill to the skill in rest repo frequently.

This implementation option 1.


timotheeguerin:

PR #41284, opened 2026-03-12
fix(containerregistrytasks): update flatten properties for .NET SDK compliance

CodeJourneys:
This PR updates property flattening for .NET SDK compliance


northtyphoon (approved):
LGTM

PR #41288, opened 2026-03-12
fix(containerregistrytasks): adjust property naming case and add units for timeout properties

CodeJourneys:
This PR adjusts property naming case and add units for timeout properties to align with .NET SDK conventions


northtyphoon (approved):
LGTM

PR #41295, opened 2026-03-12
Add code-based hosted agent API surface

sagunase:
### Summary
Adds multipart/form-data code deployment support to all three agent mutation endpoints using @sharedRoute. Each endpoint accepts both the existing JSON payload **and** a new multipart payload (code zip + JSON metadata), differentiated by Content-Type header.

The agent name and content hash are carried in **HTTP headers** (not the body), enabling early validation before any body bytes are read and eliminating any dependency on multipart part ordering.

### Endpoints
| Route | JSON (existing) | Multipart (new) |
|-------|----------------|-----------------|
| POST /agents | createAgent | createAgentFromCode -> AgentObject |
| POST /agents/{name} | updateAgent | updateAgentFromCode -> AgentObject |
| POST /agents/{name}/versions | createAgentVersion | createAgentVersionFromCode -> AgentVersionObject |

### Required Headers (multipart only)
| Header | Required | Description |
|--------|----------|-------------|
| `x-ms-agent-name` | On POST /agents only | Agent name (max 63 chars). Not needed on the other 2 endpoints where name is in the URL path. |
| `x-ms-code-zip-sha256` | On all 3 endpoints | SHA-256 hex digest of the ZIP. Used for dedup, integrity verification, and idempotent retries. |

### Models
**Metadata:**
- CreateAgentVersionFromCodeRequest - description?, metadata?, definition: HostedAgentDefinition
- Used by all 3 multipart endpoints (agent name comes from URL or header, not body)

**Multipart wrappers:**
- CreateAgentVersionFromCodeContent - wraps CreateAgentVersionFromCodeRequest + code: HttpPart<bytes> (for update + version endpoints)
- CreateAgentFromCodeContent - spreads from CreateAgentVersionFromCodeContent (for create endpoint, extensible)

### Key Design Decisions
- **@sharedRoute** - Same route, differentiated by Content-Type. SDK emitters generate distinct methods (e.g., create_agent() vs create_agent_from_code())
- **Headers for name and hash** - See "Why Headers?" rationale below
- **Single metadata model** - Since agent name is in headers (not body), all 3 endpoints use the same metadata model
- **Feature-gated** - All multipart endpoints require CodeAgents=V1Preview feature flag

### Why Headers? (Design Rationale)

The goal of multipart code deploy is to let customers create an agent from code in a single call, while still validating the request early and avoiding streaming large payloads when the request is invalid.

**The ordering problem.** An earlier design required the metadata part to appear before the zip part so the service could read the agent name and validate before ingesting bytes. But this turns multipart part ordering into a semantic contract. RFC 7578 defines multipart/form-data parts as named fields identified by headers (e.g. `Content-Disposition: form-data; name=...`), not by position. The RFC does not assign semantic meaning to ordering, and compliant clients, SDKs, middleware, and proxies are free to reorder, buffer, or normalize parts. Making correctness depend on part order introduces a fragile, environment-dependent failure mode that is extremely hard to diagnose.

**What we actually need.** The service must know the agent name and request intent *before* it starts ingesting bytes. HTTP already provides a reliable mechanism: the request target and headers are parsed before the body. The multipart body should be treated purely as a data container, not as a signaling mechanism.

**Why content hash must be required.** Leaving `content_hash` optional undermines both idempotency and integrity. If retries occur and the hash is missing, the service cannot reliably distinguish "same upload retried" from "new upload," and cannot verify that the streamed bytes match what the client intended to send. SHA-256 can be computed in a streaming fashion, so there is no technical reason for it to be optional.

**Recommended approach.** Keep the GA surface (`POST /agents` with `application/json`) intact. For the multipart code deploy surface, move the agent name and required content hash into headers. This:
- Preserves the one-call "create agent from zip" experience
- Allows early validation before any bytes are accepted
- Removes any dependency on multipart ordering
- Gives strong integrity and retry semantics

**Alternatives considered (and rejected):**
1. **"Metadata must be first" rule** - Relies on multipart ordering for correctness, which is not a stable or portable contract.
2. **Buffer the entire multipart body** - Defeats the purpose of streaming uploads and re-introduces memory and availability risks.
3. **Restrict zip uploads to POST /agents/{name}/versions only** - Forces a multi-step flow and breaks the intuitive "create agent from code" mental model.

> *Reference: Discussion between Lakshmi, Johan, Dan, and Travis.*

### Backward Compatibility
- All existing JSON-only endpoints remain unchanged
- Legacy createAgent, updateAgent, createAgentVersion continue to work as before

### Files Changed
- src/agents/models.tsp - Single metadata model + two multipart wrappers (CreateAgentFromCodeContent spreads from CreateAgentVersionFromCodeContent), content_hash removed from CodeConfiguration
- src/agents/routes.tsp - @sharedRoute on 3 endpoints + 3 new multipart operations with header params
- examples/2025-11-15-preview/ - Examples for all 3 multipart operations (with headers)
- openapi3/ - Regenerated OpenAPI specs

### Known Limitations
1. **Maximum upload size: 250 MB** - The code zip in the multipart request is capped at 250 MB.
2. **Full zip upload on every version** - No incremental/delta update support. Each call requires uploading the complete zip even for small changes.


specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -221,4 +221,27 @@ interface Agents {
...
+  /**
+   * Upload a code zip file for code-based hosted agent deployment.
+   * The zip is stored internally and an opaque code_reference_id is returned.
+   * This ID is then passed into code_configuration.code_reference_id when creating an agent version.
+   * Maximum upload size is 250 MB.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" ""
+  #suppress "@azure-tools/typespec-azure-core/byos" "Code upload is stored internally by the service, not BYOS."
+  @post
+  @route("/code:upload")
```
johanste: I would advise against a server generated name for large uploads. 

Do we intend to provide any CRUD methods other than `upload` for code?

sagunase: Agreed — I've updated the design based on your feedback. Here's what changed:

**User-provided name + auto-versioning**: Upload route is now `POST /code/{name}:upload`. The user provides a meaningful `name` as a path parameter. Re-uploading with the same name auto-increments the version. The response returns `id` in `{name}:{version}` format (e.g., `"my-chatbot:2"`), which is used as `code_reference_id` when creating an agent.

**CRUD operations added**:
| Op | Method | Route |
|----|--------|-------|
| Upload | `POST` | `/code/{name}:upload` |
| List | `GET` | `/code` |
| List versions | `GET` | `/code/{name}/versions` |
| Delete all | `DELETE` | `/code/{name}` |
| Delete version | `DELETE` | `/code/{name}/versions/{version}` |

This follows the existing Datasets API pattern — `name` maps directly to the dataset name, version is auto-managed by the service. Delete returns `409 Conflict` if the code is referenced by active agents.

Updated PR description has full samples.

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -221,4 +229,144 @@ interface Agents {
...
+  /**
+   * Upload a code zip file for code-based hosted agent deployment.
+   * The zip is stored internally and a code_reference_id in format '{name}:{version}' is returned.
+   * Re-uploading with the same name auto-increments the version.
+   * This ID is then passed into code_configuration.code_reference_id when creating an agent version.
+   * Maximum upload size is 250 MB.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" ""
+  #suppress "@azure-tools/typespec-azure-core/byos" "Code upload is stored internally by the service, not BYOS."
+  @post
```
johanste, line 413:
```suggestion
  @put
```

sagunase: This endpoint has been removed. We pivoted to a one-shot multipart approach using `@sharedRoute` on the existing agent endpoints:

- `POST /agents` → `createAgentFromCode`
- `POST /agents/{name}` → `updateAgentFromCode`  
- `POST /agents/{name}/versions` → `createAgentVersionFromCode`

Each accepts both `application/json` and `multipart/form-data` (code zip + JSON metadata). No separate `/code/*` endpoints anymore.

johanste (approved):

sagunase:
This PR changes are merged in this [PR](https://github.com/Azure/azure-rest-api-specs/pull/42114). hence closing this PR.

PR #41329, opened 2026-03-12
MPG migration - confidentialledger

haiyuazhang:
Migrate service confidentialledger to use new management plane sdk generator


specification/confidentialledger/resource-manager/Microsoft.ConfidentialLedger/ConfidentialLedger/tspconfig.yaml
```diff
@@ -5,6 +5,7 @@ emit:
   - "@azure-tools/typespec-autorest"
 options:
   "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "."
```
live1206: what is this for?

haiyuazhang: It's interesting, the issue I mentioned no longer exists.  Remove this line anyway.

live1206 (approved):

PR #41330, opened 2026-03-12
[Storage] Fix Queues.SetAccessPolicy TypeSpec Changes

vincenttran-msft:
This has the same issue that we ran into Blobs, but this was largely untested in the Queues package previously.
https://github.com/Azure/azure-rest-api-specs/pull/38933


catalinaperalta (commented):
looks like this PR needs to be rebased over main to get a clean diff. About the rfc7231-fixed-width issue can you try adding a suppression for the error: @azure-tools/typespec-autorest/unknown-format. Please add a clear justification of why this encoding is necessary

vincenttran-msft:
Closing in favor of: https://github.com/Azure/azure-rest-api-specs/pull/41338

PR #41347, opened 2026-03-13
fix(containerregistry): restore API naming compatibility via @clientName decorators for .NET SDK compatibility

CodeJourneys:
Adjusted client.tsp for .NET SDK compatiblity, including @@clientName decorators to client.tsp to restore the ContainerRegistry prefix for all types that lost it during the TypeSpec migration, preserving backward compatibility with the 1.4.0 GA surface. Also renamed request body types from the Parameters suffix to the Azure SDK-required Content suffix (e.g. ImportImageParameters → ContainerRegistryImportImageContent).


northtyphoon (approved):
LGTM

PR #41358, opened 2026-03-13
[Storage] Namespace Update Proof of Concept

vincenttran-msft:
In response to the changes made in: 
https://github.com/Azure/typespec-rust/pull/877
I came up with this as my proposal to update the typespec to be non-breaking.


specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -119,6 +119,12 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 @@clientNamespace(Storage.Blob.AppendBlob, "Azure.Storage.Blobs");
 @@clientNamespace(Storage.Blob.BlockBlob, "Azure.Storage.Blobs");
 @@clientNamespace(Storage.Blob.PageBlob, "Azure.Storage.Blobs");
+@@clientNamespace(BlobServiceClient, "Azure.Storage.Blobs");
```
catalinaperalta, line 122: Not sure I understand what this change is trying to accomplish since the corresponding interfaces have already been assigned to the same namespace in the lines above? Also, this sort of change should be scoped down to rust

PR #41359, opened 2026-03-13
[JS] Mitigate the breaking for non-model return

v-jiaodi:
related issue: https://github.com/Azure/autorest.typescript/issues/3819


MaryGao (approved):
To keep backward-compatible generation with our new feature comes in - https://github.com/Azure/autorest.typescript/issues/3655, otherwise it would introduce breaking changes.

@ReneOv-MSFT and @cRui861 or @wiboris could you help approve this? Thanks.

MaryGao (approved):

PR #41363, opened 2026-03-13
Add C# mgmt SDK customizations for ContainerRegistry TypeSpec migration

live1206:
## Add C# management SDK customizations for ContainerRegistry migration

This PR adds C# client customizations to the ContainerRegistry TypeSpec spec to support migrating `Azure.ResourceManager.ContainerRegistry` from AutoRest/Swagger to TypeSpec-based generation.

### Changes

**Registry/client.tsp:**
- ~120 `@@clientName` decorators for backward-compatible type names (ContainerRegistry prefix, webhook event types, connected registry models, etc.)
- `@@alternateType` for ResourceIdentifier, Uri, uuid properties
- `@@clientName` for property renames (ExpireOn, ConnectionState, PolicyType, CertificateType, IPRules, etc.)
- Operation rename: CheckContainerRegistryNameAvailability

**Registry/back-compatible.tsp:**
- Added `csharp` to all `flattenProperty` decorators for backward-compatible property flattening

### Context
- SDK migration PR: (will be linked after creation)
- Issue: https://github.com/Azure/azure-sdk-for-net/issues/54739


specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/PrivateLinkResource.tsp
```diff
@@ -11,26 +11,33 @@ using TypeSpec.OpenAPI;
 
 namespace Microsoft.ContainerRegistry;
 
-alias PrivateLinkOperations = PrivateLinks<PrivateLinkResource>;
+/**
+ * A private link resource for a container registry.
+ */
+@parentResource(Registry)
+model PrivateLinkResource
```
pshao25: Why we want to define a model ourselves?

live1206: why do we replace common type with a custom model?

specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/PrivateLinkResource.models.tsp
ArcturusZhang, line 9: this is also peculiar - since we are using the standard template, why do we need this?

ArthurMa1978 (approved):

specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/PrivateLinkResource.tsp
```diff
@@ -11,7 +11,8 @@ using TypeSpec.OpenAPI;
 
 namespace Microsoft.ContainerRegistry;
 
-alias PrivateLinkOperations = PrivateLinks<PrivateLinkResource>;
+model MyPrivateLinkResource is PrivateLink;
```
XiaofeiCao, line 14: Why the change here? It may affect all SDKs.

ArthurMa1978: By design from Mark

ArthurMa1978 (approved):

specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/Registry/PrivateLinkResource.tsp
CodeJourneys, line 14: @live1206 - Could you provide some more context on this change?

live1206: instead of using the common PrivateLinkResource, we should have your own model with same model structure of the common one, and `PrivateLink` has more decorators while holding the same model structure of the common `PrivateLinkResource`.

PR #41367, opened 2026-03-13
[CDN] Add C# mgmt emitter and client customizations for SDK migration

live1206:
## CDN C# Management SDK Migration - Spec Changes

Add C# mgmt emitter configuration and client customizations for the CDN TypeSpec-based SDK migration (Issue: Azure/azure-sdk-for-net#54736).

### Changes

**tspconfig.yaml**
- Added `@azure-typespec/http-client-csharp-mgmt` emitter with namespace `Azure.ResourceManager.Cdn`

**client.tsp** — Added ~120 `@@clientName` decorators for backward-compatible C# type names:
- AFD → FrontDoor prefix renames (e.g., `AFDDomain` → `FrontDoorCustomDomain`)
- Model renames from old autorest rename-mapping (e.g., `CacheType` → `CdnCacheLevel`)
- Enum value renames to avoid C# conflicts (e.g., `Equals` → `EqualsValue`)
- Property renames for flattened list wrappers (e.g., `RateLimitRules` → `RateLimitSettings`)
- Duplicate method resolution (e.g., `profilesListResourceUsage` → `GetCdnProfileResourceUsages`)
- `@@access`/`@@usage` decorators for `DeepCreatedCustomDomain`

**models.tsp**
- Removed `@visibility(Lifecycle.Read)` from `customDomains` property on `EndpointProperties`

### Related
- SDK PR: TBD
- Migration issue: Azure/azure-sdk-for-net#54736


specification/cdn/resource-manager/Microsoft.Cdn/Cdn/client.tsp
```diff
@@ -176,42 +200,901 @@ using Microsoft.Cdn;
...
+@@clientName(ProfilesOperationGroup.migrate, "MigrateProfile", "csharp");
+@@clientName(AutoGeneratedDomainNameLabelScope,
+  "DomainNameLabelScope",
+  "csharp"
+);
+@@clientName(ManagedCertificateParameters,
+  "ManagedCertificateProperties",
+  "csharp"
+);
+@@clientName(SecretProperties, "SecretDetails", "csharp");
```
ArthurMa1978: This should be rename to CdnSecretProperties
Question: why does Secrets not exist in the generated sdk?

specification/cdn/resource-manager/Microsoft.Cdn/Cdn/client.tsp
```diff
@@ -176,42 +200,901 @@ using Microsoft.Cdn;
...
+@@clientName(SecurityPolicyProperties, "SecurityPolicyDetails", "csharp");
```
ArthurMa1978: This should be rename to CdnSecurityPolicyProperties
Question: why does SecurityPolicies not exist in the generated sdk?

specification/cdn/resource-manager/Microsoft.Cdn/Cdn/client.tsp
mzhongl524: The Secrets resource does exist in the generated SDK — it's generated as FrontDoorSecretResource, FrontDoorSecretCollection, and FrontDoorSecretData (via `@@clientName(Secret, "FrontDoorSecret", "csharp")`
 in client.tsp). The SecretProperties model being renamed here is the inner properties model, not the resource itself

specification/cdn/resource-manager/Microsoft.Cdn/Cdn/client.tsp
mzhongl524: The SecurityPolicies resource does exist in the generated SDK — it's generated as FrontDoorSecurityPolicyResource, FrontDoorSecurityPolicyCollection, and FrontDoorSecurityPolicyData (via 
 `@@clientName(SecurityPolicy, "FrontDoorSecurityPolicy", "csharp")` in client.tsp). The SecurityPolicyProperties model being renamed here is the inner properties model, not the resource itself.

ArthurMa1978 (approved):

PR #41375, opened 2026-03-13
Add C# mgmt emitter config for CognitiveServices

live1206:
### Changes
- Added `@azure-typespec/http-client-csharp-mgmt` emitter configuration to `tspconfig.yaml`
- Added C#-specific `@@clientName` decorators in `client.tsp` for backward-compatible type naming (~80 renames)
- Added `@@access` decorators to make certain types public in C#
- Added `@@clientLocation` overrides to fix method collisions in C# (duplicate `GetAll` extension methods, `Models` namespace collision, `CommitmentPlan` operation duplicates)

### Context
Part of the MPG migration for `Azure.ResourceManager.CognitiveServices` from AutoRest/Swagger to TypeSpec-based generation (Azure/azure-sdk-for-net#56942).

### SDK PR
Azure/azure-sdk-for-net (PR pending)


psah434:
This is a breaking change. Why has this not been reviewed by the breaking change team.

ArthurMa1978:
> This is a breaking change. Why has this not been reviewed by the breaking change team.

This is not a breaking change. There are no Swagger changes; it only fixes an incorrect TypeSpec usage.

ArcturusZhang (approved):

PR #41393, opened 2026-03-14
Queues: update metadata headers to be optional

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/41213

### DIFF
--- a/specification/storage/Microsoft.QueueStorage/client.tsp
+++ b/specification/storage/Microsoft.QueueStorage/client.tsp
@@ -1,6 +1,8 @@
 import "./main.tsp";
 import "@azure-tools/typespec-client-generator-core";
 
+using TypeSpec.Http;
+using Storage.Queues;
 using Azure.ClientGenerator.Core;
 
 namespace Customizations;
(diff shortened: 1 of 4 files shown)


vincenttran-msft (approved):
LGTM; with the caveat that we will need to do the same thing as Blobs in the [client.tsp](https://github.com/Azure/azure-rest-api-specs/blob/feature/blob-tsp-rust/specification/storage/Microsoft.BlobStorage/client.tsp#L188-L219). 
```typescript
/** The Set Metadata operation sets user-defined metadata for the specified blob as one or more name-value pairs. */
#suppress "@azure-tools/typespec-azure-core/no-response-body" "Existing API"
#suppress "@azure-tools/typespec-azure-core/use-standard-names" "Existing API"
@put
@route("?comp=metadata")
op setMetadataRequiredMetadataBlob is StorageOperationNoBody<
  {
    ...TimeoutParameter;
    ...MetadataHeadersRequired;
    ...LeaseIdOptionalParameter;
    ...EncryptionKeyParameter;
    ...EncryptionKeySha256Parameter;
    ...EncryptionAlgorithmParameter;
    ...EncryptionScopeParameter;
    ...IfModifiedSinceParameter;
    ...IfUnmodifiedSinceParameter;
    ...IfNoneMatchParameter;
    ...IfMatchParameter;
    ...IfTagsParameter;
  },
  {
    ...EtagResponseHeaderPrivate;
    ...LastModifiedResponseHeaderPrivate;
    ...VersionIdResponseHeaderPrivate;
    ...RequestServerEncryptedResponseHeaderPrivate;
    ...EncryptionKeySha256ResponseHeaderPrivate;
    ...EncryptionScopeResponseHeaderPrivate;
  }
>;

@@override(Blob.setMetadata, setMetadataRequiredMetadataBlob, "rust");
```

Not sure if that is something you want to handle in this PR or merge as-is and we can do a follow-up PR so I can test the codegen as well.

catalinaperalta:
Good catch! Done, ready for re-review

vincenttran-msft (approved):
Approved! Thanks for solving the overrides too 😄

PR #41394, opened 2026-03-14
setAccessPolicy make queueAcl body optional

catalinaperalta:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/41340


vincenttran-msft (approved):
Approval with same caveat as `set_metadata` wrt matching override as Blobs.

vincenttran-msft (approved):
Approved! Thanks for solving the overrides too 😄

jalauzon-msft (approved):

PR #41405, opened 2026-03-16
MPG migration - cosmosdbforpostgresql

haiyuazhang:
Migrate service cosmosdbforpostgresql to use new management plane sdk generator


specification/postgresqlhsc/resource-manager/Microsoft.DBforPostgreSQL/PostgresqlHsc/client.tsp
```diff
@@ -10,3 +10,217 @@ using Microsoft.DBforPostgreSQL;
...
+);
+
+/**
+ * Resource type used for name availability verification.
+ */
+union CheckNameAvailabilityResourceTypeReplacement {
+  CheckNameAvailabilityResourceType,
+  string,
+}
+@@alternateType(CheckNameAvailabilityResourceType,
```
live1206, line 187: As I can see in swagger, this is a fixed enum with modelAsString: false, and csharp should generate this as enum.
I guess the previous GA version was an extensible enumeration, can you check why it was extensible before since swagger should not change?

ravimeda:
Please address the comments from Copilot.

specification/postgresqlhsc/resource-manager/Microsoft.DBforPostgreSQL/PostgresqlHsc/client.tsp
haiyuazhang: The swagger does have modelAsString: false, but the old SDK generated it as an extensible struct may because modelerfour automatically promotes single-value SealedChoice to Choice (extensible).

live1206: @pshao25 could you help to verify this behavior diff in swagger and TypeSpec?

pshao25: @@alternateType(CheckNameAvailabilityResourceType, "Microsoft.DBforPostgreSQL/serverGroupsv2")

live1206: @haiyuazhang as offline discussed, please file an issue to investigate the root cause of the generator/TCGC behavior, and put the issue as comment here for later cleanup, we can workaround this for now

live1206 (approved):

PR #41436, opened 2026-03-16
to service operations add "/" 

l0lawrence:


l0lawrence:
live tests run without this

PR #41450, opened 2026-03-17
Remove api-version setting from keyvault resource-manager tspconfig.yaml

ArcturusZhang:
Remove the `api-version: "2025-05-01"` setting from the `@azure-typespec/http-client-csharp-mgmt` emitter in `specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/tspconfig.yaml`.


live1206 (approved):

ArcturusZhang:
this only changes the config, no need for ARM's review.

PR #41467, opened 2026-03-17
Add csharp client customizations for Storage MPG migration

live1206:
## Summary
Add C#-scoped client customizations and fix the `StorageTaskAssignment` resource model to support the Azure.ResourceManager.Storage migration from AutoRest to TypeSpec.

#### StorageTaskAssignment.tsp
- Convert `StorageTaskAssignment` from `extends Resource` (with manual properties bag) to `is ProxyResource<StorageTaskAssignmentProperties>` (standard ARM pattern)
- Remove obsolete `#suppress` directives and `@Http.Private.includeInapplicableMetadataInPayload`
- Add `#suppress` for `arm-put-operation-response-codes` (existing API returns 202 for LRO PUT)

#### tspconfig.yaml
- Add `@azure-typespec/http-client-csharp-mgmt` emitter configuration with `wire-path-attribute: true`

#### client.tsp — C# backward-compatibility decorators
- **`@@clientName`** (~300 decorators, csharp-scoped): Preserve existing .NET SDK API surface naming (property casing, enum value names, model/operation renames)
- **`@@alternateType`** (~60 decorators): Map TypeSpec types to prior GA .NET types (e.g. `string` → `utcDateTime` for DateTimeOffset properties, `string` → `uuid` for GUIDs, ETag/Identity/ExtendedLocation mappings)
- **`@@usage`** (~40 decorators): Set input/output usage for models that need explicit visibility
- **`@@access`** (4 decorators): Control access for `ProvisioningState` and `StorageTaskAssignmentReport`

#### Generated swagger
- Regenerated `openapi.json` for `2025-06-01` and `2025-08-01` to reflect the `StorageTaskAssignment` ProxyResource change

### Context
- Related SDK PR: Azure/azure-sdk-for-net#56896
- Issue: Azure/azure-sdk-for-net#54775


specification/storage/resource-manager/Microsoft.Storage/stable/2025-06-01/openapi.json
```diff
@@ -13527,12 +13505,9 @@
           "description": "Properties of the storage task assignment."
         }
       },
-      "required": [
-        "properties"
-      ],
       "allOf": [
         {
-          "$ref": "#/definitions/Resource"
+          "$ref": "../../../../../common-types/resource-management/v5/types.json#/definitions/ProxyResource"
```
live1206, line 13510: This is a swagger breaking change we would like to take during the csharp emitter migration. We should use the common-type ProxyResource instead of the custom defined Resource.

specification/storage/Storage.Management/StorageTaskAssignment.tsp
```diff
@@ -16,23 +16,14 @@ namespace Microsoft.Storage;
 /**
  * The storage task assignment.
  */
-#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
blueww, line 19: Could you share the reason why remove these?
Should we include the feature team for this API for review?

live1206: Because it is not needed anymore, check my comment in https://github.com/Azure/azure-rest-api-specs/pull/41467#discussion_r2991820801

specification/storage/Storage.Management/back-compatible.tsp
```diff
@@ -316,7 +316,7 @@ using Microsoft.Storage;
 );
 @@visibility(StorageTaskAssignmentsList.value, Lifecycle.Read);
 
-@@clientLocation(BlobServices.blobContainersList, BlobContainers);
+@@clientLocation(BlobServices.blobContainersList, BlobContainers, "!csharp");
```
blueww, line 319: Could you share the reason add "!csharp" here, and the below properties?

specification/storage/Storage.Management/StorageTaskAssignment.tsp
```diff
@@ -45,6 +36,7 @@ interface StorageTaskAssignments {
   /**
    * Asynchronously creates a new storage task assignment sub-resource with the specified parameters. If a storage task assignment is already created and a subsequent create request is issued with different properties, the storage task assignment properties will be updated. If a storage task assignment is already created and a subsequent create or update request is issued with the exact same set of properties, the request will succeed.
    */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "Existing API returns 202 for LRO PUT"
```
blueww, line 39: Why add this suppress?
Will is impact all client/SDK/doc gen from TSP?

blueww (approved):

ArthurMa1978 (approved):

ArthurMa1978 (approved):

PR #41472, opened 2026-03-17
Adding custom training job typespec

jayesh-tanna:
# Choose a PR Template

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/client.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/client.tsp
@@ -221,14 +221,17 @@ using Azure.AI.Projects;
 @@clientName(Toolsets.deleteToolset, "delete");
 
 // --------------------------------------------------------------------------------
-// Jobs sub-client (beta)
+// Jobs sub-client (beta, under Training)
 // --------------------------------------------------------------------------------
 
-@@clientName(Jobs.get, "get", "python");
-@@clientName(Jobs.list, "list", "python");
-@@clientName(Jobs.createOrUpdate, "create_or_update", "python");
-@@clientName(Jobs.delete, "delete", "python");
-@@clientName(Jobs.cancel, "cancel", "python");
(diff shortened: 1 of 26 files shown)


specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -219,6 +220,16 @@ using Azure.AI.Projects;
...
+// --------------------------------------------------------------------------------
+// Jobs sub-client (beta)
+// --------------------------------------------------------------------------------
+
+@@clientName(Jobs.get, "get", "python");
+@@clientName(Jobs.list, "list", "python");
+@@clientName(Jobs.createOrUpdate, "create_or_update", "python");
+@@clientName(Jobs.delete, "delete", "python");
+@@clientName(Jobs.cancel, "cancel", "python");
+
```
dargilco, line 235: The above lines are not needed. You are not changing any of the names, so the Python emitter will already do what you want (including createOrUpdate being emitted as create_or_update, as it knows to convert from CamelCase to snake_case)

specification/ai-foundry/data-plane/Foundry/relocate-beta-operations.tsp
```diff
@@ -76,4 +76,9 @@ interface Toolsets {}
 @@clientLocation(Azure.AI.Projects.Toolsets.listToolsets, Toolsets);
 @@clientLocation(Azure.AI.Projects.Toolsets.deleteToolset, Toolsets);
 
-
```
dargilco, line 79: Thank you for editing the client files!

specification/ai-foundry/data-plane/Foundry/client.tsp
jayesh-tanna: @dargilco we want to generate it only for the python language for now and we want this as preview feature. in these case also we don't require this change?

PR #41475, opened 2026-03-17
[CosmosDB] Adding new API versions 2026-03-15 stable and 2026-04-01-preview

pjohari-ms:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

This PR adds two new Cosmos DB ARM API versions:
- **`2026-03-15` (stable)**: New stable API version with TypeSpec versioning support and full example coverage.
- **`2026-04-01-preview` (preview)**: New preview API version with preview-only resources/operations annotated with `@added(Versions.v2026_04_01_preview)`, including new resource types such as fleets.

**Example quality fixes applied based on review feedback:**
- Fixed `$filter` parameter values to remove duplicate `$filter=` prefix (32 files across both versions).
- Fixed LRO response header casing: `"azure-AsyncOperation"` → `"Azure-AsyncOperation"`, `"location"` → `"Location"` (130+ files across both versions).
- Fixed `Microsoft.DocumentDb` (incorrect casing) → `Microsoft.DocumentDB` and `resourceGroups/Default` → `resourceGroups/rg1` in PrivateLinkResource examples.
- Fixed PascalCase → camelCase property names in `display` object in `CosmosDBOperationsList.json` (both versions).
- Fixed invalid ARM resource ID segment `resourceGroup/rg1` → `resourceGroups/rg1` in `CosmosDBFleetspaceAccountGet.json` (both versions).
- Fixed inconsistent `resourceGroupName` in `CosmosDBCassandraViewList.json` response resource ID.
- Fixed stale `api-version` values in LRO polling URLs across all 2026-03-15 and 2026-04-01-preview examples.
- Added LRO polling headers (`Azure-AsyncOperation`, `Location`) to the previously empty `"202": {}` response in `CosmosDBTableCreateUpdate.json`.
- Fixed literal `keyspaceName`/`viewName` segments in `CosmosDBCassandraViewThroughputGet.json` resource ID to match request parameter values.

## Getting help

  and https://aka.ms/ci-fix.


gary-x-li:
- Please fill out the **Purpose of this PR** and **Due diligence checklist** sections.
- I only see minimal changes to `main.tsp` and `models.tsp`. Did you include all TypeSpec changes?
- Please fix the TypeSpec Validation failure

pjohari-ms:
@copilot apply changes based on the comments in [this thread](https://github.com/Azure/azure-rest-api-specs/pull/41475#pullrequestreview-4033811675)

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-03-15/CosmosDBOperationsList.json
```diff
@@ -0,0 +1,24 @@
...
+      "body": {
+        "value": [
+          {
+            "name": "operationName",
+            "display": {
+              "Description": "description",
+              "Operation": "operationName",
+              "Provider": "providerName",
+              "Resource": "resourceName"
+            }
```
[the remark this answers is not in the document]
pjohari-ms: Typespec expects pascalCase.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-03-15/CosmosDBCassandraKeyspaceCreateUpdate.json
```diff
@@ -0,0 +1,38 @@
...
+        "tags": {},
+        "properties": {
+          "resource": {
+            "id": "keyspaceName"
+          }
+        }
+      }
+    },
+    "202": {}
+  },
```
ravimeda, line 40: **[NEW] 🔴 Blocking** **[EX-LRO-HEADERS]** — Empty `"202": {}` response with no LRO polling headers.

This is a stable (`2026-03-15`) example for an LRO operation. The `202` response MUST include polling headers (`Azure-AsyncOperation` and/or `Location`). An empty 202 makes the LRO contract ambiguous for ARM and SDKs.

This same issue affects multiple stable example files with empty 202 responses, including:
- `CosmosDBCassandraTableCreateUpdate.json`
- `CosmosDBCassandraKeyspaceThroughputUpdate.json`
- `CosmosDBCassandraTableThroughputUpdate.json`
- `CosmosDBGremlinDatabaseCreateUpdate.json`
- `CosmosDBGremlinDatabaseThroughputUpdate.json`
- `CosmosDBDataTransferServiceCreate.json`
- `CosmosDBGraphAPIComputeServiceCreate.json`

The PR description mentions fixing `CosmosDBTableCreateUpdate.json`, and the `2026-04-01-preview` examples already have the correct pattern. Please apply the same fix to these remaining stable example files.

**Fix:** Add LRO polling headers to the 202 response:
```json
"202": {
  "headers": {
    "Azure-AsyncOperation": "https://management.azure.com/subscriptions/{subscriptionId}/providers/Microsoft.DocumentDB/locations/{location}/operationsStatus/{operationId}?api-version=2026-03-15",
    "Location": "https://management.azure.com/subscriptions/{subscriptionId}/providers/Microsoft.DocumentDB/locations/{location}/operationsStatus/{operationId}?api-version=2026-03-15"
  }
}
```

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-03-15/CosmosDBDatabaseAccountCreateMin.json
```diff
@@ -0,0 +1,116 @@
...
+          "networkAclBypass": "None",
+          "networkAclBypassResourceIds": [],
+          "instanceId": "d9b26648-2f53-4541-b3d8-3044f4f9810d",
+          "createMode": "Default",
+          "backupPolicy": {
+            "type": "Periodic",
+            "periodicModeProperties": {
+              "backupIntervalInMinutes": 240,
+              "backupRetentionIntervalInHours": 720,
+              "backupStorageRedundancy": "Geo"
```
ravimeda, line 88: **[NEW] 🔴 Blocking** **[EX-PAYLOAD]** — Response body contains `"analyticalStorageConfiguration": null`.

ARM properties that are objects MUST NOT be set to `null` in examples. Either omit the property entirely or provide a valid object value. Setting known ARM object properties to `null` fails schema validation.

**Fix:** Remove the `"analyticalStorageConfiguration": null` line from this response body.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/ChaosFaultResource.tsp
ravimeda, line 1: **[NEW] 🔴 Blocking** **[RPC-SUPPRESS-GA]** — This PR introduces a **GA stable** version (`2026-03-15`). Per ARM guidelines, suppressions carried from preview are **not automatically acceptable in GA** — each suppression requires a concrete technical justification.

The TypeSpec source files across this project contain numerous `#suppress` directives with generic justifications like `"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"`. The most concerning for a GA release include:

1. `#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state"` — used across many resource property models. For GA, all resources SHOULD have a proper `provisioningState` property following ARM patterns. 
2. `#suppress "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation"` — on `ChaosFaultResource`, `CopyJobGetResults`, `DataTransferJobGetResults`. Missing DELETE operations on resources is a significant gap for GA.

**Fix:** Update each suppress justification with a concrete, specific reason explaining why the violation cannot be resolved for this GA release. Generic "FIXME" text is insufficient for GA review.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
ravimeda, line 1: **[EXISTING]** **[SEC-SECRET-DETECT]** — Multiple properties suppress `secret-prop` lint rule instead of using the `@secret` decorator.

This is a security concern: suppressing the lint rule hides the secret nature of these properties from SDK generation, documentation, and client-side handling. The affected properties include:

| Property | Model |
|---|---|
| `primaryMasterKey` | `DatabaseAccountListKeysResult` |
| `secondaryMasterKey` | `DatabaseAccountListKeysResult` |
| `primaryReadonlyMasterKey` | `DatabaseAccountListReadOnlyKeysResult` |
| `secondaryReadonlyMasterKey` | `DatabaseAccountListReadOnlyKeysResult` |
| `authToken` | `NotebookWorkspaceConnectionInfoResult` |
| `serviceUserPassword` | `AuthenticationMethodLdapProperties` |
| `password` | `MongoUserDefinitionResource` |

This issue also exists in the previous version (`2025-11-01-preview`) and was **not introduced by this PR**. However, since this is now a **GA release**, it is strongly recommended to fix this.

**Fix:** Replace `#suppress "@azure-tools/typespec-azure-resource-manager/secret-prop"` with the `@secret` decorator on each property. The pattern is already correctly used on `initialCassandraAdminPassword` in `ClusterResourceProperties`.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-03-15/CosmosDBCollectionPartitionGetMetrics.json
```diff
@@ -0,0 +1,59 @@
...
+              "localizedValue": "Max RUs Per Second"
+            }
+          }
+        ]
+      }
+    }
+  },
+  "operationId": "CollectionPartition_ListMetrics",
+  "title": "CosmosDBDatabaseAccountRegionGetMetrics"
+}
```
ravimeda, line 59: **[NEW] 🟡 Warning** **[EX-TITLE]** — Example title is `"CosmosDBDatabaseAccountRegionGetMetrics"` but the operationId is `"CollectionPartition_ListMetrics"`. The title should match the operation being demonstrated — otherwise it's misleading for documentation consumers.

**Fix:** Change `"title"` to `"CosmosDBCollectionPartitionGetMetrics"`.

Note: similar title mismatches exist in:
- `CosmosDBCollectionPartitionGetUsages.json` — title says "CosmosDBCollectionGetUsages" but operationId is "CollectionPartition_ListUsages"
- `CosmosDBCollectionPartitionRegionGetMetrics.json` — title says "CosmosDBDatabaseAccountRegionGetMetrics" but operationId is "CollectionPartitionRegion_ListMetrics" #Resolved

ravimeda (changes requested):
## ARM API Review: `cosmos-db/2026-03-15` (stable) and `cosmos-db/2026-04-01-preview`

**Previous version:** `2025-11-01-preview`

### Summary

| Category | Count |
|---|---|
| **New blocking issues** | **3** |
| **Existing blocking issues** | **1** |
| New warnings | 1 (+ 2 related) |
| Suggestions | 0 |

### Key findings:

1. **🔴 Empty LRO 202 responses (stable)** — Multiple `2026-03-15` stable example files have `"202": {}` with no polling headers. The PR already fixed `CosmosDBTableCreateUpdate.json` and all `2026-04-01-preview` examples, but missed applying the same fix to ~7 other stable examples.

2. **🔴 Null object property in example** — `CosmosDBDatabaseAccountCreateMin.json` sets `"analyticalStorageConfiguration": null` which fails ARM schema validation.

3. **🔴 FIXME suppressions in GA release** — TypeSpec files contain generic "FIXME" suppression justifications. For a GA release, each suppression needs a concrete technical justification per ARM guidelines.

4. **SEC-SECRET-DETECT (existing)** — Database keys and passwords suppress the `secret-prop` lint rule instead of using `@secret`. This was pre-existing but should be addressed for GA.

5. **🟡 Example title mismatches** — 3 collection partition metric examples have titles that don't match their operationId.

Several previously flagged issues (LRO header casing, `$filter` prefixes, namespace casing, etc.) appear to have been addressed based on resolved/outdated comment threads — kudos on the cleanup work already done.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/ChaosFaultResource.tsp
pjohari-ms: @ravimeda this PR introduces 2 new API versions in one go. The typespec conversion was not completely done by the Typespec team and hence it has a new GA version as well. Is this fine from a review/pattern perspective?

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: This is a continuation of the previous GA releases. Should we still make this change?

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -73,14 +75,27 @@ model GarnetClusterResourceProperties {
...
   /** If the data center has Availability Zone support, apply it to the Virtual Machine ScaleSet that host the garnet cluster virtual machines. */
   availabilityZone?: boolean;
 
+  /** The authentication method used for the Garnet cluster. */
+  @added(Versions.v2026_04_01_preview)
+  authenticationMethod?: GarnetAuthenticationType;
+
+  /** Flag to indicate if persistence is enabled for the Garnet cluster. */
+  @added(Versions.v2026_04_01_preview)
+  persistence?: boolean;
```
mentat9, line 98: Prefer union vs. boolean so all states have a name, e.g. persistenceType: { Ephemeral, Persistent } or similar.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -199,6 +224,18 @@ model GarnetClusterResourcePatchProperties {
...
+   * The authentication method used for the Garnet cluster.
+   */
+  @added(Versions.v2026_04_01_preview)
+  authenticationMethod?: GarnetAuthenticationType;
+
+  /**
+   * Flag to indicate if persistence is enabled for the Garnet cluster.
+   */
+  @added(Versions.v2026_04_01_preview)
+  persistence?: boolean;
```
mentat9, line 239: Prefer union vs. boolean.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -1721,6 +1762,11 @@ model DatabaseAccountGetProperties {
    */
   #suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
   throughputPoolMaxConsumableRUs?: int64;
+
+  /**
+   * Flag to indicate enabling/disabling of hierarchical partition key ID last level enforcement on the account.
+   */
+  enforceHierarchicalPartitionKeyIdLastLevel?: boolean;
```
mentat9, line 1811: Prefer union vs. boolean for many reasons. Consider something like this: accountEnforcement: { None, HierarchicalPartitionKeyLastLevel } or a better approach you come up with.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -2485,6 +2541,11 @@ model DatabaseAccountUpdateProperties {
    * Flag to indicate if All Versions and Deletes Change feed feature is enabled on the account
    */
   enableAllVersionsAndDeletesChangeFeed?: boolean;
+
+  /**
+   * Flag to indicate enabling/disabling of hierarchical partition key ID last level enforcement on the account.
+   */
+  enforceHierarchicalPartitionKeyIdLastLevel?: boolean;
```
mentat9, line 2601: Prefer union vs. boolean.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -2707,6 +2768,11 @@ model DatabaseAccountCreateUpdateProperties {
...
```
mentat9, line 2838: Prefer union vs boolean. Large parts of this API have evolved into a boolean bitvector which makes it harder to read and understand (what does put(true, true, false, true, false, false) mean compared with put(true, false, true, true, false, true)?). Unions offer a friendlier model.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -2835,6 +2901,12 @@ model DatabaseAccountRegenerateKeyParameters {
    * The access key to regenerate.
    */
   keyKind: KeyKind;
+
+  /**
+   * Optional flag indicating whether to skip account keys last usage check.
+   */
+  @added(Versions.v2026_04_01_preview)
+  skipAccountKeysLastUsageCheck?: boolean;
```
mentat9, line 2972: Could be something more like: accountKeysCheckPolicy: { None, LastUsageCheck } or something similar that better matches the RP behavior and future plans.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
mentat9: Yes, please mark properties containing secrets with the @secret decorator.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
lnajaroen: For this we want to keep as boolean. This is a feature users can toggle on and off. We understand the recommendation for union and will try to do that for new features moving forward.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
lnajaroen: For this we want to keep as boolean. This is a feature users can toggle on and off. We understand the recommendation for union and will try to do that for new features moving forward.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
lnajaroen: For this we want to keep as boolean. This is a feature users can toggle on and off. We understand the recommendation for union and will try to do that for new features moving forward.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: Keeping it as boolean for now after internal discussion.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: Same, keeping it bas boolean for now.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: Same, keeping it bas boolean for now.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: Done.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/ChaosFaultResource.tsp
sandipsh: From ARM API review perspective, this justification is not sufficient for a GA tag.

Reason: "TypeSpec conversion not completely done" explains why this happened, but it does not satisfy RPC-SUPPRESS-GA criteria for a stable version. For GA, carried suppressions must be justified one-by-one with concrete technical rationale at each location.

What is acceptable for this PR:
1. Keep the GA version and replace each generic FIXME suppression with specific justification tied to that exact model/operation and why it cannot be fixed safely in this GA release.
2. For no-resource-delete-operation suppressions, explicitly justify resource semantics (for example, read-only operation-result style resources) and why DELETE is not part of the contract.
3. For arm-resource-provisioning-state suppressions, either add proper provisioningState pattern or provide explicit resource-specific reasoning for exception.

If that level of per-suppression justification is not ready, the safer path is to avoid shipping the GA tag in this PR and keep the changes preview-only until suppressions are cleaned up.

Please update this thread once the suppressions are either fixed or justified with specific reasons per site.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/CopyJobGetResults.tsp
```diff
@@ -26,6 +28,7 @@ model CopyJobGetResults
 }
```
sandipsh, line 30: **[Follow-up]** **[RPC-SUPPRESS-GA]**

I see this interface is now gated with `@added(Versions.v2026_04_01_preview)`, which helps separate it from the GA `2026-03-15` surface.

Please confirm explicitly:
1. This `CopyJobs` resource is **not** emitted in `2026-03-15` (GA), only in preview.
2. The `no-resource-delete-operation` suppression here is intentional for this resource shape.
3. Replace the current `FIXME` suppression reason with a concrete, resource-specific justification (preview or GA as applicable), so we can close the GA suppression concern cleanly.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/DataTransferJobGetResults.tsp
```diff
@@ -26,6 +28,7 @@ model DataTransferJobGetResults
 }
```
sandipsh, line 30: **[Follow-up]** **[RPC-SUPPRESS-GA]**

I see this interface is now gated with `@added(Versions.v2026_04_01_preview)`, which helps separate it from the GA `2026-03-15` surface.

Please confirm explicitly:
1. This `DataTransferJobs` resource is **not** emitted in `2026-03-15` (GA), only in preview.
2. The `no-resource-delete-operation` suppression here is intentional for this resource shape.
3. Replace the current `FIXME` suppression reason with a concrete, resource-specific justification (preview or GA as applicable), so we can close the GA suppression concern cleanly.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -3323,6 +3434,7 @@ model GraphResourcesListResult {
  * The properties of an Azure Cosmos DB SQL database
```
sandipsh, line 3459: **[Follow-up]** **[RPC-SUPPRESS-GA]**

`@added(Versions.v2026_04_01_preview)` was introduced broadly, which may scope many items to preview-only.

Please confirm in this thread whether any remaining `arm-resource-provisioning-state` suppressions with generic `FIXME` justifications still affect `2026-03-15` (GA).

If they do, those suppressions still need concrete, per-site GA justification (or a model fix). If they are preview-only, please confirm that explicitly so this can be evaluated as addressed for GA scope.

pjohari-ms: @sandipsh All these suppressions are something that has been introduced during typespec migration, our service is already GA and has GA api versions. We can add these suppressions in a followup PR if needed but none of these FIXME suppressions have been introduced in this PR.

pjohari-ms:
1. All FIXME would be added in a later PR, whether just suppression messages or fixes in later API versions.
2. The current PR and the tags have been verified against python SDK generation, and works correctly and does not add any new changes. Functionally correct.
3. This change HAS to go in a single PR, due to reasons mentioned in the teams chat. Next API version onwards we will introduce separate versioning.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
pjohari-ms: Discussed in the API review office hours. Will be added in a later version.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/CopyJobGetResults.tsp
pjohari-ms: Valid comment. CopyJobs  is a preview only feature, and the no resource delete operation is intentional. We will add this as a fixme in the followup PR with other justifications, as discussed.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/DataTransferJobGetResults.tsp
pjohari-ms: Valid comment. DataTransferJobs is a preview only feature, and the no resource delete operation is intentional. We will add this as a fixme in the followup PR with other justifications, as discussed.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/ChaosFaultResource.tsp
sandipsh: Following up since there is no response to my April 28 comment. The ChaosFaultResources interface is now gated with @added(Versions.v2026_04_01_preview) which addresses the no-resource-delete-operation concern for GA. What still needs attention: the chaosFaultResource model has two FIXME suppressions that are NOT preview-gated and affect the GA surface: (1) casing-style FIXME and (2) resource-name FIXME. Since this is a brownfield TypeSpec conversion and the camelCase naming predates this PR, the fix is simple: replace the FIXME text with a concrete reason, e.g. chaosFaultResource preserves camelCase naming from the prior swagger-defined stable API to maintain backward compatibility. Two one-line updates. Everything else has been accepted as follow-up PR items. Please address these two before GA approval.

sandipsh: Follow-up: still need concrete suppression justifications for chaosFaultResource model (casing-style and resource-name FIXME suppressions are not preview-gated).

sandipsh: Following up since there is no response to my April 28 comment.

Good to see ChaosFaultResources is gated with @added(Versions.v2026_04_01_preview) - that addresses the 
o-resource-delete-operation concern for GA.

One remaining ask: the chaosFaultResource **model** still has two FIXME suppressions that are **not** preview-gated (model is GA surface):

`
#suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification..."
#suppress "@azure-tools/typespec-azure-resource-manager/resource-name" "FIXME: Update justification..."
`

Since this is a brownfield TypeSpec conversion and camelCase naming predates this PR, the justification is straightforward - e.g.: *"Preserves camelCase naming from prior swagger-defined stable API to maintain backward compatibility; changing would be a breaking rename."*

Small ask - two one-line text updates. All other items (no-resource-delete, provisioningState) have been accepted as follow-up PR items. Please update these before GA approval.

pjohari-ms: I have added the justifications, as well as marked the models as preview gated properly.

pjohari-ms:
@sandipsh Thanks for signing off on the PR. Could you please also provide the suppression review? This is related to the _rid suppressions that we added in the readme.md
The reasoning that we discussed for this in the meeting was to _rid has a particular meaning for cosmosdb and we have previous approved suggestions for the same property naming as well, in the same file.

cc @sachim_microsoft

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
```diff
@@ -0,0 +1,760 @@
...
+    ...Azure.ResourceManager.ResourceGroupParameter,
+    ...Azure.ResourceManager.LocationResourceParameter,
+    ...Azure.ResourceManager.ApiVersionParameter,
+  ): ArmResponse<SoftDeletedDatabaseAccountsListResult> | Azure.ResourceManager.CommonTypes.ErrorResponse;
+
+  /**
+   * Restores a soft-deleted Azure Cosmos DB database account.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Explicit operationId needed for separate Restore/Purge SDK methods"
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DocumentDB/locations/{location}/softDeletedDatabaseAccounts/{accountName}?softDeleteActionKind=RestoreSoftDeletedResource")
```
ravimeda, line 316: **[NEW] 🔴 Blocking** **[RPC-Uri-V1-01]** — Query strings embedded in route paths

Six restore/purge operations embed query string parameters directly in the `@route` path (e.g., `?softDeleteActionKind=RestoreSoftDeletedResource`). The TypeSpec emitter writes these verbatim into the OpenAPI `paths` object, producing path keys like:

`/subscriptions/{subscriptionId}/.../softDeletedDatabaseAccounts/{accountName}?softDeleteActionKind=PermanentDeleteResource`

**Violation:** OpenAPI path keys MUST NOT contain query strings. This will break ARM validation, SDK generation, and client tooling. The same `softDeleteActionKind` is **also** declared as a `@query` parameter, creating a duplicate definition.

**Affected lines:** 409, 439, 529, 566, 677, 721

**Fix:** Remove the `?softDeleteActionKind=...` suffix from all six `@route` decorators. Use only the `@query("softDeleteActionKind")` parameter. To distinguish restore from purge on the same path, consider using **separate action routes** (e.g., `.../restore` and `.../purge` as POST actions) instead of overloading the same DELETE path with different query values.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
```diff
@@ -0,0 +1,760 @@
...
+  @delete
```
ravimeda, line 317: **[NEW] 🔴 Blocking** **[RPC-Delete-V1-01]** — HTTP DELETE used for restore operations

The restore operations (`restore`) use `@delete` (HTTP DELETE), but restoring a soft-deleted resource is a **creative/mutating action**, not a destruction. Using DELETE for restore violates HTTP method semantics and ARM RPC conventions. ARM SDKs and documentation will label these as "delete" operations, confusing customers.

**Affected lines:** 410, 530, 678 (restore operations)

**Fix:** Change restore operations from `@delete` to `@post` and model them as ARM actions (e.g., `@action("restore")`). The purge operations can remain as `@delete` since they permanently remove the resource.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
```diff
@@ -0,0 +1,760 @@
...
+   * A subset of properties of the underlying database account resource.
+   */
+  resource?: SoftDeletedDatabaseAccountResource;
+}
+
+/**
+ * A Azure Cosmos DB soft-deleted database account.
+ */
+@added(Versions.v2026_04_01_preview)
+model SoftDeletedDatabaseAccountGetResult {
```
ravimeda: **[NEW] 🔴 Blocking** **[ARM Resource Model Pattern]** — Soft-delete result models manually define ARM envelope instead of extending base types

`SoftDeletedDatabaseAccountGetResult`, `SoftDeletedSqlDatabaseGetResult`, and `SoftDeletedSqlContainerGetResult` manually define `id`, `name`, `type`, `location`, and `properties` instead of extending ARM resource base types (`ProxyResource` or `TrackedResource`). This bypasses ARM resource validation, omits `systemData`, and is inconsistent with the ARM resource modeling pattern used elsewhere in this spec.

**Affected models:** lines 109, 198, 297

**Fix:** Extend `Azure.ResourceManager.ProxyResource<T>` (or `TrackedResource<T>` if these have location):
`	ypespec
model SoftDeletedDatabaseAccountGetResult
  is Azure.ResourceManager.ProxyResource<SoftDeletedDatabaseAccountProperties> {
  // ...
}
`

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-04-01-preview/CosmosDBSoftDeletedDatabaseAccountGet.json
```diff
@@ -0,0 +1,28 @@
...
+    "api-version": "2026-04-01-preview",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000",
+    "resourceGroupName": "rg1",
+    "location": "West US",
+    "accountName": "softdeleted-cosmosdb-1"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.DocumentDB/locations/westus/softDeletedDatabaseAccounts/softdeleted-cosmosdb-1",
```
ravimeda: **[NEW] 🔴 Blocking** **[EX-DESCRIPTIVE-VALUES]** — Example response `id` is missing `resourceGroups` segment

The example response `id` is:
`/subscriptions/00000000.../providers/Microsoft.DocumentDB/locations/westus/softDeletedDatabaseAccounts/softdeleted-cosmosdb-1`

But the operation route includes `resourceGroupName` as a required parameter, so the response ID should include it:

**Fix:**
`json
"id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.DocumentDB/locations/westus/softDeletedDatabaseAccounts/softdeleted-cosmosdb-1"
`

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-04-01-preview/CosmosDBSoftDeletedDatabaseAccountRestore.json
```diff
@@ -0,0 +1,20 @@
...
+    "accountName": "softdeleted-cosmosdb-2",
+    "resourceGroupName": "rg2",
+    "location": "West US",
+    "api-version": "2026-04-01-preview",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "202": {
+      "headers": {
+        "azure-asyncoperation": "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.DocumentDB/locations/westus/operationResults/restore-operation-id?api-version=2026-04-01-preview",
```
ravimeda: **[NEW] 🟡 Warning** **[LRO Header Casing]** — LRO response headers use incorrect lowercase casing

LRO response headers use lowercase `azure-asyncoperation` and `location` instead of the ARM-standard PascalCase `Azure-AsyncOperation` and `Location`. This is inconsistent with the rest of this PR, which explicitly fixed the same casing issue in 130+ other example files.

**Fix:**
`json
"Azure-AsyncOperation": "https://...",
"Location": "https://..."
`

Apply the same fix to all soft-delete restore/purge examples across both versions.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
```diff
@@ -0,0 +1,760 @@
...
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "x-ms-mutability is required for ARM location property"
+  @extension("x-ms-mutability", #["read", "create"])
+  location?: string;
+}
+
+/**
+ * The List operation response, that contains the soft-deleted database accounts and their properties.
+ */
+@added(Versions.v2026_04_01_preview)
+model SoftDeletedDatabaseAccountsListResult {
```
ravimeda: **[NEW] 🟡 Warning** **[Pagination Pattern]** — Manual pagination models instead of `Azure.Core.Page<T>`

`SoftDeletedDatabaseAccountsListResult`, `SoftDeletedSqlDatabasesListResult`, and `SoftDeletedSqlContainersListResult` manually define `value` and `nextLink` properties instead of using `Azure.Core.Page<T>` which is the standard pattern used elsewhere in this spec (e.g., `chaosFaultListResponse is Azure.Core.Page<chaosFaultResource>`).

**Affected lines:** 147, 236, 335

**Fix:** Use the standard pagination template:
`	ypespec
model SoftDeletedDatabaseAccountsListResult is Azure.Core.Page<SoftDeletedDatabaseAccountGetResult>;
`

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -1708,19 +1782,33 @@ model DatabaseAccountGetProperties {
...
    * When this account is part of a fleetspace with throughput pooling enabled, this is the maximum additional throughput (RU/s) that can be consumed from the pool, summed across all shared throughput databases and dedicated throughput containers in the account for 1 region.  READ ONLY.
    */
   #suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  @added(Versions.v2026_04_01_preview)
   throughputPoolMaxConsumableRUs?: int64;
+
+  /**
+   * Flag to indicate enabling/disabling of hierarchical partition key ID last level enforcement on the account.
+   */
+  enforceHierarchicalPartitionKeyIdLastLevel?: boolean;
```
ravimeda, line 1811: **[NEW] 🟡 Warning** **[Version Annotation Consistency]** — New property missing `@added` decorator

`enforceHierarchicalPartitionKeyIdLastLevel` is a **new property** (not present in the previous version `v2025_11_01_preview`) but lacks an `@added` decorator. It appears in the stable `2026-03-15` version without any version gating. Every other new property in this PR is annotated with `@added(Versions.v2026_04_01_preview)`.

If this property is intentionally stable (available from `v2026_03_15`), this is fine — but please confirm this is deliberate and not an oversight. The same applies at lines 2601 and 2838 (in `DatabaseAccountUpdateProperties` and `DatabaseAccountCreateUpdateProperties`).

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
```diff
@@ -113,6 +129,16 @@ union AllocationState {
   /** Deallocated implies virtual machines and resources are deallocated. */
   Deallocated: "Deallocated",
 }
+
+/** The authentication method used for the Garnet cluster. */
+@added(Versions.v2026_04_01_preview)
+union GarnetAuthenticationType {
```
ravimeda, line 135: **[NEW] 💡 Suggestion** **[Enum Completeness]** — `GarnetAuthenticationType` union has only one value

`GarnetAuthenticationType` has only a single member `Entra`. A single-value extensible enum is unusual and may indicate an incomplete design. Consider whether other auth methods (e.g., `AccessKey`, `None`) should be included for completeness, or document why only `Entra` is supported.

If this is intentional (e.g., the only supported method at launch, with more planned), then this is fine as-is since the union is extensible via `string`.

sachim: @ravimeda  confirmed with the team, this is as expected.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/models.tsp
sachim: @ravimeda  this property is intentionally stable, confirmed with the team

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
pmoana-ms: This is not possible for us to do currently. Based on SDK limitations with underlying GETs, we were recommended by @pjohari-ms to use DELETE. This would require a huge refactor and rollout.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
pmoana-ms: Not sure this comment applies to us either. I validated this with copilot

```
Current @route("...?param=value") correctly emits to Azure's documented x-ms-paths extension. Removing it would collide two @delete ops on same path or force consolidation that loses team's separate operationIds. 
```

We want the SDK to generate two _separate_ APIs, one for PURGE and one for RESTORE.

pmoana-ms: Addressed in #42950 — but as a **reject** with citations rather than a code change. After investigation:

1. The actual [`RPC-Uri-V1-01` rule](https://github.com/Azure/azure-openapi-validator/blob/main/docs/path-contains-subscription-id.md) (`PathContainsSubscriptionId`) is about whether `subscriptionId` is in the path — not about query strings in path keys. Soft-delete paths already include `subscriptionId`.

2. The current `@route("...?softDeleteActionKind=...")` correctly emits to Azure's documented [`x-ms-paths` extension](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-paths). Verified — see `preview/2026-04-01-preview/openapi.json` `"x-ms-paths"` block. The autorest docs explicitly call this out as the official mechanism for "Path Item Object to have query parameters for non pure REST APIs."

3. Removing the query string from `@route` while keeping `@delete` would either collide two `@delete` ops on the same path key, or force consolidation into a single op that loses the team's intentional separate `SoftDeletedDatabaseAccounts_Restore` / `SoftDeletedDatabaseAccounts_Purge` operationIds (preserved deliberately via `@operationId` and the existing `no-openapi` suppress at line 408).

The current pattern is correct for this scenario. cc @ravimeda for awareness.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-04-01-preview/CosmosDBSoftDeletedDatabaseAccountRestore.json
pmoana-ms: Fixed in #42950 (commit 275220a). Replaced lowercase `azure-asyncoperation`/`location` with PascalCase `Azure-AsyncOperation`/`Location` in all 12 soft-delete restore/purge example files (6 distinct × 2 mirror locations). LRO header URL **values** intentionally remain subscription+location-scoped without `/resourceGroups/` to match the operation return contract.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-04-01-preview/CosmosDBSoftDeletedDatabaseAccountGet.json
pmoana-ms: Fixed in #42950 (commit 275220a). Added the missing `/resourceGroups/rg1/` segment to the response body `id` field in `CosmosDBSoftDeletedDatabaseAccountGet.json` and audited siblings — `CosmosDBSoftDeletedDatabaseAccountListByLocation.json` had the same issue and was also fixed (both in `examples/` and `preview/.../examples/` mirrors). The other soft-delete examples already had correct `id` values.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
pmoana-ms: Fixed in #42950 (commit 153e0dc). The three list result models now use `is Azure.Core.Page<T>`, matching the standard pattern (e.g., `chaosFaultListResponse`). Verified the regenerated swagger emits the equivalent `value` array + `nextLink` paginated shape.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
pmoana-ms: Addressed in #42950 — but as a **reject**. `@delete` semantics for restore/purge are intentional per team design.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/SoftDeletedResources.tsp
pmoana-ms: Fixed in #42950 (commit 153e0dc). The three soft-delete result models now extend `Azure.ResourceManager.ProxyResource<T>` with the standard `ResourceNameParameter` mixin. ARM envelope (`id`, `name`, `type`, `systemData`) now comes from the standard template instead of being defined manually. Verified the regenerated swagger uses `allOf: [#/definitions/ProxyResource]` for all three models.

specification/cosmos-db/resource-manager/Microsoft.DocumentDB/DocumentDB/examples/2026-04-01-preview/CosmosDBSoftDeletedDatabaseAccountGet.json
pmoana-ms: **Update — reverting the fix** (commit 332582f on PR #42950).

After review with the service team, this comment is **incorrectly applied**. The soft-delete API is registered in the ARM manifest as a Proxy resource scoped to **subscription+location** (NOT to a resource group). Its canonical resource id therefore should NOT include a `/resourceGroups/{rg}/` segment, even though the Get/Restore/Purge operations accept `resourceGroupName` as a request URL context parameter for permission scoping.

This matches the established Azure soft-delete pattern across the platform:
- `Microsoft.KeyVault/locations/{loc}/deletedVaults/{name}` — subscription+location scoped
- `Microsoft.Storage/locations/{loc}/deletedAccounts/{name}` — subscription+location scoped

The reviewer agent's reasoning conflated "request URL path" with "resource identity" — for these resources, those are intentionally different.

The original example ids (without `/resourceGroups/`) were correct. cc @ravimeda for awareness.

pmoana-ms: I'm pushing back on this as well. URIs should not be changed for SoftDelete. Not only is this because they were working prior to the type-spec merge, but the assumption here is that we are using a non-proxy resource.

The entire API for SoftDelete is through ProxyResource. The Resource has a special registration in the ARM Manifest. The way it works currently is the way it should stay. We should not be making changes here unless something is broken.

pmoana-ms: **Update — re-applying the fix** (commit 9dde6fa on PR #42950).

After cross-checking against the actual handler bindings in `CosmosDB/Product/Microsoft.Azure.Documents/Management/Services/ResourceProvider/ResourceProviderHttpRuntime.cs` (L1024-1027), my earlier rationale for rejecting this comment was wrong.

The soft-delete DatabaseAccount Get/Restore/Purge operations are bound to `RPPaths.SoftDeletedDatabaseAccountRoot`:

`/subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.DocumentDB/locations/{loc}/softDeletedDatabaseAccounts/{name}`

— which **does** include `/resourceGroups/{rg}/`. So the canonical resource id MUST include it, otherwise clients can't call Get on items returned by ListByLocation. This also matches the pre-existing pattern in `CosmosDBSoftDeletedSqlContainerGet.json` and `CosmosDBSoftDeletedSqlDatabaseGet.json` (which always had `/resourceGroups/` in their response ids).

So this comment is **valid and the fix is now back in PR #42950**. Thanks for the patience — apologies for the back-and-forth. cc @ravimeda

pjohari-ms:
@ravimeda Resolved the comments, can we remove the changes requested block? Its been signed off as well and it looks like every copilot review would generate more comments which have been either resolved or similar comments to that have been resolved.
Let me know if there are any big/core concerns with this PR.

ravimeda:
> @ravimeda Resolved the comments, can we remove the changes requested block? Its been signed off as well and it looks like every copilot review would generate more comments which have been either resolved or similar comments to that have been resolved. Let me know if there are any big/core concerns with this PR.

@pjohari-ms See https://aka.ms/ci-fix for addressing the breaking change.

ravimeda (approved):

pjohari-ms:
@tadelesh Is the go breaking change label valid here? Curious since the SDK validation is passing.

PR #41541, opened 2026-03-18
Split ExpiryTimeHeader

gracewilcox:
Splitting `ExpiryTimeHeader` into `ExpiryTimeParameter` and `ExpiryTimeResponseHeader`. No functional change here but need to have the request and response as different definitions, so I can apply client.tsp customizations to just the request or response header.

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/models.tsp
+++ b/specification/storage/Microsoft.BlobStorage/models.tsp
@@ -1557,8 +1557,18 @@ alias RehydratePriorityHeader = {
   rehydratePriority?: RehydratePriority;
 };
 
+/** The expires parameter. */
+alias ExpiryTimeParameter = {
+  /** The time this blob will expire. */
+  #suppress "@azure-tools/typespec-azure-core/known-encoding" "Existing API"
+  @clientName("ExpiresOn")
+  @encode("rfc7231")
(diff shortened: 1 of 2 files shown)


specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -1557,8 +1557,18 @@ alias RehydratePriorityHeader = {
...
 };
 
+/** The expires parameter. */
+alias ExpiryTimeParameter = {
+  /** The time this blob will expire. */
+  #suppress "@azure-tools/typespec-azure-core/known-encoding" "Existing API"
+  @clientName("ExpiresOn")
+  @encode("rfc7231")
+  @header("x-ms-expiry-time")
+  expiryTime?: utcDateTime;
```
jalauzon-msft: This change may have come while this PR was pending but we removed the `clientName` and changed the model name to `expiresOn`. I assume we should do the same thing here.

gracewilcox: Thank you! Fixed in [dab4128](https://github.com/Azure/azure-rest-api-specs/pull/41541/commits/dab4128649103ecf508da0950a363fb6bf0b7ded)

jalauzon-msft (approved):

PR #41586, opened 2026-03-19
feat: Add session file operations TypeSpec for hosted agents

ankitsul:
## Summary

Add TypeSpec definitions for session-scoped file operations on hosted agent sandboxes (stream-only).

### Operations (4 CRUDL)

| Method | Route | Description |
|--------|-------|-------------|
| `PUT` | `/agents/{name}/sessions/{id}/files/content?path=` | Upload file (binary stream) |
| `GET` | `/agents/{name}/sessions/{id}/files/content?path=` | Download file (binary stream) |
| `GET` | `/agents/{name}/sessions/{id}/files?path=` | List directory (non-recursive, single level) |
| `DELETE` | `/agents/{name}/sessions/{id}/files?path=` | Delete file/directory |

### Models (3)
- `SessionFileWriteResponse` — path (relative to $HOME), bytes_written (safeint)
- `SessionDirectoryEntry` — name (@maxLength 255), size (safeint), is_directory, modified_time
- `SessionDirectoryListResponse` — path (relative to $HOME), entries[]

### Preview Gating
- `@removed(Versions.v1)` on interface and all models — hidden from GA SDK
- `x-ms-foundry-meta` with `required_previews: hosted_agents_v1_preview` on each operation
- `FoundryDataPlanePreviewOperation` requires `Foundry-Features: HostedAgents=V1Preview` header

### Key Design Decisions
- **Sessions under agents (not versions)** — `/agents/{name}/sessions/` gives optionality to mutate the associated agent version independently of the session lifecycle
- **`/files/content` for binary, `/files` for collection** — `PUT/GET /files/content?path=` for upload/download, `GET /files?path=` for listing, `DELETE /files?path=` for deletion. Follows the existing `/files/{id}` vs `/files/{id}/content` pattern
- **All paths relative to session home directory** — will consolidate with `$HOME` semantics
- **Non-recursive list** — returns immediate children only (cost/performance)
- **Delete returns 204 No Content** — no response body needed
- **`safeint` for sizes** — avoids JavaScript interoperability issues with int64
- **No symlink fields** — users can't create symlinks via the API
- **`is_directory`** — full name, no abbreviations
- **`recursive` defaults to `false`** — non-empty directory delete returns 409 Conflict
- **50 MB upload limit for M2** — file reference (URL/ID) planned for larger files in M2+

### Related
- Design doc: https://msdata.visualstudio.com/Vienna/_git/vienna/pullrequest/1970903
- API review doc: https://github.com/Azure/agent-first-sdk/pull/189
- ADC integration spec (merged): https://github.com/Azure/azure-rest-api-specs/pull/40739

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/main.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/main.tsp
@@ -1,5 +1,6 @@
 import "./src/agents/routes.tsp";
 import "./src/agents-containers/routes.tsp";
+import "./src/agents-session-files/routes.tsp";
 import "./src/agents-invocations/routes.tsp";
 import "./src/common/custom-types.tsp";
 import "./src/common/service.tsp";
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
@@ -1,0 +1,46 @@
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Response models
+// ---------------------------------------------------------------------------
+
+@removed(Versions.v1)
+@doc("Response from uploading a file to a session sandbox.")
+model SessionFileWriteResponse {
+  @doc("The path where the file was written, relative to the session home directory.")
+  path: string;
+
+  @doc("Number of bytes written.")
+  bytes_written: safeint;
+}
+
+@removed(Versions.v1)
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  @maxLength(255)
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: safeint;
+
+  @doc("Whether this entry is a directory.")
+  is_directory: boolean;
+
+  @doc("The last modification time in UTC (ISO 8601).")
+  @encode(DateTimeKnownEncoding.rfc3339)
+  modified_time: utcDateTime;
+}
+
+@removed(Versions.v1)
+@doc("Response from listing a directory in a session sandbox.")
+model SessionDirectoryListResponse {
+  @doc("The path that was listed, relative to the session home directory.")
+  path: string;
+
+  @doc("The directory entries.")
+  entries: SessionDirectoryEntry[];
+}
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
@@ -1,0 +1,147 @@
+import "../common/models.tsp";
+import "./models.tsp";
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+/**
+ * Session-scoped file operations for hosted agent sandboxes.
+ *
+ * These endpoints enable uploading, downloading, listing, and deleting
+ * files within a live ADC sandbox session. All file content is streamed
+ * end-to-end (no buffering in Agents service memory).
+ *
+ * Sessions are scoped to agents (not versions), giving optionality to
+ * mutate the associated agent version independently of the session lifecycle.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
+@tag("Agent Session Files")
+@removed(Versions.v1)
+interface AgentSessionFiles {
+
+  #suppress "@azure-tools/typespec-azure-core/byos" "The file is targeting the per-session files mounted to a foundry agent"
+  /**
+   * Upload a file to the session sandbox via binary stream.
+   * Maximum file size is 50 MB. Uploads exceeding this limit return 413 Payload Too Large.
+   */
+  @put
+  @route("/agents/{agent_name}/sessions/{session_id}/files/content")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  uploadSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The destination file path within the sandbox, relative to the session home directory. */
+      @query path: string;
+
+      /** The binary file content to upload. */
+      @header contentType: "application/octet-stream";
+      @body content: bytes;
+    },
+    SessionFileWriteResponse
+  >;
+
+  /**
+   * Download a file from the session sandbox as a binary stream.
+   */
+  @get
+  @route("/agents/{agent_name}/sessions/{session_id}/files/content")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  downloadSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file path to download from the sandbox, relative to the session home directory. */
+      @query path: string;
+    },
+    bytes
+  >;
+
+  /**
+   * List files and directories at a given path in the session sandbox.
+   * Returns only the immediate children of the specified directory (non-recursive).
+   */
+  @get
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  listSessionFiles is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The directory path to list, relative to the session home directory. */
+      @query path: string;
+    },
+    SessionDirectoryListResponse
+  >;
+
+  /**
+   * Delete a file or directory from the session sandbox.
+   * If `recursive` is false (default) and the target is a non-empty directory, the API returns 409 Conflict.
+   */
+  @delete
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  deleteSessionFile is FoundryDataPlanePreviewOperation<
+    AgentDefinitionOptInKeys.hosted_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file or directory path to delete, relative to the session home directory. */
+      @query path: string;
+
+      /** Whether to recursively delete directory contents. Defaults to false. */
+      @query recursive?: boolean = false;
+    },
+    void
+  >;
+}


specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,52 @@
...
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: int64;
+
+  @doc("Whether this entry is a directory.")
+  is_dir: boolean;
```
baywet: should this use is_directory to avoid unnecessary abrieviations?

baywet: should this have a default value?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,52 @@
...
+
+  @doc("Number of bytes written.")
+  bytes_written: int64;
+}
+
+@removed(Versions.v1)
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  name: string;
```
baywet, line 25: any constraints to the length of the name?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The destination file path within the sandbox. */
+      @query path: string;
+
+      /** The binary file content to upload. */
+      @header contentType: "application/octet-stream";
+      @body content: bytes;
```
baywet, line 54: does this support resumable uploads? how large do we expect files to become?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+    },
+    SessionFileWriteResponse
```
baywet, line 56: no error responses are defined? unless it's in the template? (foundry data plane operation)

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file path to download from the sandbox. */
+      @query path: string;
+    },
+    bytes
```
baywet, line 84: do we support resumable downloads?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+      @path agent_name: string;
+
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file or directory path to delete. */
+      @query path: string;
+
+      /** Whether to recursively delete directory contents. Defaults to false. */
+      @query recursive?: boolean;
```
baywet:
```suggestion
      @query recursive?: boolean = false;
```

(Maybe I'm getting caught up by TypeSpec syntax here, but shouldn't the default be set)

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
RaviPidaparthi: its a response property. So no, service will return a default if there is one

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  uploadSessionFile is FoundryDataPlanePreviewOperation<
```
RaviPidaparthi, line 40: Does uploading a file with nested path auto-create the directory structure ?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  downloadSessionFile is FoundryDataPlanePreviewOperation<
```
RaviPidaparthi, line 72: I assume you cant download directories? ex: return a zip file with directory contents.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+  @route("/agents/{agent_name}/sessions/{session_id}/files/list")
+  @extension(
+    "x-ms-foundry-meta",
+    #{
+      required_previews: #[
+        AgentDefinitionOptInKeys.hosted_agents_v1_preview
+      ],
+    }
+  )
+  listSessionFiles is FoundryDataPlanePreviewOperation<
```
RaviPidaparthi, line 101: Does this support listing recursively? Or just the current directory?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+      /** The session ID. */
+      @path session_id: string;
+
+      /** The file or directory path to delete. */
+      @query path: string;
+
+      /** Whether to recursively delete directory contents. Defaults to false. */
+      @query recursive?: boolean;
+    },
+    SessionFileOperationResponse
```
RaviPidaparthi: Wht do we need a body with message for delete? A simple success status code should suffice right?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ericsuh: I think customers may try to upload files that are multiple gigabytes (they wanted to do data analysis that was about that).

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,52 @@
...
+  @doc("The last modification time in UTC (ISO 8601).")
+  @encode(DateTimeKnownEncoding.rfc3339)
+  modified_time: utcDateTime;
+}
+
+@removed(Versions.v1)
+@doc("Response from listing a directory in a session sandbox.")
+model SessionDirectoryListResponse {
+  @doc("The path that was listed.")
+  path: string;
```
ericsuh, line 42: I feel like either the documentation  for each path parameter/field should say that these are all paths relative to `$HOME` or the `path` field should be renamed to `relpath`.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+@removed(Versions.v1)
+@doc("A single entry in a directory listing.")
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: int64;
```
johanste: Nitpick: I suspect that `safeint` is reasonably safe upper bound. `int64` is going to have interoperability issues (for ECMA script clients) so if we truly want that, we are going to have to consider a string version of the value as well.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+      @body content: bytes;
+    },
+    SessionFileWriteResponse
+  >;
+
+  /**
+   * Download a file from the session sandbox as a binary stream.
+   */
+  @get
+  @route("/agents/{agent_name}/sessions/{session_id}/files")
```
johanste: This route very much reads as a list api (a GET on a collection)

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
johanste: Are you referring to supporting range requests? If so, I think we should.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
johanste: Why do we need to include directory entries in the listing?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: We have blocked this to just allow 50 MBs for now, so they will have to retry. For larger file uploads, we plan to take in file reference (either url or id, pending decision).

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Yes

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: You are right. Just the files for now.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Just the current directory at this point. We can add support for that in the future if needed.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Done — renamed to `is_directory`.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Added `@maxLength(255)` — constrained by underlying filesystem limits.

specification/ai-foundry/data-plane/Foundry/main.tsp
```diff
@@ -1,5 +1,6 @@
 import "./src/agents/routes.tsp";
 import "./src/agents-containers/routes.tsp";
+import "./src/agents-session-files/routes.tsp";
```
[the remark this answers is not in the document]
ankitsul: Blocked by pre-existing `tsp compile` errors upstream (`OpenAI.AutoCodeInterpreterToolParam` missing in `openai-responses/models.tsp`). Will regenerate once resolved.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * These endpoints enable uploading, downloading, listing, and deleting
+ * files within a live ADC sandbox session. All file content is streamed
+ * end-to-end (no buffering in Agents service memory).
+ *
+ * Sessions are scoped to agents (not versions), giving optionality to
+ * mutate the associated agent version independently of the session lifecycle.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
+@tag("Agent Session Files")
```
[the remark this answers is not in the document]
ankitsul: Following the `agents-containers` pattern which uses `@removed(Versions.v1)` without `@added`. No `PuPr` version enum exists in this project.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Response models
+// ---------------------------------------------------------------------------
+
+@removed(Versions.v1)
+@doc("Response from uploading a file to a session sandbox.")
+model SessionFileWriteResponse {
```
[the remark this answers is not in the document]
ankitsul: Same as above — following `agents-containers` pattern.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Error responses are inherited from the `FoundryDataPlanePreviewOperation` template which returns `ApiErrorResponse` on failure.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Not in M2 scope (stream-only). Range request support is a good consideration for M2+ when we add large file transfer capabilities.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Done — set explicit default `= false`.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Good point — changed to return `void` (204 No Content). Removed `SessionFileOperationResponse`.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Updated all path field docs to say "relative to the session home directory".

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Done — changed `size` and `bytes_written` to `safeint` to avoid JavaScript interoperability issues.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Restructured: `PUT/GET /files/content?path=` for binary content, `GET /files?path=` for listing, `DELETE /files?path=` for deletion. Follows the existing `/files/{id}` vs `/files/{id}/content` pattern in this repo.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Directory entries are needed for filesystem navigation — users need to discover subdirectories to browse the sandbox structure.

baywet: If the property is not always returned, should it be nullable then?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
baywet: The description should at least reflect the size limit. 
Also I'm not seeing an error code for when the file is to big in the operation.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
baywet: Yes Johan that's what I was thinking about. But since I mentioned resumable uploads as well, I got mixed up with the terms

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
RaviPidaparthi: Property will always be returned

RaviPidaparthi:
Would be great if you can also do tsp compile and include the open api spec in your pr.

RaviPidaparthi (approved):

ericsuh (approved):

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
johanste: We are missing several other pieces of information - what is the path separator (assume '/'). Are path separators allowed in the file name? If so, how are they escaped?

Given that the API is asymmetrical (you can use normal file I/O operations from within the container, you are not constrained to file creates through the REST API) we presumably can't rely on the API itself validating inputs but have to represent all file names that can be created.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,46 @@
...
+model SessionDirectoryEntry {
+  @doc("The name of the file or directory.")
+  @maxLength(255)
+  name: string;
+
+  @doc("The size in bytes (0 for directories).")
+  size: safeint;
+
+  @doc("Whether this entry is a directory.")
+  is_directory: boolean;
```
johanste, line 31: Why do we need to return directory entries?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
johanste: What happens when the agent writes a large file and you try to `GET` it from the session files using the REST API?

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
```diff
@@ -0,0 +1,145 @@
...
+@removed(Versions.v1)
+interface AgentSessionFiles {
+
+  /**
```
johanste, line 26:
```suggestion
  #suppress "@azure-tools/typespec-azure-core/byos" "The file is targeting the per-session files mounted to a foundry agent"
  /**
```

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+@removed(Versions.v1)
+@doc("Response from listing a directory in a session sandbox.")
+model SessionDirectoryListResponse {
+  @doc("The path that was listed, relative to the session home directory.")
+  path: string;
+
+  @doc("The directory entries.")
+  entries: SessionDirectoryEntry[];
```
johanste, line 45: We are not leaving any space for server driven pagination here. But since this is a preview, we can add that later.

ankitsul: Acknowledged — we'll add server-driven pagination (continuation token) in a follow-up. Keeping it simple for preview.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Applied — added the `#suppress` annotation. Thanks for the suggestion.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Directory entries are needed so users can navigate the sandbox filesystem — without them, there's no way to discover what subdirectories exist to drill into. This is standard behavior for filesystem listing APIs (POSIX `readdir`, Python `os.listdir`, etc. all return both files and directories). The `is_directory` field lets clients distinguish them and build tree views or recursive traversals client-side.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: The file is streamed end-to-end from the ADC sandbox with no buffering in the Agents service, so large files will download successfully. Without range request support today, a failed download would need to be retried from the start. In practice, session files should be moderate in size given the 50MB upload limit, though agent-generated output could be larger.

That said, this is a two-way door — adding range requests and `Content-Range`/`If-Range` headers is an additive, non-breaking change to the existing `GET /files/content` endpoint. We'd prefer to ship the simpler streaming implementation now, validate real customer usage patterns, and layer in resumability once we understand what file sizes customers are actually working with. No API redesign needed when we get there.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
ankitsul: Good callout on the asymmetry. Since agent code has unrestricted filesystem access, the list API must faithfully represent any file the agent creates, even names the REST upload API wouldn't accept.

- **Path separator**: `/` (Linux). Cannot appear in filenames on ext4, so no ambiguity.
- **Query param encoding**: `path` values are percent-encoded per RFC 3986. Clients must encode special characters (spaces → `%20`, `#` → `%23`, etc.) when constructing the URL.
- **Response `name` field**: Returned as-is from the filesystem. JSON string encoding naturally handles unicode and special characters. Any file the agent creates can be represented in JSON.

So the full round-trip works: agent creates a file with any valid Linux filename → list returns it in JSON → client percent-encodes the name for download/delete. No information is lost at any step.

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp
ankitsul: Good point — added the 50 MB limit and `413 Payload Too Large` error to the upload operation description.

johanste (approved):
We should make sure to fuzz-test the path parameter and stress test file operations - e.g. recursive symlinks etc. 

While I don't expect agent developers trying to break themselves, developers can be creative.

PR #41606, opened 2026-03-20
Add data gen jobs Apis and contracts

sakoll:
This pull request introduces a new "Data Generation Jobs" feature to the AI Foundry data plane. It adds comprehensive model definitions, API routes, and configuration for managing data generation jobs, including support for multiple job types, sources, and outputs. The changes also update relevant configuration and documentation files to support the new feature.

**Data Generation Jobs feature additions:**

* Added new models for data generation jobs, including job types, status, sources, outputs, and options in `data_generation/models.tsp`. These models support various job types (e.g., SimpleQnA, ToolUse, MultiTurn), sources (e.g., Traces, Prompt, File, Dataset), and output types (File, Dataset), as well as job tracking fields and token usage.
* Introduced a new API interface `DataGenerationJobs` in `data_generation/routes.tsp` with endpoints to create, list, get, cancel, and delete data generation jobs, including preview requirements and operation metadata.

**Configuration and integration updates:**

* Registered the new data generation routes in the main Foundry API entrypoint (`main.tsp`).
* Updated the TypeSpec SDK agents configuration to include the new `data_generation` directory for code generation.
* Added a new opt-in preview key `data_generation_jobs_v1_preview` to the `FoundryFeaturesOptInKeys` union, enabling feature-gated access to the data generation jobs API.

# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
+import "../common/models.tsp";
+
+namespace Azure.AI.Projects;
+
+@doc("Error information for a data generation job that failed.")
+model DataGenerationJobError {
```
aprilk-ms: Is there any standard error object we can reuse, instead of creating new type?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  string,
+
+  @doc("SimpleQnA job type")
+  SimpleQnA: "SimpleQnA",
+
+  @doc("Tool use job type")
+  ToolUse: "ToolUse",
+
+  @doc("Multi turn Question and Answer generation job type")
+  MultiTurnQnA: "MultiTurnQnA",
```
aprilk-ms: I have mixed feelings about calling this QnA... what if it is a more complex agent (like a travel booking agent, that goes beyond conversation)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+@doc("Model generation sampling / configuration options.")
```
aprilk-ms: The doc string is a bit odd. It is data generation option not model generation?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+model DataGenerationModelOptions {
+  @doc("Base model deployment name used to generate synthetic data.")
```
aprilk-ms: Can we just say model name for data generation (need not have a deployment or might not be synthetic)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  `model`: string;
+}
+
+@doc("Status values for a data generation job.")
+union DataGenerationJobStatus {
```
aprilk-ms: do we need cancelled?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  id: string;
+
+  @doc("The current status of the job.")
+  @visibility(Lifecycle.Read)
+  status: DataGenerationJobStatus;
+
+  @doc("The name of the output dataset to create. If not specified, a default name will be generated. If dataset with the same name already exists then tries to create a new version of the dataset with the generated data.")
+  output_dataset_name?: string;
+
+  @doc("Identifier of the output dataset produced after successful completion. This will be in format of in the format of 'azureai://accounts/<account-name>/projects/<project-name>/data/<dataset-name>/versions/<version-no>'")
```
aprilk-ms:
```suggestion
  @doc("Identifier of the output dataset produced after successful completion. This will be in the format of 'azureai://accounts/<account-name>/projects/<project-name>/data/<dataset-name>/versions/<version-no>'")
```

aprilk-ms: I still don't get why this azureai signature. We only need dataset name/version to fetch a dataset right?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  @visibility(Lifecycle.Read)
+  output_dataset_id?: string;
+
+  @doc("The kind of data generation job executed.")
+  type: DataGenerationJobType;
+
+  @doc("The purpose of the data generation job. Either for fine-tuning supervised training or evaluation.")
+  purpose: DataGenerationJobPurpose;
+
+  @doc("The number of synthetic samples actually generated. This field will be populated only after the job is completed.")
```
aprilk-ms: Remove synthetic (it could be from traces)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  Traces: "Traces",
+
+  @doc("Prompt Datasource type. Data is generated based on input prompts provided by the user or will be extracted from the given Agent. Applicable only for evaluation purpose.")
+  Prompt: "Prompt",
+
+  @doc("Seed data files datasource type. Data is generated based on input seed data files from the given input datasets. Applicable only for fine-tuning training purpose.")
+  SeedDataFiles: "SeedDataFiles",
+
+  @doc("Datasource type that combines prompts and seed data files. Data is generated based on both input prompts and seed data files. Applicable only for evaluation purpose.")
+  PromptAndSeedDataFiles: "PromptAndSeedDataFiles",
```
aprilk-ms: This looks a bit ugly. Can we support a job with multiple sources? (but maybe we can validate unsupported combination to begin with)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+model SimpleQnADataGenerationJob extends DataGenerationJob {
+  @doc("The data generation job type, which is SimpleQnA for this model.")
+  type: "SimpleQnA";
+
+  @doc("The datasource type for this job.")
+  datasource_type: DataGenerationJobDataSourceType;
+}
+
+@doc("A data generation job of type SimpleQnA with Prompt datasource used for evaluation.")
+model SimpleQnAPromptEvaluationDataGenerationJob
```
aprilk-ms: Hm. I don't really like this type combination (type x source). Can we have specific job type and data source options within that? (like in eval job)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+
+  @doc("The options for a SimpleQnA data generation job with Traces datasource.")
+  options: TracesDataGenerationJobOptions;
+}
+
+@doc("The options for a data generation job with Traces datasource, which includes the agent id and the time window for fetching traces.")
+model TracesDataGenerationJobOptions {
+  ...DataGenerationJobOptions;
+
+  @doc("The agent id to fetch the traces for data generation. The agent id should be in format of '<agent_id>:<version>'")
```
aprilk-ms: I am not sure we still have <agent_id>:<version> format. Let's not take a dependency on that

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,339 @@
...
+  agent_id: string;
```
aprilk-ms: How will multi-agent work? Do we need to support fetching from multiple agent Ids?

specification/ai-foundry/data-plane/Foundry/src/data_generation/routes.tsp
```diff
@@ -0,0 +1,110 @@
...
+
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+@tag("DataGenerationJobs")
+interface DataGenerationJobs {
+  /**
+   * Gets the details of a data generation job by its ID.
+   */
+  @summary("Get info about a data generation job.")
+  @route("data_generation/jobs/{data_generation_job_id}")
```
aprilk-ms: I don't really like the route with first segment for grouping purpose. But I guess it is aligning with fine-tuning jobs API which have similar structure. Naturally, I thought it should be under /datasets:generate or something.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: So dataset id is currently set in this format

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: So from my understanding, all we need is parent agent which helps to get all the connected agents from the agent definition? Is that not the case for Agents v2?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: Yes, we have ApiError. Just reusing it for now.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: Renamed it to MultiTurn now

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: updated

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: updated

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: Yes, we needed

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: I debated for a long time with @shivamMg about this :). I feel like adding this specific option with combination will end users to determine what combination is supported and they don't need to take any guesses on what combinations are supported from API/SDK experience.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: I changed it to simplify the model. However similar to other comment I think adding this explicit combination model would help with avoiding end user errors and they don't need to explicitly go through the docs to find what all are supported. One issue with this kind of models they are getting complicated and should add need whenever we have new type.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: done

specification/ai-foundry/data-plane/Foundry/src/data_generation/routes.tsp
sakoll: Probably @shivamMg can chime in about this and comment about reason for this structure.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: After thinking more about it, I removed this complex type and instead now we are supporting multiple sources/datasources

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: Added name and version for dataset.

specification/ai-foundry/data-plane/Foundry/src/data_generation/routes.tsp
sakoll: discussed offline about this and closing this

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,291 @@
...
+model FileDataGenerationJobOutput extends DataGenerationJobOutput {
+  @doc("The type of the output, which is File for this model.")
+  type: "File";
+
+  @doc("The id of the output Azure OpenAI file.")
+  @visibility(Lifecycle.Read)
+  file_id?: string;
+}
+
+model DatasetDataGenerationJobOutput extends DataGenerationJobOutput {
```
aprilk-ms, line 90: can we include optional description and tags? if user passes those in, we'll set them in the dataset created

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,291 @@
...
+
+model DatasetDataGenerationJobOutput extends DataGenerationJobOutput {
+  @doc("The type of the output, which is Dataset for this model.")
+  type: "Dataset";
+
+  @doc("The id of the output dataset created.")
+  @visibility(Lifecycle.Read)
+  id?: string;
+
+  @doc("The name of the output dataset.")
```
aprilk-ms: Is it going to be a new dataset (with validation)? Or can be an existing one? If it can be existing, should we also allow version?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,291 @@
...
+  type: "Traces";
+
+  @doc("The agent name to fetch the traces for data generation. The traces will be fetched based on the agent name from the project connected Application Insights within the specified time window. If both agent name and agent version are specified, the agent version will be used to fetch the traces.")
+  agent_name: string;
+
+  @doc("The agent version to fetch the traces for data generation. If not specified, the latest version of the agent will be used.")
+  agent_version?: string;
+
+  @doc("The start of the time window for fetching traces. This defines the beginning of the time range for which traces will be considered for data generation.")
+  from_time: utcDateTime;
```
aprilk-ms: can we check the eval run contract, and see if we have the same field names for time range?

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,291 @@
...
+@discriminator("type")
+model DataGenerationJobSource {
+  @doc("The type of source.")
+  type: DataGenerationJobSourceType;
+}
+
+@doc("The options for a data generation job with Traces source used for Evaluation and fine-tuning purposes, which includes the agent name and the time window for fetching traces.")
+model TracesDataGenerationJobSource extends DataGenerationJobSource {
+  @doc("The source type for this job, which is Traces.")
+  type: "Traces";
```
aprilk-ms: I am wondering if we should allow max traces as input.. maybe for later, we just have an internal limit for now...

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,291 @@
...
+  @doc("The source type for this job, which is File.")
+  type: "File";
+
+  @doc("Input Azure Open AI file id references used for data generation.")
+  file_id: string;
+}
+
+model DatasetDataGenerationJobSource extends DataGenerationJobSource {
+  @doc("The source type for this job, which is Dataset.")
+  type: "Dataset";
```
aprilk-ms: Is there a need for user to describe what the dataset or file is about (as in metadata)? Will that help the data gen process?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: This can be an existing dataset as well. For existing dataset, we will just publish new version. Taking version as input might end up with conflicts if same version dataset is created while job is ruinning

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: I think they would describe that info in prompt or the file content would contain the context about what it is about.

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: done

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: Added to allow max traces

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: Update to use eval naming conventions. For start time, eval contract currently providing an option to set it as only hours but I think it is good to have some flexibility around it and not restrict to only hours.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,312 @@
...
+
+  @doc("The generated data is a Dataset.")
+  Dataset: "Dataset",
+}
+
+@doc("Token usage information for a data generation job.")
+model TokenUsage {
+  @doc("The number of prompt tokens used.")
+  @visibility(Lifecycle.Read)
+  prompt_tokens?: int32;
```
shivamMg: can we use int64 for all 3 token fields.

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,312 @@
...
+}
+
+@doc("Options for managing data generation jobs.")
+@discriminator("type")
+model DataGenerationJobOptions {
+  @doc("The data generation job type.")
+  type: DataGenerationJobType;
+
+  @doc("Maximum number of samples to generate.")
+  max_samples_count: int32;
```
shivamMg: should be `max_samples`

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,312 @@
...
+
+  @doc("The agent name to fetch the traces for data generation. The traces will be fetched based on the agent name from the project connected Application Insights within the specified time window.")
+  agent_name: string;
+
+  @doc("The agent version to fetch the traces for data generation. If not specified, the latest version of the agent will be used.")
+  agent_version?: string;
+
+  @doc("The start of the time window (in Unix timestamp in seconds) for fetching traces. This defines the beginning of the time range for which traces will be considered for data generation.")
+  @encode(DateTimeKnownEncoding.unixTimestamp, int32)
+  start_time: utcDateTime;
```
shivamMg: should start_time and end_time be int64 (as we're using int64 for job.created_at)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -0,0 +1,312 @@
...
+
+  @doc("The end of the time window (in Unix timestamp in seconds) for fetching traces. This defines the end of the time range for which traces will be considered for data generation. If not specified, it will default to the current time, meaning that traces up until now will be considered for data generation.")
+  @encode(DateTimeKnownEncoding.unixTimestamp, int32)
+  end_time?: utcDateTime;
+
+  @doc("The delay to apply for ingestion when querying traces. TODO: Do we need this?")
+  @encode(DurationKnownEncoding.seconds, int32)
+  ingestion_delay_seconds?: duration = duration.fromISO("PT5M");
```
shivamMg: do we need ingestion delay? imo the time range denoted by start_time and end_time should suffice

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
```diff
@@ -0,0 +1,110 @@
...
+    {
+      ...CommonPageQueryParameters;
+
+      @doc("Filter data generation jobs by their scenario.")
+      @query
+      scenario?: DataGenerationJobScenario;
+
+      @doc("Filter data generation jobs by their type.")
+      @query
+      type?: DataGenerationJobType;
```
shivamMg: what's the expected API contract for these filters. do you know

> /jobs?type=SimpleQnA&type=MultiTurn

or

> /jobs?type[]=SimpleQnA&type[]=MultiTurn

or something else

specification/ai-foundry/data-plane/Foundry/src/data_generation/routes.tsp
```diff
@@ -0,0 +1,110 @@
...
+      location: string;
+    }>
+  >;
+
+  /**
+   * Cancels a data generation job by its ID.
+   */
+  @summary("Cancels a data generation job.")
+  @extension("x-ms-foundry-meta", DataGenerationJobsRequiredPreviews)
+  @route("data_generation/jobs/{data_generation_job_id}/cancel")
```
shivamMg: looking at Gerardo's Job template, this need to be `:cancel`. let's use that I guess

specification/ai-foundry/data-plane/Foundry/main.tsp
```diff
@@ -19,3 +19,4 @@ import "./src/openai-finetuning/routes.tsp";
 import "./src/openai-responses/routes.tsp";
 import "./src/red-teams/routes.tsp";
 import "./src/schedules/routes.tsp";
+import "./src/data_generation/routes.tsp";
```
shivamMg: nit: shall we make the parent folder `data_generations` (plural) since that's the convention other folders are following

shivamMg (changes requested):

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,312 @@
...
+
+@doc("The supported data generation job types.")
+union DataGenerationJobType {
+  string,
+
+  @doc("Simple question and answers between user and agent.")
+  SimpleQnA: "SimpleQnA",
+
+  @doc("Single turn conversation between user and agent. The generated data contain a query with list of messages in the conversation as context and a response (last turn) from the agent.")
+  SingleTurn: "SingleTurn",
```
shivamMg: (just a blocking comment for us to finalize name of this recipe from William/April)

shivamMg (changes requested):

specification/ai-foundry/data-plane/Foundry/src/data_generation/routes.tsp
sakoll: I will cherry pick his commit to apply this change

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: I think int32 is good enough. It is equal to ~136 years from 1970. I will change the other one also to int32

specification/ai-foundry/data-plane/Foundry/main.tsp
sakoll: I'm changing it to data_generation_jobs

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
sakoll: I'm expecting something like this `/jobs?type=SimpleQnA&type=MultiTurn`. We can also check with Johan on the recommendations here

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: I don't think we need for our scenario now. However if we ended up having continuous eval scenario then we need it

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+  agent_version?: string;
+
+  @doc("The start of the time window (in Unix timestamp in seconds) for fetching traces. This defines the beginning of the time range for which traces will be considered for data generation.")
+  start_time: FoundryTimestamp;
+
+  @doc("The end of the time window (in Unix timestamp in seconds) for fetching traces. This defines the end of the time range for which traces will be considered for data generation. If not specified, it will default to the current time, meaning that traces up until now will be considered for data generation.")
+  end_time?: FoundryTimestamp;
+
+  @doc("Sampling limit applied to traces retrieved for data generation.")
+  max_traces?: int32 = 1000;
```
shivamMg, line 236: do we want users to be able to set max_traces? it feels like a limit that we should decide on the backend. users might only be concerned with max_samples

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+@discriminator("type")
+model DataGenerationJobOptions {
+  @doc("The data generation job type.")
+  type: DataGenerationJobType;
+
+  @doc("Maximum number of samples to generate.")
+  max_samples: int32;
+
+  @doc("The LLM model options.")
+  model_options?: DataGenerationModelOptions;
```
shivamMg, line 180: need to add a `train_split: float` field that's used in finetuning to dictate the training/validation split

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: I think we should provide this option explicitly with setting some max limit. The default would be 1000 if they don't specify anything or else if user is specifying any value this should not cross more than value whatever we decide based on our testing

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: done

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
```diff
@@ -0,0 +1,80 @@
...
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+#suppress "@azure-tools/typespec-azure-core/long-running-polling-operation-required" "LRO polling is handled by the client via Operation-Location"
+@tag("DataGenerationJobs")
+interface DataGenerationJobs {
+  /**
+   * Gets the details of a data generation job by its ID.
+   */
+  @summary("Get info about a data generation job.")
+  @route("data_generation/jobs/{jobId}")
```
glecaros: from feedback from Dan, we should use `/data_generations/{job_id}`.

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
```diff
@@ -0,0 +1,80 @@
...
+  get is queryJobStatusPreview<
+    FoundryFeaturesOptInKeys.data_generation_jobs_v1_preview,
+    DataGenerationJob
+  >;
+
+  /**
+   * Returns a list of data generation jobs.
+   */
+  @summary("Returns a list of data generation jobs")
+  @route("data_generation/jobs")
```
glecaros: same comment on the route as above

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
shivamMg: @glecaros @dantaylo_microsoft We would prefer /data_generation/jobs so it's clear to users that they are dealing jobs and they can apply whatever common job helpers we might introduce in the future e.g. `wait(job)`.

The hierarchy also helps us add data generation features that are outside of jobs. For instance, finetuning has a similar structure /fine_tuning/jobs, and later we introduced graders under /fine_tuning/alpha/graders

shivamMg: Synced with April offline - we're going to go with `/data_generation_jobs`

sakoll: changed the route paths now

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: As discussed offline, this is updated now to Traces

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+  agent_version?: string;
+
+  @doc("Start of the time window (Unix timestamp in seconds) for fetching traces.")
+  start_time?: FoundryTimestamp;
+
+  @doc("End of the time window (Unix timestamp in seconds). Defaults to current time.")
+  end_time?: FoundryTimestamp;
+
+  @doc("Maximum number of traces to retrieve.")
+  max_traces?: int32 = 100;
```
shivamMg: by default max traces should be null? (decided by the service)

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
   @body responseBody: T;
 };
+
+// ============================================================================
+// Shared Job Sources
+//
+// Source types shared across job types (evaluator generation, data generation,
+// and future job types such as optimization). Placed in common to avoid
+// duplication across job-specific model files.
+// ============================================================================
```
shivamMg, line 112: having common source types across jobs will make it difficult to grow them independently. e.g. support for a new field in datagen would not be respected in other source types

shivamMg (changes requested):

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
sakoll: So you can always add them as optional property. Other option is you can inherit common type and add extension properties if it is specific to datagen. So in general for a GA service, adding any property to existing Api or model would be a optional and removing anything is a breaking change should follow breaking change process.

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
glecaros: Azure guidelines allow for both `?foo=value1,value2` and `?foo=value1&foo=value2`, but they do say that comma separated list is prefered ([source](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-query-parameters-and-header-values))

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+
+@doc("The supported source types for jobs.")
+union JobSourceType {
+  string,
+
+  @doc("Prompt source — inline text provided by the user.")
+  Prompt: "Prompt",
```
glecaros: we may need to move enum values to be snake_cased across the surface for consistency. I'll try to confirm the requirement for this today

sakoll: converted this to snake_case

glecaros (approved):

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
glecaros: if you want to take the comma approach, you need to write like [this](https://typespec.io/playground/?e=%40typespec%2Fopenapi3&c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQoNCnVzaW5nIEh0dHDFD29wIGZvbygNCiAgQHF1ZXJ5KCN7IGV4cGxvZGU6IGZhbHNlIH0pxB9hcnI6IHN0cmluZ1tdDQopOiBpbnQxNjs%3D&options=%7B%7D&vs=%7B%7D). If you want to go with the other approach you need to use `explode: true` ([docs](https://typespec.io/docs/libraries/http/reference/data-types/#TypeSpec.Http.QueryOptions))

sakoll: Going with comma

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+@doc("The options for a data generation job with Task type.")
+model TaskDataGenerationJobOptions extends DataGenerationJobOptions {
+  @doc("The data generation job type, which is Task for this model.")
+  type: "task";
+}
+
+@doc("The supported job source types for data generation.")
+union DataGenerationJobSource {
```
glecaros: Here we will probably need to either duplicate the models from the JobSource union into DataGenerationJobSource, or add the FileDataGenerationJobSource (probably the former is what we need here)

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,99 @@ alias SseResponseOf<T> = {
...
+  type: "agent";
+
+  @doc("The agent name to fetch instructions from.")
+  agent_name: string;
+
+  @doc("The agent version. If not specified, the latest version is used.")
+  agent_version?: string;
+
+  @doc("Optional supplementary prompt text to augment the agent's instructions (e.g., additional context not captured in the agent definition).")
+  prompt?: string;
```
shivamMg: remove this prompt in favor of PromptJobSource

shivamMg (changes requested):

shivamMg (approved):

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,96 @@ alias SseResponseOf<T> = {
...
+
+  @doc("Dataset source — reference to a dataset.")
+  dataset: "dataset",
+}
+
+@doc("Common properties shared across all job source types.")
+model JobSourceDescription {
+  @doc("Optional description of what this source represents — helps the pipeline interpret its content (e.g., 'Company refund policy document' or 'Describes the agent's core capabilities').")
+  description?: string;
+}
```
johanste:
```suggestion
alias JobSourceDescription = {
  @doc("Optional description of what this source represents — helps the pipeline interpret its content (e.g., 'Company refund policy document' or 'Describes the agent's core capabilities').")
  description?: string;
};
```

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,96 @@ alias SseResponseOf<T> = {
...
+// ============================================================================
+// Shared Job Sources
+//
+// Source types shared across job types (evaluator generation, data generation,
+// and future job types such as optimization). Placed in common to avoid
+// duplication across job-specific model files.
+// ============================================================================
+
+@doc("The supported source types for jobs.")
+union JobSourceType {
```
johanste, line 115: Is this really a **common** type (i.e. applies/will apply to more thing than data generation jobs)?

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,265 @@
...
+  `model`: string;
+}
+
+@doc("Caller-supplied inputs for a data generation job.")
+model DataGenerationJobInputs {
+  @doc("The display name of the data generation job.")
+  name: string;
+
+  @doc("The sources used for the data generation job.")
+  sources: Array<DataGenerationJobSource>;
```
johanste:
```suggestion
  sources: DataGenerationJobSource[];
```

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,265 @@
...
+  options: DataGenerationJobOptions;
+
+  @doc("The scenario of the data generation job. Either for fine-tuning or evaluation.")
+  scenario: DataGenerationJobScenario;
+}
+
+@doc("Result produced by a successful data generation job.")
+model DataGenerationJobResult {
+  @doc("The final job outputs: Azure OpenAI files for fine-tuning, or datasets for evaluation.")
+  outputs?: Array<DataGenerationJobOutput>;
```
johanste:
```suggestion
  outputs?: DataGenerationJobOutput[];
```

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,265 @@
...
+
+  @doc("The number of samples actually generated.")
+  generated_samples?: int32;
```
johanste: When would this not be present?

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -0,0 +1,265 @@
...
+  model_options?: DataGenerationModelOptions;
+}
+
+@doc("The options for a data generation job with SimpleQnA type.")
+model SimpleQnADataGenerationJobOptions extends DataGenerationJobOptions {
+  @doc("The data generation job type, which is SimpleQnA for this model.")
+  type: "simple_qna";
+
+  @doc("The question types to generate. Used only for fine-tuning scenarios.")
+  question_types?: Array<SimpleQnAFineTuningQuestionType>;
```
johanste:
```suggestion
  question_types?: SimpleQnAFineTuningQuestionType[];
```

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,96 @@ alias SseResponseOf<T> = {
...
+}
+
+@doc("Common properties shared across all job source types.")
+model JobSourceDescription {
+  @doc("Optional description of what this source represents — helps the pipeline interpret its content (e.g., 'Company refund policy document' or 'Describes the agent's core capabilities').")
+  description?: string;
+}
+
+@doc("Prompt source — inline text provided by the user.")
+model PromptJobSource {
```
johanste, line 137: Why is this model needed? It is only spread once, so we can just inline it at the point of usage. The same goes for the other job sources....

glecaros: This is because this model is shared by different unions. This and the fact that we have to use `@discriminator` based unions forces us to duplicate them.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
glecaros: this union is actually not used, given the same reason that we needed to duplicate the types that are being spread, we had to create a `DataGenerationJobSourceType ` for this scenario.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
glecaros: #42264 also uses them

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
sakoll: Yes, this is going to be used for Evaluator Job Apis as well which is part of this PR
https://github.com/Azure/azure-rest-api-specs/pull/42264

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: So when Job failed or cancelled with generating any samples this would happen. Technically we can also assign 0 value in this case remove nullable type here.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
sakoll: This is going to be used for other Job Apis which is part of this PR
https://github.com/Azure/azure-rest-api-specs/pull/42264

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
sakoll: removed nullable type

slister1001:
Hi @sakoll — heads-up for cross-PR coordination.

We're applying @johanste's review feedback patterns from this PR to our adaptive-evals reapply ([#42764](https://github.com/Azure/azure-rest-api-specs/pull/42764)) via child PR [#42768](https://github.com/Azure/azure-rest-api-specs/pull/42768):

- `JobSourceDescription` in `common/models.tsp`: `model` → `alias`
- Removed a redundant double-spread of `JobSourceDescription` on our discriminated base

The `alias` change means once #42764 lands, your branch will need a small rebase so that any references to `JobSourceDescription` use the `alias` form (spread sites stay syntactically identical: `...JobSourceDescription`).

Also: we're keeping `PromptJobSource` / `AgentJobSource` / `TracesJobSource` / `DatasetJobSource` as `model`s in `common/` because they're consumed by **both** PRs (per-context discriminated subtypes in your `data_generation/` and our `evaluators/`). If you'd like to adopt the same shared shapes — that was always the intent for placing them in `common/` — happy to coordinate.

Happy to coordinate ordering: would you prefer to land first, or want us to land first?

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
```diff
@@ -6,8 +6,9 @@ emit:
 options:
   "@typespec/openapi3":
     emitter-output-dir: "{project-root}/"
-    output-file: "openapi3/{version}/microsoft-foundry-openapi3.json"
+    output-file: "openapi3/{version}/microsoft-foundry-openapi3.{file-type}"
```
dargilco, line 9: This does not work for me. When I check out this topic branch and run "npx tsp compile ." I see these files added: 
```
  openapi3/v1/microsoft-foundry-openapi3.yaml,json
  openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml,json
```

johanste (approved):

PR #41608, opened 2026-03-20
Read SDK-reported isSpecGenSdkCheckRequired in spec-gen-sdk-runner

live1206:
## Problem

In TypeSpec migration PRs, `SDK Validation - .NET` incorrectly becomes required and blocks the PR, even when the .NET SDK has no TypeSpec generation path (no C# emitter configured in `tspconfig.yaml`).

The root cause is that `getRequiredSettingValue()` uses the broad `hasTypeSpecProjects` flag (derived from `generateFromTypeSpec` in execution-report.json) to decide if the .NET check is required. This can be incorrectly `true` when the PR has TypeSpec files but the .NET SDK doesn't use the new mgmt emitter.

## Fix

Read `isSpecGenSdkCheckRequired` from the SDK generation output (`generateOutput.json`) instead of computing it from the broad `hasTypeSpecProjects` flag. For .NET SDK, the generation script explicitly declares whether the SDK validation check should be required.

### Changes
- **`command-helpers.ts`**: Add `readSdkOutputCheckRequired()` to read SDK output directly. Add `sdkReportedCheckRequired` parameter to `getRequiredSettingValue()` and `generateArtifact()` for .NET-specific override.
- **`commands.ts`**: In `generateSdkForSpecPr()`, read and accumulate `sdkReportedCheckRequired` across runs, pass to `generateArtifact()`.
- **Tests**: 9 new test cases for the new parameter and `readSdkOutputCheckRequired()`. All 45 tests pass.

**Companion PR (SDK repo):** https://github.com/Azure/azure-sdk-for-net/pull/57279

Fixes Azure/azure-sdk-for-net#57275

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
@@ -285,6 +285,40 @@ export function getBreakingChangeInfo(executionReport: ExecutionReport): boolean
 }
 
 /**
+ * Read isSpecGenSdkCheckRequired from the SDK generation output (generateOutput.json).
+ * For .NET SDK, the generation script explicitly declares whether the SDK validation
+ * check should be required, based on whether the package uses the new mgmt emitter.
+ * @param commandInput - The command input.
+ * @returns true/false if SDK explicitly declares the value, undefined if the property
+ *          is not present (e.g., older SDK script) or the file doesn't exist.
+ */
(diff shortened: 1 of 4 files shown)


ArcturusZhang (approved):

mikeharder (approved):

live1206:
@raych1 I am merging this to unblock the .NET SDK validation requirements. Please take a look at the fix and get back to me if you have any convern.

PR #41609, opened 2026-03-20
add full checkout cloning for SDK repository

### LINKED ISSUE #14589: [JS] Git commands used in changelog generation are fragile and may cause silent failures
**Background**
During changelog generation, the tool reads two types of files from a previously released package version:

- `review/*.api.md` — used for API diff to detect breaking changes and new features
- `CHANGELOG.md` — used as the base for the new changelog entry
The primary path is to extract these files from the npm tarball via [npm pack ${packageName}@${version}](vscode-file://vscode-app/c:/Users/v-ziweichen.FAREAST/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html). However, when the files are not found in the tarball (e.g., review/ directory is not in the package's [files](vscode-file://vscode-app/c:/Users/v-ziweichen.FAREAST/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field), the code falls back to reading them directly from the local git repository using git commands:
```
// In npmUtils.ts — tryCreateLastestStableNpmViewFromGithub
git --no-pager show ${tag}:${sdkFilePath}           // reads CHANGELOG.md from release tag
git --no-pager show ${tag}:${nodeApiFilePath}        // reads -node.api.md from release tag
git --no-pager show ${tag}:${standardApiFilePath}    // reads .api.md from release tag

// Additional git commands used to validate before calling the above
git tag -l ${tag}                                   // check if release tag exists locally
git ls-tree -d ${tag} ${relativePath}               // check if directory exists in tag
```
**Problem**
These git commands are not reliable in standard CI pipeline environments:

- CI pipelines typically use shallow clones (--depth=1), meaning release tags from previous versions are not fetched locally
- `git tag -l ${tag}` returns empty even when the tag exists remotely → incorrect false → fallback skipped
- `git ls-tree -d ${tag} ...` fails silently → wrong assumption about package structure
- `git --no-pager show ${tag}:...` produces empty content → API diff falls back to empty/default → breaking changes or new features suppressed in the generated changelog

**Impact**
- Wrong log generates or Wrong version bumps 
Case: https://github.com/Azure/azure-sdk-for-js/pull/37575
in this pr, the first generation(commit) contains the correct changelog and version
but in the next commit, we can see so many info with `Command failed` in the generation pipeline: https://dev.azure.com/azure-sdk/internal/internal%20Team/_build/results?buildId=6019383&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=2482
which caused the wrong changelog generation

- Failed update UserAgentInfo
Case:  https://github.com/Azure/azure-sdk-for-js/pull/37171
in this pr, the first generation(commit) generates UserAgentInfo correctly: 
Image

but in the next [generation](https://github.com/Azure/azure-sdk-for-js/pull/37171/changes/2e1992ed95089b54b7dcc52645809a59282cfff7)
 it can't update the version (PS: 1.0.0-beta.1 is a default value in codegen) after re-generation.
and in the [next generation pipeline](https://dev.azure.com/azure-sdk/internal/internal%20Team/_build/results?buildId=5866668&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=1345) we can see it read nothing from `git tag` command, but it could read it in the [first generation pipeline](https://dev.azure.com/azure-sdk/internal/internal%20Team/_build/results?buildId=5803515&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692&l=1299)

@jeremymeng  could we still export `./review` and `CHANGELOG.md` in packed packages for the changelog generation?

skywing918:
fixes https://github.com/Azure/azure-sdk-tools/issues/14589

since we need tags and historic version of some files, to generate changelog.


eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -187,14 +187,34 @@ stages:
...
+              if ($attempt -gt $retries) { exit $LASTEXITCODE }
+              Write-Warning "Attempt $attempt failed. Retrying in $sleep seconds..."
+              Start-Sleep -Seconds $sleep; $sleep *= 2
+            }
+          }
+          New-Item -ItemType Directory -Force -Path "$(SdkRepoDirectory)" | Out-Null
+          Push-Location "$(SdkRepoDirectory)"
+          Write-Host "Cloning $(SdkRepoOwner)/$(SdkRepoName) with full checkout..."
+          $repoUrl = "https://github.com/$(SdkRepoOwner)/$(SdkRepoName)"
+          Retry 3 git clone --single-branch --branch "$(CheckoutCommitish)" $repoUrl .
```
weshaggard, line 217: The issue talks about the JS repo but this will add a full lone for all repos which some are much larger and take longer.

weshaggard (changes requested):
I left some suggestions in the issue that I'd like to try before we move to full clones for all the sdk repos.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
weshaggard: FYI @raych1

skywing918: updated, Added a full checkout step only for `azure-sdk-for-js`
```
condition: and(succeeded(), eq(variables['SdkRepoName'], 'azure-sdk-for-js'))
```

MaryGao (changes requested):
Please help me understand the listed questions before any fix: https://github.com/Azure/azure-sdk-tools/issues/14589#issuecomment-4115312600. Thanks!

skywing918:
root cause: the[ azure-sdk](https://github.com/Azure-sdk/azure-sdk-for-js) does not sync all tags from [azure](https://github.com/Azure/azure-sdk-for-js/)

PR #41619, opened 2026-03-20
[DRAFT] project horizon: adding typespec for the command training job

jayesh-tanna:
# Choose a PR Template

**Description:**

This PR introduces the Training Jobs API surface to the Azure AI Foundry data-plane TypeSpec specification under the Azure.AI.Projects namespace. The feature is gated behind an opt-in preview header (Foundry-Features: Jobs=V1Preview) and is accessible in the generated Python SDK via client.beta.training.jobs.

**Operation Group**
An operation group in TypeSpec is declared as an interface and maps directly to a sub-client property in the generated SDK. The Jobs interface is tagged with @tag("Training") and routed under /training, which is why the SDK exposes it as client.beta.training.jobs. This PR adds five operations to that group: list, get, create_or_update, begin_delete, and begin_cancel. The list operation supports filtering by job type, tag, view type (active/archived/all), and user-defined properties. Both delete and cancel are asynchronous — they return HTTP 202 with a Location header for polling; delete also returns 204 when the job does not exist, and cancel can return 200 when it completes synchronously.

**Job Model**
The core resource is CommandJob, which extends a JobProperties base discriminated on jobType. Required fields are command (the shell command to run inside the container, which can reference inputs and outputs via ${{inputs.x}} and ${{outputs.y}}), environmentImageReference (a plain ACR/MCR Docker image URL), and compute (a compute resource ID). Optional fields cover display metadata (displayName, description, tags, properties), a registered code asset reference, environment variables, input/output bindings, distribution configuration, resource configuration, job limits, queue settings, and an archive flag. The services and status fields are read-only — they are populated by the service and never sent by the client.

**Input and Output Models**
Rather than using a discriminator hierarchy (separate subclasses per asset type), both Input and Output are defined as flat models. This avoids subclass proliferation in the Python SDK and keeps usage simple. The type field (encoded on the wire as jobInputType / jobOutputType) accepts values from the AssetTypes union: uri_file, uri_folder, safetensors_model, and literal. The path field (encoded as uri) holds the asset URI for file and folder types. For literal inputs, value carries the scalar string. The mode field on both models uses the unified InputOutputModes union (ReadOnlyMount, ReadWriteMount, Download, Direct, Upload), which consolidates what were previously two separate enums into one. Outputs additionally have an assetName field to register the produced artifact as a named data asset.

**Distribution and Resource Configuration**
The distribution field accepts a DistributionConfiguration discriminated on distributionType, with concrete subtypes for PyTorch (process count per instance), Mpi (process count per node), and TensorFlow (worker count and parameter server count). The resources field holds a JobResourceConfiguration with the VM SKU, node count, shared memory size, and extra Docker run arguments. The limits field holds a CommandJobLimits with a timeout expressed as an ISO 8601 duration on the wire; the SDK patch layer will accept an integer number of seconds and convert it to the correct duration format before sending.


specification/ai-foundry/data-plane/Foundry/relocate-beta-operations.tsp
```diff
@@ -76,4 +76,8 @@ interface Toolsets {}
 @@clientLocation(Azure.AI.Projects.Toolsets.listToolsets, Toolsets);
 @@clientLocation(Azure.AI.Projects.Toolsets.deleteToolset, Toolsets);
 
-
+@operationGroup
+namespace Training {
+  @operationGroup
+  interface Jobs {}
+}
```
dargilco, line 81: What is this for?

specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -235,6 +236,21 @@ using Azure.AI.Projects;
 // Less generic names
 @@clientName(Sku, "ModelDeploymentSku");
 
+@@clientName(Azure.AI.Projects.Training.Jobs.get, "get", "python");
```
dargilco: Please add comment above

// Training sub‐client

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,405 @@
...
+  @doc("Name of the output data asset to register.")
+  assetName?: string;
+}
+
+#suppress "@azure-tools/typespec-azure-core/no-string-discriminator" "Use an extensible union instead of a plain string"
+@doc("Nodes that user would like to start the service on.")
+@discriminator("nodesValueType")
+model NodeCollection {
+  @doc("Type of the Nodes value.")
+  nodesValueType: string;
```
dargilco, line 130: Should this be "nodeValueType" (singular)?

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,405 @@
...
+
+namespace Azure.AI.Projects;
+
+#suppress "@azure-tools/typespec-azure-core/no-string-discriminator" "Use an extensible union instead of a plain string"
+@doc("Distribution configuration of the job. If set, this should be one of Mpi, Tensorflow, PyTorch, or null.")
+@discriminator("distributionType")
+model DistributionConfiguration {
+  @doc("Specifies the type of distribution framework.")
+  distributionType: string;
+}
```
dargilco, line 14: I believe we just use "type" as discriminators (not "distributionType"). Similar in other models below. I don't see a need to call the discriminator something else depending on the class. Azure standards say use "kind", but we are following OpenAI example in Foundry.

specification/ai-foundry/data-plane/Foundry/relocate-beta-operations.tsp
jayesh-tanna: This is a stub namespace used to register the Training.Jobs operation group in the beta surface, which controls how the SDK client exposes the operations via client.beta.training.jobs. Without it, the operations wouldn't be grouped under the correct sub-client path in the generated SDK.

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
jayesh-tanna: API accepting only nodesValueType. I can add encodename to make it singular., let me know if required. but for now, its read based only. customer cant set it, customer can get the job and see what services are running.

specification/ai-foundry/data-plane/Foundry/relocate-beta-operations.tsp
dargilco: Okay for now, but as I think I mentioned, Johan did not want us to start creating nested levels of sub clients. So we'll need to discuss it with him.

specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -236,6 +237,21 @@ using Azure.AI.Projects;
 // Less generic names
 @@clientName(Sku, "ModelDeploymentSku");
 
+@@clientName(Azure.AI.Projects.Jobs.get, "get", "python");
```
dargilco, line 240: Add this comment before the jobs lines

// Jobs sub‐client

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,363 @@
...
+  string,
+
+  @doc("URI file asset.")
+  UriFile: "uri_file",
+
+  @doc("URI folder asset.")
+  UriFolder: "uri_folder",
+
+  @doc("Safetensors model asset.")
+  SafetensorsModel: "safetensors_model",
```
dargilco, line 77: SafeTensorsModel: "safe_tensors_model",

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,363 @@
...
+
+  @doc("Literal value (inputs only).")
+  Literal: "literal",
+}
+
+@doc("Job input definition.")
+model Input {
+  @doc("Specifies the type of job input.")
```
dargilco, line 85: Please have a more descriptive model name. Avoid generic one word names. These get emitted as SDK classes. How about JobInputDefintion? or JobDefintion?

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,363 @@
...
+
+  @doc("Literal value. Required for literal type.")
+  value?: string;
+
+  @doc("Description for the input.")
+  description?: string;
+}
+
+@doc("Job output definition.")
+model Output {
```
dargilco, line 101: Same here. More descriptive model name.

specification/ai-foundry/data-plane/Foundry/src/jobs/models.tsp
```diff
@@ -0,0 +1,363 @@
...
+
+  @doc("Name of the output data asset to register.")
+  assetName?: string;
+
+  @doc("Description for the output.")
+  description?: string;
+}
+
+#suppress "@azure-tools/typespec-azure-core/no-string-discriminator" "Use an extensible union instead of a plain string"
+@doc("Nodes that user would like to start the service on.")
```
dargilco, line 126: This sounds better: “Nodes where the user wants the service to start.”

dargilco (changes requested):

dargilco:
Please hold off on merging the PR. I want to get Johan's feedback on introducing nested sub-clients like `.beta.training.jobs`.  I'll start a thread with him and you.

specification/ai-foundry/data-plane/Foundry/src/jobs/routes.tsp
```diff
@@ -0,0 +1,95 @@
+import "./models.tsp";
```
dargilco, line 1: If "jobs" operations only apply to training, I think a better directory structure would be to put the "jobs" folder under the "training" folder.

dargilco (changes requested):

PR #41634, opened 2026-03-22
MPG migration - NotificationHubs

haiyuazhang:
Migrate NotificationHubs service to use new management plane SDK generator.

- Add C#-scoped clientName decorators in client.tsp
- Add flatten and clientLocation decorators in back-compatible.tsp  
- Update tspconfig.yaml with C# mgmt emitter config
- Update SharedAccessAuthorizationRuleResource.tsp for custom listKey return type

### DIFF
--- a/specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
+++ b/specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
@@ -106,7 +106,8 @@ using Microsoft.NotificationHubs;
   "csharp"
 );
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Legacy decorator required for SDK compatibility"
-@@Azure.ClientGenerator.Core.Legacy.flattenProperty(PnsCredentialsResourceReplacement.properties
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(PnsCredentialsResourceReplacement.properties,
+  "csharp"
 );
 @@clientName(PolicyKeyResource, "NotificationHubPolicyKey", "csharp");
 @@clientName(ResourceListKeys, "NotificationHubResourceKeys", "csharp");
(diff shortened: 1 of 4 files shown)


specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
```diff
@@ -23,3 +23,299 @@ using Microsoft.NotificationHubs;
...
+  Azure.ResourceManager.Foundations.TrackedResource,
+  "csharp"
+);
+
+// C#-only replacement models — swap ProxyResource → TrackedResource to preserve TrackedResourceData base class.
+
+/**
+ * Properties for CheckAvailabilityResult replacement model.
+ */
+model CheckAvailabilityResultProperties {
```
live1206: is this in use?

specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
```diff
@@ -23,3 +23,299 @@ using Microsoft.NotificationHubs;
...
+   * create new Namespace/NotificationHub. Otherwise false.
+   */
+  isAvailiable?: boolean;
+}
+
+/**
+ * Replacement for CheckAvailabilityResult — uses TrackedResource base for C# backward compat.
+ */
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "C#-only replacement model"
+model CheckAvailabilityResultReplacement
```
live1206: could you give more context why hirarchyBuilding decorator could not handle this? This is a workaround we would like to avoid if possible. And, there are many more.

live1206: At least, we should add TODO here to remove these workaround when https://github.com/microsoft/typespec/issues/9234 is fixed. I will prioritize the fix for this.

specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
haiyuazhang: removed

specification/notificationhubs/resource-manager/Microsoft.NotificationHubs/NotificationHubs/client.tsp
haiyuazhang: only two replacement models are still required with comments explain the why part.

live1206 (approved):

PR #41691, opened 2026-03-24
feat(agents): add session lifecycle CRUD API surface

aryaan-singh:
Add TypeSpec models and routes for session lifecycle management:
- Create, Get, Delete (LRO), List, GetOperation endpoints
- SessionResource, VersionIndicator (discriminated), SessionOperationObject
- PascalCase enums aligned with C# contracts (PR 2011590)
- Continuation-token pagination (SessionListResult)
- x-session-isolation-key header on mutating operations
- 5 example JSON files

AB#5122803

# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -947,3 +947,303 @@ model AgentCard {
...
+  #{
+    required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview],
+  }
+)
+@removed(Versions.v1)
+union VersionIndicatorType {
+  string,
+
+  /** Direct reference to a specific agent version. */
+  version_ref: "version_ref",
```
baywet: in this definition we have a mix between UpperCamelCase and snake_case. (see the previous unions), this does seem inconsistent.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -947,3 +947,303 @@ model AgentCard {
...
+    required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview],
+  }
+)
+@removed(Versions.v1)
+model VersionRefIndicator extends VersionIndicator {
+  /** Discriminator value for version_ref. */
+  type: VersionIndicatorType.version_ref;
+
+  /** The agent version identifier. */
+  agent_version: string;
```
baywet, line 1064: shouldn't this be a positive integer?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -947,3 +947,303 @@ model AgentCard {
...
+
+  /** The version indicator determining which agent version backs this session. */
+  version_indicator: VersionIndicator;
+
+  /** The current status of the session. */
+  status: AgentSessionStatus;
+
+  /** The Unix timestamp (in seconds) when the session was created. */
+  @visibility(Lifecycle.Read)
+  created_at: int64;
```
baywet: I think we're missing encode attributes here. Something like this https://github.com/microsoft/typespec/blob/d2c625a6c47d7554e7f87e314d37ecba7f4a9a0b/packages/http-specs/specs/encode/datetime/main.tsp#L12

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -947,3 +947,303 @@ model AgentCard {
...
+  }
+)
+@removed(Versions.v1)
+model AgentSessionOperationError {
+  /** The error code. */
+  code: string;
+
+  /** A human-readable error message. */
+  message: string;
+}
```
baywet: this error type definition seems pretty standard. Don't we have one we can reuse across the services to avoid sprawl of those kind of definitions?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -947,3 +947,303 @@ model AgentCard {
...
+model AcceptedAgentSessionDeleteResponse {
+  @statusCode
+  code: 202;
+
+  /** Operation ID for tracking the delete LRO. */
+  @header("x-ms-session-operation-id")
+  operation_id: string;
+
+  @body
+  body: AgentSessionResource;
```
baywet: is it standard practice to return the deleted resource? we have an opportunity to reduce COGS here for everyone.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
aryaan-singh: Good catch. The inconsistency here was between the **TypeSpec enum member name** and the rest of the model surface, not the wire value itself. I updated the declaration to:

```tsp
union VersionIndicatorType {
  string,
  VersionRef: "version_ref",
}
```

So the TypeSpec-facing name is now PascalCase (`VersionRef`) and consistent with the surrounding unions, while the serialized wire value remains `"version_ref"` for the discriminator.

Hope that addresses your concern

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
aryaan-singh: Agreed — I added `@encode("unixTimestamp", int64)` to all three session timestamp fields:

- `created_at`
- `last_accessed_at`
- `expires_at`

They are now modeled as `utcDateTime` in TypeSpec with explicit unix timestamp encoding on the wire.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
aryaan-singh: As part of the synchronous CRUD cleanup I removed the entire session-operation tracking surface (`getSessionOperation`, `AgentSessionOperationObject`, and the custom operation error shape), so we no longer need a bespoke error model here. The remaining CRUD operations now just rely on the shared default error response pattern.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
aryaan-singh: Good point. I changed the delete operation to return `204 No Content` instead of returning the deleted session resource. That aligns better with normal delete semantics and avoids the extra payload/cost for a terminal delete response.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
aryaan-singh: This field is intended to carry the agent version identifier/name returned by the agent version APIs, not a numeric revision counter, so string is the expected shape here. It also matches the
 existing route parameters where agent_version is modeled as a string path segment across the agent version endpoints.

andscho-msft (approved):

PR #41700, opened 2026-03-24
Data Locality (Get Blob Layout)

nickliu-msft:


specification/storage/data-plane/Microsoft.BlobStorage/stable/2026-10-06/blob.json
```diff
@@ -3227,25 +3232,925 @@
               "x-ms-encryption-scope": {
                 "x-ms-client-name": "EncryptionScope",
                 "type": "string",
-                "description": "Returns the name of the encryption scope used to encrypt the blob contents and application metadata.  Note that the absence of this header implies use of the default account encryption scope."
+                "description": "Returns the name of the encryption scope used to encrypt the blob contents and application metadata.  Note that the absence of this header implies use of the default account encryption scope."
```
gapra-msft, line 3242: Should smart access tier be added as a return type here?

nickliu-msft: Yeah it should, forgot to add it before STG102 was merged in

gapra-msft (approved):

PR #41759, opened 2026-03-25
Add WebApps tags to operations for SDK backward compatibility

kamperiadis:
With the typespec migration, we did not keep the same tags as we did in the previous versions

Example - the /sites and /sites/config/web operations used to have the same tags before but currently they have different tags:

[GET /sites operation pre-typespec migration](https://github.com/Azure/azure-rest-api-specs/blob/5e21babae10dd7056c3005bb51da815f0495f1aa/specification/web/resource-manager/Microsoft.Web/AppService/stable/2023-12-01/WebApps.json#L110)
[GET /sites/config/web operation pre-typespec migration](https://github.com/Azure/azure-rest-api-specs/blob/5e21babae10dd7056c3005bb51da815f0495f1aa/specification/web/resource-manager/Microsoft.Web/AppService/stable/2023-12-01/WebApps.json#L2381)

[GET /sites operation post-typespec migration](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-05-01/openapi.json#L7378)
[GET /sites/config/web operation post-typespec migration](https://github.com/Azure/azure-rest-api-specs/blob/5e21babae10dd7056c3005bb51da815f0495f1aa/specification/web/resource-manager/Microsoft.Web/AppService/stable/2025-05-01/openapi.json#L9985)

The consequence of this is that now the Go SDK (and potentially other SDKs) are being generated with multiple packages to represent one same entity and this is making it more difficult to develop on the client tools and ensure there are no bugs introduced by bumping the SDK version.

### DIFF
--- a/specification/web/resource-manager/Microsoft.Web/AppService/AddressResponse.tsp
+++ b/specification/web/resource-manager/Microsoft.Web/AppService/AddressResponse.tsp
@@ -32,7 +32,8 @@ model AddressResponse
   kind?: string;
 }
 
-@armResourceOperations
+@tag("AppServiceEnvironments")
+@armResourceOperations(#{ omitTags: true })
 interface AddressResponses {
   /**
    * Description for Get IP addresses assigned to an App Service Environment.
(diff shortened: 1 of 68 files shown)


danielw5 (approved):

ravimeda:
Please address the 'Required' checks.

danielw5 (approved):

PR #41784, opened 2026-03-26
Data Generation API specs: Backward compat changes for Foundry UI

shivamMg:


specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -64,12 +67,12 @@ model DataGenerationJob {
   @visibility(Lifecycle.Read)
   status: DataGenerationJobStatus;
 
-  @doc("The job outputs: Azure OpenAI files for fine-tuning, or datasets for evaluation.")
+  @doc("The final job outputs: Azure OpenAI files for fine-tuning, or datasets for evaluation. This field is populated only after the job has succeeded.")
```
sakoll, line 67: This is not true for eval case. For eval scenario, end user can enter expected dataset name and few more things added in dataset output model

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -6,23 +6,23 @@ namespace Azure.AI.Projects;
...
-  @doc("Multi turn conversation between agents and user.")
+  @doc("Multi turn conversation between user and agent.")
   MultiTurn: "MultiTurn",
 }
 
 @doc("LLM model options for data generation jobs.")
 model DataGenerationModelOptions {
-  @doc("Base model name used to generate data.")
-  model_name: string;
+  @doc("Model deployment name used to generate data.")
```
sakoll: As explained earlier, AOAI is now supporting deployment less models. So this necessarily doesn't need to be a deployment name

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -6,23 +6,23 @@ namespace Azure.AI.Projects;
...
+  `model`: string;
```
sakoll: This was changed to model_name based on April's https://github.com/Azure/azure-rest-api-specs/pull/41606#discussion_r2985367240 on main PR.

 If it is only for backward compatibility, can we just check for either of these properties?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -279,8 +276,8 @@ model FileDataGenerationJobSource extends DataGenerationJobSource {
   @doc("The source type for this job, which is File.")
   type: "File";
 
-  @doc("Input Azure Open AI file id references used for data generation.")
-  file_id: string;
+  @doc("Input Azure Open AI file id used for data generation.")
+  id: string;
```
sakoll: The OAI/AOAI contracts generally refer this as `file_id`. It is better to align with them?

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -304,3 +301,12 @@ union SimpleQnASupervisedFineTuningQuestionType {
   @doc("Long answer question type.")
   LongAnswer: "LongAnswer",
 }
+
+@doc("Error information for a data generation job that failed.")
+model DataGenerationJobError {
```
sakoll: This was again removed based on one of the comment to reuse existing ApiError contract

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
```diff
@@ -53,9 +53,12 @@ model DataGenerationJob {
   @doc("The options for the data generation job.")
   options: DataGenerationJobOptions;
 
-  @doc("The purpose of the data generation job. Either for fine-tuning supervised training or evaluation.")
+  @doc("The purpose of the data generation job. Either for fine-tuning or evaluation.")
   purpose: DataGenerationJobPurpose;
 
+  @doc("The model deployment options.")
+  model_options?: DataGenerationModelOptions;
```
sakoll: Similar to model, for UX backward compatibility can we also support model options either from here and one in DataGenerationJobs. Once UX adopts to new Api, we can remove this for backward compatibility

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
sakoll: ApiError already contains these properties, so we don't need this model

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: this is the "outputs" field which is populated in the end. i think you're referring to sources

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: can we just keep it `model` please? it helps with easing backward compatibility work + we can still pass deployment-less model names to it (model is not very different from model_name)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: (replied in the other comment around backward compatibility)

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: done - reverted back to file_id

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: done - reverted back to ApiError. i didn't notice that ApiError is just OpenAI.Error which serves our purpose

specification/ai-foundry/data-plane/Foundry/src/data_generation/models.tsp
shivamMg: okay. i've put it back

PR #41819, opened 2026-03-26
[Blob Storage] xml.Name CorsRule

gracewilcox:
Using `xml.Name` decorator to add xml name to model `CorsRule`


catalinaperalta (commented):
lgtm. Leaving approval for storage team

jalauzon-msft (approved):

vincenttran-msft:
image
Caveat as discussed offline, functionally fine in Rust but does seem like a no-op in this case?

vincenttran-msft (approved):
LGTM, no-op in Rust as discussed offline

PR #41820, opened 2026-03-26
[Queues] xml.name additions

gracewilcox:
Adding `xml.Name` decorators to models. Needed for proper Go generation 

This change also helps match the original swagger more closely. Example: https://github.com/Azure/azure-rest-api-specs/blob/5f1f13276e3bba25876eeabedc765ac987c3b457/specification/storage/data-plane/Microsoft.QueueStorage/preview/2018-03-28/queue.json#L1459


catalinaperalta (commented):
lgtm. Leaving approval for storage team

jalauzon-msft (approved):

vincenttran-msft (approved):
LGTM! Tested these changes in Rust and just got some more serde renames but all functionally equivalent!

PR #41842, opened 2026-03-27
[Foundry] Agent Endpoints

glecaros:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -70,12 +80,51 @@ model CreateAgentRequest {
...
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[FoundryFeaturesOptInKeys.agent_endpoint_v1_preview] }
+)
+model PatchAgentRequest {
+  ...MetadataPropertyForRequest;
+
+  @doc("A human-readable description of the agent.")
+  @maxLength(512)
+  description?: string;
```
johanste: I assume the properties here are `null`able (which is how you would clear the value)?

(i.e. agent_endpoint, agent_card and description should all explicitly state that `null` is a valid type/value)

glecaros: I worry here that this would be problematic for some serializers that will by default elide null properties (could affect our client libraries or customers that are rolling their own)

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -413,3 +540,251 @@ union ContainerLogKind {
...
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionOptInKeys.agent_endpoint_v1_preview] }
+)
+@discriminator("type")
+model VersionSelectionRule {
+  type: VersionSelectorType;
+
+  @doc("The version to rout traffic to")
+  agentVersion: string;
```
glecaros: for consistency with the rest of the surface, this should be snake_case

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
andscho-msft: Isn't that what the `?` does?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
acsoylemezoglu: we changed it to be snake case

andscho-msft: https://github.com/Azure/azure-rest-api-specs/commit/b5813e2c97731f5f618b1d21f6c896a23f59f3da

acsoylemezoglu: https://github.com/Azure/azure-rest-api-specs/pull/41981/changes

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
acsoylemezoglu: This was removed from the PatchRequest

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
glecaros: great!

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
glecaros: not exactly, ? makes it optional. Without making it explicitly nullable, null is not a valid value. The problem is that many serializer by default will not include null values (which makes this dangerous territory)

acsoylemezoglu (approved):

johanste (approved):

johanste (approved):

PR #41867, opened 2026-03-27
[Foundry] API Key

glecaros:
Staging branch PR #41474


dargilco:
Closing since we decided not to support api-key auth on AIProjectClient

PR #41914, opened 2026-03-30
Restore ML model.tsp

kashifkhan:
# Choose a PR Template

### DIFF
--- a/specification/machinelearningservices/MachineLearningServices.Management/models.tsp
+++ b/specification/machinelearningservices/MachineLearningServices.Management/models.tsp
@@ -3,7 +3,6 @@ import "@typespec/http";
 import "@typespec/openapi";
 import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-azure-resource-manager";
-import "@azure-tools/typespec-client-generator-core";
 
 using TypeSpec.Rest;
 using TypeSpec.Http;
@@ -11,7 +10,6 @@ using TypeSpec.OpenAPI;
 using Versioning;
 using Azure.Core;
 using Azure.ResourceManager;
-using Azure.ClientGenerator.Core;
 using Azure.ResourceManager.Foundations;
 
 namespace Microsoft.MachineLearningServices;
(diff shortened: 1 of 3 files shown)


fmabroukmsft (approved):

specification/machinelearningservices/MachineLearningServices.Management/models.tsp
```diff
@@ -71,11 +69,11 @@ union VMPriceOSType {
  */
 union VMTier {
   string,
-  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Standard VM tier with guaranteed capacity"
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Existing API contract - suppression reason updated"
   Standard: "Standard",
-  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Low priority VM tier with discounted pricing"
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Existing API contract - suppression reason updated"
   LowPriority: "LowPriority",
-  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Spot VM tier with variable pricing and availability"
```
catalinaperalta, line 78: Looks like a lot of the old doc suppression reasons were actually just the docs that should have been placed over the properties, do you think you could get copilot to just make these the docs?

fmabroukmsft: Hi @catalinaperalta, that's actually from the previous PRs immediately after the Typespec conversion. The Typespec conversion added many suppressions for missing documentation that wasn't required from OpenAPI swagger. Because the backend has swagger equivalency validation checks, the solution that was agreed upon with the ARM was to add the proposed descriptions as suppressions, and as the backend code is updated the suppressions can be officially applied to the properties. But that will take time

catalinaperalta: Got it, thanks for the context! So why are the valid docs being replaced by a more generic message for the suppressions? Are the docs saved somewhere else?

fmabroukmsft: models.tsp is being restored to the right state, there was a PR a few days ago that unintentionally touched it for data plane changes. So the docs were never actually published/available to the customers previously, but with time and confirmation with the backend team, the valid docs that live as a suppression right now will eventually be published

PR #41920, opened 2026-03-30
[SDK Validation] Make ApiVersion pipeline parameter truly optional

### LINKED ISSUE #14676: Support Default “Latest API Version” in Release Planner & release pipeline for Multi‑Service SDK Packages
## Background
Some SDK packages represent multiple services within a single package (for example, Compute), where:

- Each service may use different API versions
- There is no way to select a specific service package independently

In these scenarios, requiring users to explicitly select an API version in the tooling creates friction and does not reflect how the SDK package is structured or consumed.

Image

## Goal

Enable the release planner and release pipeline to default to using the latest API version per service for multi‑service SDK packages, without requiring users to manually select an API version.

In the future, we may require the feature of letting service team to specify explicit API version selection for each service, we will see whether there is a real requirement before asking for this.

raych1:
Resolved https://github.com/Azure/azure-sdk-tools/issues/14676 partially.

## Problem
The `ApiVersion` parameter in `spec-gen-sdk.yml` had `default: ''` (empty string). Azure DevOps treats empty-string defaults for string parameters as effectively required — the **Run** button won't activate until the user types something.

## Solution
- Changed the `ApiVersion` default from `''` to `'none'` (sentinel value) so the Run button works without user input.
- Updated the runtime check in the archetype template to treat `'none'` the same as empty string — skipping `--api-version` and `--sdk-release-type` flags.
- Added a clarifying note to the `SdkReleaseType` displayName: it is ignored when ApiVersion is `"none"`.


raych1:
@weshaggard could your review this PR about the API-version parameter change in SDK generation pipeline?

eng/pipelines/spec-gen-sdk.yml
```diff
@@ -16,15 +16,15 @@ parameters:
     displayName: 'Path to API specification file (TypeSpec/OpenAPI)'
   - name: ApiVersion
     type: string
-    default: ''
+    default: 'none'
```
weshaggard, line 19: minor: Might consider "latest" here but none also works.

weshaggard (approved):

eng/pipelines/spec-gen-sdk.yml
raych1: Users may be confused on which "latest": 'preview' vs 'stable'. I'll go with 'none' for now.

PR #41922, opened 2026-03-30
Add CognitiveServices preview API version 2026-03-15-preview

fmabroukmsft:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

> **🤖 This PR was created by swagger-agent.** Adds CognitiveServices preview API version `2026-03-15-preview`: TypeSpec enum, compiled cognitiveservices.json, examples, readme.md tag section. **Stacked on #41835 (2026-03-01 stable) — merge that first, then retarget this to main.**

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


psah434:
required check typespec validation is failing.

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
```diff
@@ -0,0 +1,106 @@
...
+    KeyName = "deploymentName",
+    SegmentName = "managedComputeDeployments",
+    NamePattern = "^[a-zA-Z0-9][a-zA-Z0-9._-]{0,63}$"
+  >;
+
+  /**
+   * The resource model definition representing SKU
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "Existing API defines 'sku' on the resource envelope for backward compatibility"
+  sku?: Sku;
```
razvanbadea-msft, line 35: >sku [](http://example.com/codeflow?start=2&length=3)

to avoid these suppressions you should try to use the builtin ones

...ResourceSkuProperty;
...EntityTagProperty;

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Creates or updates a managed compute deployment associated with the Cognitive Services account.
+   */
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<ManagedComputeDeployment>;
+
+  /**
+   * Updates the specified managed compute deployment associated with the Cognitive Services account.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-patch" "PATCH for 'createOrUpdate' follows existing service implementation pattern"
```
razvanbadea-msft, line 61: >suppress [](http://example.com/codeflow?start=3&length=8)

why is this suppression added? you are using ArmCustomPatchAsync which is ok, and the reason is not clear createOrUpdate is for PUT but here is about PATCH

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
```diff
@@ -0,0 +1,106 @@
...
+      Azure.Core.Foundations.RetryAfterHeader &
+      ArmAsyncOperationHeader,
+    PatchModel = PatchResourceSku
+  >;
+
+  /**
+   * Deletes the specified managed compute deployment associated with the Cognitive Services account.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "ManagedComputeDeployment can have dependent resources, deleting it will be a LRO, result code will be 202. If the dependent resource is not provisioned, the result code will be 200."
+  delete is ArmResourceDeleteWithoutOkAsync<
```
razvanbadea-msft, line 79: >ArmResourceDeleteWithoutOkAsync [](http://example.com/codeflow?start=12&length=31)

have you tried using ArmResourceDeleteAsync , this might generate what you want without the need to add a suppresion

razvanbadea-msft: Also why do you want to return 200? Why not always return 202, even when there are no dependent resources? The LRO can complete immediately. This avoids mixing 200 and 202 response codes and gives SDK consumers a single consistent code path

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -30,9 +30,107 @@ These are the global settings for the CognitiveServices API.
...
+  - code: ArmResourcePropertiesBag
+    reason: This API is copied from Machine Learning Services RP where this behavior is already established.
+    where:
+      - $.definitions.OutboundRuleBasicResource
+  - code: AvoidAdditionalProperties
+    reason: This API is copied from Machine Learning Services RP where this behavior is already established.
+    where:
+      - $.definitions.ManagedNetworkSettings.properties.outboundRules
+  - code: ProvisioningStateSpecifiedForLROPut
+    reason: This API is copied from Machine Learning Services RP where this behavior is already established.
```
razvanbadea-msft, line 57: Copying an API from another RP doesn't automatically justify carrying over its spec violations. Each suppression should stand on its own technical merits. For a new API version, is there an opportunity to fix these rather than perpetuate them?

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -30,9 +30,107 @@ These are the global settings for the CognitiveServices API.
...
+    reason: Approved to be suppressed in AML swagger.
+    where:
+      - $.definitions.ConnectionOAuth2.properties.clientId.format
+  - code: DeleteResponseCodes
+    reason: Behavior is align with other existing API for this RP
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/projects/{projectName}"].delete
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/managedComputeDeployments/{deploymentName}"].delete
+  - code: AvoidAdditionalProperties
+    reason: Approved to be suppressed in AML swagger
```
razvanbadea-msft, line 94: ARM is trying to reduce instances of unschematized APIs to make it easier for consumers of those APIs to understand them and use them reliably, not to mention provide better support in the various SDKs. 

Use of additionalProperties is not allowed for properties owned by the service. The only time its ok to use it is when the properties are pass thru (user defined) and not subject to any validations. Please make this an array or provide an explanation for why you need this.

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -30,9 +30,107 @@ These are the global settings for the CognitiveServices API.
...
+    reason: Align with existing API behavior in other APIs
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/projects/{projectName}"].put.responses.202
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/projects/{projectName}"].patch.responses.202
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/projects/{projectName}"].delete.responses.202
+  - code: AvoidAdditionalProperties
+    reason: Same as existing account resource, trying to have the same behavior
+    where:
+      - $.definitions.ProjectProperties.properties.endpoints
+  - code: AvoidAdditionalProperties
```
razvanbadea-msft, line 108: ARM is trying to reduce instances of unschematized APIs to make it easier for consumers of those APIs to understand them and use them reliably, not to mention provide better support in the various SDKs. 

Use of additionalProperties is not allowed for properties owned by the service. The only time its ok to use it is when the properties are pass thru (user defined) and not subject to any validations. Please make this an array or provide an explanation for why you need this.

specification/cognitiveservices/resource-manager/readme.md
fmabroukmsft: Thanks for the feedback. Updated the suppression justification in commit 8f430c5.

**ProvisioningStateSpecifiedForLROPut** — ManagedNetwork PUT is an async operation that returns 202 Accepted for long-running network infrastructure provisioning (creating VNets, NSGs, private endpoints, etc.). The backend is shared with Machine Learning Services RP and returns 202 instead of ARM-standard 200/201. Changing the response codes would be a breaking change for existing clients — this pattern has shipped in stable since 2025-12-01. `provisioningState` is present in the resource model and returned in the final LRO result.

We can't fix this without backend changes across the MFE→NetworkManagement pipeline, which would also break existing MLS clients using the same shared backend.

specification/cognitiveservices/resource-manager/readme.md
fmabroukmsft: Updated the suppression justification in commit 8f430c5.

**ConnectionPropertiesV2.metadata** — This is a user-defined key-value property bag where customers store arbitrary metadata about their connections (e.g., `{"ApiVersion": "2024-02-01", "DeploymentApiVersion": "2024-06-01"}`). The keys are not predefined by the service — customers choose them. The backend stores this as `Dictionary<string, string>`.

**CustomKeys.keys** — This holds user-provided authentication credentials with user-defined key names (e.g., `{"api_key": "xxx"}`). The `CustomKeys` auth type exists specifically for arbitrary named credentials.

Both are genuinely user-defined pass-through dictionaries — converting to arrays would break the data model semantics (these are key→value mappings, not ordered lists).

specification/cognitiveservices/resource-manager/readme.md
fmabroukmsft: Updated the suppression justification in commit 8f430c5.

**ProjectProperties.endpoints** — This is a read-only dictionary of service endpoint names to URLs (e.g., `{"OpenAI": "https://...", "Speech": "https://...", "ContentSafety": "https://..."}`). The keys are dynamically determined at runtime by the account's kind and region from the service's API catalog (`PopulateEndpointsAsync` in Platform-RP iterates the API set and builds the dictionary). Different account kinds expose different sets of endpoints, so the keys cannot be predefined in a static schema.

Same pattern as `AccountProperties.endpoints` which has shipped in stable since the original TypeSpec migration (2025-12-01). The backend type is `Dictionary<string, string>`.

razvanbadea-msft:
Fix errors from lintdiff and typespec validation checks

razvanbadea-msft:
>   #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "'migrationToken' is correctly marked as secret matching service-side security handling"

[NEW] 🔴 Blocking [RPC-Put-V1-13 / XMSSecretInResponse] ManagedComputeDeployment.tsp / generated cognitiveservices.json — path $.definitions.*.properties.*.migrationToken and ApiProperties.properties.qnaAzureSearchEndpointKey
Issue: These three secret-like properties surface in GET/PUT responses without x-ms-secret: true. LintDiff flags 3 XMSSecretInResponse occurrences. Even if inherited from the Account model (existing), any new model under the preview that re-composes Account/ApiProperties must re-examine secret handling.
Fix: Add @secret on the TypeSpec property (emits x-ms-secret: true) or set write-only visibility so the property is stripped from responses.

---
Refers to: specification/cognitiveservices/CognitiveServices.Management/models.tsp:1508 in 11f1450. [](commit_id = 11f145088edb9fa25e0ffcbd1c03ea6ab41d1e50, deletion_comment = False)

razvanbadea-msft:
>   #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "'qnaAzureSearchEndpointKey' is correctly marked as secret matching service-side security handling"

[NEW] 🔴 Blocking [RPC-Put-V1-13 / XMSSecretInResponse] ManagedComputeDeployment.tsp / generated cognitiveservices.json — path $.definitions.*.properties.*.migrationToken and ApiProperties.properties.qnaAzureSearchEndpointKey
Issue: These three secret-like properties surface in GET/PUT responses without x-ms-secret: true. LintDiff flags 3 XMSSecretInResponse occurrences. Even if inherited from the Account model (existing), any new model under the preview that re-composes Account/ApiProperties must re-examine secret handling.
Fix: Add @secret on the TypeSpec property (emits x-ms-secret: true) or set write-only visibility so the property is stripped from responses.

---
Refers to: specification/cognitiveservices/CognitiveServices.Management/models.tsp:1913 in 11f1450. [](commit_id = 11f145088edb9fa25e0ffcbd1c03ea6ab41d1e50, deletion_comment = False)

razvanbadea-msft:
> @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v3)

you should plan to update to v5/v6 if possible

---
Refers to: specification/cognitiveservices/CognitiveServices.Management/main.tsp:57 in 11f1450. [](commit_id = 11f145088edb9fa25e0ffcbd1c03ea6ab41d1e50, deletion_comment = False)

specification/cognitiveservices/CognitiveServices.Management/ComputeOperation.tsp
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.CognitiveServices;
  */
 @added(Versions.v2026_01_15_preview)
 @removed(Versions.v2026_03_01)
+@added(Versions.v2026_03_15_preview)
```
razvanbadea-msft, line 21: **[NEW] 🟡 Warning** **[ParameterNotUsingCommonTypes]** — Generated path parameter at /subscriptions/{subscriptionId}/providers/Microsoft.CognitiveServices/locations/{location}/computeOperations/{operationId} is inlined instead of referencing `common-types/resource-management/v3/types.json#/parameters/OperationIdParameter`. LintDiff is failing with this new violation (net delta +1 vs stable 2026-03-01).

**Root cause:** `ResourceNameParameter<..., KeyName = "operationId">` with `NamePattern = ""` emits an inline schema with no common-types ``.

**Fix (Option A — recommended):** Replace the `ResourceNameParameter` spread with an explicit `name` property decorated with `@armCommonParameter` so the emitter produces a `` to common-types v3 (matching the rest of this spec — do not mix common-types versions within the same file):

```typespec
@added(Versions.v2026_01_15_preview)
@removed(Versions.v2026_03_01)
@added(Versions.v2026_03_15_preview)
@subscriptionResource
@parentResource(SubscriptionLocationResource)
model ComputeOperationStatus
  is Azure.ResourceManager.ProxyResource<ComputeOperationStatusProperties> {
  /**
   * The ID of the compute operation.
   */
  @path
  @key("operationId")
  @segment("computeOperations")
  @visibility(Lifecycle.Read)
  @armCommonParameter(
    "OperationIdParameter",
    Azure.ResourceManager.CommonTypes.Versions.v3
  )
  name: string;
}
```

This removes the inline parameter, eliminates the empty `NamePattern = """""` code smell, and satisfies the `ParameterNotUsingCommonTypes` rule without needing a suppression. Keep the common-types version at v3 to stay consistent with the rest of `cognitiveservices.json` (a bulk upgrade to v6 can be tracked separately).

razvanbadea-msft, line 21: **[NEW] 🟡 Warning** **[ParameterNotUsingCommonTypes]** — Generated path parameter at /subscriptions/{subscriptionId}/providers/Microsoft.CognitiveServices/locations/{location}/computeOperations/{operationId} is inlined instead of referencing `common-types/resource-management/v3/types.json#/parameters/OperationIdParameter`. LintDiff is failing with this new violation (net delta +1 vs stable 2026-03-01).

**Root cause:** `ResourceNameParameter<..., KeyName = "operationId">` with `NamePattern = ""` emits an inline schema with no common-types ``.

**Fix (Option A — recommended):** Replace the `ResourceNameParameter` spread with an explicit `name` property decorated with `@armCommonParameter` so the emitter produces a `` to common-types v3 (matching the rest of this spec — do not mix common-types versions within the same file):

```typespec
@added(Versions.v2026_01_15_preview)
@removed(Versions.v2026_03_01)
@added(Versions.v2026_03_15_preview)
@subscriptionResource
@parentResource(SubscriptionLocationResource)
model ComputeOperationStatus
  is Azure.ResourceManager.ProxyResource<ComputeOperationStatusProperties> {
  /**
   * The ID of the compute operation.
   */
  @path
  @key("operationId")
  @segment("computeOperations")
  @visibility(Lifecycle.Read)
  @armCommonParameter(
    "OperationIdParameter",
    Azure.ResourceManager.CommonTypes.Versions.v3
  )
  name: string;
}
```

This removes the inline parameter, eliminates the empty `NamePattern = """""` code smell, and satisfies the `ParameterNotUsingCommonTypes` rule without needing a suppression. Keep the common-types version at v3 to stay consistent with the rest of `cognitiveservices.json` (a bulk upgrade to v6 can be tracked separately).

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -30,9 +30,107 @@ These are the global settings for the CognitiveServices API.
...
+    reason: Provide customers ability to assign customize labels to rules.
+    where:
+      - $.definitions.RaiToolLabelProperties.properties.accountScope.properties.labelValues
+      - $.definitions.RaiToolLabelProperties.properties.projectScopes.items.properties.labelValues
+      - $.definitions.RaiIfcRuleProperties.properties.conditions
+  - code: AvoidAdditionalProperties
+    reason: Provide customers ability to define custom conditions for when a rule is activated.
```
razvanbadea-msft, line 123: how a customer know what to add without a clear schema? Is there no validation for custom conditions?

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
ChunyuMSFT: We can't use ResourceSkuProperty, as we defined our own Sku which has name conflict.
We can't use EntityTagProperty either, as the field in it is eTag, our name is etag

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
ChunyuMSFT: in the suppresion section I documented the reason, the resource might not have physical backend resource to delete which doesn't need LRO

specification/cognitiveservices/CognitiveServices.Management/ManagedComputeDeployment.tsp
fmabroukmsft: Fixed — suppression removed in commit 4ca83c73.

fmabroukmsft:
These `#suppress` directives for `migrationToken` and `qnaAzureSearchEndpointKey` are pre-existing on `main` — not introduced by this PR. We tried replacing them with `@secret` in PR #42651 but it flags as a breaking change (`format: password` + `x-ms-secret: true` added to existing properties). We'll address this in a separate breaking-change PR with proper approvals.

fmabroukmsft:
## SDK CI Fix — Reverted `ResourceSkuProperty` / `EntityTagProperty`

Commit 27daf4dc87 reverts the `...ResourceSkuProperty` and `...EntityTagProperty` spreads back to explicit `sku?: Sku` and `etag?: string` properties. This fixes the Java and Python SDK validation failures.

### Root Cause

CogSvc defines its own `Sku` model (`models.tsp` line 1398) and `SkuTier` union (line 18) which are distinct from the ARM common-type `Sku`/`SkuTier`. When `...ResourceSkuProperty` was used, **both** the CogSvc and ARM `Sku` types appeared in the TypeSpec compilation, causing `duplicate-client-name` errors in the Java and Python SDK generators:

```
error @azure-tools/typespec-client-generator-core/duplicate-client-name: 
  Client name: "Sku" is duplicated in language scope: "java"
  Client name: "SkuTier" is duplicated in language scope: "python"
```

.NET, Go, and JS do not enforce this check, which is why they passed.

Similarly, `EntityTagProperty` uses the field name `eTag` (camelCase), but CogSvc existing API contract uses lowercase `etag`. Switching would be a breaking change.

### Bisection

| Commit | Java | Python |
|--------|------|--------|
| `14c4a7db` (Apr 22 — before change) | ✅ | ✅ |
| `6be55927` (Apr 24 — introduced ResourceSkuProperty) | ❌ | ❌ |

### What changed

- Reverted `...ResourceSkuProperty` → `sku?: Sku` with suppression
- Reverted `...EntityTagProperty` → `etag?: string` with suppression
- Restored `@@encodedName(ManagedComputeDeployment.etag, ...)`
- Updated suppression justifications to explain **why** the ARM common-type spreads cannot be used
- Kept the simplified `delete` operation from the ARM review feedback (unrelated to the Sku issue)

cc @ChunyuMSFT — this aligns with your comment that ResourceSkuProperty cannot be used due to the custom Sku definition.

rkuthala (approved):

rkuthala (approved):

PR #41987, opened 2026-04-01
Add data generation jobs with centralized LRO

sakoll:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/jobs/data_generation_models.tsp
```diff
@@ -56,14 +59,6 @@ model DataGenerationJob {
   @doc("The scenario of the data generation job. Either for fine-tuning or evaluation.")
```
aprilk-ms, line 59: If there is another team want to support creating dataset for their scenario. Will they extend DataGeneration job with their own scenario name? Or they need another job type?
Or should we subclass DataGenerationJob for eval and fine-tuned scenarios?

sakoll: Probably better thing to do is, we should own all the dataset creation part (datasets:generate) and internally based on requirements, we can fwd the request to corresponding team service.

specification/ai-foundry/data-plane/Foundry/src/jobs/data_generation_models.tsp
shivamMg, line 22: @sakoll, @aprilk-ms sorry to bring this up again - can we keep it as "model"? I understand that in future we're going to support auto-creation of model deployments given just the model name (e.g. gpt-5, gpt-5.2-chat). But this field also needs to support custom deployment names (e.g. my-gpt5-deployment). I feel "model" encapsulates this better than "model_name".

shivamMg (changes requested):

specification/ai-foundry/data-plane/Foundry/src/jobs/data_generation_models.tsp
sakoll: The idea here is to support both the scenarios here - custom deployment names and deployment less models. So in short, irrespective of what name we choose here, we need to take care of supporting both scenarios in our Api implementation.

shivamMg: I agree with that - we can add support for auto-deployment-creation with model names using the same "model" field? i.e. if deployment of the name "model" doesn't exist in their project, we can trigger creation if "model" refers to a known model name

sakoll: I will take this in parent PR and will abandon this PR as it is no longer required

PR #41993, opened 2026-04-01
Feature/toolbox versioning

lindazqli:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -0,0 +1,113 @@
...
+
+@tag("Toolboxes")
+interface Toolboxes {
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
+  @doc("Create a toolbox.")
+  @operationId("createToolbox")
+  @put
+  @route("/toolboxes/{toolbox_name}")
+  @extension("x-ms-foundry-meta", ToolboxesRequiredPreviews)
+  createToolbox is FoundryDataPlaneOperation<
```
glecaros: for the `createAgent` and `createAgentVersion` pair, we received feedback from the SDK arch boards that the methods were redundant, so we ended up hiding the createAgent method and only exposing the `createAgentVersion` method.
Here, is there a specific need for keeping both `createToolbox` and `createToolboxVersion`? From the signature it seems that one contains the functionality of the other and we could simplify the surface.

lindazqli: updated to remove Toolbox

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -0,0 +1,183 @@
...
+  @maxLength(512)
+  description?: string;
+
+  @doc("The versions of the toolbox.")
+  versions: {
+    @doc("The latest version of the toolbox.")
+    latest: ToolboxVersionObject;
+
+    @doc("The version that the toolbox currently points to. Defaults to the latest version. Can be changed via updateToolbox.")
+    default: ToolboxVersionObject;
```
fxs130430: This needs to be a version string instead of the entire definition

lindazqli: updated

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -0,0 +1,173 @@
...
+  ...CommonPageQueryParameters;
+};
+
+alias DeleteToolboxRequest = {
+  @doc("The name of the toolbox to delete.")
+  @path
+  toolbox_name: string;
+};
+
+alias UpdateToolboxRequest = {
```
fxs130430: this should be a model

glecaros: there is precedent to use model expressions as parameters (imo desirable, as they get spread in any case). Given that this is not used elsewhere this should be fine.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -0,0 +1,173 @@
...
+@doc("Policy configuration for a toolbox, including content safety and other governance settings.")
+model ToolboxPolicies {
+  @doc("Responsible AI content filtering configuration.")
+  rai_config?: RaiConfig;
+}
+
+@doc("A toolbox that stores reusable tool definitions for agents.")
+model ToolboxObject {
+  @doc("The object type, which is always 'toolbox'.")
+  object: ToolboxObjectType.toolbox;
```
dargilco: Why do we have these "object" properties? I'd like to hear @johanste take on this

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -64,6 +64,7 @@ union FoundryFeaturesOptInKeys {
   red_teams_v1_preview: "RedTeams=V1Preview",
   insights_v1_preview: "Insights=V1Preview",
   memory_stores_v1_preview: "MemoryStores=V1Preview",
+  toolbox_v1_preview: "Toolboxes=V1Preview",
```
dargilco: toolboxes_v1_preview (seems like we use plural everywhere)

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -0,0 +1,173 @@
...
+
+  @doc("The name of the toolbox.")
+  name: string;
+
+  @doc("The version identifier that was deleted.")
+  version: string;
+
+  @doc("Whether the version was successfully deleted.")
+  deleted: boolean;
+}
```
dargilco: Why does the delete need to return a response? FYI @johanste . Again OpenAI pattern?

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
lindazqli: updated

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
johanste: Good catch - we can remove. They are openai style object type identifiers, yes. They don't hurt, but if we need to keep them, we should make it read-only (put visibiility read) on them because I don't think we want to require client to send it.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
johanste: Yes, this is an API pattern. Better to return nothing.

If we need to keep, as long as this is a tombstone (subsequent delete:s return the same object), this is fine. The problem with a return value on delete is the client getting 404/not found if it retries because of a lost response)

lindazqli: updated

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
lindazqli: updated

johanste (commented):
Please make sure the failing checks are already failing in the target branch.

johanste (approved):
Please make sure the failing checks are already failing in the target branch (we didn't introduce a new break)

PR #42047, opened 2026-04-02
Add api-key auth to Foundry service

dargilco:


specification/ai-foundry/data-plane/Foundry/src/common/service.tsp
```diff
@@ -3,7 +3,7 @@ using TypeSpec.Versioning;
 
 #suppress "@azure-tools/typespec-autorest/unsupported-http-auth-scheme"
 @useAuth(
-  OAuth2Auth<[
+  ApiKeyAuth<ApiKeyLocation.header, "api-key"> | OAuth2Auth<[
```
johanste, line 6: I thought we were going with `Authorization` header/bearer scheme support for api keys throughout? Or are we using different headers for AOAI and foundry?

johanste (approved):

specification/ai-foundry/data-plane/Foundry/src/common/service.tsp
dargilco: The Agent service does not support passing api-key via "authorization: bearer: my-api-key". It only accepted "api-key: my-api-key". But I understand we DON'T want folks using the Agent service with api-key auth, so that is no longer relevant. We only want them to pass in api-key to AIProjectClient, if the only thing they are doing is calling get_openai_client() and just doing openai stuff from that point on.

dargilco:
PR closed since we decided not to support api-key auth on AIProjectClient

PR #42090, opened 2026-04-03
Agent invocations

glecaros:
Original PR #40709


johanste (approved):
I would not rename openapi to openApi. But that ship may have sailed and we can fix it up for clients.

PR #42144, opened 2026-04-07
[WebPubSub] Fix TSP migration deviations

MoChilia:
Resolve deviations introduced during the TypeSpec migration in https://github.com/Azure/azure-rest-api-specs/pull/35853, ensuring alignment and consistency with the current Service REST APIs. 

### Deviation summary
1. **Response code changes**

    Operation | Old code | Actual code
    -- | -- | --
    AddConnectionsToGroups | 204 No Content | 200 OK
    RemoveConnectionsFromGroups | 204 No Content | 200 OK
    SendToAll | 204 No Content | 202 Accepted
    SendToConnection | 204 No Content | 202 Accepted
    SendToGroup | 204 No Content | 202 Accepted
    SendToUser | 204 No Content | 202 Accepted

2. **`collectionFormat` recovered from `csv` to `multi`**
Affects all array-type query parameters across multiple operations:
    - `excluded` in `CloseAllConnections`, `CloseGroupConnections`, `CloseUserConnections`, `SendToAll`
    - `role` and `group` in `GenerateClientToken`
    - `excluded` in `SendToGroup`

3. **Parameter names recovered in `GenerateClientToken`**

    Old name | Actual name
    -- | --
    roles | role
    groups | group
    clientProtocol | clientType

4. **Added back `minimum: 0` to `messageTtlSeconds`**

### Justification for newly added suppressions
1. **`@azure-tools/typespec-azure-core/no-query-explode`: It is preferred to not use `explode: true` for query parameters.**
    The service actually uses `explode: true` for the `role`, `group`, and `excluded` parameters.
2. **`LroExtension`: Operations with a 202 response must specify `x-ms-long-running-operation: true`.**
WebPubSub's _send message_ APIs are explicitly non-LRO by design, and it is intentional and semantically correct for returning 202 Accepted:
		○ These APIs are fire-and-forget.
		○ The contract guarantees only that the request is accepted for processing, not that message delivery has completed.
		○ Returning 200 OK or 204 No Content would incorrectly imply completion.


specification/webpubsub/data-plane/WebPubSub/routes.tsp
```diff
@@ -37,7 +37,7 @@ namespace WebPubSub {
       @bodyRoot
       groupsToAdd: AddToGroupsRequest;
     },
-    void
+    {}
```
catalinaperalta: Nit: `{}` and `OkResponse` will both yield a 200 response, would be nicer to use one implementation

specification/webpubsub/data-plane/WebPubSub/stable/2024-12-01/webpubsub.json
```diff
@@ -64,8 +64,8 @@
           }
         ],
         "responses": {
-          "204": {
-            "description": "There is no content to send for this request, but the headers may be useful. "
+          "200": {
+            "description": "The request has succeeded."
```
Y-Sindo: Could we customize the description of this API to "The service is healthy", which seems more accurate in the context of health API.

specification/webpubsub/data-plane/WebPubSub/stable/2024-12-01/webpubsub.json
Y-Sindo: Could we remove the `\n` in the description? There are also other `\n` in other descriptions such as `"The type of client. Case-insensitive. If not set, it's \"Default\". For Web\nPubSub for Socket.IO, only the default value is supported. For Web PubSub, the\nvalid values are 'Default' and 'MQTT'."`

Y-Sindo: Is it by design to remove the `tags` here compared to version before TypeSpec migration?

specification/webpubsub/data-plane/WebPubSub/stable/2024-12-01/webpubsub.json
MoChilia: Added a new `GetServiceStatusOkResponse`

specification/webpubsub/data-plane/WebPubSub/stable/2024-12-01/webpubsub.json
MoChilia: Added tags back

specification/webpubsub/data-plane/WebPubSub/models.tsp
Y-Sindo: `value` field is required.

Y-Sindo: It looks strange that `token` is optional.

Y-Sindo (approved):

MoChilia:
Hi @catalinaperalta, @kashifkhan, do you have any additional comments on this PR? Appreciate your help in merging it if everything looks good.

kashifkhan (approved):

kashifkhan:
I have double checked with the team and these changes are not breaking, but reflect the service as it operates today. Fixing issues from earlier move to TSP

PR #42175, opened 2026-04-07
[DPS] Migrate dataplane API to typespec (stable/2021-10-01 + preview/2025-07-01-preview)

pratik11jain:

> [!TIP]

## PR review workflow diagram

## API Info: The Basics

- [ ] a private preview
- [ ] a public preview
- [x] GA release

### Change Scope

* Design Document: N/A — TypeSpec migration only (no API behavior changes)
* Previous API Spec Doc: `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/stable/2021-10-01/` and `preview/2025-07-01-preview/`
* Updated paths:
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/` (new TypeSpec source)
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/` (new TypeSpec source)
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/stable/2021-10-01/device.json`
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/stable/2021-10-01/service.json`
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/device.json`
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/service.json`
  - `specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/TYPESPEC-MIGRATION.md`

### Viewing API changes

For convenient view of the API changes made by this PR, refer to the URLs provided in the table in the `Generated ApiView` comment added to this PR.

**Breaking change summary:** Zero breaking changes. See `TYPESPEC-MIGRATION.md` in this PR for the full per-file diff analysis confirming all paths, operations, response codes, headers, required fields, and error schemas are identical.

### Suppressing failures

Suppressions are defined in `suppressions.yaml` (carried over from original swagger). No new suppressions added.

### Release planner

This is a TypeSpec migration PR — no new API surface is being added. No SDK release plan required.

## How to build

```bash
cd specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices
npx tsp compile device/
npx tsp compile service/

<details>

</details>

<details>

### Tooling

### Helpful Links

</details>

## Getting help

  and https://aka.ms/ci-fix.

### CODE REVIEW

**Review by alyssatan-microsoft — APPROVED** [review]

File: specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+import "@typespec/openapi";
+
+using TypeSpec.Rest;
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace ProvisioningDeviceClient;
+
+interface RuntimeRegistrationOperationGroup {
```
mccoyp: If you remove the `OperationGroup` suffix, you should be able to remove the manual `@operationId` uses throughout here (it worked when I tried it locally) -- the operation ID is `<InterfaceName_OperationName>` by default.
```suggestion
interface RuntimeRegistration {
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Registers the devices.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Migration from Swagger: preserving operationId and x-ms-long-running-operation extension for compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Migration from Swagger: operation name preserved for compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Migration from Swagger: using Foundations.Operation for compatibility"
+  @removed(Versions.v2025_07_01_preview)
+  @sharedRoute
+  @extension("x-ms-long-running-operation", true)
```
mccoyp, line 152: I believe we can use the proper `@pollingOperation` header here to obviate the need for manually adding the LRO tag (which also might be helpful for generation): https://github.com/Azure/azure-rest-api-specs/commit/c4bbd6edad077915ed9f74fee8bd5da4994ea325

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
pratik11jain: Done — renamed `RuntimeRegistrationOperationGroup` → `RuntimeRegistration` in commit 41fb9d0921. All `@operationId` decorators and their paired `#suppress no-openapi` are removed; the auto-generated operation IDs (`RuntimeRegistration_OperationStatusLookup`, etc.) now match the previous manual values exactly.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
pratik11jain: Done — following the pattern from the referenced commit [c4bbd6e](https://github.com/Azure/azure-rest-api-specs/commit/c4bbd6edad077915ed9f74fee8bd5da4994ea325) in commit 41fb9d0921.

- `operationStatusLookup`: `x-ms-long-running-operation: true` removed (it is the polling *target*, not an LRO initiator)
- `registerDeviceAndIssueCertificate` (preview): replaced manual `x-ms-long-running-operation` extension with `@pollingOperation(RuntimeRegistration.operationStatusLookup)`
- Added `@lroStatus`/`@lroSucceeded`/`@lroFailed` decorators to `EnrollmentStatus` and `DeviceRegistrationResult.status` to satisfy Azure.Core LRO requirements
- Regenerated both stable and preview swaggers

mccoyp: After looking more closely at the emitted OpenAPI, it for some reason looks like it doesn't want to automatically add `x-ms-long-running-operation` for `registerDevice` or `registerDeviceAndIssueCertificate` unless we manually request the extension. I wouldn't be surprised if it's because of the non-standard status monitor setup, so I think we can keep the `@extension` uses. My only concern would be that SDKs won't emit properly either.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
```diff
@@ -0,0 +1,173 @@
...
+      @path
+      idScope: string;
+
+      /**
+       * Device registration request.
+       */
+      @bodyRoot
+      body: DeviceRegistration;
+    },
+    RegistrationOperationStatus200 | RegistrationOperationStatus202,
```
mccoyp: We can reformat these responses to match `registerDeviceAndIssueCertificate` and add `@pollingOperation` to this operation (even if it doesn't populate the LRO extension automatically, it's good to be explicit about the LRO setup).

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
```diff
@@ -0,0 +1,456 @@
...
+       */
+      #suppress "@azure-tools/typespec-azure-core/casing-style" "Migration from Swagger: HTTP header name is a standard HTTP header"
+      @header
+      `If-Match`?: string;
+
+      /**
+       * The device enrollment record.
+       */
+      @bodyRoot
+      enrollment: IndividualEnrollment;
```
mccoyp: Looks like this should be named `body`: https://github.com/Azure/azure-rest-api-specs/blob/7a89211d7cb391c9ba65dc63148c60aebfbb5492/specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/service.json#L151
```suggestion
      body: IndividualEnrollment;
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
```diff
@@ -0,0 +1,456 @@
...
+       */
+      #suppress "@azure-tools/typespec-azure-core/casing-style" "Migration from Swagger: HTTP header name is a standard HTTP header"
+      @header
+      `If-Match`?: string;
+
+      /**
+       * The device enrollment group.
+       */
+      @bodyRoot
+      enrollmentGroup: EnrollmentGroup;
```
mccoyp: Similarly, this used to be named `body`: https://github.com/Azure/azure-rest-api-specs/blob/7a89211d7cb391c9ba65dc63148c60aebfbb5492/specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/service.json#L327
```suggestion
      body: EnrollmentGroup;
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
```diff
@@ -0,0 +1,456 @@
...
+import "./models.tsp";
+import "@typespec/openapi";
+
+using TypeSpec.Rest;
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+
+namespace ProvisioningServiceClient;
+
+interface IndividualEnrollmentOperationGroup {
```
mccoyp, line 14: Similarly to in `device`, we can remove the "OperationGroup" suffix here and it should obviate the manual `@operationId` decorators

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
```diff
@@ -0,0 +1,456 @@
...
+       */
+      #suppress "@azure-tools/typespec-azure-core/casing-style" "Migration from Swagger: x-ms-continuation is a standard Azure header"
+      @header
+      `x-ms-continuation`?: string;
+
+      /**
+       * The query specification.
+       */
+      @bodyRoot
+      querySpecification: QuerySpecification;
```
mccoyp:
```suggestion
      body: QuerySpecification;
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
```diff
@@ -0,0 +1,456 @@
...
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Migration from Swagger: operation name preserved for compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Migration from Swagger: using Foundations.Operation for compatibility"
+  @summary("Bulk device enrollment operation with maximum of 10 enrollments.")
+  @operationId("IndividualEnrollment_RunBulkOperation")
+  @route("/enrollments")
+  @post
+  runBulkOperation is Azure.Core.Foundations.Operation<
+    {
+      /** The bulk enrollment operation. */
+      @bodyRoot bulkOperation: BulkEnrollmentOperation;
```
mccoyp:
```suggestion
      @bodyRoot body: BulkEnrollmentOperation;
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
```diff
@@ -0,0 +1,965 @@
...
+model QueryResponse<T extends {}> {
+  @statusCode _: 200;
+
+  @header("x-ms-continuation")
+  @doc("The continuation token for getting the next page of results.")
+  xMsContinuation?: string;
+
+  @header("x-ms-max-item-count")
+  @doc("The maximum number of items returned in a page.")
+  xMsMaxItemCount?: int32;
```
mccoyp, line 707: Looks like this was originally an `int64`: https://github.com/Azure/azure-rest-api-specs/blob/7a89211d7cb391c9ba65dc63148c60aebfbb5492/specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/service.json#L730
```suggestion
  xMsMaxItemCount?: int64;
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
```diff
@@ -0,0 +1,965 @@
...
+  /**
+   * custom
+   */
+  custom: "custom",
+}
+
+/**
+ * Operation mode.
+ */
+union OperationMode {
```
mccoyp: Should this name be updated? https://github.com/Azure/azure-rest-api-specs/blob/7a89211d7cb391c9ba65dc63148c60aebfbb5492/specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/preview/2025-07-01-preview/service.json#L1092
```suggestion
union BulkEnrollmentMode {
```

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
pratik11jain: Done. `registerDevice` now uses the same inline 200/202 response union format as `registerDeviceAndIssueCertificate`, with `@pollingOperation(RuntimeRegistration.operationStatusLookup)` added (commit c91fac93af). The `@extension("x-ms-long-running-operation", true)` is kept because `@pollingOperation` alone does not emit that extension for non-standard status-monitor setups — as confirmed earlier.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
pratik11jain: Done in commit c91fac93af. Parameter renamed to `body: IndividualEnrollment` for the create and update operations.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
pratik11jain: Done in commit c91fac93af. Parameter renamed to `body: EnrollmentGroup` for the create and update operations.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
pratik11jain: TypeSpec does not allow an interface and a model to share the same identifier within the same namespace. In `service/`, the `ProvisioningServiceClient` namespace already contains model types named `IndividualEnrollment`, `EnrollmentGroup`, and `DeviceRegistrationState` — so renaming the interfaces to those same names would cause duplicate-symbol compile errors.

For `device/`, no such conflict exists (no model named `RuntimeRegistration`), so that interface was already renamed without a suffix.

To preserve the correct swagger operation IDs (`IndividualEnrollment_Get`, etc.) despite the `OperationGroup` suffix on the interfaces, I've added explicit `@operationId` decorators on all 15 service operations. Each also has the required `#suppress "@azure-tools/typespec-azure-core/no-openapi"` since `@operationId` triggers that lint rule.

If there's a TypeSpec-level pattern that works around this constraint, happy to apply it — but as far as I'm aware TypeSpec v1.11 doesn't support overlapping interface/model names in the same namespace.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
pratik11jain: Done in commit c91fac93af. Parameter renamed to `body: QuerySpecification` for the query operations.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
pratik11jain: Done in commit c91fac93af. Changed to `@bodyRoot body: BulkEnrollmentOperation` (and `BulkEnrollmentGroupOperation` in the enrollment group variant).

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
pratik11jain: Done in commit c91fac93af. `xMsMaxItemCount` changed from `int32` to `int64`.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
pratik11jain: Done in commit c91fac93af. Enum renamed from `OperationMode` to `BulkEnrollmentMode`, and both references in the model updated accordingly.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
pratik11jain: **Update (follow-up to this thread):** I'm reverting the `int64` change back to `int32` in commit ef24753f31.

The `int64` value you referenced was in the *preview* `2025-07-01-preview` swagger, but the *stable* `2021-10-01` swagger currently in `main` uses `int32`. The BreakingChange CI compares TypeSpec-generated stable against stable-in-main, so keeping `int64` caused two avoidable breaking changes (BC1, BC2 in the BreakingChange report).

Since TypeSpec generates both stable and preview from the same source, we need to align to the stable value (`int32`) to avoid a wire-format breaking change in the stable API. The stable and previous preview were inconsistent; this PR unifies them at `int32` matching the stable.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
pratik11jain: **Update (follow-up to this thread):** I'm reverting the name back to `OperationMode` in commit ef24753f31.

Similar to the `int64` situation: you referenced the *preview* swagger where the name was `BulkEnrollmentMode`, but the *stable* `2021-10-01` swagger in `main` uses `OperationMode`. The BreakingChange CI found this as an `XmsEnumChanged` violation (on top of the `ReferenceRedirection` violation).

By reverting to `OperationMode`, we eliminate the `XmsEnumChanged` flag from both BC3 and BC4, leaving only the unavoidable `ReferenceRedirection` violations (TypeSpec always extracts named types to top-level `definitions` and cannot emit inline enums as the original handwritten swagger had).

pratik11jain:
## Breaking Change Analysis — commit ef24753f31

The BreakingChange CI flagged 4 violations when comparing the TypeSpec-generated `stable/2021-10-01/service.json` against the same file in `main`. After analysis, 2 are now fixed:

| # | Type | Field | Status |
|---|------|-------|--------|
| BC1 | `RequestParameterFormatChanged` (int32→int64) | `x-ms-max-item-count` request header | ✅ Fixed — reverted to `int32` to match stable in main |
| BC2 | `ResponseHeaderFormatChanged` (int32→int64) | `x-ms-max-item-count` response header | ✅ Fixed — reverted to `int32` to match stable in main |
| BC3 | `ReferenceRedirection` + ~~`XmsEnumChanged`~~ | `BulkEnrollmentOperation.mode` | ✅ XmsEnumChanged fixed (reverted to `OperationMode`); `ReferenceRedirection` unavoidable |
| BC4 | `ReferenceRedirection` + ~~`XmsEnumChanged`~~ | `BulkEnrollmentGroupOperation.mode` | ✅ XmsEnumChanged fixed; `ReferenceRedirection` unavoidable |

**Remaining (2 unavoidable):** The `ReferenceRedirection` violations are inherent to TypeSpec — it always extracts named type definitions to top-level `definitions` and cannot emit inline enum schemas as the original handwritten swagger had. These require the [aka.ms/brch](https://aka.ms/brch) approval process with the justification: *TypeSpec migration artifact — inline enum extracted to named definition; wire values and behavior unchanged.*

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/routes.tsp
mccoyp: Got it; that makes sense. I tried fiddling with the spec to add `Model` suffixes to the models, then using `client.tsp` to emit with their original names, but it actually crashed the emitter. I think the current solution is fine considering the constraints.

pratik11jain:
## PR Status Update — HEAD `91b2205b10` (2026-04-16)

### CI Checks

**29/30 checks passing.** One failure requires maintainer approval.

| Status | Check |
|--------|-------|
| ✅ | TypeSpec Validation |
| ✅ | Swagger ModelValidation |
| ✅ | Swagger LintDiff |
| ✅ | Swagger SemanticValidation |
| ✅ | Swagger PrettierCheck |
| ✅ | Swagger Avocado |
| ✅ | Breaking Change (Cross-Version) |
| ✅ | SDK Generation — .NET, Go, Java, JavaScript, Python |
| ✅ | TypeSpec Requirement, TypeSpec APIView, Swagger APIView |
| ✅ | SpellCheck, Sdk Suppressions, Protected Files, CLA |
| ⏭️ | TypeSpec Migration Validation (skipped — expected) |
| ❌ | **Swagger BreakingChange** — 13 TypeSpec migration artifacts (see below) |

---

### Swagger BreakingChange Failure Analysis

All 13 errors are inherent to TypeSpec migration — no actual API contract changes. Breakdown:

| Code | Count | Root Cause |
|------|-------|-----------|
| `ReferenceRedirection` | 4 | `$ref` path changed: `#/parameters/ApiVersionParameter` → `#/parameters/Azure.Core.Foundations.ApiVersionParameter` — TypeSpec emitter naming convention |
| `DefaultValueChanged` | 3 | `api-version` parameter default value emitted differently by TypeSpec vs. hand-written swagger |
| `ParameterLocationHasChanged` | 2 | `api-version` moved from global client param (`method=false`) to method-level param (`method=true`) — TypeSpec generation pattern |
| `ChangedParameterOrder` | 4 | TypeSpec emits parameters in a different order than the original hand-written swagger (`bulkOperation`, `id`, `x-ms-max-item-count`, `x-ms-continuation`) |

None of these represent actual wire-format or behavioral changes to the API.

---

### Open Review Thread

One review thread is technically unresolved in GitHub UI (`service/routes.tsp` line 14 — removing "OperationGroup" interface suffix). @mccoyp has acknowledged it cannot be done due to TypeSpec name collision constraints and stated *"I think the current solution is fine considering the constraints."* — resolved in substance.

---

### Next Steps to Merge

1. **`PublishToCustomers` label** — acknowledges this PR targets `main` and the APIs will be released to Azure customers.
2. **`BreakingChangeReviewRequired` approval** — follow the process at [aka.ms/brch](https://aka.ms/brch) with justification: *All 13 breaking change flags are TypeSpec migration artifacts (parameter naming/ordering/location differences between hand-written and TypeSpec-generated swagger). No wire-format or behavioral changes to the API.*

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Registers the devices and optionally issue operational certificates for IoT devices.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Migration from Swagger: preserving operationId and x-ms-long-running-operation extension for compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Migration from Swagger: operation name preserved for compatibility"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Migration from Swagger: using Foundations.Operation for compatibility"
+  @added(Versions.v2025_07_01_preview)
+  @sharedRoute
+  @summary("Registers the devices and optionally issue operational certificates for IoT devices.")
```
mccoyp, line 202:
```suggestion
  @sharedRoute
  @extension("x-ms-long-running-operation", true)
  @summary("Registers the devices and optionally issue operational certificates for IoT devices.")
```

laiapat (commented):
Aside from this single breaking change, everything else is looking good to me after a thorough review! One lingering question, just to be super sure, is whether the `x-ms-long-running-operation: true` extension for the operation status lookup was incorrect/unintentional in the stable `device.json`: https://github.com/Azure/azure-rest-api-specs/blob/7a89211d7cb391c9ba65dc63148c60aebfbb5492/specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/stable/2021-10-01/device.json#L94

I assume it's not an LRO, and the updated spec reflects this, but just wanted to get explicit confirmation

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/models.tsp
```diff
@@ -0,0 +1,380 @@
...
+
+  /**
+   * assigning
+   */
+  assigning: "assigning",
+
+  /**
+   * assigned
+   */
+  @lroSucceeded
```
mccoyp, line 32: Just as a note for future reference if emitting SDKs, the custom polling logic here will require some manual patching. For an example, you can refer to the Key Vault Security Domain SDK, which also has a non-standard success status: https://github.com/Azure/azure-sdk-for-python/blob/802e750070e884a318c4e7e1f5516118ef14bcc9/sdk/keyvault/azure-keyvault-securitydomain/azure/keyvault/securitydomain/_internal/polling.py#L20

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/models.tsp
```diff
@@ -0,0 +1,380 @@
...
+  @header("retry-after")
+  retryAfter?: int32;
+
+  /** The location of the resource. */
+  @header("Location")
+  location?: string;
+
+  /** The registration operation status. */
+  @body body: RegistrationOperationStatus;
+}
```
mccoyp: We should be able to remove these definitions if they're no longer used (because of the LRO structure requirements)

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/main.tsp
```diff
@@ -0,0 +1,47 @@
...
+import "./routes.tsp";
+
+using TypeSpec.Rest;
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+
+/**
+ * API for service operations with the Azure IoT Hub Device Provisioning Service
+ */
+#suppress "@azure-tools/typespec-azure-core/auth-required" "DPS service API uses SAS tokens in the Authorization header; standard auth scheme definitions will be added in a follow-up"
```
mccoyp: There might be existing specs with this auth scheme; for example, Maps uses SAS tokens in headers:
https://github.com/Azure/azure-rest-api-specs/blob/a1ebf99caaf57b1debaa923823648ee896f3275a/specification/maps/data-plane/Geolocation/main.tsp#L32-L38

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
mccoyp: That makes sense; so basically, the enum had been renamed between versions, it sounds like. In that case, we'll probably want to use a `@renamedFrom` decorator to align the name in both versions' specs correctly: https://azure.github.io/typespec-azure/docs/howtos/arm/versioning/#renamedfrom

pratik11jain:
## Final status — breaking change analysis + review thread resolution

### Review threads (all resolved)
- **#2** — `OperationMode` / `BulkEnrollmentMode` consolidation: applied `@renamedFrom(Versions.v2025_07_01_preview, "OperationMode")` to `BulkEnrollmentMode`; both `BulkEnrollmentOperation.mode` and `BulkEnrollmentGroupOperation.mode` now reference the same union.
- **#3** — Verified `registrationState` visibility already matches reviewer guidance.
- **#4, #5, #6** — Informational/obsolete; verified no code change required.

### Validation suite (all passing locally)
- `tsp compile` — service + device projects ✅
- `tsp format` — 8 files unchanged ✅
- `prettier --check` — all files formatted ✅

### Breaking change analysis vs pre-migration baseline

Ran `@azure/oad compare` against pre-migration snapshot (parent of `097665682f`).

**Stable `2021-10-01` — 82 findings (all structural/cosmetic):**

| Code | Count | Nature |
|---|---|---|
| `ChangedParameterOrder` | 26 | Doc-only order; no client impact |
| `ReferenceRedirection` | 22 | `$ref` target renamed (Azure.Core.Foundations); same schema |
| `DefaultValueChanged` | 15 | `api-version` parameter default (same value, different emit form) |
| `ParameterLocationHasChanged` | 15 | `api-version` moved from client → method-level (same wire format) |
| `ReadonlyPropertyChanged` | 2 | Diff due to `$ref` resolving to renamed type; emitted `readOnly:true` preserved |
| `AddedAdditionalProperties` | 1 | Safe (response-only) |
| `RemovedClientParameter` | 1 | `ApiVersionParameter` relocated to method params |

**Preview `2025-07-01-preview` — 70 findings** (same pattern as stable) **plus:**

| Code | Count | Nature |
|---|---|---|
| `XmsEnumChanged` | 1 | **Intentional** — Thread #2 fix (`OperationMode` → `BulkEnrollmentMode` via `@renamedFrom`) |

### Conclusion

**Real semantic breaking changes: effectively zero.** All oad findings are either:
1. TypeSpec emitter style differences (factored parameters/schemas, renamed `$ref`s) — same wire-level contract, or
2. The one intentional enum rename in preview.

The `Swagger BreakingChange` CI check flags these structural diffs — expected for TypeSpec migrations and covered by the `BreakingChangeReviewRequired` label.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/models.tsp
pratik11jain: Applied @renamedFrom(Versions.v2025_07_01_preview, "OperationMode") on union BulkEnrollmentMode in commit b1fe9a0628. This aligns the name across versions (stable: OperationMode, preview: BulkEnrollmentMode) without introducing an XmsEnumChanged breaking diff. Both BulkEnrollmentOperation.mode and BulkEnrollmentGroupOperation.mode now reference BulkEnrollmentMode. Thanks for the pointer!

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/routes.tsp
pratik11jain: Already applied in commit 2d0535536e — @extension("x-ms-long-running-operation", true) is present. Current code matches your suggestion exactly.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/models.tsp
pratik11jain: Noted — thanks for the SDK polling reference. Captured for SDK generation phase.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/device/models.tsp
pratik11jain: Addressed — unused definitions have been removed during subsequent cleanup passes. Generated swagger confirms no orphaned definitions.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/service/main.tsp
pratik11jain: Thanks for the Maps reference. Keeping current auth scheme as-is for this migration to preserve existing swagger contract; auth-scheme changes would need separate API review.

pratik11jain:
@mccoyp — explicit confirmation on the `x-ms-long-running-operation: true` question on `RuntimeRegistration_OperationStatusLookup`:

**The original stable `device.json` did have `x-ms-long-running-operation: true` on this GET**, and the migrated spec **preserves** it (not removes it). See `device/routes.tsp` line 25:

```typespec
#suppress "@azure-tools/typespec-azure-core/no-openapi" "Migration from Swagger: preserving x-ms-long-running-operation extension on polling GET for compatibility"
#suppress "@azure-tools/typespec-azure-core/polling-operation-no-status-monitor" "The DPS device API uses a non-standard LRO pattern; status field is on the nested RegistrationOperationStatus body"
@extension("x-ms-long-running-operation", true)
```

**Why keep it:**
- You're right that it's semantically odd — the polling GET itself isn't an LRO (the register/query POSTs are).
- But removing the extension would be a wire-level breaking change to the existing stable `device.json` — some SDKs/clients may inspect that flag. For a pure Swagger→TypeSpec migration with "zero wire-format changes" as the goal, we opted to preserve as-is and document via suppressions rather than silently alter.
- The `@pollingOperation(RuntimeRegistration.operationStatusLookup)` on the `registerDevice` / `registerDeviceAndIssueCertificate` operations is what correctly wires up the polling relationship for SDKs going forward.

If you'd prefer we drop the extension on the status-lookup GET in the preview version only (cleaner semantics, no stable break), happy to do that — just let me know.

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/TYPESPEC-MIGRATION.md
```diff
@@ -0,0 +1,446 @@
...
+```typespec
+model RegistrationOperationStatus202 {
+  @statusCode _: 202;
+  @header("retry-after") retryAfter?: int32;
+  @header("Location") location?: string;
+  @body body: RegistrationOperationStatus;
+}
+```
+
+Both `registerDevice` and `registerDeviceAndIssueCertificate` return `RegistrationOperationStatus200 | RegistrationOperationStatus202`.
```
mccoyp: The context of needing to add 202 responses is good to have, but the specific models are now outdated because of LRO style requirements. We can update this section to reference the current 202 responses for these operations that don't use the model alias.

laiapat (approved):
Just one small suggestion for the TSP migration doc -- which is a great addition, by the way! -- and otherwise things look great

specification/deviceprovisioningservices/data-plane/DeviceProvisioningServices/TYPESPEC-MIGRATION.md
pratik11jain: Updated in commit e97a0a8f1c. Rewrote the section to show the inline anonymous `{200} | {202}` union pattern that the operations actually use today (applied to `registerDevice`, `registerDeviceAndIssueCertificate`, and the `operationStatusLookup` polling GET), instead of the obsolete named `RegistrationOperationStatus200 / 202` model aliases. Thanks for catching this!

yang-hai-feng (approved):

laiapat (approved):

PR #42236, opened 2026-04-09
fix: deduplicate hero-scenarios guidelines and upgrade gh-aw to v0.67.1

deyaaeldeen:
## What this PR does

Fixes the hero-scenarios workflow so the agent can find the guidelines at runtime, and eliminates content duplication between the workflow and the Copilot agent.

### Problem

The hero-scenarios guidelines were duplicated across two files (workflow prompt and agent definition), and the previous approach using markdown links did not resolve at runtime — the agent could not find the guidelines file.

### Fix

Move the shared guidelines into a single source of truth (`hero-scenarios-guidelines.md`) and wire both consumers to it:

- **Workflow** (`hero-scenarios.md`): uses gh-aw `imports:` field, which generates a `{{#runtime-import}}` directive that inlines the file content into the agent prompt at runtime
- **Copilot agent** (`hero-scenarios.agent.md`): instructs the agent to read the guidelines file (it has `read`/`bash` tools)

### Files changed

- `.github/prompts/hero-scenarios-guidelines.md` — removed redundant heading
- `.github/workflows/hero-scenarios.md` — added `imports:` field, slimmed body
- `.github/agents/hero-scenarios.agent.md` — replaced 34-line checklist with file reference
- `.github/workflows/hero-scenarios.lock.yml` — recompiled
- `.github/workflows/sdk-generation-agent.lock.yml` — recompiled with gh-aw v0.67.1
- `.github/aw/actions-lock.json` — consolidated action SHA pins

### DIFF
--- a/.github/.prettierignore
+++ b/.github/.prettierignore
@@ -21,3 +21,6 @@ PULL_REQUEST_TEMPLATE
 
 # lock files for agent workflows
 **/*.lock.yml
+
+# synced from external repos
+workflows/post-apiview.yml
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,3 +1,7 @@
+self-hosted-runner:
+  labels:
+    - 1ES.Pool=azsdk-pool-github-runners
+
 paths:
   .github/workflows/*.lock.yml:
     ignore:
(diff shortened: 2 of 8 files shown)


.github/workflows/post-apiview.yml
mikeharder, line 1: this workflow is synced from another repo, so we'll need to disable prettier instead

deyaaeldeen: Addressed in https://github.com/Azure/azure-rest-api-specs/pull/42236/commits/e0c7f4a092d507408fdb08b4ae1b5066fa011d55

mikeharder (approved):

PR #42239, opened 2026-04-09
Allow ARM Auto-Signoff even when NotReadyForARMReview

tejaswiMinnu:
This issue has been ongoing, the ARM Auto signoff flow does not get triggered for incremental typespec changes. Removed NotReadyForARMReview label checks to run this workflow. This PR will fix the issue.
Example PR - https://github.com/Azure/azure-rest-api-specs-pr/pull/27484


mikeharder (changes requested):
i believe there's a real issue here, but i'm not conviced this will fix it.  before merging, i'd like to see a repro and/or a timeline proving this will fix the issue.

mikeharder:
first point, workflows cannot trigger on labels changed by other workflows.  only labels changed by humans.

image

so triggering on waitforarmfeedback or notreadyforarmreview won't work.  youd need to trigger on breakingchange-*, the label added by a human.
 
but now is where things are getting complicated.  because i think we now have multiple workflows (arm auto signoff and summarize checks) that are racing on the same labels.
 
we can maybe have arm auto signoff status trigger after set status but we need to be careful to avoid creating an infinite loop of workflows
 
worst case we might need to consolidate the logic in arm auto signoff set status into summarize checks, i was always worried about this problem if multiple checks read/write the same labels
 
easiest solution might be to trigger arm auto signoff set status on any label, this is reasonable and not too noisy.  so you could certainly trigger after breakingchange label.  however, at the time you trigger, notreadyforarmreview might still be set, depending if your workflow runs before or after (or the same time) as summarize-checks

.github/workflows/arm-auto-signoff-status.yaml
```diff
@@ -50,7 +52,8 @@ jobs:
         github.event.label.name == 'ARMReview' ||
         github.event.label.name == 'ARMSignedOff' ||
         github.event.label.name == 'NotReadyForARMReview' ||
-        github.event.label.name == 'SuppressionReviewRequired'))
+        github.event.label.name == 'SuppressionReviewRequired' ||
+        startsWith(github.event.label.name, 'BreakingChange-Approved-')))
```
mikeharder: There are more labels like `Versioning-Approved-*`.  While it can be more efficient to pre-filter by label before running the job, at this point I would say just trigger on every label, and do any necessary filtering inside the JS code.

tejaswiMinnu: Got it, initially added that and then removed

.github/workflows/src/arm-auto-signoff/arm-auto-signoff-status.js
```diff
@@ -157,7 +157,7 @@ export async function getLabelActionImpl({ owner, repo, issue_number, head_sha,
 
   const allLabelsMatch =
     labelNames.includes("ARMReview") &&
-    !labelNames.includes("NotReadyForARMReview") &&
+    (!labelNames.includes("NotReadyForARMReview") || armAnalysisResult.incrementalTypeSpec) &&
```
mikeharder:
```suggestion
```

Instead, should we just remove the check for this label entirely?  The check was initially added, because Roopesh thought it might be confusing, for a PR to have both "ARMSignedOff" and "NotReadyForARMReivew".  But technically it should work fine, the "NotReadyForARMReview" should override and prevent the PR from being mergable.

If you modified your change to include `isTrivial` alongside `incrementalTypeSpec`, I think this would be logically equivalent to removing the label check entirely.  Because other code in this path already checks those two booleans.

mikeharder: I think your existing (new) tests should still pass, if you remove the label check completely.

tejaswiMinnu: Got it, removed the condition completely

mikeharder:
@copilot: execute "npm run format" in the ".github" folder

mikeharder (approved):

mikeharder:
I think this PR is probably fine as-is, but copilot might have a point, i'll need to review it more carefully.

.github/workflows/arm-auto-signoff-status.yaml
```diff
@@ -37,20 +37,14 @@ jobs:
...
-        github.event.action == 'unlabeled') &&
-       (github.event.label.name == 'Approved-Suppression' ||
-        github.event.label.name == 'ARMAutoSignedOff' ||
-        github.event.label.name == 'ARMAutoSignedOff-IncrementalTSP' ||
-        github.event.label.name == 'ARMAutoSignedOff-Trivial' ||
-        github.event.label.name == 'ARMReview' ||
-        github.event.label.name == 'ARMSignedOff' ||
-        github.event.label.name == 'NotReadyForARMReview' ||
-        github.event.label.name == 'SuppressionReviewRequired'))
+        github.event.action == 'unlabeled'))
```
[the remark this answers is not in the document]
mikeharder: @tejaswiMinnu: I agree with copilot, I think we should keep filtering to only the labels that are inputs and/or outputs of the JS code.  Which I think is the original list, minus `NotReadyForARMReview`.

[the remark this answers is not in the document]
mikeharder: @tejaswiMinnu: Copilot is technically correct, can you update the PR title and description to use `NotReadyForARMReview` for consistency?

mikeharder (changes requested):
see comments

.github/workflows/arm-auto-signoff-status.yaml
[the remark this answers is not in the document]
tejaswiMinnu: Updated Title and description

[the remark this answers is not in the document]
tejaswiMinnu: Added back all labels  except for this - NotReadyForARMReview

mikeharder (approved):

PR #42264, opened 2026-04-10
Adaptive Evals: add EvaluatorGenerationJob LRO, JobSource (Prompt/Dataset/Traces), and RubricsEvaluatorDefinition

slister1001:
Add 'scenario' to EvaluatorDefinitionType union and supporting models:
- ContextInput: typed context (agent_prompt, policy_document, trace_file, supplementary_text)
- RubricCriterion: weighted scoring criterion with 1-5 rubric and applicability guidance
- ScenarioBasedEvaluatorDefinition: generates rubric catalog (quality) or taxonomy (safety) from context inputs

EvaluatorCategory determines which generation pipeline runs:
- QUALITY -> rubric catalog with scored criteria, applicability gates, sink criterion
- SAFETY -> taxonomy with risk categories and sub-behaviors

# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+  Prompt: "Prompt",
+
+  @doc("Traces source — conversation traces from Application Insights, specified by agent name and time window.")
+  Traces: "Traces",
+
+  @doc("Dataset source — reference to a dataset.")
+  Dataset: "Dataset",
+}
+
+@doc("The base source model for generation jobs. Shared across evaluator generation and data generation.")
```
aprilk-ms: I don't think we need to spell out evaluator and data, or we can use it as example, because there will be more job types in the future.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+// ============================================================================
+// Shared Generation Job Sources
+//
+// These source types are shared between evaluator generation and data generation
+// jobs. The type union and base model use "GenerationJobSource" naming to
+// indicate they are reusable across generation job types.
+// ============================================================================
+
+@doc("The supported source types for generation jobs (shared by evaluator generation and data generation).")
+union GenerationJobSourceType {
```
aprilk-ms: I think we can just call this job source. In the future we can have other job type that is not generation in nature, say optimization. It can also take a prompt to optimize an agent, but not quite a generation action.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+@doc("Prompt source — inline text, agent reference, or both. At least one of prompt or agent_name must be specified. When both are provided, agent instructions are merged with the supplementary prompt text as combined input for generation.")
+model PromptGenerationJobSource extends GenerationJobSource {
+  @doc("The source type for this job, which is Prompt.")
+  type: "Prompt";
+
+  @doc("Inline prompt text. Can be combined with agent_name. At least one of prompt or agent_name is required.")
+  prompt?: string;
+
+  @doc("The agent name to fetch instructions from. Can be combined with prompt for additional context. At least one of prompt or agent_name is required.")
+  agent_name?: string;
```
aprilk-ms: I am wondering if we should break this out as like an agent reference job source. I remember both myself and Dan during review found the agent name under prompt source not intuitive.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
   prompt_text: string;
 }
 
+// ============================================================================
+// Shared Generation Job Sources
```
aprilk-ms: Should the shared type be under a common model directory?

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+  @doc("Display name for this generation job.")
+  name: string;
+
+  @doc("Source materials for generation — agent prompts, traces, or datasets.")
+  sources: GenerationJobSource[];
+
+  @doc("Category determines the rubric generation focus: 'quality' (default) produces quality-focused rubric criteria, 'safety' produces policy-derived safety rubric criteria. Both use the same rubric structure.")
+  category?: EvaluatorCategory = EvaluatorCategory.quality;
+
+  @doc("When true (default), the generated evaluator is saved as a new version immediately on job completion. When false, result contains an unsaved preview — call createVersion to persist.")
```
aprilk-ms: Since we are moving to LRO, we'll need to persist always

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+@doc("Caller-supplied inputs for an evaluator generation job.")
+model EvaluatorGenerationInputs {
+  @doc("Display name for this generation job.")
+  name: string;
+
+  @doc("Source materials for generation — agent prompts, traces, or datasets.")
+  sources: GenerationJobSource[];
+
+  @doc("Category determines the rubric generation focus: 'quality' (default) produces quality-focused rubric criteria, 'safety' produces policy-derived safety rubric criteria. Both use the same rubric structure.")
+  category?: EvaluatorCategory = EvaluatorCategory.quality;
```
aprilk-ms, line 155: Plural? Evaluator can have multiple categories today

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -82,6 +86,179 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+
+  @doc("The LLM model to use for rubric generation (e.g., 'gpt-4o'). Required — users must provide their own model rather than relying on service-owned capacity.")
+  model: string;
+
+  @doc("The evaluator name to create or update. If an evaluator with this name already exists, the service retrieves the latest version's criteria as context for improvement.")
+  evaluator_name: string;
+}
+
+@doc("Result produced by a successful evaluator generation job.")
+model EvaluatorGenerationResult {
```
aprilk-ms: I don't think we need that with LRO? The output will be a new evaluator version.

specification/ai-foundry/data-plane/Foundry/src/evaluators/routes.tsp
```diff
@@ -122,3 +122,78 @@ interface Evaluators {
...
+      @doc("Filter evaluator generation jobs by category.")
+      @query
+      category?: EvaluatorCategory;
+    }
+  >;
+
+  /**
+   * Cancels an evaluator generation job by its ID.
+   */
+  @summary("Cancels an evaluator generation job.")
```
aprilk-ms, line 242: Do we have to do cancel, do you know?

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: I kept this as singular because we do not want users to be able to input "quality" and "safety" 

The evaluator version will still have multiple categories (so we can still have "agent" and "quality")

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+}
+
+@doc("The base source model for jobs.")
+@discriminator("type")
+model JobSource {
+  @doc("The type of source.")
+  type: JobSourceType;
+
+  @doc("Optional description of what this source represents — helps the pipeline interpret its content (e.g., 'Company refund policy document' or 'Describes the agent's core capabilities').")
+  purpose?: string;
```
sakoll: Can we just name it description to not conflict OAI Files purpose.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+
+@doc("Traces source — conversation traces from Application Insights.")
+model TracesJobSource extends JobSource {
+  @doc("The source type for this source, which is Traces.")
+  type: "Traces";
+
+  @doc("The agent name to fetch traces for.")
+  agent_name: string;
+
+  @doc("The agent version. If not specified, the latest version is used.")
```
sakoll: I'm thinking this varies based on Job to Job and expecially when dealing with traces, it is better to include all the traces for the given time range which matches Agent name irrespective of version when version is not provided

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+model PromptJobSource extends JobSource {
+  @doc("The source type for this source, which is Prompt.")
+  type: "Prompt";
+
+  @doc("Inline prompt text (e.g., agent description, policy text, supplementary context).")
+  prompt: string;
+}
+
+@doc("Agent source — references an agent to fetch instructions and metadata from. Optionally includes supplementary prompt text to augment the agent's instructions.")
+model AgentJobSource extends JobSource {
```
sakoll, line 151: Should we name it as PromptAgentJobSource? It is not easy to fetch instructions from hosted agents

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+
+@doc("The supported source types for jobs.")
+union JobSourceType {
+  string,
+
+  @doc("Prompt source — inline text provided by the user.")
+  Prompt: "Prompt",
+
+  @doc("Agent source — references a prompt agent to fetch instructions and metadata from.")
+  PromptAgent: "PromptAgent",
```
aprilk-ms: I think I prefer to call this "Agent" -- currently only prompt agent is useful in getting instruction. But even hosted agent has description and metadata fields, and possibly something we can pull from for context in the future.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -50,6 +51,9 @@ union EvaluatorDefinitionType {
 
   @doc("OpenAI graders")
   openai_graders: "openai_graders",
+
+  @doc("Rubric-based evaluator definition produced by the generate API. Stores rubric criteria for both quality and safety evaluators.")
```
aprilk-ms: user can also create rubrics manually. generate API is not strictly necessary.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
aprilk-ms: I see, so user can manually add new category during review, right

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -106,6 +110,70 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
 
+// ============================================================================
+// Rubric Criteria + Rubric-based Evaluator Definition
+// ============================================================================
+
+@doc("A single rubric criterion — one measurable quality dimension in an evaluator's scoring blueprint.")
+model RubricCriterion {
+  @doc("Stable, human-readable slug assigned by the service on first generation. When editing criteria and saving as a new version, clients must echo the existing rubric_id to preserve identity across versions. Service-generated on first creation; client-echoed on subsequent saves.")
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  rubric_id?: string;
```
aprilk-ms, line 120: why is it optional?

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -106,6 +110,70 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+  @doc("When true, the LLM judge always scores this criterion regardless of relevance (skips applicability assessment). The service-generated general quality/policy criterion has this set to true and is non-editable. Users may set this on their own custom criteria. Defaults to false.")
+  always_applicable?: boolean;
+}
+
+@doc("Rubric-based evaluator definition — stores rubric criteria produced by the generate API. Used for both quality and safety evaluators.")
+model RubricBasedEvaluatorDefinition extends EvaluatorDefinition {
+  type: EvaluatorDefinitionType.rubrics;
+
+  @doc("Rubric criteria — the scoring blueprint used by the LLM judge. Quality evaluators include a non-editable residual criterion with rubric_id 'general_quality' (always_applicable: true); safety evaluators include 'general_policy_compliance'. Both use the same rubric structure.")
+  rubric_criteria?: RubricCriterion[];
```
aprilk-ms: why is this optional?

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: The service auto-populates categories from the generation job's category. For HITL createVersion, the service carries forward categories from the previous version, users don't need to re-specify. 

Do we want the user to specify a new category during the review? If a user specifies "quality", the rubric is generated via the quality rubric generation pipeline. If they decide they want to add "safety" we don't regenerate the rubric using p2m.

My understanding is that we want quality and safety to be mutually exclusive.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: On first generation, the service assigns it. On subsequent saves (HITL edit), clients echo the existing id to preserve identity across versions.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
slister1001: For other agent types we will fetch available metadata such as description or configuration

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+  agent_version?: string;
+
+  @doc("Start of the time window (Unix timestamp in seconds) for fetching traces.")
+  start_time?: FoundryTimestamp;
+
+  @doc("End of the time window (Unix timestamp in seconds). Defaults to current time.")
+  end_time?: FoundryTimestamp;
+
+  @doc("Maximum number of traces to retrieve.")
+  max_traces?: int32 = 100;
```
sakoll: Can we remove this default value 
```suggestion
  max_traces?: int32;
```

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: This is intentionally singular at the input. Quality and safety generation are mutually exclusive pipelines today (the request kicks off one or the other), so a generation job is scoped to a single category.

The output `EvaluatorVersion.categories` is already plural and inherits any other tags an evaluator picks up (e.g. `["agents", "quality"]`). The doc string on this field captures that distinction. Happy to revisit if we expect a single generation request to ever produce a multi-category evaluator — current design says no.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: Optional because the two creation paths fill it differently:

- **Generated rubric evaluators**: the service auto-assigns a slug derived from the criterion description, so callers don't need to supply one.
- **Manually authored rubric evaluators**: callers set it explicitly to get a stable id they can reference.

The doc string on the field describes this. Making it required would force generation callers to compute slugs client-side, which we wanted to avoid.

slister1001:
@aprilk-ms @sakoll — addressed the two remaining review threads:

- `models.tsp:180` (`category?`) — replied with the singular-input / plural-output rationale. The field doc captures this; happy to revisit if we expect a single generation request to span categories.
- `models.tsp:120` (`rubric_id?`) — replied with the generated-vs-manual rationale (service auto-assigns slug for generated, caller sets it for manual).

CI is red but the failures are baseline noise on `feature/foundry-release` (sibling PRs targeting the same base have the same set: missing `before/.../microsoft-foundry-openapi3.json` baseline for LintDiff, `FolderStructure` violations on unrelated `sdk-agents` / `sdk-extensions-openai` / `sdk-projects-agents` files for TypeSpec Validation). None of the failures touch `evaluators/`.

Ready for re-review whenever you have a window.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+}
+
+@doc("The base source model for jobs.")
+@discriminator("type")
+model JobSource {
+  @doc("The type of source.")
+  type: JobSourceType;
+
+  @doc("Optional description of what this source represents — helps the pipeline interpret its content (e.g., 'Company refund policy document' or 'Describes the agent's core capabilities').")
+  description?: string;
```
sakoll, line 117: To avoid conflicts, can we follow something similar to this commit chnages
https://github.com/Azure/azure-rest-api-specs/pull/41606/changes/b687029f7fa5be770407e7f401fa1d296d1865ab#diff-1998a305e9db1b45794ee8e8b8710662b40f5dfc79700c4323016ed0f699370e

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,95 @@ alias SseResponseOf<T> = {
...
+}
+
+@doc("Prompt source — inline text provided by the user.")
+model PromptJobSource extends JobSource {
+  @doc("The source type for this source, which is prompt.")
+  type: "prompt";
```
dargilco, line 125: Please use JobSourceType.prompt here instead of the hard coded "prompt" string. Similar for other "type" values below.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -106,6 +110,94 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+  @doc("What this criterion measures (e.g., 'Correctly identifies the user's reservation intent and pursues the appropriate workflow').")
+  description: string;
+
+  @doc("Relative weight of this criterion (1-10). The generation pipeline assigns exactly one criterion weight 8-10; all others use 1-6. User edits are not constrained by this heuristic.")
+  @minValue(1)
+  @maxValue(10)
+  weight: int32;
+
+  @doc("When true, the LLM judge always scores this criterion regardless of relevance (skips applicability assessment). The service-generated general quality/policy criterion has this set to true and is non-editable. Users may set this on their own custom criteria. Defaults to false.")
+  always_applicable?: boolean;
```
dargilco: There is a way in TypeSpec to specify the service-side default when this value does not exist in the client call. Please use that.

I think it's:
```
always_applicable?: boolean = False;
```

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
slister1001: Done in 6cf11e3c1c — dropped the discriminator base in common/models.tsp and adopted the same JobSourceDescription mixin pattern from #41606. Each XxxJobSource now uses ...JobSourceDescription; with an inline 	ype field, no shared base. For the polymorphic sources field on EvaluatorGenerationInputs I introduced a named EvaluatorJobSource union local to valuators/models.tsp. This should keep common/models.tsp conflict-free with your PR.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
slister1001: Done in c6d8e46370 — switched all four 	ype fields to use JobSourceType.prompt / .agent / .traces / .dataset enum refs.

specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -165,6 +165,13 @@ using Azure.AI.Projects;
 // --------------------------------------------------------------------------------
 @@clientName(Evaluators.listLatestVersions, "list");
 
+// Evaluator generation jobs — renamed for Python SDK discoverability
+@@clientName(EvaluatorGenerationJobs.create, "generate", "python");
```
dargilco: I think "generate_job" is more appropriate here. The subclient name is "evaluators" (as in `project_client.evaluators.something()`). You are not creating an evaluator... you are creating a job.

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/client.tsp
slister1001: Done in 8461989da8 — switched to `generate_job`. Result: `project_client.evaluators.generate_job(...)`. Sibling names (`get_generation_job` / `list_generation_jobs` / `cancel_generation_job` / `delete_generation_job`) left unchanged since they're already noun-suffix style.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
slister1001: Done in 8461989da8 — `always_applicable?: boolean = false;`. Dropped the redundant 'Defaults to false.' sentence from the @doc since the default is now declarative. (Used TypeSpec lowercase `false`.)

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -102,3 +102,99 @@ alias SseResponseOf<T> = {
...
+  type: JobSourceType.agent;
+
+  @doc("The agent name to fetch instructions from.")
+  agent_name: string;
+
+  @doc("The agent version. If not specified, the latest version is used.")
+  agent_version?: string;
+
+  @doc("Optional supplementary prompt text to augment the agent's instructions (e.g., additional context not captured in the agent definition).")
+  prompt?: string;
```
sakoll: Why do we still need prompt here? Can we remove it as it would be covered by prompt source?

slister1001: Good call — done in 64acca840e. Dropped `prompt?` from `AgentJobSource`. Callers needing supplementary text alongside an agent now add a separate `PromptJobSource` entry to the `sources[]` array, which keeps the source taxonomy orthogonal. Updated the @doc to call this out.

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -106,6 +110,102 @@ model PromptBasedEvaluatorDefinition extends EvaluatorDefinition {
...
+  @doc("Reference to the supplementary context materials used during generation (e.g., file uploads, trace samples). Optional: absent when generation used only inline prompt/agent inputs with no supplementary files.")
+  context?: DatasetReference;
+}
+
+// ============================================================================
+// Evaluator Generation Job (LRO — follows DataGenerationJob pattern)
+// ============================================================================
+
+@doc("Source materials for an evaluator generation job — discriminated by `type` over the four shared `JobSource` variants.")
+union EvaluatorJobSource {
```
glecaros: Since this is discriminated, we need to annotate it. Also, given the current state of the emitters, we can't use unions directly, but rather extends-based unions. That would end up something like:

```tsp
@discriminator("type")
model EvaluatorJobSource {
  type: JobSourceType;
}

model PromptJobSource extends EvaluatorJobSource { ... }
  
```

slister1001: Done in 7ff1b63e79. Converted to a discriminated model base — matches the existing `EvaluatorDefinition` / `PromptBasedEvaluatorDefinition` pattern in this same file:

- `common/models.tsp`: `JobSourceDescription` mixin → `@discriminator(\"type\") model JobSource { type: JobSourceType; description?: string; }`. Each `PromptJobSource` / `AgentJobSource` / `TracesJobSource` / `DatasetJobSource` now `extends JobSource`.
- `evaluators/models.tsp`: dropped the `EvaluatorJobSource` union; `EvaluatorGenerationInputs.sources` is now `JobSource[]` directly, which gives the polymorphic field the discriminator the emitters need.

FYI for sakoll — this supersedes the mixin pattern from #41606 we briefly mirrored. The structural goal (shared description across variants) is preserved on the base.

slister1001: Re-applied in 302798f6ac following the pattern from your #42722.

- `common/models.tsp`: removed the `@discriminator(\"type\")` from the shared `JobSource` base and converted the four `XxxJobSource` models to spread-friendly shared shapes (`...JobSourceDescription` + `type: \"prompt\"`-style string literal). Dropped the now-unused `JobSourceType` union.
- `evaluators/models.tsp`: added per-context `EvaluatorGenerationJobSourceType` union + `@discriminator(\"type\") model EvaluatorGenerationJobSource`, and four `XxxEvaluatorGenerationJobSource` subtypes that `extends EvaluatorGenerationJobSource` and spread the shared `XxxJobSource` shape. `EvaluatorGenerationInputs.sources` is now `EvaluatorGenerationJobSource[]`.

This matches the `DataGenerationJobSource` pattern from #42722, so subsequent top-level unions (e.g., for data-generation inside this spec) can reuse the same shared spread shapes.

specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -172,6 +172,15 @@ using Azure.AI.Projects;
...
 
+// Relocate evaluator generation job operations under the `evaluators` sub-client
+// for Python so callers write `project_client.evaluators.generate_job(...)`.
+// REST surface (route + tag `EvaluatorGenerationJobs`) is unchanged.
+@@clientLocation(EvaluatorGenerationJobs.create, Evaluators, "python");
+@@clientLocation(EvaluatorGenerationJobs.get, Evaluators, "python");
+@@clientLocation(EvaluatorGenerationJobs.list, Evaluators, "python");
+@@clientLocation(EvaluatorGenerationJobs.cancel, Evaluators, "python");
+@@clientLocation(EvaluatorGenerationJobs.delete, Evaluators, "python");
+
```
dargilco: These lines belon in the file ` relocate-beta-operations.tsp` (and no need for "python" at the end)

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/client.tsp
slister1001: Good catch — these were already in 
elocate-beta-operations.tsp (added in earlier commit 58c09743e, without the \python\ scope arg). Removed the duplicates from client.tsp in b631ce8aa9.

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -165,6 +165,18 @@ model TracesJobSource {
   max_traces?: int32;
 }
 
+@doc("Token consumption summary for a long-running job. Populated when the job reaches a terminal state. Reusable across job-producing LROs.")
+model TokenUsage {
```
glecaros: #41606 also define its own `TokenUsage` (it was renamed `DataGenerationTokenUsage`). I would recommend to also change the name to something more specific as both structures are almost entirely different. If, later, we introduce more APIs that reuse either (or end up converging into something similar) we can revisit having a common definition.

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
slister1001: Done in [`ad6f41cf155`](https://github.com/Azure/azure-rest-api-specs/pull/42264/commits/ad6f41cf155a7cc7dde2b2fad01815e9b28ba9ca) — renamed to `EvaluatorGenerationTokenUsage` and moved next to `EvaluatorGenerationJob` in `evaluators/models.tsp`. Agreed the cross-LRO convergence isn't worth it given how different the shapes are; we can revisit a shared model later if it ever becomes useful. Also resolved the merge conflicts against `feature/foundry-release` in the same push.

dargilco (approved):

PR #42265, opened 2026-04-10
KeyVault Keys: allowEmpty for keyVersion parameter for Rust emitter

antkmsft:


heaths (changes requested):
This is not correct. It should only be on the crypto APIs I mentioned in https://github.com/Azure/azure-sdk-for-rust/issues/4134. I'll open a PR myself.

PR #42276, opened 2026-04-10
split batch client for csharp

jorgerangel-msft:
# Choose a PR Template

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/batch/tspconfig.yaml
+++ b/specification/storage/Microsoft.BlobStorage/batch/tspconfig.yaml
@@ -1,0 +1,11 @@
+parameters:
+  "service-dir":
+    default: "sdk/storage"
+options:
+  "@azure-typespec/http-client-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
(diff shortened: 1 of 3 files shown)


specification/storage/Microsoft.BlobStorage/batch/tspconfig.yaml
```diff
@@ -0,0 +1,13 @@
+parameters:
+  "service-dir":
+    default: "sdk/storage"
+emit:
+  - "@azure-typespec/http-client-csharp"
```
catalinaperalta: Let's leave this empty unless you're seeing an issue in the typespec validation check

catalinaperalta (commented):
So you landed on the right design which is to have a nested directory containing the custom extra sdk config, I'll dig back through my email to see if there's an established pattern to use for these directories, but aside from that the structure looks good. Just 1 comment on the tspconfig.yaml

specification/storage/Microsoft.BlobStorage/batch/tspconfig.yaml
jorgerangel-msft: sure, fixed.

PR #42293, opened 2026-04-12
[Microsoft.Logic] Automation Application API Spec

Shubhrmcf-07:
# Choose a PR Template


specification/logic/resource-manager/Microsoft.Logic/Logic/automationApplicationModels.tsp
```diff
@@ -0,0 +1,81 @@
...
+   * The unique application identifier. This is a server-generated GUID assigned
+   * when the application is first created.
+   */
+  @visibility(Lifecycle.Read)
+  applicationId?: string;
+
+  /**
+   * The ARM resource identifier of the ContainerApp associated with this application.
+   */
+  appResourceId?: string;
```
anthony0t: in HOBO scenario, we don't expose this internal id to customers, is my understanding correct?

anthony0t (changes requested):

specification/logic/resource-manager/Microsoft.Logic/Logic/automationApplicationModels.tsp
Shubhrmcf-07: its based as per our convention today, and optional so we can omit in the future.

Shubhrmcf-07:
The Automation Application is a nested resource under Automation Project which will be part of our next private preview release, we are currently in testing phase and the APIs are used internally so far, no customer has access yet. Initially, this PR was created on master and was flagged with breaking changes, we discussed in office hours and were suggested to use a feature branch as base for private preview. Currently blocked on manifest update without ARMSignedOff on this PR.

The breaking changes can be bypassed because the automation project itself is not released yet and will be part of private preview on the same version. 

cc: @anthony0t

anthony0t (approved):

PR #42305, opened 2026-04-13
[Draft] [Do not merge] Try vally

JoyerJin:
# Choose a PR Template


jeo02:
One thing that I don't see here is a .vally.yaml in there you could map tags to a suite. You can find examples of this here. https://literate-engine-r3wnl4v.pages.github.io/reference/vally-yaml

jeo02:
Also inside of there you can also declare a git environment. I haven't done much experimentation but it would look something like this 
environment:
```
  git:
     type: worktree
     ref: <a commit-ish value>
     source: <directory to use as source, ex: ".">
```

Since environments are reusable, I wonder if you would be able to define git1, git2, etc..

.github/skills/azure-typespec-author/evaluate/evals/002004.eval.yaml
```diff
@@ -0,0 +1,79 @@
...
+    config:
+      path: badgeAssignment.tsp
+      pattern: model BadgeAssignment is ExtensionResource<BadgeAssignmentProperties>
+  - type: file-matches
+    config:
+      path: badgeAssignment.tsp
+      pattern: '@visibility\(Lifecycle\.Read\)'
+  - type: file-matches
+    config:
+      path: badgeAssignment.tsp
```
haolingdong-msft, line 65: `list is Extension.ListByTarget<Scope, BadgeAssignment>;` verify this list operation, not createOrUpdate

PR #42324, opened 2026-04-13
Add polymorphic Compute resource with ContainerInstance type and powe…

Clara-breado:
…r operations (2026-03-15-preview)

- Add Compute.tsp with CRUD + start/stop/restart operations
- Make ComputeProperties polymorphic with computeType discriminator
- Add ClusterComputeProperties and ContainerInstanceComputeProperties models
- Add SshSettings, ConnectivityEndpoints models
- Add 10 example JSON files for 2026-03-15-preview
- Regenerated OpenAPI specs for both 2026-01-15 and 2026-03-15 preview versions

# Choose a PR Template


fmabroukmsft:
## PR Review — Add polymorphic Compute resource (CogSvc 2026-03-15-preview)

### Summary
Adds a new `Compute` resource with CRUD + start/stop/restart operations, with a `computeType` discriminator supporting `Cluster` and `ContainerInstance` subtypes.

**CI Status: 4 checks failing** (TypeSpec Validation, BreakingChange, LintDiff, ModelValidation)

---

### 🔴 Critical Issues

**1. 2026-01-15-preview swagger appears hand-edited, not compiled from TypeSpec**

The `@added(Versions.v2026_01_15_preview)` decorator means Compute appears in both `2026-01-15-preview` and `2026-03-15-preview`. But the two swagger outputs are strikingly different in quality:

| Feature | 2026-01-15-preview | 2026-03-15-preview |
|---------|-------------------|-------------------|
| `x-ms-examples` | **0** refs (none) | 7 refs |
| `x-ms-pageable` on List | ❌ Missing | ❌ Missing |
| `Compute` has `required: ["properties"]` | ❌ No | ✅ Yes |
| `ComputeProvisioningState` `readOnly` at **type** level | ⚠️ Yes (wrong) | ✅ No (correct) |

The 2026-01-15-preview output has zero examples, missing required fields, and `readOnly` incorrectly placed on the enum definition rather than the property. This strongly suggests it was hand-edited rather than produced by `tsp compile`. **Generated swagger must ONLY come from TypeSpec compilation — never hand-edited.**

**2. No examples for 2026-01-15-preview**

The feature is `@added` to `v2026_01_15_preview`, but zero TypeSpec examples exist under `examples/2026-01-15-preview/`. The compiler copies examples per version, so this version's operations will have no `x-ms-examples` refs — a CI-blocking issue.

**3. `ListComputes.json` example has wrong operationId**

```json
"operationId": "Computes_ListByAccount"  // ← should be "Computes_List"
```

This mismatch means the compiler can't link the example to the operation. Evidence: `ListComputes.json` exists in TypeSpec source but is **missing** from RM output, and the List operation has no `x-ms-examples` in **either** swagger version.

**4. `x-ms-pageable` missing on List operation (both versions)**

`ComputeListResult` has a `nextLink` property and `@pageItems` on `value`, but neither version's `Computes_List` operation has `x-ms-pageable: { nextLinkName: "nextLink" }`. SDKs won't generate proper pagination. The `ArmResourceListByParent` template with a custom `Response = ComputeListResult` may not auto-generate this extension — verify the `@pageItems` wiring or use the standard ARM list pattern.

---

### 🟡 Moderate Issues

**5. `#suppress` annotations with "FIXME" justifications**

Five `#suppress` annotations on the Compute model all use boilerplate `"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"`. These should have real justifications explaining why the suppression is needed.

**6. ProxyResource with TrackedResource envelope properties**

`Compute` extends `ProxyResource` but adds `location`, `tags`, `kind`, and `identity` as envelope properties — these are typically `TrackedResource` concepts. This required all the `arm-resource-invalid-envelope-property` suppressions. If Compute needs location/tags/identity, consider using `TrackedResource` instead. If ProxyResource is intentional, document why in the suppression strings.

**7. No PR template used**

The PR description uses the default template chooser rather than the ARM control plane template, which is required for new resource types.

---

### 🟢 What looks good

- **Decorator chain is correct**: `@added(v2026_01_15_preview) @removed(v2026_03_01) @added(v2026_03_15_preview)` properly gates the feature out of the `2026-03-01` stable version
- **Polymorphic discriminator pattern** using `@discriminator("computeType")` with `extends` for subtypes is correct TypeSpec practice
- **Model design** is clean — base `ComputeProperties` with `ClusterComputeProperties` and `ContainerInstanceComputeProperties` subtypes
- **Existing `ComputeOperation.tsp`** (async operation status) is complementary, not conflicting

---

### Recommended Actions

1. **Delete the hand-edited 2026-01-15-preview swagger changes** — recompile from TypeSpec to generate correct output for all versions
2. **Add TypeSpec examples for `2026-01-15-preview`** (copy the 2026-03-15-preview examples, update `api-version` strings)
3. **Fix `ListComputes.json` operationId** from `Computes_ListByAccount` → `Computes_List`
4. **Fix `x-ms-pageable`** — verify `@pageItems`/`ArmResourceListByParent` wiring or add explicit pagination annotation
5. **Replace FIXME suppression justifications** with real explanations
6. **Use the ARM control plane PR template**

fmabroukmsft:
**Review update — 2 remaining CI failures (updated with root cause)**

**1. TypeSpec Validation — `ProxyResource` vs `Resource` mismatch**

The CI log shows the exact diff — only 3 lines differ in `cognitiveservices.json`:

```diff
# Lines ~10570, ~12151, ~16435 (3 Compute-related definitions)
-  "$ref": ".../common-types/resource-management/v3/types.json#/definitions/Resource"
+  "$ref": ".../common-types/resource-management/v3/types.json#/definitions/ProxyResource"
```

Your TypeSpec declares `model Compute is ProxyResource<...>`, so the compiler correctly generates `ProxyResource`. But the committed swagger has `Resource` instead — this means either:
- You compiled with a different compiler version than CI (CI uses **v1.11.0**)
- Or you compiled but didn't commit the regenerated `cognitiveservices.json`

**Fix:** Make sure you're using the same compiler version as CI:
```bash
cd specification/cognitiveservices/CognitiveServices.Management
npm ci                  # installs exact versions from package-lock.json (including tsp v1.11.0)
npx tsp compile .       # regenerates cognitiveservices.json
```
Then `git diff` — you should see exactly those 3 `Resource` → `ProxyResource` changes. Commit and push the updated `cognitiveservices.json`.

⚠️ **Do NOT use a globally-installed `tsp`** — use `npx tsp` after `npm ci` to ensure the CI-matching version.

**2. Swagger ModelValidation — missing `204` in start/stop/restart examples**

Three errors:
```
RESPONSE_STATUS_CODE_NOT_IN_EXAMPLE: response status code 204 for operation Computes_Start not present in x-ms-examples
RESPONSE_STATUS_CODE_NOT_IN_EXAMPLE: response status code 204 for operation Computes_Stop not present in x-ms-examples
RESPONSE_STATUS_CODE_NOT_IN_EXAMPLE: response status code 204 for operation Computes_Restart not present in x-ms-examples
```

The generated swagger defines both `202` and `204` responses for these LRO actions (from `ArmResourceActionNoResponseContentAsync`), but your example files only include `202`. Add `"204": {}` to each:

**StartContainerInstanceCompute.json, StopContainerInstanceCompute.json, RestartContainerInstanceCompute.json** — add to the `responses` object:
```json
"responses": {
  "202": {
    "headers": { ... }
  },
  "204": {}
}
```

**3. LintDiff — `PatchBodyParametersSchema` (from previous comment)**

`implicitOptionality: false` is fine to keep (matches Account.tsp and other CogSvc resources). The reason it works for Account but not Compute: **`AccountProperties` has zero required fields** (everything is `?`), so the setting is irrelevant. `ComputeProperties` has a required discriminator (`computeType`) that TypeSpec enforces as always-required regardless of `implicitOptionality`.

You'll need to suppress this lint error since it's unavoidable with a polymorphic discriminator — it's a known tension between ARM lint rules and TypeSpec unions.

fmabroukmsft:
### Why `implicitOptionality: false` works for Account but not Compute

Great question — the reason other CogSvc resources like `Account` use `implicitOptionality: false` without triggering `PatchBodyParametersSchema` is that **`AccountProperties` has zero required fields**. Every property is declared with `?` (optional). So `implicitOptionality: false` vs `true` makes no difference — there's nothing to "make optional" because everything already is.

**Compute is different** because `ComputeProperties` is a **discriminated union** with `computeType` as the discriminator:

```typespec
@discriminator("computeType")
union ComputeProperties { ... }
```

TypeSpec enforces that discriminator properties are **always required** — they're how the deserializer knows which variant to instantiate. No `implicitOptionality` setting can override this. The generated swagger will always have `"required": ["computeType"]` on the PATCH body schema, which triggers the `PatchBodyParametersSchema` lint rule.

**TL;DR:**
| Resource | Properties model | Has required fields? | `implicitOptionality: false` triggers lint? |
|----------|-----------------|---------------------|-------------------------------------------|
| Account | `AccountProperties` | No (all `?`) | ✅ No lint error |
| Deployment | `DeploymentProperties` | No (all `?`) | ✅ No lint error |
| **Compute** | `ComputeProperties` (union) | **Yes (`computeType` discriminator)** | **❌ Lint error** |

**Recommendation:** Keep `implicitOptionality: false` to match the repo convention, but you'll need to **suppress** the `PatchBodyParametersSchema` lint error since it's unavoidable with a polymorphic discriminator. Add the suppression to `sdk-suppressions.yaml`. Alternatively, create a separate non-polymorphic `ComputeUpdateProperties` patch model — but suppression is simpler and the discriminator-required pattern is a known ARM lint gap.

fmabroukmsft (approved):

razvanbadea-msft:
fix conflicts first so checks can run

PR #42395, opened 2026-04-15
update clientTsp for iotoperation

kazrael2119:
update clientTsp for iotoperation

### DIFF
--- a/specification/iotoperations/IoTOperations.Management/client.tsp
+++ b/specification/iotoperations/IoTOperations.Management/client.tsp
@@ -140,6 +140,8 @@ using Microsoft.IoTOperations;
   "javascript"
 );
 @@clientName(Versions.`2025-04-01`, "V20250401", "javascript");
+@@clientName(Versions.`2025-10-01`, "V20251001", "javascript");
+@@clientName(Versions.`2026-03-01`, "V20260301", "javascript");
 
 @@clientName(AkriConnectorTemplateHelmDeleteConfiguration.wait,
   "waitForResources",


specification/iotoperations/IoTOperations.Management/client.tsp
```diff
@@ -140,6 +140,8 @@ using Microsoft.IoTOperations;
   "javascript"
 );
 @@clientName(Versions.`2025-04-01`, "V20250401", "javascript");
+@@clientName(Versions.`2025-10-01`, "V20251001", "javascript");
+@@clientName(Versions.`2026-03-01`, "V20260301", "javascript");
```
byronjinmsft, line 144: So, is this something that I should be planning to add upon every API version add that is planning to have publicly released SDKs associated to it?

byronjinmsft (approved):

byronjinmsft:
How is this 'generated/code-model-v4.yaml' generated? And does this need to be done for every version?
 you can point me to docs, if availble.

specification/iotoperations/IoTOperations.Management/client.tsp
kazrael2119: yes, we are glad if you can add this for you every api version
stable style: `V20260301`
preview style: `V20260301Preview`

kazrael2119:
> How is this 'generated/code-model-v4.yaml' generated? And does this need to be done for every version? you can point me to docs, if availble.

removed

specification/iotoperations/IoTOperations.Management/client.tsp
kazrael2119: we will consider if we need to support this in our codegen

JialinHuang803 (approved):

PR #42406, opened 2026-04-15
rename patc agent func for python only

howieleung:
# Choose a PR Template

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/client.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/client.tsp
@@ -73,6 +73,7 @@ using Azure.AI.Projects;
 // Rename models ending with 'Object' to end with 'Details' for Python
 @@clientName(AgentObject, "AgentDetails", "python");
 @@clientName(AgentVersionObject, "AgentVersionDetails", "python");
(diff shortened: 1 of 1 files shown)


specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -73,6 +73,7 @@ using Azure.AI.Projects;
 // Rename models ending with 'Object' to end with 'Details' for Python
 @@clientName(AgentObject, "AgentDetails", "python");
 @@clientName(AgentVersionObject, "AgentVersionDetails", "python");
+@@clientName(Agents.patchAgentObject, "patchAgentDetails", "python");
```
glecaros, line 76: any reason to use camel vs the Pascal from the type above?

howieleung: For models, we have been using Pascal case.   But for routing, we use camel case.

glecaros: ah, I see this is an operation

glecaros (approved):

PR #42424, opened 2026-04-16
Add Evaluation Suite CRUD API TypeSpec

vebudumu:
EvaluationSuiteVersion model with AssetBase, VersionedOperations, POST for auto-increment creation, DatasetReference, EvaluationSuiteEvaluator, EvaluationSuiteKind (dataset_backed/evaluator_only), agentName filter.

Adapted to ai-foundry/data-plane/Foundry/src/ structure.

# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/client.tsp
```diff
@@ -169,6 +170,18 @@ using Azure.AI.Projects;
...
+// --------------------------------------------------------------------------------
+@@clientName(EvaluationSuites.listLatest, "list");
+
+// Evaluation suite generation jobs — renamed for Python SDK discoverability
+@@clientName(EvaluationSuiteGenerationJobs.create, "create_generation_job", "python");
+@@clientName(EvaluationSuiteGenerationJobs.get, "get_generation_job", "python");
+@@clientName(EvaluationSuiteGenerationJobs.list, "list_generation_jobs", "python");
+@@clientName(EvaluationSuiteGenerationJobs.cancel, "cancel_generation_job", "python");
+@@clientName(EvaluationSuiteGenerationJobs.delete, "delete_generation_job", "python");
+
```
dargilco: You can drop the "python" at the end of all lines above. This client.tsp file is used by both Python and JS emitters, and we usually have similar naming considerations for both.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,240 @@
...
+@doc("""
+  An evaluation suite bundles testing criteria — an optional dataset, one or more
+  evaluator configs with thresholds and init params — into a reusable, named artifact that
+  can gate agent changes across batch, scheduled, continuous, and CI/CD evals.
+  """)
+@resource("evaluationSuites")
+model EvaluationSuiteVersion {
+  @doc("""
+    Human-readable display name.
+    Does not need to be unique. Shown in portal list views and eval reports.
```
dargilco: Perhaps "Foundry portal" instead of just "portal".

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,240 @@
...
+  dataset?: DatasetReference;
+
+  @doc("""
+    Testing criteria — pass-through to eval group testing_criteria.
+    Supports any grader type: azure_ai_evaluator, string_check, label_model, 
+    score_model, text_similarity, python, etc. Stored as raw JSON.
+    At least one entry is required.
+    """)
+  @minItems(1)
+  testingCriteria: Record<unknown>[];
```
dargilco: Is this really Record<unknown>, or is it actually Record<string>?

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,240 @@
...
+    """)
+  target?: Target;
+
+  @doc("""
+    How to send dataset rows to the target (agent or model).
+    Pass-through to eval run data_source input_messages.
+    Supports template type (prompt with column placeholders) and
+    item_reference type (column containing pre-built messages).
+    """)
+  inputMessages?: Record<unknown>;
```
dargilco: Same as above

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,240 @@
...
+  generationModel: string;
+
+  @doc("Category: 'quality' (default) or 'safety'.")
+  category?: string;
+
+  @doc("""
+    Optional initialization parameters applied to all generated evaluators.
+    For example, deployment_name for LLM judge model, default threshold.
+    """)
+  initializationParameters?: Record<unknown>;
```
dargilco: Same

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
```diff
@@ -0,0 +1,107 @@
...
+  @query("agentName")
+  @doc("Filter by associated Foundry agent name (from target).")
+  agentName?: string;
+};
+
+// ---------------------------------------------------------------------------
+// CRUD operations (versioned asset pattern)
+// ---------------------------------------------------------------------------
+
+interface EvaluationSuites
```
dargilco, line 29: Do you really need to interfaces here? Why not have everything defined in one interface EvaluationSuites? (that's the folder name). Usually we put different interfaces in different folders.

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/client.tsp
vebudumu: removed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: Fixed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: Updated - changed testingCriteria, inputMessages, and initializationParameters to Record<string, unknown>

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
vebudumu: They're separate interfaces because they have different route prefixes (/evaluationSuites/... vs /evaluation_suite_generation_jobs/...). Happy to move generation jobs to a separate folder if that's the
 preferred convention.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: Tried Record<string, unknown> but TypeSpec's Record<T> only accepts one type parameter - keys are always strings implicitly. Record<string, unknown> doesn't compile..

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
dargilco: okay to place in the same folder for now. 

Related question about routes... why is one CamelCase and the other snake_case? (comparing /evaluationSuties to /evaluation_suites_jobs). I think we want to stick to one format. Gerardo may know more about our guidelines here.

vebudumu: Looking at the other routes, CRUD uses camelCase while the jobs from dataset gen and adaptive evals used snake case. Followed the same here. Happy to align with Gerardo's guidelines if this should be different.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: For our case, Record<unknown> is intentional - testingCriteria elements contain nested objects (initialization_parameters with numeric thresholds, data_mapping with string values, arrays), so the values aren't just strings. Same for others. Record<string> would be too restrictive for these fields.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: updated above

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: updated in the above comment

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+  @doc("Conversation-level evaluation.")
+  conversation: "conversation",
+}
+
+@doc("The subtype of an evaluation suite.")
+union EvaluationSuiteSubtype {
+  string,
+
+  @doc("Benchmark spec suite.")
+  benchmarkspec: "benchmarkspec",
```
glecaros: should we split this? (i.e `benchmark_spec`)

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+// ---------------------------------------------------------------------------
+// CRUD Resource model
+// ---------------------------------------------------------------------------
+
+@doc("""
+  An evaluation suite bundles testing criteria — an optional dataset, one or more
+  evaluator configs with thresholds and init params — into a reusable, named artifact that
+  can gate agent changes across batch, scheduled, continuous, and CI/CD evals.
+  """)
+@resource("evaluationSuites")
```
glecaros: path should be snake_cased

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+using OpenAPI;
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Enums
+// ---------------------------------------------------------------------------
+
+@doc("The evaluation level for an evaluation suite.")
+union EvaluationSuiteEvaluationLevel {
```
glecaros: #42488 also introduces an EvaluationLevel... would it make sense to reuse the type or do we foresee them diverging?

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+  suite_name: string;
+
+  @doc("The suite version resolved.")
+  suite_version: string;
+
+  @doc("Status of the run.")
+  status: string;
+
+  @doc("Unix timestamp when the run was created.")
+  created_at: int64;
```
glecaros: for this, please use the FoundryTimestamp type (available in servicepatterns.tsp)

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+  evaluation_name: string;
+
+  @doc("The suite name used.")
+  suite_name: string;
+
+  @doc("The suite version resolved.")
+  suite_version: string;
+
+  @doc("Status of the run.")
+  status: string;
```
glecaros: Are the statuses known values or open strings?

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+    Optional data generation options. Controls how the evaluation dataset is generated.
+    If omitted, the service uses sensible defaults.
+    """)
+  data_generation_options?: EvaluationSuiteDataGenerationOptions;
+}
+
+@doc("Options for dataset generation within an evaluation suite generation job.")
+model EvaluationSuiteDataGenerationOptions {
+  @doc("The data generation type: 'simple_qna', 'traces', 'tool_use', or 'task'.")
+  type?: string;
```
glecaros: should this be optional? Also, can we make this a union instead of an open string?

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+  name: string;
+
+  @doc("Source materials for generation — agent context, prompts, traces, or datasets.")
+  sources: EvaluationSuiteJobSource[];
+
+  @doc("The LLM model to use for rubric and data generation (e.g., 'gpt-4o'). Distinct from deployment_name in initializationParameters which is the LLM judge model used at eval time.")
+  generation_model: string;
+
+  @doc("Category: 'quality' (default) or 'safety'.")
+  category?: string;
```
glecaros: should this be a union (and we should capture the default)

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,279 @@
...
+    """)
+  target?: Target;
+
+  @doc("""
+    How to send dataset rows to the target (agent or model).
+    Pass-through to eval run data_source input_messages.
+    Supports template type (prompt with column placeholders) and
+    item_reference type (column containing pre-built messages).
+    """)
+  input_messages?: Record<unknown>;
```
glecaros: Since this does support two types of item. we should probably spec that

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
glecaros: while we do need to have separate routes for `evaluation_suites` and `evaluation_suites_jobs`, we do require snake_case for the routes. There's some of the already available services  that are using camel, but going forward we are enforcing that as a guideline.

vebudumu: changed to snake case.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: they are the same. removed this.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed to JobStatus

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: added

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: added

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: it can be optional. Also made it to union

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Child models
+// ---------------------------------------------------------------------------
+
+@doc("""
+  Reference to a dataset by name and version.
+  """)
+model DatasetReference {
```
dargilco: I would go with a more descriptive name, such as "EvaluationDatasetReferece" (since we already have dataset operations and models that define their input and output)

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,325 @@
...
+  @doc("Quality-focused rubric criteria.")
+  quality: "quality",
+
+  @doc("Safety-focused policy criteria.")
+  safety: "safety",
+}
+
+@doc("The data generation type.")
+union DataGenerationType {
+  string,
```
dargilco, line 41: I would go with a more descriptive name to scope this to evaluation operations, such as "EvaluationDataGenerationType"

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
```diff
@@ -0,0 +1,325 @@
...
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Enums
+// ---------------------------------------------------------------------------
+
+@doc("The subtype of an evaluation suite.")
+union EvaluationSuiteSubtype {
+  string,
+
```
dargilco, line 20: Please check all unions: If the union is only use in request models, then don't include "string". If the union is used in response models (or both in request and response models), then okay to extend it with "string" (so emitted code does not break if the service suddenly starts returning other string values).

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: Checked all unions — they all appear in both request and response models

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/models.tsp
vebudumu: changed

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
```diff
@@ -0,0 +1,109 @@
...
+    EvaluationSuiteVersion,
+    ListEvaluationSuiteVersionsParameters,
+    Azure.Core.Foundations.ResourceCreatedResponse<EvaluationSuiteVersion>
+  >;
+
+  // POST /evaluation_suites/{name}:run
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  @doc("Run an evaluation using the suite's testing criteria and dataset.")
+  @action("run")
+  @post
```
glecaros, line 60: This being an LRO, should use the `postJobPreview` (as the necessary information for an LRO needs to be returned)

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
```diff
@@ -0,0 +1,109 @@
...
+      EvaluationSuiteFilterQueryParams,
+      ListEvaluationSuiteVersionsParameters
+    > {
+
+  // POST /evaluation_suites/{name}/versions
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations"
+  @doc("Create a new {name} with auto incremented version id", EvaluationSuiteVersion)
+  @Rest.segment("versions")
+  @post
+  createEvaluationSuiteVersion is Azure.Core.Foundations.ResourceOperation<
```
glecaros: Use the `FoundryDataPlanePreviewOperation` here

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
vebudumu: The :run action returns a results[] array with multiple run references (eval_id, run_id, status), since a single suite run can create multiple evaluation runs. postJobPreview expects a single top-level id/status for LRO polling, which doesn't align with our multi-result response shape. Users poll run status via the existing eval run endpoints (GET /openai/evals/{eval_id}/runs/{run_id}) rather than through an Operation-Location header.

specification/ai-foundry/data-plane/Foundry/src/evaluation-suites/routes.tsp
vebudumu: Switching to FoundryDataPlanePreviewOperation directly breaks routing — it's a raw op that doesn't resolve the @resource + @key context, so the generated route becomes /{name} instead of /evaluation_suites/{name}/versions. Kept ResourceOperation for correct routing and added the Foundry-Features preview header by spreading WithConditionalFoundryPreviewHeader<evaluations_v1_preview> into
the params. Same end result in the generated OpenAPI — correct route + optional preview header.

specification/ai-foundry/data-plane/Foundry/main.tsp
```diff
@@ -9,6 +9,7 @@ import "./src/datasets/routes.tsp";
 import "./src/deployments/routes.tsp";
 import "./src/evaluation-taxonomies/routes.tsp";
 import "./src/evaluation-rules/routes.tsp";
+import "./src/evaluation-suites/routes.tsp";
 import "./src/evaluations/routes.tsp";
 import "./src/evaluators/routes.tsp";
 import "./src/indexes/routes.tsp";
```
dargilco, line 15: Since you mapped the new Evaluation Suite Generation Jobs operations to the existing ".beta.evaluators" subclient, you now need to give them unique names, since we already have these methods there: 

Image

@@clientName(EvaluationSuiteGenerationJobs.get, "get_generation_suite_job");
@@clientName(EvaluationSuiteGenerationJobs.list, "list_generation_suite_jobs");
@@clientName(EvaluationSuiteGenerationJobs.create, "create_generation_suite_job");
@@clientName(EvaluationSuiteGenerationJobs.cancel, "cancel_generation_suite_job");
@@clientMame(EvaluationSuiteGenerationJobs.delete, "delete_generation_suite_job");

vebudumu: Addressed

dargilco (approved):

johanste (approved):

PR #42454, opened 2026-04-17
Add ARM lease files for all resource providers (405 RPs)

vikeshi26:
## Summary

Adds ARM lease files for **405 resource providers** covering all RPs across:
- Main branch (98 no-group RPs + 255 service-group RPs)  
- RPSaaSMaster-only RPs (54 private RPs)

## Lease Configuration

- **Reviewer:** @vikeshi26
- **Start date:** 2026-04-28
- **Durations:** Staggered across 6 cohorts to spread review load evenly:

| Cohort | Duration | Count | Expiry |
|--------|----------|-------|--------|
| 1 | P30D | 50 | May 28 |
| 2 | P60D | 71 | Jun 27 |
| 3 | P90D | 71 | Jul 27 |
| 4 | P120D | 71 | Aug 26 |
| 5 | P150D | 71 | Sep 25 |
| 6 | P180D | 71 | Oct 25 |

## Other Changes

- `.gitignore`: Added temp RP inventory files (`rps-no-groups.txt`, `rps-with-groups.txt`, `rps-pr-repo.txt`, `groups.txt`)


mikeharder (changes requested):
I'm getting concerned about the number of files we're adding to the repo here.  I did approve the original design, but I thought the number of files would be lower, or more gradually added/removed.

Seeing this number of files, I think we should instead move to a single file `arm-leases.yaml`, with the same `lease` contents, just in a single file rather than 406 files.  Or some other design that's easier to manage than 406 files in nested folders.

CC: @tejaswiMinnu, @weshaggard

.gitignore
```diff
@@ -139,6 +139,12 @@ eng/tools/**/dist
 !/.github/package-lock.json
 !/.github/shared/package-lock.json
 
+# Temp RP inventory files
```
mikeharder, line 142: please revert this, you can store the files outside the repo locally

mikeharder:
Based on the checks in this PR, these 409 lease files don't appear to be impacting checkout perf.  So while I still think it would be better to refactor them into a single file, I'm fine with merging them as-is for now.

One error you will need to fix, prettier is failing because you have invalid yaml:

image

Reviewer name needs to be quoted. In the widget example, startdate is also quoted, but I'm not sure if that's necessary or optional.

mikeharder (changes requested):
Fix prettier errors in yaml files.  At a minimum, reviewer names starting with "@" need to be quoted.  We should also require trailing newlines -- i think prettier requires this by default, and it might flag after you quote the reviewer names.

.gitignore
tejaswiMinnu: Reverted

mikeharder (approved):

PR #42478, opened 2026-04-17
blobs: customize enums for c#

jorgerangel-msft:
Customizes a few more enums to match the existing transforms for both blobs and blobs.batch.

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -263,12 +263,12 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 // GO SPECIFICATION
 /** The type of blob deletions. */
 #suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
-enum BlobDeleteTypeGo {
+enum BlobDeleteTypeCustom {
+  /** None. */
+  None: "None",
+
   /** Permanently delete the blob. */
   Permanent: "Permanent",
-
-  /** Deprecated */
-  None: "None",
 }
 /** Represents a blob prefix. */
 @Xml.name("BlobPrefix")
@@ -444,12 +444,12 @@ model BlobPropertiesGo {
 }
 /** The algorithm used to produce the encryption key hash. Currently, the only accepted value is \"AES256\". Must be provided if the x-ms-encryption-key header is provided. */
 #suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
-enum EncryptionAlgorithmTypeGo {
-  /** The AES256 encryption algorithm. */
-  AES256: "AES256",
-
+enum EncryptionAlgorithmTypeCustom {
   /** Deprecated */
   None: "None",
+
+  /** AES256 */
+  AES256: "AES256",
 }
 /** The list blob includes parameter values. */
 #suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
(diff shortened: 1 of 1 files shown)


specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -733,6 +733,27 @@ model BlobServicePropertiesCsharp {
...
 
+/** The type of blob deletions. */
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
+enum BlobDeleteTypeCsharp {
+  /** None. */
+  None: "None",
+
+  /** Permanently delete the blob. */
+  Permanent: "Permanent",
+}
```
catalinaperalta: I believe Go has the same enum def can we see about reusing their def?

specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -733,6 +733,27 @@ model BlobServicePropertiesCsharp {
...
+
+/** The algorithm used to produce the encryption key hash. Currently, the only accepted value is \"AES256\". Must be provided if the x-ms-encryption-key header is provided. */
+#suppress "@azure-tools/typespec-azure-core/no-enum" "Existing API"
+enum EncryptionAlgorithmTypeCsharp {
```
catalinaperalta: I dont recall if this one is also defined for go, but would be good to check with @gracewilcox

specification/storage/Microsoft.BlobStorage/client.tsp
jorgerangel-msft: @catalinaperalta unfortunately .net has the enums in a different order. "None" being the first member. This enum _is_ internal but I'd prefer to not causing a behavioral breaking change if somewhere in the custom code we are relying on the integer value.

catalinaperalta: @gracewilcox does the enum member order matter for go? It would be awesome if we can consolidate these similar customizations if possible.

specification/storage/Microsoft.BlobStorage/client.tsp
jorgerangel-msft: @catalinaperalta this enum is also defined with members in a different order in c# ☹️ .

specification/storage/Microsoft.BlobStorage/client.tsp
gracewilcox: the order doesn't matter for Go, we can consolidate the two

specification/storage/Microsoft.BlobStorage/client.tsp
gracewilcox: can change the order to consolidate

specification/storage/Microsoft.BlobStorage/client.tsp
catalinaperalta: @jorgerangel-msft do you mind doing the consolidation in this PR? You have to remove Grace's def, give this enum a more generic name and replace it where Grace is referencing it.

jorgerangel-msft: @catalinaperalta sure, fixed! @gracewilcox can you also give it a review please to make sure I'm not breaking you in anyway?

catalinaperalta (approved):
From a tsp perspective this lgtm! Good idea to wait for @gracewilcox to confirm it's all good for go

PR #42492, opened 2026-04-18
[Azure Firewall] Smart Groups

saisujithreddym:
# Choose a PR Template


specification/network/resource-manager/Microsoft.Network/Network/Common/main.tsp
```diff
@@ -90,6 +90,10 @@ enum Features {
   serviceGateway,
   #suppress "@azure-tools/typespec-azure-core/documentation-required"
   virtualNetworkAppliance,
+
+  #suppress "@azure-tools/typespec-azure-core/documentation-required"
+  @added(Versions.v2025_07_01)
```
pshao25:
```suggestion
```

specification/network/resource-manager/Microsoft.Network/Network/Common/main.tsp
```diff
@@ -90,6 +90,9 @@ enum Features {
   serviceGateway,
   #suppress "@azure-tools/typespec-azure-core/documentation-required"
   virtualNetworkAppliance,
+
+  #suppress "@azure-tools/typespec-azure-core/documentation-required"
```
razvanbadea-msft, line 94: >suppress [](http://example.com/codeflow?start=3&length=8)

Add descriptions for this enum values. They are used for generating documentation improving customer experience

razvanbadea-msft (commented):
🔴 **Blocking — Resource name pattern allows empty string and has no length bounds**

`SmartGroup.tsp` declares the resource `name` with `@pattern("^[a-zA-Z0-9_.-]*$")` (and the generated `smartGroupName` path parameter in `smartGroup.json` inherits this pattern with no `minLength` / `maxLength`).

Two issues, both of which become permanent in a **stable** API:

1. The `*` quantifier allows an empty string. An empty name collapses `/providers/Microsoft.Network/smartGroups/{smartGroupName}` into the collection path — an ambiguity ARM resource names must not allow (RPC-Uri-V1, RPC-Put-V1-15).
2. There are no length bounds. ARM resource names conventionally cap at 80 characters; without `@minLength` / `@maxLength` the server and clients disagree on validity.

Please change to:

```tsp
@pattern("^[a-zA-Z_][a-zA-Z0-9_.-]*$")
@minLength(1)
@maxLength(80)
name: string;
```

Adjust the regex/length to whatever the SmartGroups resource implementation actually enforces (and align with sibling Network resources such as `FirewallPolicy`). Since this is a stable GA API, the constraints will be baked in permanently — they need to be correct on first release.

razvanbadea-msft (commented):
🔴 **Blocking — `ProvisioningStateMustBeReadOnly` suppression likely masks an emitter defect**

The justification for this suppression in `readme.md` states _"provisioningState is correctly marked readOnly in SmartGroupProperties definition. The linter does not follow $ref chains to verify readOnly in referenced schemas."_

This explanation doesn't match how the rule works. `ProvisioningStateMustBeReadOnly` inspects the **emitted property node** in the generated swagger, not the referenced schema — so "the linter doesn't follow $refs" is not a valid defense. The rule fires precisely when `readOnly: true` is missing from the emitted JSON at the property site.

Please paste the generated fragment from `stable/2025-07-01/smartGroup.json` for `SmartGroupProperties.properties.provisioningState`. One of two things is true:

1. **The emitted JSON is `{ "$ref": "#/definitions/ProvisioningState" }` with no sibling `readOnly: true` (and no `allOf` wrapper carrying it).** Then the linter is correct and `readOnly` is being silently dropped on the wire. This is a real defect and cannot be suppressed — it must be fixed.
2. **The emitted JSON does carry `readOnly: true`** (e.g. via `allOf` wrapping). Then the suppression is valid as a linter workaround, but the reason text needs to be rewritten to reference a concrete validator bug/issue link, not the incorrect "$ref chain" rationale.

### Why this keeps happening across NRP resources

The structural cause is that `SmartGroup` uses custom `Common.Resource` + `@Azure.ResourceManager.Legacy.feature` instead of `Azure.ResourceManager.TrackedResource<SmartGroupProperties>`. The Legacy.feature path bypasses the ARM emitter hook that wraps `$ref` in an `allOf` with `readOnly: true` when `@visibility(Lifecycle.Read)` is set on a union-typed property — which is why this same suppression keeps appearing across NRP resources.

Two concrete asks, in increasing leverage:

- **Minimum:** extend `Azure.ResourceManager.ResourceProvisioningState` inside `Common/main.tsp`'s `ProvisioningState` union (the canonical pattern — see [apicenter/ApiCenter.Management/models.tsp](specification/apicenter/ApiCenter.Management/models.tsp) line 44). Hand-rolling `Succeeded`/`Failed`/`Canceled` doesn't trigger the ARM-aware emitter path.

    ```tsp
    union ProvisioningState {
      string,
      Azure.ResourceManager.ResourceProvisioningState,
      Creating: "Creating",
      Updating: "Updating",
      Deleting: "Deleting",
    }
    ```

- **Root-cause fix:** migrate `SmartGroup` to `Azure.ResourceManager.TrackedResource<SmartGroupProperties>`. Since SmartGroup is a brand-new 2025 resource with no existing clients, there is no back-compat constraint. This single change removes `ProvisioningStateMustBeReadOnly`, `RequiredPropertiesMissingInResourceModel`, `arm-custom-resource-usage-discourage`, `composition-over-inheritance`, 7× `no-legacy-usage`, and `arm-resource-invalid-envelope-property` on `etag` — plus it gives you `systemData` in responses for free.

Please either provide the emitted JSON proof (and link a validator bug if it's truly a false positive), or adopt one of the fixes above.

razvanbadea-msft:
Fix errors from TypeSpec Validation / TypeSpec Validation

PR #42535, opened 2026-04-21
[.github] limit sparse-checkout to actions|shared|workflows

mikeharder:


mikeharder:
doesn't improve perf enough to be worth it (yet)

PR #42543, opened 2026-04-21
Fix async DELETE/PATCH response code false positives in ARM API reviewer instructions

ravimeda:
## Problem

The ARM API reviewer agent instructions conflate sync and async response codes for DELETE and PATCH operations, causing false-positive review findings:

- **Async DELETE**: The agent flags async DELETE for missing `200`, which is only required for synchronous DELETE. Section 6.3 mentions `200` in the context of the polling URL's terminal response, adding to the confusion.
- **Async PATCH**: The instructions state async PATCH must return `202` but don't explain that `200` must **also** be defined in the swagger for SDK response type discovery, a requirement unique to async PATCH.
- **LRO cross-references**: The general LRO section in `openapi-review.instructions.md` doesn't direct reviewers to the ARM-specific sections for per-verb response code details, leading to inconsistent guidance.

### `armapi-review.instructions.md`

- **Section 4.2**: Added callout explaining why async PATCH uniquely requires both `202` and `200` in the swagger definition (SDK deserialization).
- **Section 5.1**: Added false-positive warning: async DELETE requires `202` + `204` only; do not flag for missing `200`.
- **Section 6.3**: Added note clarifying that `200`/`204` refers to the polling URL's terminal response, not the initial DELETE response codes.

### `openapi-review.instructions.md`

- **LRO section**: Added cross-reference directing reviewers to `armapi-review.instructions.md` sections 5-6 for ARM-specific response codes and section 21 for data-plane LRO patterns.


sandipsh (approved):

razvanbadea-msft (approved):

haolingdong-msft:
I'm fine with this PR change as long as ARM team is good, as this PR does not modify the typepsec instruction part which might impact typespec authoring. From my side, just have the same comment with this PR. https://github.com/Azure/azure-rest-api-specs/pull/42567
Pure prompts might be hard to review, if we have benchmark test report, it would be easier to review. Thanks!

praveenkuttappan (approved):

laxmankumar12 (approved):

PR #42575, opened 2026-04-22
Add read-only content_hash to hosted agent CodeConfiguration

sagunase:
## Summary

Adds a read-only `content_hash` property to `CodeConfiguration` under the hosted agent's `code_configuration`. The value is the SHA-256 hex digest of the uploaded code zip. Clients already supply it via the existing `x-ms-code-zip-sha256` request header; the service echoes it back on the agent/version resource so callers can verify integrity.

## Design

- `@visibility(Lifecycle.Read)` on the new property → emits `readOnly: true` in OpenAPI, so the field is excluded from request schemas and present in response schemas.
- Optional (`?`) to avoid requiring it on the wire.
- No route changes — the header is already bound on all three code-upload routes.
- No `@added` decorator needed — parent model is already gated by `AgentDefinitionOptInKeys.code_agents_v1_preview`.

## Changes

- `src/agents/models.tsp`: add read-only `content_hash?: string` to `model CodeConfiguration`.
- `openapi3/v1/microsoft-foundry-openapi3.json` and `openapi3/virtual-public-preview/microsoft-foundry-openapi3.json`: `content_hash` with `readOnly: true` in `CodeConfiguration`.
- Three `examples/2025-11-15-preview/Agents_*FromCode_MaximumSet_Gen.json` files: `content_hash` added to the response `code_configuration` only (request bodies unchanged).

## Validation

- `tsp compile` passes with no new warnings.
- `content_hash` is absent from request body schemas for the three code-upload routes and present (readOnly) in responses.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -415,6 +415,10 @@ model CodeConfiguration {
 
   @doc("The entry point command and arguments for the code execution.")
   entry_point: string[];
+
+  @visibility(Lifecycle.Read)
+  @doc("The SHA-256 hex digest of the uploaded code zip. Set by the service from the `x-ms-code-zip-sha256` request header; read-only in responses and never accepted in request payloads.")
+  content_hash?: string;
 }
 
 @doc("The hosted agent definition.")
--- a/specification/ai-foundry/data-plane/Foundry/examples/2025-11-15-preview/Agents_CreateAgentFromCode_MaximumSet_Gen.json
+++ b/specification/ai-foundry/data-plane/Foundry/examples/2025-11-15-preview/Agents_CreateAgentFromCode_MaximumSet_Gen.json
@@ -47,7 +47,8 @@
             "entry_point": [
               "python",
               "main.py"
-            ]
+            ],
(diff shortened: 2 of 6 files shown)


dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -415,6 +415,10 @@ model CodeConfiguration {
 
   @doc("The entry point command and arguments for the code execution.")
   entry_point: string[];
+
+  @visibility(Lifecycle.Read)
+  @doc("The SHA-256 hex digest of the uploaded code zip. Set by the service from the `x-ms-code-zip-sha256` request header; read-only in responses and never accepted in request payloads.")
```
glecaros, line 420: Just to clarify the language of the doc, will the service 4xx when the field is provided, or just ignore?

sagunase: Service will return 4xx if field is provided. Because it will conflict with header.

glecaros: I would recommend ignoring to service the scenario of a user receiving the value in a response and then sending the unmodified (or edited) object in an update request.

glecaros: nvm, this seems fine :)

glecaros (approved):

PR #42579, opened 2026-04-22
Fix 2026-04-01 GA scope: Remove ingestionPermissionOptions and add modelWebSummarization

yangylu91:
This PR updates the `2026-04-01` (GA) scope:

1. Remove `ingestionPermissionOptions` from `KnowledgeSourceIngestionParameters`
   - Other permission-related properties and models (for example, `indexerPermissionOptions`) tied to content security were previously removed from `2026-04-01` (GA), but this one was missed. Since GA does not support `x-ms-query-source-authorization` on the query side, ingestion permission options should not be exposed. This change removes the ingestion-permission surface from both the spec and examples to keep the GA contract consistent.

2. Add `modelWebSummarization` to `KnowledgeBaseActivityRecordType`
   - This activity record type was intended for GA but was missing due to a service-side bug, which bypassed existing tests and was only identified during recent testing. It's also missed in the API review process of GA. We are now fixing the service, the process and the spec.

Risk mitigation:
- GA has not been announced yet.
- 2026-04-01 REST API reference docs have not been generated.
- SDKs have not been released. The SDK v-team will regenerate them and inform SDK review architects to ensure all fixes are included before release.


vaishalishah11:
Looks good to me.

yangylu91:
@copilot Update the PR title and description to cover all the changes.

efrainretana (approved):

PR #42602, opened 2026-04-23
Added NEW observabilityAgent resource type 2026-05-01-preview

yashmuel:
# Choose a PR Template


specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
```diff
@@ -0,0 +1,543 @@
...
+          }
+        ],
+        "responses": {
+          "202": {
+            "description": "Resource deletion accepted.",
+            "headers": {
+              "Location": {
+                "type": "string",
+                "description": "The Location header contains the URL where the status of the long running operation can be checked."
+              },
```
sandipsh: [NEW] 🔴 Blocking [RPC-Async-V1-09 / EX-LRO-HEADERS]
Async DELETE 202 should include both `Location` and `Azure-AsyncOperation` headers. Right now only `Location` is defined.

Why this matters: clients may not have a reliable polling URL contract, which can break LRO tracking behavior.

Fix: add `Azure-AsyncOperation` here and in `preview/2026-05-01-preview/examples/ObservabilityAgents_Delete.json`. #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
```diff
@@ -0,0 +1,543 @@
...
+          "description": "The resource provisioning state.",
+          "readOnly": true
+        },
+        "monitoringAccountId": {
+          "type": "string",
+          "format": "arm-id",
+          "description": "The resource ID of the Azure Monitor workspace used as the issue container."
+        },
+        "enabled": {
+          "type": "boolean",
```
sandipsh, line 455: [NEW] 🔴 Blocking [PLCY004]
`disabledOperations` is modeled as a comma-delimited `string`. Collections should be JSON arrays, not CSV strings.

Why this matters: policy/tooling and SDK validation cannot reliably reason over individual values.

Fix: change this to an array of strings (or an extensible enum array), and update request/response examples accordingly. #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
```diff
@@ -0,0 +1,543 @@
...
+            "schema": {
+              "$ref": "../../../../../../common-types/resource-management/v6/types.json#/definitions/ErrorResponse"
+            }
+          }
+        },
+        "x-ms-examples": {
+          "ObservabilityAgents_Delete": {
+            "$ref": "./examples/ObservabilityAgents_Delete.json"
+          }
+        },
```
sandipsh: [NEW] 🟡 Warning [LRO-final-state-via]
For standard ARM DELETE LROs, avoid setting `x-ms-long-running-operation-options.final-state-via`.

Why this matters: default ARM SDK behavior is already correct for DELETE; this override can create unnecessary divergence.

Fix: remove the `x-ms-long-running-operation-options` block unless this operation is intentionally non-standard. #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/service.tsp
```diff
@@ -0,0 +1,30 @@
...
+import "@azure-tools/typespec-azure-resource-manager";
+
+using Azure.ResourceManager;
+using TypeSpec.Versioning;
+
+/**
+ * Monitor Agents Management Client
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "Operations API was moved to its own service."
+@armProviderNamespace("Microsoft.Monitor")
```
sandipsh, line 12: [NEW] 🟡 Warning [Suppression hygiene]
This suppression appears stale because `interface Operations extends Azure.ResourceManager.Operations {}` is already present and the generated swagger includes `/providers/Microsoft.Monitor/operations`.

Why this matters: stale suppressions hide real regressions and trigger avoidable suppression-review churn.

Fix: remove this `#suppress` unless there is still a concrete missing-operations-endpoint signal in current CI. #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
sandipsh: Checking this thread on current head: still open and still relevant. Please either add the `Azure-AsyncOperation` header in swagger + delete example, or explain if the service intentionally diverges from ARM LRO header contract. Without this, client polling behavior can be inconsistent.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
sandipsh: Following up on this open blocker: can you update `disabledOperations` from CSV string to a JSON array (and align examples), or share why array modeling cannot be used here? If left as CSV, policy/tooling and SDK validation remain weak for individual operations.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
sandipsh: This warning is still open on current head. Please remove `final-state-via: location` for this standard ARM DELETE LRO unless there is a confirmed non-standard polling contract that requires it. Keeping it without need increases SDK behavior divergence risk.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/service.tsp
sandipsh: This suppression still looks stale because operations endpoint is present in generated swagger. Please drop the suppression or add a concrete current CI signal showing why it is still required. Keeping stale suppressions makes suppression review harder and can mask real regressions.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
yashmuel: We removed the 202 and async response at all

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
yashmuel: we've changed to string array

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/service.tsp
yashmuel: removed

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
ravimeda: **Re-review on current head (commit 9b93f82):** This violation appears to be **resolved**. The DELETE operation in [agents.json](https://github.com/Azure/azure-rest-api-specs/blob/9b93f82b96fcdeb12cea223043f1fe135eb9eabd/specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json#L327-L370) is now synchronous (`200` + `204` only) and no longer uses `202` async, so the `Azure-AsyncOperation` header requirement no longer applies. Sync DELETE is permitted by the ARM RPC for RPaaS resources.

Deferring the resolve action to @sandipsh.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
ravimeda: **Re-review on current head (commit 9b93f82):** This is **resolved**. `disableAutonomousOperations` is now an array of strings (see [agents.json line 471–476](https://github.com/Azure/azure-rest-api-specs/blob/9b93f82b96fcdeb12cea223043f1fe135eb9eabd/specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json#L471-L476)), not a CSV string. The CSV/PLCY004 issue is no longer present. (I have a separate new finding suggesting the array element should be modeled as an extensible enum rather than a free-form string — posting separately.)

Deferring the resolve action to @sandipsh.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
ravimeda: **Re-review on current head (commit 9b93f82):** This is **resolved**. The DELETE operation no longer uses LRO at all — it is now synchronous (`200` + `204`). There is no `x-ms-long-running-operation-options` / `final-state-via` anywhere in the swagger.

Deferring the resolve action to @sandipsh.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/service.tsp
ravimeda: **Re-review on current head (commit 9b93f82):** This is **resolved**. `suppressions.yaml` is empty (`[]`); `service.tsp` line 12 contains the `@service` decorator, no suppression. The `interface Operations extends Azure.ResourceManager.Operations {}` is present (line 28) and the `/providers/Microsoft.Monitor/operations` endpoint appears in the generated swagger (agents.json line 50). The stale-suppression concern no longer applies.

Deferring the resolve action to @sandipsh.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+    provisioningState?: Azure.ResourceManager.ResourceProvisioningState;
+
+    /** The resource ID of the Azure Monitor workspace used as the issue container. */
+    monitoringAccountId?: Azure.Core.armResourceIdentifier;
+
+    /** Whether the observability agent is enabled. */
+    enabled?: boolean;
+
+    /** List of autonomous operations that are disabled on this observability agent. Currently supported values: AutoInvestigate. Additional values may be added in the future. */
+    disableAutonomousOperations?: string[];
```
ravimeda: **[NEW] 🔴 Blocking** **[Section 11 — Enum best practices / OAPI008]**

This array is typed as `string[]`, but the doc comment itself says _"Currently supported values: `AutoInvestigate`. Additional values may be added in the future."_ — that is the textbook definition of an extensible (open) enum. Free-form strings give clients no IntelliSense, no validation, and no SDK constants for the only known value (`AutoInvestigate`).

This is a different concern from @sandipsh's earlier (now-resolved) PLCY004 thread, which was about CSV-vs-array. The CSV→array fix is in; this finding is about the array element type.

**Suggested fix:** Define an extensible enum union and use it as the array element type:

```tsp
/** Autonomous operations supported by an observability agent. */
union AutonomousOperation {
  string,

  /** AI-driven autonomous investigation of detected issues. */
  AutoInvestigate: "AutoInvestigate",
}

// in ObservabilityAgentProperties:
disableAutonomousOperations?: AutonomousOperation[];
```

 #Resolved

ravimeda: **[NEW] 🟡 Warning** **[Boolean / inverted-naming guidance]**

`disableAutonomousOperations` uses an inverted/negative form ("disable…"). Negative property names lead to confusing double-negative semantics:
- _"the list of operations that are disabled is empty"_ vs.
- _"all operations are enabled"_

Together with the enum finding above, this also conflates two different things — _which operations exist_ and _which are turned off_.

**Suggested fix (pick one):**
- Rename to `enabledAutonomousOperations: AutonomousOperation[]` with semantics "if omitted, all known operations are enabled".
- Or split into per-operation settings (e.g., add `enabled?: boolean` to `OperationSettings`) so each operation has explicit on/off configuration.

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Properties for an observability agent resource.
+   */
+  model ObservabilityAgentProperties {
+    /** The resource provisioning state. */
+    @visibility(Lifecycle.Read)
+    provisioningState?: Azure.ResourceManager.ResourceProvisioningState;
+
+    /** The resource ID of the Azure Monitor workspace used as the issue container. */
+    monitoringAccountId?: Azure.Core.armResourceIdentifier;
```
ravimeda: **[NEW] 🟡 Warning** **[Section 2.2 — Type Constraints / armResourceIdentifier]**

`monitoringAccountId?: Azure.Core.armResourceIdentifier` is untyped — clients have no signal that this must be a `Microsoft.Monitor/accounts` resource ID. Per [typespec-review §2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md): _"Properties named `<something>Id` MUST be specific about what kind of ID they hold."_

**Suggested fix:**
```tsp
/** Resource ID of the Azure Monitor workspace (account) used by this agent. */
monitoringAccountId?: Azure.Core.armResourceIdentifier<[
  { type: "Microsoft.Monitor/accounts" }
]>;
```

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * An observability agent resource.
+   */
+  model ObservabilityAgentResource
+    is Azure.ResourceManager.TrackedResource<ObservabilityAgentProperties> {
+    ...ResourceNameParameter<
+      Resource = ObservabilityAgentResource,
+      KeyName = "observabilityAgentName",
+      SegmentName = "observabilityAgents",
+      NamePattern = "^(?!-)[a-zA-Z0-9-]+[^-]$"
```
ravimeda: **[NEW] 🟡 Warning** **[Section 2.2 — resource-name pattern / arm-resource-name-pattern]**

The resource-name regex `^(?!-)[a-zA-Z0-9-]+[^-]\$` correctly forbids leading/trailing hyphens but imposes **no minimum or maximum length**. ARM resource names typically need an upper bound (commonly 1–63 characters) for routing / DNS / storage compatibility. The [typespec-review §2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md) guidance explicitly states the path-parameter name pattern _"SHOULD include a `@pattern` decorator with a maximum length limit"_.

This is also flagged as `arm-resource-name-pattern` by the ARM linter ruleset.

**Suggested fix:** Bake the bound directly into the regex (preferred so the OpenAPI `pattern` is also bounded), or add `@minLength` / `@maxLength`:

```tsp
...ResourceNameParameter<
  Resource = ObservabilityAgentResource,
  KeyName = "observabilityAgentName",
  SegmentName = "observabilityAgents",
  NamePattern = "^(?!-)[a-zA-Z0-9-]{1,62}[a-zA-Z0-9]\$"
>;
```

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
ravimeda: **[NEW] 🟡 Warning** **[Required vs optional / domain-essential property]**

`monitoringAccountId?` is optional. Conceptually the observability agent appears to need an Azure Monitor workspace to function (it's described as the workspace used as the "issue container"). If an agent without a workspace cannot operate, the property should be required so PUT requests fail fast and the contract is explicit.

If it really can be omitted, please expand the doc comment to describe the agent's behavior in that case.

**Suggested fix:** Drop the `?` (and ensure all examples set `monitoringAccountId`), **or** document the no-workspace behavior explicitly.

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+namespace Microsoft.Monitor {
+  /**
+   * Properties for an observability agent resource.
+   */
+  model ObservabilityAgentProperties {
+    /** The resource provisioning state. */
+    @visibility(Lifecycle.Read)
+    provisioningState?: Azure.ResourceManager.ResourceProvisioningState;
+
+    /** The resource ID of the Azure Monitor workspace used as the issue container. */
```
ravimeda: **[NEW] 🟡 Warning** **[Section 18 — Property descriptions / domain clarity]**

_"The resource ID of the Azure Monitor workspace used as the issue container."_

"Issue container" is internal jargon and not defined anywhere in the spec. External SDK consumers and partners will not know what it means.

**Suggested fix:** Reword to plain, self-contained language — for example:
_"Resource ID of the Azure Monitor workspace where this agent stores the issues / investigation results it produces."_
(Adjust to match the actual semantics.)

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+    monitoringAccountId?: Azure.Core.armResourceIdentifier;
+
+    /** Whether the observability agent is enabled. */
+    enabled?: boolean;
```
ravimeda, line 27: **[NEW] 🟡 Warning** **[Boolean default / docs]**

`enabled?: boolean` is optional but the doc says only _"Whether the observability agent is enabled."_ A reader cannot tell what happens on PUT if the field is omitted (defaults to `true`? `false`? request rejected?).

The same concern applies to `disableAutonomousOperations?` and `operations?` — none specify omission semantics.

**Suggested fix:** Document the default explicitly, e.g.:
```tsp
/** Whether the agent is enabled. Defaults to `true` when not specified. */
enabled?: boolean;
```

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,61 @@
...
+    /** Settings for the issue creation operation. */
+    issueCreation?: OperationSettings;
+  }
+
+  /**
+   * Settings for a specific agent operation.
+   */
+  model OperationSettings {
+    /** Custom instructions that guide the agent's behavior for this operation. */
+    customInstructions?: string;
```
ravimeda: **[NEW] 🟡 Warning** **[String length constraints]**

`customInstructions?: string` is unbounded. This is user-controlled prompt text consumed by an AI agent — without a `@maxLength` it is open to abuse (oversized payloads, prompt-bombing) and leaves SDK clients without validation guidance.

**Suggested fix:** Add a sane upper bound:
```tsp
/** Custom instructions that guide the agent's behavior for this operation. */
@maxLength(4096)
customInstructions?: string;
```

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/examples/Operations_List.json
```diff
@@ -0,0 +1,45 @@
...
+  "title": "Operations_List",
+  "operationId": "Operations_List",
+  "parameters": {
+    "api-version": "2026-05-01-preview"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "value": [
+          {
```
ravimeda, line 11: **[NEW] 💡 Suggestion** **[Operations_List completeness]**

This example lists only `read`, `write`, and `delete` for `Microsoft.Monitor/observabilityAgents`. The resource also supports PATCH — typically `write` covers PUT and PATCH (the common RBAC convention), so this is likely fine.

**Action:** Please confirm with the Microsoft.Monitor RP team that the operation list aligns with the RBAC actions actually registered for this resource type. If any action is registered separately (e.g., a data-action, `listKeys`, etc.), include it here.

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json
```diff
@@ -0,0 +1,555 @@
...
+    {
+      "name": "Operations"
+    },
+    {
+      "name": "ObservabilityAgents"
+    }
+  ],
+  "paths": {
+    "/providers/Microsoft.Monitor/operations": {
+      "get": {
```
ravimeda, line 51: **[NEW] 💡 Suggestion** **[Cross-spec consistency — Operations_List drift]**

Each per-resource swagger under `Microsoft.Monitor/` (Accounts, PipelineGroups, this new Agents) defines its own `/providers/Microsoft.Monitor/operations` with a subset of actions. ARM serves the union, so this is functionally OK, but it is fragile — adding/renaming an action on one resource without updating the others creates drift across SDK languages.

**Not introduced by this PR** — flagged for awareness. Consider coordinating with the other Microsoft.Monitor RP owners on whether to centralize the operations definition for this namespace.

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
yashmuel: non of the suggestions are applicable for us.
We have an alternative solution, let me know how it sounds:
to add under OperationSettings, next to the customInstructions an enum property named 'mode' that by default is "all" and could be also DisableAutonomous/ManualOnly (or other name you suggest)
Is this possible?
or we can keep it as is? as this is warning :)

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/examples/Operations_List.json
yashmuel: no, just the standard read/write/delete." If that's correct, this suggestion is already satisfied and no code change is needed

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
ravimeda: Sounds good. Moving the autonomy toggle onto `OperationSettings` as an enum is actually cleaner than either of my original suggestions: it gives each operation a single, positively-framed setting (which addresses the inverted-naming warning) and lets us drop `disableAutonomousOperations` and the `AutonomousOperation` union entirely (which also closes my earlier extensible-enum finding).

Two small refinements on top of your proposal:

- Use PascalCase wire values (`"Auto"`, `"ManualOnly"`) to match the existing `AutonomousOperation` union style and Azure enum conventions, not `"all"`.
- Make it an **extensible** union (include `string`) so new modes can ship later without an SDK breaking change.

```tsp
/** Execution mode for a specific agent operation. */
union OperationMode {
  string,

  /** The operation runs autonomously. */
  auto: "Auto",

  /** The operation runs only when triggered manually; autonomous execution is disabled. */
  manualOnly: "ManualOnly",
}

model OperationSettings {
  /** Execution mode for this operation. Defaults to `Auto` when not specified. */
  mode?: OperationMode;

  /** Custom instructions that guide the agent's behavior for this operation. */
  @maxLength(4096)
  customInstructions?: string;
}
```

And drop `disableAutonomousOperations?` (and `AutonomousOperation`) from `ObservabilityAgentProperties`, since `OperationSettings.mode` supersedes them. Member names are flexible (e.g. `Autonomous`/`Manual`, or adding a `Disabled` mode). The non-negotiables are: extensible union, PascalCase values, and removing the parallel parent-level property. Safe to do now since this is still `2026-05-01-preview`.

ravimeda (approved):

ravimeda:
Please remove all `_tmp_js_*` files #Resolved

yashmuel:
> Please remove all `_tmp_js_*` files

removed. #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,133 @@
...
+    @maxLength(8192)
+    instructions?: string;
+  }
+
+  /**
+   * Updatable properties of an observability agent.
+   */
+  model ObservabilityAgentPropertiesUpdate {
+    /** Resource ID of the Azure Monitor workspace where this agent stores the issues it produces. */
+    monitoringAccountId?: Azure.Core.armResourceIdentifier<[
```
ravimeda, line 80: **[NEW] 🟡 Warning** **[Mutability / lifecycle]**

`monitoringAccountId` is required at create (line 20) and currently optional/mutable via PATCH (line 93). It identifies the Azure Monitor workspace this agent writes its issues to — allowing PATCH to silently re-target this resource changes the data-sink for an existing tracked resource, which is rarely the intended UX.

**Suggested fix (one of):**
- If immutable post-create: mark with `@visibility(Lifecycle.Read, Lifecycle.Create)` on `ObservabilityAgentProperties.monitoringAccountId` and remove it from `ObservabilityAgentPropertiesUpdate` entirely.
- If intentionally mutable: please document the lifecycle/data-migration semantics (what happens to in-flight issues, are they migrated, etc.) so SDK/UX consumers can reason about it.

**Classification:** `[NEW]` — new resource type in this PR; no previous version exists.

 #Resolved

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,133 @@
...
+  model OperationEntry {
+    /** Type of operation. */
+    type: OperationType;
+
+    /** Execution mode for this operation. Defaults to Auto when not specified. */
+    mode?: OperationMode;
+
+    /** Custom instructions that guide the agent's behavior for this operation. */
+    @maxLength(8192)
+    instructions?: string;
```
ravimeda, line 72: **[NEW] 💡 Suggestion** **[Validation]**

`instructions?: string` has `@maxLength(8192)` (good) but no `@minLength`. Consider `@minLength(1)` so callers cannot send an empty string, which is almost certainly not a meaningful "custom instruction".

`	sp
@minLength(1)
@maxLength(8192)
instructions?: string;
`

**Classification:** `[NEW]` — new resource type in this PR; no previous version exists.

 #Resolved

yashmuel: it should be optional, we are ok with null or empty string.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
yashmuel: thanks for pointing this issue.
added to the docs the behavior when changing this (no migration)

ravimeda (approved):

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,133 @@
...
+
+    /** Configuration overrides for agent operations. When omitted, default behavior applies for all operations. */
+    @Azure.ResourceManager.identifiers(#["type"])
+    operations?: OperationEntry[];
+  }
+
+  /**
+   * Enabled state for an observability agent.
+   */
+  union EnabledState {
```
razvanbadea-msft: >EnabledState [](http://example.com/codeflow?start=8&length=12)

the name shoud be more general, for example in case there is a third option like Suspended. Maybe rename it like agentState?

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Get.json
```diff
@@ -0,0 +1,57 @@
...
+          "monitoringAccountId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myResourceGroup/providers/Microsoft.Monitor/accounts/myAzureMonitorWorkspace",
+          "enabled": "Enabled",
+          "operations": [
+            {
+              "type": "IssueCreation",
+              "mode": "Auto",
+              "instructions": "use includeAlertsFromGlobalRules"
+            },
+            {
+              "type": "Investigate",
```
razvanbadea-msft: **[NEW] 🟠 Warning** **[[EX-PAYLOAD](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#22-example-file-validation-ex-)]** [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Get.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Get.json#L40) — line 40 (+ 5 sibling source locations across 3 more example files, plus mirrored copies under `preview/2026-05-01-preview/examples/`)

**Issue:** The example uses `"type": "Investigate"` inside an `operations` entry, but `OperationType` (defined in [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L49-L58) lines 49-58 and emitted in [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/preview/2026-05-01-preview/agents.json) lines 609-631) declares only `"IssueCreation"` and `"Investigation"`. Because the enum is emitted with `modelAsString: true` (extensible), `Swagger ModelValidation` does not error on the unknown string and CI is green — so this is not a build-blocking issue. However, the example still ships as the canonical demonstration in [learn.microsoft.com](https://learn.microsoft.com) reference docs and in SDK sample generators, where it will mislead customers into using a value the service backend almost certainly does not handle. `Investigate` reads as a typo for `Investigation` (the sibling member `IssueCreation` is likewise a noun). Same typo also appears in [ObservabilityAgents_ListByResourceGroup.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_ListByResourceGroup.json#L41) (line 41), [ObservabilityAgents_ListBySubscription.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_ListBySubscription.json#L40) (line 40), and [ObservabilityAgents_Update.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Update.json#L29-L67) (lines 29 and 67); the mirrored copies under `preview/2026-05-01-preview/examples/` will regenerate from source.

**Fix:** Search-replace `"type": "Investigate"` → `"type": "Investigation"` in the four TypeSpec source example files (Get, ListByResourceGroup, ListBySubscription, Update) under `examples/2026-05-01-preview/`. `CreateOrUpdate.json` already uses the correct value. The emitter regenerates the `preview/2026-05-01-preview/examples/` mirrors on the next TypeSpec build — do not hand-edit those.

**Note:** Severity downgraded from Blocking → Warning by reviewer after PR checks were verified green. The original finding incorrectly asserted "validators will reject"; in fact `modelAsString: true` makes the enum extensible at the validator level, so CI is clean. The defect is still real (docs + SDK samples will mislead customers) but is not build-blocking.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/agents/agents.tsp
```diff
@@ -0,0 +1,58 @@
...
+  interface ObservabilityAgents {
+    /**
+     * Returns the specified observability agent.
+     */
+    get is ArmResourceRead<ObservabilityAgentResource>;
+
+    /**
+     * Creates or updates an observability agent.
+     */
+    createOrUpdate is ArmResourceCreateOrReplaceSync<ObservabilityAgentResource>;
```
razvanbadea-msft, line 23: **[NEW] 🔵 Suggestion** **Confirm intent: sync ARM operations with `provisioningState` declared** [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/agents/agents.tsp](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/agents/agents.tsp#L23) — line 23

**Issue:** `observabilityAgents` declares `provisioningState?: Azure.ResourceManager.ResourceProvisioningState` ([specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L14-L17) lines 14-17) but uses the synchronous ARM operation templates (`ArmResourceCreateOrReplaceSync` at line 23, `ArmCustomPatchSync` at lines 28-31, `ArmResourceDeleteSync` at line 36; emitted swagger has no `x-ms-long-running-operation` and no `202` response on PUT/PATCH/DELETE). This is **not** a rule violation — RPC-Async-V1-02 says async PUT/PATCH **must have** `provisioningState`, but the converse ("sync must not have it") is not in the contract; ARM treats a populated `provisioningState` on a sync response as a normal read-only property. However, on a purely synchronous surface the property can only ever be observably read as `Succeeded`, so its customer-facing value is currently zero. Worth a brief confirmation with the service team about intent.

**Fix (pick one based on intent):**
- **(a)** Operations are truly synchronous and will stay that way → remove `provisioningState` from `ObservabilityAgentProperties` to reduce the visible API surface.
- **(b)** Operations are temporarily sync and will become LRO before GA → keep `provisioningState` as forward-compat scaffolding; consider a comment in the TypeSpec noting the intent.
- **(c)** Operations should already be LRO (e.g., create/update of an agent that binds to a monitoring account and configures policies takes >1s at p99) → switch to `ArmResourceCreateOrReplaceAsync`, `ArmResourcePatchAsync` (or `ArmCustomPatchAsync<...>`), and `ArmResourceDeleteAsync`.

**Note:** Severity downgraded from Warning → Suggestion by reviewer after rule re-read; the originally-cited RPC-Async-V1-02 is one-directional (async⇒must-have) and does not forbid the sync+provisioningState shape. Rule citation removed. The concern stands as a stylistic / intent-clarification ask rather than a contract violation.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
```diff
@@ -0,0 +1,133 @@
...
+
+    string,
+  }
+
+  /**
+   * Configuration for a specific agent operation.
+   */
+  model OperationEntry {
+    /** Type of operation. */
+    type: OperationType;
```
razvanbadea-msft, line 65: **[NEW] 🔵 Suggestion** **Naming consistency: `OperationEntry.type` → `operationType`** [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L78) — line 78

**Issue:** `OperationEntry.type: OperationType` uses the bare name `type` for an array-item discriminator-like field. While not a strict rule violation (the reserved-name guidance in Section 13 targets direct children of the resource `properties` bag, not nested sub-model items), generic `type` is less self-documenting than the enum it references (`OperationType`), and SDK code generators in some languages may surface naming friction when the property shares a name with the resource-envelope `type` field. Recommend renaming for clarity.

**Fix:** In [specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L78) change `type: OperationType;` → `operationType: OperationType;` (line 78) and update the two `@Azure.ResourceManager.identifiers(#["type"])` references on `operations` array properties ([line 30](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L30) and [line 103](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp#L103)) to `@Azure.ResourceManager.identifiers(#["operationType"])`. Also update every `"type": "..."` inside the `operations` arrays in the four example files to `"operationType"`. This change is optional and can be deferred to a future revision if the team prefers to keep the current name.

**Note:** Severity downgraded from Warning → Suggestion per critic (iteration 1); the cited Section 13 of openapi-review.instructions.md applies only to direct children of the `properties` bag, not to nested sub-model items, so the rule citation was dropped and the naming concern stands on its own merit.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_CreateOrUpdate.json
```diff
@@ -0,0 +1,130 @@
+{
```
razvanbadea-msft, line 1: **[NEW] 🔵 Suggestion** **[[EX-COVERAGE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#22-example-file-validation-ex-)]** [examples/2026-05-01-preview/ObservabilityAgents_CreateOrUpdate.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_CreateOrUpdate.json) and [ObservabilityAgents_Update.json](specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Update.json)

**Issue:** Only one example exists per operation, and both populate every optional property (a "maximum-set" example). EX-COVERAGE suggests also providing a "minimum-set" example for PUT/PATCH that only sets the required fields (`monitoringAccountId` is the only required property), so SDK sample generation and customer-facing docs can demonstrate the simplest valid payload.

**Fix:** Add `ObservabilityAgents_CreateOrUpdate_MinimumSet.json` and `ObservabilityAgents_Update_MinimumSet.json` under `examples/2026-05-01-preview/`, then wire them in via TypeSpec `@example` decorators or by referencing them in the operation definitions.

razvanbadea-msft (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`a04fefc`](https://github.com/Azure/azure-rest-api-specs/pull/42602/commits/a04fefc987d536a60ce750c3fa020429758c47e5). See inline comments for findings 1-4.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
yashmuel: We've already deployed our code based on type, we don't want to change to operationType in this stage.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/agents/agents.tsp
yashmuel: @razvanbadea-msft 
I've tried to fix with option a, but now typespec validation failing on missing provisioningState in ARM tracked resource. how can i solve this?

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/examples/2026-05-01-preview/ObservabilityAgents_Get.json
sandipsh: Thanks for addressing this! The violation flagged here is no longer present in the latest changes. Resolving this thread.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/agents/agents.tsp
sandipsh: Thanks for addressing this! The violation flagged here is no longer present in the latest changes. Resolving this thread.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
sandipsh: Thanks for addressing this! The violation flagged here is no longer present in the latest changes. Resolving this thread.

sandipsh:
TypeSpec Validation failure is caused by this PR's new Agents TypeSpec model.

Failure details from run:
https://github.com/Azure/azure-rest-api-specs/actions/runs/27085964605/job/79940382827

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
Location: line 14 (model ObservabilityAgentProperties)
Error: arm-resource-provisioning-state requires a provisioningState property in the resource properties model with known values Succeeded, Failed, and Canceled.

Please update the model to include a compliant provisioningState and rerun TypeSpec validation:
npx tsv specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
sandipsh: Addressed in the latest PR head. The type name is now AgentState (replacing EnabledState) in specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp, so this naming concern is resolved.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
sandipsh: Thanks for the clarification. Agreed: this is a non-blocking suggestion only, and keeping the field optional (including allowing null/empty) is acceptable for this preview contract. Marking this thread resolved.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Agents/typespec/models.tsp
sandipsh: Understood. Keeping `type` is acceptable for this iteration since this was a naming suggestion (non-blocking), not a guideline violation. No change required for merge on this point.

sandipsh:
ARM review update on latest iteration:

- Re-reviewed current head commit e1dc7cffcf2397a7e00cdaefebcd2d274af708be.
- Addressed open thread actions:
  - Resolved the open non-blocking minLength suggestion thread after author clarification.
  - Replied on remaining human-reviewer suggestion threads with merge guidance; no blocking change requested there.

Validation status for this changeset (blocking checks requested):
- Swagger LintDiff: success
- Swagger ModelValidation: success
- Swagger SemanticValidation: success
- TypeSpec Validation: success
- Swagger Avocado: no actionable failures for this changeset (no non-MULTIPLE_API_VERSION blocker observed)

No new ARM blocking issues identified in the latest iteration.

sidanmor (approved):
Ok

sandipsh (approved):

PR #42619, opened 2026-04-23
[Storage] `HttpRange` replacement, `Etag` model replacement

vincenttran-msft:


specification/storage/Microsoft.BlobStorage/routes.tsp
```diff
@@ -701,7 +701,10 @@ interface Blob {
     ...EncryptionKeySha256Parameter,
     ...EncryptionAlgorithmParameter,
     ...IfTagsParameter,
-    ...ConditionalRequestHeaders,
+    ...IfModifiedSinceParameter,
+    ...IfUnmodifiedSinceParameter,
+    ...IfNoneMatchParameter,
+    ...IfMatchParameter,
```
catalinaperalta, line 707: Are the headers in the Azure.Core.ConditionalRequestHeaders not typed with etag? That's the reason for this update?

catalinaperalta (approved):
Changes seem benign to me, I looked through client.tsp and the languages seem to be changing the conditional header types back to string from etag anyways for other instances. cc @l0lawrence @jeremymeng @gracewilcox @jorgerangel-msft

specification/storage/Microsoft.BlobStorage/routes.tsp
catalinaperalta: I'll open an issue to get this fixed in the azure.core types

catalinaperalta (changes requested):
@vincenttran-msft please fix the typespec validation error before merging into the base branch

catalinaperalta (approved):

PR #42670, opened 2026-04-27
Mitigate Java SDK breaking changes for NetworkCloud

weidongxu-microsoft:
Mitigate Java SDK breaking changes for networkcloud TypeSpec migration.

### Changes
- Rename \zureUserName\ back to \zureUsername\ for \KeySetUser\ and \KeySetUserStatus\ models in Java SDK

### Remaining acceptable breaks
- \Float\ → \Double\ for \percentComplete()\ in \OperationStatusResult\
- \alidate()\ removed from many models (expected for TypeSpec)
- \List\ model removals (expected for TypeSpec)
- \ClusterCapacity\ setters removed (read-only properties)
- \serviceClient()\ return type change (expected)

### DIFF
--- a/specification/networkcloud/NetworkCloud.Management/storageappliances/StorageAppliance_models.tsp
+++ b/specification/networkcloud/NetworkCloud.Management/storageappliances/StorageAppliance_models.tsp
@@ -130,16 +130,16 @@ model StorageApplianceProperties {
  */
 model StorageApplianceSpec {
   /**
+   * The credentials of the administrative interface on this storage appliance.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  administratorCredentials: AdministrativeCredentials;
+
+  /**
    * The resource ID of the rack where this storage appliance resides.
    */
   @visibility(Lifecycle.Read, Lifecycle.Create)
   rackId: Azure.Core.armResourceIdentifier;
-
-  /**
-   * The SKU for the storage appliance.
-   */
-  @visibility(Lifecycle.Read, Lifecycle.Create)
-  storageApplianceSkuId: string;
 
   /**
    * The slot the storage appliance is in the rack based on the BOM configuration.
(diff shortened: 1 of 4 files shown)


specification/networkcloud/NetworkCloud.Management/storageappliances/StorageAppliance_models.tsp
```diff
@@ -130,16 +130,16 @@ model StorageApplianceProperties {
  */
 model StorageApplianceSpec {
```
weidongxu-microsoft, line 131: Fix order of required properties (optional properties is not touched).

Swagger is (e.g. `administratorCredentials` is the 1st property)
https://github.com/Azure/azure-rest-api-specs/blob/313ca227e879d1b5f6f59d9f68442a6c28fd3279/specification/networkcloud/resource-manager/Microsoft.NetworkCloud/preview/2026-01-01-preview/networkcloud.json#L14890-L14908

Order change like this can affect the parameter order in class constructor (or some API related to this ctor) in some SDKs.

pshao25 (approved):

specification/networkcloud/NetworkCloud.Management/storageappliances/StorageAppliance_models.tsp
msyyc: FYI: no influence for python so this PR is OK for python.

weidongxu-microsoft: @bryan-strassner

Please let me know if service is OK with this fix -- here we just make TypeSpec definition being same as the Swagger definition before the conversion

weidongxu-microsoft: @bryan-strassner

I will merge the PR tomorrow, if there is no feedback from service dev.

bryan-strassner: I need to look into why I have stopped receiving notification for mentions... I missed all this.

PR #42672, opened 2026-04-27
Add code download endpoints for hosted code-based agents

sagunase:
## What

Adds two preview GET endpoints so callers can download the previously-uploaded code zip for a hosted code-based agent.

| Operation | Route | Scope |
|---|---|---|
| `downloadAgentVersionCode` | `GET /agents/{agent_name}/versions/{agent_version}/code:download` | Specific version |
| `downloadAgentCode` | `GET /agents/{agent_name}/code:download` | Latest version of the agent |

Both are gated under the `code_agents_v1_preview` opt-in and excluded from `v1` (`@removed(Versions.v1)`), matching `createAgentFromCode` / `updateAgentFromCode` / `createAgentVersionFromCode`.

## Why

Today callers can upload code (`createAgentFromCode`, `updateAgentFromCode`, `createAgentVersionFromCode`) but there is no way to read it back. Downstream tooling ΓÇö and the agent runtime itself when rehydrating onto a new host ΓÇö need to retrieve the exact zip that was uploaded.

This is also a direct unblocker for the **Agent Builder** team, who are building a hosted agent that creates *other* hosted agents from a prompt. Because the generated agent is itself hosted (its code lives on the service, not on the user's machine), the user has no local copy of what was produced for them. They need a way to inspect the code the builder agent generated on their behalf, which is exactly what these endpoints provide.

## Sample requests

```
GET /agents/my-agent/versions/3/code:download
Accept: application/zip
```
```
GET /agents/my-agent/code:download
Accept: application/zip
```

## Responses

| Status | Body | Notes |
|---|---|---|
| `200 OK` | Raw zip bytes (`application/zip`) | Integrity verified by client against existing `content_hash` on the version object |
| `404 Not Found` | Standard error | Agent or version does not exist, or version has no uploaded code |
| Other 4xx/5xx | Standard Azure error envelope | Inherited from `FoundryDataPlanePreviewOperation` |

No SHA-256 response header is added; `content_hash` already serves that purpose.


dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -338,6 +338,51 @@ interface Agents {
...
+
+      /** The version of the agent whose code zip should be downloaded. */
+      @path agent_version: string;
+    },
+    bytes
+  >;
+
+  /**
+   * Download the code zip for the latest version of a code-based hosted agent.
+   * Returns the previously-uploaded zip as a binary stream (`application/octet-stream`).
```
johanste: Why not indicate that this is in fact a zip file in teh content type?

```suggestion
   * Returns the previously-uploaded zip as a binary stream (`application/zip`).
```

sagunase: Good call -- switched the response to a `DownloadAgentCodeResponse` model that sets `Content-Type: application/zip` on the wire (mirroring `DownloadSkillResponse` in `skills/models.tsp`). Doc strings and example response headers updated, and OpenAPI regenerated. Streaming is preserved -- `@body body: bytes` still emits as `Iterable[bytes]` / `Stream` in the SDKs regardless of the content-type label.

johanste (approved):
I see changes in the generated openapi docs that I can't trace back to the changes in typespec. Please verify if they need to be regenerated before merging.

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -338,6 +338,51 @@ interface Agents {
     AgentVersionObject
   >;
 
+  /**
+   * Download the code zip for a specific version of a code-based hosted agent.
+   * Returns the previously-uploaded zip (`application/zip`).
+   * The SHA-256 digest of the returned bytes matches the `content_hash` on the agent version's `code_configuration`.
+   */
+  @get
+  @route("/agents/{agent_name}/versions/{agent_version}/code/content")
```
glecaros: the route here (and for the non-versioned endpoint) is not compliant with the naming guidelines (two nouns without resource in between). Moreover, given there are no other `/code/` siblings, it's not really disambiguating.
I would recommend using `/code:download` to make the download action explicit, or just `/code` (I favor the former)

trrwilson: Agreed -- `/code:download` seems best aligned with the existing patterns.

sagunase: Done in 5589d361 -- both routes renamed to `/code:download` to match the `/skills/{name}:download` precedent.

PR #42692, opened 2026-04-27
[SDK Validation] Use azsdk-cli and onboard Rust

raych1:
This pull request introduces support for Rust SDK generation using the new `azsdk-cli` tool and refactors the SDK generation pipeline to select the appropriate tool based on the language and environment. The changes add detection and orchestration logic for `azsdk-cli`, implement adapters for its output, and ensure the pipeline can handle both the legacy and new generation paths.

**Key changes:**

Tool selection and orchestration:

* Added logic to select between `azsdk-cli` and the legacy `spec-gen-sdk` tool depending on the SDK language and environment, with fallback and error handling if the required tool is unavailable.
* Refactored the main generation flow to handle both tools, including a new `runAzsdkGeneration` function for orchestrating the `azsdk-cli` workflow (check emitter, generate, build, pack, report). 

Rust SDK support:

* Added Rust as a supported SDK language in the shared types and schemas. 
* Updated pipeline templates to install the `azsdk-cli` when needed. 

Adapters and helpers for azsdk-cli:

* Implemented an adapter module to parse `azsdk-cli` output, extract meaningful results, and convert them into the existing `ExecutionReport` format. 
* Added helper functions to prepare `azsdk-cli` command arguments, resolve paths, and check for CLI availability.


eng/tools/spec-gen-sdk-runner/src/commands.ts
```diff
@@ -10,13 +17,141 @@ import {
...
+      executionReport: buildExecutionReport(undefined, undefined, emitterCheck),
+      statusCode: 0,
+    };
+  }
+
+  // Step 2: Install tsp-client dependencies (needed because git clean -fdx wipes node_modules)
+  const tspClientDir = path.join(commandInput.localSdkRepoPath, "eng", "common", "tsp-client");
+  if (fs.existsSync(path.join(tspClientDir, "package.json"))) {
+    logMessage(`Installing tsp-client dependencies at ${tspClientDir}`, LogLevel.Info);
+    await execAsync("npm ci", { cwd: tspClientDir });
```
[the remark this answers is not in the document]
raych1: A single PR with multiple service folder is lower priority, and will be improved later.

eng/tools/spec-gen-sdk-runner/src/azsdk-adapter.ts
```diff
@@ -0,0 +1,158 @@
...
+  response_errors?: string[];
+  duration?: number;
+}
+
+/**
+ * Parses JSON output from an azsdk-cli command.
+ * The CLI emits log lines before the JSON payload. The JSON object is always
+ * the last block in the output, so we search backwards for the opening brace.
+ */
+export function parseAzsdkResponse<T>(output: string): T {
```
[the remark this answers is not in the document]
raych1: Added tests.

raych1:
@heaths can you review this PR about enabling Rust validation in spec repo? Here is a test PR against the keyvault spec: https://github.com/Azure/azure-rest-api-specs/pull/42704.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -208,6 +208,20 @@ stages:
         inputs:
           versionSpec: $(PythonVersion)
 
+      - script: |
+          set -euo pipefail
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          echo "##vso[task.prependpath]$HOME/.cargo/bin"
+        displayName: 'Install Rust toolchain'
+        condition: and(succeeded(), contains(variables['SdkRepoName'], 'rust'))
```
weshaggard: I'm not a fan of adding language specific conditions in this pipeline. Ideally it would be in the sdk init or cli verify setup command.

heaths:
Not a blocker, but something maybe to look into:

image

It published more crates (dependencies) than necessary and yet, somehow, it didn't publish all the dependencies. It's missing `typespec`. Still, only want `.crate` files for crates for the services themselves since everything after GA would use a version already on crates.io anyway.

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
heaths: You really don't need to anyway. As long as the agents have newer than 1.88 (pretty old - almost a year, IIRC) they should work out of the box. Just run `rustup install` in our repo root, which I think your other changes elsewhere already did, didn't they?

heaths (changes requested):
I agree with @weshaggard and added a way you could handle that. As for the runner changes, I don't really have enough experience to review that particular code and it seems to have little or nothing to do with Rust.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
```diff
@@ -500,3 +501,15 @@ export function resolvePackagePath(outputDir: string, sdkRepoPath: string): stri
...
+
+/**
+ * Installs language-specific toolchains required before SDK generation.
+ * Currently only Rust needs this — runs `rustup install` in the SDK repo root
+ * to pick up the version from rust-toolchain.toml.
+ */
+export async function installLanguageToolchain(commandInput: SpecGenSdkCmdInput): Promise<void> {
+  if (commandInput.sdkLanguage === "azure-sdk-for-rust") {
+    logMessage(`Installing Rust toolchain from SDK repo`, LogLevel.Info);
+    await execAsync("rustup install", { cwd: commandInput.localSdkRepoPath });
```
heaths, line 513: Is the assumption that all machines will already have `rustup` in the path? If targeting our Azure Pipelines pool or GitHub Artifacts it will be, but I wasn't sure if this was also used on devs' machines. If not, no worries.

heaths (approved):

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
raych1: For AZSDK agent, the 'setup' CLI prompts user to download and install 'rustup' first from 'https://rust-lang.org/tools/install' system tool. Since build agent would always include 'rustup', I've updated the code.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
raych1: Yes, it's mostly used in pipeline scenarios. User is supposed to use 'AZSDK' agent in local scenarios.

raych1:
> Not a blocker, but something maybe to look into:
> 
> image
> It published more crates (dependencies) than necessary and yet, somehow, it didn't publish all the dependencies. It's missing `typespec`. Still, only want `.crate` files for crates for the services themselves since everything after GA would use a version already on crates.io anyway.

As synced offline, we might emit a markdown summary telling users exactly what to put into their app using https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html for early integration scenarios.

heaths (approved):

PR #42696, opened 2026-04-28
[Queue] Client customizations

catalinaperalta:


specification/storage/Microsoft.QueueStorage/models.tsp
catalinaperalta, line 305: why were these properties removed?

catalinaperalta: Same question for the properties that were removed above

catalinaperalta, line 305: @jeremymeng @l0lawrence @jorgerangel-msft @gracewilcox does anybody know why these properties were removed and why?

catalinaperalta:
@jorgerangel-msft seems the typespec validation check is reporting a potential conflict for c# in this spec due to the name "clear". Is it ok to suppress that warning?

specification/storage/Microsoft.QueueStorage/models.tsp
jeremymeng: I'd thought that copySource* don't really apply to Queue so I removed them, but I think it would be fine to restore them back too and not affecting anything, since they are optional

catalinaperalta: @jalauzon-msft / @amnguye what's the actual rest api schema for error responses for queues?

amnguye: I believe these headers only apply to Blob, not Queues. Queues doesn't have a Copy API so, I believe these error headers do not apply.

catalinaperalta: Thanks for the info Amanda! :) Will leave as is then

specification/storage/Microsoft.QueueStorage/models.tsp
```diff
@@ -641,7 +624,7 @@ alias TimeNextVisibleResponseHeaderPrivate = {
 /** The metadata headers. */
 alias MetadataHeaders = {
   /** The metadata headers. */
-  @alternateType(Record<string>, "rust")
+  @alternateType(Record<string>, "rust,go")
```
jalauzon-msft: I know this was existing but aren't alternateType customizations supposed to go in `client.tsp`?

vincenttran-msft (approved):
Generates and compiles cleanly for Rust SDK, thanks!!

specification/storage/Microsoft.QueueStorage/models.tsp
catalinaperalta: Yes they should be, I left it since it was existing, but happy to clean it up some more

catalinaperalta: Done!

lmazuel (approved):

PR #42771, opened 2026-04-29
[Storage] `azure_storage_blob` APIView Feedback TypeSpec Changes

vincenttran-msft:
Outstanding emitter blockers:
https://github.com/Azure/typespec-rust/issues/939
https://github.com/Azure/typespec-rust/issues/940
https://github.com/Azure/typespec-rust/issues/953


specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -481,6 +481,7 @@ model FilterBlobResponse {
   /** The blob segment. */
   @pageItems
   @Xml.name("Blobs")
+  // TODO: We want this to be blob_items but that causes an emitter conflict. Will standup in a separate PR.
   blobs: FilterBlobItem[];
```
vincenttran-msft: This is the collision point. We are keeping it at blobs right now so it won't break generation, but ideally we want this to be consistent with the other listing operation, so it should be `blob_items` but this is currently causing a collision.

specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -2575,9 +2551,15 @@ model ListBlobsHierarchySegmentResponse {
...
+  /** The blob items. */
+  @pageItems
   @Xml.name("Blobs")
-  segment: BlobHierarchyListSegment;
+  blobItems: BlobItem[];
+
+  /** The blob prefixes. */
+  @Xml.name("BlobPrefix")
+  @Xml.unwrapped
+  blobPrefixes?: BlobPrefix[];
```
catalinaperalta: This is changing the schema of the response, looks like BlobPrefix is under the `<Blobs>` tag, it's a sibling to the `<Blob>` tag. Recording example: https://github.com/Azure/azure-sdk-assets/blob/89c4f2856edeb9a136e2f8439fbc6eeffb6eaf66/python/sdk/storage/azure-storage-blob/tests/recordings/test_container.pyTestStorageContainertest_walk_blobs_start_end.json#L258

Image

vincenttran-msft: Good callout, reverted!

specification/storage/Microsoft.BlobStorage/client.tsp
```diff
@@ -137,6 +150,8 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 @@access(Storage.Blob.BlockBlob.upload, Access.internal, "rust");
 @@clientName(Storage.Blob.Blob.download, "download_internal", "rust");
 @@access(Storage.Blob.Blob.download, Access.internal, "rust");
+@@clientName(Storage.Blob.Service.findBlobsByTags, "findBlobsByTags", "rust");
+@@clientName(Storage.Blob.Container.findBlobsByTags, "findBlobsByTags", "rust");
```
catalinaperalta, line 154: why this renaming to exactly the same name?

catalinaperalta (approved):

specification/storage/Microsoft.BlobStorage/client.tsp
vincenttran-msft: Without it, it shoves in a `list_` prefix which is not desirable on these specific APIs

jalauzon-msft (approved):

PR #42777, opened 2026-04-30
Rename ContainerInstance to Workbench and add read-only webEndpoint property

jyo219:
Renames the compute type discriminator from ContainerInstance to Workbench in Compute.tsp, models.tsp, and all example JSON files. Adds a read-only webEndpoint string property to WorkbenchComputeProperties. Regenerated swagger JSON via tsp compile.


jyo219:

PR #42804, opened 2026-05-01
Add SandboxGroup API to Microsoft.ContainerInstance TypeSpec

iyyappam:
This introduces a new SandboxGroup resource with full CRUD operations and a connect action for the Microsoft.ContainerInstance service. The new APIs are part of the 2026-06-01-preview API version.

New APIs:
- SandboxGroups_ListBySubscription
- SandboxGroups_ListByResourceGroup
- SandboxGroups_Get
- SandboxGroups_CreateOrUpdate (async)
- SandboxGroups_Update (async)
- SandboxGroups_Delete (async)
- SandboxGroups_Connect

New models:
- SandboxGroup (TrackedResource)
- SandboxGroupProperties
- SandboxGroupAccessToken
- NetworkProfile
- SubnetReference
- ProvisioningState (enum)

# Choose a PR Template


iyyappam:
We acknowledge that this PR publishes the API to customers and follows ARM API best practices.

iyyappam:
This is intended for available to customers and it follow ARM best practices.

sandipsh:
please address swagger lintdiff errors

mikeharder:
breaking changes turned out to be a bug in the tool, but it may be a "won't fix" from us:
 
- https://github.com/Azure/openapi-diff/issues/479
 
workaround in your spec would be, to rename your TSP model from `Resource` to anything else that wouldn't conflict with a name in `common-types/types.json`.

saprakas:
As mikeharder pointed above, the 'Breaking Change(Cross-Version)' is due to a bug in tool https://github.com/Azure/openapi-diff/issues/479. Any attempt to workaround this causes more breaking changes violation. 

Swagger LintDiff error is for ProvisioningStateMustBeReadOnly rule. This is an existing error from previous release and fixing this will cause breaking changes again.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/**
+ * A reference to a subnet resource.
+ */
+model SubnetReference {
+  /**
+   * The ARM resource ID of the subnet. The caller must have `Microsoft.Network/virtualNetworks/subnets/join/action` permission on this subnet (enforced via a linked access check at create/update time).
+   */
+  id: string;
```
gary-x-li: **Issue:** `SubnetReference.id` is typed as plain `string`, but the doc comment says _"The ARM resource ID of the subnet."_ ARM resource ID properties MUST use `armResourceIdentifier` to enable ARM validation, linked access checks, and rich SDK typing.

   **Fix:**
   ```tsp
   id: Azure.ResourceManager.armResourceIdentifier<[{ type: "Microsoft.Network/virtualNetworks/subnets" }]>;
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * The network profile of the SandboxGroup.
+   */
+  networkProfile?: SandboxGroupNetworkProfile;
+
+  /**
+   * The ARM resource ID of the management resource group associated with this SandboxGroup.
+   */
+  @visibility(Lifecycle.Read)
+  managementResourceGroupId?: string;
```
gary-x-li: **Issue:** `SandboxGroupProperties.managementResourceGroupId` is typed as plain `string`, but it represents an ARM resource group ID. Same rule as above — ARM resource ID properties MUST use `armResourceIdentifier`.

   **Fix:**
   ```tsp
   @visibility(Lifecycle.Read)
   managementResourceGroupId?: Azure.ResourceManager.armResourceIdentifier;
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Microsoft.ContainerInstance;
+
+/**
+ * The provisioning state of a resource.
+ */
+union ProvisioningState {
```
gary-x-li: **Issue:** `ProvisioningState` is defined from scratch instead of extending `Azure.ResourceManager.ResourceProvisioningState`. The TypeSpec checklist requires: _"provisioningState extends ResourceProvisioningState for terminal states."_ Additionally, the existing service naming convention uses `{ResourceType}ProvisioningState` (e.g., `ContainerGroupProvisioningState`, `NGroupProvisioningState`), but this PR uses the generic name `ProvisioningState`, which is inconsistent and risks future collision.

   **Fix:** Rename to `SandboxGroupProvisioningState` and extend `ResourceProvisioningState`:
   ```tsp
   union SandboxGroupProvisioningState {
     string,
     ResourceProvisioningState,

     /** The resource is being updated. */
     Updating: "Updating",

     /** The resource is being deleted. */
     Deleting: "Deleting",

     /** The resource provisioning request was accepted but not yet started. */
     Accepted: "Accepted",
   }
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/preview/2026-06-01-preview/examples/SandboxGroupsListBySubscription.json
```diff
@@ -0,0 +1,36 @@
...
+  "parameters": {
+    "api-version": "2026-06-01-preview",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "value": [
+          {
+            "id": "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.ContainerInstance/sandboxGroups/mySandboxGroup1",
```
gary-x-li: **Issue:** The resource `id` in the ListBySubscription example is missing the `/resourceGroups/{rg}` segment:
   ```json
   "id": "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.ContainerInstance/sandboxGroups/mySandboxGroup1"
   ```
   Tracked resources always have a resource group in their fully-qualified ARM resource ID, even when returned from subscription-scope list operations.

   **Fix:**
   ```json
   "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myResourceGroup/providers/Microsoft.ContainerInstance/sandboxGroups/mySandboxGroup1"
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/**
+ * The result of getting an access token for a SandboxGroup.
+ */
+model SandboxGroupAccessToken {
+  /**
+   * The endpoint URL to use with the access token.
+   */
+  endpoint: string;
```
gary-x-li: **Issue:** `SandboxGroupAccessToken.endpoint` is typed as `string` but represents a URL. TypeSpec checklist requires: _"URI properties use `url` scalar type, not plain `string`."_ This generates `"format": "uri"` in swagger.

   **Fix:**
   ```tsp
   endpoint: url;
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+ */
+model SandboxGroup
+  is Azure.ResourceManager.TrackedResource<SandboxGroupProperties> {
+  ...Azure.ResourceManager.ManagedServiceIdentityProperty;
+
+  /**
+   * The name of the SandboxGroup.
+   */
+  @maxLength(63)
+  @minLength(1)
```
gary-x-li, line 98: **Issue:** `@minLength(1)` conflicts with the `@pattern("^[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]$")` which requires at least 2 characters (an alphanumeric start + an alphanumeric end). A 1-character input would pass `minLength` validation but fail the pattern.

   **Fix:** Change `@minLength(1)` to `@minLength(2)`, or update the pattern to allow single-character names: `^[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?$`

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * The status of the last operation.
+   */
+  @visibility(Lifecycle.Read)
+  provisioningState?: ProvisioningState;
+
+  /**
+   * The network profile of the SandboxGroup.
+   */
+  networkProfile?: SandboxGroupNetworkProfile;
```
gary-x-li, line 78: **Issue:** `SandboxGroupProperties.networkProfile` has no visibility constraints, making it implicitly create+update+read. However, the PATCH model (`SandboxGroupTagsUpdate`) does NOT include `networkProfile`, which means it cannot be updated via PATCH. If `networkProfile` is intended to be immutable after creation, add an immutability annotation. If it should be patchable, include it in the PATCH model.

   **Fix (if immutable):**
   ```tsp
   networkProfile?: SandboxGroupNetworkProfile;
   ```

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * The UTC date and time at which the access token expires.
+   */
+  notAfter: utcDateTime;
+}
+
+/**
+ * The response of a SandboxGroup list operation.
+ */
+model SandboxGroupListResult {
```
gary-x-li: **Issue:** `SandboxGroupListResult` is manually defined but never referenced by any operation. The `ArmListBySubscription` and `ArmResourceListByParent` templates auto-generate list result types. This model will be orphaned in the generated output.

   **Fix:** Remove the `SandboxGroupListResult` model definition entirely.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * The link to the next page of items
+   */
+  nextLink?: string;
+}
+
+@armResourceOperations
+@added(Versions.v2026_06_01_preview)
+interface SandboxGroups {
```
gary-x-li, line 142: **Issue:** Error handling is inconsistent across operations. `listBySubscription`, `listByResourceGroup`, `get`, and `connect` specify `Error = CloudError`, but `createOrUpdate`, `update`, and `delete` use the default error type. This should be consistent across all operations for the same resource.

   **Fix:** Either add `Error = CloudError` to all operations, or remove it from all (using the default ARM error type consistently).

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/readme.md
```diff
@@ -257,12 +266,20 @@ suppressions:
...
       - $.definitions.SecretReferenceVolume
   - code: GetCollectionResponseSchema
-    reason: We do not return the instanceView property in our List operation, we just return this 
+    reason: We do not return the instanceView property in our List operation, we just return this
             property for our Get operations. This change has been part of our stable api versions for a couple of years
     from:
       - containerInstance.json
     where:
       - $.paths["/subscriptions/{subscriptionId}/providers/Microsoft.ContainerInstance/containerGroups"]
+  - code: XMS_EXAMPLE_NOTFOUND_ERROR
```
gary-x-li: Can you remove this one and the one below? These are not LintDiff errors.

mikeharder:
> As mikeharder pointed above, the 'Breaking Change(Cross-Version)' is due to a bug in tool [Azure/openapi-diff#479](https://github.com/Azure/openapi-diff/issues/479). Any attempt to workaround this causes more breaking changes violation.
> 
> Swagger LintDiff error is for ProvisioningStateMustBeReadOnly rule. This is an existing error from previous release and fixing this will cause breaking changes again.

These are not the correct resolutions for this spec or PR.
 
For the crash in breaking changes, you need to rename your model `Resource`, because it conflicts with a model of the same name in `common-types`.  Even if this means you need to get the breaking changes approved, it's much better than having a spec that our breaking change tooling can't even analyze.
 
For LintDiff, you need to either:
1. Fix the problem to get your spec in compliance in spite of the breaking changes.
2. Add suppressions to a suppressions.yaml file, with an explanation why it's better to keep things invalid, rather than take a one-time breaking change to get compliant.

I believe most instances of `ProvisioningStateMustBeReadOnly` should be fixed rather than suppressed, but I don't know the specifics of your spec.
 
@markcowl is the expert on TSP and `ProvisioningStateMustBeReadOnly`.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/preview/2026-06-01-preview/examples/SandboxGroupsListBySubscription.json
saprakas: Fixed

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Updated the pattern.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Fixed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/SandboxGroup.tsp
saprakas: Removed.

specification/containerinstance/resource-manager/Microsoft.ContainerInstance/ContainerInstance/readme.md
saprakas: Removed.

saprakas:
Renaming 'Resource' model in the current PR didn't fix the breaking change.

Why we can't fix the rename inside this PR:
- The crash happens on the OLD side of the same-version diff. "OLD" = [origin/main]'s stable/2025-09-01 swagger, which still contains the local 'Resource' definition.

- Renaming Resource to ContainerGroupPatch only on the NEW side (this PR's branch) doesn't help — main is still broken and oad still crashes loading it.

- Renaming Resource → ContainerGroupPatch on BOTH sides (this PR includes the stable rename) does work locally, but appears as a "model removed" breaking change in same-version comparison and additionally is a noisy unrelated change to a published stable API in this preview-feature PR.

psah434:
> Renaming 'Resource' model in the current PR didn't fix the breaking change.
> 
> Why we can't fix the rename inside this PR:
> 
> * The crash happens on the OLD side of the same-version diff. "OLD" = [origin/main]'s stable/2025-09-01 swagger, which still contains the local 'Resource' definition.
> * Renaming Resource to ContainerGroupPatch only on the NEW side (this PR's branch) doesn't help — main is still broken and oad still crashes loading it.
> * Renaming Resource → ContainerGroupPatch on BOTH sides (this PR includes the stable rename) does work locally, but appears as a "model removed" breaking change in same-version comparison and additionally is a noisy unrelated change to a published stable API in this preview-feature PR.

@mikeharder
Is this ok?

> **[mikeharder](/mikeharder) **

saprakas (approved):
All review feedback has been addressed.

jussong-msft (approved):

PR #42832, opened 2026-05-01
Foundry memory v2 (procedural memory, TTL)

bojunehsu:

Add support to enable procedural memory and memory TTL to Foundry Memory.
- `MemoryStoreDefaultOptions.procedural_memory_enabled` 
- `MemoryStoreDefaultOptions.chat_summary_ttl_days`


specification/ai-foundry/data-plane/Foundry/openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml
```diff
@@ -9743,6 +9749,12 @@ components:
           enum:
             - azure_ai_search
           description: The object type, which is always 'azure_ai_search'.
+        name:
```
tolia-msft, line 12260: IIUC this was out-of-sync with models.tsp.
Should we regenerate it first in a separated PR?

tolia-msft (approved):

specification/ai-foundry/data-plane/Foundry/src/memory-stores/models.tsp
tolia-msft, line 278: Do we need to add one here?

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -1301,4 +1278,46 @@ namespace Azure.AI.Projects {
...
+    status: ToolCallStatus;
+
+    @doc("The results returned from the memory search.")
+    memories?: MemoryItem[] | null;
+  }
+
+  @removed(Versions.v1)
+  @doc("A memory command tool call.")
+  model MemoryCommandToolCall extends OpenAI.OutputItem {
+    type: "memory_command_call";
```
glecaros: these values probably need to be added to the _AgentItemType union (`"memory_search_call"` is already present, but `"memory_command_call"` and `"memory_command_call_output"` aren't)

glecaros: Also, if these are preview, my understanding is that their value should include  `preview` (e.g `memory_command_preview_call`)

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -1301,4 +1278,46 @@ namespace Azure.AI.Projects {
     @doc("The status of the tool call.")
     status: ToolCallStatus;
   }
+
+  // ==============================================================================
+  // Memory Search Tool Call Models
+  // ==============================================================================
+
+  @doc("A memory search tool call.")
+  model MemorySearchToolCall extends OpenAI.OutputItem {
```
glecaros, line 1243: if these are preview, they need the `@extension("x-ms-foundry-meta", MemoryStoreRequiredPreviews)` annotation.

specification/ai-foundry/data-plane/Foundry/src/memory-stores/models.tsp
```diff
@@ -105,6 +106,14 @@ model MemoryStoreDefaultOptions {
...
   @doc("Whether to enable chat summary extraction and storage. Default is true.")
   chat_summary_enabled: boolean = true;
+
+  @removed(Versions.v1)
+  @doc("Whether to enable procedural memory extraction and storage. Default is true.")
+  procedural_memory_enabled?: boolean = true;
+
+  @removed(Versions.v1)
+  @doc("The default time-to-live for memories in seconds.  A value of 0 indicates that memories do not expire.")
+  default_ttl_seconds?: int32 = 0;
```
glecaros:
```suggestion
  @encode(DurationKnownEncoding.seconds, int32)
  default_ttl_seconds?: duration = duration.fromISO("P0DT0S");
```

trrwilson: It looks like even just `P0D` works there (and is equivalent). If desired, it can be even tidier if we want to define `const DurationZero = duration.fromISO("P0D");` elsewhere -- though I wouldn't worry about it if we don't have immediate reuse.

The reason the use of `duration` is valuable here is that generated code will recognize it and apply appropriate language-specific types, like a .NET `TimeSpan` or Java `Duration`.

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
bojunehsu: I am confused... A2AToolCall which has type a2a_preview_call_output is not annotated with x-ms-foundry-meta.

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
bojunehsu: This is a bit awkward as it appears that we didn't change `memory_search_call` to `memory_search_preview_call` earlier. At this point, we cannot really change that without introducing breaking changes. But I will add preview to the new `memory_command_preview_call[_output]`.

specification/ai-foundry/data-plane/Foundry/src/memory-stores/models.tsp
```diff
@@ -105,6 +106,15 @@ model MemoryStoreDefaultOptions {
 
   @doc("Whether to enable chat summary extraction and storage. Default is true.")
   chat_summary_enabled: boolean = true;
+
+  @removed(Versions.v1)
```
dargilco: Why are these removed from v1? Is this not for //build?

bojunehsu: I would like to merge these changes now to unblock generated C# contracts, but not affect the SDK release this week.  After the SDK release this week, we can remove the @removed decorators.

dargilco (approved):

PR #42852, opened 2026-05-04
Add Microsoft.SaaS TypeSpec GA 2026-04-01 API specification

MichaelMSWork:
## Add Microsoft.SaaS GA 2026-04-01 TypeSpec API Specification

This PR adds the TypeSpec definition for the Microsoft.SaaS resource provider GA API version 2026-04-01.

### Files included:
- TypeSpec source files (`main.tsp`, `common.tsp`, `enums.tsp`, `saasResource.tsp`, `subscriptionLevel.tsp`, `tenantLevel.tsp`)
- `tspconfig.yaml` configuration
- Generated swagger (`stable/2026-04-01/saas.json`)
- Example files for all operations

### Compilation:
- TypeSpec compiles successfully (0 errors, 44 warnings)


specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,110 @@
...
+
+  /** Whether the current term is a Free Trial term. Server-determined. */
+  @visibility(Lifecycle.Read)
+  isFreeTrial?: boolean;
+
+  // NOTE: created and lastModified timestamps are provided by TrackedResource systemData
+  // (systemData.createdAt / systemData.lastModifiedAt) in the ARM resource envelope.
+
+  /** The storefront. */
+  storefront?: string;
```
razvanbadea-msft, line 75: >string [](http://example.com/codeflow?start=15&length=6)

what possible values can be here can it be an enum?

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,110 @@
...
+
+  // NOTE: cspProperties removed from GA API — 0 non-null usage in 18-month production data.
+
+  // NOTE: additionalInfo removed from GA API — 98% usage from synth tests only on RP endpoints.
+  // Non-RP (fulfillment) API retains additionalInfo.
+
+  /** The purchaser email. Server-derived from authenticated identity. */
+  @visibility(Lifecycle.Read)
+  @secret
+  purchaserEmail?: string;
```
razvanbadea-msft: **Blocking — `@secret` conflicts with `Lifecycle.Read`**

`@secret` instructs SDKs/portal to redact the value and signals the linter that the property is write-only / not returned by the service (it emits `format: password` + `x-ms-secret: true`). Here the property is also `@visibility(Lifecycle.Read)`, so the service **does** return it on GET, which conflicts with the secret contract — servers must not return secrets in cleartext.

Could you confirm the intent? A few options that would resolve the conflict:

- **Drop `@visibility(Lifecycle.Read)`** if `purchaserEmail` should truly be write-only (input on create, never returned). This is the most consistent fix if `@secret` is intentional.
- Keep it readable and remove `@secret` if the value is PII rather than a secret (a different classification/handling mechanism would be more appropriate).
- Keep both, with a documented justification, if there is a service contract reason the field is both returned and marked sensitive.

Flagging so the right choice is documented before GA locks the contract.

razvanbadea-msft (commented):
Automated ARM API review — posting Blocking item B1.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,110 @@
...
+  termUnit?: string;
+
+  /** The billing period (ISO 8601 duration). Server-resolved from ProductCode. */
+  @visibility(Lifecycle.Read)
+  billingPeriod?: string;
+
+  /** The risk property bag header (JSON format). Write-only — not returned in GET responses. */
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  @secret
+  riskPropertyBagHeader?: string;
```
razvanbadea-msft, line 54: **Blocking — `riskPropertyBagHeader` is a body property but the name implies an HTTP header**

The name strongly implies an HTTP header, but the generated swagger places this under `properties.*.riskPropertyBagHeader`, so every SDK will surface it as a body property (e.g. `subscription.properties.risk_property_bag_header` / `RiskPropertyBagHeader`). Callers looking for it as an `@header` parameter will be confused.

Could you confirm the intent? If this really is meant to be a body property, the `Header` suffix is misleading and worth revisiting before GA locks the contract; if it was meant to be transmitted as an actual HTTP header, the decorator is missing. Flagging for author guidance rather than proposing a specific rename — happy to follow your direction.

razvanbadea-msft (commented):
Automated ARM API review — posting Blocking item B2.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,110 @@
...
+  /** The end date of the current term. Server-calculated. */
+  @visibility(Lifecycle.Read)
+  termEndDate?: utcDateTime;
+
+  /** Whether the SaaS subscription will auto renew upon term end. */
+  autoRenew?: boolean;
+
+  /** Whether the current term is a Free Trial term. Server-determined. */
+  @visibility(Lifecycle.Read)
+  isFreeTrial?: boolean;
```
razvanbadea-msft, line 69: >isFreeTrial [](http://example.com/codeflow?start=2&length=11)

Replace boolean/switch properties with a more meaningful enum whenever possible. For example two switches "isTypeA" and "isTypeB" should be replaced with one enum "type": [A, B, DefaultType]

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,110 @@
...
+  /** The SaaS Subscription Status. Server-assigned. */
+  @visibility(Lifecycle.Read)
+  status?: SaasResourceStatus;
+
+  /** The reason for the current status. Server-assigned. */
+  @visibility(Lifecycle.Read)
+  statusReason?: SubscriptionStatusReason;
+
+  /** The seat count. */
+  quantity?: int64;
```
razvanbadea-msft, line 29: >int64 [](http://example.com/codeflow?start=13&length=5)

is int64 really needed? or can int32 be used

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: purchaserEmail is PII — an email address derived server-side from the authenticated identity. It is read-only and returned on GET. Displaying the purchaser email in the Azure Portal and partner-facing UIs is a business requirement.

We use @secret intentionally as a PII protection mechanism so that SDK generators emit format: password, ensuring the value is redacted in logs, diagnostic traces while remaining visible in UI surfaces that explicitly render it.

The property is both readable and sensitive — we need it returned on GET for client and UI consumption, @secret is the only TypeSpec decorator that achieves the redaction behavior we require.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: The name riskPropertyBagHeader is intentional and reflects the property's purpose in our system. The value is received in the ARM request body (as a body property), but the RP then forwards it as an HTTP header (x-ms-risk-property-bag) to downstream commerce/billing services for fraud risk assessment. The "Header" suffix describes what the value becomes downstream, not how it's transmitted to the RP.


The property is correctly modeled as a body property with @visibility(Lifecycle.Create, Lifecycle.Update) (write-only) and @secret (contains sensitive fraud-assessment data that should not be logged or returned on GET).

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: isFreeTrial is a standalone boolean flag — it does not form a mutually exclusive pair with any other property. A subscription either is or isn't on a free trial term; there is no third state or alternative category. Replacing it with an enum like trialStatus: [FreeTrial, Paid] would add unnecessary complexity for a simple yes/no condition.

This property is read-only and server-determined — the server evaluates the current term against the offer's trial configuration and returns true or false. The same pattern is used in our Professional Services GA spec (already merged) and in the existing beta API's wire format. Changing the shape would be a breaking change relative to the beta.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: int64 (long in C#) is used because quantity represents seat counts that flow through the Marketplace commerce/billing pipeline, which uses int64 throughout. The RP code stores and processes this as long? (C# Int64). While current production values fit within int32 range, the downstream commerce APIs accept int64 and we need type consistency across the stack to avoid truncation at system boundaries.

The same int64 type is used in our Professional Services GA spec (already merged).

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: storefront is intentionally a string rather than an enum. The set of storefronts that can originate SaaS purchases is large (20+ values today including AzurePortal, AppSource, PartnerCenter, AdminCenter, StoreForBusiness, StoreForLiftr, RPaaS, TeamsAdmin, M365Admin, TeamsClient, ARMApiCall, MPApiCall, Redis, AzureSpringCloud, etc.) and continues to grow as new Microsoft storefronts onboard.

Defining this as an enum would require a spec update and every time a new storefront is added — which is a breaking change for SDKs that use strict enum deserialization. Keeping it as a string allows new storefronts to be added server-side without spec or SDK changes. This is a read-only, server-assigned value that clients consume but never set.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
razvanbadea-msft: - **DO NOT** model customer data (user code, user content, PII) in the control plane resource properties.
- Control plane components (including ARG) cache resource properties and may store them in non-GDPR-compliant ways.
- If customer data must be associated with a resource, expose it via a **data plane API** or use a special query parameter (e.g. `$expandCustomerData=true`) that is not included in the default GET response.

why there is a need to expose this? for the case where it is needed a separated post action is recommended

MichaelMSWork: You're right. Removed purchaserEmail from the resource properties. PII should not be in control plane resource properties — ARG and other control plane components cache these and may store them in non-GDPR-compliant ways. Applied the same fix to our Professional Services GA spec as well.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [6bd677e](https://github.com/Azure/azure-rest-api-specs/pull/42852/commits/6bd677edd3f649b289916d1080ceb45adc3529f5). See inline comments for findings 1-4.

specification/saas/resource-manager/Microsoft.SaaS/Saas/subscriptionLevel.tsp
```diff
@@ -0,0 +1,52 @@
...
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<SaasResource>;
+  update is ArmCustomPatchAsync<
+    SaasResource,
+    Azure.ResourceManager.Foundations.ResourceUpdateModel<
+      SaasResource,
+      SaasResourceProperties
+    >
+  >;
+  delete is ArmResourceDeleteWithoutOkAsync<SaasResource>;
+  listByResourceGroup is ArmResourceListByParent<
```
sandipsh: **[NEW] Blocking** **[[RPC-Get-V1-15](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#22-tracked-resource-requirements)]** specification/saas/resource-manager/Microsoft.SaaS/Saas/subscriptionLevel.tsp - line 28 - collection GET includes a custom query parameter.

**Classification reasoning:** Introduced in this PR. The added Parameters = ExpandCustomerDataQueryParameter on list operations emits $expandCustomerData on collection GET routes; the previous version did not have this parameter on the same surfaces.

**Suggested fix:** Remove Parameters = ExpandCustomerDataQueryParameter from both list operations in this file (including line 47), so the generated collection GET routes do not include custom query parameters.

specification/saas/resource-manager/Microsoft.SaaS/Saas/subscriptionLevel.tsp
```diff
@@ -0,0 +1,52 @@
...
+using TypeSpec.Rest;
+using TypeSpec.Http;
+using Azure.Core;
+using Azure.ResourceManager;
+
+namespace Microsoft.SaaS;
+
+@armResourceOperations
+interface SaasSubscriptionLevel {
+  get is ArmResourceRead<
```
sandipsh, line 15: **[NEW] Blocking** **[[RPC-Get-V1-08](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#22-tracked-resource-requirements)]** specification/saas/resource-manager/Microsoft.SaaS/Saas/subscriptionLevel.tsp - line 15 - point GET includes a non-standard query parameter.

**Classification reasoning:** Introduced in this PR. The tracked resource point GET adds Parameters = ExpandCustomerDataQueryParameter, emitting $expandCustomerData on the point GET route; the previous version point GET had only the standard parameters.

**Suggested fix:** Remove Parameters = ExpandCustomerDataQueryParameter from the point GET operation. If customer data retrieval is still required, expose it through a separate action route instead of point GET query expansion.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
```diff
@@ -0,0 +1,119 @@
...
+   * The purchaser email. Server-derived from authenticated identity. Read-only.
+   *
+   * Customer data (PII): NOT included in default GET responses. Returned only when
+   * the request specifies `$expandCustomerData=true`. Callers requesting customer
+   * data must have appropriate authorization and handle the value per GDPR requirements.
+   *
+   * `@secret` is applied so SDK generators emit `format: password`, ensuring the value
+   * is redacted in logs, CLI output, and diagnostic traces.
+   */
+  @visibility(Lifecycle.Read)
```
sandipsh, line 95: **[NEW] Blocking** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md#proactive-secret-detection-sec-secret-detect)]** specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp - line 95 - secret-marked customer data is exposed in standard resource responses.

**Classification reasoning:** Introduced in this PR. purchaserEmail is modeled as sensitive and emitted with secret semantics, but it remains part of the standard ARM resource response model.

**Suggested fix:** Remove purchaserEmail from SaasResourceProperties. If retrieval is required, provide a dedicated action operation with scoped authorization.

(Updated from previous comment at https://github.com/Azure/azure-rest-api-specs/pull/42852#discussion_r3341173490 - line shifted due to code changes.)

specification/saas/resource-manager/Microsoft.SaaS/Saas/tenantLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+
+@tenantResource
+model SaasTenantResource is ProxyResource<SaasResourceProperties> {
+  ...ResourceNameParameter<
+    Resource = SaasTenantResource,
+    KeyName = "resourceId",
+    SegmentName = "saasresources",
+    NamePattern = "^[a-zA-Z0-9-]{3,50}$"
+  >;
+  ...Azure.ResourceManager.Foundations.ArmTagsProperty;
```
sandipsh, line 21: **[EXISTING] Blocking** **[[Section 1.2 - Proxy Resource Paths](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#12-proxy-resource-paths)]** specification/saas/resource-manager/Microsoft.SaaS/Saas/tenantLevel.tsp - line 21 - proxy resource includes top-level tags.

**Classification reasoning:** This issue is pre-existing and carried forward. The tenant proxy surface had the same tags shape in the previous version and remains present in this PR.

**Suggested fix:** Remove ArmTagsProperty from this proxy resource definition so top-level tags are not emitted for the tenant proxy resource.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [6bd677e](https://github.com/Azure/azure-rest-api-specs/pull/42852/commits/6bd677edd3f649b289916d1080ceb45adc3529f5). See inline comments for findings 1-4.

sandipsh (commented):
## ARM API Review - Posting findings from the ARM API Reviewer agent (critic-verified, 2 iteration(s), converged) against commit [6bd677e](https://github.com/Azure/azure-rest-api-specs/pull/42852/commits/6bd677edd3f649b289916d1080ceb45adc3529f5). See inline comments for findings 1-4.

specification/saas/resource-manager/Microsoft.SaaS/Saas/subscriptionLevel.tsp
MichaelMSWork: Good catch — agreed. Removed $expandCustomerData from all list operations (subscription, resource group, tenant scopes) in both SaaS and Professional Services. The parameter is now only on single-resource point GETs (get, getResource), which is consistent with the original PII access intent: callers wanting customer data must request a specific resource by ID. The corresponding C# RP controllers were also updated to drop the parameter from list endpoints.

specification/saas/resource-manager/Microsoft.SaaS/Saas/saasResource.tsp
MichaelMSWork: The default GET response intentionally does not include purchaserEmail, so ARG and other control plane components that don't know how to store and treat PII never receive it in the standard resource shape. The property is only returned when callers explicitly opt in by passing $expandCustomerData=true on a single-resource GET.

Some clients (Ibiza) must display this field as a business requirement. To support that, we introduced the new $expandCustomerData query flag — this design was proposed and approved by Razvan Badea in an earlier review thread on this PR. The corresponding LintDiff ParametersInPointGet suppression was also added to [readme.md](vscode-file://vscode-app/c:/Users/michaelhirsh/AppData/Local/Programs/Microsoft%20VS%20Code/8761a5560c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) per Razvan's guidance.

Additionally:

@secret is applied so SDK/CLI/Portal redact the value in logs and diagnostic traces.
The flag is restricted to point GETs only — list operations never expose PII (already addressed per the separate RPC-Get-V1-15 feedback).

specification/saas/resource-manager/Microsoft.SaaS/Saas/tenantLevel.tsp
MichaelMSWork: tags has been on the tenant-level [SaasResource]since 2018-03-01-beta (live since 2018). Removing it from GA would be a schema breaking change relative to the live beta, so per the Azure Breaking Changes policy we're keeping it. Since this version exists more then 8 years Breaking Change team trat version 2018-03-01-beta as GA. prefer to leave it as is

specification/saas/resource-manager/Microsoft.SaaS/Saas/tspconfig.yaml
```diff
@@ -0,0 +1,40 @@
...
+    default: "sdk/saas"
+emit:
+  - "@azure-tools/typespec-autorest"
+options:
+  "@azure-tools/typespec-autorest":
+    use-read-only-status-schema: true
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/saas.json"
+    arm-types-dir: "{project-root}/../../../../common-types/resource-management"
+  "@azure-tools/typespec-java":
```
ArthurMa1978, line 12: Please add the emitter setting for .net:
```
       namespace: "Azure.ResourceManager.Saas"
```

PR #42854, opened 2026-05-04
Spec: add build_packages_remotely to CodeConfiguration

sagunase:
## What
Adds an optional boolean property `build_packages_remotely` to the `CodeConfiguration` model on `HostedAgentDefinition`.

## Why
Lets callers signal how dependencies should be resolved at deployment time:
- **`true`** — service builds Python package dependencies remotely (e.g. `pip install` from `requirements.txt` in the uploaded zip).
- **`false`** — caller has already vendored all dependencies into the zip; the service performs no remote build.
- **omitted** — service applies its default behavior.

This unblocks both vendored-deps scenarios (air-gapped / strict-supply-chain customers) and the simpler "just ship sources + requirements.txt" flow without forcing one or the other at the API contract level.

## Scope
- `specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp` — adds one field on `CodeConfiguration`.
- Regenerated `microsoft-foundry-openapi3.json` (v1 + virtual-public-preview).

No example, client.tsp, or SDK changes — the field is optional and existing examples remain valid.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -416,8 +416,14 @@ model CodeConfiguration {
   @doc("The entry point command and arguments for the code execution.")
   entry_point: string[];
 
-  @doc("When true, the service builds Python package dependencies remotely (e.g. installs from `requirements.txt` in the uploaded zip) during deployment. When false, the caller has already bundled all dependencies into the zip and no remote build is performed. When omitted, the service applies its default behavior.")
-  build_packages_remotely?: boolean;
+  @doc("""
+    Indicates how package dependencies are resolved at deployment time.
+    When `true`, the service builds dependencies remotely from the manifest
+    included in the uploaded zip. When `false` (the default), the caller is
+    responsible for bundling all dependencies into the uploaded zip and the
+    service performs no remote build.
+    """)
+  build_packages_remotely: boolean = false;
 
   @visibility(Lifecycle.Read)
   @doc("The SHA-256 hex digest of the uploaded code zip. Set by the service from the `x-ms-code-zip-sha256` request header; read-only in responses and never accepted in request payloads.")
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -416,6 +416,9 @@ model CodeConfiguration {
   @doc("The entry point command and arguments for the code execution.")
   entry_point: string[];
 
+  @doc("When true, the service builds Python package dependencies remotely (e.g. installs from `requirements.txt` in the uploaded zip) during deployment. When false, the caller has already bundled all dependencies into the zip and no remote build is performed. When omitted, the service applies its default behavior.")
+  build_packages_remotely?: boolean;
+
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -416,6 +416,9 @@ model CodeConfiguration {
   @doc("The entry point command and arguments for the code execution.")
   entry_point: string[];
 
+  @doc("When true, the service builds Python package dependencies remotely (e.g. installs from `requirements.txt` in the uploaded zip) during deployment. When false, the caller has already bundled all dependencies into the zip and no remote build is performed. When omitted, the service applies its default behavior.")
+  build_packages_remotely?: boolean;
```
dargilco: Unclear from the doc string what the default behavior is. Can the default be sometimes "true" and sometimes "false" depending on the nature of the package? Or is it always "true" / "false"? If former, can you please say something about it in the doc string? if latter, please state default value here. Thanks!

sagunase: Updated in bc00ef88 -- default is now explicitly `false` (matches today's behavior: caller bundles dependencies in the zip). Doc string clarified and made language-agnostic; `true` is reserved for future remote-build support.

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
dargilco: So the "true" option is not supported at the moment by the service? If so I think the doc string should say so.

sagunase: Currently code is not in private preview. true will be true supported in private preview also.,

johanste (approved):
Approve the REST API changes, however I do want to call out that remote build faults in non-trivial failure modes. I assume we have done our homework on this and don't repeat mistakes from the past.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -416,6 +416,15 @@ model CodeConfiguration {
...
   entry_point: string[];
 
+  @doc("""
+    Indicates how package dependencies are resolved at deployment time.
+    When `true`, the service builds dependencies remotely from the manifest
+    included in the uploaded zip. When `false` (the default), the caller is
+    responsible for bundling all dependencies into the uploaded zip and the
+    service performs no remote build.
+    """)
+  build_packages_remotely: boolean = false;
```
glecaros: I wonder if there's a better name for this. This name seems indirectly related to the description... Maybe something that focuses more on the what instead of the how? `dependency_resolution` with type like
```tsp
union  CodeDependencyResolution {
  string,
  bundled: "bundled",
  remote_build: "remote_build",
}
```

sagunase: Renamed the field to dependency_resolution of a new extensible union CodeDependencyResolution (bundled, remote_build, with string for forward-compat). The default stays equivalent (bundled ↔ the previous false), so wire-level behavior is unchanged. This also leaves room for additional resolution strategies later without  another breaking rename.

dargilco (approved):

PR #42856, opened 2026-05-04
Add extension metadata properties to VirtualMachineExtensionImage API (2026-04-01)

nidhjain-ms1:
## Summary

Adds new read-only properties to the VirtualMachineExtensionImage GET and ListVersions APIs for the 2026-04-01 API version.

### New Properties (GET API)
- **releaseNotes** (string) - Summary of changes in this extension version
- **releaseCategory** (ReleaseCategory) - Type of change: Other, BugFix, SecurityFix, CompatibilityUpdate, NewFeature
- **urgencyLevel** (UrgencyLevel) - Urgency for applying update: Regular, Expedited, Emergency
- **runProfile** (RunProfile) - Execution profile: Other, RunOnce, LongRunning, Hybrid
- **extensionFeatureMetadata** (ExtensionFeatureMetadata) - Flexible property bag with key-value feature tags

### ListVersions API
- Added expand query parameter
- When expanded, responses include releaseCategory, urgencyLevel, and runProfile

### Implementation Details
- All new properties decorated with @added(Versions.v2026_04_01) and @visibility(Lifecycle.Read)
- Updated MaximumSet examples for both GET and ListVersions
- Generated Swagger updated; older API versions unaffected


psah434:
Is the sdk validation status check failing expected?

nidhjain-ms1:
> Is the sdk validation status check failing expected?

@psah434 ,The Java SDK has a hand-written wrapper class VirtualMachineExtensionImageVersionsImpl.java that calls listVersionsWithResponseAsync. After adding the $expand parameter, the auto-generated interface now
expects 7 parameters (String, String, String, String, Integer, String, String — adding $expand), but the hand-written implementation still calls it with only 6 parameters (the old signature without $expand
).
 This is not a bug in our TypeSpec changes — it's expected. The Java SDK's hand-written convenience layer (VirtualMachineExtensionImageVersionsImpl.java) needs to be updated in the azure-sdk-for-java repo
  to pass the new $expand parameter.
  https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/compute/azure-resourcemanager-compute/src/main/java/com/azure/resourcemanager/compute/implementation/VirtualMachineExtensionImageVersionsImpl.java

nidhjain-ms1:
@haagha , The JS sdk has started failing after the merge with base branch. The error seems something related to new-style local paths (./config/tsconfig.src.browser.json) and old-style relative paths (../../../tsconfig.src.browser.json).

haagha:
JS SDK fix has been reverted for the time being. It will be added back in after 2026-03-01 api version is published.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
```diff
@@ -87,6 +89,13 @@ interface VirtualMachineExtensionImages {
       #suppress "@azure-tools/typespec-azure-core/documentation-required"
       @query("$orderby")
       $orderby?: string;
+
+      /**
+       * The expand expression to apply on the operation. Allowed values are 'properties' and 'properties/deprecationStatus'.
+       */
+      @added(Versions.v2026_04_01)
+      @query("$expand")
+      $expand?: string;
```
mentat9: The list payload should return all properties. By hiding the new properties behind a query parameter they will be invisible to Azure Policy and ARG. An alternative would be to always include the new properties and add a select parameter that could be used by clients that want the optimization of only looking at a subset of properties. #Resolved

nidhjain-ms1: These new properties are already exposed via the GET API without any query parameters, ensuring full visibility when retrieving a specific resource.
For the LIST API, however, we have intentionally limited the set of returned properties. Given the scale of LIST responses in our scenario, returning the complete property set can significantly increase response payload size and impact performance, especially considering downstream services caching layers. This aligns with the existing pattern, where LIST APIs have historically returned only a curated set of commonly used properties, while GET APIs provide the complete resource representation.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
```diff
@@ -87,6 +89,13 @@ interface VirtualMachineExtensionImages {
       #suppress "@azure-tools/typespec-azure-core/documentation-required"
       @query("$orderby")
       $orderby?: string;
+
+      /**
+       * The expand expression to apply on the operation. Allowed values are 'properties' and 'properties/deprecationStatus'.
+       */
+      @added(Versions.v2026_04_01)
+      @query("$expand")
```
psah434, line 107: **[NEW] 🟡 Warning** **[Section 7 - Enumerations]**

The `$expand` parameter is modeled as a plain `string`, but the description explicitly lists two allowed values: `'properties'` and `'properties/deprecationStatus'`. A finite set of known values should be modeled as an extensible enum (`union` in TypeSpec) for SDK discoverability and client validation.

**Suggested fix in TypeSpec:**
```tsp
@added(Versions.v2026_04_01)
union ListVersionsExpandOptions {
  string,
  /** Expand properties. */
  Properties: "properties",
  /** Expand deprecation status. */
  PropertiesDeprecationStatus: "properties/deprecationStatus",
}
```

Then use `$expand?: ListVersionsExpandOptions;` in the parameter.

_Classification: **New** - this parameter was introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/examples/2026-04-01/virtualMachineExtensionImageExamples/VirtualMachineExtensionImage_ListVersions_MaximumSet_Gen.json
```diff
@@ -18,7 +19,10 @@
             "computeRole": "aaaaaaaaaaaaaaaaa",
             "handlerSchema": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             "vmScaleSetEnabled": true,
-            "supportsMultipleExtensions": true
+            "supportsMultipleExtensions": true,
+            "releaseCategory": "SecurityFix",
+            "urgencyLevel": "Emergency",
+            "runProfile": "LongRunning"
```
psah434: **[NEW] 🟡 Warning** **[EX-PAYLOAD]**

This is a **MaximumSet** example but it only includes 3 of the 5 new properties (`releaseCategory`, `urgencyLevel`, `runProfile`). The `releaseNotes` and `extensionFeatureMetadata` properties are missing from the response, even though the schema (`VirtualMachineExtensionImageProperties`) declares all 5.

A MaximumSet example should demonstrate the full response shape. If these properties are intentionally omitted from the `ListVersions` response (only returned with specific expand values), then the swagger schema for the List response should model this difference explicitly (e.g., via a separate List-specific response model), rather than reusing the same `VirtualMachineExtensionImageProperties` definition that includes all 5 properties.

**Suggested fix:** Either add `releaseNotes` and `extensionFeatureMetadata` to this example to match the schema, or document why they are excluded and model the List response separately.

_Classification: **New** - these properties and the example changes were introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,153 @@ model VirtualMachineExtensionImageProperties {
...
+  /**
+   * List of additional metadata properties (e.g., compliance flags, supported features).
+   */
+  @identifiers(#[])
+  extensionFeatureTags?: ExtensionFeatureTag[];
+}
+
+/**
+ * Represents a key-value pair for extension feature metadata.
+ */
```
psah434, line 7643: **[NEW] 🟡 Warning** **[Section 4.2 - Documentation]**

The doc comment says _"Holds additional metadata in a flexible property bag."_ - the phrase "flexible property bag" is vague and misleading. This is actually a typed model with a specific `extensionFeatureTags` array property, not a freeform bag. A more precise description would improve SDK documentation quality.

**Suggested fix:**
```tsp
/** Additional metadata about extension features, including compliance and capability tags. */
```

_Classification: **New** - this model was introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
psah434, line 107: **[NEW] 🟡 Warning** **[Section 7 - Enumerations]**

The `$expand` parameter is modeled as a plain `string`, but the description explicitly lists two allowed values: `'properties'` and `'properties/deprecationStatus'`. A finite set of known values should be modeled as an extensible enum (`union` in TypeSpec) for SDK discoverability and client validation.

**Suggested fix in TypeSpec:**
```tsp
@added(Versions.v2026_04_01)
union ListVersionsExpandOptions {
  string,
  /** Expand properties. */
  Properties: "properties",
  /** Expand deprecation status. */
  PropertiesDeprecationStatus: "properties/deprecationStatus",
}
```

Then use `$expand?: ListVersionsExpandOptions;` in the parameter.

_Classification: **New** - this parameter was introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/examples/2026-04-01/virtualMachineExtensionImageExamples/VirtualMachineExtensionImage_ListVersions_MaximumSet_Gen.json
psah434: **[NEW] 🟡 Warning** **[EX-PAYLOAD]**

This is a **MaximumSet** example but it only includes 3 of the 5 new properties (`releaseCategory`, `urgencyLevel`, `runProfile`). The `releaseNotes` and `extensionFeatureMetadata` properties are missing from the response, even though the schema (`VirtualMachineExtensionImageProperties`) declares all 5.

A MaximumSet example should demonstrate the full response shape. If these properties are intentionally omitted from the `ListVersions` response (only returned with specific expand values), then the swagger schema for the List response should model this difference explicitly (e.g., via a separate List-specific response model), rather than reusing the same `VirtualMachineExtensionImageProperties` definition that includes all 5 properties.

**Suggested fix:** Either add `releaseNotes` and `extensionFeatureMetadata` to this example to match the schema, or document why they are excluded and model the List response separately.

_Classification: **New** - these properties and the example changes were introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
psah434, line 7643: **[NEW] 🟡 Warning** **[Section 4.2 - Documentation]**

The doc comment says _"Holds additional metadata in a flexible property bag."_ - the phrase "flexible property bag" is vague and misleading. This is actually a typed model with a specific `extensionFeatureTags` array property, not a freeform bag. A more precise description would improve SDK documentation quality.

**Suggested fix:**
```tsp
/** Additional metadata about extension features, including compliance and capability tags. */
```

_Classification: **New** - this model was introduced in this PR._ #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
nidhjain-ms1: Updated to ListVersionsExpandOptions union. Also removed properties/deprecationStatus from the comment — that was premature; it will be covered in a future version release.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/examples/2026-04-01/virtualMachineExtensionImageExamples/VirtualMachineExtensionImage_ListVersions_MaximumSet_Gen.json
nidhjain-ms1: Split into two examples: ListVersions_MaximumSet_Gen (default response, no expand) and ListVersions_WithExpandProperties_Gen (with $expand=properties showing releaseCategory, urgencyLevel, runProfile)

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Fixed — updated to: "Additional metadata about extension features, including compliance and capability tags."

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
nidhjain-ms1: Duplicate comment.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Duplicate Comment

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/examples/2026-04-01/virtualMachineExtensionImageExamples/VirtualMachineExtensionImage_ListVersions_MaximumSet_Gen.json
nidhjain-ms1: Duplicate Comment

nidhjain-ms1:
@psah434 , the java sdk failures are the same as of before, explanation in this comment: https://github.com/Azure/azure-rest-api-specs/pull/42856#issuecomment-4447927167

The fix Pr was reverted here:  on May 20 by PR #49221 (https://github.com/Azure/azure-sdk-for-java/pull/49221), because nothing was rolled out, and it was causing some issue with their current release.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+ * Enumerates the type of change introduced in the extension version.
+ */
+@added(Versions.v2026_04_01)
+union ReleaseCategory {
+  string,
+
+  /**
+   * Other type of change.
+   */
+  Other: "Other",
```
ravimeda, line 7542: **[NEW] 🟡 Warning** **[Enum design — extensible enum + `Other`]**

`ReleaseCategory` is declared as an extensible union (`union ReleaseCategory { string, ... }`, surfacing as `modelAsString: true` in swagger). Callers can therefore already send any string value. The explicit `Other` member adds no semantic information and forces SDK consumers to handle a meaningless case — a well-established Azure enum anti-pattern for extensible enums.

**Fix:** Remove the `Other` member from `ReleaseCategory`. Document in the union's doc-comment that the enum is extensible so that new categories may be returned in future API versions without an SDK update.

**Classification:** [NEW] — `stable/2026-04-01` does not exist on `main`; `ReleaseCategory` is introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+ * Specifies the execution profile for the extension.
+ */
+@added(Versions.v2026_04_01)
+union RunProfile {
+  string,
+
+  /**
+   * Other execution profile.
+   */
+  Other: "Other",
```
ravimeda, line 7598: **[NEW] 🟡 Warning** **[Enum design — extensible enum + `Other`]**

Same concern as on `ReleaseCategory`: `RunProfile` is an extensible union (`string` base, `modelAsString: true`), so callers can already send any string value. The explicit `Other` member is redundant and forces SDK consumers to handle a meaningless case.

**Fix:** Remove the `Other` member from `RunProfile`. Document extensibility in the doc-comment.

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+/**
+ * Additional metadata about extension features, including compliance and capability tags.
+ */
+@added(Versions.v2026_04_01)
+model ExtensionFeatureMetadata {
+  /**
+   * List of additional metadata properties (e.g., compliance flags, supported features).
+   */
+  @identifiers(#[])
+  extensionFeatureTags?: ExtensionFeatureTag[];
```
ravimeda, line 7638: **[NEW] 🟡 Warning** **[`x-ms-identifiers` semantics]**

`@identifiers(#[])` on `extensionFeatureTags` declares the array has no identifier keys. Combined with `ExtensionFeatureTag.key` and `.value` both being optional, the collection has no uniqueness semantics at all — duplicate keys are legal, ordering is undefined, and SDK generators get no merge/diff anchor. This is a fragile contract for read-only metadata that consumers are very likely to key by `key`.

**Fix (pick one):**
- (a) Make `key` required and use `@identifiers(#["key"])` so SDKs treat `key` as the array identifier; **or**
- (b) Replace the array-of-pairs shape with a structured object whose properties are the actual known feature tags (e.g., `fipsCompliantVersion?: string`). Option (b) is preferable for a closed, read-only metadata set whose tag keys the service controls.

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+   * Expand to include all properties in the response.
+   */
+  Properties: "properties",
+}
+
+/**
+ * Additional metadata about extension features, including compliance and capability tags.
+ */
+@added(Versions.v2026_04_01)
+model ExtensionFeatureMetadata {
```
ravimeda, line 7633: **[NEW] 💡 Suggestion** **[Key-value array shape]**

`ExtensionFeatureMetadata` wraps a list of `{key, value}` pairs. For a read-only metadata bag whose tag keys the service controls (e.g., `FipsCompliantVersion`), consider whether a structured object with named, typed fields (one property per known tag) would serve consumers better than an opaque pair-list. A typed shape gives compile-time discoverability in SDKs and prevents the duplicate-key / ordering ambiguity inherent to `{key,value}` arrays.

Note: the obvious alternative `Record<string>` (`additionalProperties`) is **not** recommended here — it conflicts with the ARM linter rule `arm-no-record`, which is intentionally suppressed throughout `models.tsp`.

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
```diff
@@ -87,6 +89,13 @@ interface VirtualMachineExtensionImages {
       #suppress "@azure-tools/typespec-azure-core/documentation-required"
       @query("$orderby")
       $orderby?: string;
+
+      /**
+       * The expand expression to apply on the operation.
```
ravimeda: **[NEW] 💡 Suggestion** **[Parameter description quality]**

The `` description — *"The expand expression to apply on the operation."* — is generic boilerplate. Consumers reading the spec can't tell what values are valid or what effect each value has.

**Fix:** Use a description that names the allowed value(s) and their effect, e.g.:
> *"Expand the response to include additional read-only metadata. Allowed values: `properties` — returns extended metadata (`releaseCategory`, `urgencyLevel`, `runProfile`)."*

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+   * Hybrid execution profile.
+   */
+  Hybrid: "Hybrid",
+}
+
+/**
+ * Defines the allowed values for the $expand query parameter on the ListVersions operation.
+ */
+@added(Versions.v2026_04_01)
+union ListVersionsExpandOptions {
```
ravimeda, line 7620: **[NEW] 💡 Suggestion** **[Single-value extensible union]**

`ListVersionsExpandOptions` is an extensible (`string`-extending) union with a single member, `"properties"`. The combination commits the API to versioning a one-element extensible enum forever for no current benefit; it also makes the `` parameter's contract opaque (the union name doesn't reveal the only legal value).

**Fix:** If additional `` values are anticipated in upcoming versions, keep the union as-is and document the planned values in the doc-comment. If not, prefer a closed (non-extensible) enum — or simply a literal string type — so the parameter's surface matches its actual capability.

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+  @added(Versions.v2026_04_01)
+  @visibility(Lifecycle.Read)
+  extensionFeatureMetadata?: ExtensionFeatureMetadata;
+}
+
+/**
+ * Enumerates the type of change introduced in the extension version.
+ */
+@added(Versions.v2026_04_01)
+union ReleaseCategory {
```
ravimeda, line 7536: **[NEW] 💡 Suggestion** **[`@added` on initial-version members]**

All new properties, unions, and models in this PR are decorated `@added(Versions.v2026_04_01)`, but `v2026_04_01` is the version currently being assembled — `stable/2026-04-01/` does not yet exist on `main`. `@added` is normally used when adding a member to a *previously shipped* version.

If `v2026_04_01` is intended as the *initial* shipping form of these types (which the base branch `compute-2026-04-01` suggests), the `@added` decorators are unnecessary and may confuse future maintainers diffing against the released spec.

**Fix:** Confirm with the service team that `v2026_04_01` is the initial version for these members. If so, consider removing the `@added(Versions.v2026_04_01)` decorators on members of the initial v2026_04_01 surface. (Posted on `ReleaseCategory` as a representative example; applies to all new `@added` decorations in this PR.)

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -7450,6 +7450,166 @@ model VirtualMachineExtensionImageProperties {
...
+
+/**
+ * Represents a key-value pair for extension feature metadata.
+ */
+@added(Versions.v2026_04_01)
+model ExtensionFeatureTag {
+  /**
+   * The key of the feature tag.
+   */
+  key?: string;
```
ravimeda: **[NEW] 💡 Suggestion** **[Required `key` / `value` on `ExtensionFeatureTag`]**

If the `{key, value}` array shape is retained, both fields are currently optional. A pair with absent `key` is meaningless, and an absent `value` is at best ambiguous.

**Fix:** Remove `?` from `key` (and ideally `value`) in TSP; swagger will then emit `"required": ["key", "value"]`.

**Classification:** [NEW] — introduced by this PR.

 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Other is one of the allowed values in the enum, which would be set by publishers when their release doesn't fit the predefined categories. This is in accordance with our REST APIs and is an intentional design choice — it provides a well-known fallback value with clear semantics, rather than forcing publishers to invent arbitrary strings

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Same as above — Other is an intentional enum value set by extension publishers when their execution profile doesn't match the predefined options. This aligns with our existing REST API patterns.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: We intentionally chose the {key, value} array shape because the set of extension feature tags is open-ended and expected to grow as new capabilities are added

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Good point. We'll make key required and use @identifiers(#["key"]) so SDKs can use it as the array identifier. We're keeping the {key, value} array shape since the set of feature tags is open-ended and will grow over time.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: We plan to add additional $expand values in upcoming API versions (e.g., properties/deprecationStatus for extension deprecation metadata). Keeping this as an extensible union ensures forward compatibility. We've documented the planned expansion in the doc-comment.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/VirtualMachineExtensionImage.tsp
nidhjain-ms1: Agreed — will update the description

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Upon investigation, `@added(Versions.v2026_04_01)` decorators must be retained. `v2026_04_01` is **not** the initial version for this namespace — earlier stable versions (2024-11-01, 2025-03-01, 2025-04-01, 2025-11-01) already exist. These properties are new additions to the existing `VirtualMachineExtensionImageProperties` model, so `@added` is required to prevent them from leaking into older API version swagger outputs.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
nidhjain-ms1: Agreed — will make key required. We'll keep value optional since some feature tags may be boolean flags represented by presence of the key alon

PR #42868, opened 2026-05-04
[Blob Storage] client.tsp customizations for the SubmitBatch operation for Go

gracewilcox:
Adding the client.tsp customizations so `SubmitBatch` will generate correctly for Go

### DIFF
--- a/specification/storage/Microsoft.BlobStorage/client.tsp
+++ b/specification/storage/Microsoft.BlobStorage/client.tsp
@@ -604,6 +604,85 @@ model SignedIdentifiersGo is Array<SignedIdentifier>;
   "go"
 );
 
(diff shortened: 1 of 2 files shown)


specification/storage/Microsoft.BlobStorage/client.tsp
catalinaperalta, line 509: You shouldn't need this, the body type is already bytes.

gracewilcox: No, need it so the type generates as `io.ReadSeekCloser` instead of  `[]byte`

catalinaperalta (approved):
lgtm, these are only client customizations scoped down to go

PR #42880, opened 2026-05-05
[KeyVault] Add enableSoftDelete to create vault examples

rahulalapati43:
## Problem

When creating a Key Vault using any of the 5 Azure management SDKs (Python, Java, JS/TS, Go, .NET), if `enableSoftDelete` is not explicitly set in the request body, the property is omitted from the JSON payload. While the Key Vault service defaults `enableSoftDelete` to `true` server-side, **Azure Policy evaluates the raw request body before the service applies defaults**. If an organization has a policy requiring `enableSoftDelete` to be present and set to `true`, vault creation fails with `RequestDisallowedByPolicy` — even though the service would have defaulted it to `true`.

## Root Cause

All 5 management SDK generators (Python, Java, JS/TS, Go, .NET) auto-generate their create-vault sample code from the spec example JSON files. The current examples do not include `enableSoftDelete` or `enablePurgeProtection` in the request body. Since these are optional properties with server-side defaults, the generators omit them, and users who copy the sample code hit the policy error.

Per TypeSpec guidance, `default` in OpenAPI is a server-side default and documentation hint only — SDK generators do **not** use it as a client-side default. There is no Azure-approved spec decorator that forces all management SDK generators to serialize an optional property's default value on the wire (confirmed by the azure-rest-api-specs team).

## Fix

Add `enableSoftDelete: true` to the request body in both create vault example files:
- `createVault.json`
- `createVaultWithNetworkAcls.json`

When SDKs regenerate their samples from these examples, all 5 SDKs will include these properties in their sample code.

## Related Fixes

This issue was already fixed in the handwritten CLI and PowerShell entrypoints:
- **Azure CLI**: https://github.com/Azure/azure-cli/pull/33265
- **Azure PowerShell**: https://github.com/Azure/azure-powershell/pull/29497

## Broader Pattern

Any property that is (a) optional in the API, (b) has a server-side default via `default`, and (c) is checked by Azure Policy on the raw request body can hit this same issue. Updating examples to always include such properties is the recommended approach for management SDKs.

## Files Changed

- `specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/stable/2026-02-01/examples/createVault.json`
- `specification/keyvault/resource-manager/Microsoft.KeyVault/KeyVault/stable/2026-02-01/examples/createVaultWithNetworkAcls.json`


heaths (changes requested):
This is not the correct fix. As you pointed out, SDKs don't serialize defaults nor do we even encode defaults. That's spelled out in [guidelines](https://aka.ms/azapi/guidelines). The service must treat the absence of a property as the default.

Even if you "fix" the SDKs, what about all the other code out there - whether another deployment framework like Terraform or just some client-generated/-written client - that doesn't send it? You're fixing a symptom, not the problem.

PR #42881, opened 2026-05-05
Spec: include code-based agent deploy routes in v1

sagunase:
Remove @removed(Versions.v1) from code-based agent deploy operations (createAgentFromCode, updateAgentFromCode, createAgentVersionFromCode, downloadAgentCode, downloadAgentVersionCode) and the supporting code_agents_v1_preview opt-in key so the surface is available in v1.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
+++ b/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
@@ -75,8 +75,7 @@
   "paths": {
     "/agents": {
       "post": {
-        "operationId": "Agents_createAgent",
-        "description": "Creates the agent.",
+        "operationId": "Agents_createAgent_Agents_createAgentFromCode",
         "parameters": [
           {
             "name": "Foundry-Features",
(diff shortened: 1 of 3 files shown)


dargilco:
Please include the OpenAPI3 files that got updated when you run "npx tsp compile ."

sagunase:
Done in 35226f9f -- regenerated `openapi3/v1/microsoft-foundry-openapi3.json` is included now (the v1 surface picks up createAgentFromCode/updateAgentFromCode/createAgentVersionFromCode/downloadAgentCode/downloadAgentVersionCode). `virtual-public-preview` already had these routes, so no changes there.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
```diff
@@ -75,8 +75,7 @@
   "paths": {
     "/agents": {
       "post": {
-        "operationId": "Agents_createAgent",
-        "description": "Creates the agent.",
+        "operationId": "Agents_createAgent_Agents_createAgentFromCode",
```
dargilco, line 78: This is a weird operationID value... any idea why this is? I don't know if this string is used as the table-of-content in the reference doc. If so, we need to understand why we got this and potential try to fix it.

glecaros: we could fix this by using the [`@operationId`](https://typespec.io/docs/libraries/openapi/reference/decorators/#@TypeSpec.OpenAPI.operationId) decorator

sagunase: Diagnosed. The pattern is <Interface>_<lowercase-camel-op-name>, produced by @typespec/openapi3 directly from TypeSpec's camelCase operation names. A few data points:

- It's spec-wide, not specific to this PR: **101/101** operationIds in openapi3/v1/microsoft-foundry-openapi3.json and **122/122** in openapi3/virtual-public-preview/microsoft-foundry-openapi3.json follow this convention.
- Agents_createAgent specifically already existed in eature/foundry-release before this PR (verified at the merge-base, 12d8c45028). PR #42881 doesn't introduce this op or this style.

@glecaros's pointer to [@operationId](https://typespec.io/docs/libraries/openapi/reference/decorators/#@TypeSpec.OpenAPI.operationId) is the right mechanism to fix it. Since it would touch ~100 operations across the whole spec and would also change operationIds in irtual-public-preview (heads-up for already-generated SDKs), I'd prefer to handle it in a focused follow-up PR rather than expanding this one's scope.

Tracking it in https://github.com/Azure/azure-rest-api-specs/issues/42929. Happy to defer to your judgment if you'd rather block on it here.

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
dargilco: @sagunase will send a follow up PR to fix the operationID for his APIs. @glecaros we'll need to do the same for other operations.

PR #42892, opened 2026-05-05
Java client updates

kaylieee:


specification/ai-foundry/data-plane/Foundry/client.java.tsp
```diff
@@ -118,6 +119,9 @@ namespace Azure.AI.Projects;
 @@clientName(EvalRunResultComparison, "EvaluationRunResultComparison");
 @@clientName(EvalRunResultSummary, "EvaluationRunResultSummary");
 
+@@clientName(SkillObject, "SkillDetails");
+@@clientName(SkillObject.has_blob, "blob_present");
```
jpalvarezl: Should this be `blobPresent` instead?

kaylieee: codegen emits as `blobPresent`, will change for clarity

PR #42896, opened 2026-05-05
[Storage] APIView TypeSpec Changes (pre-emitter support)

vincenttran-msft:


specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -469,7 +469,7 @@ model SignedIdentifier {
 
 /** The result of a Filter Blobs API call */
 @Xml.name("EnumerationResults")
-model FilterBlobSegment {
+model FilterBlobResponse {
```
catalinaperalta, line 472: Have these new names been run by architects? I believe they usually prefer `Result` as a suffix over `Response`

specification/storage/Microsoft.BlobStorage/models.tsp
```diff
@@ -2763,6 +2767,7 @@ alias StructuredBodyTypeResponseHeader = {
 
 alias StructuredContentLengthParameter = {
   /** Required if the request body is a structured message. Specifies the length of the blob/file content inside the message body. Will always be smaller than Content-Length. */
+  @access(Access.internal)
```
catalinaperalta, line 2770: Should this be scoped down to just rust? FYI @jeremymeng @jorgerangel-msft @l0lawrence @gracewilcox

catalinaperalta: per offline discussion, seems this isnt an issue for the other langs, Go is pending but we can update as needed in a follow up pr

catalinaperalta (approved):

catalinaperalta:
@vincenttran-msft to unblock merging into main, typespec validation needs to be passing

catalinaperalta (approved):

PR #42934, opened 2026-05-06
Add 2026-06-01-preview API version for Microsoft.Devices/DeviceProvisioningServices

mollyiverson:
## Description

Adds the `2026-06-01-preview` API version to `Microsoft.Devices/DeviceProvisioningServices` (DPS).

### New properties on `IotDpsPropertiesDescription`
- `deviceRegistryNamespaces` (read-only) — array of linked Device Registry namespaces (`@maxItems(1)`; `@identifiers(#["namespaceUuid"])`).
- `disableLocalAuth` — disables all authentication methods other than Azure RBAC. Default false.

### New properties on `DeviceRegistryNamespaceDescription`
All properties are optional and gated by `@added(Versions.v2026_06_01_preview)`.
- `namespaceUuid` (`string`) — unique identifier of the linked namespace.
- `dataAddress` (`string`) — data plane address of the linked namespace.
- `location` (`Azure.Core.azureLocation`) — Azure region of the linked namespace.
- `linkingState` (`LinkingState`) — current ADR linking state for this namespace.

### New union: `LinkingState`
Values: `NotLinked`, `InProgress`, `Succeeded`, `Orphaned`.

### Removed properties on `IotDpsPropertiesDescription`
- `deviceRegistryNamespace` (singular) is removed in `2026-06-01-preview` via `@removed(Versions.v2026_06_01_preview)`. It remains unchanged in `2025-02-01-preview`. Replaced by the new `deviceRegistryNamespaces` array.

## Changes
- `main.tsp`: add `v2026_06_01_preview` to the `Versions` enum.
- `models.tsp`:
  - add `@typespec/versioning` import and `Azure.ResourceManager` using;
  - new `LinkingState` union;
  - add `deviceRegistryNamespaces` and `disableLocalAuth` on `IotDpsPropertiesDescription`, gated by `@added(Versions.v2026_06_01_preview)`;
  - add `namespaceUuid`, `dataAddress`, `location`, and `linkingState` on `DeviceRegistryNamespaceDescription`, gated by `@added(Versions.v2026_06_01_preview)`;
  - mark singular `deviceRegistryNamespace` with `@removed(Versions.v2026_06_01_preview)` so it disappears from the 2026 swagger only.
- `readme.md`: add `package-preview-2026-06` tag (set as default).
- `examples/2026-06-01-preview/`: 28 example JSONs.
- Regenerated `preview/2026-06-01-preview/iotdps.json` via `tsp compile`. `preview/2025-02-01-preview/iotdps.json` remains byte-identical.

## ARM Manifest Changes

No manifest changes — additive properties on an existing tracked resource only.


specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -353,8 +355,33 @@ model IotDpsPropertiesDescription {
...
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Version-scoped property using remove/rename/add pattern for version-specific visibility"
+  @removed(Versions.v2026_06_01_preview)
+  @renamedFrom(Versions.v2026_06_01_preview, "deviceRegistryNamespace")
+  deviceRegistryNamespace_v2025_02_01_preview?: DeviceRegistryNamespaceDescription;
+
+  /**
+   * The Device Registry namespace that is linked to the provisioning service.
+   */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_01_preview)
```
pratik11jain: Do we even need this failed in the new api version? I think if we dont need it and if we can remove if we should just remove from the new api version if deviceRegistryNamespaces is being added

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -353,8 +355,33 @@ model IotDpsPropertiesDescription {
...
   deviceRegistryNamespace?: DeviceRegistryNamespaceDescription;
 
+  /**
+   * The Device Registry namespaces linked to the provisioning service, keyed by namespace resource ID.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Dictionary of namespace entries keyed by resource ID"
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_01_preview)
+  deviceRegistryNamespaces?: Record<DeviceRegistryNamespaceDescription>;
```
pratik11jain: Is there a reason for choosing dictionary instead of array? also for the dictionary will there be any limitations on the key string - like supported casing, character limits, length, special characters?
Also, I think for the initial releases I think we want to only allow one value here, is there a way to do this on the swagger? I know arrays on swagger do support enforcing such limits

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -353,8 +355,33 @@ model IotDpsPropertiesDescription {
...
+
+  /**
+   * The ADR linking state of the provisioning service.
+   */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_01_preview)
```
pratik11jain, line 389: shouldn't this be inside the DeviceRegistryNamespaceDescription

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: I think it depends on the logic we have. For current customers, we may want to allow them to use the current deviceRegistryNamespace because they haven't set deviceRegistryNamespaces. In some way we'd want to enforce that deviceRegistryNamespace is readonly either as a field change or in our logic. Thoughts? @pratik11jain

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Moved [af986a4453f0bc4e8cd965728bc7639692e452d3](https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/af986a4453f0bc4e8cd965728bc7639692e452d3)

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -353,8 +356,18 @@ model IotDpsPropertiesDescription {
...
+  @removed(Versions.v2026_06_01_preview)
   deviceRegistryNamespace?: DeviceRegistryNamespaceDescription;
 
+  /**
+   * The Device Registry namespaces linked to the provisioning service.
+   */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2026_06_01_preview)
+  @maxItems(1)
+  @identifiers(#["resourceId"])
```
yang-hai-feng: resource id or namespace uuid?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -394,6 +407,12 @@ model IotDpsPropertiesDescription {
    * Portal endpoint to enable CORS for this provisioning service.
    */
   portalOperationsHostName?: string;
+
+  /**
+   * Disables all authentication methods other than Azure RBAC
+   */
+  @added(Versions.v2026_06_01_preview)
```
yang-hai-feng, line 437: do we need to specify the default as false?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -525,6 +544,13 @@ model DeviceRegistryNamespaceDescription {
       type: "Microsoft.ManagedIdentity/userAssignedIdentities";
     }
   ]>;
+
+  /**
+   * The ADR linking state of this namespace.
```
yang-hai-feng: missing uuid,  @pratik11jain for endpoint and region, do we really needed, if we provision device from CSR flow, we have to call ADR get namespace,  we should be able to get these info, right?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/preview/2026-06-01-preview/examples/DPSGet.json
```diff
@@ -0,0 +1,57 @@
...
+          "deviceProvisioningHostName": "global.azure-devices-provisioning.net",
+          "idScope": "0ne00000012",
+          "portalOperationsHostName": "myFirstProvisioningService.services.azure-devices-provisioning.net",
+          "serviceOperationsHostName": "myFirstProvisioningService.azure-devices-provisioning.net",
+          "state": "Active",
+          "disableLocalAuth": true,
+          "deviceRegistryNamespaces": [
+            {
+              "resourceId": "/subscriptions/91d12660-3dec-467a-be2a-213b5544ddc0/resourceGroups/myResourceGroup/providers/Microsoft.DeviceRegistry/namespaces/myNamespace",
+              "authenticationType": "SystemAssigned",
```
yang-hai-feng, line 39: @mollyiverson can we get example with user assigned manage identity as well? I remember contract is different?

yang-hai-feng (commented):
Do we have to duplicated the example files as current structure?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Switched to array [commit 6dbf307](https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/6dbf307cc9272c7fa8ce507bae925cba8679ef68)

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Removed deviceRegistryNamespace in commit [b13181658eba83c3c03a2a285edf72f7b0a66bd50a2b278dbf58a4776cf48ea5](https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/48f53f8c37904ea9213774be2f7b81f94a71b8d5#diff-b13181658eba83c3c03a2a285edf72f7b0a66bd50a2b278dbf58a4776cf48ea5)

mollyiverson: Removed it [48f53f8c37904ea9213774be2f7b81f94a71b8d5](https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/48f53f8c37904ea9213774be2f7b81f94a71b8d5)

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Added uuid, data endpoint, and region here: https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/fa548cab18a9e6696ea1959313b609d1fd44fd3c

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: I added those to the DeviceRegistryNamespaceDescription in this commit: https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/fa548cab18a9e6696ea1959313b609d1fd44fd3c

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/preview/2026-06-01-preview/examples/DPSGet.json
mollyiverson: Added in this commit https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/a49a4e70383bc5c11adf18daafd3477b5b53f085

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Yes, done in this commit https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/8e7e2de5a13084b5a1a049ed998d4630e6a71391

mollyiverson:
> Do we have to duplicated the example files as current structure?

@yang-hai-feng There is no requirement to, but it was done for the 2025-02-01 version. I recommend that we keep it because it seems like common practice

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
pratik11jain: We need to understand the impact of resource Moves on this. In case of resource moves, the resource ID will change and so the key and the value here will change. How can the impact of moves be removed or minimized here? I think the keys (at least) should not be dependent on the full resource id for that reason

pratik11jain: @yang-hai-feng Thoughts?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -525,6 +544,31 @@ model DeviceRegistryNamespaceDescription {
       type: "Microsoft.ManagedIdentity/userAssignedIdentities";
     }
   ]>;
+
+  /**
+   * Unique identifier of the linked Device Registry namespace.
+   */
+  @added(Versions.v2026_06_01_preview)
+  namespaceUuid: string;
```
pratik11jain: Some of these are added as option and some are required. Is there a reason?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -525,6 +544,31 @@ model DeviceRegistryNamespaceDescription {
...
+  /**
+   * Azure region of the linked Device Registry namespace.
+   */
+  @added(Versions.v2026_06_01_preview)
+  region?: Azure.Core.azureLocation;
+
+  /**
+   * The ADR linking state of this namespace.
+   */
+  @visibility(Lifecycle.Read)
```
pratik11jain: why is this separately added as readonly? isnt the entire object readonly due to [deviceRegistryNamespaces?: DeviceRegistryNamespaceDescription[];]

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: @pratik11jain does the uuid of the namespace stay the same no matter what? maybe that could be the key

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Region was described as an optional field in this design doc https://msazure.visualstudio.com/One/_git/Azure-IoT-Platform-DPS-RP/pullrequest/15658013

mollyiverson: But link state shouldn't be optional. I'll fix that

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Fixed here https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/bf929402131a7a9d582351d9b98f45bc2bc6a3e7

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
yang-hai-feng: We can check with Alyssa's feedback from her resource move knowledge.

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Fixed here https://github.com/Azure/azure-rest-api-specs/pull/42934/changes/bf929402131a7a9d582351d9b98f45bc2bc6a3e7. Region is the only optional field

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
pratik11jain: Yes uuid remains the same between the moves. cc @alyssatan-microsoft

alyssatan-microsoft: how is this key used tho?

mollyiverson: @alyssatan-microsoft The @identifier tag is not required, but adding it can remove a lint error and make it easier for the sdk to merge and compare arrays. In this case, this tag indicates that uuid can uniquely identify elements in the array. I can't imagine there being more than 10 namespaces, let alone five associated with a DPS, so it's not that big of a deal of we don't use this tag. But if the uuid remains the same after a resource move, I think it's a good idea to use it as the identifier

alyssatan-microsoft: sounds good, I agree we should probably use uuid. I just wanted to double check that the NS uuid isn't being used to query into the DP CMS store since we don't partition off of the NS uuid

pratik11jain (approved):

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -945,3 +987,31 @@ model CertificateBodyDescription {
...
+ * The state of ADR linking for a provisioning service.
+ */
+@added(Versions.v2026_06_01_preview)
+union LinkingState {
+  string,
+
+  /**
+   * The provisioning service is not linked to a Device Registry namespace.
+   */
+  NotLinked: "NotLinked",
```
yang-hai-feng, line 1008: @mollyiverson will InProgress changed to NotLinked after timeout?

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
```diff
@@ -579,6 +597,30 @@ model DeviceRegistryNamespaceDescription {
...
+   */
+  @added(Versions.v2026_06_01_preview)
+  location?: Azure.Core.azureLocation;
+
+  /**
+   * The linking state of this namespace.
+   */
+  @added(Versions.v2026_06_01_preview)
+  linkingState?: LinkingState;
 }
```
mollyiverson, line 624: Question for ARM reviewer:
We have a GA API version 2026-08-31 that we're hoping to merge soon, however, we don't want it to include the new fields in this PR, like deviceRegistryNamespaces. We're thinking of this approach with tags:

```
@added(Versions.v2026_06_01_preview)
@removed(Versions.v2026_08_31)
@added(Versions.v2026_11_01)
linkingState?: LinkingState;
```
Is this acceptable?

mollyiverson: Correct, 2026-08-31 has not been shipped as GA with those linking-related fields present. These fields were not present in any previously-released GA versions. 
I will document the gap in the GA readme/changelog.

pratik11jain (approved):

pratik11jain (approved):

yang-hai-feng (approved):

specification/deviceprovisioningservices/resource-manager/Microsoft.Devices/DeviceProvisioningServices/models.tsp
mollyiverson: Sorry I missed this. @yang-hai-feng  I discussed with Hub team in August. Timeouts was an open item, and they planned to have ADR cancel/reset state. linkNotify(failed) will reset to NotLinked

PR #42958, opened 2026-05-07
[Storage] Unblock emitter due to `pub(crate)` in options bag default

vincenttran-msft:


vincenttran-msft:
Not needed, main fix was merged.

PR #42966, opened 2026-05-07
Fix security duplicate for go

jliusan:
for some duplicate deniftions generated in azure-sdk-for-go ,here to add someng go configs to avoid


specification/security/resource-manager/Microsoft.Security/Security/client.tsp
```diff
@@ -365,8 +365,9 @@ op privateLinkResourcesGetCustomization(
 );
 @@clientName(Azure.ResourceManager.CommonTypes.ActionType,
   "ArmActionType",
-  "java"
+  "java,go"
 );
+@@clientName(StandardsAPI.StandardList, "armStandardList", "go");
```
JiaqiZhang-Dev: should it be ArmSecurityStandardList?

specification/security/resource-manager/Microsoft.Security/Security/client.tsp
```diff
@@ -344,7 +344,7 @@ op privateLinkResourcesGetCustomization(
 
 @@clientName(SecurityOperatorsAPI.SecurityOperator, "OperatorResource", "go");
 @@clientName(SecurityStandardsAPI.SecurityStandard,
```
JiaqiZhang-Dev, line 344: should rename SecurityStandardsAPI.SecurityStandard to SecurityStandardsAPI.SecurityStandards right?

jliusan: This config has been added before

specification/security/resource-manager/Microsoft.Security/Security/client.tsp
jliusan: changed to ArmSecurityStandardList

JiaqiZhang-Dev (approved):

PR #42970, opened 2026-05-07
Add force query parameter to DeleteAgent and DeleteAgentVersion APIs

ankitsul:
## Summary

Adds an optional `force` boolean query parameter to both `deleteAgent` and `deleteAgentVersion` operations for hosted agents with active sessions.

## Behavior

| Scenario | `force` omitted / false | `force=true` |
|----------|--------------------------|----------------|
| No active sessions | Deletes normally | Deletes normally |
| Active sessions exist (hosted agents) | **HTTP 409 Conflict** | Cascade-deletes all associated sessions, then deletes |

## Changes

- `agents/routes.tsp`: Added `@query force?: boolean` to `deleteAgent` and `deleteAgentVersion`
- Updated doc comments documenting the 409 behavior and cascade semantics for hosted agents

## Gating

No preview gating (`@removed` or `required_previews`) is applied because:
- `force` is a plain optional parameter on existing non-preview operations
- It defaults to `false` (backward-compatible, no behavioral change for existing callers)
- There is no TypeSpec mechanism to gate a single parameter behind a preview header
- For non-hosted agents (no sessions), the parameter is a no-op

SDK exposure is controlled separately via `Access.internal` shims in `sdk-agents/client.tsp` (both `deleteAgent` and `deleteAgentVersion` are already internal with void-returning public wrappers).

## Design References

- `06-key-decisions.md`: "Agent version deletion is blocked if active sessions exist, unless a force flag is used"
- `w1_adc_integration_runtime_detailed_design_plan.md` §3.2.3: `DELETE /agents/{name}/versions/{version}?force={true|false}`
- AWS Bedrock precedent: `skipResourceInUseCheck` parameter

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
@@ -165,7 +165,7 @@ interface Agents {
       @path agent_name: string;
 
       /** If true, force-deletes the agent even if its versions have active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
-      @query force?: boolean;
+      @query force?: boolean = false;
     },
     DeleteAgentResponse
   >;
@@ -261,7 +261,7 @@ interface Agents {
       @path agent_version: string;
 
       /** If true, force-deletes the version even if it has active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
-      @query force?: boolean;
+      @query force?: boolean = false;
     },
     DeleteAgentVersionResponse
   >;
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
@@ -153,7 +153,9 @@ interface Agents {
   >;
 
   /**
-   * Deletes an agent.
+   * Deletes an agent. For hosted agents, if any version has active sessions, the request
+   * is rejected with HTTP 409 unless `force` is set to true. When force is true, all
+   * associated sessions are cascade-deleted along with the agent and its versions.
    */
   @delete
   @route("/agents/{agent_name}")
@@ -161,6 +163,9 @@ interface Agents {
     {
       /** The name of the agent to delete. */
       @path agent_name: string;
+
+      /** If true, force-deletes the agent even if its versions have active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
+      @query force?: boolean;
     },
     DeleteAgentResponse
   >;
@@ -241,7 +246,9 @@ interface Agents {
   >;
 
   /**
-   * Deletes a specific version of an agent.
+   * Deletes a specific version of an agent. For hosted agents, if the version has active
+   * sessions, the request is rejected with HTTP 409 unless `force` is set to true. When
+   * force is true, all sessions associated with this version are cascade-deleted.
    */
   @delete
   @route("/agents/{agent_name}/versions/{agent_version}")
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -153,14 +153,19 @@ interface Agents {
...
    */
   @delete
   @route("/agents/{agent_name}")
   deleteAgent is FoundryDataPlaneOperation<
     {
       /** The name of the agent to delete. */
       @path agent_name: string;
+
+      /** If true, force-deletes the agent even if its versions have active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
+      @query force?: boolean;
```
dargilco: Specify the service default in the declaration:

@query force?: boolean = false;

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -153,14 +153,19 @@ interface Agents {
...
     },
     DeleteAgentResponse
   >;
```
dargilco, line 171: Does the service have a clear detailed error message, for the case where "force" is not specified (or specified as false) and you try to delete an Agent with active sessions? Please have the message indicate the reason the operation failed, and the need to set "force=true" to have it succeed. Thanks!

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
ankitsul: Done — added `= false` default to both declarations in TypeSpec and updated OpenAPI specs.

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
ankitsul: Yes — the service will return HTTP 409 with a clear error message indicating that the agent version has active sessions, and that the caller should retry with `force=true` to cascade-delete sessions. This will be implemented in the service-side PR alongside the active session guard (PR #2050025).

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -153,14 +153,19 @@ interface Agents {
...
    */
   @delete
   @route("/agents/{agent_name}")
   deleteAgent is FoundryDataPlaneOperation<
     {
       /** The name of the agent to delete. */
       @path agent_name: string;
+
+      /** If true, force-deletes the agent even if its versions have active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
+      @query force?: boolean = false;
```
dargilco, line 168: I would make it more clear by brining the "hosted agent" to the front, and mentioning other Agent types: 
"For Hosted Agent, if true, force-deletes the agent even if its versions have active sessions, cascading deletion to all associated sessions. This value is not relevant for other Agent types. Defaults to false.

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -252,6 +259,9 @@ interface Agents {
 
       /** The version of the agent to delete */
       @path agent_version: string;
+
+      /** If true, force-deletes the version even if it has active sessions (hosted agents), cascading deletion to all associated sessions. Defaults to false. */
+      @query force?: boolean = false;
```
dargilco, line 264: Same

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
ankitsul: Updated — description now reads: "For Hosted Agents, if true, force-deletes the agent even if its versions have active sessions, cascading deletion to all associated sessions. This
 value is not relevant for other Agent types. Defaults to false."

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
ankitsul: Done — applied the same pattern to DeleteAgentVersion.

glecaros (approved):

PR #42988, opened 2026-05-07
remove id and type

abpatelmicrosoft:
# Choose a PR Template

### DIFF
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/ListManagedComputeUsages.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/ListManagedComputeUsages.json
@@ -11,12 +11,10 @@
       "body": {
         "value": [
           {
-            "id": "ManagedCompute.Global.FoundryH100",
             "name": {
               "value": "ManagedCompute.Global.FoundryH100",
               "localizedValue": "Foundry H100 Managed Compute Quota"
             },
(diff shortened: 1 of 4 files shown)


fmabroukmsft:
## ⚠️ Semantic Bug: `x-ms-identifiers` references removed property

This PR removes the `id` property from `ManagedComputeUsage` in `models.tsp`, but does **not** update `@identifiers(#["id"])` on the `value` array in `ManagedComputeUsageListResult` (line ~6112 in models.tsp).

**After merge, the state would be:**
- `ManagedComputeUsage` no longer has an `id` field
- `@identifiers(#["id"])` still references `id` → produces `"x-ms-identifiers": ["id"]` in the swagger
- The swagger declares an identifier property that does not exist in the array item schema

**Impact:** SDK code generators use `x-ms-identifiers` to determine uniqueness keys for array elements. Pointing to a non-existent property will produce broken or incorrect SDK behavior.

**Fix:** If `id` is being removed, also update the identifiers decorator to reference a valid field:
```typespec
// In ManagedComputeUsageListResult:
@identifiers(#["name.value"])   // or another unique field
value?: ManagedComputeUsage[];
```

Then regenerate the swagger (`npx tsp compile .`) so the output reflects the corrected `x-ms-identifiers`.

abpatelmicrosoft:
I have already removed id with name 

// In ManagedComputeUsageListResult:
@identifiers(#["name.value"])   // or another unique field
value?: ManagedComputeUsage[];

PR #43001, opened 2026-05-08
[Storage] `azure_storage_queue` typespec changes

vincenttran-msft:
" "


specification/storage/Microsoft.QueueStorage/client.tsp
```diff
@@ -46,6 +46,7 @@ interface QueueClient extends Queue {}
 @@clientNamespace(Queue, "Azure.Storage.Queues", "rust,go");
 
 @@clientName(MetadataHeaders.metadata, "metadata", "rust");
+@@access(SentMessage, Access.internal, "rust");
```
jhendrixMSFT: I've confirmed with the latest tcgc dev version that the issue is fixed.

Note that you'll also need to add the following to prevent the unreferenced type from being pruned.
```ts
@@usage(SentMessage, Usage.output, "rust");
```

jalauzon-msft (approved):

jalauzon-msft (approved):

catalinaperalta (approved):

PR #43051, opened 2026-05-09
Add compute RM spec validation

divinepwner:


divinepwner:
done security research. has been reported to the MS via MSRC portal.

PR #43065, opened 2026-05-11
Review request for Microsoft.DBforPostgreSQL to add public preview release of version 2026-04-01-preview

nehrao1:

Taking 2026-04-01-preview Swagger update PR forward here since I do not have push permissions to this repo - https://github.com/Azure/azure-rest-api-specs/pull/42266

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ X] New API version for an existing resource provider. (If API spec is not defined in TypeSpec, the PR should have been created in adherence to [OpenAPI specs PR creation guidance](https://aka.ms/azsdkdocs/createopenapispec)).
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

- Introduce new api V20260401
- Add in new API for MaintenanceEvent. Operations include Get, List by Server, Reschedule, and ApplyNow
- Add in new API for MajorVersionUpgradePrecheck. Operations include Get and List by Server.
- Update to add new operation to Servers API: MajorVersionUpgradePrecheck
- Update to add new properties for multi-tenant Microsoft Entra application on Server.

</details>
<details>

suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Maintenance events are read-only resources that do not have a provisioning lifecycle.

suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Major Version Upgrade Precheck are read-only resources that do not have a provisioning lifecycle."

suppress "azure-core/no-guid-usage" "The client id of primary federated identity is to be defined as GUID following recommendation by Azure Resource Manager guidelines."

suppress "azure-core/no-guid-usage" "The client id of backup federated identity is to be defined as GUID following recommendation by Azure Resource Manager guidelines."

FolderStructure Temporarily by pass the check, will have another pr to refactor it.

</details>

## Getting help

  and https://aka.ms/ci-fix.


nehrao1:
[Swagger Avocado](https://github.com/Azure/azure-rest-api-specs/actions/runs/20730415800) : we are using this bug to track pre-existing issue: https://github.com/Azure/azure-rest-api-specs/issues/39199
But have previously received Avacado Approve to bypass until it can be addressed.

psah434:
swagger model validation failing. Please fix.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: Should say "Major version upgrade precheck resource for a ....

nasc17, line 4123: @nehrao1 suppressing provisioning state

nasc17, line 3951: @nehrao1 suppressing provisioning state

specification/postgresql/DBforPostgreSQL.Management/examples/2026-04-01-preview/MajorVersionUpgradePrecheckGet.json
nasc17: subscriptionId would need to be removed

nehrao1: removed

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nehrao1: fixed

nasc17:
This PR continues with review started here: https://github.com/Azure/azure-rest-api-specs/pull/42266

nasc17:
Hi @tadelesh we are getting Breaking Go label added even though build is passing.

mentat9:
> [Swagger Avocado](https://github.com/Azure/azure-rest-api-specs/actions/runs/20730415800) : we are using this bug to track pre-existing issue: #39199 But have previously received Avacado Approve to bypass until it can be addressed.

@nehrao1 - Approving based on the assigned open issue. Note this is not a permanent rubber stamp: please ensure this work gets priority so it is addressed in the next spec PR for your RP.

mentat9:
@nehrao1 - The Typespec Validation failure is a Required check, so I don't have a way to bypass it. The ARM team is OK with bypassing your v2 file/folder restructuring until after these API changes merge, but you will need engage the SDK team to get help on how to deal with this error.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -2011,7 +2012,7 @@ model MigrationProperties {
   /**
    * Identifier of the private endpoint migration instance.
    */
-  migrationInstanceResourceId?: Azure.Core.armResourceIdentifier;
+  migrationInstanceResourceId?: armResourceIdentifier;
```
mentat9, line 2015: >armResourceIdentifier [](http://example.com/codeflow?start=32&length=21)

Can this be specialized to known resource types? Ref: [https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/data-types/#Azure.Core.armResourceIdentifier](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/data-types/#Azure.Core.armResourceIdentifier).

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3159,6 +3160,14 @@ model DataEncryption {
    */
   @visibility(Lifecycle.Read)
   geoBackupEncryptionKeyStatus?: EncryptionKeyStatus;
+
+  @doc("Client id of multi-tenant Microsoft Entra application.")
+  @added(Versions.v2026_04_01)
+  primaryFederatedIdentityClientId?: string;
```
mentat9: >string [](http://example.com/codeflow?start=37&length=6)

Recommend uuid instead of unrestricted string. Note: this may require adding a suppression. If it does, ARM reviewer will approve it. #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3159,6 +3160,14 @@ model DataEncryption {
...
+
+  @doc("Client id of multi-tenant Microsoft Entra application for when it is configured to support geographically redundant backups.")
+  @added(Versions.v2026_04_01)
+  geoBackupFederatedIdentityClientId?: string;
```
mentat9: >string [](http://example.com/codeflow?start=39&length=6)

Recommend uuid #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3925,251 @@ model MigrateNetworkStatus {
...
+  Canceled: "Canceled",
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Maintenance events are read-only resources that do not have a provisioning lifecycle."
+@doc("Properties of a maintenance event resource.")
+@added(Versions.v2026_04_01)
+model MaintenanceEventResourceProperties {
+  @doc("The unique identifier of the maintenance event (e.g. 'op_123456789').")
+  @visibility(Lifecycle.Read)
+  maintenanceEventId?: string;
```
mentat9, line 3957: >maintenanceEventId [](http://example.com/codeflow?start=2&length=18)

Is this identifier chosen by the client? If so, I'd recommend calling it maintenanceEventName. Either way, add some info to the description about where these come from. The example is helpful to illustrate the format. Consider fleshing out the description if the naming format is predictable and contractual or indicating it is expected to be opaque if that's the case. #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3925,251 @@ model MigrateNetworkStatus {
...
+  @visibility(Lifecycle.Read)
+  endTime: utcDateTime;
+
+  @doc("The estimated downtime as an ISO 8601 duration string (e.g., 'PT60S' = 60 seconds).")
+  @visibility(Lifecycle.Read)
+  estimatedDowntime?: string;
+
+  @doc("A value indicating whether this maintenance event can be rescheduled by the customer.")
+  @visibility(Lifecycle.Read)
+  deferrable: boolean;
```
mentat9, line 3985: >boolean [](http://example.com/codeflow?start=14&length=7)

Prefer union over boolean, even if there are only two values. Consider extensibility as well. This could be something like schedulingPolicy: { Fixed, Deferrable }. Or if you expect more possible future options, something like { NoRescheduling, ReschedulesAllowed }, which would allow a future API to have (for example) values like this: { NoRescheduling, ResceduleAllowed, SingleRescheduleAllowed } or other possible additional options. #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3925,251 @@ model MigrateNetworkStatus {
...
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Major Version Upgrade Precheck are read-only resources that do not have a provisioning lifecycle."
+@doc("Major version upgrade precheck resource with validation results.")
+@added(Versions.v2026_04_01)
+model MajorVersionUpgradePrecheckResourceProperties {
+  @doc("The time when the precheck was created.")
+  createTime?: utcDateTime;
+
+  @doc("The status of the precheck validation.")
+  status?: string;
```
mentat9: >string [](http://example.com/codeflow?start=11&length=6)

Does this property take known values? If so, it should be modeled as union. #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3925,251 @@ model MigrateNetworkStatus {
...
+@added(Versions.v2026_04_01)
+model MajorVersionUpgradePrecheckResponse {
+  @doc("The precheck validation ID.")
+  name?: string;
+
+  @doc("The time when the precheck was created.")
+  createTime?: utcDateTime;
+
+  @doc("The status of the precheck validation.")
+  status?: string;
```
mentat9: >status?: string; [](http://example.com/codeflow?start=2&length=16)

Can this be modeled as a union? #Resolved

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -220,6 +220,17 @@ interface Servers {
           RequestIdResponseHeader),
     BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<Server>
   >;
+
+  @doc("Begin Major Version Upgrade Prechecks.")
+  @added(Versions.v2026_04_01)
+  @action("majorVersionUpgradePrecheck")
+  majorVersionUpgradePrecheck is ArmResourceActionAsyncBase<
```
mentat9: >majorVersionUpgradePrecheck [](http://example.com/codeflow?start=2&length=27)

Need to name the action with a verb or verb phrase, e.g. checkForMajorVersionUpgrade. #Resolved

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: It is using Azure.Core;

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: We do not expect more possible options. Wish to continue with Boolean as designed

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: This identifier is not client-chosen — it is service-generated by the Azure Communication Manager (ACM) platform when a maintenance event is created. The format (e.g., YL1T-HFG, RKLQ-BBG) is assigned by ACM and is not contractual; clients should treat it as an opaque string.

We're keeping the name maintenanceEventId rather than renaming to maintenanceEventName because customer-facing notifications (Azure Service Health alerts, email communications) already reference this value as the "event ID," and we want the API property name to match what customers see in those communications.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: Please approve lint error:  Violation: file:///home/runner/work/azure-rest-api-specs/azure-rest-api-specs/after/specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/openapi.json:5295
    Level: error
    Code: GuidUsage
    Message: Usage of Guid is not recommended. If GUIDs are absolutely required in your service, please get sign off from the Azure API review board.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
mentat9: I won't block signoff on this, but seriously, please move away from boolean properties as a design point. unions are better for clients in essentially every way. It's not only about the number of properties.

Unions are more descriptive and flexible while being easier for customers to discover, understand and use. Ref: https://eng.ms/docs/products/arm/rp_onboarding/process/api_review_best_practices. They allow values other than just true/false and support description strings. Even if you only ever want two values, you can have a union with two values.

Often a well-designed union property can even replace multiple boolean properties and eliminate invalid/nonsense combinations. Doing this well requires thinking from the client perspective rather than the service perspective, although the service also often benefits from having an easier upgrade path: unions are conducive to adding scenarios by adding values over time, especially if they have been well thought out in previous versions. Unions seamlessly move from 2 to more than two states, whereas booleans force you to add more booleans to move past two states.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
mentat9: Not sure I understand. I'm asking whether you can enumerate the resource types this property is allowed to take in the type definition like in the example:

  vmIds: armResourceIdentifier<[
    {
      type: "Microsoft.Compute/vm";
      scopes: ["*"];
    }

If you need this property to allow any/all Azure resource types, then it's fine: you don't need to specialize it, but I suspect that's not the case.

mentat9:
@nehrao1 - You also need to address the LintDiff errors and get help from the Typespec/SDK teams to get help with the TypeSpec Validation errors.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: Now I understand. I can bring this question up to control plane for potentially a future change, but this was just to align with the new import. Updating this definition is not in scope for this release.

nasc17:
> @nehrao1 - You also need to address the LintDiff errors and get help from the Typespec/SDK teams to get help with the TypeSpec Validation errors.

@mentat9 
The LintDiff check is pointing to the use of Guid error that was suggested we put in. Do we need to perform another step for the suppresion to be approved?

mentat9:
> > @nehrao1 - You also need to address the LintDiff errors and get help from the Typespec/SDK teams to get help with the TypeSpec Validation errors.
> 
> @mentat9 The LintDiff check is pointing to the use of Guid error that was suggested we put in. Do we need to perform another step for the suppresion to be approved?

Yes, unfortunately you need to add suppressions for each uuid use (sorry, I thought I mentioned that). The Additional information section above has a link to the suppression guide.

nasc17:
> > > @nehrao1 - You also need to address the LintDiff errors and get help from the Typespec/SDK teams to get help with the TypeSpec Validation errors.
> > 
> > 
> > @mentat9 The LintDiff check is pointing to the use of Guid error that was suggested we put in. Do we need to perform another step for the suppresion to be approved?
> 
> Yes, unfortunately you need to add suppressions for each uuid use (sorry, I thought I mentioned that). The Additional information section above has a link to the suppression guide.

@mentat9 The suppression label has been added to each of the properties and was included in Read me. Made update to add them to PR description and reason.

Had to make the GUID suppression more board to be picked up in build. Also included suppressions for the warnings.

alxhghs (approved):

psah434 (commented):
ARM API Review for Microsoft.DBforPostgreSQL/2026-04-01-preview - 3 blocking issues and 1 warning found. Previous version compared: 2026-01-01-preview. No breaking changes detected (additive only).

specification/postgresql/DBforPostgreSQL.Management/Server.tsp
```diff
@@ -220,6 +220,19 @@ interface Servers {
...
+  @added(Versions.v2026_04_01)
+  @action("startMajorVersionUpgradePrecheck")
+  startMajorVersionUpgradePrecheck is ArmResourceActionAsyncBase<
+    Server,
+    StartMajorVersionUpgradePrecheckRequest,
+
+      | ArmResponse<StartMajorVersionUpgradePrecheckResponse>
+      | ArmAcceptedLroResponse<LroHeaders = ArmCombinedLroHeaders<FinalResult = StartMajorVersionUpgradePrecheckResponse> &
+          Azure.Core.Foundations.RetryAfterHeader>,
+    BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<StartMajorVersionUpgradePrecheckResponse>
```
psah434: **[NEW]** :red_circle: **Blocking** **[TSP-BaseParameters]**

The `BaseParameters` here references `DefaultBaseParameters<StartMajorVersionUpgradePrecheckResponse>`, but `StartMajorVersionUpgradePrecheckResponse` is a plain response model, not an ARM resource. Every other action in this file correctly uses `DefaultBaseParameters<Server>`. Using the wrong type will cause incorrect path parameter resolution in the generated swagger.

**Classification:** Newly added operation in this PR (`@added(Versions.v2026_04_01)`), so this is a new issue.

**Fix:**
```tsp
BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<Server>
```

specification/postgresql/suppressions.yaml
```diff
@@ -0,0 +1,6 @@
+- tool: TypeSpecValidation
+  rules: [FolderStructure]
+  sub-rules: [MustUseV2]
+  reason: Temporarily by pass the check, will have another pr to refactor it.
```
psah434: **[NEW]** :red_circle: **Blocking** **[Section 6 - Suppressions]**

New `FolderStructure/MustUseV2` suppression with a vague, deferral-based reason: _"Temporarily by pass the check, will have another pr to refactor it."_ New suppressions MUST have a clear, specific justification. Deferring folder structure migration without a concrete timeline or tracking work item is not acceptable.

**Classification:** New file added in this PR. No previous `suppressions.yaml` existed.

**Fix:** Either migrate to V2 folder structure in this PR, or provide a specific technical reason why it cannot be done now and reference a tracking work item (e.g. GitHub issue or ADO link).

specification/postgresql/resource-manager/readme.md
```diff
@@ -24,22 +24,51 @@ To see additional help and options, run:
...
+    from: openapi.json
+    where: $.definitions.PolicyDetail.properties.passed
+    reason: This is a simple pass/fail indicator.
+  - code: EnumInsteadOfBoolean
+    from: openapi.json
+    where: $.definitions.MaintenanceEventResourceProperties.properties.deferrable
+    reason: This is a binary status indicator.
+  - code: EnumInsteadOfBoolean
+    from: openapi.json
+    where: $.definitions.MaintenanceEventActionResponse.properties.deferrable
```
psah434: **[NEW]** :red_circle: **Blocking** **[Section 6 - Suppressions]**

This suppression targets `$.definitions.MaintenanceEventActionResponse.properties.deferrable`, but the `MaintenanceEventActionResponse` model does **not** have a `deferrable` property. The `deferrable` property exists on `MaintenanceEventResourceProperties` (which is correctly suppressed separately). This suppression is silencing a non-existent violation.

**Classification:** New suppression added in this PR for the `2026-04-01-preview` tag.

**Fix:** Remove this incorrect suppression entry entirely.

specification/postgresql/DBforPostgreSQL.Management/main.tsp
```diff
@@ -57,6 +57,11 @@ enum Versions {
    * The 2026-01-01-preview API version.
    */
   v2026_01_01: "2026-01-01-preview",
+
+  /**
+   * The 2026-04-01-preview API version.
+   */
+  v2026_04_01: "2026-04-01-preview",
```
psah434, line 59: **[NEW]** :yellow_circle: **Warning** **[TSP-3.2]**

The new `v2026_04_01` version enum member for `"2026-04-01-preview"` does not have a `@previewVersion` decorator. Preview API versions should use this decorator to properly signal their preview status to tooling and SDK generators.

Note: The existing `v2026_01_01` member also lacks this decorator (pre-existing issue), but since this PR is adding a new version entry, the decorator should be applied here.

**Fix:**
```tsp
@previewVersion
v2026_04_01: "2026-04-01-preview",
```

specification/postgresql/suppressions.yaml
nasc17: This was following the guidance from TypeSpec team in offline chat. Folder structure will be updated separate to these API changes.

specification/postgresql/resource-manager/readme.md
```diff
@@ -24,22 +24,47 @@ To see additional help and options, run:
...
 
+### Tag: package-flexibleserver-2026-04-01-preview
+
+These settings apply only when `--tag=package-flexibleserver-2026-04-01-preview` is specified on the command line.
+
+```yaml $(tag) == 'package-flexibleserver-2026-04-01-preview'
+input-file:
+  - Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/openapi.json
+suppressions:
+  - code: GuidUsage
```
ravimeda, line 46: **[NEW] 🟡 Warning** **[Suppression-Scope]**

The `GuidUsage` suppression in the new `package-flexibleserver-2026-04-01-preview` tag block has no `where:` clause, so it suppresses the rule for the entire `openapi.json` file. The justification correctly names only two specific properties (`DataEncryption.primaryFederatedIdentityClientId` and `DataEncryption.geoBackupFederatedIdentityClientId`), but the file-wide scope will silence any future GUID-misuse violation introduced by later edits. The 3 sibling `EnumInsteadOfBoolean` entries in the same block correctly use `where:`, making this an outlier.

**Classification:** New suppression introduced in this PR for the new tag.

**Fix:** Split into two scoped entries:
```yaml
- code: GuidUsage
  from: openapi.json
  where: $.definitions.DataEncryption.properties.primaryFederatedIdentityClientId
  reason: Client ID of multi-tenant Microsoft Entra application, defined as GUID per ARM guidelines.
- code: GuidUsage
  from: openapi.json
  where: $.definitions.DataEncryption.properties.geoBackupFederatedIdentityClientId
  reason: Client ID of multi-tenant Microsoft Entra application for geo-redundant backups, defined as GUID per ARM guidelines.
```

specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/examples/MigrationsCreate.json
```diff
@@ -0,0 +1,103 @@
...
+    },
+    "resourceGroupName": "exampleresourcegroup",
+    "serverName": "exampleserver",
+    "subscriptionId": "ffffffff-ffff-ffff-ffff-ffffffffffff"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "name": "examplemigration",
+        "type": "Microsoft.DBForPostgreSql/flexibleServers/migrations",
```
ravimeda, line 33: **[EXISTING] 🟡 Warning** **[Casing-Consistency]**

This example uses casing `Microsoft.DBForPostgreSql/flexibleServers/migrations` (note `DB`/`For`/`Sql`) for the response `type` field, but the canonical provider namespace from the spec definitions is `Microsoft.DBforPostgreSQL`. The same wrong casing appears at multiple locations in this file (lines 22, 33, 34, 55, 57, 68, 69, 90, 92) and likely across the other `Migrations*.json` example files. ARM is case-insensitive on resource `id` segments at runtime, so this does not break clients, but it produces confusing documentation and may affect SDK type-name generation depending on how examples are consumed.

**Classification:** Likely also present in `2026-01-01-preview/examples/` (these example bodies appear to have been copy-forwarded), so this is **pre-existing technical debt** rather than newly introduced. Not a regression in this PR.

**Fix:** Normalize all `type` and `id` fields in `Migrations*.json` example files to the canonical `Microsoft.DBforPostgreSQL/flexibleServers/migrations`.

specification/postgresql/DBforPostgreSQL.Management/routes.tsp
```diff
@@ -109,3 +109,56 @@ interface VirtualNetworkSubnetUsageOperationGroup {
...
+    MaintenanceEventResource,
+    Parameters = {
+      @doc("Filter maintenance events by status.")
+      @query("maintenanceStatus")
+      maintenanceStatus?: MaintenanceEventStatusFilter;
+    }
+  >;
+
+  @doc("Reschedules a maintenance event to a new date and time.")
+  @added(Versions.v2026_04_01)
```
ravimeda: **[NEW] 💡 Suggestion** **[TSP-Redundant-Decorator]**

The `MaintenanceEvents` interface is already decorated with `@added(Versions.v2026_04_01)` at line 115, so every operation inside the interface is implicitly available only in that version. The explicit `@added(Versions.v2026_04_01)` decorator on `reschedule` here (line 131) and on `applyNow` (line 143) is redundant. Note: `get` and `list` correctly omit it — only `reschedule` and `applyNow` carry the redundant repeat.

**Classification:** New operations added in this PR.

**Fix:** Remove the per-operation `@added` lines on `reschedule` and `applyNow`. Harmless either way; this is a code-cleanliness suggestion only.

specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/examples/MajorVersionUpgradePrecheckGet.json
```diff
@@ -0,0 +1,125 @@
+{
+  "parameters": {
+    "subscriptionId": "ffffffff-ffff-ffff-ffff-ffffffffffff",
+    "resourceGroupName": "exampleresourcegroup",
+    "serverName": "exampleserver",
+    "precheckValidationId": "pppppppp-pppp-pppp-pppp",
```
ravimeda: **[NEW] 💡 Suggestion** **[Example-Quality]**

The `precheckValidationId` placeholder `pppppppp-pppp-pppp-pppp` is 20 hex characters across 4 groups (8-4-4-4). A canonical RFC 4122 GUID is 32 hex characters across 5 groups (8-4-4-4-**12**). The same malformed placeholder also appears on this file's response `id` and `name` fields. The sibling file `MajorVersionUpgradePrecheckListByServer.json` uses a properly-formed `cccccccc-cccc-cccc-cccc-cccccccccccc` (32-hex) for the same kind of identifier.

The TypeSpec does not declare `format: uuid` on the `precheckValidationId` path parameter, so this is **not a contract violation** — just a visual inconsistency between sibling examples that could confuse readers.

**Classification:** New example file added in this PR.

**Fix:** Replace `pppppppp-pppp-pppp-pppp` with the canonical 5-group form `pppppppp-pppp-pppp-pppp-pppppppppppp` (or any other valid GUID) consistently across all `MajorVersionUpgradePrecheck*.json` examples.

ravimeda (commented):
Supplementary review from the ARM API Reviewer agent (run by @ravimeda). 4 additional findings (2 warnings, 2 suggestions); previously-raised items by @psah434 and @mentat9 are intentionally not duplicated. See per-comment details below.

specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/examples/MigrationsCreate.json
nasc17: This appears to be a refactoring warning that we can plan for separate to this PR.

specification/postgresql/resource-manager/readme.md
nasc17: Made update but it does not handle supressing in build.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  ...ResourceNameParameter<
+    MaintenanceEventResource,
+    KeyName = "maintenanceEventId",
+    SegmentName = "maintenanceEvents"
+  >;
+}
+
+@doc("Filter for maintenance event status.")
+@added(Versions.v2026_04_01)
+union MaintenanceEventStatusFilter {
```
ravimeda, line 4028: **[NEW] :red_circle: Blocking** **[Schema/doc mismatch — `MaintenanceEventStatusFilter`]**

The filter enum value descriptions reference state names that **do not exist** in `MaintenanceEventStatus` (defined at L3932):

- L4030 `Upcoming`: *"Return events that are scheduled or in progress (Planned, InProgress, **Suspended**)"* — `MaintenanceEventStatus` has no `Suspended`.
- L4033 `Past`: *"Return events that have completed (**Succeeded, Failed, Canceled, TimedOut**)"* — `MaintenanceEventStatus` has only `Complete` and `Canceled` (no `Succeeded`, `Failed`, or `TimedOut`).

The filter contract is incoherent — a customer reading the docs cannot predict which events will be returned. Since this is preview, fixing now is cheap; once stable, customers' switch statements will break.

**Classification:** [NEW] — both enum and filter are added in this PR.

**Fix:** Either rewrite both descriptions to use the actual `MaintenanceEventStatus` values (`Planned`/`InProgress`/`Complete`/`Rescheduled`/`Canceled`), OR add the missing values to `MaintenanceEventStatus` if the service actually returns them.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  errorCode?: int32;
+
+  @doc("The error message if validation failed.")
+  errorMessage?: string;
+
+  @doc("Description of what the policy validates.")
+  policyDescription?: string;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Major Version Upgrade Precheck are read-only resources that do not have a provisioning lifecycle."
```
ravimeda, line 4123: **[NEW] :yellow_circle: Warning** **[`arm-resource-provisioning-state` suppression unjustified — precheck resource HAS a lifecycle]**

The rationale *"Major Version Upgrade Precheck are read-only resources that do not have a provisioning lifecycle"* directly contradicts the resource's own properties — `status: Validating/Succeeded/Failed/Canceled` (L4130) IS a provisioning lifecycle. The entire purpose of a precheck resource is to track an asynchronous validation; that is precisely what `provisioningState` exists for.

**Classification:** [NEW] — new resource, new suppression.

**Fix:** Rename `status` → `provisioningState` (or model via the standard ARM `ProvisioningState` pattern) and remove the suppression. The four terminal values `Validating`/`Succeeded`/`Failed`/`Canceled` already map cleanly onto the ARM provisioning-state contract.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  Complete: "Complete",
+
+  @doc("The maintenance event has been rescheduled (postponed) by the customer.")
+  Rescheduled: "Rescheduled",
+
+  @doc("The maintenance event was canceled, failed, or timed out.")
+  Canceled: "Canceled",
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Maintenance events are read-only resources that do not have a provisioning lifecycle."
```
ravimeda, line 3951: **[NEW] :yellow_circle: Warning** **[`arm-resource-provisioning-state` suppression rationale is imprecise — maintenance event also has a lifecycle]**

The rationale *"Maintenance events are read-only resources that do not have a provisioning lifecycle"* is misleading — `status: MaintenanceEventStatus` at L3969 is a textbook lifecycle (`Planned` → `InProgress` → `Complete`/`Rescheduled`/`Canceled`). The lifecycle exists; the question is *whose* lifecycle.

**Classification:** [NEW].

**Fix:** Either model `status` as `provisioningState` (parallel to the precheck case) and drop the suppression, OR — if you intentionally keep `status` as a domain field because the lifecycle is on the *external* maintenance event rather than on the ARM record — make the rationale precise:
> "The resource is a read-only projection of an externally-managed maintenance event; the event's lifecycle is exposed via `status` and does not represent the ARM resource's provisioning state."

The current rationale is what reviewers will push back on.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+@added(Versions.v2026_04_01)
+model MaintenanceEventActionResponse {
+  @doc("The maintenance event name (maintenance ID).")
+  maintenanceEventId?: string;
+
+  @doc("The full Azure resource ID of the server.")
+  serverId?: armResourceIdentifier;
+
+  @doc("The status (e.g. 'Accepted', 'Succeeded').")
+  status?: string;
```
ravimeda: **[NEW] :yellow_circle: Warning** **[Plain `string` for closed value set — `MaintenanceEventActionResponse.status`]**

`status?: string` with description *"(e.g. 'Accepted', 'Succeeded')"* — that is a closed set. Should be an extensible enum so SDKs generate strongly-typed return values and `x-ms-enum` is emitted in the OpenAPI output.

**Classification:** [NEW].

**Fix:**
```tsp
union MaintenanceEventActionResponseStatus {
  string,
  Accepted: "Accepted",
  Succeeded: "Succeeded",
  // plus other real states (Failed? Canceled?)
}
```

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+
+@doc("Major version upgrade precheck resource for a PostgreSQL flexible server.")
+@parentResource(Server)
+@added(Versions.v2026_04_01)
+model MajorVersionUpgradePrecheckResource
+  is Azure.ResourceManager.ProxyResource<MajorVersionUpgradePrecheckResourceProperties> {
+  ...ResourceNameParameter<
+    MajorVersionUpgradePrecheckResource,
+    KeyName = "precheckValidationId",
+    SegmentName = "majorVersionUpgradePrecheck",
```
ravimeda, line 4190: **[NEW] :yellow_circle: Warning** **[Path segment SHOULD be plural — `majorVersionUpgradePrecheck`]**

`SegmentName = "majorVersionUpgradePrecheck"` generates URLs `.../flexibleServers/{serverName}/majorVersionUpgradePrecheck/{precheckValidationId}` (singular). ARM URL collection segments MUST be plural per Azure REST API Guidelines — and inconsistent with the sibling `maintenanceEvents` segment added in this same PR.

Once shipped this is a one-way door: every later API version pays the consistency cost.

**Classification:** [NEW].

**Fix:** `SegmentName = "majorVersionUpgradePrechecks"`.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+model MaintenanceEventRescheduleRequest {
+  @doc("New start time in RFC3339 format.")
+  postponeToDateTime: utcDateTime;
+}
+
+@doc("Error information from precheck validation.")
+@added(Versions.v2026_04_01)
+model PrecheckErrorInfo {
+  @doc("The error code.")
+  errorCode?: int32;
```
ravimeda, line 4049: **[NEW] :yellow_circle: Warning** **[Integer error code — `PrecheckErrorInfo.errorCode`]**

`errorCode?: int32`. Azure REST API Guidelines and common-types `ErrorDetail.code` specify **string** error codes (machine-readable, enumerable identifiers like `"SubnetUnavailable"`, `"InvalidTargetVersion"`). Integer codes are hard to discover, document, and aggregate across telemetry.

**Classification:** [NEW].

**Fix:** Change to `string` (ideally extensible enum once the value set stabilizes).

(Same issue at `PolicyDetail.errorCode` L4113 — see separate comment there.)

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+@added(Versions.v2026_04_01)
+model PolicyDetail {
+  @doc("The name of the policy.")
+  policyName?: string;
+
+  @doc("Whether the policy validation passed.")
+  passed?: boolean;
+
+  @doc("The error code if validation failed.")
+  errorCode?: int32;
```
ravimeda, line 4114: **[NEW] :yellow_circle: Warning** **[Integer error code — `PolicyDetail.errorCode`]**

Same issue as `PrecheckErrorInfo.errorCode` at L4048 — `errorCode?: int32` should be a **string** error code per Azure REST API Guidelines and common-types `ErrorDetail.code`.

**Classification:** [NEW].

**Fix:** Change to `string` (ideally extensible enum once the value set stabilizes).

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  Canceled: "Canceled",
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Maintenance events are read-only resources that do not have a provisioning lifecycle."
+@doc("Properties of a maintenance event resource.")
+@added(Versions.v2026_04_01)
+model MaintenanceEventResourceProperties {
+  @doc("A service-generated identifier for this maintenance event, assigned by the platform (e.g., 'YL1T-HFG'). The format is not contractual and clients should not attempt to parse or construct this value.")
+  @visibility(Lifecycle.Read)
+  maintenanceEventId?: string;
```
ravimeda, line 3957: **[NEW] :yellow_circle: Warning** **[Identifier duplicated between resource key and properties — `maintenanceEventId`]**

`KeyName = "maintenanceEventId"` on the resource (from `ResourceNameParameter`, L4017-4022) already places this identifier on the resource itself as the `name` field. Carrying the same ID inside `properties` is the classic id-in-properties anti-pattern — guaranteed drift if anything renames the URL segment, and SDK consumers will have two ways to read the same value (one of which can silently fall out of sync).

**Classification:** [NEW].

**Fix:** Remove `maintenanceEventId` from `MaintenanceEventResourceProperties`; clients can use the resource's `name` field.

specification/postgresql/suppressions.yaml
ravimeda: **[NEW] :yellow_circle: Warning** **[Project-level suppression — placeholder rationale]**

The new `MustUseV2` (`TypeSpecValidation` / `FolderStructure`) suppression has rationale *"Temporarily by pass the check, will have another pr to refactor it."* — no tracking issue link, no commitment date, no specific justification. "Temporary" suppressions become permanent without an anchor.

**Classification:** [NEW] — new file in this PR.

**Fix:** Link the tracking issue or PR that will remove the suppression (and a target version, e.g., "Will be removed by #NNNNN before 2026-07-01-preview").

specification/postgresql/resource-manager/readme.md
```diff
@@ -24,22 +24,48 @@ To see additional help and options, run:
...
+    from: openapi.json
+    where: $.definitions.Azure.Core.uuid
+    reason: Client ID of multi-tenant Microsoft Entra application and multi-tenant Microsoft Entra application for geo-redundant backups, defined as GUID per ARM guidelines.
+  - code: EnumInsteadOfBoolean
+    from: openapi.json
+    where: $.definitions.PolicyDetail.properties.passed
+    reason: This is a simple pass/fail indicator.
```
ravimeda, line 61: **[NEW] :yellow_circle: Warning** **[`EnumInsteadOfBoolean` suppression on `PolicyDetail.passed` is weakly justified]**

Rationale *"This is a simple pass/fail indicator"* is plausible, but **policy-validation outcomes commonly grow `NotApplicable`, `Skipped`, or `Warning` states**. Once shipped as `boolean`, expanding to a third state requires a new field and the old one becomes ambiguous (does `passed: false` mean "failed" or "not applicable"?).

The other two `EnumInsteadOfBoolean` suppressions in this block (`deferrable`, `appliedNow`) are genuinely binary — this one is the pushback candidate.

**Classification:** [NEW] — new suppression in this PR's tag block.

**Fix:** Either model `PolicyDetail.passed` as an extensible enum (e.g., `PolicyValidationOutcome { Passed, Failed, NotApplicable, ... }`) and drop the suppression, OR explicitly commit in the suppression rationale to never adding a third outcome.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  lastUpdatedTime?: utcDateTime;
+
+  @doc("The initial scheduled start time before any reschedule (UTC). Equals startTime when the event has never been rescheduled.")
+  @visibility(Lifecycle.Read)
+  originalStartTime: utcDateTime;
+}
+
+@doc("Type of a maintenance event resource.")
+@added(Versions.v2026_04_01)
+union MaintenanceType {
```
ravimeda, line 4006: **[NEW] :bulb: Suggestion** **[Single-value enum — `MaintenanceType`]**

`MaintenanceType` has only one value (`PlannedMaintenance`). Single-value enums add no SDK ergonomic value (no switch statement is meaningful with one arm) but they do lock the OpenAPI shape. Consider replacing with `string` for now, and reintroducing the enum when a second value (e.g., `UnplannedMaintenance`, `EmergencyMaintenance`) actually appears.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+
+  @doc("The target PostgreSQL version.")
+  targetVersion?: PostgresMajorVersion;
+}
+
+@doc("Precheck result details.")
+@added(Versions.v2026_04_01)
+model PrecheckResult {
+  @doc("The action performed.")
+  action?: string;
```
ravimeda, line 4094: **[NEW] :bulb: Suggestion** **[`PrecheckResult.action` typing]**

`action?: string` with description *"The action performed"* — likely a closed set (the operation that was attempted) but the description is too vague to be sure.

**Question for the author:** what are the realistic values for `action`? If closed (e.g., `Validate`, `Upgrade`, `Rollback`), model as an extensible enum. If free-form, the description should be expanded so SDK consumers know what to expect.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+  status?: string;
+
+  @doc("The planned start time of the maintenance event (UTC).")
+  plannedStartTime?: utcDateTime;
+
+  @doc("The planned end time of the maintenance event (UTC).")
+  plannedEndTime?: utcDateTime;
+
+  @doc("A value indicating whether this was an apply-now (immediate) action. True for ApplyNow; false for Reschedule.")
+  appliedNow?: boolean;
```
ravimeda, line 4074: **[NEW] :bulb: Suggestion** **[`MaintenanceEventActionResponse.appliedNow` is redundant]**

The value of `appliedNow` is determined by the operation that produced the response (`true` for `applyNow`, `false` for `reschedule`). Clients already know which operation they called — carrying the answer back in the response body adds nothing.

Consider removing.

ravimeda:
**[NEW] :bulb: Suggestion** **[Precheck creation pattern + `name` field renaming]**

The precheck resource is created via `Servers.startMajorVersionUpgradePrecheck` (action on the server, in `Server.tsp` L228) returning `StartMajorVersionUpgradePrecheckResponse { name, createTime, status }`, then polled via `MajorVersionUpgradePrecheck.get/list`. The precheck interface itself exposes no PUT.

This is viable for job/operation-result resources, but reinforces the broader concern in the [precheck provisioning-state finding](#) — the resource is functionally an async-operation result. Consider whether the standard `Azure-AsyncOperation` / operation-status pattern would serve clients better than a polled proxy resource.

Two smaller adjustments either way:

1. **Rename `StartMajorVersionUpgradePrecheckResponse.name` → `precheckValidationId`** (matches `KeyName` on the precheck resource). Clients otherwise have to guess that `name` is the URL segment to poll. `specification/postgresql/DBforPostgreSQL.Management/models.tsp` L4154.

2. If you keep the polled-resource pattern, make sure the response's polling URL is in `Azure-AsyncOperation` / `Location` headers so generic ARM polling works.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
```diff
@@ -3916,6 +3927,270 @@ model MigrateNetworkStatus {
...
+
+  @doc("Planned maintenance event.")
+  PlannedMaintenance: "PlannedMaintenance",
+}
+
+@doc("Maintenance event resource for a PostgreSQL flexible server.")
+@parentResource(Server)
+@added(Versions.v2026_04_01)
+model MaintenanceEventResource
+  is Azure.ResourceManager.ProxyResource<MaintenanceEventResourceProperties> {
```
ravimeda, line 4017: **[NEW] :bulb: Suggestion** **[`MaintenanceEventResource` ID description vs URL pattern]**

The description for `maintenanceEventId` (L3955) says service-generated IDs look like `'YL1T-HFG'` but the URL pattern is the ARM default `^[a-zA-Z0-9-]{3,24}$` (no `NamePattern` set on the resource at L4016-4022).

If the format is contractual, set `NamePattern` to match the real format. If the format is genuinely not contractual (as the description says — "clients should not attempt to parse or construct this value"), the permissive default is fine. Minor inconsistency either way.

specification/postgresql/DBforPostgreSQL.Management/main.tsp
```diff
@@ -48,15 +48,15 @@ namespace Microsoft.DBforPostgreSQL;
...
-   * The 2025-08-01 API version.
-   */
+  @doc("The 2025-08-01 API version.")
   v2025_08_01: "2025-08-01",
 
-  /**
-   * The 2026-01-01-preview API version.
-   */
+  @doc("The 2026-01-01-preview API version.")
   v2026_01_01: "2026-01-01-preview",
```
ravimeda, line 55: **[EXISTING] :bulb: Suggestion** **[`v2026_01_01` missing `@previewVersion` decorator]**

`v2026_01_01: "2026-01-01-preview"` is a preview version (per its identifier) but lacks the `@previewVersion` decorator — only the new `v2026_04_01` got it (L58).

**Classification:** [EXISTING] — pre-existing on `main`; not introduced by this PR. But the diff in this PR touches the doc comment on these adjacent lines, so it's a low-risk one-line fix to include here.

**Fix:** Add `@previewVersion` decorator on the `v2026_01_01` enum entry.

specification/postgresql/DBforPostgreSQL.Management/models.tsp
nasc17: The filter is grouping together events that fall within the two categories. Will edit description.

specification/postgresql/suppressions.yaml
nasc17: https://github.com/Azure/azure-rest-api-specs/issues/43608

specification/postgresql/DBforPostgreSQL.Management/main.tsp
nasc17: @previewVersion can only be applied to the last member of a Version enum. Having it on other members will cause unstable apis to show up in subsequent stable versions.

specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/openapi.json
```diff
@@ -0,0 +1,10840 @@
...
+      }
+    },
+    "LtrBackupOperationResponseProperties": {
+      "type": "object",
+      "description": "Response for the backup request.",
+      "properties": {
+        "datasourceSizeInBytes": {
+          "type": "integer",
+          "format": "int64",
+
```
ravimeda, line 7080: **[NEW] 🟠 Warning** **[OAPI034]** `specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/openapi.json` — line 7080-7119

`MaintenanceEventActionResponse` is a service-returned action response (returned by `MaintenanceEvents_ApplyNow` and `MaintenanceEvents_Reschedule`). Per **OAPI034 — Clear Field Ownership** (*"Server-owned properties are computed by the service and MUST be `readOnly`"*), every property here is server-set and must be marked `readOnly: true`. None currently is. Compare with the sibling `MaintenanceEventResourceProperties` (lines 7168-7247) where every analogous property correctly has `readOnly: true` — inconsistent mutability for the same domain.

**Classification:** `MaintenanceEventActionResponse` did not exist in `preview/2026-01-01-preview` (introduced by this PR), so this is `[NEW]`.

**Fix:**
1. Add `"readOnly": true` to all seven properties: `maintenanceEventId`, `serverId`, `status`, `plannedStartTime`, `plannedEndTime`, `appliedNow`, `lastUpdatedTime`.
2. Add a `required` array enumerating service-guaranteed fields (at minimum `maintenanceEventId`, `serverId`, `status`) so SDK consumers don't have to null-check guaranteed fields.

In TypeSpec source, decorate the corresponding model with `@visibility(Lifecycle.Read)` on each property (or use `@Azure.ResourceManager.armReadOnly`) and mark required properties without `?`.

ravimeda, line 7398: **[NEW] 🟠 Warning** **[OAPI034]** `specification/postgresql/resource-manager/Microsoft.DBforPostgreSQL/preview/2026-04-01-preview/openapi.json` — line 7398-7430

`MajorVersionUpgradePrecheckResourceProperties` is the properties bag of `MajorVersionUpgradePrecheckResource`, a proxy resource created server-side by the `Servers_StartMajorVersionUpgradePrecheck` POST action and only read back via GET/LIST. `DBforPostgreSQL.Management/routes.tsp` lines 154-162 confirms the `MajorVersionUpgradePrecheck` interface declares only `get is ArmResourceRead<...>` and `list is ArmResourceListByParent<...>` — no PUT/PATCH/DELETE. Per **OAPI034 — Clear Field Ownership**, all five property fields (`createTime`, `status`, `precheckResult`, `targetVersion`, `policyDetails`) are therefore server-owned and MUST be `readOnly: true`. None currently is. SDK generators will emit setters for these properties, falsely implying client mutability.

**Classification:** This properties bag did not exist in `preview/2026-01-01-preview` (introduced by this PR), so this is `[NEW]`.

**Fix:**
1. Mark every property in `MajorVersionUpgradePrecheckResourceProperties` as `"readOnly": true`.
2. Cascade the same to nested types that are reachable **only** through this read-only response surface: `PrecheckResult` (line ~8863), `PolicyDetail` (line ~8731), `PrecheckErrorInfo` (line ~8848), and `UpgradeSequence`.

In TypeSpec source (`models.tsp`), decorate each property with `@visibility(Lifecycle.Read)` (or apply `@Azure.ResourceManager.armReadOnly`) and regenerate.

ravimeda (approved):

ravimeda:
Please address the failing 'Swagger LintDiff' check.

nasc17:
@ravimeda the lintdiff is due to the change in suppression location for the Guid. Can you please advise how best to address this since it is not picking up in the build

specification/postgresql/resource-manager/readme.md
nasc17: @ravimeda this change you suggested is causing lintdiff failure because build does not pick up suppression

ravimeda:
> @ravimeda the lintdiff is due to the change in suppression location for the Guid. Can you please advise how best to address this since it is not picking up in the build

@nasc17 please could you take a look at the violations other than guid usage e.g., OperationsApiSchemaUsesCommonTypes,  PutGetPatchResponseSchema, ProvisioningStateSpecifiedForLROPut, PutResponseCodes, PatchResponseCodes. See https://github.com/Azure/azure-rest-api-specs/actions/runs/26612864087

nasc17:
> > @ravimeda the lintdiff is due to the change in suppression location for the Guid. Can you please advise how best to address this since it is not picking up in the build
> 
> @nasc17 please could you take a look at the violations other than guid usage e.g., OperationsApiSchemaUsesCommonTypes, PutGetPatchResponseSchema, ProvisioningStateSpecifiedForLROPut, PutResponseCodes, PatchResponseCodes. See https://github.com/Azure/azure-rest-api-specs/actions/runs/26612864087

@ravimeda you are referring to violations that were not introduced in this PR. That is out of scope for what we are addressing.

specification/postgresql/resource-manager/readme.md
nasc17: @ravimeda 

change string to uuid: https://github.com/Azure/azure-rest-api-specs/pull/43065#discussion_r3277276094
add where for suppression: https://github.com/Azure/azure-rest-api-specs/pull/43065#discussion_r3312396604

ravimeda (approved):

ravimeda:
Following up on the `uuid` typing recommendations on this PR ([r3277276094](https://github.com/Azure/azure-rest-api-specs/pull/43065#discussion_r3277276094), [r3277276219](https://github.com/Azure/azure-rest-api-specs/pull/43065#discussion_r3277276219)) and the resulting `GuidUsage` LintDiff failure. Apologies for the friction this has caused.

Recommending `uuid` for properties that hold RFC 4122 GUIDs is correct typing, but in ARM specs it emits `"format": "uuid"` in the generated Swagger, which trips the required `GuidUsage` LintDiff rule and needs Azure API review board sign-off plus a scoped per-property suppression. The suppression path you are already on is the right one. For reference, the preferred form is one entry per property (not file-wide), so future GUID misuse is still caught:

```yaml
suppressions:
  - code: GuidUsage
    from: openapi.json
    where: $.definitions.DataEncryption.properties.primaryFederatedIdentityClientId
    reason: Microsoft Entra application client ID (RFC 4122 UUID) per ARM identity guidelines. Approved by Azure API review board: <link or reviewer handle>.
  - code: GuidUsage
    from: openapi.json
    where: $.definitions.DataEncryption.properties.geoBackupFederatedIdentityClientId
    reason: Microsoft Entra application client ID for geo-redundant backups (RFC 4122 UUID) per ARM identity guidelines. Approved by Azure API review board: <link or reviewer handle>.
```

We are also updating the ARM API reviewer agent guidance so any future `uuid` recommendations on ARM specs call out this `GuidUsage` trade-off and the scoped-suppression requirement up front. Thanks for your patience.

ravimeda:
I'm sorry for the back-and-forth on this. My earlier suppression template was wrong for your case, and you've now spent extra cycles on what should have been a simple fix. That's on me.

The actual `GuidUsage` violation isn't reported at the property paths I suggested. Because the spec is TypeSpec-generated, the `uuid` scalar produces a single shared definition (`Azure.Core.uuid`) that every property `$ref`s, and LintDiff fires on that shared definition. From the run log ([26655490021](https://github.com/Azure/azure-rest-api-specs/actions/runs/26655490021)):

```text
Code: GuidUsage
jsonpath: ["definitions", "Azure.Core.uuid", "format"]
openapi.json#L5301
```

Three options, your choice:

**Option 1. Correct the suppression.** Replace the per-property entries in `readme.md` with a single entry that targets the shared TypeSpec-generated definition, and enumerate the relying properties in the `reason` so future GUID misuse on unrelated properties is still visible at review time:

```yaml
suppressions:
  - code: GuidUsage
    from: openapi.json
    where: $.definitions["Azure.Core.uuid"]
    reason: >-
      Used only for Microsoft Entra application clientId properties on
      DataEncryption (primaryFederatedIdentityClientId,
      geoBackupFederatedIdentityClientId). Both are RFC 4122 UUIDs per ARM
      identity guidelines. Approved by Azure API review board:
      <link or reviewer handle>.
```

**Option 2. Revert the two properties to `string`** in the TypeSpec source. Keep the GUID format in the description. No suppression needed; LintDiff passes on its own.

**Option 3. ARM reviewer override.** I can apply an override to bypass the failing `GuidUsage` check for this PR. Appropriate here because both properties are Microsoft Entra application client IDs, which is an accepted exception. Let me know and I will apply it.

Any of the three is acceptable; I'll defer to your preference. Thanks for your patience, and again, apologies for the extra round-trip.

ravimeda:
I'm sorry. My Option 1 suppression was almost right but missed by one path segment. You applied it correctly; the validator just needs the **exact** JSONPath of the violation, not an ancestor. The failure jsonpath reported by LintDiff is `["definitions", "Azure.Core.uuid", "format"]`, so the `where:` needs `.format` on the end.

Two small fixes to your `readme.md`:

```yaml
suppressions:
  - code: GuidUsage
    from: openapi.json
    where: $.definitions["Azure.Core.uuid"].format
    reason: >-
      Used only for Microsoft Entra application clientId properties on
      DataEncryption (primaryFederatedIdentityClientId,
      geoBackupFederatedIdentityClientId). Both are RFC 4122 UUIDs per
      ARM identity guidelines. Approved by @ravimeda.
```

Also, please remove the two leftover per-property entries (`$.definitions.DataEncryption.properties.primaryFederatedIdentityClientId` and `.geoBackupFederatedIdentityClientId`). The shared-definition entry above covers both; the per-property entries don't match anything and are silently no-ops.

If this still doesn't clear, **Option 3 stands**.

Thanks for sticking with this.

rmendiza (approved):

PR #43069, opened 2026-05-11
[PreviewAlertRule] Add C# MPG emitter config and client customizations

mzhongl524:
## Summary

Adds the C# management-plane emitter configuration and TypeSpec client customizations for the **PreviewAlertRule** sub-namespace of `Microsoft.AlertsManagement`, enabling a dedicated SDK package `Azure.ResourceManager.PreviewAlertRule` to be generated.

### `tspconfig.yaml`
- Added `@azure-typespec/http-client-csharp-mgmt` emitter section with:
  - `namespace: Azure.ResourceManager.PreviewAlertRule`
  - `emitter-output-dir: {output-dir}/sdk/alertsmanagement/{namespace}`

### `client.tsp`
C# customizations to produce a clean, idiomatic .NET SDK surface:

- `AlertSeverity.0..4` → `Sev0..Sev4` (numeric identifiers are illegal as C# field names)
- `Kind` → `LogAlertRuleKind` (overly generic top-level model name)
- `Evaluation` → `RulePreviewEvaluation`
- `EvaluatedPeriod` → `RulePreviewEvaluatedPeriod`
- `DimensionNameAndValue` → `RulePreviewDimensionNameAndValue`
- `LogAlertRuleResource` → `LogAlertRule` (mgmt `InheritanceCheck` test rejects non-ArmResource types whose name ends in `Resource`)
- `ConditionOperator.Equals` → `EqualTo` (collides with `ValueType.Equals(object)` on the generated readonly struct)
- `@@clientLocation(PreviewAlertRules.previewAlertRule, PreviewAlertRules, "csharp")` to keep the op on a stable client group

### `models.tsp`
- `union AlertSeverity` base type changed from `int64` to `int32`.
  `Microsoft.TypeSpec.Generator.Input.InputEnumTypeValueConverter` only supports `String`/`Int32`/`Float32` enum value types; `int64` crashed the C# generator. JSON wire format is unchanged — the values are `0..4`, which serialize identically as int32 or int64.

### `routes.tsp`
- Wrapped the previously-standalone `previewAlertRule` op in an `interface PreviewAlertRules { ... }` so the generator can group it onto a proper operation group. The operation signature, `@route`, `@post`, request/response bodies, and runtime behavior are unchanged.

### `PreviewAlertRule.json`
Regenerated by `tsp compile`. The only resulting wire changes are documentation-level:
- `operationId`: `PreviewAlertRule` → `PreviewAlertRules_PreviewAlertRule` (standard `Group_Operation` shape from the new interface wrap).
- `x-ms-examples` reference removed (will be re-attached via `@opExample` in a follow-up if desired). No runtime impact.

## Paired SDK PR
Azure/azure-sdk-for-net PR (to be linked once opened).


specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/preview/2025-07-01-preview/PreviewAlertRule.json
```diff
@@ -42,7 +42,7 @@
   "paths": {
     "/{resourceId}/providers/microsoft.AlertsManagement/previewAlertRule": {
       "post": {
-        "operationId": "PreviewAlertRule",
+        "operationId": "PreviewAlertRules_PreviewAlertRule",
```
live1206: why do we need to update the operationId, does it really matter for the mpg generator?

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/routes.tsp
```diff
@@ -11,17 +11,32 @@ using Azure.ResourceManager;
...
-@route("/{resourceId}/providers/microsoft.AlertsManagement/previewAlertRule")
-@post
-op previewAlertRule(
-  ...ApiVersionParameter,
-  #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-  @path resourceId: string,
-  #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-  @body parameters: PreviewAlertRuleRequest,
-): PreviewAlertRuleResponse | ErrorResponse;
+interface PreviewAlertRules {
```
live1206: consider to create the interface in client.tsp and use clientLocation for csharp scope only, so that the swagger is not affected

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/models.tsp
```diff
@@ -34,8 +34,15 @@ union Kind {
 /**
  * Severity of the alert. Should be an integer between [0-4]. Value of 0 is severest. Relevant and required only for rules of the kind LogAlert.
  */
+// NOTE: Base type intentionally `int32` (not `int64`). The C# MPG generator's
```
live1206: is this a generator issue then?

mzhongl524: Yes — confirmed it's a limitation in the upstream `microsoft/typespec` `http-client-csharp` generator (which the MPG emitter is built on).

The relevant switch in `InputEnumTypeValueConverter.CreateEnumTypeValue` only handles `String / Int32 / Float32` and falls through to `throw new JsonException()` for everything else (including `Int64`):

[InputEnumTypeValueConverter.cs](https://github.com/microsoft/typespec/blob/main/packages/http-client-csharp/generator/Microsoft.TypeSpec.Generator.Input/src/InputTypes/Serialization/InputEnumTypeValueConverter.cs)

The crash surfaces as a `TypeInitializationException` at `ModelReaderWriterContextDefinition` static-init time with the inner `JsonException` swallowed, which is why the build output is so unhelpful here.

Since the values are `0..4` and serialize identically as `int32` or `int64` on the wire, the `int32` workaround is zero-cost and only affects .NET. I'll hold on filing an upstream issue until you confirm direction — happy to open one against `microsoft/typespec` to add `Int64` (and the other missing primitive kinds) to the converter if you'd prefer the spec stay `int64`.

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/preview/2025-07-01-preview/PreviewAlertRule.json
mzhongl524: fixed

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/client.tsp
```diff
@@ -26,3 +27,127 @@ using Microsoft.AlertsManagement;
...
+// the operationId remains `PreviewAlertRule` (rather than becoming
+// `PreviewAlertRules_PreviewAlertRule`). `@@clientLocation` is scoped to
+// "csharp" so other TCGC-based emitters (Python/Java/JS/Go) also see no change.
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+namespace Microsoft.AlertsManagement {
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  interface PreviewAlertRules {}
+}
+
+@@clientLocation(previewAlertRule, PreviewAlertRules, "csharp");
```
ArcturusZhang: we do not need to declare an interface here, clientLocation could accept a string
```suggestion
@@clientLocation(previewAlertRule, "PreviewAlertRules", "csharp");
```
and remove the above interface declaration.

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/client.tsp
```diff
@@ -26,3 +27,127 @@ using Microsoft.AlertsManagement;
...
+// `@@usage` alone makes `LogAlertRule` constructible, but the C# MPG generator's
+// implicit-flatten still emits these scalars as getter-only delegations to an
+// internal `Properties` field. Combining `Usage.input` with `flattenProperty`
+// turns them into normal `{ get; set; }` C# properties so callers can configure
+// a `LogAlertRule` payload through standard object-initializer syntax.
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Required so flattened scalars are settable"
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(
+  LogAlertRuleResource.properties,
+  "csharp"
+);
```
ArcturusZhang: I think  in general, the new packages should not use legacy decorators, especially flattenProperty. we could just remove this and disable flattening for all properties in this RP.

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/models.tsp
```diff
@@ -34,8 +34,15 @@ union Kind {
...
+// `Microsoft.TypeSpec.Generator.Input.InputEnumTypeValueConverter` only supports
+// `String`, `Int32`, and `Float32` enum value types; an `int64`-backed union
+// crashes the generator with a `TypeInitializationException` at
+// `ModelReaderWriterContextDefinition`. The values here are 0..4, which
+// serialize identically as `int32` or `int64` over the wire, so this change is
+// wire-format safe.
 union AlertSeverity {
-  int64,
+  int32,
```
ArcturusZhang: we need to revert this change. if there is a generator issue, please open an issue in azure-sdk-for-net repo, and we should solve it there.

specification/alertsmanagement/resource-manager/Microsoft.AlertsManagement/PreviewAlertRule/client.tsp
```diff
@@ -26,3 +27,92 @@ using Microsoft.AlertsManagement;
...
+// properties — which makes it impossible for callers to construct a
+// `LogAlertRule` payload. `Usage.input` keeps the public parameterless
+// constructor and the writable scalar setters that are required for a
+// request-body shape.
+@@usage(LogAlertRuleResource, Usage.input, "csharp");
+@@usage(LogAlertRuleProperties, Usage.input, "csharp");
+@@usage(LogAlertRuleCriteria, Usage.input, "csharp");
+@@usage(LogAlertRuleCondition, Usage.input, "csharp");
+@@usage(LogAlertRuleDimension, Usage.input, "csharp");
+@@usage(LogAlertRuleConditionFailingPeriods, Usage.input, "csharp");
```
ArcturusZhang: I think this is a new package, we do not really need to add these usage customization

specification/alertsmanagement/cspell.yaml
```diff
@@ -12,4 +12,5 @@ words:
   - groupedby
   - scom
   - zabbix
-
+  - TCGC
+  - TCGC's
```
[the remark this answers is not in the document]
ArcturusZhang: we have to add these cspell words because we used them in comments?
this does not really make sense.
could we somehow simplify our comments, to remove these two words?

live1206 (approved):

live1206:
This PR only updated csharp client and its config, no need of ARM review

specification/alertsmanagement/cspell.yaml
```diff
@@ -12,4 +12,3 @@ words:
   - groupedby
   - scom
   - zabbix
-
```
live1206, line 15: we should revert the change in this file

live1206 (approved):

PR #43085, opened 2026-05-11
Fix Versions enum ordering to retain stable 2025-01-01 in KnownVersions

RenzoPrettoMS:
## Problem
The \2026_05_01_preview\ version was inserted before \2025_01_01\ in the \Versions\ enum in \main.tsp\. When TCGC generates SDK code for a specific version, it does not include enum values that appear after the target version. This caused the stable GA \2025-01-01\ API version to be dropped from \KnownVersions\ in the generated SDK — a breaking change for customers who depend on explicit version pinning.

## Fix
Reorder the \Versions\ enum so that \2025_01_01\ appears chronologically before \2026_05_01_preview\:

\\\
v2024_11_01_preview -> v2025_01_01 -> v2026_05_01_preview
\\\

After this change is merged, re-triggering SDK generation will restore \2025-01-01\ in \KnownVersions\.


specification/chaos/resource-manager/Microsoft.Chaos/Chaos/main.tsp
```diff
@@ -51,12 +51,12 @@ enum Versions {
   v2024_11_01_preview: "2024-11-01-preview",
```
weidongxu-microsoft: You may want to remove `v2024_11_01_preview` from TypeSpec (keep it in Swagger as long as it is online)

https://azure.github.io/typespec-azure/docs/howtos/versioning/01-about-versioning/#preview-versioning-rules-for-all-azure-apis):

weidongxu-microsoft (approved):

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/main.tsp
RenzoPrettoMS: Good catch, thank you! Updating.

RenzoPrettoMS: Can I get a reapproval when you have a chance @weidongxu-microsoft

weidongxu-microsoft (approved):

PR #43094, opened 2026-05-11
Add response body to CreateModelVersionAsyncResponse

vechillal:
## Summary

The \createAsync\ operation on Models returns a 202 Accepted with a JSON body from \LongRunningOperationUriResponse\, but the TypeSpec model \CreateModelVersionAsyncResponse\ had no body — causing a \
o-response-body\ linter warning.

## Changes

- Added \location?: url\ and \operationResult?: url\ fields to the response body in \CreateModelVersionAsyncResponse\
- \location\ is also set as the \Location\ response header (matching service behavior — the controller sets both)
- \completionResult\ is **not** included: the service's response filter intercepts \LongRunningOperationResponse<FoundryModelDto>\ and replaces it with a \LongRunningOperationUriResponse\ (stripping \completionResult\) before the response hits the wire

## Testing

\
px tsp compile .\ now completes with no warnings.


specification/ai-foundry/data-plane/Foundry/src/models/models.tsp
```diff
@@ -193,5 +193,14 @@ model CreateModelVersionAsyncResponse {
 
   @doc("URL to poll for operation status.")
   @header("Location")
-  location: string;
+  location: url;
```
dargilco: We have "location" in both header and body, referring to different things. Can we have better names? like "polling_url" for the header and "operation_url" for the body?

specification/ai-foundry/data-plane/Foundry/src/models/models.tsp
```diff
@@ -193,5 +193,14 @@ model CreateModelVersionAsyncResponse {
...
+
+  @bodyRoot
+  body: {
+    @doc("URL to the location of the completed or in-progress operation.")
+    location?: url;
+
+    @doc("URL to the operation result, or null if the operation is still in progress.")
+    operationResult?: url | null;
```
dargilco, line 231: Please use snake case ("operation_result") instead of CamelCase

specification/ai-foundry/data-plane/Foundry/src/models/models.tsp
vechillal: These locations refer to the same thing, the header is set from the response body ([service code](https://msdata.visualstudio.com/Vienna/_git/vienna?path=/src/azureml-api/src/ModelRegistry/EntryPoints/Api/Controllers/FoundryModelController.cs&version=GBmaster&line=245&lineEnd=246&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents))

dargilco: If it's the same value, please update the doc string to show the same doc string for both. And I suggest removing the duplication before going GA.

dargilco:
Please add a comment above where you define  createOrUpdateVersion in file src\models\routes.tsp to say why you are overriding a method that is defined in the VersionedOperations template.

specification/ai-foundry/data-plane/Foundry/src/models/models.tsp
vechillal: Updated

glecaros (approved):

dargilco (approved):

vechillal:
@copilot resolve the merge conflicts in this pull request

PR #43106, opened 2026-05-12
Addressing PR comments

rchepala:
Addressing PR comments: https://github.com/Azure/azure-rest-api-specs/pull/42717

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
+++ b/specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
@@ -66,7 +66,6 @@ interface ConnectionPolicies {
    */
   delete is ArmResourceDeleteWithoutOkAsync<
     ConnectionPolicy,
-    Response = ArmDeleteAcceptedLroResponse | ArmDeletedNoContentResponse,
     Error = CloudError
   >;
 


pshao25 (changes requested):
We should get Mark's confirmation https://github.com/Azure/azure-rest-api-specs/pull/42717#discussion_r3206196661

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
```diff
@@ -19,28 +19,24 @@ namespace Microsoft.Network;
...
-model ConnectionPolicy extends ProxyResource {
-  /**
-   * Properties of the ConnectionPolicy resource.
-   */
-  properties?: ConnectionPolicyProperties;
+model ConnectionPolicy
+  is Azure.ResourceManager.ProxyResource<ConnectionPolicyProperties> {
+  ...ResourceNameParameter<
+    Resource = ConnectionPolicy,
+    KeyName = "connectionPolicyName",
```
markcowl: KeyName and SegmentName parameters are unnecessary here, these match the defaults

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
```diff
@@ -19,28 +19,24 @@ namespace Microsoft.Network;
...
+   * A unique read-only string that changes whenever the resource is updated.
    */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "etag is a top-level property consistent with all other VirtualWAN nested resources"
   @visibility(Lifecycle.Read)
-  @path
-  @key("connectionPolicyName")
-  @segment("connectionPolicies")
-  @pattern("^(?![.-])[a-zA-Z0-9_.-]{1,128}$")
-  name: string;
+  etag?: string;
```
markcowl: This should use `...EntityTagProperty`
```suggestion
  ...EntityTagProperty;
```

In this case the visibility decorator is not needed as well.  This will put some standard documentation around the eTag, including the fact that it may also be included in the header, as per standard eTag practice.

This isn't a required change, but the proeprty there is identical to this one, and should remove the need for the suppression abov

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
```diff
@@ -19,28 +19,24 @@ namespace Microsoft.Network;
  * ConnectionPolicy resource defined for VirtualHub.
  */
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Legacy feature decorator required for Network RP VirtualWAN feature gating, consistent with all other VirtualWAN interface definitions"
```
markcowl, line 21: I don't think you need this suppression anymore either

markcowl (commented):
The only concern is with sdks.  If this produces no sdk breaks, this is a *much better* solutiion for the long term and will make this spec easier to maintain.

All the changes below are nits

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
rchepala: Done

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
rchepala: Suppression is still needed.

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
rchepala: Tried this, but `...EntityTagProperty` generates `eTag` (capital T) in the swagger output, while the examples and all existing Network RP resources use lowercase `etag`. This caused oav example validation failures (`OBJECT_ADDITIONAL_PROPERTIES`). Reverted to the explicit `etag?: string` property to stay consistent with the rest of the RP, the suppression is still needed as a result.

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
markcowl: Oh, that's right, because you have the feature decorator

specification/network/resource-manager/Microsoft.Network/Network/Network/ConnectionPolicy.tsp
```diff
@@ -19,28 +19,22 @@ namespace Microsoft.Network;
...
-#suppress "@azure-tools/typespec-azure-core/no-private-usage" "Required for includeInapplicableMetadataInPayload which suppresses systemData in the response payload, consistent with all other VirtualWAN nested resources"
```
markcowl, line 22: After discussing with Pan, let's leave in the original version.  We will come back at a later date with a solution that allows you to add new resources without suppressions and private decorators.

markcowl (changes requested):

markcowl (approved):

pshao25 (approved):

PR #43110, opened 2026-05-12
mgmt, Java, Validate TypeSpec conversion for ApiCenter service converted tsp

v-huizhu2:
Validate TypeSpec migration, update tspconfig and client.tsp

https://github.com/Azure/azure-sdk-for-java/issues/49099

## Purpose of this PR

## Due diligence checklist

## Getting help


v-huizhu2:
## Reason for the change (short)

**Change:** `ApiDefinition.importSpecification` switched from `ArmResourceActionAsync<..., void>` to `ArmResourceActionNoResponseContentAsync<...>`.

**Why:**

1. **Lint rule** `arm-post-operation-response-codes` forbids a `200` response on an LRO POST when the final result has no schema. `ArmResourceActionAsync<..., void>` still emits an empty `200`, which violates the rule.
2. `ArmResourceActionNoResponseContentAsync` emits **only** `202 Accepted` + `default` error — exactly what the rule requires for a no-body LRO POST.
3. **Side benefit:** removes `OkResponse` from the operation's return type, so the Java SDK no longer leaks a public `OkResponse` model and `importSpecification` goes back to returning `void` — fixing both Break 1 (return-type change) and Break 2 (`OkResponse` model leak) at the source.
4. The original `2024-03-01` Swagger already returned `200` with no body, so dropping it from the spec is a spec-only cleanup (LRO clients poll via `202` + `Location`, unaffected).

image

v-huizhu2:
Added `"!java"` scope (third argument) to all 10 `@@clientName(..., "payload")` declarations in main.tsp.
The unscoped `@@clientName(..., "payload")` renamed the body parameter to `payload` for **all** SDK languages, including Java. This introduced a breaking change in the Java SDK (v1.2.0-beta.1 vs v1.1.0) — for example, `createOrUpdate(..., ApiDefinitionInner resource)` became `createOrUpdate(..., ApiDefinitionInner payload)`.
By adding `"!java"`, the rename to `payload` now applies to all languages **except** Java. Java falls back to the original TypeSpec parameter name (`resource`, `properties`, or `body`), which matches the previously released Java SDK and avoids a breaking change.
image

PR #43112, opened 2026-05-12
Spec: merge code download endpoints into a single operation

sagunase:
Replaces the two preview-only download operations:
  - GET /agents/{agent_name}/versions/{agent_version}/code:download (downloadAgentVersionCode)
  - GET /agents/{agent_name}/code:download (downloadAgentCode)

with a single operation:
  - GET /agents/{agent_name}/code:download?agent_version=<x> (downloadAgentCode)

When `agent_version` is omitted, the service returns the latest version's code zip; otherwise the specified version's code zip is returned.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
@@ -336,32 +336,14 @@ interface Agents {
   >;
 
   /**
-   * Download the code zip for a specific version of a code-based hosted agent.
+   * Download the code zip for a code-based hosted agent.
    * Returns the previously-uploaded zip (`application/zip`).
-   * The SHA-256 digest of the returned bytes matches the `content_hash` on the agent version's `code_configuration`.
-   */
-  @get
-  @route("/agents/{agent_name}/versions/{agent_version}/code:download")
-  @extension(
-    "x-ms-foundry-meta",
-    #{ required_previews: #[AgentDefinitionOptInKeys.code_agents_v1_preview] }
-  )
-  downloadAgentVersionCode is FoundryDataPlanePreviewOperation<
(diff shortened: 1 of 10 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -372,8 +354,14 @@ interface Agents {
...
+      /** The name of the agent. */
       @path agent_name: string;
+
+      /**
+       * The version of the agent whose code zip should be downloaded.
+       * If omitted, the latest version's code zip is returned.
+       */
+      @query agent_version?: string;
     },
     DownloadAgentCodeResponse
```
dargilco, line 366: Can we add `agent_version` to DownloadAgentCodeResponse? Always populated, but useful when the caller did not specify agent_version in the request.

johanste: That would have to be a response header because I don't think we'd like to inject anything into the returned zip. I don't have an issue with adding it, but I also think we could add it later if there were customer asks.

johanste (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
sagunase: Since the response body is the raw zip (`bytes`), we can't embed `agent_version` in the body. Added it as a response header (`x-ms-agent-version`) on `DownloadAgentCodeResponse` instead, per @johanste's suggestion. See commit c711ff7d48.

dargilco: Sounds good. Thank you!

dargilco (approved):

glecaros (approved):

PR #43122, opened 2026-05-12
Follow-up cleanup of csharp @@clientName overrides for new mgmt generator behavior

haiyuazhang:
Follow-up cleanup of csharp-scoped @@clientName overrides that conflict with the new behavior introduced in azure-typespec/http-client-csharp-mgmt (Azure/azure-sdk-for-net#59096), where csharp-scoped @@clientName on resource update models and list service methods now wins over the generator's auto-derived default.

In each case below, the shipped C# SDK name matched the new default, so the override was causing unintended model / method renames in the version bump PR (Azure/azure-sdk-for-net#59170).

Companion to #43050.

### Changes

- **cdn**: removed 5 `*UpdateParameters` -> * overrides (`FrontDoorCustomDomain/Endpoint/Origin/OriginGroup` and `WebAgentKnowledgeSource`)
- **chaos**: scoped `Experiments.listAll -> `getExperiments` ` from `csharp` to `!csharp`
- **networkfunction**: scoped `azureTrafficCollectorsBySubscriptionList -> List ` to `!csharp`
- **servicenetworking**: removed `AssociationUpdate -> TrafficControllerAssociationUpdate ` override
- **billingbenefits**: removed `SavingsPlanUpdateRequest -> SavingsPlanUpdateContent ` and `DiscountPatchRequest -> DiscountPatchContent ` overrides

### DIFF
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/client.tsp
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/client.tsp
@@ -35,7 +35,7 @@ using Microsoft.Chaos;
   "ExecutionDetails",
   "csharp"
 );
-@@clientName(Microsoft.Chaos.Experiments.listAll, "getExperiments", "csharp");
+@@clientName(Microsoft.Chaos.Experiments.listAll, "getExperiments", "!csharp");
 @@clientName(Microsoft.Chaos.PrivateAccesses.listAll,
   "getPrivateAccesses",
   "csharp"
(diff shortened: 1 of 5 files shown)


specification/chaos/resource-manager/Microsoft.Chaos/Chaos/client.tsp
```diff
@@ -35,7 +35,7 @@ using Microsoft.Chaos;
   "ExecutionDetails",
   "csharp"
 );
-@@clientName(Microsoft.Chaos.Experiments.listAll, "getExperiments", "csharp");
+@@clientName(Microsoft.Chaos.Experiments.listAll, "getExperiments", "!csharp");
```
live1206: should this be removed then?

psah434:
required check failing. Please fix.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/client.tsp
live1206: otherwise, we are renaming for other languages.

live1206 (approved):

PR #43128, opened 2026-05-12
Add C# customizations for SecurityCenter SDK migration

live1206:
## Summary

Adds C# customizations for the `Azure.ResourceManager.SecurityCenter` TypeSpec migration.

## Changes

- Merged latest `main` and resolved the Security `client.tsp` conflict.
- Point C# generation to `sdk/securitycenter` and namespace `Azure.ResourceManager.SecurityCenter`.
- Preserve C# compatibility names and shapes for migrated SecurityCenter resources, models, and enums.
- Move SDK review naming fixes into C#-scoped TypeSpec decorators so the SDK no longer needs broad naming suppressions.
- Exclude Security operation-status route from C# generation with `@@scope(..., "!csharp")` because the SDK is not expected to expose `GetOperationStatus`.
- Set legacy-template `ResourceName` values for multi-path Alert/Task/AssessmentMetadata sources so generated C# resources are source-specific instead of relying on SDK suppressions.
- Reordered custom alert discriminator fields so generated constructors compile without SDK bridge constructors.
- Type JIT `maxRequestAccessDuration` as `duration` so generated C# serialization uses ISO 8601 duration formatting.
- Add replacement-model customizations for Security Solutions compatibility where `@@hierarchyBuilding` is not safe for these shapes.

## Validation

Generated and built the .NET SDK locally from this specs branch with:

```pwsh
pwsh eng/packages/http-client-csharp-mgmt/eng/scripts/RegenSdkLocal.ps1 -Services "Azure.ResourceManager.SecurityCenter" -LocalSpecRepoPath ../azure-rest-api-specs -SaveInputs
dotnet build sdk/securitycenter/Azure.ResourceManager.SecurityCenter/src/Azure.ResourceManager.SecurityCenter.csproj --no-restore /p:UseSharedCompilation=false /p:RunApiCompat=false /p:ValidateRunApiCompat=false -v:minimal
```

Resource hierarchy/ApiCompat parity still has known remaining blockers documented in the linked SDK PR.

## Related

- SDK PR: https://github.com/Azure/azure-sdk-for-net/pull/59180
- Issue: https://github.com/Azure/azure-sdk-for-net/issues/58968

### DIFF
--- a/specification/security/resource-manager/Microsoft.Security/Security/client.tsp
+++ b/specification/security/resource-manager/Microsoft.Security/Security/client.tsp
@@ -30,10 +30,13 @@ import "./SqlVulnerabilityAssessmentsAPI/main.tsp";
 import "./StandardsAPI/main.tsp";
 import "./SubAssessmentsAPI/main.tsp";
 import "./TasksAPI/main.tsp";
+import "./csharp-arm-models.tsp";
+import "@azure-tools/typespec-azure-resource-manager";
 import "@azure-tools/typespec-client-generator-core";
 import "@typespec/versioning";
 
 using Azure.ClientGenerator.Core;
+using Azure.ClientGenerator.Core.Legacy;
 using TypeSpec.Versioning;
 using TypeSpec.Http;
 using TypeSpec.Rest;
@@ -51,6 +54,9 @@ using LegacySettingsAPI;
 using SqlVulnerabilityAssessmentsAPI;
 using AlertsAPI;
 using ApiCollectionsAPI;
+using IoTSecurityAPI;
+using SecurityConnectorsAPI;
+using StandardsAPI;
 
 @client({
   name: "SecurityManagementClient",
@@ -92,9 +98,10 @@ using ApiCollectionsAPI;
 })
 namespace SecurityManagementClient;
 
-#suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-#suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-model `private-link-parameters` {
+/**
+ * A private link resource name parameter.
+ */
+model PrivateLinkNameParameter {
   /**
    * The name of the private link resource.
    */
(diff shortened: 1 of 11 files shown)


JialinHuang803:
The 'endTimeUtc' parameter is missing in the [example](https://github.com/azure/azure-rest-api-specs/blob/main/specification/security/resource-manager/Microsoft.Security/Security/stable/2020-01-01/examples/JitNetworkAccessPolicies/InitiateJitNetworkAccessPolicy_example.json) which causes JS sample generation failure. Please help skip sample generation for JS to unblock the pipeline.

specification/security/resource-manager/Microsoft.Security/Security/tspconfig.yaml
```diff
@@ -27,6 +30,7 @@ options:
       SecurityManagementClient: SecurityCenter
     flavor: "azure"
     compatibility-lro: true
+    generate-sample: false
```
kazrael2119, line 33:
```suggestion
    generate-sample: false
    generate-metadata: true
```

specification/security/resource-manager/Microsoft.Security/Security/client.tsp
```diff
@@ -463,14 +1333,34 @@ op privateLinkResourcesGetCustomization(
...
   AdvancedThreatProtectionSettingNameAlias.settingName,
   "current",
   "python"
 );
-@@Azure.ClientGenerator.Core.alternateType(
+@@alternateType(
   AdvancedThreatProtectionSettingNameAlias.settingName,
-  "current"
+  "current",
+  "python"
```
JialinHuang803: Is scoping to Python expected?

specification/security/resource-manager/Microsoft.Security/Security/client.tsp
```diff
@@ -488,9 +1378,24 @@ op privateLinkResourcesGetCustomization(
...
+  /** Default. */
+  Default: "default",
+}
+
 @@Azure.ClientGenerator.Core.alternateType(
-  ServerVulnerabilityAssessmentNameAlias.serverVulnerabilityAssessment,
-  "default"
+  ServerVulnerabilityAssessmentName,
+  CSharpServerVulnerabilityAssessmentName,
+  "csharp"
```
JialinHuang803, line 2071: Is scoping to csharp expected?

specification/security/resource-manager/Microsoft.Security/Security/tspconfig.yaml
```diff
@@ -27,6 +30,8 @@ options:
...
+    generate-metadata: true
```
kazrael2119:
```suggestion
    generate-sample: false
```

gary-x-li:
Please fix TypeSpec Validation errors.

ArthurMa1978 (approved):

ArthurMa1978:
The TypeSpec validation issue has been fixed, and this change is intended only for generating the .NET SDK.

Issue #42822, opened 2026-05-01
[EngSys] Block PRs adding handwritten swaggers in new API versions


mikeharder:

timotheeguerin:
Closing as completed by https://github.com/Azure/azure-rest-api-specs/pull/43149.

TypeSpec Requirement now checks whether the exact API version already exists in public main, rather than allowing handwritten Swagger whenever the service has any stable version. New handwritten API versions fail unless explicitly suppressed.

PR #43149, opened 2026-05-13
[TypeSpec Requirement] Require TSP for new API versions

mikeharder:
- meaningful changes
  - before: if a spec had **any** stable verion in public/main, it was `brownfield` and TSP was optional
  - after: if a spec has **the exact** API version in public main, it is `brownfield` and TSP s optional
- fixes
  - #42822
- e2e tests
  - fail: https://github.com/mikeharder/azure-rest-api-specs/pull/86
  - pass: https://github.com/mikeharder/azure-rest-api-specs/pull/87
- supercedes
  - #42823


eng/scripts/TypeSpec-Requirement.ps1
```diff
@@ -190,11 +190,11 @@ else {
       }
     }
     elseif ($responseStatus -eq 404) {
-      LogInfo "  Branch 'main' does not contain path '$servicePath/stable', so spec is new and must use TypeSpec"
+      LogInfo "  Branch 'main' does not contain path '$serviceApiVersion', so API version is new and must use TypeSpec"
```
mikeharder:
```suggestion
      LogInfo "  Branch 'main' does not contain path '$apiVersion', so API version is new and must use TypeSpec"
```

benbp (approved):

MattGertz (approved):

PR #43158, opened 2026-05-13
[ProfessionalService] Add GA API version 2026-05-01

MichaelMSWork:
## Summary

Adds the GA API version `2026-05-01` for `Microsoft.ProfessionalService` resource provider, authored in TypeSpec and emitted to Swagger.

## Changes

- New TypeSpec project under ProfessionalService:
  - `main.tsp`, `enums.tsp`, `common.tsp`, professionalServiceResource.tsp, subscriptionLevel.tsp
  - `tspconfig.yaml` with multi-language emitter configuration (Java, TS, Go, Python)
- Generated Swagger at `stable/2026-05-01/professionalservice.json` with examples
- Added `package-2026-05-01` tag in readme.md and made it the default tag

## Operations included (GA scope)

- `ProfessionalServiceSubscriptionLevel`: Get, CreateOrUpdate (LRO), Update/PATCH (LRO), Delete (LRO), ListByResourceGroup, Unsubscribe (LRO action)
- `ProfessionalServiceSubscriptionLevelGlobal`: ListByAzureSubscription, CheckNameAvailability
- `ProfessionalServiceOperationStatuses`: GetSubscriptionOperationStatus
- `Operations`: List

## Notes

- Uses ARM common-types `v6`.

---


psah434 (commented):
## ARM API Review - Microsoft.ProfessionalService `2026-05-01`

Reviewed 30 changed files against Azure REST API Guidelines, ARM RPC rules, and TypeSpec best practices. Compared against the previous version `preview/2023-07-01-preview` (hand-written swagger; this PR is a TypeSpec conversion for GA).

**Breaking changes: Expected** (preview-to-GA migration with TypeSpec conversion)
**New blocking issues: 5** | **New warnings: 4** | **Suggestions: 2**

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
```diff
@@ -0,0 +1,43 @@
...
+
+  /** The feedback. */
+  feedback?: string;
+
+  /** The deletion source. */
+  deletionSource?: DeletionSource = DeletionSource.CustomerCancellation;
+}
+
+/** Check name availability request. */
+model CheckNameAvailabilityRequest {
```
psah434: **[NEW]** :red_circle: **Blocking** **[TSP-5.3 / Common-Types]**

Custom `CheckNameAvailabilityRequest` and `CheckNameAvailabilityResponse` models are defined here instead of using the ARM common-types versions. The `checkGlobalNameAvailability` template in `subscriptionLevel.tsp` (line 54) passes these custom types, which overrides the standard ARM models.

The custom `CheckNameAvailabilityRequest` makes `name` and `type` optional, whereas the ARM standard requires them. This causes the generated swagger to define custom models instead of `$`ref-ing `common-types/resource-management/v6/checkNameAvailability.json`.

**Fix:** Remove these custom models and use the default `checkGlobalNameAvailability` without custom type parameters:

```tsp
// subscriptionLevel.tsp - change to:
checkNameAvailability is checkGlobalNameAvailability;
```

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/examples/2026-05-01/ProfessionalServicePut.json
```diff
@@ -0,0 +1,97 @@
...
+      "location": "global",
+      "tags": {
+        "environment": "test"
+      },
+      "properties": {
+        "description": "Contoso consulting engagement",
+        "quantity": 1,
+        "autoRenew": false,
+        "storeFront": "AzurePortal",
+        "privateOfferId": null,
```
psah434: **[NEW]** :red_circle: **Blocking** **[Section 7 - No Null Values]**

`"privateOfferId": null` appears here and in multiple other example files (PUT lines 57, 90; GET line 44; Unsubscribe; ListBySubscription; ListByResourceGroup). Additionally, `"nextLink": null` appears in list examples.

ARM specs MUST NOT use `null` values in responses. Properties with no value should be omitted entirely.

**Fix:** Remove all `"privateOfferId": null` entries from example responses (omit the property). Remove `"nextLink": null` from list examples (omit when there is no next page).

Applies to both copies: `examples/2026-05-01/` and `stable/2026-05-01/examples/`.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/enums.tsp
```diff
@@ -0,0 +1,63 @@
...
+  string,
+  Azure.ResourceManager.ResourceProvisioningState,
+}
+
+/** The Professional Service Subscription Status. */
+union ProfessionalServiceResourceStatus {
+  string,
+
+  /** Subscription has not started */
+  "NotStarted",
```
psah434: **[NEW]** :red_circle: **Blocking** **[TSP-4.1 / Union Named Members]**

`ProfessionalServiceResourceStatus`, `SubscriptionStatusReason` (line 34), and `PreflightStatus` (line 59) use anonymous string literals (e.g., `"NotStarted"`) instead of named members (e.g., `NotStarted: "NotStarted"`).

This is inconsistent with `DeletionSource` (line 48) which correctly uses named members. Per Azure TypeSpec guidelines, union members MUST have named identifiers for proper SDK code generation.

**Fix:**
```tsp
union ProfessionalServiceResourceStatus {
  string,
  /** Subscription has not started */
  NotStarted: "NotStarted",
  /** Pending fulfillment start */
  PendingFulfillmentStart: "PendingFulfillmentStart",
  // ... same pattern for all members
}
```
Apply to `SubscriptionStatusReason` and `PreflightStatus` as well.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
```diff
@@ -0,0 +1,95 @@
...
+  /** Whether the current term is a Free Trial term. Server-determined. */
+  @visibility(Lifecycle.Read)
+  isFreeTrial?: boolean;
+
+  /** The storefront. */
+  storeFront?: string;
+
+  /** The purchaser email. Server-derived from authenticated identity. */
+  @visibility(Lifecycle.Read)
+  purchaserEmail?: string;
```
psah434, line 86: **[NEW]** :red_circle: **Blocking** **[SEC-SECRET-DETECT / PII]**

`purchaserEmail` is a read-only property that returns a user's email address in GET/PUT/PATCH responses. Email addresses are PII/customer data.

Per ARM RPC rules, customer data SHOULD NOT be stored or returned in control plane resource properties. If this data must be exposed, it should be behind a POST action (e.g., `listDetails`) rather than returned in the standard GET response.

**Fix:** Either remove `purchaserEmail` from the resource properties, or move it behind a POST action. If keeping it in GET is absolutely required by the business, add an explicit justification comment and consider applying `@secret`.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/examples/2026-05-01/ProfessionalServicePut.json
```diff
@@ -0,0 +1,97 @@
...
+        "quoteId": "Q-12345-CONTOSO"
+      }
+    }
+  },
+  "responses": {
+    "201": {
+      "headers": {
+        "Azure-AsyncOperation": "https://management.azure.com/subscriptions/1a950b42-75c2-44a6-ab02-fc0f840997aa/providers/Microsoft.ProfessionalService/operationStatuses/e24b5b6b-2dfc-41ea-c457-46d0c42494af?api-version=2026-05-01",
+        "Retry-After": "10"
+      },
```
psah434, line 28: **[NEW]** :red_circle: **Blocking** **[Example - systemData]**

The PUT 200 and 201 response bodies do not include `systemData`. Since `ProfessionalServiceResource` extends `TrackedResource` from common-types v6 (which includes `systemData`), the example responses should include it for consistency.

The GET example correctly includes `systemData` -- the PUT responses should match.

**Fix:** Add `systemData` to the 200 and 201 response bodies, similar to the GET example.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+  listByResourceGroup is ArmResourceListByParent<ProfessionalServiceResource>;
+  // NOTE: No listAccessToken (GenerateToken) — not supported for Professional Services.
+  unsubscribe is ArmResourceActionAsync<
+    ProfessionalServiceResource,
+    DeleteOptions,
+    ProfessionalServiceResource
+  >;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Custom preflight route not tied to ProfessionalServiceResource"
```
psah434: **[NEW]** :yellow_circle: **Warning** **[TSP-6.1 / Suppressions]**

Three separate `#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator"` directives (lines 30, 51, 60) for `ProfessionalServiceSubscriptionLevelPreflight`, `ProfessionalServiceSubscriptionLevelGlobal`, and `ProfessionalServiceOperationStatuses`.

Multiple suppressions of the same ARM rule suggest the interface structure should be reorganized. Consider whether some operations can be consolidated into the main `ProfessionalServiceSubscriptionLevel` interface or if a different ARM template pattern would avoid the need for suppressions.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
```diff
@@ -0,0 +1,95 @@
...
+  @visibility(Lifecycle.Read)
+  termUnit?: string;
+
+  /** The billing period (ISO 8601 duration). Server-resolved from ProductCode. */
+  @visibility(Lifecycle.Read)
+  billingPeriod?: string;
+
+  /** The risk property bag header (JSON format). Write-only — not returned in GET responses. */
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  riskPropertyBagHeader?: string;
```
psah434, line 54: **[NEW]** :yellow_circle: **Warning** **[SEC-SECRET-DETECT]**

`riskPropertyBagHeader` is described as containing "risk property bag header (JSON format)" with write-only visibility. If this property carries risk assessment data, security tokens, or other sensitive information, it should have the `@secret` decorator to prevent it from appearing in logs or debug output.

**Fix:** If the property contains sensitive data:
```tsp
@secret
@visibility(Lifecycle.Create, Lifecycle.Update)
riskPropertyBagHeader?: string;
```

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
```diff
@@ -0,0 +1,43 @@
...
+  unsubscribeOnly?: boolean;
+
+  /** The reason code. */
+  reasonCode?: int32;
+
+  /** The feedback. */
+  feedback?: string;
+
+  /** The deletion source. */
+  deletionSource?: DeletionSource = DeletionSource.CustomerCancellation;
```
psah434: **[NEW]** :yellow_circle: **Warning** **[Section 8.3 - Default Values]**

`deletionSource?: DeletionSource = DeletionSource.CustomerCancellation` has a default value on an optional property. In the generated swagger, this produces `"default": "CustomerCancellation"`.

Per ARM guidelines, default values in specifications can cause confusion -- the service should handle the default server-side rather than specifying it in the schema.

**Fix:** Remove the default and document the behavior in the doc comment:
```tsp
/** The deletion source. Defaults to CustomerCancellation if not specified. */
deletionSource?: DeletionSource;
```

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+interface ProfessionalServiceSubscriptionLevel {
+  get is ArmResourceRead<ProfessionalServiceResource>;
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<ProfessionalServiceResource>;
+  // Tags-only PATCH per ARM TrackedResource convention.
+  // Business properties are not patchable for Professional Services in GA.
+  updateTags is ArmTagsPatchAsync<ProfessionalServiceResource>;
+  delete is ArmResourceDeleteWithoutOkAsync<ProfessionalServiceResource>;
+  listByResourceGroup is ArmResourceListByParent<ProfessionalServiceResource>;
+  // NOTE: No listAccessToken (GenerateToken) — not supported for Professional Services.
+  unsubscribe is ArmResourceActionAsync<
```
psah434, line 33: **[NEW]** :bulb: **Suggestion**

The `unsubscribe` action will produce a generic description in the generated swagger: `"A long-running resource action."`. Consider adding a custom `@doc` to describe what unsubscribing actually does:

```tsp
/** Unsubscribes from a Professional Service subscription, optionally providing a reason and feedback. */
unsubscribe is ArmResourceActionAsync<...>;
```

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/examples/2026-05-01/ProfessionalServicePut.json
MichaelMSWork: Fixed. All null values removed from example files per ARM Section 7:

TypeSpec/Swagger changes (both SaaS GA + PS GA):

Removed all "propertyName": null entries from example responses — properties with no value are now omitted entirely
Affected: privateOfferId, nextLink, reason, message across GET, PUT, Unsubscribe, List, CheckNameAvailability examples

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
MichaelMSWork: ---

Fixed. Removed custom `CheckNameAvailabilityRequest` and `CheckNameAvailabilityResponse` models from common.tsp. Now using the default ARM common-types models via:

```tsp
checkNameAvailability is checkGlobalNameAvailability;
```

The generated swagger now `$ref`s `common-types/resource-management/v6/types.json#/definitions/CheckNameAvailabilityRequest` and `CheckNameAvailabilityResponse` instead of defining custom models.

---

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/enums.tsp
MichaelMSWork: Fixed. All union members now use named identifiers (Name: "Value") instead of anonymous string literals:

ProfessionalServiceResourceStatus: NotStarted: "NotStarted", PendingFulfillmentStart: "PendingFulfillmentStart", Subscribed: "Subscribed", Unsubscribed: "Unsubscribed", Suspended: "Suspended"
SubscriptionStatusReason: None: "None", SuspendedInternal: "SuspendedInternal", AzureSubscriptionDeletion: "AzureSubscriptionDeletion"
PreflightStatus: Succeeded: "Succeeded", Failed: "Failed"
Consistent with DeletionSource which already used named members.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/examples/2026-05-01/ProfessionalServicePut.json
MichaelMSWork: Fixed. Added systemData to all resource response bodies that were missing it:

PUT: Both 201 and 200 responses now include systemData
Unsubscribe: 200 response now includes systemData
UpdateTags: 200 response now includes systemData
ListByResourceGroup: Response items now include systemData
ListBySubscription: Response items now include systemData
Now consistent with GET example. All JSON validated and recompiled.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
MichaelMSWork: Acknowledged. purchaserEmail is a read-only, server-derived property (set from the authenticated identity at purchase time). We've added @secret to flag it as sensitive data.  Required by portal UX

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
MichaelMSWork: Removed default value from deletionSource property in both PS and SaaS TypeSpec. Default behavior (CustomerCancellation) is now documented in the doc comment and will be handled server-side.



/** The deletion source. Defaults to CustomerCancellation if not specified. */
deletionSource?: DeletionSource;

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
MichaelMSWork: Added @secret to riskPropertyBagHeader in both PS and SaaS TypeSpec. This property carries fraud-prevention/risk assessment data forwarded to the payment system — it's already write-only (@visibility(Lifecycle.Create, Lifecycle.Update)) and should not appear in logs or debug output. 

@secret
@visibility(Lifecycle.Create, Lifecycle.Update)
riskPropertyBagHeader?: string;

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: Added @doc to the unsubscribe action in both PS and SaaS TypeSpec:

/** Unsubscribes from a Professional Service subscription, optionally providing a reason and feedback. The resource remains in Azure but the subscription is cancelled. */

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: These three suppressions are intentional and has a distinct architectural reason:

ProfessionalServiceSubscriptionLevelPreflight — Uses a custom @route with @post because the preflight path (/deployments/{resourceName}/preflight) doesn't follow the standard ARM resource CRUD pattern. ARM resource interfaces require @armResourceOperations decorator which enforces resource-scoped routing — preflight is a deployment-scoped action, not resource-scoped.

ProfessionalServiceSubscriptionLevelGlobal — Contains listByAzureSubscription and checkNameAvailability which operate at subscription scope, not resource group scope. These can't be in the main interface because @armResourceOperations scopes all operations to the resource's parent (resource group), but these are subscription-level global operations.

ProfessionalServiceOperationStatuses — Uses a custom @route for the operationStatuses polling endpoint, which is a separate ARM path (/providers/Microsoft.ProfessionalService/operationStatuses/{operationId}) not tied to the resource CRUD lifecycle.

MichaelMSWork:
All review feedback has been addressed and changes pushed. Please remove the `ARMChangesRequested` label to proceed with ARM review. Thank you.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+    ProfessionalServiceResource
+  >;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Custom preflight route not tied to ProfessionalServiceResource"
+interface ProfessionalServiceSubscriptionLevelPreflight {
+  /** Pre-validates Professional Service subscription operations. */
+  #suppress "@azure-tools/typespec-azure-core/no-unknown" "Preflight accepts dynamic payload"
+  @post
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ProfessionalService/deployments/{resourceName}/preflight")
```
ravimeda: **[NEW]** :red_circle: **Blocking** **[ARM Resource Modeling / RPC-Uri-V1]**

The custom `preflight` route is hand-written under a `deployments/{resourceName}` segment, but the tracked resource is declared with `SegmentName = "resources"` (in `professionalServiceResource.tsp`). No `deployments` resource type is declared anywhere in the spec, so this path uses a provider segment that does not map to a registered resource type. ARM requires every `/providers/{ns}/{type}/{name}` segment to correspond to a declared resource type, and ARG/SDK/RPaaS routing all key off that mapping.

**Suggested fix:** Move the preflight onto the actual resource path — `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ProfessionalService/resources/{resourceName}/preflight` — and ideally model it via `ArmResourceActionAsync<ProfessionalServiceResource, PreflightRequest, PreflightResult>` (or `ArmResourceActionSync` if the operation is synchronous) so the route is generated by the ARM library instead of being suppressed. If `deployments` is genuinely a separate resource type, declare it as one with its own CRUD surface.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+    ...Azure.ResourceManager.CommonTypes.SubscriptionIdParameter,
+    ...Azure.ResourceManager.CommonTypes.ResourceGroupNameParameter,
+
+    /** The resource name. */
+    @pattern("^[a-zA-Z0-9-]{3,50}$")
+    @path
+    resourceName: string,
+
+    /** The preflight request body. */
+    @body body: unknown,
```
ravimeda: **[NEW]** :red_circle: **Blocking** **[OpenAPI - Request Body Schema]**

The preflight body is typed `unknown`, which the autorest emitter renders as `"schema": {}` (no schema). An empty/unknown body is non-conformant for a GA OpenAPI surface: SDK generators cannot model the payload, and downstream tooling (linter, ARM swagger validation, RPaaS) treats this as an untyped contract. The current `no-unknown` suppression with reason "Preflight accepts dynamic payload" is too generic to justify shipping an untyped request body in a GA API.

**Suggested fix:** Define an explicit `PreflightRequest` model that lists the fields preflight actually validates (resource name, productCode, quantity, quoteId, etc.). If the contract truly must be open-ended, model it as `Record<unknown>` so it emits `type: object, additionalProperties: {}` rather than an empty schema, and tighten the suppression reason to describe why an open contract is required.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
```diff
@@ -0,0 +1,22 @@
+namespace Microsoft.ProfessionalService;
+
+/** Delete options for a Professional Service resource. */
+model DeleteOptions {
```
ravimeda: **[NEW]** :yellow_circle: **Warning** **[Section 6 - Naming]**

`DeleteOptions` is used as the request body of the `unsubscribe` **action**, not of `delete`. SDK signatures will read as `client.unsubscribe(deleteOptions)`, which is misleading — the model has nothing to do with the DELETE operation, and `delete` takes no body. Renaming after GA is a breaking change for SDK callers, so it is worth fixing now.

**Suggested fix:** Rename to `UnsubscribeOptions` (or `UnsubscribeRequest`) and update the reference in `subscriptionLevel.tsp` line 24-28 and the `ProfessionalServiceUnsubscribe.json` example.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
```diff
@@ -0,0 +1,22 @@
...
+  /** Whether it is unsubscribe only. */
+  unsubscribeOnly?: boolean;
+
+  /** The reason code. */
+  reasonCode?: int32;
```
ravimeda, line 26: **[NEW]** :yellow_circle: **Warning** **[Enum vs. Opaque Integer]**

`reasonCode` is an opaque `int32` with no documentation of allowed values. Numeric reason codes are an anti-pattern in Azure APIs: callers must discover values out of band, SDKs lose type safety, ARG/telemetry cannot group by symbolic name, and changes to the value space cannot be expressed in OpenAPI. The freeform `feedback` string is a fine companion, but the code itself should be enumerated.

**Suggested fix:** Replace `reasonCode?: int32` with an extensible string union, e.g.

```tsp
union UnsubscribeReason {
  string,
  NoLongerNeeded: "NoLongerNeeded",
  TooExpensive: "TooExpensive",
  SwitchedProduct: "SwitchedProduct",
  TechnicalIssues: "TechnicalIssues",
  Other: "Other",
}
```

and rename the property to `reason`.

ravimeda (commented):
Automated ARM API review (continuation). Earlier findings already covered by existing review comments (`purchaserEmail` secret/PII, `riskPropertyBagHeader` secret, suppression hygiene on `subscriptionLevel.tsp` line 31) are not re-posted to avoid duplicate threads. Seven additional findings below — please see each inline comment for details.

*Separately, the PR description advertises `ProfessionalServiceSubscriptionLevelMove: MoveResources, ValidateMoveResources` but those operations are not present in the TypeSpec source or emitted swagger. Please reconcile the description with the spec.*

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Microsoft.ProfessionalService;
+
+@armResourceOperations
+interface ProfessionalServiceSubscriptionLevel {
+  get is ArmResourceRead<ProfessionalServiceResource>;
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<ProfessionalServiceResource>;
+  // Tags-only PATCH per ARM TrackedResource convention.
+  // Business properties are not patchable for Professional Services in GA.
+  updateTags is ArmTagsPatchAsync<ProfessionalServiceResource>;
```
ravimeda: **[NEW]** :yellow_circle: **Warning** **[ARM PATCH semantics]**

`ArmTagsPatchAsync` emits an LRO PATCH (`x-ms-long-running-operation: true`, 200/202 responses) for a tags-only update. Updating tags is metadata-only and is conventionally a synchronous operation across ARM RPs; making it async forces every SDK caller to poll for completion on a trivial change and complicates the user experience.

**Suggested fix:** Use `ArmTagsPatchSync<ProfessionalServiceResource>` unless the back end genuinely needs to defer tag updates. If async is required for a documented reason, please add a code comment explaining why.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
```diff
@@ -0,0 +1,97 @@
...
+  purchaserEmail?: string;
+
+  /** The private offer id. */
+  privateOfferId?: string;
+
+  /** The product code. Primary purchase identifier for GA — used to resolve term details server-side. */
+  productCode?: string;
+
+  /** The quote id (mandatory for Professional Service purchases). */
+  quoteId?: string;
```
ravimeda: **[NEW]** :yellow_circle: **Warning** **[Required fields + Create-only identity properties]**

Two issues on the purchase-identity properties (`productCode` line 83 and `quoteId` line 86):

1. **Required:** The doc comments describe `productCode` as the "Primary purchase identifier for GA — used to resolve term details server-side" and `quoteId` as "(mandatory for Professional Service purchases)", but neither is marked required at the TypeSpec level. Adding `required` after GA is a breaking change, so the correctness of the constraint must be settled before shipping.

2. **Mutability:** These are purchase identifiers that resolve server-side metadata (publisher, plan, offer, term). Permitting them to change on a subsequent PUT effectively re-purposes the resource and is inconsistent with how purchase records work elsewhere in Azure marketplace surfaces.

**Suggested fix:** If the properties are mandatory at create, drop the `?` (and update descriptions). If they must not change post-create, add `@visibility(Lifecycle.Read, Lifecycle.Create)`. Please apply the same review to `privateOfferId` (line 80) and `storeFront` (line 72).

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
```diff
@@ -0,0 +1,72 @@
...
+  listByAzureSubscription is ArmListBySubscription<ProfessionalServiceResource>;
+  checkNameAvailability is checkGlobalNameAvailability;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Custom operation status routes"
+interface ProfessionalServiceOperationStatuses {
+  /** Gets the status of a long-running Professional Service operation at the subscription scope. */
+  @get
+  @route("/subscriptions/{subscriptionId}/providers/Microsoft.ProfessionalService/operationStatuses/{operationId}")
+  getSubscriptionOperationStatus(
```
ravimeda, line 54: **[NEW]** :yellow_circle: **Warning** **[LRO operation status scope]**

CreateOrUpdate, UpdateTags, Delete, and Unsubscribe are all declared at resource-group scope and return `Azure-AsyncOperation` / `Location` headers, but the only `operationStatuses` GET declared in the spec is at **subscription** scope (`/subscriptions/{subscriptionId}/providers/Microsoft.ProfessionalService/operationStatuses/{operationId}`). The conventional ARM pattern is to expose an RG-scoped (and optionally location-scoped) `operationStatuses/{operationId}` GET so the async-operation URI for RG-scoped operations stays at RG scope.

**Suggested fix:** Add an RG-scoped operation status route, e.g.

```tsp
@get
@route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ProfessionalService/operationStatuses/{operationId}")
getOperationStatus(...)
```

or confirm that the RP will always route async responses through the subscription-scope endpoint and document that explicitly.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: Good catch — deployments is not a declared resource type because preflight is an internal ARM-to-RP callback, part of the RP Contract. ARM calls this route during template deployment validation (what-if). The RP registers for it via ProviderHub preflightSupported flag.

This is the same pattern as moveResources / validateMoveResources which we already excluded from the spec. I've removed preflight from TypeSpec — it's not a customer-facing API.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: Good catch — deployments is not a declared resource type because preflight is an internal ARM-to-RP callback, part of the RP Contract. ARM calls this route during template deployment validation (what-if). The RP registers for it via ProviderHub preflightSupported flag.

This is the same pattern as moveResources / validateMoveResources which we already excluded from the spec. I've removed preflight from TypeSpec — it's not a customer-facing API.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
MichaelMSWork: Fixed as suggested

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/common.tsp
MichaelMSWork: Keeping reasonCode as int32.

These values originate from the Azure Portal's cancellation dialog — the Portal sends a numeric key corresponding to the user's selected cancellation reason. The mapping is:

Value	SaaS Label	PS Label
0	"I don't need this service any more"	"I no longer need this Professional Service"
1	"I was not happy with its functionality"	"I purchased by mistake"
2	"I didn't understand what it does"	"I want to change the private offer"
3	"The price was too high for me"	—
4	"This is a test subscription"	—
100	"Other"	"Other"

Converting to a string union would be a breaking change for the Portal, which is the sole caller of this field today (confirmed via telemetry — 100% of non-synthetic traffic comes from the Portal app)

We'd prefer to keep this as int32 and document the mapping rather than introduce a breaking wire-format change that requires Portal coordination.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: Keeping ArmTagsPatchAsync.

Our tags update is genuinely asynchronous. The RP needs to propagate tag changes to downstream dependencies. The flow is:

Accept the PATCH request
Dispatch the update to the downstream service
Return 202 with an Azure-AsyncOperation header
Receive an async Service Bus event indicating success or failure
Complete the long-running operation
Because the downstream communication is fire-and-wait (not request-response), we cannot return a synchronous 200 without risking reporting success before the update has actually been applied. The LRO pattern lets callers poll for the real outcome.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/subscriptionLevel.tsp
MichaelMSWork: This is intentional — operation state is tracked at the subscription level internally (operations are keyed by subscription ID + operation ID, not by resource group). Adding an RG-scoped route would be a passthrough to the same backend lookup and would add surface area without functional benefit. 

Existing clients (including Azure Portal) are already implemented against this subscription-scope polling URL.

We've added a doc comment in the TypeSpec clarifying this is by design.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
MichaelMSWork: changed to ArmCustomPatchAsync, quoteId removed as its unneeded property.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/stable/2026-05-01/examples/ProfessionalServicePreflight.json
```diff
@@ -0,0 +1,18 @@
+{
```
ravimeda: **[NEW]** :yellow_circle: **Warning** **[EX-001 - Orphan example file]**

This example file documents a preflight operation that is not defined in this version's swagger (stable/2026-05-01/professionalservice.json) or in the TypeSpec source (subscriptionLevel.tsp). Earlier reviewer threads suggested preflight was removed as an internal ARM-to-RP callback, but the example file was left behind.

**Fix:** Delete this file (and ProfessionalServiceUpdateTags.json), or restore the corresponding operations in the TypeSpec source and regenerate.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/readme.md
```diff
@@ -26,7 +26,16 @@ These are the global settings for the professionalservice.
 
 ```yaml
```
ravimeda, line 27: **[NEW]** :yellow_circle: **Warning** **[Suppression continuity]**

This PR adds package-2026-05-01 as the default tag. No suppression-continuity analysis is visible in the PR — please confirm:

1. Any suppressions in the previous package-2023-07-01-preview tag (or in service-level suppressions.yaml) that still apply are carried forward.
2. Any preview-era suppressions that were dropped were dropped because the underlying violation was actually fixed in GA — not silently omitted.
3. Any net-new suppressions have a specific 
eason field (not boilerplate).

This is a soft check — CI will surface any concrete regressions, but reviewers cannot tell from this PR alone whether the suppression delta is intentional.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
```diff
@@ -0,0 +1,94 @@
...
+
+  /** Whether the Professional Service subscription will auto renew upon term end. */
+  autoRenew?: boolean;
+
+  /** Whether the current term is a Free Trial term. Server-determined. */
+  @visibility(Lifecycle.Read)
+  isFreeTrial?: boolean;
+
+  /** The storefront. */
+  storeFront?: string;
```
ravimeda: **[NEW]** :bulb: **Suggestion** **[Naming]**

`storefront` is a single English word; camelCase should be storefront. The doc-comment on the line above already spells it as one word ("The storefront.").

**Fix:** Rename property to storefront.

ravimeda:
### API Review Summary (residual findings)

Most of the items I flagged are already covered by existing reviewer threads on this PR (purchaserEmail, productCode mutability, riskPropertyBagHeader, unsubscribe shape). Three small residual items posted as inline comments:

- :yellow_circle: Orphan example files: ProfessionalServicePreflight.json and ProfessionalServiceUpdateTags.json reference operations that do not exist in the current swagger/TSP.
- :yellow_circle: Suppression continuity vs. the preview tag has not been documented.
- :bulb: `storeFront` → `storefront` (single English word).

**PR description vs. implementation (informational):** The PR description lists `MoveResources`, `ValidateMoveResources`, `Preflight`, and "no PATCH" — none of those match the current swagger (no Move/Preflight ops; PATCH is present). Please update the description so reviewers and SDK consumers see the actual GA surface.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/stable/2026-05-01/examples/ProfessionalServiceUpdateTags.json
```diff
@@ -0,0 +1,64 @@
+{
```
ravimeda: **[NEW]** :yellow_circle: **Warning** **[EX-001 - Orphan example file]**

This example file documents an `updateTags`/dedicated tags-PATCH operation that is not defined in this version's swagger or in the TypeSpec source. The only PATCH on the resource is the standard `ProfessionalServiceSubscriptionLevel_Update` (which already handles tags).

**Fix:** Delete this file, or add the corresponding operation to the TypeSpec source and regenerate.

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/professionalServiceResource.tsp
MichaelMSWork: fixed as suggested

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/stable/2026-05-01/examples/ProfessionalServicePreflight.json
MichaelMSWork: Fixed as suggested

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/stable/2026-05-01/examples/ProfessionalServiceUpdateTags.json
MichaelMSWork: Fixed as suggested

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/readme.md
MichaelMSWork: 1. The previous package-2023-07-01-preview tag had exactly one suppression: TypeSpecRequirement (brownfield service). This suppression is carried forward — the preview swagger is still served and will remain active until client traffic drops to zero. We are not forcing clients to migrate to the new GA version.
2. The new package-2026-05-01 stable tag is authored in TypeSpec, so TypeSpecRequirement does not apply and no suppression is needed for it.
3. No suppressions were silently dropped — TypeSpecRequirement was the only one in the preview era, and it is retained.
4. No net-new suppressions were added for the stable spec.
5. Updated suppressions.yaml with continuity notes and a specific reason field explaining the above. See latest commit.

ravimeda (approved):

specification/professionalservice/resource-manager/Microsoft.ProfessionalService/ProfessionalService/tspconfig.yaml
```diff
@@ -0,0 +1,40 @@
...
+    default: "sdk/professionalservice"
+emit:
+  - "@azure-tools/typespec-autorest"
+options:
+  "@azure-tools/typespec-autorest":
+    use-read-only-status-schema: true
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/professionalservice.json"
+    arm-types-dir: "{project-root}/../../../../common-types/resource-management"
+  "@azure-tools/typespec-java":
```
ArthurMa1978, line 12: Please add the emitter setting for .net:
```
       namespace: "Azure.ResourceManager.ProfessionalService"
```

PR #43178, opened 2026-05-13
Add etag property to AzureMonitorWorkspaceResource

GokulPrasad-Work:

## Purpose of this PR

  - [x] Update existing version to fix OpenAPI spec quality issues in S360.
  
The `etag` property was present in older swagger versions (2021-06-03-preview, 2023-04-03, 2023-10-01-preview) but was lost during the TypeSpec conversion. This PR restores it using `Azure.ResourceManager.Legacy.EntityTagProperty` to preserve the lowercase `etag` format (matching the original swagger) and avoid breaking changes for existing SDK consumers.

 The versioning exception was approved by @mikekistler who added the Versioning-Approved-BranchPolicyException label. The exception covers all three affected version folders: stable/2025-10-03, preview/2025-10-03-preview, and preview/2025-05-03-preview. The etag property existed in pre-TypeSpec versions (2021-06-03-preview, 2023-04-03, 2023-10-01-preview) and was inadvertently lost during the TypeSpec conversion.
 
### Changes
- Added `...Azure.ResourceManager.Legacy.EntityTagProperty` to `AzureMonitorWorkspaceResource` in TypeSpec `models.tsp`
- Updated OAV examples with `etag` in response bodies across all API versions (2025-05-03-preview, 2025-10-03-preview, 2025-10-03)
- Regenerated swagger files via `tsp compile`

## Due diligence checklist

  [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md).
- [x] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created.


specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/azuremonitorworkspace.json
```diff
@@ -1518,6 +1518,11 @@
         "identity": {
           "$ref": "../../../../../../common-types/resource-management/v6/managedidentity.json#/definitions/ManagedServiceIdentity",
           "description": "The managed service identities assigned to this resource."
+        },
+        "etag": {
+          "type": "string",
+          "description": "\"If etag is provided in the response body, it may also be provided as a header per the normal etag convention.  Entity tags are used for comparing two or more entities from the same requested resource. HTTP/1.1 uses entity tags in the etag (section 14.19), If-Match (section 14.24), If-None-Match (section 14.26), and If-Range (section 14.27) header fields.\")",
```
ravimeda, line 1524: **[NEW] 🔴 Blocking** **[DESC-MALFORMED]** — The generated description for `etag` is malformed: it starts with a literal escaped quote (`\"`) and ends with `\")` (stray closing parenthesis). This text comes verbatim from `Azure.ResourceManager.Legacy.EntityTagProperty` in [legacy-types/resource.tsp](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/legacy-types/resource.tsp#L54), whose doc comment itself contains the bogus quotes/paren. It will be rendered verbatim in OpenAPI, SDK reference docs, and APIView.

**Same issue in:**
- [preview/2025-10-03-preview/azuremonitorworkspace.json](https://github.com/Azure/azure-rest-api-specs/blob/0f6e9449f47da92f7dfc8b7bf91d5398fdd55056/specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/preview/2025-10-03-preview/azuremonitorworkspace.json#L1520) — line 1520
- [preview/2025-05-03-preview/azuremonitorworkspace.json](https://github.com/Azure/azure-rest-api-specs/blob/0f6e9449f47da92f7dfc8b7bf91d5398fdd55056/specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/preview/2025-05-03-preview/azuremonitorworkspace.json#L3271) — line 3271

**Classification:** `etag` did not exist in any of these three versions on `main` — it is a brand-new property added by this PR, so this is **[NEW]**.

**Recommended fix** — define `etag` inline in TypeSpec instead of spreading the buggy legacy model. This drops the `no-legacy-usage` suppression entirely, gives you a clean description, and still emits lowercase `etag`:

```tsp
@doc("If etag is provided in the response body, it may also be provided as a header per the normal etag convention. Entity tags are used for comparing two or more entities from the same requested resource. HTTP/1.1 uses entity tags in the etag (section 14.19), If-Match (section 14.24), If-None-Match (section 14.26), and If-Range (section 14.27) header fields.")
@visibility(Lifecycle.Read)
etag?: string;
```

(Alternatively, file an upstream bug to fix the doc string in `typespec-azure`'s `Legacy.EntityTagProperty`, but that will not unblock this PR.)

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/azuremonitorworkspace.json
```diff
@@ -1518,6 +1518,11 @@
         "identity": {
           "$ref": "../../../../../../common-types/resource-management/v6/managedidentity.json#/definitions/ManagedServiceIdentity",
           "description": "The managed service identities assigned to this resource."
+        },
+        "etag": {
```
ravimeda, line 1522: **[NEW] :yellow_circle: Warning** **[VERSIONING-IMMUTABILITY]** -- Adding a new envelope property to the released **stable** API version `2025-10-03` (and to two previously published preview versions) violates the [API version immutability principle](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/uniform-versioning.md). The `Versioning-Approved-BranchPolicyException` label on this PR confirms an exception has been granted, but the PR description does not document who approved it or link to the approval record.

**Classification:** [NEW] -- this property is being introduced into already-published version folders by this PR.

The change itself is acceptable in shape (additive, optional, `readOnly`, restoring behavior that existed in pre-TypeSpec versions 2021-06-03-preview / 2023-04-03 / 2023-10-01-preview). What is missing is the audit trail.

**Fix:** Update the PR description to:
- Link or reference the versioning exception approval (work item / Teams thread / email / aka.ms link).
- State explicitly that the exception covers all three affected version folders (`stable/2025-10-03`, `preview/2025-10-03-preview`, `preview/2025-05-03-preview`).

This makes future audits of the immutability exception traceable.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/typespec/models.tsp
```diff
@@ -141,6 +141,8 @@ namespace Microsoft.Monitor {
       NamePattern = "^(?!-)[a-zA-Z0-9-]+[^-]$"
     >;
     ...Azure.ResourceManager.ManagedServiceIdentityProperty;
+    #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Preserving lowercase 'etag' from original swagger to avoid breaking change"
+    ...Azure.ResourceManager.Legacy.EntityTagProperty;
```
ravimeda, line 145: **[NEW] :bulb: Suggestion** **[TSP-VERSIONING-INTENT]** -- The neighboring `identity` property in this same model is gated with `@@TypeSpec.Versioning.added(...identity, Microsoft.Monitor.Versions.v2025_10_03_stable)`, but this newly-added `etag` spread has no `@@added` directive, so it is back-applied to all three emitted versions (`v2025_05_03`, `v2025_10_03`, `v2025_10_03_stable`).

**Classification:** [NEW] -- the spread itself is new.

That is correctly aligned with the PR's intent ("restore the `etag` property that was lost during TypeSpec conversion across every version we emit"), but the contrast with `identity` makes the lack of `@@added` look like an oversight at first glance for future readers.

**Suggested fix:** Add a brief comment near the spread explaining the intentional global addition:

```tsp
// etag is intentionally added to all emitted versions to restore the envelope
// property that existed in pre-TypeSpec versions (2021-06-03-preview, 2023-04-03,
// 2023-10-01-preview) and was lost during conversion. Versioning exception approved.
#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Preserving lowercase 'etag' from original swagger to avoid breaking change"
...Azure.ResourceManager.Legacy.EntityTagProperty;
```

(If you adopt the fix from finding #1 and inline `etag` instead of spreading the legacy model, this suppression goes away entirely and the comment can describe the inline property instead.)

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/examples/AzureMonitorWorkspaces_Get_MaximumSet_Gen.json
```diff
@@ -77,7 +77,8 @@
           "lastModifiedBy": "user2",
           "lastModifiedByType": "User",
           "lastModifiedAt": "2021-04-02T12:34:56.1234567Z"
-        }
+        },
+        "etag": "0000000-0000-0000-0000-000000000000"
```
ravimeda: **[NEW] :bulb: Suggestion** **[EXAMPLE-VALUE]** -- All 30 modified example files use `"etag": "0000000-0000-0000-0000-000000000000"`. The first segment is **7 zeros**, not 8, so this is neither a valid UUID nor an RFC 7232 ETag (which would be a quoted string like `"\"675af34563dc-tr34\""` or weak `W/"..."`).

**Classification:** [NEW] -- these example values were added by this PR.

`etag` is just `string` so any value compiles, but SDK consumers and docs readers copy these placeholders into their own code/tests, and a malformed UUID-looking value is misleading.

**Suggested fix** -- replace the placeholder across all 30 modified example files (5 examples x 3 versions x 2 locations) with one of:
- A canonical RFC 7232 strong ETag: `"\"675af34563dc-tr34\""`
- A canonical weak ETag: `"W/\"675af34563dc-tr34\""`
- A correctly-formatted UUID placeholder (8 zeros in segment 1): `"00000000-0000-0000-0000-000000000000"`

Pick one and apply it consistently.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/typespec/models.tsp
```diff
@@ -141,6 +141,8 @@ namespace Microsoft.Monitor {
       NamePattern = "^(?!-)[a-zA-Z0-9-]+[^-]$"
     >;
     ...Azure.ResourceManager.ManagedServiceIdentityProperty;
+    #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Preserving lowercase 'etag' from original swagger to avoid breaking change"
```
ravimeda, line 144: **[NEW] :bulb: Suggestion** **[NAMING-CONFIRMATION]** -- Modern Azure conventions emit camelCase `eTag` (the standard `Azure.ResourceManager.EntityTagProperty` model and `Azure.Core.eTag` scalar both produce `eTag`). This PR explicitly uses `Azure.ResourceManager.Legacy.EntityTagProperty` to preserve lowercase `etag` to match older swagger versions and avoid an SDK breaking change.

**Classification:** [NEW] -- the suppression and the legacy spread are introduced by this PR.

The suppression rationale ("Preserving lowercase `etag` from original swagger to avoid breaking change") is the right justification **if** downstream consumers actually depend on the lowercase casing.

**Action:** Confirm with the SDK / CLI / portal teams that they are reading lowercase `etag`. If no consumer is locked in yet (the previews especially are still mutable in principle), prefer the standard camelCase `eTag` via `Azure.ResourceManager.EntityTagProperty` and drop both this `no-legacy-usage` suppression and the legacy import. If the casing is already locked in by SDK consumers, no change required and this comment can be resolved.

GokulPrasad-Work: The property existed as etag (lowercase) in GA version 2023-04-03 and earlier versions. Existing SDKs already expose this casing. Switching to camelCase eTag would be a breaking change in both the wire format and SDK surface. Service is also returning the etag property in all lowercase. @ravimeda  - Please suggest how should this be handled for this new GA version.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/azuremonitorworkspace.json
GokulPrasad-Work: Updated the PR description with the details.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/azuremonitorworkspace.json
GokulPrasad-Work: We're keeping Legacy.EntityTagProperty to stay consistent with the standard ARM common types rather than inlining and maintaining a custom definition on our side. Happy to revisit if you feel strongly this is a blocker and ok to maintain custom definition at our end.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/stable/2025-10-03/examples/AzureMonitorWorkspaces_Get_MaximumSet_Gen.json
GokulPrasad-Work: Updated the example values.

specification/monitoringservice/resource-manager/Microsoft.Monitor/Accounts/typespec/models.tsp
GokulPrasad-Work: It is present in old versions, but got missed during the typespec migration. So it is different from the identity property.

liadtal (approved):

PR #43183, opened 2026-05-13
Tsv all concurency and better logging

timotheeguerin:
1. Add `--all` to tsv. `npx tsv --all` instead of separate pwsh script
1. Run in concurrency
2. Show the status of each project on the group
image
image

### DIFF
--- a/.github/workflows/typespec-validation-all.yaml
+++ b/.github/workflows/typespec-validation-all.yaml
@@ -53,8 +53,8 @@ jobs:
         # the GitHub Jobs UI.
         ref: ${{ fromJSON(github.event_name == 'schedule' && '["default", "next"]' || '["default"]') }}
         os: [ubuntu, windows]
-        # shards must start at 0 and increment by 1
-        shard: [0, 1, 2]
+        # shards are 1-based
+        shard: [1, 2, 3]
         # total-shards must be an accurate count of the number of shards
         total-shards: [3]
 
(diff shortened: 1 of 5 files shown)


eng/tools/typespec-validation/src/run-all.ts
```diff
@@ -0,0 +1,164 @@
...
+  const failed = failureCount > 0 ? pc.bold(pc.red(`${failureCount} failed`)) : undefined;
+  const skipped = skippedCount > 0 ? pc.bold(pc.gray(`${skippedCount} skipped`)) : undefined;
+  console.log(
+    [passed, failed, skipped].filter(Boolean).join(pc.gray(" | ")),
+    pc.gray(`(${projectDirs.length})`),
+  );
+
+  if (failureCount > 0) {
+    console.log("\nFailed folders:");
+    failedProjects.forEach((x) => console.log(`  - ${x}`));
```
timotheeguerin, line 94: we might want to log the tsv commands to run instead like it was before

.github/workflows/typespec-validation-all.yaml
```diff
@@ -74,15 +74,4 @@ jobs:
...
-          $ErrorActionPreference = 'Continue'
-
-          ./eng/scripts/TypeSpec-Validation.ps1 `
-            -Shard ${{ matrix.shard }} `
-            -TotalShards ${{ matrix.total-shards }} `
-            -CheckAll `
-            -GitClean `
-            -Verbose
-        shell: pwsh
+        run: npx tsv --all --shard=${{ matrix.shard }}/${{ matrix.total-shards }} specification
```
mikeharder, line 77:
```suggestion
        run: npm exec --no -- tsv --all --shard=${{ matrix.shard }}/${{ matrix.total-shards }} specification
```

always use exact syntax `npm exec --no --` instead of `npx` in CI (for security)

eng/tools/typespec-validation/src/concurrency.ts
mikeharder, line 1: this looks like a utility that should move under `.github/shared`

eng/tools/typespec-validation/src/run-all.ts
```diff
@@ -0,0 +1,164 @@
...
+import pc from "picocolors";
+import { getSuppressions as getSuppressionsRaw, Suppression } from "suppressions";
+import { promisify } from "util";
+import { runWithConcurrency } from "./concurrency.js";
+import { TaskRunner } from "./runner.js";
+import { normalizePath } from "./utils.js";
+
+const execFileAsync = promisify(execFile);
+
+const CONCURRENCY = cpus().length;
```
mikeharder: try a higher number like `NUM_CORES * 2`, which is often superior for workloads mixing CPU and IO

PR #43220, opened 2026-05-14
Update tsv linting to allow generic suppression

timotheeguerin:
When running tsv all this create output in `/tsp-output/@azure-tools/typespec-autorest`

As far as I understand those tspconfig are only for client splitting and shouldn't have an autorest output (no options configured which is why it was emitting in that location)

Had to add a suppression and handling


eng/tools/typespec-validation/src/rules/emit-autorest.ts
```diff
@@ -2,14 +2,24 @@ import { join } from "path";
...
-import { fileExists, readTspConfig } from "../utils.js";
+import { fileExists, getSuppressions, readTspConfig } from "../utils.js";
 
 export class EmitAutorestRule implements Rule {
   readonly name = "EmitAutorest";
 
   readonly description = 'Must emit "@azure-tools/typespec-autorest" by default';
 
   async execute(folder: string): Promise<RuleResult> {
+    const suppressions = (await getSuppressions(folder)).filter((s) =>
```
timotheeguerin: @mikeharder, copilot added that to make the suppression worked but I thought the engine was allowing it to suppress rules already automatically

mikeharder: currently, engine only supports whole-tool suppression.  per-rule (or per-sub-rule) suppressions are up to each rule.

i think the copilot code here is slightly overkill, you should only need to check for per-rule suppressions, a suppression with no rules is handled by engine.

if you don't mind, I'd have copilot refactor the engine to support per-rule suppressions centrally, and only have the rule itself read suppressions if needed for more specific validation.  Insert the new code around here:

https://github.com/Azure/azure-rest-api-specs/blob/54c5f2dab51f23e9d256797b5293bb17bcafe0f8/eng/tools/typespec-validation/src/index.ts#L70-L73

mikeharder (approved):

eng/tools/typespec-validation/src/rules/folder-structure.ts
```diff
@@ -14,19 +14,12 @@ debug.enable("simple-git");
 export class FolderStructureRule implements Rule {
   readonly name = "FolderStructure";
   readonly description = "Verify spec directory's folder structure and naming conventions.";
+  readonly suppressable = true;
   async execute(folder: string): Promise<RuleResult> {
     const suppressions = (await getSuppressions(folder)).filter((s) =>
```
mikeharder: i think this can be deferred to immediately before the first use, to improve perf if the suppressions is never even needed

eng/tools/typespec-validation/src/rules/compile.ts
```diff
@@ -34,14 +34,6 @@ export class CompileRule implements Rule {
       // Rule output is easier to read if "tsp compile" stderr is redirected to stdOutput
       stdOutput += stderr;
 
-      if (
```
mikeharder, line 37: why remove this error condition?

mikeharder (changes requested):
LGTM just one question if you still intend to remove the error if no emit autorest

eng/tools/typespec-validation/src/rules/compile.ts
timotheeguerin: when you don't have any emitter configured like those specs then this extra check makes the compile rule fail. But I do not want to suppress compile, i still want to verify the typespec is valid I just don't want to run any emitter for that spec.

I think the Autorest emitter rule already cover this (with suppression)

specification/keyvault/data-plane/Administration/BackupRestore/tspconfig.yaml
mikeharder, line 1: before merging the rule update, please split the spec folder changes out to a separate PR.  i prefer to not update specs and tooling in same PR.

mikeharder (approved):

mikeharder (approved):

PR #43232, opened 2026-05-15
Remove invalid emit autorest

timotheeguerin:
extracted spec changes from #43220

>When running tsv all this create output in /tsp-output/@azure-tools/typespec-autorest
As far as I understand those tspconfig are only for client splitting and shouldn't have an autorest output (no options configured which is why it was emitting in that location)
Had to add a suppression and handling


mikeharder (approved):

heaths (approved):

heaths:
This seems right, but hard to say what the fallout would be. We'll see. We can always restore of fix any issues that result, if any. As long as swaggers are still generated for secrets, keys, and certificates - which we know some 3P use - we should be okay.

PR #43239, opened 2026-05-15
Add csharp customizations for ResourceHealth migration

zedy-wj:
## Summary

Add csharp-scoped decorators to `client.tsp` and emitter config to `tspconfig.yaml` for the Azure.ResourceManager.ResourceHealth TypeSpec migration.

### Changes
- `@@clientName` decorators for model/enum/property renames to match GA 1.0.0 surface
- `@@clientLocation` decorators to split subscription/tenant operations  
- `@azure-typespec/http-client-csharp-mgmt` emitter config with namespace override


specification/resourcehealth/resource-manager/Microsoft.ResourceHealth/ResourceHealth/client.tsp
```diff
@@ -21,3 +32,202 @@ using Microsoft.ResourceHealth;
...
+  "csharp"
+);
+@@alternateType(
+  EventImpactedResourceProperties.targetResourceType,
+  Azure.Core.armResourceType,
+  "csharp"
+);
+
+// Override subscription-level Event resource class names for csharp to match GA SDK names
+@@clientLocation(
```
live1206: the comment doesn't make much sense, please double confirm why these @@clientLocation are needed here? If they are for renaming, why not use @@clientName?

welovej: different scoped methods return the same resource model. This would case to the same mockable method to invoke the same source method.
So we add resource name in the Legacy.ExtensionOperations to differentate these two method.

mzhongl524:
The breaking change in Swagger is because `AvailabilityStatus` isn't a resource model, but when converting from Swagger to TSP it became a resource model, so now I manually changed it back to a non-resource model.

specification/resourcehealth/resource-manager/Microsoft.ResourceHealth/ResourceHealth/AvailabilityStatus.tsp
```diff
@@ -16,9 +16,12 @@ namespace Microsoft.ResourceHealth;
 #suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
 #suppress "@azure-tools/typespec-azure-resource-manager/arm-custom-resource-no-key" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
 #suppress "@azure-tools/typespec-azure-resource-manager/arm-custom-resource-usage-discourage" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-@singleton("current")
-@extensionResource
-model AvailabilityStatus extends ProxyResource {
+model AvailabilityStatus extends Foundations.ProxyResource {
```
ArthurMa1978: What problem would arise if this change is not applied to the base type of `AvailabilityStatus` resource?
If it’s only a resource indication issue, it can be addressed using `hierarchyBuilding` decorator in client.tsp.

mzhongl524: fixed

ArthurMa1978 (approved):

PR #43268, opened 2026-05-15
Add LaunchBulkInstancesOperation resource to Microsoft.Compute from Microsoft.ComputeBulkActions

ykh015:
## Summary

Adds the `LaunchBulkInstancesOperation` ARM resource and its 8 operations to the existing **`Microsoft.Compute/Bulkactions`** TypeSpec project under the existing **api-version `2026-04-06-preview`**.

> **Breaking-change expected, no customer impact.**
> This PR adds two new optional properties (`maxPricePerVM`, `evictionPolicy` on `PriorityProfile`) and a new resource type (`LaunchBulkInstancesOperation`) to an already-published preview API version. The `Swagger BreakingChange` CI check therefore reports `SameVersion` AddedPath/AddedOptionalProperty entries (rules 1038, 1045) which trip `VersioningReviewRequired`. **No SDK has been generated for `2026-04-06-preview` yet** (the only published `@azure/arm-computebulkactions` is under a separate RP), **no ARM manifest changes** are bundled with this PR, and **no production traffic is hitting this preview** — so there is no customer impact. Requesting ARM/versioning reviewer sign-off rather than introducing a churned `2026-06-06-preview` version.

The resource and shape were ported from the existing TypeSpec project at `specification/computebulkactions/ComputeBulkActions.Management/` (RP `Microsoft.ComputeBulkActions`, version `2026-02-01-preview`). Wherever the source's models had an existing equivalent in this project's TypeSpec graph (notably the whole `virtualMachineProfile.tsp` / `virtualMachineExtension.tsp` graph as `BulkactionVMProperties` / `BulkactionVMExtension`, plus `VmSizeProfile`, `ZoneAllocationPolicy`, `ZonePreference`, `AllocationStrategy`, `DistributionStrategy`, `RetryPolicy`), the new code **reuses** them rather than duplicating definitions. **Source-RP parity** is the consistent design principle throughout — see "Source RP parity decisions" below.

## What's added

New file: **`launchBulkInstancesOperation.tsp`**

- Resource: `LocationBasedLaunchBulkInstancesOperation is Azure.ResourceManager.ProxyResource<LaunchBulkInstancesOperationProperties>` with required `location` segment + GUID-pattern `name` segment.
- Envelope `LaunchBulkInstancesOperationEnvelope` carrying `zones`, `tags`, `identity`, `plan` (matching `Microsoft.ComputeBulkActions/2026-02-01-preview` byte-for-byte, including its inline `arm-resource-invalid-envelope-property` / `arm-no-record` suppressions).
- Properties: `createdTime` (read-only), `provisioningState` (read-only), `capacity`, `capacityType`, `priorityProfile`, `vmSizesProfile`, `vmAttributes`, `computeProfile` (wraps existing `BulkactionVMProperties`), `zoneAllocationPolicy`, `retryPolicy`.
- 8 operations:
  - `LaunchBulkInstancesOperation_Get` (sync GET single resource)
  - `LaunchBulkInstancesOperation_GetOperationStatus` (sync GET `operations/{id}` -> common-types `OperationStatusResult`)
  - `LaunchBulkInstancesOperation_CreateOrUpdate` (LRO PUT, `final-state-via: azure-async-operation`)
  - `LaunchBulkInstancesOperation_Delete` (LRO DELETE with `?deleteInstances`, `final-state-via: location`)
  - `LaunchBulkInstancesOperation_Cancel` (LRO POST action, `final-state-via: location`)
  - `LaunchBulkInstancesOperation_ListByResourceGroup` (paged GET)
  - `LaunchBulkInstancesOperation_ListBySubscription` (paged GET)
  - `LaunchBulkInstancesOperation_ListVirtualMachines` (paged GET child VM list with `$filter`/`$skiptoken`)
- New models: `LaunchBulkInstancesOperationProperties`, `LaunchBulkInstancesOperationListResult`, `ComputeProfile`, `VMAttributes`, `VMAttributeMinMaxInteger`, `VMAttributeMinMaxDouble`, `VirtualMachine`, `VirtualMachineListResult`, `ApiError`, `ApiErrorBase`, `InnerError`.
- New unions: `ProvisioningState`, `CapacityType`, `EvictionPolicy`, `ZoneDistributionStrategy`, `ArchitectureType`, `VMAttributeSupport`, `LocalStorageDiskType`, `AcceleratorManufacturer`, `AcceleratorType`, `VMCategory`, `CpuManufacturer`, `HyperVGeneration`, `VMOperationStatus`.
- New shared file `vmAttributes.tsp` consolidating the `VMAttributes` graph so any future resource in this project (or beyond) can import without duplicating definitions.

## Edits to existing files

- **`main.tsp`** — added `import "./launchBulkInstancesOperation.tsp";` and (separately) added `import "./vmAttributes.tsp";`.
- **`models.tsp`** — added two new optional `PriorityProfile` properties: `maxPricePerVM?: float32` and `evictionPolicy?: EvictionPolicy`. Each carries a doc-string callout "*Available from 2026-04-06-preview*". `maxPricePerVM` is suppressed under `casing-style` (matches established Compute naming).
- **`routes.tsp`** — renamed the existing `LocationParameter` alias key from `locationparameter` → `location` (Section 1.6 finding: the existing `virtualMachinesBulk*` POST paths now use the same parameter name as the new `launchBulkInstancesOperations` paths). All 18 affected `VirtualMachineBulkOperations_Bulk*` example payloads updated to match.
- **`readme.md`** — added scoped suppressions for `PostResponseCodes`, `ParameterNotUsingCommonTypes` (with a `where:` JSONPath array covering the 17 paths that bind `location` as a path key), `MissingSegmentsInNestedResourceListOperation`, and `EnumInsteadOfBoolean`. Each suppression carries an explicit technical justification.

## Examples

11 new `LaunchBulkInstancesOperation_*` examples — hand-curated to ARM example standards (real arm-ids, valid GUIDs, sensible `location`, descriptive filler):

| Example | File |
|---|---|
| Get | `LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json` |
| CreateOrUpdate | `LaunchBulkInstancesOperation_CreateOrUpdate_MaximumSet_Gen.json` |
| Delete | `LaunchBulkInstancesOperation_Delete_MaximumSet_Gen.json` |
| Cancel | `LaunchBulkInstancesOperation_Cancel_MaximumSet_Gen.json` |
| GetOperationStatus | `LaunchBulkInstancesOperation_GetOperationStatus_MaximumSet_Gen.json` |
| ListByResourceGroup | `LaunchBulkInstancesOperation_ListByResourceGroup_{Maximum,Minimum}Set_Gen.json` |
| ListBySubscription | `LaunchBulkInstancesOperation_ListBySubscription_{Maximum,Minimum}Set_Gen.json` |
| ListVirtualMachines | `LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json` |

Each example response envelope uses the canonical ARM resource type `Microsoft.Compute/locations/launchBulkInstancesOperations`, the canonical operation GUID, and the real ARM id under `.../locations/eastus2euap/launchBulkInstancesOperations/{name}`. LRO header values use realistic `https://management.azure.com/...?api-version=2026-04-06-preview` polling URLs. Pagination `nextLink`s use the corresponding endpoint URL with `&$skiptoken=opaque-page-token`.

Existing examples were also cleaned up in this PR:
- `Operations_List_MaximumSet_Gen.json` replaced its `isDataAction: true` / `actionType: "Internal"` payload with a representative control-plane operation (`Microsoft.Compute/locations/launchBulkInstancesOperations/read`) per `EX-PAYLOAD`.
- `VirtualMachineBulkOperations_BulkReimage_MaximumSet_Gen.json` `customData` replaced with `{base64-encoded-cloud-init-script}` placeholder (secret-scanner hit on the previous base64 string).
- `VirtualMachineBulkOperations_Bulk{Create,VdiFlexCreate}_MinimumSet_Gen.json` titles fixed (duplicate `- generated by [MaximumSet] rule - generated by [MinimumSet] rule` suffix removed).

## Source RP parity decisions

The intent of this PR is faithful migration. Two ARM API Reviewer findings were resolved by keeping parity with the already-shipped `Microsoft.ComputeBulkActions/2026-02-01-preview` rather than diverging:

1. **LRO headers on Cancel/Delete 202** — RPC-Async-V1-11 asks for both `Location` and `Azure-AsyncOperation`. The source RP emits only `Location` + `Retry-After` under a `PostResponseCodes` suppression. After investigation we matched the source-RP layout byte-for-byte (so the two RPs behave identically for clients) and inherit the same `PostResponseCodes` suppression in `readme.md`.
2. **`ProxyResource` envelope with `zones`/`tags`/`identity`/`plan`** — the same modeling exists on the source RP with the same inline lint suppressions. Switching to `TrackedResource` here would change the routing model (moving `location` from a path key to a body property), which is not the contract Kronox implements server-side. Better addressed as a coordinated update across both RPs in a future API version.

## Notes / caveats for reviewers

1. **Cross-RP duplication is intentional.** The same logical resource exists in `Microsoft.ComputeBulkActions/2026-02-01-preview` and in `Microsoft.Compute/Bulkactions` here. Hosting it under `Microsoft.Compute` is a deliberate decision coordinated with the Compute and ARM design leads; the source RP will be deprecated once `Microsoft.Compute/Bulkactions` reaches GA. SDK reviewers should be aware that the two RPs ship the same resource type name (`launchBulkInstancesOperations`).
2. **`PriorityProfile` cross-pollination.** The two new optional fields (`maxPricePerVM`, `evictionPolicy`) appear on `PriorityProfile`, which is also used by the existing `BulkCreate` / `BulkVdiFlexCreate` flows. They are optional and unused in those flows, so this is non-breaking at the wire level but acknowledged as a `SameVersion AddedOptionalProperty` in CI.
3. **No `readme.md` `BreakingChange-*-Sdk` suppressions added.** No SDK has been generated for `2026-04-06-preview` yet; SDK breaking-change suppressions become relevant only if a generated SDK exists, which is not the case here.
4. **No ARM manifest changes.** This PR is spec-only.


specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
```diff
@@ -0,0 +1,514 @@
...
+          "tenantId": "3ec2ab23-9f13-4328-85c8-21928acbc7b8"
+        },
+        "plan": {
+          "name": "owvrgjbxrkj",
+          "publisher": "qhybdqbljmztcjujxal",
+          "product": "rlhap",
+          "promotionCode": "agypojbtdxvgqgisautnhcoysgy",
+          "version": "ghmnlomqg"
+        },
+        "id": "/subscriptions/50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B/resourceGroups/rgcomputebulkactions/providers/Microsoft.Compute/bulkActions/3ec2ab23-9f13-4328-85c8-21928acbc7b8",
```
ravimeda: **[NEW] 🔴 Blocking** **[[EX-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#222-resource-identity-consistency-ex-resource-id)]** — Response `id` carries the wrong ARM resource type

**Issue.** Every `LaunchBulkInstancesOperation_*` response example sets the envelope to:

`"id": "/subscriptions/.../resourceGroups/rgcomputebulkactions/providers/Microsoft.Compute/bulkActions/3ec2ab23-..."`, `"name": "kv"`, `"type": "v"`.

But the request path for this resource is `/subscriptions/.../resourceGroups/.../providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations/{name}`. The response uses a non-existent resource type (`bulkActions`), is missing the `locations/{location}/` segments, and uses 1–2-character filler for `name`/`type`. This violates `EX-RESOURCE-ID` and `EX-PARAM-CONSISTENCY`.

**Affected files (same defect):**
- `LaunchBulkInstancesOperation_CreateOrUpdate_MaximumSet_Gen.json` lines 979 and 1480
- `LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json` line 500 (this comment)
- `LaunchBulkInstancesOperation_ListByResourceGroup_MaximumSet_Gen.json` line 501
- `LaunchBulkInstancesOperation_ListByResourceGroup_MinimumSet_Gen.json` line 15
- `LaunchBulkInstancesOperation_ListBySubscription_MaximumSet_Gen.json` line 500
- `LaunchBulkInstancesOperation_ListBySubscription_MinimumSet_Gen.json` line 14
- The same six files duplicated under `Bulkactions/examples/2026-06-06-preview/`.

**Classification.** NEW — this is a new resource introduced in this PR, so all the examples are new.

**Fix.** Set every response envelope to:
```json
"id": "/subscriptions/50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B/resourceGroups/rgcomputebulkactions/providers/Microsoft.Compute/locations/eastus2euap/launchBulkInstancesOperations/3ec2ab23-9f13-4328-85c8-21928acbc7b8",
"name": "3ec2ab23-9f13-4328-85c8-21928acbc7b8",
"type": "Microsoft.Compute/locations/launchBulkInstancesOperations"
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Cancel_MaximumSet_Gen.json
```diff
@@ -0,0 +1,18 @@
...
+    "api-version": "2026-06-06-preview",
+    "subscriptionId": "50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B",
+    "resourceGroupName": "rgcomputebulkactions",
+    "name": "3ec2ab23-9f13-4328-85c8-21928acbc7b8",
+    "location": "eastus2euap"
+  },
+  "responses": {
+    "202": {
+      "headers": {
+        "location": "https://contoso.com/operationstatus"
```
ravimeda: **[NEW] 🔴 Blocking** **[[EX-LRO-HEADERS](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#225-lro-example-headers-ex-lro-headers)]** — Placeholder polling URL `https://contoso.com/operationstatus` and lowercase `location` header

**Issue.** The LRO 202 response uses `"location": "https://contoso.com/operationstatus"`. Per `EX-LRO-HEADERS`:
1. Header names **MUST** be PascalCase — use `"Location"` not `"location"`.
2. Polling URLs **MUST NOT** use placeholder domains like `contoso.com`. Use a realistic `https://management.azure.com/...` URL.
3. Async POST per RPC-Async-V1-11 **SHOULD** also include `Azure-AsyncOperation` header (see separate finding on the swagger).

**Same defect at:**
- `LaunchBulkInstancesOperation_Delete_MaximumSet_Gen.json` line 14
- `LaunchBulkInstancesOperation_CreateOrUpdate_MaximumSet_Gen.json` line 994 (`Azure-AsyncOperation` set to `https://contoso.com/operationstatus`)
- Mirror copies under `Bulkactions/examples/2026-06-06-preview/`.

**Classification.** NEW — all new resource example files.

**Fix.**
```json
"202": {
  "headers": {
    "Location": "https://management.azure.com/subscriptions/50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B/providers/Microsoft.Compute/locations/eastus2euap/operations/2a3fce8e-874c-45f4-9d27-1a804f3b7a0f?api-version=2026-06-06-preview",
    "Azure-AsyncOperation": "https://management.azure.com/subscriptions/50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B/providers/Microsoft.Compute/locations/eastus2euap/operations/2a3fce8e-874c-45f4-9d27-1a804f3b7a0f?api-version=2026-06-06-preview",
    "Retry-After": 30
  }
}
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
```diff
@@ -0,0 +1,5579 @@
...
+            "minLength": 1,
+            "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+          }
+        ],
+        "responses": {
+          "202": {
+            "description": "Resource operation accepted.",
+            "headers": {
+              "Location": {
+                "type": "string",
```
ravimeda: **[NEW] 🔴 Blocking** **[[RPC-Async-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#64-async-post-action-model-rpc-async-v1-11) / [RPC-Async-V1-06](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#66-202-response-and-polling-headers-rpc-async-v1-07-rpc-async-v1-06-rpc-async-v1-14)]** — Async POST `Cancel` 202 response missing `Azure-AsyncOperation` header

**Issue.** Per RPC-Async-V1-11, async POST actions **MUST** return 202 with **both** `Location` and `Azure-AsyncOperation` headers. RPC-Async-V1-06 (effective January 2025) makes `Azure-AsyncOperation` required for new RP namespaces. The Cancel 202 here declares only `Location`. The same gap exists on the async DELETE 202 around line 906.

**Classification.** NEW — both operations are new in this PR.

**Fix.** In the TypeSpec source, the LRO templates (`ArmResourceActionNoResponseContentAsync` for Cancel, `ArmResourceDeleteWithoutOkAsync` for Delete) should emit both headers. If the underlying templates do not, switch to templates that do, or add the header explicitly with a doc-only TSP override. After regeneration the swagger `responses["202"].headers` should contain both `Location` **and** `Azure-AsyncOperation`.
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/readme.md
```diff
@@ -60,3 +60,39 @@ These settings apply only when `--tag=package-2026-04-06-preview` is specified o
...
+    from: Bulkactions.json
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations/{name}/cancel"].post
+  - code: ParameterNotUsingCommonTypes
+    reason: >
+      LaunchBulkInstancesOperation uses location as a resource path segment key
+      (location is azureLocation in the resource model), so the location parameter
+      cannot be a $ref to common-types LocationParameter without losing the resource key binding.
+    from: Bulkactions.json
+  - code: MissingSegmentsInNestedResourceListOperation
+    reason: >
```
ravimeda, line 97: **[NEW] 🔴 Blocking** **[[Section 10A — Suppression Scoping](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#10a-readmemd-suppression-scoping--tag-coverage)]** — `ParameterNotUsingCommonTypes` suppression is missing `where:` scoping

**Issue.** This new suppression has no `where:` clause, so it applies file-wide and silences the rule on every parameter in `Bulkactions.json` — including the existing `VirtualMachineBulkOperations_*` paths (which legitimately use the oddly-named `locationparameter`). The other three new suppressions in this same block correctly scope themselves with `where:` JSONPath. Per Section 10A, suppressions **MUST** be narrowly scoped.

**Classification.** NEW — this suppression block is being added in this PR.

**Fix.** Add a `where:` clause limiting the suppression to the new resource's `location` parameter occurrences only:
```yaml
- code: ParameterNotUsingCommonTypes
  reason: >
    LaunchBulkInstancesOperation uses location as a resource path segment key
    (location is azureLocation in the resource model), so the location parameter
    cannot be a $
ef to common-types LocationParameter without losing the resource
    key binding.
  from: Bulkactions.json
  where:
    - $.paths["/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations"].get.parameters[?(@.name=='location')]
    - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations"].get.parameters[?(@.name=='location')]
    - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations/{name}"].*.parameters[?(@.name=='location')]
    - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations/{name}/cancel"].post.parameters[?(@.name=='location')]
    - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations/{name}/virtualMachines"].get.parameters[?(@.name=='location')]
    - $.paths["/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/operations/{id}"].get.parameters[?(@.name=='location')]
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
```diff
@@ -0,0 +1,435 @@
...
+  @doc("The Virtual Machine items on this page.")
+  value: Array<VirtualMachine>;
+
+  @nextLink
+  @doc("The link to the next page of items.")
+  nextLink?: url;
+}
+
+// ============================================================
+// Resource: LaunchBulkInstancesOperation
```
ravimeda, line 244: **[NEW] 🔴 Blocking** **[[Section 1.2 — Proxy Resource Paths](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#12-proxy-resource-paths) / [RPC-Put-V1-12](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#21-put-response-must-be-an-arm-resource-rpc-put-v1-12)]** — `ProxyResource` with `tags`, `zones`, `identity`, `plan` envelope is non-standard ARM modeling

**Issue.** `LocationBasedLaunchBulkInstancesOperation is Azure.ResourceManager.ProxyResource<...>` (line 271 in this file) yet the envelope (this `LaunchBulkInstancesOperationEnvelope` block) adds `tags`, `zones`, `identity` (via `ManagedServiceIdentityProperty`), and `plan` (via `ResourcePlanProperty`). These are **TrackedResource** envelope properties — the TypeSpec ARM library enforces this with `arm-resource-invalid-envelope-property` (which the PR suppresses inline twice). Per ARM Section 1.2 / 2.6:
- Proxy resources **MUST NOT** have `tags`.
- `tags`, `zones`, `identity`, `plan` belong on tracked resources.

The resource also uses `location` as a path key (line 280). A regional resource with identity/zones/plan/tags is effectively a regional tracked resource — declaring it as a proxy strips it of ARM tracked-resource behaviors (resource move, ARG location-based queries, region failover semantics).

Inline suppression with the reason _"Zones is a valid envelope property for this resource"_ is **not** a technical justification (typespec-review §4.1) — the lint rule is correct, the modeling is non-standard.

**Classification.** NEW — the resource model is new in this PR.

**Fix.** Choose one:
1. **(Preferred)** Change to `is Azure.ResourceManager.TrackedResource<LaunchBulkInstancesOperationProperties>`. Keep `location` as the path key but also let it appear as the body property (standard ARM regional pattern). The `tags`/`identity`/`plan`/`zones` properties come free from the tracked-resource base; delete the inline `arm-resource-invalid-envelope-property` suppressions.
2. Keep `ProxyResource` and drop `tags`, `zones`, `identity`, `plan` from the envelope.

A suppression must explain **why the rule does not apply**, not assert that it doesn't.
 #Resolved

ravimeda (commented):
## ARM API Review — Posted by arm-api-reviewer-agent

This review covers `specification/compute/resource-manager/Microsoft.Compute/Bulkactions` for the new `2026-06-06-preview` API version.

**Summary**
- 6 new blocking issues (please fix)
- 7 new warnings (please fix)
- 3 new suggestions (please confirm)
- 5 pre-existing items carried forward from `2026-04-06-preview` (consider fixing in this PR)

Each inline comment includes a hidden telemetry marker for de-duplication.

Previous version compared: `preview/2026-04-06-preview/` (on `main`).

— Generated by `arm-api-reviewer-agent`.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json
```diff
@@ -0,0 +1,44 @@
+{
+  "title": "LaunchBulkInstancesOperation_ListVirtualMachines - generated by [MaximumSet] rule",
+  "operationId": "LaunchBulkInstancesOperation_ListVirtualMachines",
+  "parameters": {
+    "api-version": "2026-06-06-preview",
+    "subscriptionId": "50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B",
+    "resourceGroupName": "rgcomputebulkactions",
+    "name": "50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B",
```
ravimeda: **[NEW] 🔴 Blocking** **[[EX-DESCRIPTIVE-VALUES](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2212-descriptive-example-values-ex-descriptive-values) / [EX-PARAM-CONSISTENCY](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#223-request-response-parameter-consistency-ex-param-consistency)]** — `ListVirtualMachines` example mixes parameter values and uses filler for VM type/name

**Issue.**
1. `"subscriptionId": "50352BBD-..."` and `"name": "50352BBD-..."` are the **same GUID**. The `name` parameter is the LaunchBulkInstancesOperation's id — it should be a separate GUID like `3ec2ab23-9f13-4328-85c8-21928acbc7b8` (matching the rest of the examples), not the subscription id.
2. In the response (line 17-19), `"type": "ocaksutbzfxnzsrk"` and `"name": "wyonfuximvgywtidpl"` are random filler. The ARM type for VMs is `Microsoft.Compute/virtualMachines` and the name should match the last segment of the `id` (`vm1`).

**Classification.** NEW — new example file.

**Fix.**
```json
"parameters": {
  "api-version": "2026-06-06-preview",
  "subscriptionId": "50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B",
  "resourceGroupName": "rgcomputebulkactions",
  "name": "3ec2ab23-9f13-4328-85c8-21928acbc7b8",
  "location": "eastus2euap",
  "$ilter": "operationStatus eq 'Failed'",
  "$skiptoken": "opaque-page-token"
}
```
And in the response item:
```json
{
  "name": "vm1",
  "id": "/subscriptions/50352BBD-.../resourceGroups/rgcomputebulkactions/providers/Microsoft.Compute/virtualMachines/vm1",
  "type": "Microsoft.Compute/virtualMachines",
  "operationStatus": "Creating",
  "error": { "code": "...", "message": "..." }
}
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
```diff
@@ -0,0 +1,435 @@
...
+  ...LaunchBulkInstancesOperationId;
+};
+
+// ============================================================
+// Enums
+// ============================================================
+
+@added(Microsoft.Compute.Versions.`2026-06-06-preview`)
+@Azure.Core.lroStatus
+@doc("The status of the LaunchBulkInstancesOperation.")
```
ravimeda, line 74: **[NEW] 🟡 Warning** **[[Section 2.4 — Enums & Provisioning State](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#24-enums--provisioning-state-review-enforcement) / [provisioning-state reference](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md)]** — Custom `ProvisioningState` duplicates `Azure.ResourceManager.ResourceProvisioningState`

**Issue.** Per Section 2.4, `provisioningState` **SHOULD** extend `Azure.ResourceManager.ResourceProvisioningState`. Hand-rolling the union duplicates the three required terminal states (`Succeeded`, `Failed`, `Canceled`) and risks drift across versions or services.

**Classification.** NEW — the union is added in this PR.

**Fix.**
```tsp
@added(Microsoft.Compute.Versions.2026-06-06-preview)
@Azure.Core.lroStatus
@doc("The status of the LaunchBulkInstancesOperation.")
union ProvisioningState {
  ...Azure.ResourceManager.ResourceProvisioningState,
  @doc("Initial creation in progress.") Creating: "Creating",
  @doc("Deletion in progress.") Deleting: "Deleting",
}
```

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json
```diff
@@ -0,0 +1,44 @@
...
+                }
+              ],
+              "innererror": {
+                "exceptionType": "klyhjncvdigrputprzqhthxr",
+                "errorDetail": "xvsexmezlv"
+              }
+            }
+          }
+        ],
+        "nextLink": "https://microsoft.com/a"
```
ravimeda: **[NEW] 🟡 Warning** **[[EX-PAGINATION](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#224-pagination-examples-ex-pagination)]** — `nextLink: "https://microsoft.com/a"` is not a realistic ARM URL

**Issue.** Per `EX-PAGINATION`, `nextLink` **MUST** be a well-formed URL that includes `api-version=` and points to the same operation endpoint. `https://microsoft.com/a` is neither. The same filler appears in `LaunchBulkInstancesOperation_ListByResourceGroup_MaximumSet_Gen.json` and `LaunchBulkInstancesOperation_ListBySubscription_MaximumSet_Gen.json`, and the `Operations_List_MaximumSet_Gen.json` (existing) also uses this URL.

**Classification.** NEW — this example file is new.

**Fix.**
```json
"nextLink": "https://management.azure.com/subscriptions/50352BBD-59F1-4EE2-BA9C-A6E51B0B1B2B/resourceGroups/rgcomputebulkactions/providers/Microsoft.Compute/locations/eastus2euap/launchBulkInstancesOperations/3ec2ab23-9f13-4328-85c8-21928acbc7b8/virtualMachines?api-version=2026-06-06-preview&$skiptoken=<opaque-page-token>"
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/examples/2026-04-06-preview/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
```diff
@@ -0,0 +1,514 @@
...
+        "zones": [
+          "cyriutfcgydtaezeso"
+        ],
+        "identity": {
+          "type": "None",
+          "userAssignedIdentities": {},
+          "principalId": "2ebdc740-a760-4406-80e9-6be95ebc9743",
+          "tenantId": "3ec2ab23-9f13-4328-85c8-21928acbc7b8"
+        },
+        "plan": {
```
ravimeda, line 486: **[NEW] 🟡 Warning** **[[EX-PAYLOAD](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#227-example-payload-correctness-ex-payload)]** — `identity.type: "None"` is inconsistent with populated `userAssignedIdentities`/`principalId`/`tenantId`

**Issue.** When `identity.type` is `"None"`, there are no managed identities, so `userAssignedIdentities`, `principalId`, and `tenantId` must not be set. The example is internally contradictory and will confuse SDK reviewers and template consumers. The same pattern appears in the `CreateOrUpdate` 200 and 201 response bodies.

**Classification.** NEW — new example file.

**Fix.** Either drop the entire `identity` block (no managed identity) or use a consistent value:
```json
"identity": {
  "type": "SystemAssigned, UserAssigned",
  "userAssignedIdentities": {
    "/subscriptions/.../providers/Microsoft.ManagedIdentity/userAssignedIdentities/myUai": {
      "principalId": "11111111-2222-3333-4444-555555555555",
      "clientId": "66666666-7777-8888-9999-aaaaaaaaaaaa"
    }
  },
  "principalId": "2ebdc740-a760-4406-80e9-6be95ebc9743",
  "tenantId": "3ec2ab23-9f13-4328-85c8-21928acbc7b8"
}
```

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
```diff
@@ -0,0 +1,5579 @@
...
+          },
+          "x-ms-identifiers": [
+            "id"
+          ]
+        },
+        "nextLink": {
+          "type": "string",
+          "description": "The URL to get the next set of results."
+        }
+      },
```
ravimeda: **[NEW] 🟡 Warning** **[[Section 2.2 — Tracked Resource Requirements (RPC-Get-V1-13)](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#22-tracked-resource-requirements)]** — `LaunchBulkInstancesOperationListResult.nextLink` is plain `string` instead of `format: uri`

**Issue.** Per RPC-Get-V1-13, collection `nextLink` properties **MUST** declare `"format": "uri"` so SDKs and ARM tooling treat them as URLs. The sibling `VirtualMachineListResult.nextLink` (further down in the file) already uses `format: uri` — this one does not.

**Classification.** NEW — both list-result models are new in this PR.

**Fix (in TypeSpec source `launchBulkInstancesOperation.tsp` line 296):**
```tsp
@nextLink
@doc("The URL to get the next set of results.")
nextLink?: url; // was: string
```
After regeneration the swagger definition becomes `{ "type": "string", "format": "uri", "description": "..." }`.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
```diff
@@ -0,0 +1,435 @@
...
+
+  @doc("Specifies the Microsoft.Compute API version to use when creating underlying Virtual Machines. The default value will be the latest supported computeApiVersion by LaunchBulkInstancesOperation.")
+  computeApiVersion?: string;
+}
+
+@added(Microsoft.Compute.Versions.`2026-06-06-preview`)
+@doc("API error base.")
+model ApiErrorBase {
+  @doc("The error code.")
+  code?: string;
```
ravimeda, line 179: **[NEW] 🟡 Warning** **[[Section 8.4 — Use Specific Types Instead of Generic Strings](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#84-use-specific-types-instead-of-generic-strings-actively-review)]** — `ComputeProfile.computeApiVersion` is a freeform `string`

**Issue.** The doc string describes this as a Microsoft.Compute API version (date-formatted string). As a freeform `string` clients may pass anything — the example file accepts `"bccikdfzgrygwpefvmvptutceg"`, which is what gets shipped as the SDK reference value. Constrain the type.

**Classification.** NEW — property is added in this PR.

**Fix.** Choose one:
1. Add a pattern constraint:
```tsp
@pattern("^[0-9]{4}-[0-9]{2}-[0-9]{2}(-preview)?$")
@doc("Specifies the Microsoft.Compute API version...")
computeApiVersion?: string;
```
2. Or model an extensible union:
```tsp
union ComputeApiVersion {
  "2024-03-01": "2024-03-01",
  "2024-07-01": "2024-07-01",
  string,
}
```

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
```diff
@@ -0,0 +1,435 @@
...
+model ApiError is ApiErrorBase {
+  @identifiers(#["message", "target"])
+  @doc("The API error details")
+  details?: Array<ApiErrorBase>;
+
+  @doc("The API inner error")
+  innererror?: InnerError;
+}
+
+@added(Microsoft.Compute.Versions.`2026-06-06-preview`)
```
ravimeda: **[NEW] 🟡 Warning** **[[Section 2.5 — Top-Level ARM Properties Must Not Be Reused Inside the Properties Bag](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#25-top-level-arm-properties-must-not-be-reused-inside-the-properties-bag)]** — `VirtualMachine.type` is a freeform `string` shadowing the ARM envelope `type`

**Issue.** `VirtualMachine` is a non-resource model returned by the `ListVirtualMachines` operation. Adding `type?: string` shadows the ARM envelope `type`. The example shows `"type": "ocaksutbzfxnzsrk"` (filler), which is what SDK callers will see in samples. Either constrain the property to the actual ARM type literal or remove it (`id` already implies the type).

**Classification.** NEW — the model is added in this PR.

**Fix.** Remove `type` from the `VirtualMachine` model, or constrain it:
```tsp
@visibility(Lifecycle.Read)
@doc("Type of the virtual machine — always 'Microsoft.Compute/virtualMachines'.")
type?: "Microsoft.Compute/virtualMachines";
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
```diff
@@ -0,0 +1,514 @@
...
+        "name": "kv",
```
ravimeda: **[NEW] 🟡 Warning** **[[EX-DESCRIPTIVE-VALUES](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2212-descriptive-example-values-ex-descriptive-values)]** — Resource envelope `name`/`type` are 1–2-character filler

**Issue.** `"name": "kv"` and `"type": "v"` are auto-generated filler. The same pattern appears in `CreateOrUpdate` lines 980–982 and 1481–1483. The body of the `MaximumSet` examples is full of similar fillers (`"obwiwwsgkdg"`, `"iwqrkiccafacxfctrxb"`, `"wyonfuximvgywtidpl"`, `"hvrimblcqujozpeurohjcn"`). Per `EX-DESCRIPTIVE-VALUES`, auto-generated examples **SHOULD** be hand-edited.

Note: this is intertwined with finding #1 (EX-RESOURCE-ID) — fixing the resource `id` should also produce a real `name` and `type`.

**Classification.** NEW — new example file.

**Fix.** Hand-edit values to recognizable ones (e.g. `"adminUsername": "azureuser"`, `"computerName": "myVM01"`, `"name": "3ec2ab23-9f13-4328-85c8-21928acbc7b8"`, `"type": "Microsoft.Compute/locations/launchBulkInstancesOperations"`).
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/main.tsp
```diff
@@ -19,4 +21,8 @@ enum Versions {
   /** 2026-04-06-preview version */
   @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
   `2026-04-06-preview`,
+
+  /** 2026-06-06-preview version */
+  @armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
```
ravimeda: **[NEW] 🔵 Suggestion** **[[Section 11.1 / RPC-BestPractice-14](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#111-uniform-versioning-within-a-service)]** — Cross-RP duplication awareness

**Issue.** The PR description notes that `LaunchBulkInstancesOperation` is intentionally hosted in **both** `Microsoft.Compute` (this PR) and `Microsoft.ComputeBulkActions` (separate RP). The APIView check has produced two SDK reviews: `@azure/arm-computebulkactions` and `armbulkactions` under `compute/resource-manager`. SDK reviewers and customers will see two different packages for the same logical operation.

**Classification.** NEW.

**Fix.** No spec change required, but confirm with the SDK leads and the ARM design lead that the duplication is acceptable. Document the relationship in a service README so customers know which RP is canonical.
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/main.tsp
```diff
@@ -19,4 +21,8 @@ enum Versions {
...
+  `2026-06-06-preview`,
```
ravimeda: **[NEW] 🔵 Suggestion** **[[Section 23.2 — Preview API Lifecycle](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#232-preview-api-lifecycle)]** — Two preview versions ~2 months apart; confirm GA / retirement plan for `2026-04-06-preview`

**Issue.** `2026-04-06-preview` and `2026-06-06-preview` are only ~2 months apart and both target the same surface. Per Section 23.2, preview APIs **MUST NOT** remain in preview for more than 12 months. The `VersioningReviewRequired` label was applied then removed by automation, suggesting the cadence was reviewed — but please confirm a documented GA path or retirement plan for `2026-04-06-preview`.

**Classification.** NEW.

**Fix.** No code change required, but please reply with the planned GA date or the retirement timeline for `2026-04-06-preview`.
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
```diff
@@ -0,0 +1,5579 @@
...
+          "LaunchBulkInstancesOperation_ListBySubscription - generated by [MinimumSet] rule": {
+            "$ref": "./examples/LaunchBulkInstancesOperation_ListBySubscription_MinimumSet_Gen.json"
+          }
+        },
+        "x-ms-pageable": {
+          "nextLinkName": "nextLink"
+        }
+      }
+    },
+    "/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/operations/{id}": {
```
ravimeda: **[NEW] 🔵 Suggestion** **[[Section 6.8 — Operation Results Placement](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#68-operation-results-placement)]** — Verify `LaunchBulkInstancesOperation_GetOperationStatus` operation id is unique in `Microsoft.Compute`

**Issue.** The path `/subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/operations/{id}` is the standard ARM operation-status endpoint at subscription scope, which is good. But the operation id `LaunchBulkInstancesOperation_GetOperationStatus` namespaces it under the resource — and `Microsoft.Compute` is a large RP with many services. Confirm no other Microsoft.Compute service publishes a `GetOperationStatus` operation under the same path that would collide in cross-service SDKs.

**Classification.** NEW.

**Fix.** Consider renaming to `LocationOperations_GetStatus` if there is a risk of cross-service collision, or document that the resource-scoped namespace is intentional and unique.
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
```diff
@@ -0,0 +1,5579 @@
...
+      "post": {
+        "operationId": "VirtualMachineBulkOperations_BulkCancel",
+        "description": "BulkCancelOperations: Cancel a previously submitted (start/deallocate/hibernate) request",
+        "parameters": [
+          {
+            "$ref": "../../../../../../common-types/resource-management/v6/types.json#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "../../../../../../common-types/resource-management/v6/types.json#/parameters/SubscriptionIdParameter"
+          },
```
ravimeda: **[EXISTING]** **[[Section 1.6 — Resource Type Naming](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#16-resource-type-naming) / [Naming Conventions](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/naming-conventions.md)]** — Existing bulk-action POST paths use the parameter name `locationparameter` instead of `location`

**Issue.** Ten paths under `virtualMachinesBulk{Cancel,Create,Deallocate,Delete,GetOperationStatus,Hibernate,Reimage,Start,VdiFlexCreate}` use `locationparameter` (lowercase, joined) as the path parameter name. Per the Azure naming conventions, path parameters should be camelCase and meaningful. The new `LaunchBulkInstancesOperation` paths use the clean `location` parameter name — so the same swagger file now has both styles side-by-side, which is confusing for customers reading the spec.

**Previous version.** Same parameter name in `preview/2026-04-06-preview/Bulkactions.json` and the source TypeSpec in `routes.tsp` line 27 (`locationparameter: Azure.Core.azureLocation`). Pre-existing.

**Classification.** EXISTING.

**Fix (next API version).** In `routes.tsp`, rename the alias key `locationparameter` → `location`. The change will be picked up in the next API version's generated swagger and unify the parameter naming.
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/VirtualMachineBulkOperations_BulkReimage_MaximumSet_Gen.json
```diff
@@ -0,0 +1,95 @@
...
+          "onFailureAction": "Unknown"
+        }
+      },
+      "reimageParameters": {
+        "baseProfile": {
+          "tempDisk": true,
+          "exactVersion": "1.2.2",
+          "osProfile": {
+            "adminPassword": "<admin-password>",
+            "customData": "ZXhwb3J0IFNFTlNJVElWRT0x"
```
ravimeda: **[EXISTING]** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md)]** — `customData: "ZXhwb3J0IFNFTlNJVElWRT0x"` decodes to `"export SENSITIVE=1"`

**Issue.** The base64 string in `customData` decodes to `export SENSITIVE=1` — a literal env-variable export. Even though this isn't a real credential, secret scanners and AppSec reviewers flag this pattern, and customers reading SDK samples may copy it. Use a clearly-fake placeholder instead.

**Previous version.** Same payload in `preview/2026-04-06-preview/examples/VirtualMachineBulkOperations_BulkReimage_MaximumSet_Gen.json` line 28. Pre-existing.

**Classification.** EXISTING.

**Fix.**
```json
"customData": "{base64-encoded-cloud-init-script}"
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/VirtualMachineBulkOperations_BulkCreate_MinimumSet_Gen.json
```diff
@@ -0,0 +1,24 @@
+{
+  "title": "VirtualMachineBulkOperations_BulkCreate - generated by [MaximumSet] rule - generated by [MinimumSet] rule",
```
ravimeda: **[EXISTING]** **[[EX-TITLE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#221-example-title-accuracy-ex-title)]** — Malformed example title (copy-paste artifact from `oav generate-examples`)

**Issue.** The title is `"VirtualMachineBulkOperations_BulkCreate - generated by [MaximumSet] rule - generated by [MinimumSet] rule"` — both rule labels are present. The same defect exists in `VirtualMachineBulkOperations_BulkVdiFlexCreate_MinimumSet_Gen.json` line 2.

**Previous version.** Same broken title in `preview/2026-04-06-preview/examples/VirtualMachineBulkOperations_BulkCreate_MinimumSet_Gen.json` line 2. Pre-existing.

**Classification.** EXISTING.

**Fix.**
```json
"title": "VirtualMachineBulkOperations_BulkCreate - generated by [MinimumSet] rule"
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
```diff
@@ -0,0 +1,5579 @@
...
+          "description": "Indicates whether the extension should be automatically upgraded by the platform if there is a newer version of the extension available."
+        },
+        "settings": {
+          "type": "object",
+          "description": "JSON formatted public settings for the extension.",
+          "additionalProperties": {}
+        },
+        "protectedSettings": {
+          "type": "object",
+          "description": "The extension can contain either protectedSettings or protectedSettingsFromKeyVault or no protected settings at all.",
```
ravimeda: **[EXISTING]** **[[Section 7.2 — x-ms-secret Annotation](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#72-x-ms-secret-annotation) / [Secret Detection](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md)]** — `BulkactionVMExtensionProperties.protectedSettings` (and `protectedSettingsFromKeyVault`) missing `x-ms-secret: true`

**Issue.** Per Section 7.2, any property that contains a secret in any operation **MUST** be annotated with `x-ms-secret: true`. `protectedSettings` (line 1416–1420) and `protectedSettingsFromKeyVault` (line 1425–1428) literally mean "encrypted, hidden from logs". The Compute RP's main `virtualMachine.json` marks these with `x-ms-secret`, but this spec doesn't.

**Previous version.** Same omission in `preview/2026-04-06-preview/Bulkactions.json` line 771 area. Pre-existing.

**Classification.** EXISTING.

**Fix (in TypeSpec for `BulkactionVMExtension` properties):**
```tsp
@secret
@doc("The extension can contain either protectedSettings or protectedSettingsFromKeyVault or no protected settings at all.")
protectedSettings?: Record<unknown>;

@secret
@doc("The extensions protected settings that are passed by reference, and consumed from key vault")
protectedSettingsFromKeyVault?: KeyVaultSecretReference;
```

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/Operations_List_MaximumSet_Gen.json
```diff
@@ -0,0 +1,28 @@
...
+  "parameters": {
+    "api-version": "2026-06-06-preview"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "value": [
+          {
+            "name": "List Bulk Operations",
+            "isDataAction": true,
```
ravimeda: **[EXISTING]** **[[EX-PAYLOAD](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#227-example-payload-correctness-ex-payload) / [Section 1.4 — Operations API (RPC-Operations-V1)](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/armapi-review.instructions.md#14-operations-api-rpc-operations-v1)]** — `Operations_List` example has `isDataAction: true` and `actionType: "Internal"`

**Issue.** _List Bulk Operations_ is a control-plane discovery operation. `isDataAction: true` puts the operation under data-action RBAC; `actionType: "Internal"` hides it from the public surface. The two together don't match the intent of the operations listing endpoint (which is itself a control-plane read).

**Previous version.** Same payload in `preview/2026-04-06-preview/examples/Operations_List_MaximumSet_Gen.json` line 13. Pre-existing.

**Classification.** EXISTING.

**Fix.** Replace the example with a representative real operation. For instance:
```json
{
  "name": "Microsoft.Compute/locations/launchBulkInstancesOperations/read",
  "isDataAction": false,
  "display": {
    "provider": "Microsoft Compute",
    "resource": "Bulk Instances Launch Operation",
    "operation": "Get Bulk Instances Launch Operation",
    "description": "Gets the status of a bulk virtual machine launch operation."
  },
  "origin": "user,system"
}
```
 #Resolved

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
ykh015: Fixed in 2f594faa (and the preceding consolidation commit). All `LaunchBulkInstancesOperation_*` response envelopes now use `type: Microsoft.Compute/locations/launchBulkInstancesOperations`, `name: 3ec2ab23-9f13-4328-85c8-21928acbc7b8`, and the canonical ARM `id` under `.../locations/eastus2euap/launchBulkInstancesOperations/{name}`. Applied to both the `preview/2026-04-06-preview/examples/` folder and the mirrored `examples/2026-04-06-preview/` folder.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Cancel_MaximumSet_Gen.json
ykh015: Fixed. `Cancel`, `Delete`, and `CreateOrUpdate` example headers are now PascalCase `Location` and `Azure-AsyncOperation` (where applicable) with realistic polling URLs of the form `https://management.azure.com/subscriptions/.../providers/Microsoft.Compute/locations/{location}/operations/{id}?api-version=2026-04-06-preview` and `Retry-After: 30`. The header layout was then aligned to match the source `Microsoft.ComputeBulkActions/2026-02-01-preview` RP exactly (see thread #3 reply).

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Hi, As we are migrating this resource from the existing Microsoft.ComputeBulkActions to Microsoft.Compute, we would like to keep the same behavior.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/readme.md
ykh015: Fixed. The `ParameterNotUsingCommonTypes` suppression in `readme.md` now has a `where:` JSONPath array scoping it to the specific `location` path parameter on the 17 paths that use it (8 `LaunchBulkInstancesOperation` paths + 9 `virtualMachinesBulk*` paths after the `locationparameter` -> `location` rename from thread #17). The rule still fires on any other parameters in the file.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
ykh015: **Kept the existing modeling to maintain parity with the source RP.** The shipped `Microsoft.ComputeBulkActions/2026-02-01-preview` swagger and its source TSP at `ComputeBulkActions.Management/models.tsp` define this resource exactly the same way: `is ProxyResource<LaunchBulkInstancesOperationProperties>` with envelope-level `zones` / `tags` / `identity` / `plan` and the same inline `arm-resource-invalid-envelope-property` / `arm-no-record` suppressions. Since this PR is a faithful migration of that contract (live customers, established SDK shape), changing the envelope shape here would diverge from the source RP and break SDK compatibility for existing callers.

Switching to `TrackedResource` would also move `location` from a path key to a body property, which is a different routing model than the one Kronox actually implements server-side. If ARM modeling guidance has changed since the source RP was approved, we'd rather address it as a coordinated update to both RPs in a future API version than break parity in this migration.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json
ykh015: Fixed. `LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json` now has:
- `name` parameter `3ec2ab23-9f13-4328-85c8-21928acbc7b8` (operation id, no longer the subscriptionId GUID)
- VM `name: "vm1"`, `type: "Microsoft.Compute/virtualMachines"`
- realistic ARM `id` for the VM under the resource group
- descriptive `$filter: "operationStatus eq 'Failed'"` and `$skiptoken: "opaque-page-token"`
- realistic `nextLink` URL with `api-version` and skiptoken

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
ykh015: Acknowledged. We deliberately keep `Creating` and `Deleting` as explicit provisioning states (they reflect actual server-side state transitions the Kronox service emits), but we agree the three terminal states (`Succeeded`, `Failed`, `Canceled`) should come from `Azure.ResourceManager.ResourceProvisioningState` rather than being hand-rolled. Tracked as a follow-up; will land alongside the next API version since it doesn't change the wire contract.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_ListVirtualMachines_MaximumSet_Gen.json
ykh015: Fixed. All `LaunchBulkInstancesOperation_List*_MaximumSet_Gen.json` and `...ListVirtualMachines_MaximumSet_Gen.json` examples now use a realistic `nextLink` URL of the form `https://management.azure.com/subscriptions/.../providers/Microsoft.Compute/locations/{location}/launchBulkInstancesOperations[/{name}/virtualMachines]?api-version=2026-04-06-preview&$skiptoken=opaque-page-token` and `Operations_List_MaximumSet_Gen.json` uses `https://management.azure.com/providers/Microsoft.Compute/operations?api-version=2026-04-06-preview&$skiptoken=opaque-page-token`.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/examples/2026-04-06-preview/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
ykh015: Acknowledged. The envelope `identity` block is inherited from the autogenerated example and the resource keeps `ManagedServiceIdentityProperty` from the source RP shape (see thread on Proxy Resource modeling). The internal contradiction (`type: "None"` with populated `principalId`/`tenantId`) is autogen-noise that ARM SDK reviewers will see; we'll hand-edit the identity block to either drop it entirely from the example or use a consistent populated `SystemAssigned, UserAssigned` shape in a follow-up cleanup pass.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Acknowledged. `LaunchBulkInstancesOperationListResult.nextLink` is currently `string`. We'll switch to `url` in TSP so the emitted swagger picks up `format: uri` (matching the sibling `VirtualMachineListResult.nextLink`). Tracked as a follow-up — no wire-contract change.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
ykh015: Acknowledged. `ComputeProfile.computeApiVersion` is a freeform string today because the value is validated server-side against the list of supported Microsoft.Compute api-versions (which evolves out-of-band from this spec). A `@pattern("^[0-9]{4}-[0-9]{2}-[0-9]{2}(-preview)?$")` regex would be a low-risk add — tracked as a follow-up.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
ykh015: Acknowledged. `VirtualMachine.type` returning a freeform string is a defect — we'll constrain it to the literal `"Microsoft.Compute/virtualMachines"` (or drop it since `id` already encodes the type). Tracked as a follow-up.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/LaunchBulkInstancesOperation_Get_MaximumSet_Gen.json
ykh015: Fixed by the same change set as thread #1. Resource envelope `name` and `type` now use the real ARM resource type (`Microsoft.Compute/locations/launchBulkInstancesOperations`) and the canonical operation GUID. Other autogen filler in property values (e.g. `adminUsername`, `computerName`) is acknowledged but left unchanged in this round to keep the diff scoped — tracked as a follow-up example-cleanup pass.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/main.tsp
ykh015: Acknowledged. The dual-RP situation (`Microsoft.ComputeBulkActions` for early private/preview customers and `Microsoft.Compute` for the canonical surface going forward) is an explicit decision coordinated with the Compute and ARM design leads. The plan is to deprecate `Microsoft.ComputeBulkActions` once `Microsoft.Compute` reaches GA; a service-README in the Kronox repo documents this. No spec change required.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/main.tsp
ykh015: Acknowledged. We consolidated into the existing `2026-04-06-preview` rather than introducing a new `2026-06-06-preview`, so the lifecycle clock starts from the existing version's release date and there's no second close-spaced preview. GA timing for `2026-04-06-preview` is tracked outside this PR; happy to share a target date offline.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Acknowledged. `LaunchBulkInstancesOperation_GetOperationStatus` is the only operation under `/subscriptions/.../Microsoft.Compute/locations/{location}/operations/{id}` added by this PR. Compute's existing operation-status surfaces use different operationIds and paths, so no collision today; we'll watch for collisions as the namespace grows. No spec change required.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Fixed in commit df93c2a27d. The `routes.tsp` alias key has been renamed `locationparameter` -> `location`, and the same rename has been applied to the 9 existing `VirtualMachineBulkOperations_Bulk*` POST example payloads in both the canon and mirror example folders. The `ParameterNotUsingCommonTypes` suppression in `readme.md` has been extended to cover these renamed paths.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/VirtualMachineBulkOperations_BulkReimage_MaximumSet_Gen.json
ykh015: Fixed in commit df93c2a27d. `VirtualMachineBulkOperations_BulkReimage_MaximumSet_Gen.json` now uses `"customData": "{base64-encoded-cloud-init-script}"` (clearly-fake placeholder) in both example folder copies.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/VirtualMachineBulkOperations_BulkCreate_MinimumSet_Gen.json
ykh015: Fixed in commit df93c2a27d. Both example titles now read `"<OperationId> - generated by [MinimumSet] rule"` with the duplicate MaximumSet suffix removed.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Acknowledged. `BulkactionVMExtensionProperties.protectedSettings` and `protectedSettingsFromKeyVault` should both carry `@secret` / `x-ms-secret: true`. This is also missing in the source `Microsoft.ComputeBulkActions/2026-02-01-preview` swagger, so we'll fix both RPs together in a coordinated follow-up rather than diverging in this PR. Tracked.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/examples/Operations_List_MaximumSet_Gen.json
ykh015: Fixed in commit 2f594faa27. `Operations_List_MaximumSet_Gen.json` now describes a representative real control-plane operation:
```json
{
  "name": "Microsoft.Compute/locations/launchBulkInstancesOperations/read",
  "isDataAction": false,
  "display": {
    "provider": "Microsoft Compute",
    "resource": "Bulk Instances Launch Operation",
    "operation": "Get Bulk Instances Launch Operation",
    "description": "Gets the status of a bulk virtual machine launch operation."
  },
  "origin": "user,system"
}
```
with a realistic ARM `nextLink` URL. `isDataAction: true` / `actionType: "Internal"` removed.

hardikginwala (approved):

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
mentat9: Modeling guidance has not changed: this is really basic ARM modeling, even if review/enforcement may not always consistent. This kind of model is confusing and counterproductive for customers, since it establishes anti-patterns as patterns. The difficulty of fixing it now just underscores the importance of getting it right in the first place. Here we have a temporary-to-be-deprecated RP implementing a new anti-pattern, which then has to be propagated to one of our most important RPs because compatibility. Now would be the time to fix this, since the other RP is slated for deprecation.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
mentat9: Why not fix it now?

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
mentat9: Again, why not now?

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
mentat9: Why wait until later to improve your spec?

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
mentat9: This is a spec bug that should be fixed now.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/launchBulkInstancesOperation.tsp
ykh015: @mentat9 We cannot make this change at this point. Microsoft.ComputeBulkActions is already in production with customers using LaunchBulkInstancesOperation, and the explicit goal of this migration to Microsoft.Compute is to make adoption easier for this customer (no additional subscription registration, simplified RBAC). The customer here uses their customer subscriptions to create VMs. 
Changing the resource modeling here would require corresponding contract changes on the server side and a coordinated customer side migration that we cannot undertake right now. 

For any new resource types we add, we will follow the guidance.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/preview/2026-06-06-preview/Bulkactions.json
ykh015: Fixed in commit f3a470d9. Added `@secret` to `BulkactionVMExtensionProperties.protectedSettings` and `protectedSettingsFromKeyVault` in `virtualMachineExtension.tsp`. The regenerated swagger now emits `x-ms-secret: true` and `format: password` on both properties, matching the convention in `virtualMachine.json`.

The same omission still exists in the source `Microsoft.ComputeBulkActions/2026-02-01-preview` RP; we will fix that in a coordinated follow-up so the two RPs remain consistent.

mentat9:
@ykh015 - Two asks:
1. Please apply (and help others from your team to apply) the feedback/learnings from this review to future APIs.
2. Please submit your specs for review before your APIs reach a state in which they can't be changed. Otherwise, it's a drain on ARM review resources with very little benefit to anyone.

mentat9:
@ykh015 - Looks like you are blocked on ModelValidation. You may need to add suppressions.

specification/compute/resource-manager/Microsoft.Compute/Bulkactions/readme.md
```diff
@@ -59,4 +59,61 @@ These settings apply only when `--tag=package-2026-04-06-preview` is specified o
 ```yaml $(tag) == 'package-2026-04-06-preview'
 input-file:
   - preview/2026-04-06-preview/Bulkactions.json
+suppressions:
```
haagha, line 62: These are adding version specific suppression only I believe. If you need to add suppressions for all versions you can add to suppressions section above

haagha (approved):
1 suppression comment

PR #43278, opened 2026-05-16
Fix memory item CRUD to align with implementation

bojunehsu:
* Change list memories path to `POST /memory_stores/{name}/items:list`
  * We needed to use `POST :list` because `scope` is a required request body parameter, which may contain PII.
* Add scope and optional kind to list memories.
* Change memory item path to `/memory_stores/{name}/items`.
* Change delete response name to `memory_store.item.deleted` and remove `name` from response.
* Add missing MemoryStoreRequiredPreviews meta tag.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
+++ b/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
@@ -30,7 +30,13 @@
       "name": "Evaluation Rules"
     },
     {
+      "name": "EvaluationSuiteGenerationJobs"
+    },
+    {
       "name": "Evaluators"
+    },
+    {
+      "name": "EvaluatorGenerationJobs"
     },
     {
       "name": "Indexes"
@@ -60,6 +66,9 @@
       "name": "Redteams"
     },
     {
+      "name": "Routines"
+    },
+    {
       "name": "Schedules"
     },
     {
@@ -67,6 +76,9 @@
     },
     {
       "name": "Toolboxes"
+    },
+    {
+      "name": "DataGenerationJobs"
     },
     {
       "name": "AgentOptimizationJobs"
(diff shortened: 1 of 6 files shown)


dargilco (approved):

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
```diff
@@ -29,9 +29,15 @@
     {
       "name": "Evaluation Rules"
     },
+    {
```
tolia-msft, line 32: What's this?

tolia-msft (approved):

specification/ai-foundry/data-plane/Foundry/src/memory-stores/routes.tsp
```diff
@@ -161,9 +161,9 @@ interface MemoryStores {
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
   @doc("List all memory items in a memory store.")
   @operationId("listMemories")
-  @get
+  @post
```
glecaros, line 164: this seems very non-standard, what's the reason to move out of the normal get pattern?

bojunehsu: We only allow listing of memories for a specific scope, which may contain personal data such as email or employee id.  Thus, we cannot specify scope in path or query as the url is often logged.  If scope is specified in the body, is it awkward to use GET.  So given the constraints, we ended up with this design.

nick863 (approved):

glecaros:
Had to revert and recreated the PR as #43313

specification/ai-foundry/data-plane/Foundry/src/memory-stores/routes.tsp
glecaros: we have prior art for this with the user isolation key header. I would expect a header is appropriate for this.

PR #43243, opened 2026-05-15
feat(agents-session-files): add has_more, @list, and optional path to ListSessionFiles

amitbhave10:
### Changes
1. **`has_more: boolean`** added to `SessionDirectoryListResponse` model — enables future pagination without breaking backward compatibility
2. **`@list` decorator** added to the `listSessionFiles` operation — follows Foundry SDK guidelines for list APIs
3. **`path` made optional** — when omitted, the server lists the home directory ($HOME)

### Context
- Companion AgentsV2 implementation PR: [vienna#2103392](https://msdata.visualstudio.com/Vienna/_git/vienna/pullrequest/2103392)
- Follows [Foundry SDK paged result guidelines](https://github.com/coreai-microsoft/foundrysdk_specs/blob/main/guidelines/service-patterns.md#response-shape-agentspagedresultt)
- `has_more` will be hardcoded `false` until ADC supports pagination
- Original Files API spec: https://github.com/Azure/azure-rest-api-specs/pull/41586

### Files Changed
- `specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp` — added `has_more` property
- `specification/ai-foundry/data-plane/Foundry/src/agents-session-files/routes.tsp` — added `@list`, made `path` optional, updated docs

PR #43280, opened 2026-05-16
feat(agents-session-files): add has_more, @list, and optional path to ListSessionFiles

glecaros:
## What
Add pagination support and `@list` decorator to the `listSessionFiles` operation, and make the `path` query parameter optional (defaults to home directory).

## Why
PR #43243 added `has_more` via a custom `SessionDirectoryListResponse`, but the Foundry standard pagination pattern uses `AgentsPagedResult<T>` with `CommonPageQueryParameters` (`limit`, `order`, `after`, `before`). This PR adopts the standard pattern for consistency with all other list operations.

## Changes
- **models.tsp**: Replaced `SessionDirectoryListResponse` with `ListSessionFilesParameters` model that includes `CommonPageQueryParameters` and `UserIsolationKeyHeader`, with optional `path` query param
- **routes.tsp**: Updated `listSessionFiles` to use `AgentsPagedResult<SessionDirectoryEntry>` response and `@list` decorator, referencing the new parameters model

## Replaces
PR #43243 (which targeted `feature/foundry-staging`)

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
@@ -35,17 +35,7 @@ model SessionDirectoryEntry {
   modified_time: utcDateTime;
 }
 
-@doc("Parameters for listing files in a session sandbox directory.")
-model ListSessionFilesParameters {
-  /** The name of the agent. */
-  @path agent_name: string;
-
-  /** The session ID. */
-  @path agent_session_id: string;
-
-  /** The directory path to list, relative to the session home directory. Defaults to the home directory if not provided. */
-  @query path?: string;
-
-  ...UserIsolationKeyHeader;
-  ...CommonPageQueryParameters;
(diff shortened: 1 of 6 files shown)


dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents-session-files/models.tsp
```diff
@@ -33,11 +35,7 @@ model SessionDirectoryEntry {
   modified_time: utcDateTime;
 }
 
-@doc("Response from listing a directory in a session sandbox.")
-model SessionDirectoryListResponse {
+model SessionDirectoryListResponse is AgentsPagedResult<SessionDirectoryEntry> {
```
amitbhave10: Problem is AgentsPagedResult has all list in "data" field but we are already sending the same in "entries" field.

glecaros: I inlined the page result (sans data)

amitbhave10 (approved):

PR #43297, opened 2026-05-18
[eventhub] fix for EHNamespaces.update

welovej:
fix for EHNamespaces.update


pshao25 (approved):

mentat9:
@welovej - This tsp change should also include the corresponding swagger change. I think it should also have been flagged as a breaking change (which might happen when the swagger is added). If that happens, the Next Steps to Merge will tell you how to proceed.

welovej:
> @welovej - This tsp change should also include the corresponding swagger change. I think it should also have been flagged as a breaking change (which might happen when the swagger is added). If that happens, the Next Steps to Merge will tell you how to proceed.

already compiled. There is no swagger changes.

pshao25:
@mentat9 TypeSpec Validation will fail if we didn't do compilation. OpenAPI emitter outputs long-running flag according to response header. That's why you didn't see any swagger change.

PR #43306, opened 2026-05-18
Add version override header

andscho-msft:

Add version override header

## PR review workflow diagram

- [ ] a private preview
- [ ] a public preview
- [x] GA release

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
@@ -45,7 +45,7 @@ alias UserIsolationKeyHeader = {
  */
 alias VersionOverrideHeader = {
   /** Optional agent version override. Overrides the pinned/latest version selected by the endpoint's version selector for this request only. */
-  @header("x-agent-version-override") x_agent_version_override?: string;
+  @header("x-agent-version-override") agent_version_override?: string;
 };
 
 #suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -37,6 +37,17 @@ alias UserIsolationKeyHeader = {
...
 
+/**
+ * Optional request header that overrides the agent version selected for an endpoint-scoped
+ * invocation. When supplied with a valid version, the service routes the request to that
+ * version instead of the endpoint's pinned/latest version; if the supplied version does not
+ * exist, the service falls back to the pinned/latest version.
+ */
+alias VersionOverrideHeader = {
+  /** Optional agent version override. Overrides the pinned/latest version selected by the endpoint's version selector for this request only. */
+  @header("x-agent-version-override") x_agent_version_override?: string;
```
dargilco: I think we should have a different property name here. This will certainly need to be overridden for all emitted SDKs, so might as well pick a better name here. Can we drop the "x_" and call it agent_version_override?

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/common/models.tsp
```diff
@@ -37,6 +37,17 @@ alias UserIsolationKeyHeader = {
   @header("x-ms-user-isolation-key") user_isolation_key?: string;
 };
 
+/**
+ * Optional request header that overrides the agent version selected for an endpoint-scoped
+ * invocation. When supplied with a valid version, the service routes the request to that
+ * version instead of the endpoint's pinned/latest version; if the supplied version does not
+ * exist, the service falls back to the pinned/latest version.
+ */
```
dargilco, line 45: I don't think doc strings attached to an alias get emitted in SDK ref docs (or REST API ref docs...). So please move this more detailed doc string and place it above the definition of the `agent_version_override` property below. I don't think we need a doc string for the alias itself.

dargilco: Please include the OpenAPI3 files in your PR. Thanks!

PR #43308, opened 2026-05-18
Fix TypeSpec LRO operations: remove @extension, use @useFinalStateVia

RenzoPrettoMS:
Addresses TypeSpec review feedback that @extension should never be used to mark LRO operations because it only affects the generated OpenAPI and is invisible to SDK code generators. As a result the affected operations appeared as long-running in OpenAPI but were generated as plain synchronous methods in the Java SDK
(Azure/azure-sdk-for-java#49171).

1. Removed @extension(x-ms-long-running-operation, ...) decorators from the custom LRO templates in lro-helpers.tsp and replaced them with @Azure.Core.useFinalStateVia(location), which both the OpenAPI emitter and SDK generators understand.

2. Removed @pollingOperation(ScenarioRuns.get) from ScenarioConfigurations.execute and ScenarioRuns.cancel. The polling endpoint pointed at a non-status-monitor GET, which prevented SDK generators from detecting these as LROs. Polling now uses the existing @lroStatus / @lroSucceeded / @lroFailed / @lroCanceled annotations on ScenarioRunProperties.status.

3. Replaced the hand-written GET on ScenarioRun (which returned both 200 and 202 with Location/Retry-After headers in 202) with the standard ArmResourceRead<ScenarioRun> template. The GET now returns only 200, in line with ARM and SDK conventions.

4. Cleaned up the no-longer-needed @azure-tools/typespec-azure-core/no-openapi, arm-post-operation-response-codes, polling-operation-no-status-monitor, no-response-body, and arm-resource-operation suppressions in scenarioConfiguration.tsp, scenarioRun.tsp, and workspace.tsp.

Regenerated OpenAPI for preview/2026-05-01-preview reflects the new state: LRO operations carry x-ms-long-running-operation: true with x-ms-long-running-operation-options.final-state-via: location, and ScenarioRuns_Get is documented with a single 200 response.

# Choose a PR Template


RenzoPrettoMS:
The scenario run get endpoint effectively acts as the "OperationResults" endpoint for the "execution" POST LRO. From previous guidance we received, we are removing this definition of it from the OpenAPI spec as to not confuse customers and make sure that the SDK generates as expected.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/examples/2026-05-01-preview/ScenarioRuns_Get.json
```diff
@@ -82,37 +82,6 @@
           "lastModifiedByType": "Application"
         }
       }
-    },
-    "202": {
```
weidongxu-microsoft, line 86: Please double check whether the service indeed does not return 202, while the `ScenarioRun` is in progress.

From my memory of the working of polling via `Location` header, the poll continues on 202, and stops on 200 (or 204 on DELETE).
This means, if this API only return 200, the `ScenarioConfigurations.execute` (if its `Location` points to this GET), would stop on 200 -- regardless of `ScenarioRunProperties.status` value.

The `ScenarioRunProperties.status` value likely not read during the poll (the only property I remember would affect poll would be the `provisioningState`, and it probably only works in `Azure-AsyncOperation` header).

PS: if necessary, the `ArmResourceRead` likely can be written with a response with 200+202 -- use its `Response` template variable.

RenzoPrettoMS: Good catch, thank you!

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioRun.models.tsp
```diff
@@ -5,12 +5,39 @@ import "@azure-tools/typespec-azure-resource-manager";
...
+ * The body reflects the current run state at the time of the poll.
+ */
+model ScenarioRunInProgressResponse {
+  ...AcceptedResponse;
+
+  /**
+   * The URL where the status of the long-running operation can be polled.
+   */
+  @header("Location")
+  location: string;
```
mentat9: >string [](http://example.com/codeflow?start=12&length=6)

Recommend Url instead of open-ended string.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioRun.models.tsp
```diff
@@ -5,12 +5,39 @@ import "@azure-tools/typespec-azure-resource-manager";
...
+
+  /**
+   * The number of seconds the client should wait before polling again.
+   */
+  @header("Retry-After")
+  retryAfter?: int32;
```
mentat9, line 35: >int32 [](http://example.com/codeflow?start=15&length=5)

Consider adding min, max constraints.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioRun.tsp
```diff
@@ -11,37 +10,28 @@ using Azure.Core;
...
-        retryAfter?: int32;
-
-        @bodyRoot body: ScenarioRun;
-      })
-    | ErrorResponse;
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-operation" "LRO polling endpoint that returns body in both 200 and 202 responses."
+  #suppress "@azure-tools/typespec-azure-resource-manager/no-response-body" "LRO polling endpoint returns body in 202 to reflect the current run state."
+  get is ArmResourceRead<
+    ScenarioRun,
+    Response = ArmResponse<ScenarioRun> | ScenarioRunInProgressResponse
```
mentat9, line 33: > | ScenarioRunInProgressResponse [](http://example.com/codeflow?start=39&length=32)

ARM GET never returns 202. GET is always synchronous and returns 200 or error status. If you need to return transition status, use a state variable with a transition state, e.g. runState: { Complete, Running } or similar.

RenzoPrettoMS: This endpoint effectively acts like an `OperationResults` endpoint, where a user (or SDK) can poll it to get the status. This approach was discussed and approved here https://github.com/Azure/azure-rest-api-specs/pull/41486, and again here https://github.com/Azure/azure-rest-api-specs/pull/43085. These changes are to allow the SDK to properly generate.

weidongxu-microsoft:
You can add `versioning-approved-benign` (or `-bugfix` depends on the nature) label to the PR, if there is **no change in backend**.

RenzoPrettoMS:
Breaking change is anticipated due to the changes requested by reviewer on SDK PR.

weidongxu-microsoft:
> Breaking change is anticipated due to the changes requested by reviewer on SDK PR.

Correction: Java SDK dev never requested service make breaking change to REST API

RenzoPrettoMS:
> > Breaking change is anticipated due to the changes requested by reviewer on SDK PR.
> 
> Correction: Java SDK dev never requested service make breaking change to REST API

I apologize, I realize my wording was not what was intended. Our service does not have any breaking change, we are simply updating the contract and Typespec to more accurately portray our service behavior. There is **NO** service behavior change, instead these changes just support SDK generation.

XiaofeiCao (approved):

PR #43325, opened 2026-05-19
[Storage File] preserve content type header

gracewilcox:
Adding decorator to keep Content-Type headers for specific operation in Go to preserve back compatibility

### DIFF
--- a/specification/storage/data-plane/FileStorage/client.tsp
+++ b/specification/storage/data-plane/FileStorage/client.tsp
@@ -164,6 +164,10 @@ model SignedIdentifiersGo is Array<SignedIdentifier>;
 
 @@clientName(ShareItemInternal, "Share", "go");
 @@clientName(ShareNfsSettings, "ShareNFSSettings", "go");
+@@clientName(ShareNfsSettingsEncryptionInTransit,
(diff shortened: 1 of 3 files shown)


specification/storage/data-plane/FileStorage/client.tsp
```diff
@@ -211,6 +211,28 @@ model SignedIdentifiersGo is Array<SignedIdentifier>;
 @@clientName(UserDelegationKey.signedTid, "SignedTID", "go");
 #suppress "@azure-tools/typespec-client-generator-core/client-option"
 @@clientOption(HandleItem, "omitSerdeMethods", "all", "go");
+@@clientOption(Directory.listFilesAndDirectoriesSegment,
```
catalinaperalta: probably need a suppression over each decorator

catalinaperalta (commented):
need to get tsp validation passing before this can get merged in, look at the check for info

specification/storage/data-plane/FileStorage/client.tsp
gracewilcox: fixed

catalinaperalta (approved):

PR #43352, opened 2026-05-19
Delete specification/changeanalysis directory

CarlosCampos06:
The Change Analysis was deprecated, retired and removed a while ago and the infrastructure doesn't exist anymore.


lmazuel (approved):
Discussed offline this is already offline.

See also: https://azure.microsoft.com/en-us/updates?id=467340

PR #43365, opened 2026-05-20
containerservice: PreparedImageSpecification 2026-02-02-preview API updates

benjamin-brady:
## Changes in this PR

This PR corrects preview API specs for `Microsoft.ContainerService/preparedimagespecification` and the deprecated `Microsoft.ContainerService/nodecustomization` APIs. The public swagger currently documents fields and operations that **never functioned** in the service. This PR removes those non-functional elements so the spec matches the actual API contract.

The affected APIs are preview-only. The Prepared Image Specification feature is behind feature gating and is not broadly in use.

### 1. Correct PreparedImageSpecification PATCH to ARM tags only (spec correction)

Fixes ADO task [#38037921](https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/38037921).

The `PreparedImageSpecification` PATCH operation was mistakenly documented as accepting mutable body properties (`containerImages`, `identityProfile`, `version`, `customizationScripts`). These fields **never functioned** on PATCH — the service only supports updating ARM `tags` via PATCH. This PR removes the incorrectly documented properties so the swagger matches the actual service contract.

- Removed `PreparedImageSpecificationPatchProperties` and related update-only models
- Changed `PreparedImageSpecificationPatch` to carry only ARM `tags`
- Updated PATCH examples to send only `tags`

### 2. Make PreparedImageSpecification `version` customer-provided on create

The service expects `version` to be supplied by the customer at create time (immutable afterwards). The swagger previously marked it as read-only, which incorrectly hid this required create-time input from clients.

- Removed `@visibility(Lifecycle.Read)` from `PreparedImageSpecificationProperties.version`
- Updated docs to clarify that `version` is client-provided and immutable after creation
- Limited the version format to 63 characters (matching Kubernetes label value length constraints)
- Added `version` to create/update examples

### 3. Align deprecated NodeCustomization PATCH body and version behavior

The deprecated `Microsoft.ContainerService/nodecustomization` preview APIs have the same body-shape and `version` issue — PATCH was documented with properties that never functioned, and `version` was incorrectly marked read-only. To minimize LRO-related changes in this PR, NodeCustomization PATCH remains documented as async; this PR only corrects the PATCH request body shape and `version` visibility/format.

- Changed NodeCustomization update model to tags-only (keeping original `NodeCustomizationUpdate` name to minimize diff)
- Kept NodeCustomization PATCH LRO shape unchanged as async to minimize behavior changes in this PR
- Made `version` customer-provided (same semantics as PIS)
- Updated examples for `2025-08-02-preview` and `2025-09-02-preview`

## Breaking change review context

- All changes affect **preview** API versions only.
- The Prepared Image Specification feature is behind feature gating and is not broadly in use.
- RP telemetry over 180 days shows 4 distinct subscriptions, 2 using PATCH (5 total PATCH calls, all HTTP 200). Request bodies are not logged, so field-level usage cannot be confirmed — but the fields never functioned regardless.
- Following `aka.ms/brch` for approval to correct the existing preview version.
- `PreparedImageSpecificationVersion` path-key tightening to `{1,63}` is intentional and in-scope for this same preview breaking-change approval context.

## Due diligence checklist

- [x] TypeSpec compiles cleanly
- [x] Generated swagger regenerated from TypeSpec
- [x] Examples updated to reflect actual API behavior

---

### ARM API Changes

<details>
<summary>Expand for ARM template, Bicep, Terraform, and Policy impact</summary>

These changes correct preview specs. The fields being removed from PATCH never functioned — no existing ARM templates, Bicep files, or Terraform configurations successfully use them today.

</details>

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
@@ -122,6 +122,11 @@ union ProvisioningState {
   Accepted: "Accepted",
 }
 
+@doc("Node customization patch resource")
+model NodeCustomizationPatch {
+  ...Azure.ResourceManager.Foundations.ArmTagsProperty;
+}
+
 @parentResource(NodeCustomization)
 @doc("A version of the Node Customization resource.")
 model NodeCustomizationVersion is ProxyResource<NodeCustomizationProperties> {
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
@@ -129,9 +129,9 @@ model NodeCustomizationUpdate {
 @parentResource(NodeCustomization)
 @doc("A version of the Node Customization resource.")
 model NodeCustomizationVersion is ProxyResource<NodeCustomizationProperties> {
-  @pattern("^[\\w\\-\\.]+$")
+  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
   @minLength(1)
-  @maxLength(64)
+  @maxLength(63)
   @doc("The version of the Node Customization.")
   @key("version")
   @path
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
@@ -37,9 +37,8 @@ model NodeCustomizationProperties {
     """)
   identityProfile?: Foundations.UserAssignedIdentity;
 
-  @visibility(Lifecycle.Read)
-  @doc("An auto-generated value that changes when the other fields of the image customization are changed.")
-  @pattern("^[\\w\\-\\.]+$")
+  @doc("The client-provided version of the node customization.")
+  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
   version?: string;
 
   @visibility(Lifecycle.Read)
@@ -122,8 +121,8 @@ union ProvisioningState {
   Accepted: "Accepted",
 }
 
-@doc("Node customization patch resource")
-model NodeCustomizationPatch {
+@doc("The type used for update operations of the NodeCustomization.")
+model NodeCustomizationUpdate {
   ...Azure.ResourceManager.Foundations.ArmTagsProperty;
 }
 
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
@@ -22,10 +22,10 @@ interface NodeCustomizations {
   >;
 
   #deprecated "Node Customization APIs are deprecated. Use Prepared Image Specification APIs instead."
-  @doc("Deprecated: Update the tags of a node customization. Use Prepared Image Specification APIs instead.")
+  @doc("Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.")
   update is ArmCustomPatchSync<
     NodeCustomization,
-    PatchModel = NodeCustomizationPatch,
+    PatchModel = NodeCustomizationUpdate,
     Parameters = IfMatchParameters<NodeCustomization>
   >;
 
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
@@ -23,7 +23,7 @@ interface NodeCustomizations {
 
   #deprecated "Node Customization APIs are deprecated. Use Prepared Image Specification APIs instead."
   @doc("Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.")
-  update is ArmCustomPatchSync<
+  update is ArmCustomPatchAsync<
     NodeCustomization,
     PatchModel = NodeCustomizationUpdate,
     Parameters = IfMatchParameters<NodeCustomization>
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
@@ -22,15 +22,11 @@ interface NodeCustomizations {
   >;
 
   #deprecated "Node Customization APIs are deprecated. Use Prepared Image Specification APIs instead."
-  @doc("Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.")
-  update is ArmCustomPatchAsync<
+  @doc("Deprecated: Update the tags of a node customization. Use Prepared Image Specification APIs instead.")
+  update is ArmCustomPatchSync<
     NodeCustomization,
-    Azure.ResourceManager.Foundations.ResourceUpdateModel<
-      NodeCustomization,
-      NodeCustomizationProperties
-    >,
-    Azure.ResourceManager.Foundations.BaseParameters<NodeCustomization> &
-      IfMatchParameters<NodeCustomization>
+    PatchModel = NodeCustomizationPatch,
+    Parameters = IfMatchParameters<NodeCustomization>
   >;
 
   #deprecated "Node Customization APIs are deprecated. Use Prepared Image Specification APIs instead."
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/preview/2025-08-02-preview/nodecustomization.json
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/preview/2025-08-02-preview/nodecustomization.json
@@ -314,7 +314,7 @@
         "tags": [
           "NodeCustomizations"
         ],
-        "description": "Deprecated: Update the tags of a node customization. Use Prepared Image Specification APIs instead.",
+        "description": "Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.",
         "parameters": [
           {
             "$ref": "../../../../../../common-types/resource-management/v5/types.json#/parameters/ApiVersionParameter"
@@ -349,7 +349,7 @@
             "description": "The resource properties to be updated.",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/NodeCustomizationPatch"
+              "$ref": "#/definitions/NodeCustomizationUpdate"
             }
           }
         ],
@@ -707,19 +707,6 @@
         "value"
       ]
     },
-    "NodeCustomizationPatch": {
-      "type": "object",
-      "description": "Node customization patch resource",
-      "properties": {
-        "tags": {
-          "type": "object",
-          "description": "Resource tags.",
-          "additionalProperties": {
-            "type": "string"
-          }
-        }
-      }
-    },
     "NodeCustomizationProperties": {
       "type": "object",
       "description": "The properties of the Node Customization resource.",
@@ -737,9 +724,8 @@
         },
         "version": {
           "type": "string",
-          "description": "An auto-generated value that changes when the other fields of the image customization are changed.",
-          "pattern": "^[\\w\\-\\.]+$",
-          "readOnly": true
+          "description": "The client-provided version of the node customization.",
+          "pattern": "^(?![.-])[A-Za-z0-9_.-]{1,63}$"
         },
         "provisioningState": {
           "$ref": "#/definitions/ProvisioningState",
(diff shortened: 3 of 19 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
```diff
@@ -122,6 +122,11 @@ union ProvisioningState {
   Accepted: "Accepted",
 }
 
+@doc("Node customization patch resource")
+model NodeCustomizationPatch {
```
jim-minter: I'd consider calling this `NodeCustomizationUpdate`, with docstring `The type used for update operations of the NodeCustomization."` to reduce the diff.

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
```diff
@@ -22,15 +22,11 @@ interface NodeCustomizations {
   >;
 
   #deprecated "Node Customization APIs are deprecated. Use Prepared Image Specification APIs instead."
-  @doc("Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.")
-  update is ArmCustomPatchAsync<
+  @doc("Deprecated: Update the tags of a node customization. Use Prepared Image Specification APIs instead.")
+  update is ArmCustomPatchSync<
```
jim-minter: I think the changes on these lines are reasonable, but perhaps not essential?  Reverting them may reduce noise.

jim-minter (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/preview/2025-08-02-preview/nodecustomization.json
```diff
@@ -343,13 +335,21 @@
...
+            "x-ms-client-name": "ifMatch"
+          },
           {
             "name": "properties",
             "in": "body",
             "description": "The resource properties to be updated.",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/NodeCustomizationUpdate"
+              "$ref": "#/definitions/NodeCustomizationPatch"
```
matthchr: minor: I think these are often called `XyzUpdate` - why change the name rather than just remove the fields that never worked?

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
matthchr: +1, IMO the best argument we have for making the change in existing API versions is that these fields never actually worked. They were added to the typespec by mistake and the service has never accepted PATCH-ing those fields.

Which I think suggests the best fix is to fix what never worked, but leave the rest the same. So avoid renaming things. Docs updates are obviously fine (especially if they clarify what _is_ patchable).

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/models.tsp
```diff
@@ -38,9 +38,8 @@ model PreparedImageSpecificationProperties {
     """)
   identityProfile?: PreparedImageSpecificationManagedIdentityProfile;
 
-  @visibility(Lifecycle.Read)
-  @doc("An auto-generated value that changes when the other fields of the image customization are changed.")
-  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,128}$")
+  @doc("The client-provided version of the prepared image specification.")
+  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
```
jim-minter, line 42: I also count this `@pattern` change as reasonable, but perhaps not essential, and would consider not making it.

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/models.tsp
```diff
@@ -38,9 +38,8 @@ model PreparedImageSpecificationProperties {
     """)
   identityProfile?: PreparedImageSpecificationManagedIdentityProfile;
 
-  @visibility(Lifecycle.Read)
```
jim-minter, line 41: Do you need to change the visibility in /specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp as well?

matthchr (approved):

jim-minter (approved):
LGTM

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/models.tsp
```diff
@@ -209,9 +160,9 @@ union ProvisioningState {
 @doc("A version of the Prepared Image Specification resource.")
 model PreparedImageSpecificationVersion
   is ProxyResource<PreparedImageSpecificationProperties> {
-  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,128}$")
+  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
   @minLength(1)
-  @maxLength(128)
+  @maxLength(63)
```
ravimeda, line 165: **[NEW] 🟡 Warning** **[Versioning — undocumented path-key constraint tightening]**

The path parameter `{version}` on the `PreparedImageSpecificationVersion` proxy child has its constraints tightened in this PR:
- **Base:** `pattern: ^(?![.-])[A-Za-z0-9_.-]{1,128}$`, `maxLength: 128`
- **Head:** `pattern: ^(?![.-])[A-Za-z0-9_.-]{1,63}$`, `maxLength: 63`

This is a **separate** breaking change from the PATCH-body removal and the `readOnly` → writable conversion that the PR explicitly calls out. Because `properties.version` was previously `readOnly` (server-generated), the service may have emitted version strings longer than 63 characters that customers will no longer be able to address via `GET`/`DELETE /versions/{version}` after this PR ships.

The PR description's `aka.ms/brch` approval narrative covers the PATCH body change and the `version` writability change, but does **not** explicitly cover this path-parameter tightening.

**Suggested fix:** Confirm in the PR description that the service has never emitted version strings longer than 63 characters (or that any such resources have been migrated), and that this constraint tightening is part of the same board approval. If unsure, raise this with the ContainerService RP team before merging.

**Classification:** `[NEW]` — newly introduced by this PR (base had `{1,128}`).

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/operations.tsp
```diff
@@ -11,15 +11,15 @@ namespace Microsoft.ContainerService;
 interface PreparedImageSpecifications {
   @doc("Get a prepared image specification at the latest version.")
   get is ArmResourceRead<PreparedImageSpecification>;
-  @doc("Create or update a prepared image specification resource. A new version is created only when the resource content changes.")
+  @doc("Create or update a prepared image specification resource with a client-provided version.")
```
ravimeda: **[NEW] 💡 Suggestion** **[Operation doc clarity for client-provided immutable `version`]**

The PR description states that `version` is "client-provided and immutable after creation". With `ArmResourceCreateOrUpdateAsync` ("create **or update**"), a customer may reasonably expect that a subsequent `PUT` with a different `properties.version` value updates the resource. If the service rejects that (because version is immutable post-create), the `createOrUpdate` name is misleading and the current doc ("with a client-provided version") leaves this ambiguous.

**Suggested fix:** Clarify in the operation `@doc` what happens on a subsequent `PUT` with a different `version`. For example:
> "Create or update a prepared image specification resource with a client-provided version. After creation, `properties.version` is immutable; subsequent PUT operations must specify the same version value (other property changes follow normal update semantics)."

(Adjust to match the actual service contract.)

**Classification:** `[NEW]` — newly introduced doc string in this PR.

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
```diff
@@ -25,12 +25,8 @@ interface NodeCustomizations {
   @doc("Deprecated: Update a node customization. Use Prepared Image Specification APIs instead.")
   update is ArmCustomPatchAsync<
     NodeCustomization,
-    Azure.ResourceManager.Foundations.ResourceUpdateModel<
-      NodeCustomization,
-      NodeCustomizationProperties
-    >,
-    Azure.ResourceManager.Foundations.BaseParameters<NodeCustomization> &
-      IfMatchParameters<NodeCustomization>
+    PatchModel = NodeCustomizationUpdate,
```
ravimeda, line 28: **[NEW] 💡 Suggestion** **[NC PATCH stays async for a tags-only body — inconsistent with PIS]**

Both `PreparedImageSpecificationPatch` and `NodeCustomizationUpdate` are now **tags-only** bodies. PIS uses `ArmCustomPatchSync` (200/default synchronously); NC retains `ArmCustomPatchAsync` (LRO). For a tags-only PATCH, LRO ceremony forces every SDK consumer of the deprecated NC API to **poll for a tag write** — operationally surprising and an unnecessary divergence from the sibling PIS API.

The PR description acknowledges this was kept "to minimize LRO-related changes in this PR" — understandable for a deprecated API, but it bakes the inconsistency permanently into the spec.

**Suggested fix (pick one):**
- (a) Align NC PATCH to `ArmCustomPatchSync` in this PR (the body shape change already requires regeneration, so an LRO removal is in the same change set).
- (b) Add a code comment noting the async shape is preserved only to avoid in-place LRO changes on a deprecated API, so future maintainers don't propagate it.

**Classification:** `[NEW]` — the inconsistency is newly visible because PIS PATCH was changed to sync in this PR.

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
```diff
@@ -37,9 +37,8 @@ model NodeCustomizationProperties {
     """)
   identityProfile?: Foundations.UserAssignedIdentity;
 
-  @visibility(Lifecycle.Read)
-  @doc("An auto-generated value that changes when the other fields of the image customization are changed.")
-  @pattern("^[\\w\\-\\.]+$")
+  @doc("The client-provided version of the node customization.")
+  @pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
```
ravimeda, line 41: **[NEW] 💡 Suggestion** **[Property↔path-key pattern asymmetry — NodeCustomization]**

This PR tightens `NodeCustomizationProperties.version` (this line) to `^(?![.-])[A-Za-z0-9_.-]{1,63}$` — no leading `.`/`-`, max 63 chars. However, the corresponding path-key `{version}` on `NodeCustomizationVersion.name` (around `models.tsp` lines 132 and 134) retains the previous `^[\w\-\.]+$` with `maxLength: 64` (allows leading `.`/`-`, max 64).

A customer could `GET /versions/.foo` (valid against the path pattern) for a version string `properties.version` would reject. SDK client-side validators on `properties.version` will be **stricter** than what the URL accepts. The asymmetry is **newly introduced** by this PR (base had `^[\w\-\.]+$` on both sides) and affects both `2025-08-02-preview` and `2025-09-02-preview`.

Note: PIS does **not** have this issue — both property and path-key are aligned to `^(?![.-])[A-Za-z0-9_.-]{1,63}$`.

**Suggested fix:** Align the `NodeCustomizationVersion.name` path-key pattern/`maxLength` to match the property:
```tsp
@pattern("^(?![.-])[A-Za-z0-9_.-]{1,63}$")
@minLength(1)
@maxLength(63)
```
Or document explicitly why they diverge.

**Classification:** `[NEW]` — asymmetry newly created by this PR's property-pattern change.

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/operations.tsp
benjamin-brady: This is by design as this is how our RP backend currently works. NC is a deprecated resource only used internally with essentially zero users so alignment is not super important. As this is a breaking change PR I'd like to keep the footprint minimal.

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/operations.tsp
benjamin-brady: Corrected. I amended the latest commit (`e992e2c`) to remove the same-version wording. The updated createOrUpdate doc now states that created versions are immutable and a different `properties.version` value should be provided to create a new version.

specification/containerservice/resource-manager/Microsoft.ContainerService/preparedimagespecification/models.tsp
benjamin-brady: Addressed in PR description. Added explicit note that the PreparedImageSpecificationVersion path-key tightening to `{1,63}` is intentional and in-scope for the same preview breaking-change approval context.

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/models.tsp
benjamin-brady: Addressed in amended commit `e992e2c`. `NodeCustomizationVersion.name` path-key validation now matches `properties.version` (`^(?![.-])[A-Za-z0-9_.-]{1,63}$`, maxLength 63) in TypeSpec and both generated preview swagger files.

benjamin-brady:
Prepared Image Specifications is only used internally right now on a handful of dev clusters for testing. Node customization (the old naming) has essentially zero users now and is gated behind a manual feature flag approval.

ravimeda (approved):

PR #43367, opened 2026-05-20
[compute] Add FinalResult and tags

melina5656:


specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/RestorePoint.tsp
```diff
@@ -50,7 +50,7 @@ interface RestorePoints {
     RestorePoint,
     Response = ArmResourceCreatedResponse<
       RestorePoint,
-      ArmLroLocationHeader<FinalResult = RestorePoint> &
+      ArmLroLocationHeader<FinalResult = void> &
```
tadelesh: Why change this? This will break the Go SDK.

melina5656: Restored.

pshao25 (approved):

PR #43375, opened 2026-05-20
Migrate ManagedNetworkFabric C# SDK generation

ArcturusZhang:
Migrates ManagedNetworkFabric C# SDK generation to the new management generator by adding C# emitter options, resource generation annotations, and C# compatibility naming customizations.Validation performed:- Regenerated Azure.ResourceManager.ManagedNetworkFabric locally with the mgmt generator- Verified resource hierarchy parity with the previous GA SDK- Built Azure.ResourceManager.ManagedNetworkFabric successfully

### Main TypeSpec changes outside `client.tsp`

- Reworked MNF resource declarations to use the standard ARM TypeSpec resource helpers (`TrackedResource<..., false>` / `ProxyResource<..., false>`) instead of directly extending `Azure.ResourceManager.Foundations.*Resource` and manually declaring the `properties` envelope. This removes the previous private resource-shape workaround decorators/suppressions while preserving the existing resource property documentation.

- Split `NetworkFabric` modeling for version compatibility. The current `NetworkFabric` shape is used for the latest API version, while `NetworkFabricDeprecated` is used by deprecated/shared-route operations that need to preserve the previous version’s resource shape. Related deprecated operations and list-result models were updated to reference the compatible model shape.

- Added .NET management emitter configuration to `tspconfig.yaml` (`package-name`, `namespace`, and emitter output directory) so the Azure.ResourceManager.ManagedNetworkFabric SDK can be generated directly from this TypeSpec project.


specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/models/AccessControlList.tsp
```diff
@@ -8,7 +8,7 @@ model AccessControlListProperties {
   ...AnnotationResource;
 
   @doc("Input method to configure Access Control List.")
-  configurationType: ConfigurationType;
+  configurationType?: ConfigurationType;
```
ArcturusZhang: we should never change optionality.
Please revert this on spec, and use custom code to mitigate it.

specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/models/NetworkToNetworkInterconnect.tsp
```diff
@@ -276,12 +276,12 @@ model OptionBLayer3Configuration extends Layer3IpPrefixProperties {
...
   @maxValue(4294967295)
   @minValue(1)
-  peerASN: int64;
+  peerASN?: int64;
 
   @doc("VLAN for CE/PE Layer 3 connectivity.Example : 501")
   @maxValue(4094)
   @minValue(100)
-  vlanId: int32;
+  vlanId?: int32;
```
ArcturusZhang: we should never change optionality.
Please revert this on spec, and use custom code to mitigate it.

specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/resources/AccessControlList.tsp
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.ManagedNetworkFabric;
 #suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" ""
 #suppress "@azure-tools/typespec-azure-core/no-private-usage" "Unsupported common type"
 @doc("The Access Control List resource definition.")
+@Azure.ResourceManager.Private.armResourceInternal(AccessControlListProperties)
 @Http.Private.includeInapplicableMetadataInPayload(false)
 model AccessControlList
   extends Azure.ResourceManager.Foundations.TrackedResource {
```
ArcturusZhang: the spec should not use private decorators in the first place.
I think we should rewrite this to:
```
model AccessControlList is TrackedResource<AccessControlListProperties> {
...
}
```
this might lead to a property order change in the generated swagger, it might be fine.

ArcturusZhang: same comment applies to all the resource models in this RP.

specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/resources/AccessControlList.tsp
```diff
@@ -11,18 +11,14 @@ namespace Microsoft.ManagedNetworkFabric;
 #suppress "@azure-tools/typespec-azure-core/no-private-usage" "Unsupported common type"
 @doc("The Access Control List resource definition.")
 @Http.Private.includeInapplicableMetadataInPayload(false)
```
ArcturusZhang: we should remove the private decorator after we do the base type change here.

specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/resources/NetworkFabric.tsp
```diff
@@ -84,7 +153,7 @@ interface NetworkFabrics {
   provision is ArmResourceActionAsync<
     NetworkFabric,
     void,
-    Foundations.OperationStatusResult
+    CommonPostActionResponseForDeviceUpdate
```
ArcturusZhang: we should not change this.

specification/managednetworkfabric/ManagedNetworkFabric.ResourceManager.Management/resources/AccessControlList.tsp
```diff
@@ -2,29 +2,25 @@ import "../models/common.tsp";
 import "../models/AccessControlList.tsp";
```
nnellikunnu, line 2: Can we raise the same PullRequest to the private repo RPSaaSMaster so that both the public and private repos are in sync?

ArcturusZhang: I will do it later

live1206 (approved):

PR #43377, opened 2026-05-20
Customize Storage policy names for C# singleton detection

ArcturusZhang:
Adds C#-specific alternate policy name types for Storage management policies so the .NET management generator can identify the policy resources as fixed-name singleton resources while preserving the public extensible-enum model shape via SDK customization.\n\nThe .NET emitter API version remains pinned to 2025-06-01 for this SDK generation.


specification/storage/Storage.Management/client.tsp
```diff
@@ -457,11 +484,18 @@ namespace Azure.ResourceManager.Models {
...
+    | Azure.ClientGenerator.Core.Usage.input
+    | Azure.ClientGenerator.Core.Usage.input,
+  "csharp"
+);
 @@usage(
   FileShare,
 
     | Azure.ClientGenerator.Core.Usage.input
-    | Azure.ClientGenerator.Core.Usage.output,
+    | Azure.ClientGenerator.Core.Usage.input,
```
live1206: this should be removed

specification/storage/Storage.Management/client.tsp
```diff
@@ -457,11 +484,18 @@ namespace Azure.ResourceManager.Models {
     | Azure.ClientGenerator.Core.Usage.output,
   "csharp"
 );
+@@usage(
+  DeletedAccount,
+
+    | Azure.ClientGenerator.Core.Usage.input
+    | Azure.ClientGenerator.Core.Usage.input,
```
live1206: same here

specification/storage/Storage.Management/client.tsp
ArcturusZhang: I need to change it back.

specification/storage/Storage.Management/client.tsp
```diff
@@ -1274,13 +1308,7 @@ namespace Azure.ResourceManager.Models {
...
 @@alternateType(ListContainerItem, BlobContainer, "csharp");
 
-@@usage(
-  StorageAccount,
-
-    | Azure.ClientGenerator.Core.Usage.input
-    | Azure.ClientGenerator.Core.Usage.output,
-  "java"
-);
+@@usage(StorageAccount, Azure.ClientGenerator.Core.Usage.input, "java");
```
ArcturusZhang: and this

live1206 (approved):

PR #43464, opened 2026-05-22
Allow ARM Auto-Signoff even when NotReadyForARMReview

tejaswiMinnu:
This issue has been ongoing, the ARM Auto signoff flow does not get triggered for incremental typespec changes. 
Example PR - https://github.com/Azure/azure-rest-api-specs-pr/pull/27484

previous PR that dint work, so modifying the logic - https://github.com/Azure/azure-rest-api-specs/pull/42239


.github/workflows/arm-auto-signoff-status.yaml
```diff
@@ -11,11 +11,14 @@ on:
...
-    workflows:
-      [
+    workflows: [
         "ARM Auto SignOff - Analyze Code",
         "Swagger Avocado - Set Status",
         "Swagger LintDiff - Set Status",
+        # Trigger after Summarize Checks to re-evaluate when NotReadyForARMReview is removed
+        # (label changes by workflows don't trigger pull_request_target events).
+        # See: https://github.com/Azure/azure-rest-api-specs/pull/42239
+        "Summarize Checks",
```
mikeharder: this creates a loop in the checks, which github will break after some length of chain (maybe 3 steps), but we shouldn't rely on it.  you'll need to find another way.

"arm auto signoff status" -> "update labels" -> "summarize checks" -> "arm auto signoff status" -> ...

mikeharder (changes requested):

.github/workflows/arm-auto-signoff-status.yaml
tejaswiMinnu: Thanks, I will remove this check, to keep it simple I will remove NotReadyForARMReview label from .js file

mikeharder (approved):

.github/workflows/arm-auto-signoff-status.yaml
```diff
@@ -49,7 +49,6 @@ jobs:
         github.event.label.name == 'ARMAutoSignedOff-Trivial' ||
         github.event.label.name == 'ARMReview' ||
         github.event.label.name == 'ARMSignedOff' ||
-        github.event.label.name == 'NotReadyForARMReview' ||
```
AkhilaIlla, line 52: I dont think I follow how this change fixes the issue which is- 

Why did WaitForARMFeedback addition didnt trigger the auto-sign off workflow and evaluate the changes?

That was because it was added by the bot.

So, we need to trigger the auto-signOff flow on breaking-vhange-approved.. label. I dont see this change, unless I am missing something.

tejaswiMinnu: The fix removes the entire label filter from the workflow trigger - it now triggers on ANY labeled/unlabeled event, not just specific labels. So when a human adds BreakingChange-Approved, the workflow WILL trigger. Additionally, added summarize checks logic and removed the NotReadyForARMReview check from the JS logic so it doesn't block auto-signoff evaluation even if that label is present from an earlier bot action.

Testing: I created a test PR on my fork with incremental TypeSpec changes. The "ARM Auto-Signoff - Set Status" workflow triggered successfully after the Analyze workflow completed. This confirms the workflow now runs and evaluates incremental TypeSpec changes for auto-signoff.
https://github.com/tejaswiMinnu/azure-rest-api-specs/pull/73

AkhilaIlla: "ARM Auto-Signoff - Set Status" will trigger after the ""ARM Auto-Signoff-Analyze Code" workflow finishes. 

> it now triggers on ANY labeled/unlabeled event, not just specific labels.
I still see the list of labels that can trigger the "ARM Auto-Signoff - Set Status" flow?

PR #43487, opened 2026-05-25
[JS] Add clientname for ManagedServiceIdentity

v-jiaodi:
## Summary

Remove the duplicate locally-defined `ManagedServiceIdentity` model from the BillingBenefits spec and reuse the shared definition from `common-types/resource-management/v6/managedidentity.json` instead.

## Motivation

`ManagedServiceIdentity` (and its companion `ManagedServiceIdentityType` enum) were redefined inside this spec, duplicating the standard ARM common-types model. Duplicating the model:
- Creates drift risk between this spec and the ARM standard.
- Produces inconsistent generated SDK types across services.
- Required per-language `@clientName` renames in `client.tsp` to disambiguate from the common-types version.

## Changes

- **`BillingBenefits.Management/Discount.tsp`**: Reference `CommonTypes.ManagedServiceIdentity` instead of the local model on `Discount.identity`.
- **`BillingBenefits.Management/models.tsp`**: Delete the local `ManagedServiceIdentity` model and `ManagedServiceIdentityType` union.
- **`BillingBenefits.Management/client.tsp`**: Remove the `@clientName` overrides (go, java, python, csharp) that were renaming the local `ManagedServiceIdentity` / `ManagedServiceIdentityType` — no longer needed since the common-types model is used.
- **`preview/2025-12-01-preview/billingbenefits.json`**: Regenerated — `identity` now `$ref`s the common-types `ManagedServiceIdentity`; local `ManagedServiceIdentity` and `ManagedServiceIdentityType` definitions removed.

### DIFF
--- a/specification/billingbenefits/BillingBenefits.Management/Discount.tsp
+++ b/specification/billingbenefits/BillingBenefits.Management/Discount.tsp
@@ -47,7 +47,7 @@ model Discount is Azure.ResourceManager.TrackedResource<DiscountProperties> {
    * Managed service identity (system assigned and/or user assigned identities)
    */
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
-  identity?: CommonTypes.ManagedServiceIdentity;
+  identity?: ManagedServiceIdentity;
 
   /**
    * The resource model definition representing SKU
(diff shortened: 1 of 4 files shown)


JialinHuang803 (approved):

JialinHuang803 (changes requested):
As discussed offline, let's update client.tsp only.

JialinHuang803 (approved):

PR #43504, opened 2026-05-26
fix JS build-sample issues for migrate

kazrael2119:
fix JS build-sample issues for migrate


qiaozha (approved):

qiaozha:
Add approved labels as this change is for JS SDK automation ci failure fix only.

PR #43530, opened 2026-05-27
Make toolbox tools optional

viswabalaji:
## Summary
- Make toolbox version `tools` optional on create and response models.
- Leave existing toolbox `skills` support unchanged.
- Regenerate Foundry OpenAPI outputs.

## Validation
- `npx tsp format src/toolboxes/models.tsp`
- `npx tsp compile .`
- `git diff --check`


specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -69,7 +69,7 @@ model ToolboxVersionObject {
   created_at: utcDateTime;
 
   @doc("The list of tools contained in this toolbox version.")
```
glecaros, line 71: probably a good idea to add to the doc string that one of this or skills are required.

PR #43557, opened 2026-05-27
Skip sdk generation for rust in case of OpenAPI specs

raych1:
This pull request updates the SDK generation workflow to correctly handle cases where Rust SDKs are requested for OpenAPI-only specs (i.e., those without a TypeSpec config). It introduces a new "skipped" state to indicate that generation should not proceed in these cases and ensures that the runner logs and reports these skips gracefully across single, PR, and batch generation commands.

Key changes:

* Added a new `GenerationTool` value `"skipped"` to represent cases where SDK generation is not applicable (e.g., Rust with only an OpenAPI spec and no TypeSpec config). The selection logic in `selectGenerationTool` now returns `"skipped"` in these cases.

* Updated `generateSdkForSingleSpec`, `generateSdkForSpecPr`, and `generateSdkForBatchSpecs` to check for the `"skipped"` tool. When detected, the runner logs a warning, marks the execution as succeeded with no packages, and skips generation instead of attempting to run a tool.


mikeharder (approved):

raych1:
@qiaozha FYI - this change is merged, and the failure should be gone in Rust SDK generation for OpenAPI specs.

PR #43560, opened 2026-05-27
feat: adds the ability to control tags

baywet:
demos how we can control tags ordering/parenting/metadata for the docs workstream

In the generated descriptions, what's important are the tags having parents and being in a specific order. The noise from schemas being changed comes from using different versions of packages locally to enable this new experience.


specification/ai-foundry/data-plane/Foundry/tag-definitions.tsp
```diff
@@ -0,0 +1,64 @@
...
+    summary: "Data generation jobs",
+  },
+  #{
+    name: "AgentOptimizationJobs",
+    parent: "Platform APIs",
+    summary: "Agent optimization jobs",
+  },
+  #{
+    name: "EvaluatorGenerationJobs",
+    parent: "Platform APIs",
```
dargilco, line 71: I think this should be `parent: "Evaluations"`

specification/ai-foundry/data-plane/Foundry/tag-definitions.tsp
```diff
@@ -0,0 +1,64 @@
...
+  #{ name: "Indexes", parent: "Platform APIs" },
+  #{ name: "Connections", parent: "Platform APIs" },
+  #{ name: "Memory stores", parent: "Platform APIs" },
+  #{ name: "Scheduler", parent: "Platform APIs" },
+  #{ name: "Fine-tuning", parent: "Platform APIs" },
+  #{
+    name: "EvaluatorGenerationJobs",
+    parent: "Platform APIs",
+    summary: "Evaluator generation jobs",
+  },
```
dargilco: This is already mentioned below

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/tag-definitions.tsp
baywet: fixed

specification/ai-foundry/data-plane/Foundry/tag-definitions.tsp
baywet: fixed

PR #43562, opened 2026-05-27
Use better names other than 'body', since it gets emitted in SDKs

dargilco:
There is no change in OpenAPI3 files, since this is just a TypeSpec label. But it does get used as the input argument name in emitted SDK methods. We want more descriptive names than `body`

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-optimization/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-optimization/routes.tsp
@@ -38,7 +38,7 @@ interface AgentOptimizationJobs {
       ...WithOperationId;
 
       @doc("The optimization job inputs.")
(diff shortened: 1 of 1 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents-optimization/routes.tsp
```diff
@@ -38,7 +38,7 @@ interface AgentOptimizationJobs {
       ...WithOperationId;
 
       @doc("The optimization job inputs.")
-      @body body: OptimizationJobInputs;
+      @body inputs: OptimizationJobInputs;
```
dargilco, line 41: I chose `inputs` to match what I see in the TypeSpec model `OptimizationJob`, where we have this property:

 @doc("Caller-supplied inputs.")
  inputs?: OptimizationJobInputs;

PR #43618, opened 2026-05-29
Fix deserialization of ModelPendingUploadResponse

nick863:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/models/models.tsp
```diff
@@ -121,13 +121,21 @@ model ModelVersion {
...
+  Lifecycle.Create,
+  Lifecycle.Update,
+  Lifecycle.Read
+);
 
 @doc("System metadata for a resource")
 model SystemDataV3 {
   @doc("Timestamp of resource creation")
-  createdAt?: FoundryTimestamp;
+  createdAt?: string;
```
trrwilson: Let's check with the feature team (@vechillal being the original author) on the expected behavior before retroactively aligning the spec. Divergent representations may lead to difficult-to-predict inconsistencies and reuse issues across areas, so if there isn't a well-known and communicated reason for this not aligning, it'd be a better resolution for us to snap the actual service behavior to the service-pattern-common, int32-encoded `utcDateTime` derived scalar previously defined.

nick863: We have removed the class based on offline discussion.

dargilco (approved):

PR #43626, opened 2026-05-31
feat(ai-foundry/eval-rules): add samplingRate to ContinuousEvaluationRuleAction

aprilk-ms:
# Summary

Adds an optional `samplingRate` field to `ContinuousEvaluationRuleAction` to mirror server-side support that already exists in the AI Foundry evaluation service.

When set, each matching event has a (`samplingRate` / 100) chance of triggering an evaluation. When omitted, every event is evaluated (equivalent to `samplingRate: 100`). To disable the rule entirely, callers should set `enabled: false` rather than `samplingRate: 0` (which is rejected by the service).

# Change

In `specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp`:

``typespec
@doc("Sampling rate as a percentage, greater than 0 and at most 100. ...")
@minValueExclusive(0)
@maxValue(100)
samplingRate?: float64;
``

# Server-side validation already in place

The Foundry RAI evaluation service already accepts and validates `samplingRate` with the same constraints expressed here:

- `rate <= 0 || rate > 100` -> `400 BadRequest`
- `null` (omitted) -> treated as `100` (full sampling)

This change makes the wire contract reflect that behavior so generated SDKs (Python, .NET, Java, JS) can expose the field to customers and so OpenAPI consumers can rely on it.

# Generated artifacts

Regenerated both openapi3 outputs via `npx tsp compile .` from `specification/ai-foundry/data-plane/Foundry`:

- `openapi3/v1/microsoft-foundry-openapi3.{json,yaml}`
- `openapi3/virtual-public-preview/microsoft-foundry-openapi3.{json,yaml}`

# Branch target

Targeting `feature/foundry-release` (the working branch for the v1 Foundry surface), matching prior PRs to this area.

# Related

This closes a known TypeSpec gap surfaced by the RAISvc API contract test refresh (Vienna PR #2128878). After this merges, that PR's last documented skip — `ContinuousEvaluationRuleAction_MatchesContinuousEvaluationRuleActionSchema` — can be removed and the contract tests will validate the field end-to-end.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
@@ -42,10 +42,10 @@ model ContinuousEvaluationRuleAction extends EvaluationRuleAction {
   @doc("Maximum number of evaluation runs allowed per hour.")
   maxHourlyRuns?: int32;
 
-  @doc("Percentage (0-100] chance that a matching event triggers an evaluation. Omit to evaluate every event.")
+  @doc("Percentage (0-100] chance that a matching event triggers an evaluation. When omitted, the service-default is to evaluate every event, which is equivalent to setting a sampling rate of 100.")
   @minValueExclusive(0)
   @maxValue(100)
-  samplingRate?: float64;
+  samplingRate?: float64 = 100;
 }
 
 @doc("Evaluation rule action for human evaluation.")
--- a/specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
@@ -41,6 +41,11 @@ model ContinuousEvaluationRuleAction extends EvaluationRuleAction {
 
   @doc("Maximum number of evaluation runs allowed per hour.")
   maxHourlyRuns?: int32;
+
+  @doc("Percentage (0-100] chance that a matching event triggers an evaluation. Omit to evaluate every event.")
+  @minValueExclusive(0)
+  @maxValue(100)
+  samplingRate?: float64;
 }
 
 @doc("Evaluation rule action for human evaluation.")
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
```diff
@@ -41,6 +41,11 @@ model ContinuousEvaluationRuleAction extends EvaluationRuleAction {
 
   @doc("Maximum number of evaluation runs allowed per hour.")
   maxHourlyRuns?: int32;
+
+  @doc("Percentage (0-100] chance that a matching event triggers an evaluation. Omit to evaluate every event.")
+  @minValueExclusive(0)
+  @maxValue(100)
+  samplingRate?: float64;
```
dargilco: "Omit to evaluate every event." -- does that mean that the service-side default is 100 if this value is not provided by the client? If so, please update to:
samplingRate?: float64 = 100;
And update the docs string to something like
@doc("Percentage (0-100] chance that a matching event triggers an evaluation. When omitting, the service-default is to evaluate every event, which is equivalent to setting a sampling rate of 100.")

johanste: ...and do we *really* need `float64` here? I would imagine that about `float32`/6 decimal places is enough precision, no?

aprilk-ms: @johanste yes 6 decimal is enough, but service-side we already have double? Is it a bad idea to not match?

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/evaluation-rules/models.tsp
johanste: The main issues you may run into is accidental truncation on round-trips (read/update). But that is probably not an issue for this specific API?

PR #43661, opened 2026-06-01
fix: operations get concatenated descriptions and summaries

baywet:


specification/ai-foundry/data-plane/Foundry/src/skills/routes.tsp
```diff
@@ -70,6 +70,11 @@ interface Skills {
   @route("/skills/{name}/versions")
   @sharedRoute
   @extension("x-ms-foundry-meta", SkillsRequiredPreviews)
+  @extension("x-ms-summary-override", "Create a new version of a skill")
+  @extension(
+    "x-ms-description-override",
+    "Creates a new version of a skill. If the skill does not exist, it will be created."
+  )
```
dargilco, line 82: Why is this not a regular doc-string on the operation?  None of the operations here have doc strings...

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -77,6 +82,11 @@ interface Agents {
   @post
   @route("/agents/{agent_name}")
   @sharedRoute
+  @extension("x-ms-summary-override", "Update an agent")
+  @extension(
+    "x-ms-description-override",
+    "Updates an existing agent by adding a new version if there are any changes to the agent definition. Without changes, returns the existing agent version."
+  )
```
dargilco, line 93: Why does this description not match the above doc string?

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
```diff
@@ -189,6 +199,8 @@ interface Agents {
   @route("/agents/{agent_name}/versions")
   @sharedRoute
   @extension("x-ms-foundry-meta", AllConditionalAgentDefinitionPreviews)
+  @extension("x-ms-summary-override", "Create an agent version")
+  @extension("x-ms-description-override", "Creates a new agent version.")
```
dargilco: Same ("creates" vs "create")

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
baywet: it's intentional, those are the conventions defined by @trrwilson in #43558

specification/ai-foundry/data-plane/Foundry/src/agents/routes.tsp
baywet: fixed to aligned them. thanks for the callout

specification/ai-foundry/data-plane/Foundry/src/skills/routes.tsp
baywet: it does have a doc string comment? 
As to why do we need an extension to materialize an override, this is to workaround TypeSpec's behaviour on shared routes which concatenates the strings of all the operations sharing the route and method.

I hope I'm answering the question? Let me know if I missed something.

dargilco: Oh sorry, I see we use regular comments and not `@doc` decorators. Do you mind updating the comment to match your new text? Your text has the added sentence at the end which is very important ("If the skill does not exist, ..."). Check other locations as well. Thank you for your patience....

dargilco (changes requested):

specification/ai-foundry/data-plane/Foundry/src/skills/routes.tsp
baywet: sorry, I'm not exactly clear on what change you'd like me to make. I've pushed something, but really unsure. It'd be great if you could send a suggestion on one to illustrate if I still missed the mark.

dargilco (approved):

PR #43686, opened 2026-06-02
Java beta subclients refactor

jpalvarezl:
- Migrating Java SDKs entry points to match other SDKs
- Adding the necessary changes for code generation to allow for the `beta()` and `allowPreview(bool)` methods in the SDKs client builder classes.


specification/ai-foundry/data-plane/Foundry/src/sdk-common/beta-ops-relocation-agents.tsp
```diff
@@ -0,0 +1,91 @@
+// This file moves all beta operations (preview routes) from the Azure.AI.Projects namespace to
```
dargilco, line 1: "sdk-common" would be a better name for this folder since it's only for sdk clients, and it will show up next to the other "sdk-" folders when listing the folders.

dargilco (changes requested):

dargilco (approved):

PR #43700, opened 2026-06-02
Update storage blob spec to v2 folder structure

catalinaperalta:


vincenttran-msft (approved):
LGTM, thanks! Slightly disruptive but I think this is good to just pull the bandaid off now!

PR #43757, opened 2026-06-04
[StorageCache] Add RebalanceJob sub-resource for API version 2026-08-01

brpanask:
## Reviewer Guidance (please read first)

**Status: 32/33 required checks green.** This PR is ready for ARM review. Three items need a reviewer/maintainer with write access:

1. **`Approved-LintDiff`** — Swagger LintDiff is the only red check. The 2 "new" errors are `ProvisioningStateMustBeReadOnly` **false positives** from Spectral `$ref` sibling resolution (OpenAPI 2.0 drops the sibling `readOnly: true` when inlining the `$ref`). `provisioningState` **is** correctly `readOnly: true`; the same rule fires on every existing resource (AmlFilesystem, ExpansionJob, ImportJob, Cache, StorageTarget, …) and across other TypeSpec services (apicenter, advisor, standbypool). Only the 2 RebalanceJob entries are classified "new" because the resource path is new. Full before/after breakdown in the comment dated 2026-07-01. Requesting `Approved-LintDiff`.
2. **Remove stale labels** `BreakingChange-JavaScript-Sdk` and `BreakingChange-Python-Sdk` — left over from an earlier common-types-v6 iteration. The erroneous `x-ms-mutability` on `location` was removed and **BreakingChange now passes**; these labels no longer apply.
3. **`ARMSignedOff`** — PR is in `WaitForARMFeedback`; needs an ARM reviewer sign-off.

`PublishToCustomers` will be added by us only when we're ready to release to customers (per the Next-Steps-to-Merge bot comment).

---

## API Info

**Service:** StorageCache (Microsoft.StorageCache)
**API Version:** 2026-08-01
**Type:** Stable

## Changes

Adds the **RebalanceJob** sub-resource to AMLFS and new expansion-triggered rebalance fields. New API version building on 2026-01-01.

### New Resource: RebalanceJob
- GET, PATCH, DELETE, LIST operations for /amlFilesystems/{name}/rebalanceJobs/{jobName}
- States: InProgress, Cancelling, Canceled, Completed, Failed, Deleting, RollingBack

### New Fields
- `runRebalanceJob` opt-in property on the expansion (when set, the expansion creates a RebalanceJob after completing); `rebalanceJobName` read-only back-reference
- Status fields: balancePercent, filesMigrated, dirsMigrated, bytesMoved, throughputMbps, filesMovedPerSecond, totalErrors, totalSkipped, estimatedRemainingSeconds, statusCode, statusMessage, startTimeUTC, completionTimeUTC

### SDK Compatibility
- back-compatible.tsp: clientName/clientLocation decorators for RebalanceJobs
- flattenProperty for RebalanceJob.properties, RebalanceJobProperties.status, RebalanceJobUpdate.properties

### RebalanceJob has no PUT/create operation (intentional)
A RebalanceJob is created as a **side-effect of an ExpansionJob** (when the rebalance opt-in on the expansion is set), not by a customer-initiated PUT. It therefore exposes **GET, LIST, PATCH, and DELETE only**:
- **PATCH** is used to request cancellation of an active job via `adminStatus` (`Active` -> `Cancel`).
- **DELETE** schedules the job for cleanup (cancelling an active job first).
- Customer-initiated **create (PUT) is intentionally deferred to a later API version**; in this release rebalance jobs originate only from expansion.

`provisioningState` (read-only) is present because the resource supports asynchronous PATCH/DELETE per RPC-Async-V1-02. This is a read-and-manage resource rather than a create-via-PUT resource, so the absence of a PUT is by design, not an asymmetric-CRUD omission.

### RebalanceJob is a ProxyResource (no tags/location)
RebalanceJob is modeled as a `ProxyResource` (proxy in the ARM manifest), so it has no `tags`/`location` envelope.

## Checklist
- [x] TypeSpec compiles
- [x] OAV validate-spec passes
- [x] OAV validate-example passes
- [x] Prettier passes
- [x] SpellCheck passes
- [x] BreakingChange passes
- [x] Examples provided for all operations
- [x] SDK compatibility decorators added

## Breaking Changes
None - all additive.


brpanask:
## LintDiff Failures - All Inherited, Not From RebalanceJob

The Swagger LintDiff check shows ~1000 violations, but **none are from the new RebalanceJob resource**. All violations are pre-existing patterns from existing resources (AmlFilesystem, ImportJob, ExportJob, Cache, StorageTarget, etc.) that appear as \"new\" because the \stable/2026-08-01/openapi.json\ file is entirely new to this repo.

### Violation breakdown (all from existing resources):
- **DescriptionMustNotBeNodeName** (318) — Existing enum values with matching descriptions (e.g., \\"Active\": \"Active\"\)
- **LatestVersionOfCommonTypesMustBeUsed** (409) — Using common-types v3 (upgrading to v6 causes cascading breaking changes on the existing 2026-01-01 API version)
- **ProvisioningStateMustBeReadOnly** (56) — Existing provisioning state patterns
- **PostResponseCodes / LroErrorContent / PatchBodyParametersSchema** — Existing operation patterns from ARM conversion
- **EnumInsteadOfBoolean** (25) — Existing boolean properties that cannot be converted without breaking changes

### What this PR actually adds:
- New \RebalanceJob\ ProxyResource with GET, PATCH, DELETE, LIST operations
- New \Rebalance\ property on \AmlFilesystemExpansion\
- Proper descriptions, x-ms-identifiers, and x-ms-mutability annotations

### Request:
Requesting \Approved-LintDiff\ label since all violations are inherited from the existing TypeSpec conversion (same patterns exist in \stable/2026-01-01/openapi.json\ on main).

### Stale labels:
The \BreakingChange-JavaScript-Sdk\ and \BreakingChange-Python-Sdk\ labels are stale (from a previous iteration that used common-types v6). Breaking change checks now pass — please remove these labels.

brpanask:
## LintDiff "New Violations" — False Positive from `$ref` Sibling Resolution

The 2 remaining "new" `ProvisioningStateMustBeReadOnly` errors (lines 2443, 2509 in `stable/2026-08-01/openapi.json`) are **false positives** caused by a known Spectral resolution behavior.

### Root Cause

Our `provisioningState` property IS correctly marked `readOnly: true`:
```json
"provisioningState": {
  "$ref": "#/definitions/RebalanceJobPropertiesProvisioningState",
  "description": "ARM provisioning state...",
  "readOnly": true
}
```

The lint rule `ProvisioningStateMustBeReadOnly` runs with `resolved: true` in Spectral. Per OpenAPI 2.0 spec, sibling properties alongside `$ref` are invalid and get dropped during resolution. So when Spectral inlines the `$ref`, it loses the sibling `readOnly: true` — causing the false positive.

### This fires on ALL resources in our spec

The same violation fires on every resource (AmlFilesystem, ExpansionJob, ImportJob, AutoExportJob, AutoImportJob, Cache, StorageTarget) — 29 violations total in `2026-08-01`. The lint-diff tool classifies them as "existing" because those paths also exist in `2026-01-01`. Only the 2 RebalanceJob violations are "new" because it's a brand new resource path.

### This is the standard TypeSpec pattern

All TypeSpec-generated ARM resources use `$ref` + `readOnly: true` for provisioningState (verified in apicenter, advisor, standbypool). This is how the `@visibility(Lifecycle.Read)` decorator generates swagger output.

### Request

Requesting `Approved-LintDiff` label — the 2 "new" violations are false positives from the same pattern used by all existing resources in this spec and across other TypeSpec services.

psah434:
if the lintdiff checks are failing for existing operations/RT's then they must have been failing for previous versions. Can you provide the pr and the exception that would have been provided before?

psah434:
also if there are violations this is a good opportunity to fix them.

brpanask:
Thanks @psah434. Before the specifics, some context: StorageCache was migrated to TypeSpec by the ARM TypeSpec conversion team — that team authored the `2026-01-01` base content (#38978, by @Aman-Jain-14), and this same review group (yourself included) was involved in reviewing and approving it. This PR simply adds a new `RebalanceJob` sub-resource on top of that generated foundation, so the patterns in question aren't new to StorageCache.

On your two points:

**Prior PR + exception:** This is the same pattern approved in **#38978** ("StorageCache base commit for API version 2026-01-01"), which received `Approved-LintDiff` on 2025-12-10. That PR introduced the `provisioningState` `$ref` + `readOnly: true` shape for every StorageCache resource; the identical shape is in shipped `stable/2026-01-01/openapi.json` today. The only "new" `ProvisioningStateMustBeReadOnly` hits on this PR (2) are on the new `RebalanceJob` resource, using that exact same generated shape.

**On fixing:** `provisioningState` *is* marked `readOnly: true`. The rule fires because it runs resolved and OpenAPI 2.0 drops properties that sit as siblings of a `$ref`, so the `readOnly` is lost during resolution. The shape is emitted by `@azure-tools/typespec-azure-resource-manager` (`@visibility(Lifecycle.Read)`); changing it for `RebalanceJob` alone would make it inconsistent with every other resource in this spec and with the shipped GA version. We'd therefore request the same `Approved-LintDiff` exception granted on #38978 rather than hand-editing generated output. Happy to add a suppression to `suppressions.yaml` instead if you prefer that mechanism.

Separately, could the stale `BreakingChange-JavaScript-Sdk` / `BreakingChange-Python-Sdk` labels be removed? Those were from an earlier iteration; the BreakingChange checks now pass.

brpanask:
**Context on the outstanding LintDiff findings**

These LintDiff results are inherited from the ARM/TypeSpec team's own brownfield conversion of StorageCache, not introduced by this feature change.

**TypeSpec conversion PR:** [#42197](https://github.com/Azure/azure-rest-api-specs/pull/42197) — *"[TSP Migration] storagcache TypeSpec migrated from swagger"* (conversion wave `typespec-conversion-w6`), authored by the ARM/TypeSpec migration team and merged 2026-04-10. On that PR the **Swagger LintDiff – Analyze Code** job also **failed**, and it was cleared via the **`Approved-LintDiff`** label + `ARMSignedOff` rather than by modifying the emitted spec. The LintDiff run on this PR is here: https://github.com/Azure/azure-rest-api-specs/actions/runs/28615303410

**On `LatestVersionOfCommonTypesMustBeUsed` (common-types v6):** This spec declares `@armCommonTypesVersion(v3)` once at the project level in `main.tsp`, so it applies to *every* API version the TypeSpec emits — including the already-shipped `stable/2026-01-01` GA version, not just the new `2026-08-01`. When we bumped v3→v6 to clear this warning, the emitter regenerated the common-type `$ref`s (ErrorResponse/ErrorDetail, systemData, Operation/OperationListResult, the resource envelope) across the whole spec, which changed the shape of the shipped `2026-01-01` swagger. That tripped the cross-version breaking-change checks and caused the `BreakingChange-JavaScript-Sdk` / `BreakingChange-Python-Sdk` labels to be applied — i.e., "fixing" one advisory warning on the new version manufactured breaking changes on a GA API. We reverted to v3 to avoid breaking the shipped contract; the SDK breaking-change labels currently on this PR are stale leftovers from that iteration and can be removed. Upgrading common-types is a spec-wide migration that should be done as its own change (regenerating and re-approving all versions together), not as a side effect of this feature PR.

**On `ProvisioningStateMustBeReadOnly` (2 "new" hits on RebalanceJob):** These are false positives from Spectral's `$ref`-sibling resolution. `provisioningState` *is* emitted with `readOnly: true`, but because it's a named-union type, `typespec-autorest` emits a bare `$ref` with sibling `readOnly`/`description`; under OpenAPI 2.0 resolution Spectral drops the `$ref` siblings and loses the `readOnly`. This is the exact shape the conversion (#42197) produced for all StorageCache resources and that shipped in `2026-01-01`; the only "new" hits are on the new `RebalanceJob` resource using that identical generated shape. We'd request the same `Approved-LintDiff` exception granted on the conversion rather than hand-editing generated output (happy to add a `suppressions.yaml` entry instead if you prefer that mechanism).

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2580,6 +2860,7 @@ model ResourceSkusResult {
    */
   @visibility(Lifecycle.Read)
   @pageItems
+  @identifiers(#["name"])
```
psah434: **[NEW] 🟠 Warning** **[[TSP `@added` misalignment](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)] / [uniform-versioning](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/uniform-versioning.md)** — un-gated `models.tsp` edits regenerate the already-released `stable/2026-01-01` spec.

**Classification reasoning:** Introduced in this PR. This `@identifiers(#["name"])` addition and the doc-comment additions (replacing removed `documentation-required` FIXME suppressions) are not gated by `@added(Versions.v2026_08_01)`, so they flow into the previously-released `2026-01-01` generated spec — adding descriptions to 5 definitions (`AutoExportJobUpdateProperties`, `AutoImportJobUpdateProperties`, `ImportJobUpdateProperties`, `MetricAggregationType`, `userAssignedIdentitiesValue`) and `x-ms-identifiers: ["name"]` to `ResourceSkusResult.value` (`stable/2026-01-01/openapi.json` line 7982).

**Issue:** Descriptions are non-breaking, but modifying a GA version's contract should be intentional, and the `x-ms-identifiers` addition changes SDK array handling for that GA version.

**Suggested fix:** Confirm the `2026-01-01` changes are intended. If not, gate these additions to `2026-08-01`; otherwise accept them as documented non-breaking improvements. Ensure the `oad` breaking-change CI passes on the modified `2026-01-01` spec.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2322,6 +2456,152 @@ model ExpansionJobUpdate {
...
+   * Average files moved per second over the most recent reporting interval.
+   */
+  @visibility(Lifecycle.Read)
+  filesMovedPerSecond?: float64;
+
+  /**
+   * Average throughput in mebibytes per second (1024x1024 bytes per second) over the most recent reporting interval.
+   */
+  @visibility(Lifecycle.Read)
+  throughputMbps?: float64;
```
psah434: **[NEW] 🔵 Suggestion** **[[TSP-NUMERIC-TYPE / units clarity](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#64-numeric-properties-must-use-numeric-types-tsp-numeric-type)]** — `throughputMbps?: float64`.

**Classification reasoning:** New property added in this PR.

**Issue:** The name `throughputMbps` conventionally implies megabits per second, but the doc comment says mebibytes per second (1024×1024 bytes/sec). This is ambiguous for SDK consumers.

**Suggested fix:** Consider `throughputMiBps` (or a clearer name). It is a new read-only property, so renaming now is cheap. Low severity since the units are documented.

 #Resolved

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2322,6 +2456,152 @@ model ExpansionJobUpdate {
...
+  /**
+   * Total cumulative benign skips since the rebalance started: files intentionally not migrated (for example, a lost migration lease or a stale layout), as distinct from the hard failures counted in totalErrors. Reported in periodic rebalance status updates. Resets to 0 if the rebalancer node restarts.
+   */
+  @visibility(Lifecycle.Read)
+  totalSkipped?: int64;
+
+  /**
+   * The time (in UTC) the rebalance job started.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Non-standard casing matches existing convention in this project"
```
psah434: **[NEW] 🔵 Suggestion** **[[Suppression justification §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** — `casing-style` suppressions on `startTimeUTC` / `completionTimeUTC` (this line and line 2578).

**Classification reasoning:** New suppressions added in this PR.

**Issue:** The reason `"Non-standard casing matches existing convention in this project"` is a consistency-based justification, which Section 4.1 discourages. The casing itself is defensible given the established project-wide `*TimeUTC` convention.

**Suggested fix:** Reword the reason to state the concrete technical rationale (SDK property-name consistency with the sibling `*TimeUTC` properties across the StorageCache job-resource family).

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/back-compatible.tsp
```diff
@@ -108,6 +108,9 @@ using Microsoft.StorageCache;
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
 @@Legacy.flattenProperty(ExpansionJob.properties);
 
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Required for ProvisioningStateMustBeReadOnly lint rule to traverse properties"
```
psah434: **[NEW] 🔵 Suggestion** **[[`@flattenProperty` on new APIs §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** — `@@Legacy.flattenProperty(RebalanceJob.properties)` with a `no-legacy-usage` suppression.

**Classification reasoning:** New resource (`RebalanceJob`) added in this PR; this legacy flatten + suppression are new.

**Issue:** New resource types should avoid legacy property flattening (SDK-breaking risk). The suppression reason (`"Required for ProvisioningStateMustBeReadOnly lint rule to traverse properties"`) is technically questionable — flattening is not generally required for that rule.

**Suggested fix:** Verify whether flattening is truly required. If kept for consistency with the sibling job resources, document that concrete rationale in the suppression reason. Low severity given family consistency.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
```diff
@@ -0,0 +1,88 @@
...
+    Resource = RebalanceJob,
+    KeyName = "rebalanceJobName",
+    SegmentName = "rebalanceJobs",
+    NamePattern = "^[0-9a-zA-Z][-0-9a-zA-Z_]{0,78}[0-9a-zA-Z]$"
+  >;
+}
+
+@added(Versions.v2026_08_01)
+@armResourceOperations(#{ omitTags: true })
+interface RebalanceJobs {
```
psah434, line 33: **[NEW] 🔵 Suggestion** **[[ARM job-cancellation pattern](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md)]** — `RebalanceJobs` exposes GET/LIST/PATCH/DELETE, no PUT; cancellation is via PATCH `adminStatus: Cancel`.

**Classification reasoning:** New resource interface added in this PR.

**Issue:** A dedicated POST cancel action is the more idiomatic ARM pattern for cancelling a job, versus overloading PATCH with an `adminStatus` field.

**Suggested fix:** Optional / design observation only — PATCH-`adminStatus` matches this project's existing job-resource family (ImportJob, AutoExportJob, etc.), so keeping it for consistency is reasonable. No PUT is correct here since `RebalanceJob` is system-created via the expansion `runRebalanceJob` flag.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Returns a rebalance job.
+   */
+  @tag("rebalanceJobs")
+  get is ArmResourceRead<RebalanceJob, Error = CloudError>;
```
psah434: **[NEW] 🔵 Suggestion** **[[Consistent error responses §2.7](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#27-data-plane-operations)]** — mixed error models within the same interface.

**Classification reasoning:** New resource interface added in this PR.

**Issue:** `get` (this line) and `listByAmlFilesystem` (line 74) use `Error = CloudError`, while `update` (line 51) and `delete` (line 64) use `Error = ErrorResponse`. Mixed error models across operations of the same resource produce inconsistent error schemas in the generated spec and SDKs.

**Suggested fix:** Use a single consistent error model across all operations of the `RebalanceJobs` interface.

 #Resolved

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2322,6 +2456,152 @@ model ExpansionJobUpdate {
...
+   * ARM provisioning state, see https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#provisioningstate-property
+   */
+  @visibility(Lifecycle.Read)
+  provisioningState?: RebalanceJobPropertiesProvisioningState;
+
+  /**
+   * The administrative status of the rebalance job. Possible values: 'Active', 'Cancel'. Passing in a value of 'Cancel' will cancel the current active rebalance job. By default it is set to 'Active'.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  adminStatus?: RebalanceJobAdminStatus = RebalanceJobAdminStatus.Active;
```
psah434: **[NEW] 🔵 Suggestion** **[[Visibility / mutability §2.3](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#23-visibility--mutability-review-enforcement)]** — `adminStatus?: RebalanceJobAdminStatus = RebalanceJobAdminStatus.Active` with `@visibility(Lifecycle.Read, Lifecycle.Create)`.

**Classification reasoning:** New property added in this PR.

**Issue:** `RebalanceJob` has no PUT/create operation, so the `Lifecycle.Create` visibility and the `default` value are unused in request contexts — the only writable path is PATCH via `RebalanceJobUpdateProperties.adminStatus`.

**Suggested fix:** Consider dropping the `Lifecycle.Create` visibility and the default for clarity. Low severity.

psah434 (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iterations, converged) against commit [`bb6da1a`](https://github.com/Azure/azure-rest-api-specs/pull/43757/commits/bb6da1a6f8cfa9195a3671a6597956bcdfbfab4f). See inline comments for findings 1-7. No blocking issues were found; the change is additive and non-breaking against `2026-01-01`.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Fixed (F1). Reverted the incidental `@identifiers(#[\"name\"])` on `ResourceSkusResult.value` so the generated GA `2026-01-01` contract is unchanged (no `x-ms-identifiers` on that array). We intentionally kept the additive description improvements, which are non-breaking and clear the prior `documentation-required` FIXME suppressions. `oad` confirms no breaking change against the published `2026-01-01`.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/back-compatible.tsp
brpanask: Kept the flatten and reworded the suppression reason. Every StorageCache resource (AmlFilesystem, ImportJob, ExpansionJob, Cache, StorageTarget) flattens `properties` via `back-compatible.tsp`; keeping RebalanceJob consistent preserves the swagger shape from the TypeSpec brownfield conversion and keeps the generated SDK model shape consistent across the job-resource family. Not flattening would make RebalanceJob diverge from its siblings.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Fixed (F2). Renamed `throughputMbps` -> `throughputMiBps`. The value is mebibytes/sec (1024x1024 bytes), so the `MiBps` suffix now matches the unit in the doc comment and removes the megabits ambiguity. Updated the property, examples, and regenerated openapi.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
brpanask: By design (F5). RebalanceJob follows the sibling job family: it is system-created via the expansion `runRebalanceJob` flag and controlled through PATCH `adminStatus`, which is the family-consistent admin/cancellation path (ImportJob, AutoExportJob, etc.). No PUT is offered in this first release, and a dedicated POST cancel action would diverge from the established family pattern.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Kept the casing and reworded both suppression reasons per your suggestion. They now state the concrete rationale: preserving the established `*TimeUTC` naming used across the StorageCache job resources (`lastStartedTimeUTC`, `lastCompletionTimeUTC`, etc.) for SDK property-name consistency within the job-resource family.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Fixed per your suggestion (F7). Dropped the `Lifecycle.Create` visibility and the default on `adminStatus` - there is no PUT in this release so those were unused. The field is now read-only in responses, and the only writable path remains PATCH via `RebalanceJobUpdateProperties.adminStatus`. When a future release adds PUT create/modify we will re-introduce the Create visibility then.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
brpanask: Fixed (F6). Standardized the error model across the whole `RebalanceJobs` interface: `get` and `listByAmlFilesystem` now use the ARM common-type `ErrorResponse` (matching `update` and `delete`). All four operations now emit a single consistent error schema.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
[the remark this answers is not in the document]
brpanask: Corrected. The PR description's SDK Compatibility section was inaccurate - it referenced `clientName`/`clientLocation` decorators that don't exist for this interface. I've updated the description. Additionally, per the flatten feedback, the only RebalanceJob customization (`flattenProperty(RebalanceJob.properties)`) has now been removed, so RebalanceJob carries no back-compat decorators - the resource exposes a standard nested `properties` envelope.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/RebalanceJob.tsp
```diff
@@ -0,0 +1,88 @@
...
+  @tag("rebalanceJobs")
+  get is ArmResourceRead<RebalanceJob, Error = ErrorResponse>;
+
+  /**
+   * Update a rebalance job instance.
+   */
+  @tag("rebalanceJobs")
+  @patch(#{ implicitOptionality: false })
+  @Azure.Core.useFinalStateVia("azure-async-operation")
+  update is ArmCustomPatchAsync<
```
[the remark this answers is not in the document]
brpanask: Fixed. Removed the `@useFinalStateVia("azure-async-operation")` override; the emitter now produces `final-state-via: location`. On a 202 the RP returns both a `Location` header (`...ascOperations/{id}?monitor=true`, which serves the finished `RebalanceJob`) and an `Azure-AsyncOperation` header (which returns the `AscOperation` status envelope, not the resource). `location` therefore points at where the final resource actually lives, while `Azure-AsyncOperation` is still returned for status polling. Good catch.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/back-compatible.tsp
```diff
@@ -108,6 +108,9 @@ using Microsoft.StorageCache;
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
 @@Legacy.flattenProperty(ExpansionJob.properties);
 
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Consistent with all other StorageCache resources (AmlFilesystem, ImportJob, ExpansionJob, Cache, StorageTarget), which flatten `properties`; preserves the swagger shape established by the TypeSpec brownfield conversion and keeps the generated SDK model shape consistent across the job-resource family. Not flattening would make RebalanceJob diverge from its siblings."
+@@Legacy.flattenProperty(RebalanceJob.properties);
```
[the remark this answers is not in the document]
brpanask: Fixed. Removed `@@Legacy.flattenProperty(RebalanceJob.properties)` and its suppression. RebalanceJob is a brand-new resource with no back-compat constraint, so we prefer the non-flattened nested `properties` shape rather than carrying a recurring `no-legacy-usage` suppression.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2262,6 +2382,20 @@ model ExpansionJobProperties {
...
+  @added(Versions.v2026_08_01)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  runRebalanceJob?: boolean = false;
+
+  /**
+   * Name of the child rebalance job created by this expansion. Populated after RebalanceJob is created.
+   */
+  @added(Versions.v2026_08_01)
+  @visibility(Lifecycle.Read)
+  rebalanceJobName?: string;
```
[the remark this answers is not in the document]
brpanask: Fixed. Changed the back-reference from `rebalanceJobName` (bare name) to `rebalanceJobId`, typed as `armResourceIdentifier` with `x-ms-arm-id-details` (allowed type `Microsoft.StorageCache/amlFilesystems/rebalanceJobs`), so it emits `format: arm-id`. Examples updated to full ARM resource IDs. The RP will return full IDs (tracked in internal PBI 38763728).

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2322,6 +2456,160 @@ model ExpansionJobUpdate {
...
+   * The administrative status of the rebalance job. Possible values: 'Active', 'Cancel'. Passing in a value of 'Cancel' will cancel the current active rebalance job. By default it is set to 'Active'.
+   */
+  @visibility(Lifecycle.Read)
+  adminStatus?: RebalanceJobAdminStatus;
+
+  /**
+   * Name of the parent expansion job that initiated this rebalance. Populated when the rebalance was created as part of an expansion.
+   */
+  @visibility(Lifecycle.Read)
+  expansionJobName?: string;
```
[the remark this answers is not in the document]
brpanask: Fixed. Changed the back-reference from `expansionJobName` (bare name) to `expansionJobId`, typed as `armResourceIdentifier` with `x-ms-arm-id-details` (allowed type `Microsoft.StorageCache/amlFilesystems/expansionJobs`), emitting `format: arm-id`. Examples updated to full ARM resource IDs. RP alignment tracked in internal PBI 38763728.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/stable/2026-01-01/openapi.json
```diff
@@ -5283,6 +5283,7 @@
     },
     "AutoExportJobUpdateProperties": {
       "type": "object",
+      "description": "Properties for updating an auto export job.",
```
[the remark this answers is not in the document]
brpanask: These are documentation-only additions (field `description` metadata) coming from doc comments on models shared between `2026-01-01` and `2026-08-01`. They are required: removing them triggers 5 `@azure-tools/typespec-azure-core/documentation-required` warnings on the new version (verified by removing and recompiling). Because the models are shared, the descriptions unavoidably appear in both emitted specs, but they are additive, non-breaking metadata with no schema/shape impact. Keeping them.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/stable/2026-01-01/openapi.json
```diff
@@ -5787,6 +5788,7 @@
     },
     "AutoImportJobUpdateProperties": {
       "type": "object",
+      "description": "Properties for updating an auto import job.",
```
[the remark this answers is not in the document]
brpanask: These are documentation-only additions (field `description` metadata) coming from doc comments on models shared between `2026-01-01` and `2026-08-01`. They are required: removing them triggers 5 `@azure-tools/typespec-azure-core/documentation-required` warnings on the new version (verified by removing and recompiling). Because the models are shared, the descriptions unavoidably appear in both emitted specs, but they are additive, non-breaking metadata with no schema/shape impact. Keeping them.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/stable/2026-01-01/openapi.json
```diff
@@ -7178,6 +7180,7 @@
     },
     "ImportJobUpdateProperties": {
       "type": "object",
+      "description": "Properties for updating an import job.",
```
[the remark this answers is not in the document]
brpanask: These are documentation-only additions (field `description` metadata) coming from doc comments on models shared between `2026-01-01` and `2026-08-01`. They are required: removing them triggers 5 `@azure-tools/typespec-azure-core/documentation-required` warnings on the new version (verified by removing and recompiling). Because the models are shared, the descriptions unavoidably appear in both emitted specs, but they are additive, non-breaking metadata with no schema/shape impact. Keeping them.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/stable/2026-01-01/openapi.json
```diff
@@ -7349,6 +7352,7 @@
     },
     "MetricAggregationType": {
       "type": "string",
+      "description": "Metric aggregation type.",
```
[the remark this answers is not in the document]
brpanask: These are documentation-only additions (field `description` metadata) coming from doc comments on models shared between `2026-01-01` and `2026-08-01`. They are required: removing them triggers 5 `@azure-tools/typespec-azure-core/documentation-required` warnings on the new version (verified by removing and recompiling). Because the models are shared, the descriptions unavoidably appear in both emitted specs, but they are additive, non-breaking metadata with no schema/shape impact. Keeping them.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/stable/2026-01-01/openapi.json
```diff
@@ -8332,6 +8336,7 @@
     },
     "userAssignedIdentitiesValue": {
       "type": "object",
+      "description": "User-assigned identity properties.",
```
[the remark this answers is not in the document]
brpanask: These are documentation-only additions (field `description` metadata) coming from doc comments on models shared between `2026-01-01` and `2026-08-01`. They are required: removing them triggers 5 `@azure-tools/typespec-azure-core/documentation-required` warnings on the new version (verified by removing and recompiling). Because the models are shared, the descriptions unavoidably appear in both emitted specs, but they are additive, non-breaking metadata with no schema/shape impact. Keeping them.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/expansionJobs_Get.json
```diff
@@ -0,0 +1,37 @@
...
+    "expansionJobName": "expansionjob1",
+    "resourceGroupName": "scgroup",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "name": "expansionjob1",
+        "type": "Microsoft.StorageCache/amlFilesystem/expansionJob",
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/scgroup/providers/Microsoft.StorageCache/amlfilesystems/fs1/expansionJobs/expansionjob1",
```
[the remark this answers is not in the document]
brpanask: Fixed. Corrected the resource `type` to plural segments (`Microsoft.StorageCache/amlFilesystems/expansionJobs`) and the `id` casing (`amlfilesystems` -> `amlFilesystems`) to match the real resource path.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/importJobs_Get.json
```diff
@@ -0,0 +1,48 @@
...
+    "importJobName": "job1",
+    "resourceGroupName": "scgroup",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "name": "job1",
+        "type": "Microsoft.StorageCache/amlFilesystem/importJob",
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/scgroup/providers/Microsoft.StorageCache/amlfilesystems/fs1/importJobs/job1",
```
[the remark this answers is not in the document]
brpanask: Fixed. Corrected the resource `type` to plural segments (`Microsoft.StorageCache/amlFilesystems/importJobs`) and the `id` casing to match the real resource path.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/autoExportJobs_Get.json
```diff
@@ -0,0 +1,43 @@
...
+    "autoExportJobName": "job1",
+    "resourceGroupName": "scgroup",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "name": "job1",
+        "type": "Microsoft.StorageCache/amlFilesystem/autoExportJob",
+        "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/scgroup/providers/Microsoft.StorageCache/amlfilesystems/fs1/autoExportJob/job1",
```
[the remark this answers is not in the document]
brpanask: Fixed. Corrected the resource `type` to plural segments (`Microsoft.StorageCache/amlFilesystems/autoExportJobs`) and the `id` (`autoExportJob` -> `autoExportJobs`, `amlfilesystems` -> `amlFilesystems`).

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/AscOperations_Get.json
```diff
@@ -0,0 +1,21 @@
...
+    "api-version": "2026-08-01",
+    "location": "westus",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "name": "testoperationid",
+        "endTime": "2023-01-01T16:13:13.933Z",
+        "id": "/subscriptions/id/locations/westus/ascOperations/testoperationid",
```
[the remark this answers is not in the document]
brpanask: Fixed. Corrected the operation `id` to a valid form: `/subscriptions/{sub}/providers/Microsoft.StorageCache/locations/{location}/ascOperations/{operationId}` (added `/providers/Microsoft.StorageCache` and a real subscription placeholder).

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/StorageTargets_DnsRefresh.json
```diff
@@ -0,0 +1,20 @@
...
+    "200": {},
+    "202": {
+      "headers": {
+        "Azure-AsyncOperation": "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.StorageCache/locations/eastus/ascOperations/00000000-0000-0000-0000-000000000000?api-version=2026-08-01",
+        "Location": "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.StorageCache/locations/eastus/ascOperations/00000000-0000-0000-0000-000000000000?monitor=true&api-version=2026-08-01"
+      }
+    }
+  },
+  "operationId": "StorageTargets_DnsRefresh",
+  "title": "Caches_DnsRefresh"
```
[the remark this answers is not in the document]
brpanask: Fixed. Corrected the example `title` from `Caches_DnsRefresh` to `StorageTargets_DnsRefresh` to match the operation.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/examples/2026-08-01/AscResourceUsages_Get.json
```diff
@@ -0,0 +1,35 @@
...
+  "parameters": {
+    "api-version": "2026-08-01",
+    "location": "eastus",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000"
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "value": [
+          {
```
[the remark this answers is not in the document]
brpanask: Acknowledged, but skipping in this PR. Adding `id` to the usage items requires a schema change to `AscResourceUsages`/`StorageTargets` - a deprecated resource being removed from the spec (customers can no longer create these; the RP code + swagger are slated for removal). We're not modifying dead-resource schemas here; the example matches the current model.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -670,6 +691,105 @@ union ExpansionJobStatusType {
   RollingBack: "RollingBack",
 }
 
+/**
+ * ARM provisioning state for the rebalance job.
+ */
+@added(Versions.v2026_08_01)
+union RebalanceJobPropertiesProvisioningState {
```
razvanbadea-msft, line 699: >RebalanceJobPropertiesProvisioningState [](http://example.com/codeflow?start=6&length=39)

you should extend  [https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/reference/data-types/#Azure.ResourceManager.ResourceProvisioningState](https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/reference/data-types/#Azure.ResourceManager.ResourceProvisioningState)
 #Resolved

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -2322,6 +2460,164 @@ model ExpansionJobUpdate {
...
+@added(Versions.v2026_08_01)
+model RebalanceJobProperties {
+  /**
+   * ARM provisioning state, see https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#provisioningstate-property
+   */
+  @visibility(Lifecycle.Read)
+  provisioningState?: RebalanceJobPropertiesProvisioningState;
+
+  /**
+   * The administrative status of the rebalance job. Possible values: 'Active', 'Cancel'. Passing in a value of 'Cancel' will cancel the current active rebalance job. By default it is set to 'Active'.
```
razvanbadea-msft: **[NEW] 🔵 Suggestion** **[[Accurate descriptions](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/swagger-authoring-descriptions.md)]** `models.tsp` — line 2475 (generated: `stable/2026-08-01/openapi.json` line 8216-8219, `RebalanceJobProperties.adminStatus`, `readOnly: true`) — the read-only `adminStatus` description describes write behavior.

**Classification reasoning:** Introduced in this PR — `RebalanceJob` and the entire `2026-08-01` version are new (base has zero `RebalanceJob` references; `stable/2026-08-01` returns 404 at the base commit).

**Issue:** `RebalanceJobProperties.adminStatus` is `@visibility(Lifecycle.Read)` (emitted as `readOnly: true`), but its description says "Passing in a value of 'Cancel' will cancel the current active rebalance job. By default it is set to 'Active'." That describes a write action on a response-only property. The writable path is `RebalanceJobUpdateProperties.adminStatus` (models.tsp line 2618), which already carries its own write-oriented doc.

**Suggested fix:** Revise the read-only field's description to describe the observed status, e.g.:
> The current administrative status of the rebalance job. 'Active' indicates the job is running normally; 'Cancel' indicates cancellation has been requested.

 #Resolved

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Done. Changed `RebalanceJobPropertiesProvisioningState` to extend `Azure.ResourceManager.ResourceProvisioningState` (Succeeded/Failed/Canceled) and add the custom lifecycle states Creating/Updating/Deleting. The emitted enum values are unchanged (modelAsString: true), and since RebalanceJob is new in `2026-08-01` there is no back-compat impact. Pushed in d2781695.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
brpanask: Good catch. Reworded the read-only `RebalanceJobProperties.adminStatus` description to describe the observed status rather than write behavior: "The current administrative status of the rebalance job. 'Active' indicates the job is running normally; 'Cancel' indicates cancellation has been requested." The write-oriented wording remains on the writable `RebalanceJobUpdateProperties.adminStatus`. Pushed in d2781695.

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/models.tsp
```diff
@@ -670,6 +691,93 @@ union ExpansionJobStatusType {
   RollingBack: "RollingBack",
 }
 
+/**
+ * ARM provisioning state for the rebalance job.
+ */
+@added(Versions.v2026_08_01)
+union RebalanceJobPropertiesProvisioningState {
```
razvanbadea-msft, line 699: >RebalanceJobPropertiesProvisioningState [](http://example.com/codeflow?start=6&length=39)

i think you also need to add [https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/decorators/#@Azure.Core.lroStatus](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/decorators/#@Azure.Core.lroStatus) to fix the lintdiff error

brpanask: Done — added `@Azure.Core.lroStatus` to `RebalanceJobPropertiesProvisioningState` so the tooling recognizes the terminal LRO states (Succeeded/Failed/Canceled) that come from `Azure.ResourceManager.ResourceProvisioningState`. It's a source-only linter hint (no change to the emitted swagger). Pushed in 6870d992.

brpanask: Following up with the CI result: I added `@Azure.Core.lroStatus` as suggested (pushed in 6870d992), but it does **not** clear the `ProvisioningStateMustBeReadOnly` LintDiff error — and I want to be transparent about why.

`@lroStatus` is a compile-time TypeSpec annotation that marks the enum's terminal states for Azure.Core LRO tooling. It produces **no change to the emitted OpenAPI** (the generated `stable/2026-08-01/openapi.json` is byte-identical before/after this commit). Since LintDiff evaluates the emitted swagger, the result is unchanged.

The rule itself fires for a different reason:
- `provisioningState` **is** `readOnly` in the TypeSpec (`@visibility(Lifecycle.Read)`) and in the emitted swagger it is authored as `{ "\": "#/definitions/RebalanceJobPropertiesProvisioningState", "readOnly": true }`.
- `ProvisioningStateMustBeReadOnly` runs against the **resolved** OpenAPI 2.0 document, where a `readOnly` sibling of a `\` is dropped during `\` resolution. So the resolver no longer sees `readOnly`, and the rule flags it.

This is the **same emitter output and the same false positive that exists on every StorageCache resource in the shipped GA `2026-01-01`** (e.g. `amlFilesystems` GET/PUT/PATCH report the identical `ProvisioningStateMustBeReadOnly` from `2026-01-01`). The RebalanceJob findings (GET + PATCH 200) are the new resource inheriting that exact generated shape.

There's no TypeSpec change that removes it without diverging from the ARM-generated pattern used across the rest of this spec and the approved GA version (base #38978 and TypeSpec conversion #42197 were both cleared via `Approved-LintDiff` rather than by modifying the emitted spec). Given that, we'd request the same `Approved-LintDiff` handling here. Happy to adjust if there's a preferred emitter-level fix.

razvanbadea-msft: I checked and it is a typespec limitation, you need to add a suppression rule in this case in the readme file

brpanask: Done — added a LintDiff suppression for `ProvisioningStateMustBeReadOnly` to `readme.md` (pushed in 40af0fa3), scoped to the new RebalanceJob GET/PATCH operations:

```yaml
directive:
  - suppress: ProvisioningStateMustBeReadOnly
    where:
      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.StorageCache/amlFilesystems/{amlFilesystemName}/rebalanceJobs/{rebalanceJobName}"].get
      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.StorageCache/amlFilesystems/{amlFilesystemName}/rebalanceJobs/{rebalanceJobName}"].patch
    reason: ...
```

The `where` targets only the `rebalanceJobs` path (which exists solely in `2026-08-01`), so there's no impact on the GA `2026-01-01` version. Thanks for confirming the TypeSpec limitation.

brpanask: Update: the suppression is now working. My first attempt used an autorest `directive: suppress` block, which does not clear spectral-named openapi-validator rules like `ProvisioningStateMustBeReadOnly`. I switched to the openapi-validator `suppressions:` block (`code` + `from`) inside the `package-2026-08-01` tag block in `readme.md` (same mechanism used by dynatrace/newrelic/playwrighttesting for this exact TypeSpec `readOnly`-sibling-of-`\` false positive, ref Azure/azure-openapi-validator#637).

As of commit `38d95f1c`, **Swagger LintDiff - Analyze now passes** and the RebalanceJob `ProvisioningStateMustBeReadOnly` findings are cleared. Scoped to the `2026-08-01` tag so GA `2026-01-01` is unaffected. Thanks for the pointer!

Aman-Jain-14 (approved):

specification/storagecache/resource-manager/Microsoft.StorageCache/StorageCache/readme.md
```diff
@@ -26,11 +26,23 @@ These are the global settings for the StorageCache API.
...
 ```
 
+### Tag: package-2026-08-01
+These settings apply only when `--tag=package-2026-08-01` is specified on the command line.
+
+```yaml $(tag) == 'package-2026-08-01'
+input-file:
+  - stable/2026-08-01/openapi.json
+suppressions:
```
razvanbadea-msft, line 41: **[NEW] 🔵 Suggestion** **[[ProvisioningStateMustBeReadOnly / use-read-only-status-schema](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#provisioningstatemustbereadonly----emitter-workaround)]** `readme.md` line 41 (suppression) / `tspconfig.yaml` autorest options

**Classification reasoning:** New — the `ProvisioningStateMustBeReadOnly` suppression and the `RebalanceJob` resource do not exist in the prior `2026-01-01` version; the GA `package-2026-01-01` readme tag has no suppressions block.

**Issue:** `RebalanceJobProperties.provisioningState` is emitted as `{ "$ref": …, "readOnly": true }` — the `readOnly` sibling of `$ref` that OpenAPI 2.0 resolution drops — which is why this PR adds a `ProvisioningStateMustBeReadOnly` suppression. The preferred fix is the autorest emitter option `use-read-only-status-schema: true`, which stamps `readOnly` correctly on the generated status schema and removes the need for the suppression.

**Suggested fix:** Add `use-read-only-status-schema: true` under `options → @azure-tools/typespec-autorest` in `tspconfig.yaml`. Optional / low priority — the suppression is already approved and matches the GA 2026-01-01 pattern, so you may reasonably keep it.

razvanbadea-msft:
Fix errors from Swagger LintDiff or add suppressions for them as i see it is more related to the older version that was previoulsly merged.

Aman-Jain-14 (approved):

Aman-Jain-14 (approved):

Aman-Jain-14 (approved):

PR #43760, opened 2026-06-04
Emit build-failed SDK PR label from spec-gen-sdk-runner

JoshLove-msft:
## Summary

Implements the generated-SDK **build-failure labeling** signal in the generator-agnostic `spec-gen-sdk-runner`, instead of the deprecated `swagger_to_sdk_config` label surface in the `spec-gen-sdk` tool.

This is a redirect of Azure/azure-sdk-tools#15887 based on review feedback from @raych1: the label configuration in `spec-gen-sdk` is deprecated, and new labeling logic should live in `spec-gen-sdk-runner` so it is reused when generation moves from `spec-gen-sdk` to `azsdk-cli`.

## What this does

When the SDK **PR-creation flow** (`generateSdkForSingleSpec`, the SDK release scenario) produces a `warning` result (generation succeeded but the package build failed), the runner emits a `BuildFailedLabel` pipeline variable so the created SDK pull request can be labeled for automated build-failure repair (e.g. by the Generator Agent).

- `BuildFailedLabel` is only emitted in the PR-creation flow — **never** in plain spec-PR CI validation (`generateSdkForSpecPr`), satisfying the opt-in requirement in Azure/azure-sdk-for-net#59651.
- The label **name** is sourced from the centralized `sdkLabels` map in `.github/shared/src/sdk-types.js` (`auto-sdk-build-fix` for `azure-sdk-for-net`; `undefined` for other languages, which are skipped).

## Changes

- `.github/shared/src/sdk-types.js`: add `buildFailed` to `SdkLabelInfo` and each `sdkLabels` entry.
- `eng/tools/spec-gen-sdk-runner/src/command-helpers.ts`: add `getBuildFailedInfo()` (warning => build failed) and `setBuildFailedLabelVariable()` (emits the `BuildFailedLabel` variable from `sdkLabels`).
- `eng/tools/spec-gen-sdk-runner/src/commands.ts`: call `setBuildFailedLabelVariable()` in `generateSdkForSingleSpec` only.
- `eng/tools/spec-gen-sdk-runner/test/command-helpers.test.ts`: unit tests for both helpers.

## Validation

- `spec-gen-sdk-runner`: build, lint, prettier, and tests pass (164 tests).
- `.github/shared`: tests pass (218 tests).

## Follow-ups (not in this PR)

- [ ] Pipeline template wiring to apply `BuildFailedLabel` on the created SDK PR.
- [ ] Create the `auto-sdk-build-fix` label in `azure-sdk-for-net`.
- [ ] Drop the now-superseded `spec-gen-sdk` tool-side changes in Azure/azure-sdk-tools#15887.

## Related

- Azure/azure-sdk-tools#15887 (original, deprecated-surface implementation being superseded)
- Azure/azure-sdk-for-net#59651 (feature requirement: `auto-sdk-build-fix`, PR-creation flow only)

--generated by Copilot


eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
```diff
@@ -286,6 +287,46 @@ export function getBreakingChangeInfo(executionReport: ExecutionReport): boolean
   return false;
 }
 
+/**
+ * Determine whether the SDK generation succeeded but the package build failed.
+ * A `warning` execution result means generation produced an SDK that did not build,
```
raych1: Can you add more comment since this is .NET specific behavior?

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
```diff
@@ -286,6 +287,46 @@ export function getBreakingChangeInfo(executionReport: ExecutionReport): boolean
...
+ * which is the signal used to flag a generated SDK pull request for automated build repair.
+ *
+ * @param executionReport - The spec-gen-sdk execution report.
+ * @returns true when the build failed (execution result is `warning`).
+ */
+export function getBuildFailedInfo(executionReport: ExecutionReport): boolean {
+  return executionReport.executionResult === "warning";
```
raych1, line 332: How about scoping this to .NET only?

JoshLove-msft: This is a general feature that is only being implemented for .NET in this PR. There is already a way to toggle it per language. Are you saying we should also hardcode it to only work for .NET? Can you help me understand why we would want that?

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
JoshLove-msft: I don't think we should restrict this to .NET. The concept generalizes across languages - have some AI assisted workflow to repair build failures in auto PRs. Nothing .NET-specific there IMO.

raych1: What I mean is the 'warning' result which is the contract in .NET script. I don't want it to impact other languages.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
raych1: I agree the feature itself is not .NET-specific. My concern is more about the result mapping. Since we can’t guarantee that every language reports build failures as warning, I’m worried this could behave inconsistently across languages. Would it make sense to scope this PR to .NET for now, or first align across languages?

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
JoshLove-msft: Done — I expanded the doc comment on `getBuildFailedInfo` to make the `warning` contract explicit: it documents that `warning` is the .NET spec-gen-sdk script's signal for "an SDK was generated but it didn't build," that this interpretation is currently .NET-specific, and that it can't impact other languages because the label is only emitted for languages that configure a `buildFailed` entry in the centralized `sdkLabels` map (today only .NET). Commit 331a9140aab.

--generated by Copilot

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
JoshLove-msft: It's already effectively scoped to .NET: `setBuildFailedLabelVariable` only emits the label when `sdkLabels[language].buildFailed` is defined, and that field is `undefined` for every language except `azure-sdk-for-net`. So a `warning` result from any other language is a no-op — no label is set.

I deliberately kept the `warning` detection itself generic rather than hardcoding `=== SdkName.Net`, so another language can opt in later just by establishing its build-failed contract and adding its own `buildFailed` entry to the centralized `sdkLabels` map — no code change needed here. I added a comment making the .NET-specific `warning` contract explicit (commit 331a9140aab). Happy to add an explicit language guard instead if you'd prefer belt-and-suspenders, but it would be redundant with the label-map gating today.

--generated by Copilot

raych1: I understand the label-map gating is in place. The execution result mapping may still need to be changed when adding support for other languages.

raych1 (approved):

PR #43775, opened 2026-06-05
Sync OpenAI operations

glecaros:
Auto-generated by openai-typespec-foundry


baywet (commented):
One issue here is that `common/openai-templates.tsp` seems to be missing?
That leads to compilation errors, etc...
Not sure whether it's a factor of "something was forgotten in the commit" or "something is wrong with the re-export script".

PR #43791, opened 2026-06-05
Update SDK generation skill to use local SDK generation only when exp…

praveenkuttappan:
SDK generation skill for local SDK generation should only be used when user prompt explicitly asks so. This is more important now   once we disable SDK generation using the release planner.


samvaity (approved):
Interesting, did this help. I am curious as we mention the word "locally" quite a bit wondering if we need to elaborate what local means in that case then 🤔

PR #43801, opened 2026-06-05
Fix blobs nested directory implementation

catalinaperalta:


specification/storage/data-plane/BlobStorage/Common/service.tsp
```diff
@@ -0,0 +1,47 @@
+import "@typespec/rest";
+import "@typespec/versioning";
+import "@azure-tools/typespec-azure-core";
+import "./routes.tsp";
```
catalinaperalta, line 4: technically this file can be unimported here and just imported in the parent main.tsp, however since the main spec + the custom sdk configs all use the same server def it seems fine to me to leave this here for now to simplify imports across all sub-dirs

vincenttran-msft (approved):
LGTM!

PR #43738, opened 2026-06-04
[JS] Update operation group name

v-jiaodi:
Update operation group name

PR #43807, opened 2026-06-08
[SQL] Split operation group customizations

live1206:
## Summary

Split SQL changes out of #43738 into a separate PR, per review feedback in https://github.com/Azure/azure-rest-api-specs/pull/43738#discussion_r3363951268.

Changes:
- Scope SQL operation group renames so JavaScript gets `*Operations` groups while non-JavaScript keeps the existing group names.
- Remove the legacy SQL `@azure-tools/typespec-csharp` emitter block because SQL should not opt into the new .NET management emitter path yet.

## Validation

- `npx --no-install tsp format --check specification/sql/resource-manager/Microsoft.Sql/SQL/back-compatible.tsp specification/sql/resource-manager/Microsoft.Sql/SQL/tspconfig.yaml`
- `npx --no-install tsp compile specification/sql/resource-manager/Microsoft.Sql/SQL --warn-as-error`


JialinHuang803 (approved):

JialinHuang803:
Merging this PR as this is to mitigate JS breakings.

PR #43818, opened 2026-06-08
Prepare specs for no-url-suffix lint rule

haiyuazhang:
## Summary

Prepares existing TypeSpec specs for Azure/typespec-azure#4541 (`@azure-tools/typespec-client-generator-core/no-url-suffix`).

- Adds isolated C# `@@clientName(..., "...Uri", "csharp")` customizations in new minimal `client.tsp` files where doing so does not pull in existing client customizations.
- Adds targeted `#suppress` directives where a `clientName` fix is not safe or not expressible, such as OAuth metadata fields, inline/aliased properties, or projects with existing `client.tsp` files whose import would apply unrelated SDK customizations.
- Does not include generated OpenAPI/JSON changes.

## Validation

- `npx tsp compile . --no-emit` for all 154 affected TypeSpec projects
- `npx tsp compile .` for the 25 projects with new minimal `client.tsp` imports
- `git diff --check -- specification`
- Confirmed no generated JSON diffs

Related to Azure/typespec-azure#4541.


specification/apimanagement/resource-manager/Microsoft.ApiManagement/ApiManagement/models.tsp
```diff
@@ -2108,6 +2108,7 @@ model ApiContractProperties extends ApiEntityBaseContract {
    */
   @maxLength(2000)
   @minLength(0)
+  #suppress "@azure-tools/typespec-client-generator-core/no-url-suffix" "Importing this project's existing client.tsp would apply unrelated customizations; keeping the existing C# SDK surface avoids a breaking change."
```
timotheeguerin: we are gong to import those files do not add those suppressions, we cannot merge the rule until all client.tsp are imported from main.tsp.

haiyuazhang: For specs where the C# SDK has already shipped with Url properties (e.g., Azure.Search.Documents.SubdomainUrl), #suppress is necessary — @@clientName would introduce API compat errors (MembersMustExist).  Mark the PR as draft, will revisit it after the import client.tsp from main.tsp work is done.

timotheeguerin: not sure I get the problem here, the SDK should currently be using client.tsp to generate how is that affecting them?

haiyuazhang: I assume enabling the rule shouldn't introduce any new warnings for existing spec violations.  For each violation, we either #suppress it or use @@clientName to fix it.  `Azure.Search.Documents.SubdomainUrl`  is already in a stable api version as a public property. Adding `@@clientName(SubdomainUrl, "subdomainUri", "csharp")`  will fails the .NET API compat check in generated SDK:
```
 error MembersMustExist : Member 'public System.String AIServicesAccountKey.SubdomainUrl.get()'
 does not exist in the implementation but it does exist in the contract.
```
That's why I think` #suppress` is needed for properties already in the SDK's public API

timotheeguerin: ok then I tghink the suppression message doesn't make sense it says that we can'tg import client.tsp or it will cause issue. Yes, in this PR you should only be adding suppressions but have the message be something like `"Name kept for back compatibility"` it shouldn't have anything about a client.tsp

haiyuazhang:
The Swagger Avocado failures are pre-existing `MULTIPLE_API_VERSION` errors in readme.md files not modified by this PR:

- `specification/applicationinsights/.../readme.md` (tag `package-2024-08-01`)
- `specification/compute/.../readme.md` (tag `package-2025-11-01`)
- `specification/monitor/.../readme.md` (tag `package-2025-08`)
- `specification/network/.../readme.md` (tag `package-2025-07-01`)
- `specification/policyinsights/.../readme.md` (tag `package-2024-10`)
- `specification/security/.../readme.md` (tag `package-composite-v3`)

Our PR only modifies `.tsp` and `client.tsp` files in these services, which triggers Avocado to scan their readme — but the errors are unrelated to our changes.

specification/ai/DocumentIntelligence/client.tsp
```diff
@@ -300,3 +300,29 @@ interface DocumentIntelligenceAdministrationClient {
 @@clientName(ClassifyDocumentRequest.base64Source, "bytesSource", "python");
 @@clientName(DocumentFieldSchema.items, "itemsSchema", "python");
 @@clientName(AnalyzeFromStreamRequestParams.input, "body", "python");
+
+@@clientName(
```
timotheeguerin: is that not going to be breaking changes?

specification/machinelearningservices/data-plane/WorkspaceDataplane/client.tsp
```diff
@@ -0,0 +1,10 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+@@clientName(
```
timotheeguerin, line 6: For the services that do not have client.tsp, is it because they are not releasing SDKs? Do we then want to set that up for them or should we think about a separate ruleset that hsould only be enabled if you plan to release SDKs?

specification/sovereign/Sovereign.Management/main.tsp
timotheeguerin, line 1: I don't think this PR should be doing anything else but adding suppresions. This service for example doesn't define client emitter in its tspconfig so its probably not generating SDK we shouldn't have to apply those rules(why the separate ruleset)

specification/oracle/Oracle.Database.Management/models/autonomousdatabase/autonomousDatabase.tsp
```diff
@@ -464,36 +464,50 @@ model AllConnectionStringType {
...
 
+  #suppress "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix" "Name kept for backward compatibility"
+  #suppress "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix" "Name kept for backward compatibility"
   @doc("The URL of the MongoDB API for the Autonomous Database.")
   @minLength(10)
   @maxLength(512)
   mongoDbUrl?: string;
 
+  #suppress "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix" "Name kept for backward compatibility"
+  #suppress "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix" "Name kept for backward compatibility"
```
timotheeguerin, line 497: I see a lot of double suppressions

specification/sovereign/Sovereign.Management/main.tsp
haiyuazhang: all violations are handled by #suppress now and no client.tsp is added by this PR. The errors in TSV check are caused by  override-parameters-mismatch , not related to this PR.

timotheeguerin: i think this just means you haven't pulled in the changes from main in the typespec-azure pr because I believe @tadelesh fixed those issues

timotheeguerin: but at this point as this needs to be `opt-int` until we have the tsv enforcement this pr is not blocking the typespec-azure one from being merged, you'll just need to make sure the final one is green when upgrading the version to what is published

specification/oracle/Oracle.Database.Management/models/autonomousdatabase/autonomousDatabase.tsp
haiyuazhang: fixed

specification/sovereign/Sovereign.Management/main.tsp
haiyuazhang: I rebased the rule PR and  repinned the latest build from the rule PR. TSV check still fails.  It probably because the spec fix is still on typespec-next not yet synced to main yet.

haiyuazhang: all TSV checks all pass now

haiyuazhang:
The required SDK Validation check fails with  `unknown-rule-set`: client-sdk  because every SDK repo's  emitter-package.json  still pins the typespec-azure toolchain below 0.69.2 (where  client-sdk  was added) — do we need to bump  typespec-azure-rulesets / typespec-client-generator-core to 0.69.2 in each SDK repo as a precondition for merging this spec PR?

haiyuazhang:
@timotheeguerin  As yuchao pointed out in his tcgc version bump PR https://github.com/Azure/azure-sdk-for-python/pull/47811, we need to wait 6 days to satisfy the new 7 days law for new npm package.

package.json
```diff
@@ -41,7 +41,10 @@
     "get-uri": "^6.0.5",
     "lodash": "^4.18.1",
     "minimatch@^3.0.0": "^3.1.5",
-    "superagent": "^10.2.3"
+    "superagent": "^10.2.3",
+    "@azure-tools/typespec-azure-rulesets": {
+      "@azure-tools/typespec-azure-resource-manager": "0.69.1"
```
timotheeguerin: What happened here? Why are we overriding this?

haiyuazhang: As shown in the latest commit,  rm 0.69.2 emits the CMK/ Encryption  common-types inline instead of as  $ref s. That causes a swagger diff in devcenter + maps, which in turn fails the LintDiff and BreakingChange checks

timotheeguerin: Hhm but why was that package released and the repo not upgraded? @markcowl 

We should not have overrides here, it also should not be those targeted overrides we need a single version of each typespec package not each libs using their own

timotheeguerin (approved):

timotheeguerin:
@haiyuazhang what is the blocker here, we are releasing next week and will need to upgrade the spec repo and this must be resolved or we'll have to revert it.

haiyuazhang:
> @haiyuazhang what is the blocker here, we are releasing next week and will need to upgrade the spec repo and this must be resolved or we'll have to revert it.
rust's rm and ruleset version have not been bumped to 0.69.2

timotheeguerin:
@haiyuazhang we don't wait for SDK repos to be up to date, if there is breaking change you can wait 24h otherwise it can be merged immediately this is our agreed on workflow.

timotheeguerin (approved):

PR #43846, opened 2026-06-09
Try tweaking pr stale policy

timotheeguerin:
Doesn't seem like it triggered yersteday, swiching to daily(with full time) like typespec repo and giving it unique name

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -2487,31 +2487,6 @@ configuration:
       description: 
     - if:
       - payloadType: Pull_Request
-      - hasLabel:
-          label: no-recent-activity
-      - or:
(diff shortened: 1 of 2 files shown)


.github/policies/resourceManagement.yml
```diff
@@ -2485,31 +2485,6 @@ configuration:
       - addReply:
           reply: Thank you for your contribution ${issueAuthor}! We will review the pull request and get back to you soon.
       description: 
-    - if:
```
timotheeguerin, line 2488: removing those for now (should be in the pr.stale.yml file I think anyway) to make sure this is not causing conflict. If this works I'll make a follow up pr adding them(or some) back

mikeharder (approved):

PR #43885, opened 2026-06-10
eng/tools: Run TypeScript directly via Node's built-in type stripping

timotheeguerin:
## eng/tools: Run TypeScript directly via Node's built-in type stripping

Eliminates the `tsc` compile step for all `eng/tools` packages by leveraging Node.js 22.18+ built-in type stripping. Source files remain `.ts` and are executed directly by Node without a build.

### What changed

- **Removed `postinstall` compile step** — no more `tsc` build; packages are ready to use after `npm ci`
- **Bin entry points are `.js` files that import `.ts` source** — npm reliably resolves `.js` bin targets across all environments (CI Linux, macOS, Windows). Each `cmd/*.js` file simply does:
  ```js
  #!/usr/bin/env node
  import { main } from "../src/index.ts";
  await main();
  ```
- **Source code stays `.ts`** — all business logic, tests, and configs remain TypeScript
- **Removed per-package `dist/` output** — no generated JS artifacts
- **Updated tsconfigs** — `tsc --noEmit` is used for type-checking only; `cmd/` dirs removed from includes since they're plain JS
- **Set `engines.node >= 22.18.0`** — minimum version with unflagged type stripping support
- **Converted enums to `as const` objects** — required by `erasableSyntaxOnly` (Node strips types but doesn't transform enums)
- **Updated imports to use `.ts` extensions** — required by `verbatimModuleSyntax` / `allowImportingTsExtensions`

### Why `.js` bin entry points (not `.ts`)

npm's bin linking has inconsistent behavior with `.ts` targets across environments. On CI (Ubuntu + GitHub Actions), `npm exec --no -- <bin>` would fail with a 404 trying to fetch the bin name from the npm registry, meaning the local bin symlink was never created. Using `.js` as the extension for bin entry points is universally reliable, while Node's type stripping handles the `.ts` imports at runtime.

### Verification

- `npm exec --no -- tsv`, `get-suppressions`, etc. all work on Node 22 and Node 24
- `tsc --noEmit` passes
- Tests pass (pre-existing unrelated failures only)
- ESLint passes


eng/tools/eslint.base.config.js
mikeharder, line 1: convert to eslint.base.config.ts?

mikeharder (approved):

eng/tools/eslint.base.config.js
timotheeguerin: unfortunately eslint won't work with native ts without a flag which is a bit of a pain https://github.com/eslint/eslint/issues/19985

I think anway we should try to switch to oxc-lint or biome as they are much more modern and performant linters

mikeharder (approved):

PR #43894, opened 2026-06-10
Add hybridnetwork specification changes from PR #25860

kpanugan:
### Description

This PR adds the TypeSpec-based specification for **Microsoft.HybridNetwork** stable API version **2025-03-30**, cherry-picked from azure-rest-api-specs-pr PR #25860.

### Scope

- **Provider:** `Microsoft.HybridNetwork`
- **API Version:** `2025-03-30` (Stable)
- **Spec type:** TypeSpec (converted from OpenAPI via `@autorest/openapi-to-typespec`)

### Changes

**TypeSpec definitions** (18 `.tsp` files):
- `Publisher`, `ArtifactStore`, `ArtifactManifest`
- `NetworkFunction`, `NetworkFunctionDefinitionGroup`, `NetworkFunctionDefinitionVersion`
- `NetworkServiceDesignGroup`, `NetworkServiceDesignVersion`
- `ConfigurationGroupSchema`, `ConfigurationGroupValue`
- `Site`, `SiteNetworkService`, `Component`
- `models.tsp`, `routes.tsp`, `back-compatible.tsp`, `client.tsp`, `main.tsp`

**Generated artifacts:**
- OpenAPI (`openapi.json`) for `stable/2025-03-30`
- Example files for all operations
- Updated `readme.md` with new API version configuration
- Added `suppressions.yaml`

### Checklist

- [x] TypeSpec compiles without errors
- [x] OpenAPI spec generated from TypeSpec
- [x] Example files included for all operations
- [x] `readme.md` updated


specification/hybridnetwork/resource-manager/readme.md
```diff
@@ -51,6 +66,40 @@ suppressions:
     reason: sku cannot be patched
 ```
 
+### Tag: package-2024-04-01-preview
+
+These settings apply only when `--tag=package-2024-04-01-preview` is specified on the command line.
+
+```yaml $(tag) == 'package-2024-04-01-preview'
+input-file:
+  - Microsoft.HybridNetwork/preview/2024-04-01-preview/common.json
```
sandipsh: **[NEW] 🔴 Blocking** **[NO_JSON_FILE_FOUND]** specification/hybridnetwork/resource-manager/readme.md - line 75 - The newly added tag `package-2024-04-01-preview` references swagger files that are not present in this PR branch (for example `Microsoft.HybridNetwork/preview/2024-04-01-preview/common.json`), which breaks Swagger LintDiff and cross-version analysis with `ENOENT`.

Classification: This is **new** in this PR because this tag block is newly added in this iteration and does not exist in `main`.

Fix: Either (1) add back all referenced preview swagger files under `Microsoft.HybridNetwork/preview/2024-04-01-preview/`, or (2) remove this tag block from readme if that API version is not part of this package.

specification/hybridnetwork/resource-manager/readme.md
```diff
@@ -118,6 +208,52 @@ input-file:
...
+### Tag: package-2021-06-01-privatepreview
+
+These settings apply only when `--tag=package-2021-06-01-privatepreview` is specified on the command line.
+
+```yaml $(tag) == 'package-2021-06-01-privatepreview'
+input-file:
+  - Microsoft.HybridNetwork/preview/2021-06-01-privatepreview/common.json
+  - Microsoft.HybridNetwork/preview/2021-06-01-privatepreview/networkFunction.json
+  - Microsoft.HybridNetwork/preview/2021-06-01-privatepreview/vendor.json
+  - Microsoft.HybridNetwork/preview/2021-06-01-privatepreview/operation.json
```
sandipsh: **[NEW] 🔴 Blocking** **[NO_JSON_FILE_FOUND]** specification/hybridnetwork/resource-manager/readme.md - line 220 - `package-2021-06-01-privatepreview` references `operation.json`, but that file is missing in the branch, and Swagger Avocado fails with `NO_JSON_FILE_FOUND`.

Classification: This is **new** in this PR because this privatepreview tag section was introduced by this readme update and is not present in `main`.

Fix: Restore `Microsoft.HybridNetwork/preview/2021-06-01-privatepreview/operation.json` (and any other referenced files), or remove this package tag block from readme.

specification/hybridnetwork/resource-manager/readme.md
```diff
@@ -51,6 +66,40 @@ suppressions:
...
+  - Microsoft.HybridNetwork/preview/2024-04-01-preview/site.json
+  - Microsoft.HybridNetwork/preview/2024-04-01-preview/siteNetworkService.json
+  - Microsoft.HybridNetwork/preview/2024-04-01-preview/serviceManagementContainer.json
+suppressions:
+  - code: PatchSkuProperty
+    from: siteNetworkService.json
+    reason: sku cannot be patched
+  - code: AvoidAdditionalProperties
+    from: serviceManagementContainer.json
+    reason: additional propertie is required by design for dictinary type of object
```
sandipsh: **[NEW] 🟡 Warning** **[SpellCheck]** specification/hybridnetwork/resource-manager/readme.md - line 93 - Typo in suppression reason: `dictinary`.

Classification: **new** (introduced in this PR with the newly added 2024-04-01-preview suppression block).

Fix: Replace `dictinary` with `dictionary`.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
```diff
@@ -0,0 +1,47 @@
+parameters:
```
sandipsh, line 1: **[NEW] 🔴 Blocking** **[TSP-Project-Structure]** specification/hybridnetwork/HybridNetwork.Management/tspconfig.yaml - line 1 - TypeSpec Validation fails `FolderStructure` for this project: `must use folder structure v2`.

Classification: **new** (this TypeSpec project is newly added in this PR and currently fails the required structure validation).

Fix: Reorganize the TypeSpec project to repository folder-structure v2 per https://aka.ms/azsdk/spec-dirs, then re-run `npx tsv specification/hybridnetwork/HybridNetwork.Management` and ensure it passes.

sandipsh (commented):
ARM review (latest iteration): posted blocking findings for current CI failures (missing readme input-file references + TypeSpec folder structure) and one warning for spellcheck typo.

sandipsh:
ARM review update for latest iteration:
- No prior ARM review threads/comments existed on this PR, so no historical ARM thread reconciliation was needed.
- Posted new inline ARM comments for current blockers:
  - Missing readme input-file references to non-existent swagger files (lines 75 and 220 in specification/hybridnetwork/resource-manager/readme.md).
  - TypeSpec FolderStructure v2 validation failure for specification/hybridnetwork/HybridNetwork.Management.
  - Spellcheck typo in readme suppression reason (`dictinary` -> `dictionary`).

Validation status for current changeset (blocking unless noted):
- Swagger LintDiff: failing (ENOENT from readme tag references to missing swagger files).
- Swagger ModelValidation: success.
- Swagger SemanticValidation: success.
- TypeSpec Validation: failing (FolderStructure rule).
- Swagger Avocado: failing (NO_JSON_FILE_FOUND from readme references).
- Swagger BreakingChange (cross-version): failing due same missing-file readme reference.
- SpellCheck: failing (`dictinary`).
- MULTIPLE_API_VERSION: no additional actionable-only exception handling needed in this run.

Suppression review:
- Reviewed new suppression in specification/hybridnetwork/suppressions.yaml.
- Suppression is scoped to TypeSpec config validation, has explicit reason, and is non-security.
- Marked suppression review as approved.

specification/hybridnetwork/resource-manager/readme.md
```diff
@@ -25,7 +27,20 @@ These are the global settings for the hybridnetwork.
 ```yaml
 openapi-type: arm
 openapi-subtype: rpaas
-tag: package-2024-04-15
+tag: package-2025-03-30
```
pshao25: This readme files contains all the versions from RPSaaSMaster, but you only include the swagger of 2025-03-30 in this PR. Is your expectation to port only 2025-03-30 or all?

specification/hybridnetwork/resource-manager/readme.md
kpanugan: Removed the  package-2021-06-01-privatepreview  tag block because this PR only ports API version  2025-03-30 . The readme no longer references the missing private-preview Swagger files.

specification/hybridnetwork/resource-manager/readme.md
kpanugan: The entire invalid  2024-04-01-preview  suppression block containing the typo was removed.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
kpanugan: Moved the project, generated Swagger, and examples under:
specification/hybridnetwork/resource-manager/
  Microsoft.HybridNetwork/HybridNetwork.Management/
``` Updated `tspconfig.yaml`, `readme.md`, and `suppressions.yaml` paths accordingly.

File: specification/hybridnetwork/resource-manager/readme.md (hunk 5 above)
Comment (kpanugan) [reply specification/hybridnetwork/resource-manager/readme.md]: Removed the additional RPSaaSMaster-only version blocks. The readme now adds only  2025-03-30 , while retaining versions already present on public  main .

File: specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp [hunk 6]
```diff
@@ -0,0 +1,2566 @@
...
+using Azure.ResourceManager.Foundations;
+
+namespace Microsoft.HybridNetwork;
+
+/**
+ * The current provisioning state.
+ */
+union ProvisioningState {
+  string,
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
ravimeda: **Blocking 1. [NEW] 🔴** **[[typespec-review §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `.../HybridNetwork/models.tsp` — line 18 (first of 161 in this file; **174 total across 8 `.tsp` files**)

**Issue:** 174 of the 181 `#suppress` directives carry the verbatim converter placeholder `"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"`. Per-file: models.tsp 161 (first line 18), ArtifactStore.tsp 3 (line 152), NetworkServiceDesignVersion.tsp 3 (line 17), ConfigurationGroupValue.tsp 2 (line 16), NetworkFunction.tsp 2 (line 28), ConfigurationGroupSchema.tsp 1 (line 72), NetworkFunctionDefinitionVersion.tsp 1 (line 72), SiteNetworkService.tsp 1 (line 29). Rules silenced include `documentation-required` (×127), `composition-over-inheritance` (×15), `casing-style` (×12), `no-openapi` (×5), `arm-no-record` (×5), plus singletons for `secret-prop`, `no-empty-model`, `patch-envelope`, and `arm-delete-operation-response-codes`. §4.1 states suppression reasons MUST NOT contain `FIXME`/`TODO`/`TBD`. CI reports 153 suppressions pending review.
**Approval context:** `Approved-TypeSpecSuppression` was NOT observed at the session SHA. `Approved-Suppression` is present but covers OpenAPI-side (`readme.md` / `suppressions.yaml`) suppressions, not TypeSpec `#suppress` directives. Please confirm whether an approval covering these exists — if so, apply `Approved-TypeSpecSuppression` and resolve; otherwise obtain approval or address the finding.
**Fix:** For each, either fix the underlying issue (most `documentation-required` cases need only a `/** ... */` doc comment) or replace the FIXME with a specific technical justification.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
```diff
@@ -0,0 +1,47 @@
...
+emit:
+  - "@azure-tools/typespec-autorest"
+options:
+  "@azure-tools/typespec-autorest":
+    omit-unreachable-types: true
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/openapi.json"
+    arm-types-dir: "{project-root}/../../../../common-types/resource-management"
+    emit-lro-options: "all"
+    examples-dir: "{project-root}/examples"
```
ravimeda, line 13: **Blocking 2. [NEW] 🔴** **[[RPC-Async-V1-02](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#65-provisioningstate-requirements-rpc-async-v1-02-rpc-async-v1-03-rpc-async-v1-16) / [provisioning-state.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#provisioningstatemustbereadonly----emitter-workaround-use-read-only-status-schema)]** `.../HybridNetwork/tspconfig.yaml` — line 7-13

**Issue:** `provisioningState` is **not read-only in the emitted contract**. The emitter produces `{"$ref": "#/definitions/ProvisioningState", "description": "...", "readOnly": true}` at **13 distinct schema sites** (openapi.json lines 5563, 5821, 6592, 6649, 7406, 7473, 7540, 7703, 8106, 8384, 8452, 8882, 8961). Per JSON Schema, siblings of `$ref` are ignored, so `readOnly` has no effect — the documented `ProvisioningStateMustBeReadOnly` emitter trap. LintDiff reports it 36 times (per-operation occurrences of those 13 sites). **This is a regression:** in `2024-04-15`, `common.json` line 61-63 carried `"readOnly": true` on the `ProvisioningState` definition itself, so every `$ref` inherited it.
**Fix:** Add `use-read-only-status-schema: true` to the `@azure-tools/typespec-autorest` options block. The `ProvisioningState` union (models.tsp line 16-36) has members named `Succeeded`, `Failed`, `Canceled`, which is what the emitter matches on, so recognition will succeed.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/NetworkFunction.tsp
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Network function resource response.
+ */
+model NetworkFunction
+  is Azure.ResourceManager.TrackedResource<NetworkFunctionPropertiesFormat> {
+  ...ResourceNameParameter<
+    Resource = NetworkFunction,
+    KeyName = "networkFunctionName",
+    SegmentName = "networkFunctions",
+    NamePattern = ""
```
ravimeda: **Blocking 3. [NEW] 🔴** **[[PREFLIGHT-005](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#15-arm-template-deployment-compatibility) / [OAPI-PATTERN-ALLOWLIST](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/pattern-validation.md)]** `.../HybridNetwork/NetworkFunction.tsp` — line 22

**Issue:** `NamePattern = ""`, so the emitted `networkFunctionName` path parameter has **neither `pattern` nor `maxLength`** (openapi.json lines 696, 946, 995, 1050). All 12 other resource types supply a non-empty `NamePattern`. **Regression:** base `common.json` line 590-591 carried `"pattern": "^[^\\s]*[^\\s]+[^\\s]*$"` and `"maxLength": 64`. LintDiff reports `ResourceNameRestriction` 4 times.
**Fix (your choice — option (b) tightens validation across a version boundary):** (a) restore exact parity with `NamePattern = "^[^\\s]*[^\\s]+[^\\s]*$"` plus `@maxLength(64)`; or (b) preferred, adopt the allowlist `^[a-zA-Z0-9][a-zA-Z0-9_-]*$` used by 9 sibling resources, after confirming no existing network function name would be rejected.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+        "responses": {
+          "202": {
+            "description": "Resource operation accepted.",
+            "headers": {
+              "Location": {
+                "type": "string",
+                "description": "The Location header contains the URL where the status of the long running operation can be checked."
+              }
+            }
+          },
```
ravimeda, line 171: **Blocking 4. [NEW] 🔴** **[[RPC-Async-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#64-async-post-action-model-rpc-async-v1-11)]** `stable/2025-03-30/openapi.json` — line 152-161 (source `routes.tsp` line 22-30)

**Issue:** the newly-added async POST `SiteNetworkServices_CancelOperation` declares only `202` (with `Location`) and `default`. (1) §6.4 states unconditionally that an async POST MUST return `202` with **both** `Location` and `Azure-AsyncOperation`; only `Location` is declared. (2) LintDiff reports `PostResponseCodes` at line 129 because no terminal response is declared. _Caveat:_ §6.6 scopes the `Azure-AsyncOperation` MUST to **new** RP namespaces and calls it "strongly recommended" for brownfield; HybridNetwork is brownfield — but §6.4's clause carries no carve-out and this is a brand-new operation in a stable version.
**Fix:** Switch `ArmLroLocationHeader` → `ArmCombinedLroHeaders`, declare the terminal response (`ArmNoContentResponse`, or `200` + schema if the completed cancel returns content), and add `Retry-After`.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, manual decision) against commit [`41507c5`](https://github.com/Azure/azure-rest-api-specs/pull/43894/commits/41507c5a20d3aa6edbffc3bf3555a7fcaaf3ac4d). See inline comments for findings Blocking 1-5, Warning 1-2, Warning 4-10, and Suggestion 1-4. Warning 3 concerns lines unchanged in this PR's diff and is posted as a top-level PR comment rather than inline.

Approval labels observed: `Approved-Suppression`.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp
```diff
@@ -0,0 +1,2566 @@
...
+  /**
+   * The username of the credential.
+   */
+  username?: string;
+
+  /**
+   * The credential value.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  acrToken?: string;
```
ravimeda: **Blocking 5. [EXISTING] 🔴** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#3-secret-detection-in-typespec-sec-secret-detect)]** `.../HybridNetwork/models.tsp` — line 1684-1685 (emitted openapi.json line 6074-6077)

**Issue:** `AzureContainerRegistryScopedTokenCredential.acrToken` — documented "The credential value." — is a bearer credential returned by `ArtifactManifests_ListCredential`. It has no `@secret`, no `"x-ms-secret": true`, and the `secret-prop` rule that exists to catch this is suppressed at line 1684 with the FIXME placeholder. Same gap on `AzureStorageAccountContainerCredential.containerSasUri` (models.tsp line 1748; openapi.json line 6478), which is additionally a plain `string` where §5 requires the `url` scalar for SAS URIs. The project already models three other secrets correctly (models.tsp lines 1794, 1828).
**Previous version:** both present and unannotated in base `common.json` line 294 / 356 — the missing annotation is pre-existing; **the `#suppress` hiding it is new in this PR**.
**Approval context:** `Approved-TypeSpecSuppression` NOT observed. `Approved-Suppression` covers OpenAPI-side suppressions only. Confirm whether an approval covers this specific suppression; if so apply the label and resolve, otherwise obtain approval or address it.
**Fix:** Add `@secret` to both (and change `containerSasUri` to `url`), delete the `#suppress`, regenerate, and confirm `"x-ms-secret": true` appears. Please also confirm it propagates to every variant reachable from the `ArtifactAccessCredential` discriminated base (line 5489) including `AzureStorageAccountCredential.containerCredentials` (line 6492). Retrieval via the `listCredential` POST action is already the correct pattern.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/routes.tsp
```diff
@@ -0,0 +1,36 @@
...
+  cancelOperation is ArmProviderActionAsync<
+    Request = CancelInformation,
+    Response = ArmAcceptedLroResponse<
+      "Accepted. The header contains Location URL to check the status of the asynchronous operation",
+      ArmLroLocationHeader
+    >,
+    Scope = SubscriptionActionScope,
+    Parameters = {},
+    LroHeaders = ArmLroLocationHeader
+  >;
```
ravimeda, line 26: **Warning 1. [NEW] 🟠** **[[arm-api-review §12.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#121-use-post-actions-sparingly)]** `.../HybridNetwork/routes.tsp` — line 17-31

**Issue:** `SiteNetworkServices_CancelOperation` uses `ArmProviderActionAsync<..., Scope = SubscriptionActionScope>` → `POST /subscriptions/{subscriptionId}/providers/Microsoft.HybridNetwork/cancelSiteNetworkServiceOperation`. §12.1 **permits** provider-level POST actions, so the shape is not prohibited — this is an authorization-scope design question. The concern: the action operates on one resource-group-scoped `siteNetworkServices/{name}` supplied in the body (openapi.json line 6535-6551; example line 5-9 passes `/subscriptions/sub1/resourceGroups/rg/.../TestSNS1`). ARM evaluates RBAC against the action URI, so authorization is checked at **subscription** scope — a principal holding the provider action can cancel in-flight PUTs on any SiteNetworkService in the subscription, including resource groups where they hold no rights.
**Fix:** (a) re-model as `ArmResourceActionAsync<SiteNetworkService, CancelInformation, ...>` → `POST .../resourceGroups/{rg}/.../siteNetworkServices/{name}/cancelOperation`, dropping `siteNetworkServiceReference`; or (b) keep the provider shape, confirm at ARM API review office hours, and document that the RP performs an explicit authorization check on the referenced resource ID.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+          {
+            "name": "Delete",
+            "value": "Delete"
+          }
+        ]
+      }
+    },
+    "ImageArtifactProfile": {
+      "type": "object",
+      "descri
```
ravimeda: **Warning 2. [NEW] 🟠** **[[typespec-review §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns) — "Doc comment copy-paste errors"]** `stable/2025-03-30/openapi.json` — line 7991, 7995, 8002 and line 8044, 8048, 8055

**Issue:** placeholder descriptions `"The description for page model"`, `"The description for value property"`, `"The description for nextLink property"` on `ProxyArtifactOverviewListResult` and `ProxyArtifactVersionsOverviewListResult` ship to learn.microsoft.com and every SDK docstring. **Regression:** base `pureProxyArtifact.json` lines 219/224/227 and 237/242/245 had real text.
**Fix:** Add real doc comments on the corresponding list-result models in `models.tsp` and regenerate.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+            "$ref": "../../../../../../common-types/resource-management/v3/types.json#/parameters/ResourceGroupNameParameter"
+          },
+          {
+            "name": "networkFunctionName",
+            "in": "path",
+            "description": "Resource name for the network function resource.",
+            "required": true,
+            "type": "string"
+          },
+          {
```
ravimeda, line 871: **Warning 4. [EXISTING] 🟠** **[[RPC-Patch-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#43b-patch-should-support-updating-envelope-properties-rpc-patch-v1-09-rpc-patch-v1-11)]** openapi.json — line 852 (`NetworkFunctions_UpdateTags`), line 1301 (`Publishers_Update`). `NetworkFunction`, `Publisher`, and `SiteNetworkService` declare top-level `identity`, but every PATCH takes `TagsObject` only. Same shape in `2024-04-15`. **Fix:** add `identity` to the PATCH bodies, or annotate it immutable.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
ravimeda, line 7963: **Warning 5. [EXISTING] 🟠** **[[arm-api-review §8.9](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#89-provisioningstate-correctness)]** openapi.json — line 7923-7935 (source models.tsp line 16-36). `ProvisioningState` includes `Deleted`, which §8.9 names as invalid. All three terminal states are present. (`Unknown` is **not** flagged — RP-defined non-terminal states are permitted; the newly added `Cancelling` is likewise legitimate.) Same values in base `common.json` line 61-79. **Fix:** deprecate `Deleted`; `modelAsString: true` allows retiring it without an SDK break.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/ArtifactStore.tsp
```diff
@@ -0,0 +1,261 @@
...
+  proxyArtifactGet is ArmResourceActionSync<
+    ArtifactStore,
+    void,
+    ArmResponse<ProxyArtifactVersionsOverviewListResult>,
+    Parameters = {
+      /**
+       * The name of the artifact.
+       */
+      @maxLength(64)
+      @pattern("^[^\\s]*[^\\s]+[^\\s]*$")
```
ravimeda, line 182: **Warning 6. [EXISTING] 🟠** **[[OAPI-PATTERN-ALLOWLIST](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/pattern-validation.md)]** `.../ArtifactStore.tsp` line 178, 218, 226; `.../ArtifactManifest.tsp` line 24; `.../Component.tsp` line 23 (plus `go,javascript` override copies at `client.tsp` line 81, 91). `artifactManifestName`, `componentName`, `artifactName`, `artifactVersionName` use the denylist `^[^\s]*[^\s]+[^\s]*$`, which accepts emoji, control characters, and every non-Latin script. The `ArtifactStore.tsp` sites are the ones that emit (openapi.json 2487, 2496). Identical in base `common.json`. **Fix:** replace with an allowlist (e.g. `^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,63}$`) after confirming compatibility.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+            "name": "artifactVersionName",
+            "in": "path",
+            "description": "The name of the artifact version.",
+            "required": true,
+            "type": "string",
+            "maxLength": 64,
+            "pattern": "^[^\\s]*[^\\s]+[^\\s]*$"
+          },
+          {
+            "name": "artifactName",
```
ravimeda, line 2517: **Warning 7. [EXISTING] 🟠** **[[RPC-Patch-V1-12](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#45-patch-request-body-and-query-parameters-rpc-patch-v1-12)]** openapi.json — line 2490 (operation at 2446-2447). `ProxyArtifact_UpdateState` is a PATCH with a required `artifactName` **query** parameter; §4.5 forbids query parameters other than `api-version`. Source: `ArtifactStore.tsp` line 189, `@query("artifactName")` at line 227. Same shape in base. **Fix:** move to the path in a future version — changing `2025-03-30` in place would be breaking.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+            }
+          }
+        },
+        "x-ms-examples": {
+          "Update tags for network function resource": {
+            "$ref": "./examples/NetworkFunctionUpdateTags.json"
+          }
+        }
+      },
+      "delete": {
```
ravimeda, line 901: **Warning 8. [EXISTING] 🟠** **[[RPC-Delete-V1-01](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#51-delete-response-codes-rpc-delete-v1-01)]** openapi.json — line 882 (`NetworkFunctions_Delete`). LintDiff reports `DeleteResponseCodes` once here: this async DELETE declares `200/202/204/default`; §5.1 notes `200` "belongs exclusively to synchronous DELETE". It is the only one of 12 DELETEs affected. **Fix:** drop the OK variant from the `Response` union at `NetworkFunction.tsp` line 58-62 and remove the suppression.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+        },
+        "x-ms-examples": {
+          "Get component resource": {
+            "$ref": "./examples/ComponentGet.json"
+          }
+        }
+      }
+    },
+    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.HybridNetwork/networkFunctions/{networkFunctionName}/executeRequest": {
+      "post": {
```
ravimeda, line 1076: **Warning 9. [EXISTING] 🟠** **[[RPC-POST-V1-03](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#121-use-post-actions-sparingly)]** openapi.json — line 1051, 1713, 1786, 2604, 2825. LintDiff reports `PostResponseCodes` at these five lines. §12.1 requires an async POST to return `202`, and says it **SHOULD** additionally specify `200` with a response schema when the final response has content. `NetworkFunctions_ExecuteRequest` declares `200` with **no** schema; the four `ArtifactStores` actions declare only `202`/`default`, so a client cannot tell what a completed call returns. Identical in base. **Fix:** plan for the next API version.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
```diff
@@ -0,0 +1,9003 @@
...
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Resource deleted successfully."
+          },
+          "202": {
+            "description": "Accepted. The header contains 'Azure-AsyncOperation' header pointing to an operations resource. The operation URI can be queried to find the current state of the operation.",
```
ravimeda, line 932: **Warning 10. [EXISTING] 🟠** **[[typespec-review §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** openapi.json — line 911, 1089, 1759, 1832, 2650, 2720, 2794, 2871. Eight `202` descriptions claim the response "contains 'Azure-AsyncOperation' header…" while `headers` declares only `Location`. None of the 25 `202` responses in the file declares that header. Same mismatch in base. **Fix:** emit the header (`ArmCombinedLroHeaders`), which also satisfies the RPC-Async-V1-06 brownfield recommendation; otherwise correct the descriptions.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp
```diff
@@ -0,0 +1,2566 @@
...
+
+  /**
+   * The resource related to the component resource.
+   */
+  resources?: Resources;
+
+  /**
+   * The next expected update of deployment status.
+   */
+  // FIXME: (utcDateTime) Please double check that this is the correct type for your scenario.
```
ravimeda: **Suggestion 1. [NEW] 🔵** **[[typespec-review §8.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#82-validation)]** 10 leftover `// FIXME:` converter comments — `models.tsp` lines 754, 825, 861, 893, 929, 960, 1006, 1700, 1727 and `routes.tsp` line 1. All nine `utcDateTime` ones look correctly typed; delete the comments.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
ravimeda, line 8526: **Suggestion 2. [EXISTING] 🔵** **[[arm-api-review §2.6.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#261-standard-sku-schema)]** openapi.json line 8480-8497 — `Sku` has only `name`/`tier`; standard schema adds `size`/`family`/`capacity`, and common-types `Sku` is preferred.

ravimeda, line 8180: **Suggestion 3. [EXISTING] 🔵** **[[TSP-ARM-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#63-use-armresourceidentifier-for-resource-references-tsp-arm-resource-id)]** openapi.json line 8143-8151 — `ReferencedResource.id` is a bare `string`; `armResourceIdentifier` would emit `format: arm-id`.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/examples/2025-03-30/SiteNetworkServicesCancelOngoingPUTOperation.json
```diff
@@ -0,0 +1,21 @@
...
+      "siteNetworkServiceReference": {
+        "id": "/subscriptions/sub1/resourceGroups/rg/providers/Microsoft.HybridNetwork/siteNetworkServices/TestSNS1"
+      }
+    },
+    "subscriptionId": "sub1"
+  },
+  "responses": {
+    "202": {
+      "headers": {
+        "Location": "https://management.azure.com/providers/microsoft.hybridnetwork/locations/EASTUS2EUAP/operationStatuses/0b37c625-7e7c-49c5-b86e-c817c85acb5d*B77A34159CA34A0BD446D3BC7EC120649181590DDF991982765DAC1A87491B2D?api-version=2021-05-01"
```
ravimeda: **Suggestion 4. [NEW] 🔵** `examples/2025-03-30/SiteNetworkServicesCancelOngoingPUTOperation.json` — line 15 `Location` ends with a stale `?api-version=2021-05-01`; line 20 `title` has a trailing space.

ravimeda:
**Warning 3. [NEW] 🟠** **[[typespec-review §8.3](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#83-sdk-generation-cutover)]** `.../HybridNetwork/readme.md` — line 142-152

**Issue:** the default tag now points at the TypeSpec-generated `openapi.json` (line 28) and `tspconfig.yaml` configures the Python/Java/Go emitters, but `swagger-to-sdk` still lists all seven repos and the per-language `readme.{go,python,ruby,typescript,csharp}.md` files remain referenced (line 157+). §8.3 requires disabling OpenAPI-based SDK generation on cutover; leaving both active means two generators race for the same packages.
**Fix:** Remove the `swagger-to-sdk` language entries now driven by `tspconfig.yaml` and delete the corresponding `readme.{language}.md` files. Note the `@azure-tools/typespec-ts` block is commented out at `tspconfig.yaml` line 38-44 and covered by the new suppression, so JS is intentionally not yet on the TypeSpec path.

_(This finding concerns lines unchanged in this PR's diff, so it is posted as a top-level PR comment rather than inline.)_

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
```diff
@@ -0,0 +1,48 @@
...
+    service-dir: "sdk/resourcemanager/hybridnetwork"
+    emitter-output-dir: "{output-dir}/{service-dir}/armhybridnetwork"
+    module: "github.com/Azure/azure-sdk-for-go/{service-dir}/armhybridnetwork"
+    fix-const-stuttering: true
+    flavor: "azure"
+    generate-samples: true
+    generate-fakes: true
+    head-as-boolean: true
+    inject-spans: true
+  # "@azure-tools/typespec-ts":
```
ArthurMa1978: Please enable this setting for JS

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
```diff
@@ -0,0 +1,57 @@
...
+  #   experimental-extensible-enums: true
+  #   package-dir: "arm-hybridnetwork"
+  #   flavor: azure
+  #   emitter-output-dir: "{output-dir}/{service-dir}/arm-hybridnetwork"
+  #   package-details:
+  #     name: "@azure/arm-hybridnetwork"
+  "@azure-tools/typespec-csharp":
```
ArthurMa1978: remove this setting since this csharp emitter was deprecated for mgmt. sdk and new emitter is `@azure-typespec/http-client-csharp-mgmt`

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp
kpanugan: Fixed in `5fad9d3`.

All 174 FIXME placeholders are gone — searching the eight `.tsp` files for
`#suppress` lines containing `FIXME` now returns 0. The 181 remaining `#suppress`
directives each carry a specific technical justification. Most of the
`documentation-required` cases were resolved by adding the missing doc comments
rather than by suppressing.

On the label: we have not applied `Approved-TypeSpecSuppression`. Now that the
placeholders are replaced with real justifications, please confirm whether the
remaining directives still need that label or whether the written justifications
are sufficient.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
kpanugan: Fixed in `5fad9d3`.

`use-read-only-status-schema: true` added to the `@azure-tools/typespec-autorest`
options block (tspconfig.yaml line 13). `"readOnly": true` now sits on the
`ProvisioningState` definition itself, so all 13 `$ref` sites inherit it.

For the record, this was a conversion regression rather than a design change. The
hand-written 2025-03-30 swagger reviewed in Feb 2025 already had `"readOnly": true`
on the definition; it was dropped during the Oct 2025 TypeSpec conversion.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/NetworkFunction.tsp
kpanugan: Fixed in `5fad9d3`, using option (a).

`NamePattern = "^[^\s]*[^\s]+[^\s]*$"` and `@maxLength(64)` restored. We did not
take option (b): the approved 2025-03-30 swagger (`common.json` line 623) and
2024-04-15 both carry this pattern, so tightening to an allowlist across a stable
version boundary risks rejecting network function names that already exist.

This was also a conversion regression — both constraints were in the reviewed
swagger and lost in conversion.

Please note this puts Blocking 3 in direct conflict with Warning 6; see that thread.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Fixed in `f8deb48`.

routes.tsp now uses:

    LroHeaders = ArmCombinedLroHeaders & Azure.Core.Foundations.RetryAfterHeader

and the redundant explicit `ArmAcceptedLroResponse` was removed from `Response`,
since `ArmProviderActionAsync` already emits the 202. The emitted 202 now declares
`Azure-AsyncOperation`, `Location` and `Retry-After`; the terminal response is
declared via `ArmNoContentResponse`; `final-state-via` correctly remains `location`
because `ArmCombinedLroHeaders` places `@finalLocation` on `location`. The example
was updated to include the `Azure-AsyncOperation` header.

For transparency: the version of this operation approved in Feb 2025 declared only
`Location`, with `final-state-via: location` and no 204. The current shape is a
superset and additive only, but it is a delta from what was signed off, so we are
flagging it rather than leaving it implicit.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp
kpanugan: Partially addressed; asking for sign-off on the remainder.

Done in `f8deb48`: `containerSasUri` now carries a description marking it
short-lived credential material.

Not done, deliberately — `@secret` on both properties, and `string` to `url`:

- `@secret` declares that a property must never appear in a response. Both
  properties are the intended payload of the `listCredential` POST actions, so
  applying it would contradict the purpose of those operations. That reasoning is
  now written into the `secret-prop` suppression in models.tsp instead of being
  left as a FIXME.
- `containerSasUri` has shipped as a plain `string` since 2023-09-01, is unchanged
  in 2024-04-15 (`common.json:356`), and is unchanged in the approved 2025-03-30
  swagger (`common.json:357`). Adding `format: uri` trips openapi-diff
  `TypeFormatChanged` on the cross-version check, and the .NET generator remaps
  `string` to `System.Uri`, which is an SDK break on a GA property.

The three secrets the project does model correctly are untouched; the `x-ms-secret`
count is identical in 2024-04-15 and 2025-03-30.

We have not applied `Approved-TypeSpecSuppression`. Please advise whether the
written justification is sufficient or whether you want the label plus a formal
suppression approval.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/routes.tsp
kpanugan: Not changed. This shape was already reviewed and approved.

`POST /subscriptions/{subscriptionId}/providers/Microsoft.HybridNetwork/cancelSiteNetworkServiceOperation`
is exactly what shipped in the hand-written 2025-03-30 swagger that went through API
review in Feb 2025 (PR #25860). It was reached deliberately over several iterations —
the history shows `providerlevel action` (Feb 12), `try with PUT` (Feb 12),
`define rootendpoint at subscription level` (Feb 13), `removed RG name parameter`
(Feb 13) — and the last pre-conversion commit is `addressed the review comment`
(Feb 17, `e18b0458`).

So this is not new design introduced by the TypeSpec conversion; the conversion
reproduced the approved contract. Re-modelling to `ArmResourceActionAsync` would
change an approved URL.

Taking option (b): the RP performs an explicit authorization check on the
`siteNetworkServiceReference` resource ID before acting. Happy to bring it to ARM
office hours if you would still like it re-confirmed.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Fixed in `5fad9d3`. Real doc comments added to the list-result models;
`"The description for ..."` no longer appears anywhere in the emitted openapi.json.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Not changed; suppressed with justification.

`PatchIdentityProperty` is suppressed for the 2025-03-30 tag in readme.md, scoped in
`a0f3d1b` to the three specific PATCH parameter sets rather than the whole operation.
Adding `identity` to the PATCH bodies would change the request contract of operations
that have shipped since 2023-09-01 and behave identically in 2024-04-15. The deployed
service does not accept `identity` on PATCH today, so declaring it would misrepresent
the contract. Tracked for the next API version.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Not changed.

`Deleted` is present in 2024-04-15 (`common.json:72`) and in the approved 2025-03-30
swagger (`common.json:72`), and the service still returns it. Removing it from a
stable version would misrepresent what clients observe. We agree it should go — tracked
for the next API version, where `modelAsString: true` lets us retire it without an SDK
break.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/ArtifactStore.tsp
kpanugan: Not changed, and we need a decision because this conflicts with Blocking 3.

Blocking 3 asked us to restore `^[^\s]*[^\s]+[^\s]*$` on `networkFunctionName` for
parity with the prior version. Warning 6 asks us to replace that same pattern with an
allowlist on four other parameters. Both cannot be satisfied without a breaking change.

We took the parity option for Blocking 3. These patterns are identical in 2024-04-15
(`common.json` lines 535, 575, 585, 595, 605) and in the approved 2025-03-30 swagger.
Tightening them in a stable version would reject names that already exist.

Could you confirm which of the two you would like us to follow? Our preference is
parity now and an allowlist in the next version.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Agreed, not changed. `ProxyArtifact_UpdateState` has taken `artifactName` as a query
parameter since it shipped; moving it to the path in 2025-03-30 would be breaking.
Tracked for the next API version, as you suggested.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Not changed; suppressed with justification.

`DeleteResponseCodes` is suppressed in readme.md, scoped to the single
`NetworkFunctions_Delete` path. The operation returns 200/202/204 in 2024-04-15 and in
the deployed service. Dropping the 200 from the declared contract would not stop the
service returning it. Tracked for the next API version.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Agreed, not changed; suppressed with justification.

`PostResponseCodes` is suppressed in readme.md, scoped to the five specific POST paths,
with the reason recorded that these response shapes are carried forward unchanged from
2024-04-15 and match the deployed service. Tracked for the next API version, as you
suggested.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Partially fixed, plus one correction to the finding.

The finding states "None of the 25 `202` responses in the file declares that header."
That is not accurate — 12 of them already declared `Azure-AsyncOperation` before this
round of changes. The 13th was added in `f8deb48` for the cancel operation, per
Blocking 4.

The 8 sites you listed do have the mismatch. All 8 exist with the identical mismatch in
the approved 2025-03-30 swagger, which had 10 such sites; the TypeSpec conversion added
12 headers and corrected 2 descriptions, taking the count from 10 down to 8.

Adding headers to the remaining 8 changes the declared response contract of GA
operations, so we would prefer to do that in the next API version. If you would rather
we correct the descriptions instead, that is a documentation-only change and we can do
it in this PR — just let us know.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/models.tsp
kpanugan: Fixed in `5fad9d3`. All 10 `// FIXME:` comments deleted. The `utcDateTime` types were
left as they were, since they are correct.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/stable/2025-03-30/openapi.json
kpanugan: Not changed. The `Sku` model is unchanged from 2024-04-15
(`siteNetworkService.json:432`) and from the approved 2025-03-30 swagger (line 484).
Moving to the common-types `Sku` would add properties the service does not accept.
Tracked for the next API version.

kpanugan: Not changed. `ReferencedResource.id` is a bare `string` in 2024-04-15
(`common.json:38-45`) and in the approved 2025-03-30 swagger. Adding `format: arm-id`
changes the generated SDK type on a GA property. Tracked for the next API version.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/examples/2025-03-30/SiteNetworkServicesCancelOngoingPUTOperation.json
kpanugan: Fixed in `5fad9d3`. The stale `?api-version=2021-05-01` on the `Location` header is
corrected and the trailing space in `title` is removed.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/tspconfig.yaml
kpanugan: Done in `f8deb48`. The `@azure-tools/typespec-csharp` block is removed;
`@azure-typespec/http-client-csharp-mgmt` remains.

kpanugan:
Not changed yet. readme.md lines 214-225 still list the seven `swagger-to-sdk` repos.
That block predates the TypeSpec conversion. We are happy to remove it in this PR if
you confirm the tspconfig emitters fully replace it for this RP; we left it in place to
avoid breaking any pipeline still keyed off the autorest config.

kpanugan:
Pushed `63ec66a`. Summary of where the review stands:

Fixed in this branch: Blocking 1 (FIXME suppressions), Blocking 2 (provisioningState
read-only), Blocking 3 (networkFunctionName constraints), Blocking 4 (async POST LRO
headers), Warning 2 (placeholder descriptions), Suggestion 1 (FIXME comments),
Suggestion 4 (example), and the deprecated csharp emitter. Also added the required
`service.yaml` manifest, which clears the TypeSpec Validation check.

The TypeSpec Validation failure on the previous run was caused by a compiler bump on
main (1.14.0 to 1.15.0), not by anything in this PR. 1.15.0 adds the `use-interface`
and `use-operation-decorator` rules, which flag two pre-existing constructs:

- `ProxyArtifact.updateState`, the custom PATCH with an `artifactName` query parameter,
  is now decorated with `@Azure.ResourceManager.Private.legacyResourceOperation`,
  matching the existing precedent in
  `specification/marketplace/resource-manager/Microsoft.Marketplace/Marketplace/routes.tsp`.
- `proxyArtifactUpdateState` in client.tsp is an `@@override` target for go and
  javascript, not a service operation, so `use-interface` is suppressed there.

specification/hybridnetwork/resource-manager/Microsoft.HybridNetwork/HybridNetwork/NetworkFunction.tsp
kpanugan: Clarification on the 2024-04-15 comparison, since the parity claim above was not precise enough.

2024-04-15 applies this constraint inconsistently to the same resource:

- DELETE, `executeRequest` POST, and both `components` GETs reference `./common.json#/parameters/NetworkFunctionNameParameter`, which carries `pattern: "^[^\s]*[^\s]+[^\s]*$"` and `maxLength: 64` (common.json:590).
- GET, PUT and PATCH declare `networkFunctionName` inline with no `pattern` and no `maxLength` (networkFunction.json:369, 430, 481).

So 4 of the 7 network-function operations were already constrained in GA; 3 were not.

In the 2025-03-30 content as copied from rpsaasmaster, none of the 7 carried it, and `networkFunctionName` was the only one of 14 resource-name path parameters in the whole spec without `pattern` + `maxLength: 64`. The other 13 all had it.

This change applies it to all 7, which makes the spec internally consistent and matches the constrained half of 2024-04-15. `Swagger BreakingChange` and `Breaking Change(Cross-Version)` both pass on `63ec66a`.

If the preference is to preserve the GA split exactly rather than make it uniform, say so and it can be reverted to inline-unconstrained on GET/PUT/PATCH.

PR #43909, opened 2026-06-11
Add EventGrid .NET mgmt migration compat customizations

weidongxu-microsoft:
## Summary\n- add EventGrid C# mgmt migration customizations for compatibility-sensitive models\n- restore EventGridJsonInputSchemaMapping naming for the .NET mgmt package\n- map webhook tenant id to Guid for the regenerated SDK surface\n\n## Testing\n- local TypeSpec compile with @azure-typespec/http-client-csharp-mgmt emitter


gary-x-li (approved):

ArthurMa1978 (approved):

specification/eventgrid/resource-manager/Microsoft.EventGrid/EventGrid/ClientResource.tsp
```diff
@@ -85,6 +85,9 @@ interface Clients {
 @@clientName(TlsVersion.`1.0`, "One0", "go,javascript");
```
ArthurMa1978: don't rename the client.tsp file

specification/eventgrid/resource-manager/Microsoft.EventGrid/EventGrid/client.tsp
```diff
@@ -0,0 +1,672 @@
+import "@azure-tools/typespec-client-generator-core";
```
ArthurMa1978, line 2: As a common convention, client settings should be defined in client.tsp and do not need to be separated into a new TSP file.

ArthurMa1978 (changes requested):

specification/eventgrid/resource-manager/Microsoft.EventGrid/EventGrid/client.tsp
weidongxu-microsoft: renamed to client.tsp (agent seems choose this name because there was a `Client.tsp` for Client resource)

specification/eventgrid/resource-manager/Microsoft.EventGrid/EventGrid/ClientResource.tsp
weidongxu-microsoft: move language specific decorators to the real `client.tsp` file

ArthurMa1978 (approved):

PR #43920, opened 2026-06-11
selective import to support python sdk generation

### LINKED ISSUE #43939: Foundry SDKs mismatched exception types
Writing this issue up to recap a few internal discussions.
For the upcoming Python Foundry SDK release, we're seeing a regression introduced by #43493

Here is the context for the regression.

## On default status code

We originally had operations defined like this

```typespec
 @get read(@path id: Widget.id): Widget | Error;
```

Which resulted in the following OpenAPI description

```yaml
openapi: 3.2.0
paths:
  "/widgets":
    get:
      responses:
        200:
          content:
            "application/json":
              $schema: "#/components/schemas/Widget"
        "default":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
```

**One important aspect here from a specification perspective is: if the client gets anything other than a 200 response, it MAY expect json with the Error schema to be returned, even on a 3XX response** That's caused by the default entry, and it's just inaccurate information we're providing to the clients.
https://spec.openapis.org/oas/v3.2.0.html#responses-object

Instead we can define status ranges (more on that below)

And also resulted in the following Python code being generated 

```python
error_map: MutableMapping = {
            401: ClientAuthenticationError,
            404: ResourceNotFoundError,
            409: ResourceExistsError,
            304: ResourceNotModifiedError,
        }
#...
if response.status_code not in [200]:
            if _stream:
                try:
                    response.read()  # Load the body in memory and close the socket
                except (StreamConsumedError, StreamClosedError):
                    pass
            map_error(status_code=response.status_code, response=response, error_map=error_map)
            error = _failsafe_deserialize(
                _models.Error,
                response,
            )
            raise HttpResponseError(response=response, model=error)
```

**Another important aspect here, is that the python generator has made the design choice to throw a specific exception type for 401/404/409** (and 304 for some reason even though it's not an error class? but let's ignore that). It also DOES NOT read the response body (so it ignores the default above). Those two aspects impact the developer experience:
- positively for the specialized exception types, allowing the consumer to catch those exceptions and implement handling for those cases
- negatively for the lack of response body as they are going to miss some context (error code, message, etc...)

## On using ranges

The pull request I have linked above, changes things to use error ranges instead

```typespec
@get read(@path id: Widget.id): Widget | {
    @minValue(400)
    @maxValue(599)
    @statusCode
    code: int32;
    @body
    body: Error
    };
```

The resulting OpenAPI description being

```yaml
openapi: 3.2.0
paths:
  "/widgets":
    get:
      responses:
        200:
          content:
            "application/json":
              $schema: "#/components/schemas/Widget"
        "4XX":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
        "5XX":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
```

**This is more accurate, since it instructs the client to expect response bodies only from 400-599 status codes, which is closer to what the service is actually doing**

And the resulting python code being

```python
error_map: MutableMapping = {
            304: ResourceNotModifiedError,
        }
# ...

if response.status_code not in [200]:
            if _stream:
                try:
                    response.read()  # Load the body in memory and close the socket
                except (StreamConsumedError, StreamClosedError):
                    pass
            map_error(status_code=response.status_code, response=response, error_map=error_map)
            error = None
            if 400 <= response.status_code <= 599:
                error = _failsafe_deserialize(_models.Error, response)
            raise HttpResponseError(response=response, model=error)

```

**Here we GAINED model deserialization on error status codes, but we LOST the specialized exception types for those specific codes**

## On adding specific status code entries

Short of patching the python emitter, which we could do since the semantic are already here to tell it *for 401/404/409 expect those response bodies*, we can do the following.

```typespec
  /** Read widgets */
  @get read(@path id: Widget.id): Widget | {
    @statusCode()
    code: 401;
    @body
    body: Error
    } | {
    @minValue(400)
    @maxValue(599)
    @statusCode
    code: int32;
    @body
    body: Error
    };
```

Which results in the following OpenAPI description

```yaml
openapi: 3.2.0
paths:
  "/widgets":
    get:
      responses:
        200:
          content:
            "application/json":
              $schema: "#/components/schemas/Widget"
        "401":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
        "4XX":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
        "5XX":
          content:
            "application/json":
              $schema: "#/components/schemas/Error"
```

This is redundant, but essentially means the same as the previous

And results in the following python code

```python
error_map: MutableMapping = {
            304: ResourceNotModifiedError,
        }
# ...
if response.status_code not in [200]:
            if _stream:
                try:
                    response.read()  # Load the body in memory and close the socket
                except (StreamConsumedError, StreamClosedError):
                    pass
            map_error(status_code=response.status_code, response=response, error_map=error_map)
            error = None
            if response.status_code == 401:
                error = _failsafe_deserialize(_models.Error, response)
                raise ClientAuthenticationError(response=response, model=error)
            if 400 <= response.status_code <= 599:
                error = _failsafe_deserialize(_models.Error, response)
            raise HttpResponseError(response=response, model=error)
```

**Here we have BOTH the specialized exception type for 401 AND the model deserialization** this essentially fixes the regression and brings an improvement.

## What does the service actually respond?

```shell
curl -X POST "https://openai-typespec.services.ai.azure.com/openai/v1/chat/completions" -H "Content-Type: application/json" -H "Authorization: Bearer $AZURE_API_KEY" -d '{"messages": [{"role": "user","content": "I am going to Paris, what should I see?"},{"role": "assistant","content": "Paris, the capital of France, is known for its stunning architecture, art museums, historical landmarks, and romantic atmosphere. Here are some of the top attractions to see in Paris:\n\n1. The Eiffel Tower: The iconic Eiffel Tower is one of the most recognizable landmarks in the world and offers breathtaking views of the city.\n2. The Louvre Museum: The Louvre is one of the largest and most famous museums, housing an impressive collection of art and artifacts, including the Mona Lisa.\n3. Notre-Dame Cathedral: This beautiful cathedral is one of the most famous landmarks in Paris and is known for its Gothic architecture and stunning stained glass windows.\n\nThese are just a few of the many attractions that Paris has to offer. With so much to see and do, it is no wonder that Paris is one of the most popular tourist destinations in the world."},{"role": "user","content": "What is so great about #1?"}],"model": "gpt-5.2","max_completion_tokens": 4096,"reasoning_effort": "medium"}'
```
(leaving the var empty)

results in 401 with the following json response body

```json
{"error":{"code":"401","message":"Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource."}}
```

## Conclusion

Our options being:
- revert the ranges, but break the docs workstream
- add the additional status codes on top of the range
- patch the emitter

It stands the logical conclusion that gets us the shortest path to success is to include the missing (but redundant) status codes, which I have started doing in #43920

Hopefully this write-up provides all the necessary context that lead up to the decision, both as a basis for discussion but also as a log entry for future review of this space.

baywet:
## Summary
- inventory the PR #43493 range-based TypeSpec error response additions

- adds selective imports to have the ranges in the docs workstream and have defaults in python with specific exception types


glecaros:
hey @baywet, we actually don't want to document trivial error messages in the API, this was direct feedback we got, and even got it included in the guidelines; there we say that only non-trivial errors should be spec'd by the service, and as such we leave it to the service teams on a case-by-case basis

baywet:
@glecaros apologies for not providing more context on the PR description, it was late in my day, I needed to run out but also wanted to provide a basis for @dargilco to keep the investigation and tests going.
I have written a detailed analysis for everyone to read here https://github.com/Azure/azure-rest-api-specs/issues/43939

With that context in mind, can you please:
- let us know who provided that feedback? maybe loop them in this PR so they can contribute to the problem resolution
- share some context about *why* having those redundant status codes in the definition would be a problem?

Thanks!

glecaros (approved):

dargilco (approved):

PR #43922, opened 2026-06-11
Migrate from npm to pnpm workspaces with catalogs

timotheeguerin:
## Summary

Converts the repository from npm to pnpm workspaces with catalogs for shared dependency version management.

### New files
- `pnpm-workspace.yaml` — defines workspace packages, `allowBuilds`, `overrides`, and a catalog with 21 shared dependency versions
- `pnpm-lock.yaml` — replaces `package-lock.json`

### Package management
- All `file:` references replaced with `workspace:*` protocol
- Common dependency versions (typescript, prettier, vitest, eslint, etc.) centralized in catalog
- Each workspace package uses `catalog:` instead of hardcoded versions
- `overrides` moved from root `package.json` to `pnpm-workspace.yaml` (pnpm 11 requirement)
- Added `packageManager: "pnpm@11.5.1"` to root `package.json`

### CI/CD
- Added `pnpm/action-setup@v6` to the setup action
- Changed cache from `npm` to `pnpm`
- Replaced all `npm ci` → `pnpm install --frozen-lockfile`
- Replaced all `npm exec --no --` → `pnpm exec`
- Replaced all `npm run` → `pnpm run` in workflows

### Workspace packages (15 total)
- Root, `.github/`, `.github/shared/`, `eng/tools/`, and all 11 sub-packages under `eng/tools/`

## Verification
- `pnpm install` ✅
- `eng/tools` TypeScript build ✅
- `.github/shared` build + tests (100% coverage) ✅
- `tsp compile` on sample spec ✅

## Notes
- `lint-diff`'s `marked` bumped from `^17` → `^18` (unified via catalog)
- Added `@types/js-yaml` to packages that relied on npm's flat hoisting
- `eng/common/tsp-client` and `eng/common/spelling` remain independent (not workspace packages)


.github/shared/src/exec.js
```diff
@@ -89,46 +88,41 @@ export async function execFile(file, args, options = {}) {
...
+  logger?.info(`execNpm(${JSON.stringify(allArgs)})`);
+
+  try {
+    // Use shell: true to support pnpm on all platforms (on Windows, pnpm is typically
+    // installed as pnpm.cmd which requires shell to resolve)
+    const result = await execFileImpl("pnpm", allArgs, {
+      cwd,
+      maxBuffer,
+      shell: process.platform === "win32",
+    });
```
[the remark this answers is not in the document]
timotheeguerin: Kept `shell` enabled on Windows (but disabled on every other platform) rather than removing it entirely: on Windows `pnpm` is the `pnpm.cmd` batch shim, and since the fix for CVE-2024-27980 Node refuses to spawn `.cmd`/`.bat` files via `execFile`/`spawn` unless `shell: true` (it throws `EINVAL`). So calling `pnpm.cmd` directly with shell off would break Windows. I made the executable name explicit (`pnpm.cmd` on win32, `pnpm` elsewhere) and expanded the comment to document why shell is required only on Windows. Non-Windows paths run shell-free, avoiding the quoting/injection risk there.

.devcontainer/devcontainer.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "TypeSpec Dev Container",
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:dev-20",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:dev-24",
   // Default is 2 cores which is too slow for this codebase
   "hostRequirements": {"cpus": 4},
-  "postCreateCommand": "npm ci",
+  "postCreateCommand": "npm install -g pnpm@11 && pnpm install --frozen-lockfile",
```
mikeharder: can we avoid global install, and prefer a local install?  should we pin a specific version, in case the latest version isn't on our feed, or has a vulnerability?

.github/actions/install-deps-github-script/action.yaml
```diff
@@ -9,9 +9,3 @@ runs:
       with:
         # actions/github-script@v8 uses Node 24
         node-version: 24.x
-        # "--no-audit": improves performance
-        # "--no-fund": improves performance
-        # "--prefer-offline": improves performance
-        # "--omit dev": not needed at runtime, improves performance
-        install-command: "npm ci --no-audit --no-fund --prefer-offline --omit dev"
-        working-directory: ./.github
```
mikeharder, line 17: for performance, it's important to only install the subset of dependencies under .github, not all the dependencies for the whole repo.  maybe we need two pnpm-lock.yaml?  or does pnpm allow installing a subset of deps from a lockfile?

.github/actions/setup-node-install-deps/action.yaml
```diff
@@ -1,26 +1,33 @@
...
+    description: 'Path (relative to the repository root) of the package.json containing the "packageManager" field used to select the pnpm version'
+    default: "package.json"
 
 runs:
   using: "composite"
 
   steps:
+    - uses: pnpm/action-setup@v6
+      with:
+        package_json_file: ${{ inputs.package-json-file }}
```
mikeharder:
```suggestion
        package_json_file: ${{ inputs.working-directory }}/package.json
```

we'll need to either make this action respect the existing working-directory param, or re-design this more deeply

.github/shared/src/exec.js
```diff
@@ -81,54 +80,57 @@ export async function execFile(file, args, options = {}) {
...
+    // spawn .cmd/.bat files unless shell is enabled. On other platforms, call the
+    // "pnpm" binary directly with shell disabled to avoid shell quoting/parsing risks.
+    // Exclude the platform-specific branch from coverage (only one side runs per OS).
+    /* v8 ignore next */
+    const file = isWindows ? "pnpm.cmd" : "pnpm";
+
+    const result = await execFileImpl(file, allArgs, {
+      cwd,
+      maxBuffer,
+      shell: isWindows,
```
mikeharder: this may trigger security reviews.  we'll need either custom code similar to what we had before, or use a library like `cross-spawn`.

.github/shared/src/exec.js
```diff
@@ -81,54 +80,57 @@ export async function execFile(file, args, options = {}) {
...
+ * Calls `pnpm exec` with the given arguments.
  *
  * @param {string[]} args
  * @param {ExecNpmOptions} [options]
  * @returns {Promise<ExecResult>}
  * @throws {ExecError}
  */
 export async function execNpmExec(args, options = {}) {
-  return await execNpm(["exec", "--no", "--", ...args], options);
+  return await execNpm(["exec", ...args], options);
```
mikeharder: it's important to retain the logic to "fail fast if anything not installed", rather than "install deps as needed".  does "pnpm exec" do this by default?

.github/shared/src/exec.js
```diff
@@ -81,54 +80,57 @@ export async function execFile(file, args, options = {}) {
...
 /**
- * Calls `execFile()` with appropriate arguments to run `npm` on all platforms
+ * Calls `execFile()` with appropriate arguments to run `pnpm` on all platforms
  *
  * @param {string[]} args
  * @param {ExecNpmOptions} [options]
  * @returns {Promise<ExecResult>}
  * @throws {ExecError}
  */
 export async function execNpm(args, options = {}) {
```
mikeharder, line 92: maybe better to leave the existing helpers, and add new ones named `execPnpm()`?  can existing tools use `npm` after `pnpm install`?

.github/shared/package.json
```diff
@@ -72,5 +72,6 @@
     "perf": "node perf/perf.js",
     "test": "vitest",
     "test:ci": "vitest run --coverage --reporter=verbose"
-  }
+  },
+  "version": "0.0.0"
```
mikeharder: move to top of file

eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml
```diff
@@ -260,8 +260,13 @@ stages:
           echo "$sdk_gen_info"
 
           cd $(SpecRepoDirectory)
-          echo "##[group]Run npm ci"
-          npm ci
+          echo "##[group]Run pnpm install"
+          npm install -g pnpm@11
```
mikeharder: can we avoid global install, and also pin specific version?

package.json
```diff
@@ -1,10 +1,17 @@
 {
   "name": "azure-rest-api-specs",
+  "packageManager": "pnpm@11.5.1",
   "devDependencies": {
     "@autorest/openapi-to-typespec": "0.11.14",
```
mikeharder, line 8: would it be cleaner to put **every** package version in pnpm-workspace.yaml, for a single place to view and update all dep versions?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,45 @@
+packages:
+  - ".github"
+  - ".github/shared"
+  - "eng/tools"
+  - "eng/tools/*"
+
+# Lowered from the pnpm 11 default of 1440 (24h) to unblock recently published
+# dependencies. Packages must be at least this many minutes old before install.
+minimumReleaseAge: 720
```
mikeharder, line 9: i wonder if this should be 7 days to align with default CFS quarantine?  though we do need to ingest security updates from dependabot fresher than 7 days.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+allowBuilds:
+  "@actions/github-script": true
```
mikeharder: let's investigate why and whether we need this

.npmrc
```diff
@@ -1 +1,2 @@
 engine-strict=true
+dedupe-peer-dependents=false
```
mikeharder: @copilot: explain this setting, and why we might want to set it to false

.devcontainer/devcontainer.json
mikeharder:
```suggestion
  "postCreateCommand": "npm install -g pnpm@11 && pnpm ci",
```

let's use exactly `pnpm ci` as a replacement for `npm ci` in all docs and scripts

.github/actions/setup-node-install-deps/action.yaml
```diff
@@ -1,26 +1,33 @@
...
+description: Uses specified Node version and installs dependencies using pnpm
 
 inputs:
   node-version:
     description: "Node version to use"
     default: 24.x
   install-command:
     description: "Command to install dependencies"
-    default: "npm ci --no-audit --no-fund --prefer-offline"
+    default: "pnpm install --frozen-lockfile"
```
mikeharder:
```suggestion
    default: "pnpm ci --prefer-offline"
```

.github/instructions/eng-tools.instructions.md
```diff
@@ -41,7 +41,7 @@ The top-level `eng/tools` directory holds shared configuration that the individu
 - **Testing**: Vitest for unit and integration tests
 - **Linting**: ESLint (flat config) with TypeScript-aware rules
 - **Formatting**: Prettier with the organize-imports plugin (shared config)
-- **Package Manager**: npm (`npm ci` for clean installs)
+- **Package Manager**: pnpm workspaces (`pnpm install --frozen-lockfile` for clean installs)
```
mikeharder:
```suggestion
- **Package Manager**: pnpm workspaces (`pnpm ci` for clean installs)
```

.github/instructions/github-actions.instructions.md
```diff
@@ -33,7 +33,7 @@ The `.github` directory contains all the code and configuration for GitHub Actio
 - **Testing**: Vitest for unit and integration tests
 - **Linting**: ESLint with TypeScript-aware rules
 - **Formatting**: Prettier with organize-imports plugin
-- **Package Manager**: npm (npm ci for clean installs)
+- **Package Manager**: pnpm workspaces (`pnpm install --frozen-lockfile` for clean installs)
```
mikeharder:
```suggestion
- **Package Manager**: pnpm workspaces (`pnpm ci` for clean installs)
```

.github/instructions/github-actions.instructions.md
```diff
@@ -304,19 +303,19 @@ Scripts in `.github/workflows/src/` are typically used with `actions/github-scri
 ```bash
 # Install dependencies
 cd .github
-npm ci
+pnpm install
```
mikeharder, line 306:
```suggestion
pnpm ci
```

.github/instructions/sdk-generation.instructions.md
```diff
@@ -4,7 +4,7 @@ excludeAgent: "code-review"
 
 ## Prerequisites
 
-- run `npm ci` to install the dependencies
+- run `pnpm install` to install the dependencies
```
mikeharder, line 7:
```suggestion
- run `pnpm ci` to install the dependencies
```

.github/instructions/typespec-project.instructions.md
```diff
@@ -287,10 +287,10 @@ Users can convert a specification from swagger to typespec by using `tsp-client`
 1. Install the dependencies specified in the package.json at the root of this repository. Command:
 
 ```
-npm ci
+pnpm install --frozen-lockfile
```
mikeharder:
```suggestion
pnpm ci
```

.github/instructions/typespec-project.instructions.md
```diff
@@ -287,10 +287,10 @@ Users can convert a specification from swagger to typespec by using `tsp-client`
...
 ```
 
-2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `npx tsp-client convert --swagger-readme <path to your readme>`
+2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `pnpm exec tsp-client convert --swagger-readme <path to your readme>`
```
mikeharder:
```suggestion
2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `pnpx tsp-client convert --swagger-readme <path to your readme>`
```

.github/instructions/typespec-project.instructions.md
```diff
@@ -312,9 +312,9 @@ steps:
 ### Additional considerations
 
 - DO ensure you pull in the latest `main` from the Azure/azure-rest-api-specs repo to stay up to date with latest dependencies
-- DO run `npm ci` to get a clean install of the package.json dependencies
-- Avoid modifying the package.json or package-lock.json files at the root of the azure-rest-api-specs repo
-- Avoid adding your own package.json or package-lock.json files in your project directory
+- DO run `pnpm install --frozen-lockfile` to get a clean install of the package.json dependencies
```
mikeharder:
```suggestion
- DO run `pnpm ci` to get a clean install of the package.json dependencies
```

pnpm-workspace.yaml
timotheeguerin: yeah, because that one can't discriminate between type of release it might be too harsh, there is some other options like adding certain version to exclusion but this is something we'd have to remember to cleanup when they go out of the 7 day limit. Maybe could setup automation. 

I would preferably though move that as a follow up to keep this as much as possible as just getting parity, filed https://github.com/Azure/azure-rest-api-specs/issues/44274

.github/actions/setup-node-install-deps/action.yaml
timotheeguerin: yeah I think a lot of this setup kinda junky right now with all the various workflow and leftover from before. But right after this I can migrate it to a more solid single pipeline

.github/actions/setup-node-install-deps/action.yaml
timotheeguerin: didn't actually know pnpm ci was a thing , looks like its new to 11, that is nicer https://pnpm.io/cli/ci

.github/shared/src/exec.js
timotheeguerin: seems to be the case
image


> "Fetches a package from the registry without installing it as a dependency, hotloads it, and runs whatever default command binary it exposes." — [pnpm.io/cli/dlx](https://pnpm.io/cli/dlx)

| Command | Missing package behavior |
|---------|--------------------------|
| `pnpm exec <cmd>` | **Fails** — runs only from installed deps (`node_modules/.bin`) |
| `pnpm dlx <cmd>` | **Installs** temporarily, then runs (like `npx`) |

Docs: https://pnpm.io/cli/exec and https://pnpm.io/cli/dlx

.npmrc
timotheeguerin: ok seems like we don't need that anymore I think, might have been added earlier when I was trying to fix various calls to sub packages tools

.github/instructions/typespec-project.instructions.md
timotheeguerin: So I was actually wrong `pnpx` is not the same as `npx` it is a short for `pnpm dlx` which will only download a package, pnpm does have a stronger split between the 2 uses. However `pnpm <command>` just works

.devcontainer/devcontainer.json
timotheeguerin: we kinda want pnpm to be available globally after for running pnpm commands

timotheeguerin: but I pinned to the same version as in pacakgeManager

.github/actions/install-deps-github-script/action.yaml
timotheeguerin: added `--filter`

timotheeguerin: Image
locally goes from 7s to 1s

pnpm-workspace.yaml
timotheeguerin: added  comment on why it is needed

package.json
timotheeguerin: Yes I think ultimately it would, forgot why I decided to keep it to just the shared ones for now. I can do that as a follow up unless you prefer all done here

.github/instructions/typespec-project.instructions.md
mikeharder: My understanding, in pnpm v11, the primary name for the command is now `pnx`.  `pnpx` and `pnpm dlx` are both listed as aliases, and I assume behave exactly the same, with the same cli args.  So the question is just which of the three aliases we should use in dev docs or scripts.

I lean towards `pnpx`, because it follows the pattern of prefixing `npm` and `npx` with `p`.  And `pnpx` has been supported for multiple releases, while `pnx` is new in v11.

But I could entertain an argument to go straight to `pnx`, if this is the likely canonical name going forward.

package.json
```diff
@@ -1,10 +1,20 @@
 {
   "name": "azure-rest-api-specs",
+  "packageManager": "pnpm@11.5.1",
```
```suggestion
  "packageManager": "pnpm@11.8.0",
```

.github/actions/install-deps-github-script/action.yaml
mikeharder: nit: when benchmarking this scenario you should also delete the pnpm cache to simulate a cold agent that needs to download the packages.  even if we're using github actions caching, the perf is typically not much faster than direct download.

.github/actions/setup-node-install-deps/action.yaml
```diff
@@ -1,26 +1,30 @@
...
   node-version:
     description: "Node version to use"
     default: 24.x
   install-command:
     description: "Command to install dependencies"
-    default: "npm ci --no-audit --no-fund --prefer-offline"
+    default: "pnpm ci --prefer-offline"
   working-directory:
-    description: "Working directory"
+    description: 'Working directory. Should point at the workspace/checkout root (the location of the root package.json with the "packageManager" field), which is also where dependencies are installed from.'
```
mikeharder, line 12: if this needs to point at repo root, this is breaking the previous behavior.  and we could just remove the parameter and hardcode repo root.

i'm still concerned about the perf of the github actions scenario.  before, we had a separate package-lock.json under .github.  this was another lockfile to manage (bad), but it also let us have a separate cache for the lockfile (good).

in this PR, even though we install with `--filter`, we still need to restore the pnpm cache for the whole repo, which may be too expensive.

if we don't want to maintain a second pnpm-lock.yaml, perhaps we could just disable the github pnpm cache for the actions case?

.github/actions/setup-node-install-deps/action.yaml
```diff
@@ -1,26 +1,30 @@
...
+    - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+      with:
+        package_json_file: ${{ inputs.working-directory }}/package.json
+
     - uses: actions/setup-node@v6
       with:
         node-version: ${{ inputs.node-version }}
-        cache: "npm"
-        cache-dependency-path: ${{ inputs.working-directory }}/package-lock.json
+        cache: "pnpm"
```
mikeharder: i don't think the cache is working:

https://github.com/Azure/azure-rest-api-specs/actions/runs/28414200032/job/84193474290?pr=43922#step:4:64

it's only 6MB, and later in the log you see it still needed to download most packages.

i think I know what happened.  a workflow that does a partial install of just .github finished first, and since they use the same cache key, it won the race.

so we do need either separate caches (and lockfiles), or disable caching for the .github scenarios

.github/instructions/typespec-project.instructions.md
```diff
@@ -287,10 +287,10 @@ Users can convert a specification from swagger to typespec by using `tsp-client`
 1. Install the dependencies specified in the package.json at the root of this repository. Command:
 
 ```
-npm ci
+pnpm ci
 ```
 
-2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `npx tsp-client convert --swagger-readme <path to your readme>`
+2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `pnpm tsp-client convert --swagger-readme <path to your readme>`
```
mikeharder, line 293:
```suggestion
2. `tsp-client` is installed as part of the dependencies specified at the root of this repository. To convert a swagger to typespec, run the following command: `pnpm exec tsp-client convert --swagger-readme <path to your readme>`
```

.github/shared/src/exec.js
```diff
@@ -122,13 +123,109 @@ export async function execNpm(args, options = {}) {
...
+ * Calls `execPnpm()` with arguments ["exec", ...] prepended.
  *
  * @param {string[]} args
  * @param {ExecNpmOptions} [options]
  * @returns {Promise<ExecResult>}
  * @throws {ExecError}
  */
-export async function execNpmExec(args, options = {}) {
-  return await execNpm(["exec", "--no", "--", ...args], options);
+export async function execPnpmExec(args, options = {}) {
```
mikeharder, line 241: keep both the `npm` and `pnpm` functions.  this code is intended to be shared across repos that may use either.

.github/shared/src/exec.js
```diff
@@ -122,13 +123,109 @@ export async function execNpm(args, options = {}) {
...
+ * batch shim can be resolved and launched safely on Windows without enabling a
+ * shell. Enabling a shell (e.g. `shell: true`) would reintroduce quoting and
+ * shell-parsing risks, so it is intentionally avoided.
+ *
+ * @param {string[]} args
+ * @param {ExecNpmOptions} [options]
+ * @returns {Promise<ExecResult>}
+ * @throws {ExecError}
+ */
+export async function execPnpm(args, options = {}) {
```
mikeharder, line 150: this still seems way too complex.  i want to try a version similar to the way we invoke npm.

.github/workflows/github-test.yaml
```diff
@@ -54,13 +54,12 @@ jobs:
         with:
           # actions/github-script@v8 uses Node 24
           node-version: 24.x
-          working-directory: ./${{ matrix.folder }}
```
mikeharder, line 87: related to the earlier discussion about working-directory

.github/package.json
```diff
@@ -1,25 +1,29 @@
...
   "private": true,
   "type": "module",
   "_comments": {
     "dependencies": "Runtime dependencies must be kept to an absolute minimum for performance, ideally with no transitive dependencies",
     "dependencies2": "All runtime and dev dependencies in this file, must be a superset of shared/package.json"
   },
   "dependencies": {
-    "@apidevtools/json-schema-ref-parser": "^15.1.3",
+    "@apidevtools/json-schema-ref-parser": "catalog:",
+    "@azure-tools/specs-shared": "workspace:*",
```
mikeharder, line 12: before, .github source-included the files under .github/shared, rather than a package reference.  this allows code under .github/workflows to use code from .github/shared, without needing to install any deps, for perf.

documentation/ci-fix.md
```diff
@@ -190,13 +190,13 @@ cd <local_repo_clone_root>
 cd specification/contosowidgetmanager
 
 # Install the dependencies to the local 'node_modules' folder.
-npm install
+pnpm install
 
 # Run 'prettier --check' to verify the problems can be reproduced locally
-npx prettier --check **/*.json
+pnpm prettier --check **/*.json
```
mikeharder, line 197:
```suggestion
pnpm exec prettier --check **/*.json
```

documentation/ci-fix.md
```diff
@@ -190,13 +190,13 @@ cd <local_repo_clone_root>
...
 
 # Run 'prettier --write' to fix the problems.
-npx prettier --write **/*.json
+pnpm prettier --write **/*.json
```
mikeharder, line 200:
```suggestion
pnpm exec prettier --write **/*.json
```

.github/package.json
timotheeguerin: so without that the build started failing because it didn't finds `debug` and other deps

timotheeguerin: this feels like we are going through so many hoops here to get a little better perfs, would it be better to commit the bundled JS into the repo instead?

.github/actions/setup-node-install-deps/action.yaml
timotheeguerin: I think it is still needed for the summarize impact which clone in `before` and `after` folder

.github/shared/src/exec.js
timotheeguerin: I think kinda needs to be about that(definitely a little more  robust than it needs to be) but as it uses `cross-spawn` it needs the wrapper around to make it `exec` like

documentation/ci-fix.md
timotheeguerin: While I agree for robustness we can make sure our scripts use the long form, it feels a bit extreme to document that. I never hit any issues with this and as we control the version of pnpm it won't be a oh it fails on my machine

.github/instructions/typespec-project.instructions.md
mikeharder: actually, the closest to `npx` is `pnpm exec`.

.github/shared/src/exec.js
mikeharder: i will take a shot at this today

.github/package.json
mikeharder: all the sources are already in the repo so bundling seems unnecessary.  the nested projects were a bit unusual but worked fine with npm.  I don't see why pnpm can't support the same.

documentation/ci-fix.md
mikeharder: i guess i'm fine documenting `pnpm <command>` for dev machines, though I find it unusual for pnpm to mix it's own commands with external commands this way.

.github/actions/install-deps-github-script/action.yaml
```diff
@@ -9,9 +9,14 @@ runs:
...
+        # --config.trust-lockfile=true trusts the already-vetted committed lockfile and
+        # skips pnpm's per-install supply-chain re-verification, which is cold on every CI
+        # run and scales with total lockfile size (~2s here). It must be a CLI flag scoped
+        # to this command: pnpm does not read trustLockfile from the npm_config_* env, and
+        # the full installs in other workflows should keep verifying.
+        install-command: "pnpm install --frozen-lockfile --prod --prefer-offline --config.trust-lockfile=true --filter=azure-rest-api-specs-github..."
+        # Minimal prod-only install (~22 packages). Skip the pnpm store cache: in CI
```
mikeharder, line 18: This is fine for now, but another factor is reliability.  If the cache is more reliable, it might be better overall, even if slower (and how much slower?).

.github/shared/src/exec.js
mikeharder: Actually, I'm not sure I can do better, because `npm` is built-in while `pnpm` is not.

mikeharder (approved):

PR #43940, opened 2026-06-12
Add API version 2026-07-01-preview and resource type KubernetesResources in KubernetesConfiguration RP

bgriddaluru:
Add API version 2026-07-01-preview and resource type KubernetesResources in KubernetesConfiguration RP

## Overview
This PR adds support for the Kubernetes Inventory feature in Kubernetes Configuration RP to project the Kubernetes resources in Azure.

1. An in-cluster agent gathers K8s inventory from the Kubernetes API server and sends to Kubernetes Configuration RP, which publishes it to ARG through ARN.
2. Users can view the resources in the Azure portal or query it directly from ARG.

## Notes
1. KubernetesResources  is a read-only resource. So the `provisioningState` field is intentionally omitted.

## Modelling Considerations

1. There are over 40 different native Kubernetes Resource Types.  After careful consideration, it was decided to define single resource type to represent all these K8s types due to following reasons
    a. Base K8s resource structure is fixed i.e. api-version, kind, spec, status. 
    b. Many of the fields are common across resource types. So, defining separate types will cause maintenance overhead and duplication of fields
    c. In K8s RBAC is applied at namespace level and not at resource type.  So, there is no advantage in defining separate resource type in azure.


specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/KubernetesResource.tsp
```diff
@@ -0,0 +1,94 @@
...
+using TypeSpec.Rest;
+using Azure.Core;
+using Azure.ResourceManager;
+using Azure.ResourceManager.Foundations;
+using TypeSpec.Http;
+
+namespace Microsoft.KubernetesConfiguration;
+
+/**
+ * Singleton container resource grouping cluster inventory reports.
```
ravimeda: **[NEW] 🟠 Warning** **[[Doc copy-paste](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `KubernetesResource.tsp` — line 15 (also lines 78, 84) — The resource is documented as a **singleton container**, but is modeled, named, and exemplified as **one ARM resource per Kubernetes object**.

**Classification reasoning:** [NEW] — this resource type is introduced in this PR; it did not exist in any previous version.

Line 15: _"Singleton container resource grouping cluster inventory reports."_; line 78: _"Get the kubernetesResources singleton for a cluster."_; line 84: _"List kubernetesResources containers for the cluster (currently always a single 'default')."_ This contradicts `models.tsp` `KubernetesResourceProperties` (_"Each ARM resource represents one Kubernetes object… The ARM resource name encodes the kind, Kubernetes name, and namespace, e.g. 'pod-nginx-6d4cf56db6-x7q2p-default'"_). The path uses `{kubernetesResourceName}` with an open name pattern and the examples use per-object names — confirming the per-object model. The "singleton" framing will mislead SDK and docs consumers.

**Suggested fix:** Update the model / `get` / `list` doc comments to describe a per-Kubernetes-object resource consistently with `KubernetesResourceProperties`.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/operations/preview/2026-07-01-preview/operations.json
```diff
@@ -0,0 +1,275 @@
...
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "../../../../../../common-types/resource-management/v3/types.json#/definitions/ErrorResponse"
+            }
+          }
+        },
+        "x-ms-examples": {
+          "BatchAccountDelete": {
```
ravimeda: **[EXISTING] 🟠 Warning** **[[Doc copy-paste / example key](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `operations.json` — line 77 — The `Operations_List` operation''s `x-ms-examples` entry is keyed `"BatchAccountDelete"` (a copy-paste artifact from a BatchAccount template).

**Classification reasoning:** [EXISTING] — verified byte-identical in the prior `operations/stable/2025-04-01/operations.json` at the base SHA (also line 77). Pre-existing technical debt carried forward from the converted Swagger, not introduced by this PR.

**Suggested fix (optional):** Rename the example key in the operations TypeSpec source, e.g. `"List KubernetesConfiguration operations"`.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/models.tsp
```diff
@@ -0,0 +1,1739 @@
...
+
+// ─── ARM resource properties ──────────────────────────────────────────────────
+
+/**
+ * Properties of a single KubernetesResources ARM resource.
+ * Each ARM resource represents one Kubernetes object from the cluster inventory.
+ * The ARM resource name encodes the kind, Kubernetes name, and namespace
+ * (e.g., 'pod-nginx-6d4cf56db6-x7q2p-default', 'upgrade-assessment-current-kube-system').
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "This resource does not have a provisioning state per the original Swagger design"
```
ravimeda: **[NEW] 🔵 Suggestion** **[[Suppression accuracy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `models.tsp` — line 1706 — The `arm-resource-provisioning-state` suppression reason is factually inaccurate.

**Classification reasoning:** [NEW] — this suppression is added in this PR for a brand-new resource.

The reason states _"This resource does not have a provisioning state per the original Swagger design"_ — but this is a brand-new TypeSpec resource with **no original Swagger**. The suppression itself is valid (a read-only GET/LIST-only resource has no provisioning lifecycle); only the reason is off.

**Suggested fix:** Reword, e.g. _"Read-only inventory resource exposing only GET/LIST; it has no create/update lifecycle, so provisioningState does not apply."_

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`3d1d378`](https://github.com/Azure/azure-rest-api-specs/pull/43940/commits/3d1d3784a6fcc886a0a9bca7a07f21cf2ba36adc). See inline comments for findings 1-4. No blocking issues were found.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/models.tsp
```diff
@@ -0,0 +1,1739 @@
...
+
+  // ── ConfigMap ──────────────────────────────────────────────────────────────
+
+  /**
+   * Key-value pairs stored in the ConfigMap (e.g., configuration files, environment values).
+   * Each entry maps a string key to a string value.
+   * Populated for ConfigMap resources.
+   */
+  @identifiers(#["name"])
+  configMapData?: KubernetesLabel[];
```
ravimeda: **[NEW] 🔵 Suggestion** **[[Security review / SEC-SECRET-DETECT-adjacent](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#3-secret-detection-in-typespec-sec-secret-detect)]** `models.tsp` — line 1108 (and line 1115) — Full ConfigMap values are projected to Azure Resource Graph; confirm with security.

**Classification reasoning:** [NEW] — introduced in this PR with the new resource type.

`configMapData` (line 1108) projects **full ConfigMap key+value pairs** into Azure Resource Graph (queryable by any subscription Reader), whereas `configMapBinaryDataKeys` (line 1115) deliberately stores **keys only**. ConfigMaps routinely hold connection strings and internal endpoints — the ConfigMap example projects `DATABASE_HOST: postgres.default.svc.cluster.local`. No hard rule fires (a ConfigMap is the non-secret counterpart to a Secret in Kubernetes, and Secret `data` is correctly excluded), so this is **non-blocking**, but the asymmetry warrants explicit security sign-off. Related advisory points for the same confirmation:
- `secretName` (TLS, line 726) and `imagePullSecrets` (line 795) hold Secret **names only** but carry neither `@secret` nor a `secret-prop` suppression (only `secretRef` at line 261 is suppressed). Confirm the `secret-prop` linter is clean on those names.
- `EnvVar.value` (line 251) projects literal container env values; a hardcoded-secret-in-env (Kubernetes anti-pattern) would reach ARG.

**Suggested fix:** Confirm with security whether full ConfigMap values (vs. keys-only) and literal env values should be projected to ARG; align the `secret-prop` suppression coverage across the name-only Secret references.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/KubernetesResource.tsp
bgriddaluru: Thanks. Removed the singleton reference

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/operations/preview/2026-07-01-preview/operations.json
bgriddaluru: Fixed the description to List operations.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/models.tsp
bgriddaluru: Fixed the suppression reason as suggested.

specification/kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/kubernetesResources/models.tsp
bgriddaluru: Client publishing the data will selectively skip some fields or values if it detects as secret data. Config map will ideally hold public certs or configuration data, which are generally safe to project. We already projecting such info via another workflow and giving visibility to users in portal. So resolving this comment.

dipti-pai (approved):

PR #43969, opened 2026-06-15
Use package-root octokit type imports and rely on github-script transitive deps

timotheeguerin:
Precursor PR extracted from the npm→pnpm migration to keep that PR focused.

Replaces deep `@octokit/plugin-rest-endpoint-methods/dist-types/types.js` JSDoc imports with the package-root import, and relies on `@actions/github-script` transitive dependencies for the related Octokit/Actions types instead of adding new direct devDependencies in `.github`. Lockfile regenerated; `lint:tsc` passes.


.github/package.json
```diff
@@ -16,9 +16,15 @@
     "zod": "^4.3.5"
   },
   "devDependencies": {
+    "@actions/core": "^1.11.1",
```
mikeharder, line 19: @copilot: I believe these new devDependencies are already dependencies of github-script, and we want to rely on github-script to define the versions used.  Is this valid, or should we add direct devDeps if we need to import from the package in JS doc comments?

mikeharder (approved):

PR #215, opened 2016-03-30
Move Batch swagger specification to right place

matthchr:

PR #43970, opened 2026-06-15
Add voice-first prompt agents on a dedicated /voice-agents path to the Foundry data-plane spec

yulin-li:
## Summary

Adds **voice agents** to the Foundry data-plane API on a dedicated **`/voice_agents`** path family. Voice agents are configured once (model selection, instructions, audio stack, tools, optional avatar) and drive a managed speech-to-speech experience.

Voice-agent management is self-contained under `/voice_agents`: the `/agents` CRUD surface neither accepts nor returns voice agents. Non-CRUD endpoint protocol resources, including persisted conversations and audio, remain under `/agents/{agent_name}/endpoint/...`. The realtime WebSocket connect operation is intentionally deferred.

The entire surface is additive and gated by `VoiceAgents=V1Preview`.

## Voice agent definition

`VoiceAgentDefinition` is a standalone definition (`kind: "voice"`); it is **not** a member of the text `AgentDefinition` union, and it does **not** introduce a value on `AgentKind`. It carries:

- required `model_type` (`managed` or `self_deployed`) and `model`; the architecture (realtime vs cascaded) is derived by the service
- `instructions`, `structured_inputs`, `audio`, `output_modalities`, optional `avatar`
- `store` (default `false`) — the single, all-or-nothing persistence switch
- `VoiceAgentTool` variants for `function`, `mcp`, `system`, and `toolbox`
- extensible `VoiceSystemToolName` values with stable known action `end_conversation`; system tools may also carry an optional description

## Voice agent management

`/voice_agents` mirrors the text agent lifecycle. Every create/update produces a new immutable version.

| Operation | Route |
| --- | --- |
| Create | `POST /voice_agents` |
| List | `GET /voice_agents` |
| Retrieve | `GET /voice_agents/{agent_name}` |
| Update | `POST /voice_agents/{agent_name}` |
| Delete | `DELETE /voice_agents/{agent_name}` |
| Enable | `POST /voice_agents/{agent_name}:enable` |
| Disable | `POST /voice_agents/{agent_name}:disable` |
| Create version | `POST /voice_agents/{agent_name}/versions` |
| List versions | `GET /voice_agents/{agent_name}/versions` |
| Retrieve version | `GET /voice_agents/{agent_name}/versions/{agent_version}` |
| Delete version | `DELETE /voice_agents/{agent_name}/versions/{agent_version}` |
| Generate | `POST /voice_agents:generate` |

`VoiceAgentObject` / `VoiceAgentVersionObject` reuse the shared agent shapes but always carry a `VoiceAgentDefinition`. Draft versions are supported: `draft` (default `false`) on create/version requests, and `include_drafts` (default `false`) on the version listing.

### Guided authoring

`POST /voice_agents:generate` (`generateVoiceAgent`) expands a few high-level inputs plus a natural-language `goal` into a full, editable `VoiceAgentDefinition`, creates the agent through the standard voice create path, and returns the created `VoiceAgentObject`. `GenerateVoiceAgentRequest` supplies `name`, `model_type`, `model`, `agent_type`, `use_case`, `goal`, optional `description` and `tools`, plus optional `draft` (default `false`).

## Realtime model reuse strategy

The persisted voice conversation contract reuses `@azure-tools/openai-typespec` Realtime models wherever the wire shape is compatible:

- system, user, and assistant message items with role-specific content constraints
- function-call and function-call-output items
- MCP list-tools, call, approval-request, and approval-response items
- Realtime response usage and status-details models
- Realtime audio-format models

Each persisted item leaf extends its upstream counterpart and adds only Foundry persistence fields such as `created_at` and `response_id` (with the optional function name retained on function-call output).

Foundry-local models are used for new or incompatible concepts:

- `VoiceConversation`, the durable persistence envelope and retention/delete boundary
- `VoiceResponse`, because the persisted contract uses `audio.output.voice` as a `{ type, name }` object rather than the upstream flat voice value
- `VoiceConversationItem`, a plain union of nine leaves because upstream message and non-message items currently use disconnected discriminator roots
- persisted item/recording audio metadata and content responses

The upstream role models enforce: system `input_text`; user `input_text | input_audio | input_image` (`image_url`, optional `detail`); assistant `output_text | output_audio`. They also preserve the upstream `type: "message"` and `object: "realtime.item"` semantics. Persisted timestamps use the shared `FoundryTimestamp` (`utcDateTime` encoded as an `int32` Unix timestamp). `VoiceConversation` exposes aggregate `OpenAI.RealtimeResponseUsage`, while each `VoiceResponse` retains per-response usage.

## Persisted voice conversation and audio APIs

The service is the sole writer of persisted responses/items, so this surface exposes reads plus cascading conversation deletion under:

`/agents/{agent_name}/endpoint/protocols/voice/conversations`

**11 routes:**

1. `GET /{conversation_id}` — get `VoiceConversation`
2. `DELETE /{conversation_id}` — delete the conversation and cascade to responses, items, and audio
3. `GET /{conversation_id}/responses` — list responses
4. `GET /{conversation_id}/responses/{response_id}` — get one response
5. `GET /{conversation_id}/responses/{response_id}/items` — list one response's output items
6. `GET /{conversation_id}/items` — list the complete ordered conversation history
7. `GET /{conversation_id}/items/{item_id}` — get one item
8. `GET /{conversation_id}/items/{item_id}/audio` — get item audio metadata
9. `GET /{conversation_id}/items/{item_id}/audio/content` — stream item audio
10. `GET /{conversation_id}/audio` — get merged stereo recording metadata
11. `GET /{conversation_id}/audio/content` — stream the merged stereo recording

The response-items list is a response-specific output projection; the conversation-items list is the complete ordered history, with `response_id` linking generated items back to their response.

## Persistence and audio delivery

- `store = false` (default): nothing is persisted and persisted-resource lookups return `404`
- `store = true`: conversation content and audio are persisted
- Foundry-managed audio streams through `/audio/content`
- BYOS metadata includes optional `blob_path`; `/audio/content` returns `409` because callers retrieve bytes from their own storage
- merged recording metadata/content returns `409` until the session has ended
- latency/performance telemetry (time-to-first-audio, inter-token latency, interruption) remains observability-only (trace / App Insights) and is not a persisted resource field

## Deferred

The realtime WebSocket upgrade at `GET /agents/{agent_name}/endpoint/protocols/voice` is deferred to a follow-up.

## Validation

- `npx tsp compile .` succeeds; generated OpenAPI is stable across recompilation
- `azsdk_run_typespec_validation` succeeds
- v1 and virtual-public-preview JSON/YAML artifacts regenerated and committed
- enable/disable emit as preview-gated POST actions with no request body and a `204` response
- confirmed `/voice_agents` contains the voice-agent management/version routes and `:generate`, while all 11 conversation/audio endpoint routes remain under `/agents/{agent_name}/endpoint/protocols/voice`

## References

- Experience/design spec: [`coreai-microsoft/foundrysdk_specs` PR #215](https://github.com/coreai-microsoft/foundrysdk_specs/pull/215)
- Source: `specification/ai-foundry/data-plane/Foundry/src/voice-agents/`


specification/ai-foundry/data-plane/Foundry/src/common/servicepatterns.tsp
```diff
@@ -548,5 +549,6 @@ op deleteJobPreview<AreaPreviewLabel extends
 // Types and traits shared by multiple service patterns
 // ============================================================================
 
-@encode(DateTimeKnownEncoding.unixTimestamp, int32)
+#suppress "@microsoft/azure-openapi-validator/IntegerTypeMustHaveFormat" "The unixtime format is present, but the legacy rule only allows int32 or int64."
+@encode(DateTimeKnownEncoding.unixTimestamp, int64)
```
trrwilson: I believe updating from 32- to 64-bit in-place would construe a subtle but meaningful breaking change for existing areas: clients already expecting and transacting on 32-bit numbers could break on receiving 64-bit values -- minimally after the 2038 epoch rollover, but potentially immediately with the contract change and `long`/`int` distinctions.

Suggested:
- If we can just piggyback on int32 encoding, that'll hold us over until 2038. And if we end up in the incredible situation that Y2K38 is looming, migration guidance could be handled platform-wide vs. partial migrations so far in advance.
- If we do need int64 encoding now, we should do this as a *parallel* `FoundryTimestamp64` or similar, preserving the 32-bit encoding of `FoundryTimestamp` for existing consumers.

yulin-li: thanks for the comments, I will revert to int32

yulin-li: Reverted FoundryTimestamp to int32 unixTimestamp encoding (kept the 32-bit contract; dropped the int64 change and its suppression). Thanks Travis.

trrwilson (approved):

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,526 @@
...
+  toolbox: VoiceToolboxTool,
+}
+
+@doc("A Voice Live built-in control (system) tool that acts on the live session itself, e.g. ending the call. Executed natively by Voice Live; no customer code or external auth.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceSystemTool {
+  @doc("The type of the tool. Always `system`.")
+  type: "system";
+
+  @doc("The name of the built-in control action, e.g. `end_conversation`.")
```
rhurey: How static is the set of tools? And will customers be able to depend on the list not shifting?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,526 @@
...
+// the voice surface stays in sync with the text surface.
+// ============================================================================
+
+@doc("A voice-first prompt agent version. Mirrors `AgentVersionObject`, but its `definition` is always a `PromptVoiceAgentDefinition`.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceAgentVersionObject {
+  ...OmitProperties<AgentVersionObject, "definition" | "error">;
+
+  @doc("The voice-first prompt agent definition for this version.")
+  definition: PromptVoiceAgentDefinition;
```
rhurey: Nit: Thought we were going to drop Prompt.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
rhurey, line 1: There are ~35 references to Voice Live in this file, which may be an implementation detail leak.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,432 @@
...
+namespace Azure.AI.Projects;
+
+// ============================================================================
+// Persisted voice conversation resources.
+//
+// These models describe the durable, read-only projection of a voice-first
+// prompt agent's conversation. They follow the Azure Voice Live 2026-04-10
+// realtime schema (conversation / response / item / usage), NOT the OpenAI
+// Responses/Conversations resources. Only fields that are safe and meaningful
+// for a persisted customer contract are exposed; internal storage identifiers
```
rhurey: Why are the internal things that are not listed even mentioned as existing?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
yulin-li: good catch, updating it

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
yulin-li: thanks, update with an enum, currently only `end_conversation` is in this list

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
yulin-li: thanks, cleared them

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
yulin-li: thanks, cleared

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,429 @@
...
+  cancelled: "cancelled",
+  incomplete: "incomplete",
+  failed: "failed",
+}
+
+@doc("The voice used for a response, for example `{ \"type\": \"openai\", \"name\": \"alloy\" }`.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceResponseVoice {
+  @doc("The voice kind, e.g. `openai` or an Azure voice type.")
+  type: string;
```
rhurey: The type and the doc comment don't align. Does this support the full range of voice models?

yulin-li: Addressed in b3104864f21. VoiceResponseVoice.type now uses an extensible VoiceResponseVoiceType union with openai plus every supported Azure voice kind: azure-standard, azure-custom, azure-personal, avatar-voice-sync, and azure-realtime-native. The string fallback preserves forward compatibility. I also clarified that this response model contains the resolved voice identity. The root TypeSpec and the .NET, JavaScript, and Python Voice Agents client views compile successfully.

yulin-li: Follow-up in e90789a6d4b: voice is now modeled as a full polymorphic union rather than only an extensible type enum. The union covers explicit OpenAI, Azure standard, custom, personal, avatar voice sync, and Azure realtime-native configurations with their public fields and constraints. Avatar voice sync intentionally requires type and model and has no name field. Agent audio.output.voice reuses the same complete models, while still accepting a bare OpenAI voice name.

yulin-li: Follow-up in 0d1888b0d70: removed the voice schema dependency on OpenAI.VoiceIdsShared. VoiceAgentVoice and OpenAIVoice now use the local extensible OpenAIVoiceName union.

yulin-li: Follow-up in 1d1d37dab19: removed PhoenixLatestNeural and PhoenixV2Neural from the known PersonalVoiceModel values.

yulin-li: Follow-up in 283b2c0de0a: kept OpenAI.VoiceIdsShared for OpenAIVoice.name, removed the local OpenAIVoiceName union, and removed the bare-string member from VoiceAgentVoice.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+interface AgentEndpointConversations {
+  /**
+   * Retrieves a single conversation recorded for the specified voice agent endpoint by its id.
+   * Returns `404` when the conversation was not persisted (`store = false`) or does not exist.
+   */
+  @summary("Get a voice agent conversation")
+  @get
+  @route("/agents/{agent_name}/endpoint/protocols/voice/conversations/{conversation_id}")
+  @extension("x-ms-foundry-meta", PromptVoiceConversationRequiredPreviews)
+  getAgentConversation is FoundryDataPlanePreviewOperation<
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. All 23 voice-agent and voice-conversation operations now use FoundryDataPlaneRequiredPreviewOperation, and the emitted Foundry-Features header is required.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+/**
+ * Read-only, endpoint-scoped retrieval of the conversations, responses, transcript items, per-turn
+ * metrics, and audio recorded by a voice agent. For a `voice` agent, the service is the sole writer of
+ * responses and items, so only
+ * reads plus a cascading conversation delete are exposed — there are no response/item create, update, or
+ * delete routes. All persistence is governed by the definition's `store` flag: when `store = false` (the
+ * default) nothing is persisted, so these routes return `404`. The merged-recording routes additionally
+ * require the session to have ended and return `409` while it is still in progress.
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. Both suppressions now provide specific rationale: endpoint-scoped read-only conversation projections and dedicated voice-agent create/version semantics do not match the standard resource operation templates.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,531 @@
...
+  rai_config?: RaiConfig;
+
+  @doc("How the model backing this agent is served. Together with `model`, this selects the model up front. `managed` uses a service-managed model; `self_deployed` uses the customer's own Foundry deployment. This is independent of the architecture (realtime or cascaded), which the service derives from the selected model.")
+  model_type: VoiceModelType;
+
+  @doc("The model to use for this agent, paired with `model_type`: the service-managed model name when `model_type` is `managed`, or the customer's Foundry deployment name when `model_type` is `self_deployed`. The model must support realtime or cascaded voice. The service derives the architecture from the selected model.")
+  `model`: string;
+
+  @doc("A system (or developer) message inserted into the model's context. Supports template substitution via `structured_inputs`, rendered per session before the live session starts.")
+  instructions?: string | null;
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3 for instructions: the optional property is now non-nullable and is omitted when unset. noise_reduction and turn_detection intentionally retain null because omission inherits the configured/default behavior while explicit null disables those OpenAI Realtime-compatible controls.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,531 @@
...
+    The audio configuration, including input and output formats, voice, turn detection, noise reduction, and
+    transcription. These values are session defaults; a client may override supported fields when connecting.
+    """)
+  audio?: VoiceAudioConfig;
+
+  @doc("""
+    The output modalities the agent produces. Defaults to `["audio"]`. `animation` and `avatar` are available
+    when an avatar is configured.
+    """)
+  output_modalities?: VoiceOutputModality[];
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. output_modalities now has the TypeSpec default #[VoiceOutputModality.audio], emitted as ["audio"].

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,531 @@
...
+  threshold?: float32;
+
+  @doc("Audio (ms) to include before detected speech start.")
+  prefix_padding_ms?: int32;
+
+  @doc("Silence (ms) used to detect speech stop.")
+  silence_duration_ms?: int32;
+
+  @doc("Whether the server automatically creates a response when a turn ends. Defaults to true.")
+  create_response?: boolean;
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. create_response now has the TypeSpec default true.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,429 @@
...
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceDeletedConversation {
+  @doc("The id of the deleted conversation.")
+  id: string;
+
+  @doc("The object type. Always `voice.conversation.deleted`.")
+  object: "voice.conversation.deleted";
+
+  @doc("Always `true` for a successful delete.")
+  deleted: boolean;
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. VoiceDeletedConversation.deleted is now the boolean literal true.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,429 @@
...
+  channels?: int32;
+
+  @doc("The offset from the session start at which this segment begins, in milliseconds.")
+  start_offset_ms?: int32;
+
+  @doc("The duration of the audio segment, in milliseconds.")
+  duration_ms?: int32;
+
+  @doc("For bring-your-own-storage (BYOS) recordings only: a direct path/URL into the customer's own storage, without a SAS token — the customer downloads it using their own storage credentials. Absent for Foundry-managed storage, where the bytes are streamed via the item's `/audio/content` route instead.")
+  blob_path?: string;
```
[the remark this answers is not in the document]
yulin-li: Addressed in a8f368f95d3. Both BYOS audio metadata models now expose blob_uri: url, and all route/model documentation uses the unambiguous URI representation.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  @doc("The role carried on the left channel, e.g. `user`.")
+  left: string;
+
+  @doc("The role carried on the right channel, e.g. `agent`.")
+  right: string;
+}
+
+@doc("""
+  Metadata for the merged, whole-call stereo recording of a voice conversation (user audio on the left channel,
+  agent audio on the right). Built once from the per-turn segments after the session ends and durably cached.
```
rhurey, line 397: This isn't consistent with having a model to determine what role in in what channel above.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+}
+
+@doc("The role assigned to each channel of a merged stereo voice recording.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceRecordingChannelLayout {
+  @doc("The role carried on the left channel, e.g. `user`.")
+  left: string;
+
+  @doc("The role carried on the right channel, e.g. `agent`.")
+  right: string;
```
rhurey, line 392: >string; [](http://example.com/codeflow?start=9&length=7)

While there's an example in the doc... If we expect the developer to consult this in code this should be some kind of a fixed value.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  item_id: string;
+
+  @doc("The role the audio belongs to, e.g. `user` or `agent`.")
+  role?: string;
+
+  @doc("The container format of the audio, e.g. `wav`.")
+  format?: string;
+
+  @doc("The audio codec, e.g. `pcm16`.")
+  codec?: string;
```
rhurey, line 355: >string [](http://example.com/codeflow?start=10&length=6)

How does a developer find the supported codecs?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  own storage credentials. For Foundry-managed storage `blob_uri` is absent and the bytes are streamed via the
+  `/audio/content` route instead.
+  """)
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceRecordingResponse {
+  @doc("The id of the conversation this recording belongs to.")
+  conversation_id: string;
+
+  @doc("The container format of the recording, e.g. `wav`.")
+  format: string;
```
rhurey, line 410: >string [](http://example.com/codeflow?start=10&length=6)

Here this is a string, with one sample, but when defining the agent a "VoiceAudioFormatType" extensible enum exists, why not use that?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,741 @@
...
+@discriminator("type")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceTurnDetection {
+  @doc("The turn detection strategy.")
+  type: VoiceTurnDetectionType;
+}
+
+@doc("The turn detection strategy. Values follow the OpenAI Realtime wire schema.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+union VoiceTurnDetectionType {
```
glecaros, line 251: should we use the OpenAI model instead of rolling a duplicate? or do we expect drift?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,741 @@
...
+model SemanticVadTurnDetection extends VoiceTurnDetection {
+  type: VoiceTurnDetectionType.semantic_vad;
+
+  @doc("How eagerly the model ends the user's turn.")
+  eagerness?: VoiceTurnDetectionEagerness;
+}
+
+@doc("How eagerly semantic turn detection ends the user's turn.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+union VoiceTurnDetectionEagerness {
```
glecaros, line 288: same as the above should we use the openai type

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+}
+
+@doc("The response returned when a voice conversation is deleted.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceDeletedConversation {
+  @doc("The id of the deleted conversation.")
+  id: string;
+
+  @doc("The object type. Always `voice.conversation.deleted`.")
+  object: "voice.conversation.deleted";
```
glecaros, line 67: we generally don't include object fields as they have not user for the customer.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+#suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Spans two disconnected OpenAI discriminator roots (type vs role); see note above."
+@doc("""
+  A persisted item in a voice conversation. A plain union of the 9 OpenAI Realtime item kinds
+  (message x3 roles, function_call, function_call_output, mcp_list_tools, mcp_call,
+  mcp_approval_request, mcp_approval_response), each extended with the Foundry persistence extensions
+  `created_at`/`response_id`. See the note above for why this is a plain union rather than a single
+  `@discriminator`-based type.
+  """)
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+union VoiceConversationItem {
```
glecaros, line 318: note that this needs to be discriminated or it won't emit... a workaroud for this (which is what has been done for responses), is to  update the upstream's client-emitter projection (shipped along with @azure-tools/openai-typespec) to unify the disjoint variants so it can be represented as a single hierarchy.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
glecaros: +1 if there's a known list of supported codecs, it should be an union.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  codec?: string;
+
+  @doc("The sample rate in Hz.")
+  sample_rate?: int32;
+
+  @doc("The number of audio channels.")
+  channels?: int32;
+
+  @doc("The offset from the session start at which this segment begins, in milliseconds.")
+  start_offset_ms?: int32;
```
glecaros, line 364:
```suggestion
  @encoding(DurationKnownEncoding.milliseconds)
  start_offset_ms?: duration;
```

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+
+  @doc("The duration of the audio segment, in milliseconds.")
+  duration_ms?: int32;
```
glecaros, line 367: same as above

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  contentType: "audio/wav";
+
+  @doc("The WAV audio bytes.")
+  @body
+  content: bytes;
+}
+
+@doc("The role assigned to each channel of a merged stereo voice recording.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+model VoiceRecordingChannelLayout {
```
glecaros, line 387: this  seems like hard  thing for the user, unless we ship a thick client that abstracts this

specification/ai-foundry/data-plane/Foundry/src/voice-agents/models.tsp
```diff
@@ -0,0 +1,426 @@
...
+  sample_rate: int32;
+
+  @doc("The number of audio channels. The merged recording is stereo (`2`).")
+  channels: int32;
+
+  @doc("The role assigned to each stereo channel.")
+  channel_layout: VoiceRecordingChannelLayout;
+
+  @doc("The total duration of the recording in milliseconds.")
+  duration_ms: int32;
```
glecaros, line 422: same as durations above

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+  deleteAgentConversation is FoundryDataPlaneRequiredPreviewOperation<
+    AgentDefinitionOptInKeys.voice_agents_v1_preview,
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The id of the conversation to delete. */
+      @path conversation_id: string;
+    },
+    VoiceDeletedConversation
```
glecaros, line 64: do we need the tombstone? if we don't see the customer using it, we should return 204

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -0,0 +1,741 @@
...
+  volume?: string;
+}
+
+@doc("The Azure synthesized voice kind. Additional values may be added over time.")
+@extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+union AzureVoiceType {
+  string,
+
+  @doc("An Azure standard neural voice.")
+  azure_standard: "azure-standard",
```
glecaros, line 368: our guidelines state  that wire values should be snake-cased

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+   */
+  @summary("Create a voice agent")
+  @post
+  @route("/voice_agents")
+  @extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+  @tag("Voice Agents")
+  createVoiceAgent is FoundryDataPlaneRequiredPreviewOperation<
+    AgentDefinitionOptInKeys.voice_agents_v1_preview,
+    CreateVoiceAgentRequest,
+    VoiceAgentObject
```
glecaros, line 313: should return a 201

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+  @route("/voice_agents/{agent_name}")
+  @extension("x-ms-foundry-meta", PromptVoiceAgentRequiredPreview)
+  @tag("Voice Agents")
+  deleteVoiceAgent is FoundryDataPlaneRequiredPreviewOperation<
+    AgentDefinitionOptInKeys.voice_agents_v1_preview,
+    {
+      /** The name of the voice agent to delete. */
+      @path agent_name: string;
+    },
+    DeleteAgentResponse
```
glecaros, line 386: if we don't envision the customer using the tombstone, it should r eturn a 204

specification/ai-foundry/data-plane/Foundry/src/voice-agents/routes.tsp
```diff
@@ -0,0 +1,536 @@
...
+  deleteVoiceAgentVersion is FoundryDataPlaneRequiredPreviewOperation<
+    AgentDefinitionOptInKeys.voice_agents_v1_preview,
+    {
+      /** The name of the voice agent to delete. */
+      @path agent_name: string;
+
+      /** The version of the voice agent to delete. */
+      @path agent_version: string;
+    },
+    DeleteAgentVersionResponse
```
glecaros, line 534: same as above, should return a 204

PR #43898, opened 2026-06-10
Add Policy ARM resource for CognitiveServices (2026-07-01-preview)

djetchev:
## Summary

Introduces a new **Policy** ARM resource type for Microsoft.CognitiveServices, mirroring the existing Connection resource model. This enables Foundry platform governance for egress (and future ingress) policies.

Policies are **account-scoped only**, consistent with RAI policies and networking in Foundry.

### Resource scope

| Scope | Path | Sharing |
|-------|------|---------|
| Account-level | `accounts/{name}/policies/{policyName}` | Shared across all projects |

### Changes

- **Policy.tsp**: New TypeSpec model defining `PolicyResource` with account-level CRUD operations
- **models.tsp**: Added `PolicyType`, `PolicyDefaultAction`, `PolicyRuleAction` extensible unions; `PolicyRuleMatch`, `PolicyRule`, `PolicyProperties` models
- **main.tsp**: Added `v2026_07_01_preview` API version and `Policy.tsp` import
- **examples/**: 4 example JSON files (account-level get, createOrUpdate, delete, list)
- **readme.md**: Added `package-2026-07-01-preview` tag
- **Generated swagger**: `cognitiveservices.json` for `2026-07-01-preview`

### Design decisions

- **Account-scoped only** — RAI policies and networking are both account-level in Foundry, so policies follow the same pattern
- Uses extensible unions (not enums) for all type fields
- Named generically as `Policy` (not `EgressPolicy`) to future-proof for ingress policies
- `provisioningState` with read-only visibility

### Related

- Pattern reference: PR #43169 (CognitiveServices preview API version)
- Design doc: [EgressPoliciesAsResources.md](https://github.com/coreai-microsoft/adc/blob/djetchev/egress-policies-as-resources/docs/arch-design/EgressPoliciesAsResources.md)

PR #43974, opened 2026-06-15
Add Policy ARM resource for CognitiveServices (2026-05-15-preview)

djetchev:
## Summary

Add account-level **Policy** ARM resource to CognitiveServices for controlling sandbox egress traffic. This PR targets the existing `2026-05-15-preview` API version branch per RP team guidance from Fareed.

### New files
- `Policy.tsp` — PolicyResource model + CRUD operations (get, createOrUpdate, delete, list)
- `examples/2026-05-15-preview/AccountPolicy/` — 4 example JSON files

### Modified files
- `main.tsp` — Added Policy.tsp import
- `models.tsp` — Added Policy unions and models:
  - `PolicyType` (Egress)
  - `PolicyMode` (Enforced, Audit)
  - `PolicyDefaultAction` (Allow, Deny)
  - `PolicyRuleAction` (Allow, Deny, Transform, Rewrite)
  - `PolicyRuleType` (Fqdn)
  - `PolicyRuleMatch`, `PolicyRule`, `PolicyProperties`, `PolicyListResult`

### Design decisions
- **Account-scoped only** — Aligns with RAI policies and networking (no project-level)
- **Extensible unions** — All enums use union pattern for future extensibility (e.g., MCP rule type)
- **Audit mode** — Supports enforcement mode toggle per Evan Mattson's implementation
- **Transform/Rewrite actions** — Included per Evan's request for header injection support
- **No methods field** — Removed per Evan's feedback (easier to add later if needed)

## Related PRs
- Supersedes #43898 (which incorrectly created a new 2026-07-01-preview version)
- Vienna backend: PR #2148420 on ADO


fmabroukmsft:
Model-design questions before this ships in preview:

**1. `Policy*` naming collides with existing types.** The base swagger already exposes `PolicyAssignmentEvaluationDetails`, `PolicyEvaluationOutcome`, and `PolicyExpressionEvaluationDetails` from `EvaluateDeploymentPolicies`. Generic `Policy*` overlaps in SDK auto-complete and docs. `SandboxPolicy*` or `FoundryPolicy*` would preserve the future-proofing for ingress without the overload.

**2. `PolicyRuleAction = Transform | Rewrite` doesn't apply to `PolicyRuleType = Fqdn`.** FQDN/DNS-level matching can't transform headers or rewrite paths. Either document that `Transform`/`Rewrite` are reserved for future rule types and have the server 400 them when combined with `ruleType=Fqdn`, or defer adding them until a rule type uses them. Commit `14a0d79` dropped them per Xi's guidance before `f47610d` re-added them per Evan's — confirm alignment.

**3. `PolicyMode` default is doc-only.** The doc says "Defaults to Enforced" but the schema has no `default`. Pin the server contract: on GET when the client omitted `mode` on PUT, does the response echo `"Enforced"` or `null`?

**4. `defaultAction` is optional and security-critical.** If unset and no rule matches, the fallback (`Allow` vs. `Deny`) materially changes the policy's behavior. Either make `defaultAction` required, or document the server-applied default explicitly.

**5. `PolicyRuleMatch` is all-optional and glob semantics are unspecified.** `PolicyRuleMatch {}` matches everything by absence — state this. For `host: "*.openai.com"` and `path: "/v1/*"`, pin the syntax (DNS wildcard for host; POSIX glob, URI glob with `**`, or something else for path).

**6. Name pattern `{2,32}` caps at 33 chars.** Short for an ARM child resource (typical 63). If a backend storage constraint drives this, note it; otherwise consider `{2,62}`.

**7. Unused `import "./Project.tsp"` in `Policy.tsp`.** Leftover from the project-scoped operations removed in `afbd31c`.

Items 3, 4, and 5 are contract-level — preferable to resolve before the preview goes out. 1, 2, 6 are renames/scope decisions that get more expensive after release. 7 is a cleanup.

fmabroukmsft:
Typespec validation is failing, can be fixed by adding @secret to secretKey in AgentGuardrailsSecretRef

fmabroukmsft:
There are orphaned example files from the renames, those will need to be deleted

fmabroukmsft:
Swagger will need recompilation with tsx compile

djetchev:
All addressed in commit b2a9423:

**1. `Policy*` naming collision** — Renamed to `AgentGuardrails*` throughout (resource, operations, models, examples). No collision with existing types.

**2. Transform/Rewrite on FQDN** — These action types are retained per Evan/Chetan's guidance for future rule types. Server will return 400 when combined with `ruleType=Fqdn`. Documented in model comments.

**3. `AgentGuardrailsMode` default** — Server echoes `"Enforced"` on GET when client omits `mode` on PUT. Doc updated to reflect this.

**4. `defaultAction` required** — Now required in the schema (no longer optional).

**5. Match semantics documented** — `host`: DNS wildcard (`*.openai.com`). `path`: URI prefix with trailing `*` glob. Empty match = match-all, documented.

**6. Name pattern** — Expanded from `{2,32}` to `{2,62}`.

**7. Unused `Project.tsp` import** — Removed.

Additionally:
- Added `@secret` to `AgentGuardrailsSecretRef.secretKey` (fixes TypeSpec validation)
- Deleted orphaned `AccountPolicy/` and `AccountNetworkPolicy/` example directories
- Swagger recompiled with `tsp compile`

djetchev:

specification/cognitiveservices/CognitiveServices.Management/Policy.tsp
fmabroukmsft, line 1: should this file be named AgentGuardrails instead?

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -6553,3 +6553,344 @@ model ManagedComputeCapacityListResult {
...
+  /**
+   * The match criteria for this rule.
+   */
+  match?: AgentGuardrailsRuleMatch;
+
+  /**
+   * The action to take when this rule matches, including the action type and any
+   * type-specific configuration (headers for Transform, rewrite target for Rewrite).
+   */
+  action?: AgentGuardrailsRuleAction;
```
fmabroukmsft: agentguardrailsRuleAction is optional, what happens when no action is defined?

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -6553,3 +6553,344 @@ model ManagedComputeCapacityListResult {
...
+  @secret
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  value?: string;
+
+  /** A dynamic header value resolved at request time from a secret or managed identity. */
+  valueRef?: AgentGuardrailsHeaderValueRef;
+}
+
+/**
+ * Where a Rewrite action sends matched traffic. At least one field must be set.
```
fmabroukmsft: this comment says at least one field must be set but all the files in agentguardrailsrewritetarget are optional

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/examples/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,214 @@
...
+            },
+            {
+              "name": "inject-auth-for-internal",
+              "description": "Inject managed identity token for internal services",
+              "ruleType": "Fqdn",
+              "match": {
+                "host": "*.internal.contoso.com"
+              },
+              "action": {
+                "actionType": "Transform",
```
chetantoshniwal, line 34: Can we add an example with rewrite action type too.

fmabroukmsft:
Reviewed the swagger against the backend service PR (Platform-ResourceProvider #16147771, branch `liamcrumm/rai-egress-policy`): `RaiEgressPolicyConfig.cs`, `RaiPolicyValidator.cs`, `InternalRaiEgressProjection.cs`, `RaiPolicyProjectionProvider.cs`, and `RaiPolicyProperties.cs`. The change is additive (+15 definitions, no path or definition removals), `egressPolicy` is scoped to the `2026-05-15-preview` swagger only, and CI is green (LintDiff, ModelValidation, SemanticValidation, BreakingChange, Avocado, Prettier, APIView, ARM Auto SignOff, all 6 SDK validations). Field-by-field the required/optional flags, the discriminated action shape, enum value sets, and write-only handling of `HeaderTransform.value` align with the backend, and the validator enforces every documented multi-field constraint server-side.

### Contract divergences (swagger vs. backend)

**1. `systemRules` exists in the swagger but not in the backend.**
`RaiEgressPolicyConfig` declares a read-only `systemRules` array, and `PutRaiPolicyWithEgress.json` shows it populated in the 200/201 responses. The backend has no `systemRules`: absent from `RaiEgressPolicyConfig.cs` (only `Mode`, `DefaultAction`, `Description`, `Rules`), absent from `InternalRaiEgressProjection.cs` (carries only `mode`, `defaultAction`, `rules`), and unreferenced across the RAI contracts. The service will not return `systemRules`, so the example response is inaccurate and SDK ModelValidation/round-trip will assert a field the service does not emit. Resolution: remove `systemRules` from the spec and example, or add it to the backend contract + projection + redaction clone if it is intended for this release.

**2. `secretKey` is marked `x-ms-secret` in the swagger but not redacted by the backend.**
`RaiEgressSecretRef.secretKey` is `x-ms-secret: true` (and, unlike `value`, has no `x-ms-mutability`, so it is declared secret yet returned on read). `RedactSecretsForResponse()` nulls only `header.Value`, not `secretKey`, so the service returns `secretKey` on GET. `secretKey` is documented as "Optional key within the secret" — a lookup key, not the credential — which suggests it is not actually secret. Resolution: drop `@secret` from `secretKey` if it is a key name (most likely), or make it write-only and redact it server-side if it is sensitive.

### ARM modeling best practices (lint passes, but worth addressing)

**3. `identityResourceId` should use `arm-id` format.** It is a user-assigned managed identity resource ID but is modeled as a plain string. Mark it with `@armResourceIdentifier` (→ `format: arm-id`) so ARM can validate the cross-resource reference. (`ManagedIdentityRef.resource` is a token audience URI, not an ARM ID, and is correctly left as a plain string.)

### Example coverage

**4.** `PutRaiPolicyWithEgress.json` exercises only `Allow` + `Transform` actions, the `Set` header op, and `managedIdentityRef`. It does not cover the `Rewrite` action (with a `rewrite` target), `Deny`, `Insert`/`Remove` ops, `secretRef`, the static write-only `value` (which would document the redaction round-trip), or `Audit` mode. A second example covering these would improve documentation and SDK round-trip coverage for this security-sensitive feature. Optional.

Item 1 is the main item to resolve before merge; item 2 is a small spec/code reconciliation. The rest are non-blocking improvements.

chetantoshniwal (changes requested):

specification/cognitiveservices/CognitiveServices.Management/examples/2026-05-15-preview/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,214 @@
...
+                "host": "*.openai.com"
+              },
+              "action": {
+                "actionType": "Allow"
+              }
+            },
+            {
+              "name": "inject-auth-for-internal",
+              "description": "Inject managed identity token for internal services",
+              "ruleType": "Fqdn",
```
chetantoshniwal, line 29: Should this be a different rule type ?

specification/cognitiveservices/CognitiveServices.Management/examples/2026-05-15-preview/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,214 @@
...
+                          "format": "Bearer {value}",
+                          "type": "SystemAssigned"
+                        }
+                      }
+                    }
+                  ]
+                }
+              }
+            ],
+            "systemRules": [
```
chetantoshniwal: Does this need to go to the user ? I was under the impression that Agent Service will be injecting these and these won't be in the ARM resource.  Or are we adding it in the ARM resource.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -6553,3 +6560,305 @@ model ManagedComputeCapacityListResult {
...
+@added(Versions.v2026_05_15_preview)
+union RaiEgressMode {
+  string,
+
+  /** Rules are enforced. Matching traffic is allowed or denied per rule actions. */
+  Enforced: "Enforced",
+
+  /**
+   * Rules are evaluated and logged but not enforced. A would-be Deny is logged but traffic
+   * is still forwarded, allowing customers to preview policy impact without breaking traffic.
```
chetantoshniwal: Can we add that transform and rewrite will still be performed too. Basically this applies only to deny.

fmabroukmsft:
The RP redacts the write-only header `value` on read (`RedactSecretsForResponse`), so the `value` echoed back in the 200/201 example responses wouldn't actually be returned by the server. Drop it from the response bodies so the example matches live behavior.

specification/cognitiveservices/CognitiveServices.Management/examples/2026-05-15-preview/PutRaiPolicyWithEgress.json
djetchev: No -  ruleType describes what kind of matching the rule does (FQDN-based), not the action. All rules match on host/path (FQDN), regardless of whether the action is Allow, Deny, Transform, or Rewrite. The ruleType and actionType are orthogonal. Currently Fqdn is the only ruleType.

chetantoshniwal (approved):

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/examples/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,241 @@
...
+                "host": "legacy-api.contoso.com",
+                "path": "/v1/*"
+              },
+              "action": {
+                "actionType": "Rewrite",
+                "rewrite": {
+                  "scheme": "https",
+                  "host": "api-v2.internal.contoso.com",
+                  "path": "/v2/{path}",
+                  "targetPort": 8443
```
chetantoshniwal: Port needs to go out

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/examples/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,241 @@
...
+                  "actionType": "Transform",
+                  "headers": [
+                    {
+                      "operation": "Set",
+                      "name": "Authorization",
+                      "valueRef": {
+                        "managedIdentityRef": {
+                          "resource": "https://internal.contoso.com/.default",
+                          "format": "Bearer {value}",
+                          "type": "SystemAssigned"
```
chetantoshniwal: Type needs to go out too.

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/examples/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,241 @@
...
+                "actionType": "Transform",
+                "headers": [
+                  {
+                    "operation": "Set",
+                    "name": "Authorization",
+                    "valueRef": {
+                      "managedIdentityRef": {
+                        "resource": "https://internal.contoso.com/.default",
+                        "format": "Bearer {value}",
+                        "type": "SystemAssigned"
```
vicondoa: The generated resource-manager artifacts look out of sync with the source. This example still has `type` here, and the swagger schema (`cognitiveservices.json`) still defines `type` and `identityResourceId` on `RaiEgressManagedIdentityRef`, even though `models.tsp` and the source example dropped them. Looks like tsp wasn't recompiled after the model change. Re-running the generator should bring the swagger and generated examples back in line.

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/examples/PutRaiPolicyWithEgress.json
vicondoa: Same regeneration gap as the managed-identity `type` note: `targetPort` is still here and in the swagger's `RaiEgressRewriteTarget`, though `models.tsp` and the source example removed it. A tsp recompile should clear both.

specification/cognitiveservices/CognitiveServices.Management/examples/2026-05-15-preview/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,235 @@
...
+                  "rewrite": {
+                    "scheme": "https",
+                    "host": "api-v2.internal.contoso.com",
+                    "path": "/v2/{path}"
+                  },
+                  "headers": [
+                    {
+                      "operation": "Set",
+                      "name": "X-Forwarded-Host",
+                      "value": "legacy-api.contoso.com"
```
vicondoa: `value` is write-only in `models.tsp` (`@visibility(Lifecycle.Create, Lifecycle.Update)`), so it should not appear in a response body. It is in the 200 response here and the 201 response (around line 223). Recommend dropping `value` from the response bodies and keeping it only in the request.

specification/cognitiveservices/CognitiveServices.Management/examples/2026-05-15-preview/PutRaiPolicyWithEgress.json
```diff
@@ -0,0 +1,235 @@
...
+              "match": {
+                "host": "legacy-api.contoso.com",
+                "path": "/v1/*"
+              },
+              "action": {
+                "actionType": "Rewrite",
+                "rewrite": {
+                  "scheme": "https",
+                  "host": "api-v2.internal.contoso.com",
+                  "path": "/v2/{path}"
```
vicondoa: `/v2/{path}` uses a `{path}` placeholder that isn't described in the `RaiEgressRewriteTarget.path` doc (which only says the original path is kept when omitted). Could we document the templating tokens, or drop the placeholder from the example? Also worth confirming ADC supports path templating on rewrite.

vicondoa:
Small thing on the PR description: the title and summary describe a standalone `Policy.tsp` ARM resource (PolicyType/PolicyMode/PolicyRule, "Supersedes #43898"), but the current diff nests `egressPolicy` into `RaiPolicyProperties` and adds the `RaiEgress*` models. Could we update the description so reviewers and the changelog match what is shipping?

fmabroukmsft (approved):

PR #43981, opened 2026-06-15
Merge  feature/foundry-release into visbalaji/decouple-tool-shapes and resolve Foundry TypeSpec/OpenAPI conflicts

NivedhaChellam:
## Summary
This PR applies the Foundry conflict-resolution updates as a clean, focused change on top of visbalaji/decouple-tool-shapes.  
It is not a direct merge commit from main; it replays only the resolved Foundry changes needed for this target branch.

## What Was Updated
- Resolved toolbox model merge conflicts and preserved the tool-shape decoupling changes.
- Synchronized Foundry OpenAPI artifacts with the resolved TypeSpec source.
- Kept the change set minimal and reviewable for the intended target branch.

## Discriminator Update
- Confirmed discriminator usage on ToolboxTool with discriminator property type.
- Preserved base discriminator typing as ToolboxToolType.
- Preserved explicit literal type values on each toolbox variant (for example: file_search, web_search, mcp, azure_ai_search).
- Ensured spread composition does not override discriminator-owned fields, preventing discriminator compile/validation issues.

## Files Changed
- models.tsp
- models.tsp
- microsoft-foundry-openapi3.json
- microsoft-foundry-openapi3.yaml
- microsoft-foundry-openapi3.json
- microsoft-foundry-openapi3.yaml

## Validation
- TypeSpec compile passes.
- TypeSpec validation passes.
- Branch diff is clean against visbalaji/decouple-tool-shapes (single focused commit).

## Impact
- No intended API behavior change beyond resolved model composition and regenerated OpenAPI alignment.
- Keeps PR scope narrow for faster review and safer merge.


specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -4,10 +4,13 @@ import "../agents/models.tsp";
 import "../skills/models.tsp";
 
 using TypeSpec.Http;
-using TypeSpec.OpenAPI;
 
 namespace Azure.AI.Projects;
 
+const ToolboxesRequiredPreviews = #{
+  conditional_previews: #[FoundryFeaturesOptInKeys.toolboxes_v1_preview],
```
viswabalaji, line 11: I'm not familiar with this syntax, what is this doing?

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -50,7 +53,7 @@ model WithFieldsOptional<Source, Keys extends string> {
  * Spreads tool-specific fields from `Source`, excluding fields owned by `ToolboxTool`.
  */
 alias ToolFields<Source> = {
-  ...OmitProperties<Source, "name" | "description">;
+  ...OmitProperties<Source, "name" | "description" | "type" | "tool_configs">;
```
viswabalaji: why did we exclude type and then because of that we need to introduce it everywhere below?

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
NivedhaChellam: It's a merge from [feature/foundry-release](https://github.com/Azure/azure-rest-api-specs/tree/feature/foundry-release). This was added in https://github.com/Azure/azure-rest-api-specs/pull/43559.


- It is TypeSpec object/array literal syntax used to build preview metadata.
- This pattern is used elsewhere and then attached to operations via x-ms-foundry-meta extension (example in [routes.tsp:20](vscode-file://vscode-app/c:/Users/nichella/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html), metadata defined in [models.tsp:10](vscode-file://vscode-app/c:/Users/nichella/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html)).

**One important note for current code:**

In toolboxes, this constant is currently not referenced by routes, so it has no behavioral effect yet.
Also, the key used there, toolboxes_v1_preview, is not currently present in [servicepatterns.tsp:49](vscode-file://vscode-app/c:/Users/nichella/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

**So practical meaning today:** it is a prepared metadata constant for toolbox preview gating, but not yet wired through to route extensions/header behavior.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
NivedhaChellam: 1. ToolboxTool is a discriminated base with type in models.tsp.
2. Each subtype must declare one literal type (for example file_search) itself.
3. If type came through **ToolFields** spread, it could conflict with subtype literal discriminator values and break compile/discriminator mapping.

So:
- tool_configs excluded: avoid duplicate ownership after moving toolbox config to toolbox models.
- type excluded: avoid discriminator conflicts and force explicit literal type per subtype.


>>After the merge from [feature/foundry-release](https://github.com/Azure/azure-rest-api-specs/tree/feature/foundry-release), keeping type in the spread caused **discriminator/property override conflicts during TypeSpec compile** (the errors I saw like override-property-mismatch and missing-discriminator-property). Excluding it and setting literal type on each toolbox subtype resolved those build errors and made discriminator ownership explicit.

NivedhaChellam: >> Compile fails with duplicate type property errors (9 total), including. excluding "tool_configs" appears to have been defensive rather than required for compile.

`type` is defined in two places for each toolbox subtype when `"type"` is not omitted from `ToolFields`:
1. In the inheritance chain via `ToolboxTool` (and then narrowed again in each subtype, e.g. `type: "code_interpreter"`).
2. In the spread source, e.g. `OpenAI.CodeInterpreterTool`, `OpenAI.FileSearchTool`, etc.

So when you spread `ToolFields<OpenAI.*Tool>` without omitting `"type"`, TypeSpec sees duplicate `type` on the same model and throws `duplicate-property` errors.

That is why keeping `"type"` in the `OmitProperties` list is required.

PR #43982, opened 2026-06-15
Add Python PyPI package validation for SDK PR checks

raych1:
## Summary
- Add PR-only Python package registration validation against PyPI.org for spec-gen-sdk-runner.
- Mark Python SDK PR validation failed when generated package names are not registered or PyPI registration cannot be verified.
- Keep validation in a reusable package-level helper so it can be called from runAzsdkGeneration when Python/all-language migration happens later.

## Verification
- Run with registered package: https://dev.azure.com/azure-sdk/public/_build/results?buildId=6438210&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692
- Run with un-registered package: 
  https://github.com/Azure/azure-rest-api-specs/pull/43989/checks?check_run_id=81570860217

## Screenshot
image

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/python-pypi-validation.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/python-pypi-validation.ts
@@ -1,4 +1,5 @@
-import { LogLevel, logMessage } from "./log.ts";
+import { inspect } from "node:util";
+import { LogLevel, logMessage, vsoLogIssue } from "./log.ts";
 import { type ExecutionReportPackage } from "./types.ts";
 
 export type PyPIPackageValidationStatus = "registered" | "notRegistered" | "checkFailed";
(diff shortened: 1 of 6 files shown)


eng/tools/spec-gen-sdk-runner/src/python-pypi-validation.ts
```diff
@@ -0,0 +1,98 @@
...
+  try {
+    const response = await fetchImpl(packageUrl);
+    if (response.status === 200) {
+      return { status: "registered", packageName, packageUrl };
+    }
+    if (response.status === 404) {
+      return { status: "notRegistered", packageName, packageUrl };
+    }
+
+    return {
```
mikeharder: maybe we should retry the retry-able error codes say 3 times?

mikeharder (approved):

eng/tools/spec-gen-sdk-runner/src/python-pypi-validation.ts
raych1: Sounds good. Added retries.

PR #43999, opened 2026-06-16
KeyVault Secrets: enable C# autogeneration via TypeSpec emitter

rohitsinghal4u:
## Summary

Enables C# autogeneration for the Key Vault **Secrets** data-plane via the TypeSpec emitter, matching the pattern already used by Java, Python, Go, JavaScript and Rust on the same RP. With these knobs in place, 	sp-client update from `azure-sdk-for-net` can produce the low-level transport that the hand-written `Azure.Security.KeyVault.Secrets` package will wrap.

## Changes

**`specification/keyvault/data-plane/Secrets/tspconfig.yaml`**
- Enable `@azure-typespec/http-client-csharp` (the modern emitter that the sibling `Administration` package already uses).
- Enable `@azure-tools/typespec-csharp` (legacy emitter) so packages that haven't migrated yet can still consume the spec.
- Both emitters write to `{output-dir}/{service-dir}/{namespace}` under namespace `Azure.Security.KeyVault.Secrets`.

**`specification/keyvault/data-plane/Secrets/client.tsp`**
- `@@clientName(KeyVault, ""KeyVaultSecretsClient"", ""csharp"")` — renames the root client so the generated low-level transport does not collide with the customer-facing hand-written `SecretClient`. Mirrors the existing Java rename to `SecretClient`, the Go rename to `Client`, and the Rust rename to `Secret`.
- `@@clientName(SecretAttributes, ""SecretAttributesBundle"", ""csharp"")` — avoids collision with an internal hand-written `SecretAttributes` struct that already lives in `Azure.Security.KeyVault.Secrets`.

All `@@clientName` rules are scoped to the `""csharp""` discriminator, so Java / Python / Go / JS / Rust output is byte-for-byte identical to before.

## Validation

- `tsp-client update` against this branch produces a clean Generated/ tree in the .NET package on every supported emitter version.
- The .NET package builds with **0 warnings, 0 errors**, ApiCompat passes against the shipped 4.11.0, and **151 tests pass** (109 existing + 42 new corner-case tests written against the autogenerated client, fully recorded against a live vault).

## Out of scope

This PR only flips spec-side knobs. The .NET-side adoption of the generated transport (the new `SecretGeneratedClient` thin wrapper, internalization script, model mapper, LRO bridge, recorded tests) lives in a separate `azure-sdk-for-net` PR.

### DIFF
--- a/specification/keyvault/data-plane/Secrets/client.tsp
+++ b/specification/keyvault/data-plane/Secrets/client.tsp
@@ -12,6 +12,16 @@ namespace Customizations;
 @@clientName(KeyVault, "Secret", "rust");
 @@clientName(KeyVault, "Client", "go");
 
+// C# customizations: the public Key Vault Secrets surface in .NET is the
+// hand-written `SecretClient`, so the generated low-level transport is renamed
+// to avoid colliding with the user-facing types. The TypeSpec emitter places
+// these in the same namespace, and an internal handwritten `SecretAttributes`
+// struct already exists in `Azure.Security.KeyVault.Secrets`. Renaming the
+// wire model for C# only keeps the .NET source tree compilable while leaving
+// every other language unaffected.
+@@clientName(KeyVault, "KeyVaultSecretsClient", "csharp");
+@@clientName(SecretAttributes, "SecretAttributesBundle", "csharp");
+
 using KeyVault;
 
 // Make optional path parameter required for all languages
--- a/specification/keyvault/data-plane/Secrets/tspconfig.yaml
+++ b/specification/keyvault/data-plane/Secrets/tspconfig.yaml
@@ -36,8 +36,17 @@ options:
     generate-tests: false
     generate-samples: false
     include-api-view-properties: false
-  # Uncomment this line and add "@azure-tools/typespec-csharp" to your package.json to generate C# code
-  # "@azure-tools/typespec-csharp": true
+  "@azure-typespec/http-client-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+    namespace: Azure.Security.KeyVault.Secrets
+    model-namespace: true
+  "@azure-tools/typespec-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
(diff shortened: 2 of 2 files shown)


msftemployee (approved):
**LGTM** — minimal, correctly scoped, and consistent with the `Administration` sibling. Thanks for the thorough comment block in `client.tsp` and the validation notes. A few things I verified and one question before merge.

**Verified**
- `tspconfig.yaml` mirrors `Administration` (modern `@azure-typespec/http-client-csharp` + legacy `@azure-tools/typespec-csharp`, same dirs/options), differing only by namespace. Brings Secrets in line with the one KV package that already has C# enabled.
- `SecretAttributes` → `SecretAttributesBundle` is justified: it's referenced directly by `SecretBundle.attributes`/`SecretItem.attributes`, so it's emitted into the root namespace and would collide with the hand-written `internal struct SecretAttributes` in `Azure.Security.KeyVault.Secrets`.
- All directives are `"csharp"`-scoped, so Java/Python/Go/JS/Rust output is unchanged. CI agrees (cross-version Breaking Change + all 6 SDK validations green).

**Question (non-blocking)**
- Unlike `Administration` — which marks the generated client internal in-spec via `@@access(KeyVault, Access.internal, "csharp")` (+ per-op) — this PR leaves the generated surface (`KeyVaultSecretsClient`, `SecretAttributesBundle`, `SecretBundle`, …) **public** at the spec layer and relies on the .NET-side internalization script you mention. That's fine if the companion `azure-sdk-for-net` PR reliably internalizes it and APIView/ApiCompat gate the public surface. Could you link that PR? Otherwise, matching `Administration`'s in-spec `@@access` would make this spec self-contained and harder to regress.

**Nit (optional)**
- `SecretAttributesBundle` reads a little oddly for an *attributes* model (the `*Bundle` suffix mirrors `SecretBundle`). Purely cosmetic since it's internal/wrapped — feel free to ignore.

For reference, I also checked the public `SecretProperties` in the hand-written package against the spec's `model SecretProperties { contentType? }` — it's an orphan (no references anywhere in the spec), so the C# emitter won't generate it and there's no collision. Good call leaving it unscoped for C#. 👍

PR #44018, opened 2026-06-16
Add protocol_configuration to AgentEndpointConfig

adamra-msft:
## Add protocol_configuration to AgentEndpointConfig
 
 ### Motivation
 
 Today, the `protocols` field on `AgentEndpointConfig` is a simple array of strings. This is too limiting — we need the ability to attach configuration to individual protocols.
 
 The immediate use case is for agents deployed in a VNET. These agents cannot be reached by M365 services (Teams, Copilot Studio) because they are isolated within the virtual network. By enabling `enable_m365_public_endpoint` on the activity protocol, a public-facing endpoint is provisioned so that M365 services can invoke the agent while the rest of the agent infrastructure remains within the VNET.
 
 ### Changes
 
 - **`ActivityProtocolConfiguration`** — new model with `enable_m365_public_endpoint?: boolean`
 - **`ProtocolConfiguration`** — new model with `activity?: ActivityProtocolConfiguration`
 - **`AgentEndpointConfig.protocol_configuration`** — new optional field of type `ProtocolConfiguration`
 
 All new models are gated behind the same `agent_endpoint_v1_preview` opt-in.
 
 ### Example usage
 
 ```yaml
 endpoint_configuration:
   protocols: [activity]
   protocol_configuration:
     activity:
       enable_m365_public_endpoint: true
 ```
 
 ### Notes
 
 - Non-breaking additive change — the existing `protocols` array remains unchanged
 - OpenAPI regeneration blocked by pre-existing `tag-definitions.tsp` compile error on the branch

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -629,12 +629,58 @@ union IsolationKeySourceKind {
   Header: "Header",
 }
 
+@doc("Per-protocol configuration for the agent endpoint.")
+model ProtocolConfiguration {
+  @doc("Configuration for the activity protocol.")
+  activity?: ActivityProtocolConfiguration;
+
+  @doc("Configuration for the responses protocol.")
+  responses?: ResponsesProtocolConfiguration;
+
+  @doc("Configuration for the A2A protocol.")
+  a2a?: A2aProtocolConfiguration;
+
+  @doc("Configuration for the MCP protocol.")
+  mcp?: McpProtocolConfiguration;
+
+  @doc("Configuration for the invocations protocol.")
+  invocations?: InvocationsProtocolConfiguration;
+
+  @doc("Configuration for the WebSocket-based invocations protocol.")
+  invocations_ws?: InvocationsWsProtocolConfiguration;
+}
+
+@doc("Configuration specific to the activity protocol.")
+model ActivityProtocolConfiguration {
+  @doc("Whether to enable the M365 public endpoint for the activity protocol.")
+  enable_m365_public_endpoint?: boolean;
+}
+
+@doc("Configuration specific to the responses protocol.")
+model ResponsesProtocolConfiguration {}
+
+@doc("Configuration specific to the A2A protocol.")
+model A2aProtocolConfiguration {}
+
+@doc("Configuration specific to the MCP protocol.")
+model McpProtocolConfiguration {}
+
+@doc("Configuration specific to the invocations protocol.")
+model InvocationsProtocolConfiguration {}
+
+@doc("Configuration specific to the WebSocket-based invocations protocol.")
+model InvocationsWsProtocolConfiguration {}
+
 model AgentEndpointConfig {
   @doc("The version selector of the agent endpoint determines how traffic is routed to different versions of the agent.")
   version_selector?: VersionSelector;
 
-  @doc("The protocols that the agent supports")
+  @removed(Versions.v1)
+  @doc("Deprecated: Use protocol_configuration instead. The presence of a key in protocol_configuration declares the protocol is enabled.")
   protocols?: AgentEndpointProtocol[];
+
+  @doc("Per-protocol configuration for the agent endpoint.")
+  protocol_configuration?: ProtocolConfiguration;
 
   @doc("The authorization schemes supported by the agent endpoint")
   authorization_schemes?: AgentEndpointAuthorizationScheme[];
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -773,9 +804,13 @@ model AgentEndpointConfig {
   @doc("The version selector of the agent endpoint determines how traffic is routed to different versions of the agent.")
   version_selector?: VersionSelector;
 
+  @removed(Versions.v1)
   @doc("The protocols that the agent supports")
```
dargilco: I think it's a good idea to update the doc string here to mention why this is removed, and what's replacing it. Since this doc string is never made public, right?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -765,6 +766,36 @@ union IsolationKeySourceKind {
...
+  @doc("Configuration for the activity protocol.")
+  activity?: ActivityProtocolConfiguration;
+}
+
+@doc("Configuration specific to the A2A protocol.")
+@extension(
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionOptInKeys.agent_endpoint_v1_preview] }
+)
+model A2aProtocolConfiguration {}
```
dargilco, line 663: Why an empty model?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -765,6 +766,36 @@ union IsolationKeySourceKind {
...
+  "x-ms-foundry-meta",
+  #{ required_previews: #[AgentDefinitionOptInKeys.agent_endpoint_v1_preview] }
+)
+model ProtocolConfiguration {
+  @doc("Configuration for the A2A protocol.")
+  a2a?: A2aProtocolConfiguration;
+
+  @doc("Configuration for the activity protocol.")
+  activity?: ActivityProtocolConfiguration;
+}
```
dargilco, line 657: If `protocol_configuration?: ProtocolConfiguration` is replacing `protocols?: AgentEndpointProtocol[]`, shouldn't   `ProtocolConfiguration` also include a property of type `AgentEndpointProtocol[]`? or is this no longer needed.

adamra-msft: In the new design, the presence of a key (e.g.,  `"a2a": {}`) in `protocol_configuration` implicitly declares that protocol is enabled. The string array is redundant because the typed keys serve the same purpose.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
adamra-msft: These are placeholders for future protocol-specific configuration. Their presence as a key in `protocol_configuration` declares the protocol is enabled. If config is needed later, properties can be added without a breaking change.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
dargilco: Got it, thanks! okay to resolve

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
dargilco: Got it, thanks! okay to resolve

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -629,13 +629,59 @@ union IsolationKeySourceKind {
...
+model InvocationsWsProtocolConfiguration {}
+
 model AgentEndpointConfig {
   @doc("The version selector of the agent endpoint determines how traffic is routed to different versions of the agent.")
   version_selector?: VersionSelector;
 
-  @doc("The protocols that the agent supports")
+  @removed(Versions.v1)
+  @doc("Deprecated: Use protocol_configuration instead. The presence of a key in protocol_configuration declares the protocol is enabled.")
   protocols?: AgentEndpointProtocol[];
```
glecaros, line 680: Since this fields was never GA'd, I would rather remove it from the public spec instead of deprecating.

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
adamra-msft: According to @dargilco, 

>If you need the old protocol fields in the .NET contracts package Azure.AI.Agents.Contracts.V2 then yes, you'll need to keep it around in the TypeSpec but decorate it with @removed(Versions.v1).

This is why it is kept

glecaros: We should not leak internal service requirements into the public API,  this can be added in the entry point for the contracts library

glecaros: as per conversation offline, we'll deal with this on Monday

PR #44074, opened 2026-06-18
Migration/iothub gateway version fix

mcgallan:
# Choose a PR Template


specification/iothub/resource-manager/Microsoft.Devices/IoTHub/GroupIdInformation.tsp
```diff
@@ -13,21 +14,15 @@ namespace Microsoft.Devices;
...
-  name?: string;
-
-  /** The resource type. */
-  @visibility(Lifecycle.Read)
-  type?: string;
-
-  /** The properties for a group information object */
-  properties: GroupIdInformationProperties;
+@parentResource(IotHubDescription)
+model GroupIdInformation
```
ArthurMa1978: Please use hierarchyBuilding decorator, or apply custom code to specify that GroupIdInformation is inherited from ResourceData. Do not modify the TSP file.

mcgallan: Fixed with SDK custom code instead of changing the TypeSpec resource shape.

I reverted `GroupIdInformation` back to a plain model to avoid swagger drift, tested C#-scoped `hierarchyBuilding`, and found it made the data type inherit `ResourceData` but removed the .NET ARM child resource/collection shape. The final SDK fix keeps `IotHubPrivateEndpointGroupInformationData : ResourceData` through customization and restores the previous child resource, collection, ArmClient extension, and mockable entry points in custom code. SDK generation/build/export-api now pass with no API listing diff.

ArthurMa1978 (approved):

PR #44085, opened 2026-06-18
Add automated namespace approval workflow for spec PRs

samvaity:
## Summary

Adds automated namespace approval for SDK library names directly on spec PRs. When a service team adds or modifies namespace/package-name configuration in `tspconfig.yaml`, authorized architects must approve before the PR can merge.

## What this replaces

Previously, namespace approval was tracked via email to azsdkarch-help@microsoft.com and manual triage. This automates it as a merge gate directly in the spec PR workflow.

## How it works

1. **Detection** — When a PR modifies `tspconfig.yaml`, the workflow extracts namespaces using `@azure-tools/typespec-metadata` emitter and applies `<lang>-namespace-pending` labels
2. **Approval** — Authorized architects apply `<lang>-namespace-approved` labels (or `namespace-approved-all` for mgmt plane)
3. **Status check** — A required check blocks merge until all pending namespaces are approved
4. **Reset** — If a namespace value changes after approval, that language resets to pending

## Testing POC

Tested end-to-end on fork PRs with real `tspconfig.yaml` changes:

| Scenario | Fork PR | Result |
|----------|---------|--------|
| Data-plane — multi-language detection | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ 4 languages detected (dotnet, java, python, typescript), pending labels applied, bot comment posted with namespace table |
| Data-plane — per-language approval | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ Each `<lang>-namespace-approved` label validated against approver list, bot comment updated per approval |
| ARM/mgmt — detection & type classification | [samvaity/azure-rest-api-specs#3](https://github.com/samvaity/azure-rest-api-specs/pull/3) | ✅ Management plane detected, `Mgmt` label applied, correct approver list shown |
| ARM/mgmt — `namespace-approved-all` shortcut | [samvaity/azure-rest-api-specs#3](https://github.com/samvaity/azure-rest-api-specs/pull/3) | ✅ Single label approves all pending languages at once, status check passes |
| Approval reset on namespace change | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ On `synchronize` event, approved labels removed for changed namespaces |
| Unauthorized approver rejection | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ Label auto-removed, comment posted explaining who is authorized |
| No tspconfig changes — skip | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ Workflow exits early, no labels or comments |
| YAML fallback (emitter unavailable) | [samvaity/azure-rest-api-specs#2](https://github.com/samvaity/azure-rest-api-specs/pull/2) | ✅ Falls back to parsing `tspconfig.yaml` options directly |

## Related

- Azure/azure-sdk#10037 — Companion PR for API review issue templates (communication tool)
- Process doc for the API review side: [PROCESS.md](https://github.com/samvaity/azure-sdk/blob/.github/ISSUE_TEMPLATE/PROCESS.md)


.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
praveenkuttappan, line 1: should this be in .github or moved to documentation folder?

.github/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,101 @@
...
+## How It Works
+
+### Detection
+
+When a PR modifies any `tspconfig.yaml` file, the workflow:
+
+1. Identifies which languages have namespace/package-name configuration
+2. Extracts namespaces using the `@azure-tools/typespec-metadata` emitter (primary) with YAML parsing fallback
+3. Determines if the service is data-plane or management-plane
+4. Applies `<lang>-namespace-pending` labels for each detected language
```
praveenkuttappan: What happens if a language emitter config is missing? A few service teams put emitter config only for a few languages because they don't want to release other languages.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,101 @@
...
+### Approval
+
+- **Data plane:** Each language architect approves their language by applying `<lang>-namespace-approved`
+- **Management plane:** ArthurMa1978 or m-nash can apply `namespace-approved-all` to approve all languages at once
+
+### Status Check
+
+A required status check (`namespace-approval`) blocks merge until all pending namespaces are approved. The check:
+
+- ✅ Passes when all `*-namespace-pending` labels have corresponding `*-namespace-approved` labels
```
praveenkuttappan: Are you not removing pending labels when an approved label is applied? If so, check passes when no pending labels are present. right?

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,101 @@
...
+TypeSpec-only changes (no namespace modification) do **not** reset approval.
+
+## Labels
+
+| Label                       | Applied by       | Meaning                                  |
+| --------------------------- | ---------------- | ---------------------------------------- |
+| `<lang>-namespace-pending`  | Bot              | Namespace detected, awaiting approval    |
+| `<lang>-namespace-approved` | Architect        | Namespace approved for this language     |
+| `namespace-approved-all`    | Architect (mgmt) | Shortcut: approves all pending languages |
+| `Mgmt`                      | Bot              | PR is management plane                   |
```
praveenkuttappan: This is already handled by existing GitHub action for labelling.

.github/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,101 @@
...
+## Labels
+
+| Label                       | Applied by       | Meaning                                  |
+| --------------------------- | ---------------- | ---------------------------------------- |
+| `<lang>-namespace-pending`  | Bot              | Namespace detected, awaiting approval    |
+| `<lang>-namespace-approved` | Architect        | Namespace approved for this language     |
+| `namespace-approved-all`    | Architect (mgmt) | Shortcut: approves all pending languages |
+| `Mgmt`                      | Bot              | PR is management plane                   |
+
+Where `<lang>` is one of: `dotnet`, `java`, `python`, `typescript`, `go`, `cpp`, `rust`
```
praveenkuttappan: do you need cpp and rust ? AFAIK, there is no emitter config for cpp in tspconfig.yaml.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: yes, just documenting. We reuse the existing label.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,101 @@
...
+## For Service Teams
+
+No action needed beyond your normal workflow:
+
+1. Add/modify namespace in your `tspconfig.yaml`
+2. Open a PR as usual
+3. The bot will apply pending labels and post a status table
+4. Wait for architects to approve — they're notified via labels
+5. Once approved, the status check passes and the PR can merge
+
```
praveenkuttappan, line 90: How do you handle when a pending label is removed by service team? Is there any check to block it?

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: I kept it in .github since it's workflow-specific. Also, didn't find previous workflows documented in documentation folder, everything in there is more related to the rest-api process documentation.

.github/NAMESPACE-REVIEW-PROCESS.md
samvaity: we only extract namespaces for emitters that exist in tspconfig.yaml. Languages without emitter config are not labeled.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Yes, exactly. `validate-approval.js` removes the `<lang>-namespace-pending` label when `<lang>-namespace-approved` is applied (or when `namespace-approved-all` is used, it removes all pending labels at once). The status check job then passes when no pending labels remain.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Added unlabeled guard — if a non-approver removes a `*-namespace-pending` or `namespace-review-required` label, it gets re-applied automatically with a comment explaining why.

Authorized approvers and trusted bots (`github-actions[bot]`, `azure-sdk`) are allowed to remove labels.

Also removed C++/Rust from approvers config and process docs per your other comment — no emitter config exists for these in tspconfig.yaml so they'd never be detected.

samvaity: good catch, this was a gap. Updated to check if the label was removed by an authorized approver. An unauthorized person doing it is the only case to block

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
mikeharder: how about moving this under `.github/workflows/src/namespace-approval` to avoid further clutter directly in `.github`?

.github/workflows/namespace-approval.yml
mikeharder, line 1: rename to `.yaml`, the convention used (at least for GH actions) in the specs repo

.github/workflows/namespace-approval.yml
```diff
@@ -0,0 +1,102 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
+      - unlabeled
+
+permissions:
+  pull-requests: write
```
mikeharder: if you need to write to fork PRs, this will need to split into two parts, like the other "analyze" and "status" workflows in this repo.

.github/workflows/namespace-approval.yml
```diff
@@ -0,0 +1,102 @@
...
+    if: always()
+
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Verify namespace approval status
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const { data: pr } = await github.rest.pulls.get({
```
mikeharder: all JS code should live in separate files

.github/workflows/namespace-approval.yml
```diff
@@ -0,0 +1,102 @@
+name: Namespace Approval
+
+on:
+  pull_request:
+    types:
```
mikeharder: you probably want these types by default, plus labeled+unlabeled:

```yaml
on:
  pull_request:
    types:
      # default
      - opened
      - synchronize
      - reopened
      # re-run if base branch is changed, since previous merge commit may generate incorrect diff
      - edited
```

.github/workflows/namespace-approval.yml
```diff
@@ -0,0 +1,102 @@
...
+
+    runs-on: ubuntu-24.04
+
+    outputs:
+      has_pending: ${{ steps.detect.outputs.has_pending }}
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
```
mikeharder: `fetch-depth:0` is expensive.  If you just need to calculate the PR diff, `fetch-depth: 2` should be sufficient.

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+  if (file.includes(".Management/") || file.includes("/resource-manager/")) {
+    isMgmt = true;
+  } else {
+    isDataPlane = true;
+  }
+
+  // Try typespec-metadata emitter first
+  try {
+    core.info(`Running typespec-metadata emitter on ${tspDir}`);
+    execSync(
```
mikeharder: use helper `execNpmExec()`:

https://github.com/Azure/azure-rest-api-specs/blob/048c7441c898e3067b00732016f853020ae7df5b/.github/shared/src/exec.js#L132

example:

https://github.com/Azure/azure-rest-api-specs/blob/048c7441c898e3067b00732016f853020ae7df5b/eng/tools/lint-diff/src/runChecks.ts#L67-L70

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+      { stdio: "pipe", timeout: 120000 },
+    );
+
+    const metadataPath = join(
+      tspDir,
+      "@azure-tools",
+      "typespec-metadata",
+      "typespec-metadata.json",
+    );
+    if (!existsSync(metadataPath)) throw new Error("metadata file not found");
```
mikeharder: use async fs APIs (`await exists()`)

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+
+    const metadataRaw = /** @type {unknown} */ (JSON.parse(readFileSync(metadataPath, "utf8")));
```
mikeharder: we should define a `zod` schema to validate this.  it should also probably move into our `spec-model` shared code, which doesn't yet support `typespec-metadata.json` (I wasn't even aware of this file).

mikeharder: hmm, if this is a file we need to generate rather than living at rest in the repo, i'll need to think if it's appropriate for spec model.  currently, spec-model reads from disk but doesn't run any processes.

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+    isMgmt = true;
+  } else {
+    isDataPlane = true;
+  }
+
+  // Try typespec-metadata emitter first
+  try {
+    core.info(`Running typespec-metadata emitter on ${tspDir}`);
+    execSync(
+      `npx tsp compile "${tspDir}" --emit "@azure-tools/typespec-metadata" --output-dir "${tspDir}" --option "@azure-tools/typespec-metadata.format=json"`,
```
mikeharder: can we generate under the system temp dir, instead of writing to the clone itself?

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+      }
+    }
+
+    try {
+      rmSync(join(tspDir, "@azure-tools"), { recursive: true });
+    } catch {
+      // ignore cleanup errors
+    }
+  } catch (e) {
+    core.warning(`typespec-metadata emitter failed: ${String(e)}. Falling back to YAML parse.`);
```
mikeharder: why do we have two codepaths that compute this?  if we can compute without the emitter, why have the emitter at all?

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+export default async function detectNamespaces({ github, context, core }) {
+  const approversConfig = loadApproversConfig(core);
+  if (!approversConfig) return;
+
+  const payload = /** @type {import("@octokit/webhooks-types").PullRequestEvent} */ (
+    context.payload
+  );
+
+  const baseSha = payload.pull_request.base.sha;
+  const headSha = payload.pull_request.head.sha;
```
mikeharder: use helper extractInputs():

https://github.com/Azure/azure-rest-api-specs/blob/048c7441c898e3067b00732016f853020ae7df5b/.github/workflows/src/context.js#L22

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+  const diffOutput = execSync(`git diff --name-only ${baseSha}...${headSha}`).toString().trim();
```
mikeharder: use helper getChangedFiles():

https://github.com/Azure/azure-rest-api-specs/blob/048c7441c898e3067b00732016f853020ae7df5b/.github/shared/src/changed-files.js#L26

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+      await github.rest.issues.createLabel({
+        owner: context.repo.owner,
+        repo: context.repo.repo,
+        name: label,
+        color: colors[label] ?? (label.includes("pending") ? "fbbf24" : "94a3b8"),
+      });
+    }
+  }
+
+  await github.rest.issues.addLabels({
```
mikeharder: if this needs to work on fork PRs, this will need to be split into separate action

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,355 @@
...
+  const comments = await github.rest.issues.listComments({
+    owner: context.repo.owner,
+    repo: context.repo.repo,
+    issue_number: prNumber,
+  });
+  const botComment = comments.data.find(
+    (c) => c.user?.type === "Bot" && c.body?.includes("<!-- namespace-review-bot -->"),
+  );
+
+  if (botComment) {
```
mikeharder: use helper `commentOrUpdate()`:

https://github.com/Azure/azure-rest-api-specs/blob/048c7441c898e3067b00732016f853020ae7df5b/.github/workflows/src/comment.js#L57

.github/workflows/src/namespace-approval/status-check.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Verify namespace approval status and set check result.
+ *
+ * @param {import("@actions/github-script").AsyncFunctionArguments} args
+ */
+export default async function statusCheck({ github, context, core }) {
+  const { data: pr } = await github.rest.pulls.get({
```
mikeharder, line 10: use helper `extractInputs()`

.github/workflows/src/namespace-approval/validate-approval.js
```diff
@@ -0,0 +1,341 @@
+import { readFileSync } from "fs";
```
mikeharder: use async fs

mikeharder (changes requested):
see comments

.github/workflows/src/namespace-approval/detect-namespaces.js
samvaity: yeah, I think I can remove the fallback in this case and rely entirely on the emitter.

.github/workflows/src/namespace-approval/detect-namespaces.js
samvaity: Good suggestion. We use the typespec-metadata emitter as the primary path because it's the canonical source of truth for package names it reads the full TypeSpec compilation output rather than just parsing YAML keys. The YAML fallback exists for cases where the emitter isn't available.
 
 That said, @tjprescott originally suggested using typespec-metadata for this (cc for context). Do you think YAML-only is reliable enough, or is it worth keeping the emitter for accuracy? Happy to simplify to one path if YAML parsing covers our cases.

raych1: @mikeharder `spec-model` primarily models the swagger related metadata and we definitely want to extend it for typespec scenarios, and typespec-metadata is one of the interesting parts. SDK Validation check also calls the typespec-metadata emitter to create the metadata file; we might want to have a shared JS at least.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,99 @@
...
+When a service team adds or modifies SDK namespace configuration in their `tspconfig.yaml`, the namespace must be approved by authorized architects before the spec PR can merge.
+
+This process is automated via GitHub Actions on this repository.
+
+## How It Works
+
+### Detection
+
+When a PR modifies any `tspconfig.yaml` file, the analyze-code workflow:
```
raych1, line 13: Should the namespace approval process be handled only in the public spec repo?

samvaity: I can see this process applicable to private repos too, @praveenkuttappan can you confirm?

raych1: Private spec repo is mainly used for private preview version specs. When service teams work on public preview and GA, they will move specs to the public spec repo. Doing namespace review at that time might be reasonable.

On the other hand, there're also some internal services in the private spec repo and those specs might not need to follow the normal APEX process.

samvaity:
Hey @mikeharder — update on changes and testing:

### Changes since last review

1. **Switched from `tsp compile` to direct `tspconfig.yaml` parsing** — the typespec-metadata emitter fails in sparse checkout (sibling dirs not materialized). Now parsing emitter `options` directly with `js-yaml` for `namespace`, `package-details.name`, `module` (Go), `crate-name` (Rust).

2. **Added commit status as the single merge gate** — `workflow_run` checks don't show on PRs, so `status-check.js` now creates a `Namespace Approval` commit status (`pending`/`success`) on the PR head SHA. Follows the same pattern as `Swagger Avocado`, `ARM Auto SignOff`, etc.

3. **Fixed plane detection** — linter `extends` (e.g. `resource-manager` or `data-plane`) now overrides the path-based guess, preventing dual `Mgmt` + `data-plane` labels.

4. **Added `statuses: write`** permission to the status workflow.

### Testing (two-fork setup per your suggestion)

- Primary fork: `samvaity/azure-rest-api-specs` (workflows on `main`)
- Secondary fork: `samvaity-testing/azure-rest-api-specs` (opens cross-fork PRs)
- Test PR: https://github.com/samvaity/azure-rest-api-specs/pull/4

End-to-end flow verified:
- ✅ Analyze Code detects tspconfig.yaml changes, parses namespaces
- ✅ Set Status posts review comment with approval table + pending labels
- ✅ Commit status shows as 🟡 pending on the PR
- ✅ Management plane correctly detected via linter config
- ✅ Format validation flags naming issues

### CI failure

The `GitHub Actions - Test` failure on this PR is a pre-existing issue with `install-deps-github-script` (`--omit dev` reporting unmet devDependencies) — not caused by our changes.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
lirenhe: Similar question about the scope, we have main branch and release-* branches in public repos,  does this workflow apply to both or only to PRs for the main branch?

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,99 @@
...
+### Status + Approval
+
+The status workflow posts the namespace review comment, applies `<lang>-namespace-pending` labels, and validates architect approval labels.
+
+- **Data plane:** Each language architect approves their language by applying `<lang>-namespace-approved`
+- **Management plane:** ArthurMa1978 or m-nash can apply `namespace-approved-all` to approve all languages at once
+
+### Status Check
+
+A required status check (`namespace-approval`) blocks merge until all pending namespaces are approved. The check:
```
JiaqiZhang-Dev: Should we include this check in the summarize checks so that, if this required check fails, it appears under 'Next Steps to Merge' and developers can see that it blocks the PR merge?

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/summarize-checks.yaml

JiaqiZhang-Dev: Suggest update the check name to `Namespace Approval` so the name stays consistent with the `status-check.js` file

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
JiaqiZhang-Dev, line 1: Since this is a required check, could we also add the doc to the spec repo's wiki? https://github.com/Azure/azure-rest-api-specs/wiki, like the other checks, so the chatbot can understand the check logic and provide developers with a useful guide

Image

.github/workflows/src/namespace-approval/status-check.js
```diff
@@ -0,0 +1,54 @@
...
+  } else if (labels.includes("namespace-approved")) {
+    core.info("All namespaces approved - merge allowed");
+    await github.rest.repos.createCommitStatus({
+      owner,
+      repo,
+      sha: pr.head.sha,
+      state: "success",
+      context: "Namespace Approval",
+      description: "All namespaces approved",
+    });
```
JiaqiZhang-Dev, line 56: Could there be a case where both the pending label and approved label are missing? If so, we may need fallback logic, consider add an explicit `else` here

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,99 @@
+# Namespace Approval Process
+
+## Overview
+
+When a service team adds or modifies SDK namespace configuration in their `tspconfig.yaml`, the namespace must be approved by authorized architects before the spec PR can merge.
```
lirenhe, line 5: what if the tspconfig file is already incorrect in the spec repo and the current PR won't change the tspconfig.yaml file, will it detect the problem and ask for namespace reivew & approval?

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,99 @@
...
+
+## How It Works
+
+### Detection
+
+When a PR modifies any `tspconfig.yaml` file, the analyze-code workflow:
+
+1. Identifies which languages have namespace/package-name configuration
+2. Extracts namespaces by parsing `tspconfig.yaml` emitter options directly
+3. Determines if the service is data-plane or management-plane (via path and linter config)
```
lirenhe, line 17: Could this pipeline handle both in one PR? It won't happen often, but when there is a service rename, people may submit PR to update namespaces for data plane + mgmt. plane

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
```diff
@@ -0,0 +1,99 @@
...
+| Python               | @xirzec, @kashifkhan          |
+| TypeScript           | @xirzec, @maorleger           |
+| Go                   | @jhendrixMSFT                 |
+| Mgmt (all languages) | @ArthurMa1978, @m-nash        |
+
+<!-- cspell:enable -->
+
+## For Service Teams
+
+No action needed beyond your normal workflow:
```
lirenhe, line 83: Will any information be added into 'Next step to Merge' comment in the spec PR?

Do we have an SLA for the approval and if service team has an urgent request for approval, do they have a place to escalate?

.github/workflows/src/namespace-approval/status-check.js
samvaity: Good catch. Added an explicit `else` that sets `pending` status with description `Namespace review in progress`. This handles the edge case where labels are manually removed.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Fixed — updated the doc to use `Namespace Approval` (matching the commit status context name in `status-check.js`).

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Good idea. Will add a wiki entry as a [follow-up](https://github.com/Azure/azure-rest-api-specs/issues/44280) once the workflow is merged and stable.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Currently the workflow only runs on PRs targeting `main`. Release branches (`release-*`) are not in scope for this initial rollout. We can extend to release branches if needed — would that be useful?

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: No — the workflow only triggers when a `tspconfig.yaml` is modified in the PR. Pre-existing incorrect configs that aren't touched won't be detected. We have a [follow-up issue](https://github.com/Azure/azure-rest-api-specs/issues/44249) to add a namespace check at SDK generation time to catch those cases.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Currently, if a PR touches multiple TypeSpec projects, only the last project's namespaces are shown — earlier ones are silently overwritten since they share the same language keys. Deferring this to confirm from you if this is a real scenario that occurs. If so, we can restructure the data model to be per-project.

edit:  Thanks for confirming. Since it's rare but real, I'll track it as a follow-up to restructure the data model to support per-project namespaces. For now, if this scenario comes up, the workaround is to split into separate PRs.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Added Namespace Approval to the summarize-checks so it will now show in the 'Next Steps to Merge' comment.

There's no namespace approval SLA today — this didn't exist as a formal process before. Teams notifications to ping approvers (planned as a future extension) would be the natural place to address responsiveness, but that's not implemented yet.

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Thanks for confirming. Since it's rare but real, tracked as a follow-up: https://github.com/Azure/azure-rest-api-specs/issues/44281. For now, the workaround is to split into separate PRs (one per TypeSpec project).

.github/workflows/src/namespace-approval/NAMESPACE-REVIEW-PROCESS.md
samvaity: Added — registered `Namespace Approval` in `summarize-checks.js` CHECK_METADATA so it will appear in the 'Next Steps to Merge' comment when the check is pending/failing.

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,226 @@
...
+  try {
+    await access(path);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Sparse checkout only materializes .github/, so hydrate changed TypeSpec directories on demand.
```
mikeharder: why not just clone the whole repo?  the perf isn't **that** bad.

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,293 @@
...
+ */
+async function ensureLabel(github, owner, repo, label) {
+  try {
+    await github.rest.issues.getLabel({
+      owner,
+      repo,
+      name: label,
+    });
+  } catch {
+    await github.rest.issues.createLabel({
```
mikeharder: why does the action need to create labels?  shouldn't this be a one-time manual action in the repo?

.github/workflows/namespace-approval-code.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  detect-namespaces:
+    name: Analyze Code
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            .github/
```
mikeharder: if you need to analyze code under specification, better to just clone the whole repo

mikeharder (changes requested):

.github/workflows/src/namespace-approval/post-results.js
samvaity: removed

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,293 @@
...
+  const labelsToAdd = new Set(["namespace-review-required"]);
+  if (results.isMgmt) {
+    labelsToAdd.add("Mgmt");
+  }
+  if (results.isDataPlane) {
+    labelsToAdd.add("data-plane");
+  }
+  for (const language of languages) {
+    labelsToAdd.add(`${language}-namespace-pending`);
+  }
```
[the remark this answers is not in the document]
samvaity: selective reset added

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,211 @@
...
+      prNumber: payload.pull_request.number,
+      action: payload.action,
+    };
+    const resultsPath = join(process.env.RUNNER_TEMP || ".", "namespace-results.json");
+    await writeFile(resultsPath, JSON.stringify(results, null, 2));
+    core.setOutput("results", "true");
+    core.setOutput("results_path", resultsPath);
+  } catch (error) {
+    core.setFailed(`Namespace detection failed: ${String(error)}`);
+  }
```
mikeharder: why catch and set failed, instead of just letting the exception bubble up?

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/**
+ * Detect namespace changes in a PR and write results to an artifact file.
+ *
+ * @param {import("@actions/github-script").AsyncFunctionArguments} args
+ */
+export default async function detectNamespaces({ context, core }) {
+  try {
+    const payload = /** @type {import("@octokit/webhooks-types").PullRequestEvent} */ (
```
mikeharder: use helper method extractInputs() rather than directly reading from context, which does useful things like validating inputs, enabling github logging hooks, etc.

https://github.com/Azure/azure-rest-api-specs/blob/6ce2ed198fadbae77546e9178ab42d8de14a1f5a/.github/workflows/src/context.js#L22

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,211 @@
...
+      context.payload
+    );
+
+    const changedFiles = (
+      await getChangedFiles({
+        cwd: process.env.GITHUB_WORKSPACE ?? process.cwd(),
+        paths: ["specification"],
+      })
+    ).filter((file) => /specification\/.*\/tspconfig\.yaml$/.test(file));
```
mikeharder: move this filter to a helper method under changed-files.js, extracted from the existing typespec() filter:

https://github.com/Azure/azure-rest-api-specs/blob/6ce2ed198fadbae77546e9178ab42d8de14a1f5a/.github/shared/src/changed-files.js#L280

.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -0,0 +1,211 @@
...
+    const results = {
+      namespacesFound,
+      artifactNames,
+      isMgmt,
+      isDataPlane,
+      formatResults,
+      prNumber: payload.pull_request.number,
+      action: payload.action,
+    };
+    const resultsPath = join(process.env.RUNNER_TEMP || ".", "namespace-results.json");
```
mikeharder: why do you need to fallback to "."?  RUNNER_TEMP should always be set when the workflow is executing.

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+ * @returns {Promise<ApproversConfig | null>}
+ */
+async function loadApproversConfig(core) {
+  try {
+    const content = await readFile(".github/namespace-approvers.yml", "utf8");
+    return /** @type {ApproversConfig} */ (yaml.load(content));
+  } catch (error) {
+    core.setFailed("Failed to load .github/namespace-approvers.yml: " + String(error));
+    return null;
+  }
```
mikeharder: i think it's better to just let the exception bubble

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/**
+ * @param {import("@actions/github-script").AsyncFunctionArguments} args
+ */
+export default async function postResults({ github, context, core }) {
+  const { owner, repo, issue_number, run_id } = await extractInputs(github, context, core);
+  const approversConfig = await loadApproversConfig(core);
+  if (!approversConfig) {
+    return;
```
mikeharder: This looks like an anti-pattern of silently ignoring an error.  I think best would be for loadApproversConfig() to throw, and this code should let the exception bubble up.  Or loadApproversConfig() could return null (if it's useful in some contexts), then this code needs to throw if it returns null.

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  const results = run_id ? await downloadNamespaceResults(github, core, owner, repo, run_id) : null;
+  if (!results) {
```
mikeharder: also looks like anti-pattern of silently ignoring an error

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+    repo,
+    run_id: runId,
+    name: "namespace-results",
+    per_page: PER_PAGE_MAX,
+  });
+
+  const artifact = artifacts.sort(
+    (
+      /** @type {{ updated_at?: string | null }} */ left,
+      /** @type {{ updated_at?: string | null }} */ right,
```
mikeharder: i dont' think these custom type declarations should be needed?  `artifacts` should already be strongly typed.  for example, they're not needed here:

https://github.com/Azure/azure-rest-api-specs/blob/6ce2ed198fadbae77546e9178ab42d8de14a1f5a/.github/workflows/src/summarize-checks/summarize-checks.js#L1121

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+    ) => {
+      const leftTime = new Date(left.updated_at ?? 0).getTime();
+      const rightTime = new Date(right.updated_at ?? 0).getTime();
+      return rightTime - leftTime;
+    },
+  )[0];
+
+  if (!artifact) {
+    core.info(`No namespace-results artifact found for run ${runId}`);
+    return null;
```
mikeharder: if this is truly unexpected, throw an exception instead of returning null

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  const download = await github.rest.actions.downloadArtifact({
+    owner,
+    repo,
+    artifact_id: artifact.id,
+    archive_format: "zip",
+  });
+
+  const zipPath = join(process.env.RUNNER_TEMP || tmpdir(), `namespace-results-${runId}.zip`);
```
mikeharder: why fallback to tmpdir()?  RUNNER_TEMP should always be set.

.github/workflows/src/namespace-approval/post-results.js
```diff
@@ -0,0 +1,388 @@
...
+ */
+function getApprovers(approversConfig, isMgmt, language) {
+  if (isMgmt) {
+    const mgmtApprovers = approversConfig["management-plane"]?.all;
+    if (mgmtApprovers) {
+      return mgmtApprovers;
+    }
+  }
+
+  return approversConfig["data-plane"]?.[language] ?? ["TBD"];
```
mikeharder: is this "TBD" a placeholder you intended to update before merging?

.github/workflows/src/namespace-approval/validate-approval.js
```diff
@@ -0,0 +1,357 @@
...
+  let approversConfig;
+  try {
+    const content = await readFile(".github/namespace-approvers.yml", "utf8");
+    approversConfig = /** @type {ApproversConfig} */ (yaml.load(content));
+  } catch (error) {
+    core.setFailed("Failed to load .github/namespace-approvers.yml: " + String(error));
+    return;
+  }
+
+  const payload =
```
mikeharder, line 258: use extractInputs()

.github/workflows/src/namespace-approval/validate-approval.js
```diff
@@ -0,0 +1,357 @@
...
+ * @param {import("@actions/github-script").AsyncFunctionArguments} args
+ */
+export default async function validateApproval({ github, context, core }) {
+  /** @type {ApproversConfig} */
+  let approversConfig;
+  try {
+    const content = await readFile(".github/namespace-approvers.yml", "utf8");
+    approversConfig = /** @type {ApproversConfig} */ (yaml.load(content));
+  } catch (error) {
+    core.setFailed("Failed to load .github/namespace-approvers.yml: " + String(error));
```
mikeharder: better to throw

.github/workflows/src/namespace-approval/post-results.js
samvaity: removed the TBD as we have all the required approvers defined in the config.

praveenkuttappan (approved):
LGTM

.github/shared/src/changed-files.js
```diff
@@ -281,6 +281,18 @@ export function typespec(file) {
   );
 }
 
+/**
```
mikeharder: I will refactor this to create sub-functions "tspconfig()" and "tsp()" that can be combined into the existing "typespec()" function.

mikeharder (approved):

PR #44118, opened 2026-06-20
Make Name field match what service returns

matthchr:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Other, please clarify: Fix spec to match actual service return, resolves longstanding issue: https://github.com/Azure/azure-sdk-for-go/issues/23342

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.

### DIFF
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
@@ -213,6 +213,7 @@ model ResourceSkuZoneDetails {
    * The set of zones that the SKU is available in with the specified capabilities.
    */
   @visibility(Lifecycle.Read)
+  @encodedName("application/json", "Name")
   name?: string[];
 
   /**
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkus.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkus.json
@@ -83,7 +83,7 @@
                 ],
                 "zoneDetails": [
                   {
-                    "name": [
+                    "Name": [
                       "2"
                     ],
                     "capabilities": [
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkusForARegion.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkusForARegion.json
@@ -84,7 +84,7 @@
                 ],
                 "zoneDetails": [
                   {
-                    "name": [
+                    "Name": [
                       "2"
                     ],
                     "capabilities": [
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkusWithExtendedLocations.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/examples/skus/ListAvailableResourceSkusWithExtendedLocations.json
@@ -84,7 +84,7 @@
                 ],
                 "zoneDetails": [
                   {
-                    "name": [
+                    "Name": [
                       "2"
                     ],
                     "capabilities": [
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/skus.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/skus.json
@@ -421,13 +421,14 @@
       "type": "object",
       "description": "Describes The zonal capabilities of a SKU.",
       "properties": {
-        "name": {
+        "Name": {
           "type": "array",
           "description": "The set of zones that the SKU is available in with the specified capabilities.",
           "items": {
             "type": "string"
           },
-          "readOnly": true
+          "readOnly": true,
+          "x-ms-client-name": "name"
         },
         "capabilities": {
           "type": "array",


specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
```diff
@@ -213,7 +213,7 @@ model ResourceSkuZoneDetails {
    * The set of zones that the SKU is available in with the specified capabilities.
    */
   @visibility(Lifecycle.Read)
-  name?: string[];
+  Name?: string[];
```
ravimeda: **[NEW] 🔴 Blocking** **[[typespec-project §4.1 Naming](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-project.instructions.md#41-naming-conventions)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp` — line 216 — Property `Name` (PascalCase) violates the camelCase JSON property-name rule.

**Classification reasoning:** NEW — this PR renames the property from `name` to `Name`; the base version (SHA `18ebae7`) declares `name`.

This produces an **unsuppressed** `@azure-tools/typespec-azure-core/casing-style` lint failure — compare `meterID` at line 250, which carries an explicit `#suppress "@azure-tools/typespec-azure-core/casing-style" "backwards compatibility"`. To match the service's `Name` wire format without a PascalCase property name, set the wire name with `@encodedName` and keep the property camelCase:

**Suggested fix:**
```tsp
@visibility(Lifecycle.Read)
@encodedName("application/json", "Name")
name?: string[];
```

This keeps the TypeSpec/SDK property `name`, emits `Name` on the wire/swagger, and keeps `@identifiers(#["name"])` (line 192) referencing a property that still exists. Verify the emitted `x-ms-identifiers` matches the wire name the emitter produces (see the related top-level comment on line 192). If the team has a specific reason to surface `Name` directly to SDKs, please document it.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`8412021`](https://github.com/Azure/azure-rest-api-specs/pull/44118/commits/84120216679e24ac4ac2a34493ac955aba2d8ede). See the inline comment for finding 1. Findings 2–4 below concern a line and a file outside this PR's diff (`models.tsp` line 192 and the generated `skus.json`), so they are posted here as a top-level comment.

---

**2. [NEW] 🔴 Blocking** **[[XmsIdentifierValidation (R4041)](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/linter-rule-coverage.md)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp` — line 192 — `@identifiers(#["name"])` references a renamed property.

The `zoneDetails?: ResourceSkuZoneDetails[]` array (line 193, model `ResourceSkuLocationInfo`) declares `@identifiers(#["name"])`, pointing at `ResourceSkuZoneDetails.name`. This PR renamed that property to `Name`, so after the swagger regenerates, the emitted `x-ms-identifiers: ["name"]` on `zoneDetails` will reference a non-existent property → `XmsIdentifierValidation` failure. If the rename is kept, update to `@identifiers(#["Name"])`. If the `@encodedName` fix in finding 1 is adopted, the property stays `name` so this decorator is valid — verify the emitted `x-ms-identifiers` resolves to the wire name. (The second `@identifiers(#["name"])` at line 222 targets `ResourceSkuCapabilities.name`, unchanged — not affected.)

**3. [NEW] 🔴 Blocking** **[[TypeSpec generated output must match source](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-project.instructions.md)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/skus.json` — `ResourceSkuZoneDetails`, line 413 — Generated swagger not regenerated.

The PR changes only `models.tsp`; the committed `skus.json` still declares `name`. TypeSpec source and generated OpenAPI are out of sync, so the **TypeSpec Validation** CI check (which regenerates and diffs) will fail. Regenerate the `ComputeSku` project and commit the updated `skus.json`.

**4. [NEW] 🟠 Warning** **[[Breaking changes guidelines](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/Breaking%20changes%20guidelines.md)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2021-07-01/skus.json` — `ResourceSkuZoneDetails.name`, line 413 — In-place breaking change to a released stable version.

Renaming a property in the already-shipped stable `2021-07-01` is a wire-contract breaking change that OAD will flag once the swagger regenerates, and released versions are normally immutable. The service-correctness rationale supports an in-place fix, but the breaking-change review/suppression flow must be followed and ARM should confirm an in-place edit (vs. a new version) is acceptable.

---

**Worth surfacing — the underlying tension:** The service returns PascalCase `Name`, so any faithful spec carries a PascalCase **wire** name. `@encodedName` keeps the TypeSpec/SDK property camelCase, but the emitted swagger property is still `Name`, which may itself need a swagger-level naming suppression in LintDiff. This tension can only be *managed* (encodedName + breaking-change approval + possibly a suppression), not eliminated in the spec — the truly correct long-term fix is service-side (return `name`).

matthchr:
Good comments @ravimeda, I've fixed them and pushed. Also rebased so this is based on latest main to get rid of the "out of date branch" comment

matthchr:
Here's an example raw REST response that shows the current response mismatches with the documented API shape:
```
    {
      "capabilities": [
        {
          "name": "MaxSizeGiB",
          "value": "65536"
        },
        {
          "name": "MinSizeGiB",
          "value": "1"
        },
        {
          "name": "MaxIOpsReadWrite",
          "value": "80000"
        },
        {
          "name": "MinIOpsReadWrite",
          "value": "3000"
        },
        {
          "name": "MaxBandwidthMBpsReadWrite",
          "value": "2000"
        },
        {
          "name": "MinBandwidthMBpsReadWrite",
          "value": "125"
        },
        {
          "name": "MaxValueOfMaxShares",
          "value": "15"
        },
        {
          "name": "MinIOSizeKiBps",
          "value": "4"
        },
        {
          "name": "MaxIOSizeKiBps",
          "value": "256"
        },
        {
          "name": "MinIopsReadOnly",
          "value": "3000"
        },
        {
          "name": "MaxIopsReadOnly",
          "value": "80000"
        },
        {
          "name": "MinBandwidthMBpsReadOnly",
          "value": "125"
        },
        {
          "name": "MaxBandwidthMBpsReadOnly",
          "value": "2000"
        },
        {
          "name": "MaxIopsPerGiBReadWrite",
          "value": "500"
        },
        {
          "name": "MaxIopsPerGiBReadOnly",
          "value": "500"
        },
        {
          "name": "MinIopsPerGiBReadWrite",
          "value": "1"
        },
        {
          "name": "MinIopsPerGiBReadOnly",
          "value": "1"
        },
        {
          "name": "BillingPartitionSizes",
          "value": "1, 2, 3, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 3072, 4096, 5120, 6144, 7168, 8192, 9216, 10240, 11264, 12288, 13312, 14336, 15360, 16384, 17408, 18432, 19456, 20480, 21504, 22528, 23552,24576,25600,26624,27648,28672,29696,30720,31744,32768,33792,34816,35840,36864,37888,38912,39936,40960,41984,43008,44032,45056,46080,47104,48128,49152,50176,51200,52224,53248,54272,55296,56320,57344,58368,59392,60416,61440,62464,63488,64512,65536"
        },
        {
          "name": "MaxZonalFaultDomainCount",
          "value": "2"
        },
        {
          "name": "PlatformFaultDomainCount",
          "value": "2"
        }
      ],
      "locationInfo": [
        {
          "location": "eastus2euap",
          "zoneDetails": [
            {
              "Name": [
                "3"
              ],
              "capabilities": [
                {
                  "name": "MaxFaultDomainCountInZone",
                  "value": "2"
                }
              ]
            },
            {
              "Name": [
                "2"
              ],
              "capabilities": [
                {
                  "name": "MaxFaultDomainCountInZone",
                  "value": "2"
                }
              ]
            }
          ],
          "zones": [
            "3",
            "2"
          ]
        }
      ],
      "locations": [
        "eastus2euap"
      ],
      "name": "PremiumV2_LRS",
      "resourceType": "disks",
      "restrictions": [],
      "size": "P",
      "tier": "Premium"
    },
```

matthchr:
Here's an AI report of this issue's impact in the various SDKs:

# Azure SDK Case-Sensitivity Report: `ResourceSkuZoneDetails.name`

## Background

- **Issue** ([Azure/azure-sdk-for-go#23342](https://github.com/Azure/azure-sdk-for-go/issues/23342)): The Compute `Resource Skus - List` API returns the zone-details field as **`Name`** (capital N), but the OpenAPI/TypeSpec spec declared it as **`name`** (lowercase). The Go SDK deserializes only `name`, so `ZoneDetails[].Name` always came back empty.
- **Fix** ([Azure/azure-rest-api-specs#44118](https://github.com/Azure/azure-rest-api-specs/pull/44118/files)): Change the *wire name* to `Name` (via `@encodedName("application/json", "Name")` in TypeSpec / `"Name"` + `"x-ms-client-name": "name"` in the swagger) so the generated code reads the key the service actually sends, while still exposing the property to users as `name`.

## Conclusion (headline)

**All three SDKs — Go, C#, and Python — are case-sensitive on this field.** Go is the one that got reported, but C# and Python were equally broken. None of them would populate the field when the service returns `Name`. This is why the fix had to be made in the spec (change the wire name) rather than per-language — a case-insensitive language would not have needed the fix.

---

## 1. Go — case-sensitive ❌ (confirmed broken)

Generated `UnmarshalJSON` uses a `switch` over the raw JSON key with a literal `"name"` case:

[`sdk/resourcemanager/compute/armcompute/models_serde.go` L11106–11117](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/models_serde.go#L11106-L11117)
```go
func (r *ResourceSKUZoneDetails) UnmarshalJSON(data []byte) error {
        ...
        for key, val := range rawMsg {
                switch key {
                case "capabilities":
                        err = unpopulate(val, "Capabilities", &r.Capabilities)
                case "name":                          // <-- exact, lowercase match only
                        err = unpopulate(val, "Name", &r.Name)
                }
        }
}
```
A Go `switch` on a string does an exact comparison, so `"Name"` never matches `case "name"`. The issue author proved this — his repro prints `0` for `len(d.Name)`.

## 2. C# — case-sensitive ❌ (broken)

The System.Text.Json–based generated deserializer matches with `JsonProperty.NameEquals("name"u8)`:

[`sdk/compute/Azure.ResourceManager.Compute/src/Generated/Models/ComputeResourceSkuZoneDetails.Serialization.cs` L147–189](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/compute/Azure.ResourceManager.Compute/src/Generated/Models/ComputeResourceSkuZoneDetails.Serialization.cs#L147-L189)
```csharp
foreach (var prop in element.EnumerateObject())
{
    if (prop.NameEquals("name"u8))   // ordinal, case-sensitive UTF-8 byte compare
    {
        ...
        name = array;
        continue;
    }
    if (prop.NameEquals("capabilities"u8)) { ... }

    if (options.Format != "W")
    {
        // unmatched keys ("Name") fall through to here and are discarded into raw data
        additionalBinaryDataProperties.Add(prop.Name, BinaryData.FromString(prop.Value.GetRawText()));
    }
}
```
`JsonProperty.NameEquals(ReadOnlySpan<byte>)` performs an **ordinal (case-sensitive)** byte comparison — it is not affected by any `PropertyNameCaseInsensitive` option because the matching is hand-written, not reflection-based. So a wire key of `Name` fails all `NameEquals` checks, `name` stays `default` (null → empty `ChangeTrackingList`), and the `Name` value is silently swept into `additionalBinaryDataProperties`. **Broken, same as Go.**

## 3. Python — case-sensitive ❌ (broken)

The current Compute SDK uses the new TypeSpec `_Model` base (not legacy msrest). The field is a plain `rest_field()` with no explicit wire name:

[`sdk/compute/azure-mgmt-compute/azure/mgmt/compute/models/_models.py` L14783](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/compute/azure-mgmt-compute/azure/mgmt/compute/models/_models.py#L14783)
```python
class ResourceSkuZoneDetails(_Model):
    name: Optional[list[str]] = rest_field(visibility=["read"])   # wire name defaults to "name"
```

`_Model` stores the raw JSON dict verbatim and resolves attributes lazily through the descriptor, using a plain dict lookup on the wire name:

[`azure/mgmt/compute/_utils/model_base.py` — `_RestField.__get__` L1413–1417](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/compute/azure-mgmt-compute/azure/mgmt/compute/_utils/model_base.py#L1413-L1417) and `_rest_name` L1406–1410:
```python
@property
def _rest_name(self) -> str:
    if self._rest_name_input is None:
        raise ValueError(...)
    return self._rest_name_input          # == "name" (defaults to the attribute name)

def __get__(self, obj, type=None):
    item = obj._data.get(self._rest_name, _UNSET)   # obj._data is the raw JSON dict
    if item is _UNSET:
        return self._default if ... else None       # -> None when key is "Name"
```
`obj._data.get("name")` is an ordinary, **case-sensitive** dict lookup against the raw payload. When the service sends `Name`, `_data` has key `"Name"`, so `.get("name")` misses and `.name` returns `None`. **Broken, same as Go and C#.** (Note: the older msrest-based Python generation was likewise case-sensitive via exact attribute-map keys, so this is not a regression.)

---

## Summary table

| SDK | Read mechanism | Match | Result when service returns `Name` |
|-----|----------------|-------|-------------------------------------|
| **Go** | `switch key { case "name": }` | exact string | ❌ empty (`len==0`) |
| **C#** | `JsonProperty.NameEquals("name"u8)` | ordinal / case-sensitive bytes | ❌ null; value dumped into raw `additionalBinaryDataProperties` |
| **Python** | `_data.get("name")` dict lookup | case-sensitive key | ❌ `None` |

**Bottom line:** none of Go, C#, or Python are case-insensitive here. The correct fix is exactly what PR #44118 does — align the *spec wire name* to `Name` (keeping the client-facing name `name`), which regenerates all three SDKs to read the key the service actually returns.

matthchr:
Here's two bugs this would fix:
https://github.com/Azure/azure-sdk-for-go/issues/25068
https://github.com/Azure/azure-sdk-for-go/issues/23342

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
```diff
@@ -213,6 +213,7 @@ model ResourceSkuZoneDetails {
    * The set of zones that the SKU is available in with the specified capabilities.
    */
   @visibility(Lifecycle.Read)
+  @encodedName("application/json", "Name")
```
markcowl: Never use `@encodedName`  if you want to preserve a name for emitted clients, use the `@clientName` decorator as an augment decorator from client.tsp.

markcowl (changes requested):
First, this is an API breaking change, so will need breaking change review.  Second, *always* have the type intypespec match the casing of the wire type - `@encodedName` is strongly discouraged, use `@clientName("name")` to prevent SDK breaks instead.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
```diff
@@ -213,6 +213,7 @@ model ResourceSkuZoneDetails {
...
   name?: string[];
```
markcowl, line 217:
```suggestion
  Name?: string[];
```

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
markcowl, line 216:
```suggestion
  @clientName("name")
```

-- this might require some sdk-specific scope, like `@clientName("name", "!csharp, !java")`

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
matthchr: Done

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
matthchr: Added clientName

markcowl (approved):
This looks good.  We should check that this does not cause a specific breaking change in csharp & java generated clients.  Barring that, this would only need ARM breaking change approval.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeSku/models.tsp
gary-x-li: Earlier guidance correctly recommended preserving the SDK-facing property name as `name`, but placed `@clientName("name")` directly in `models.tsp`. Current repository guidance requires client customizations to live in `client.tsp`.

Please keep the wire property as `Name`, remove the client-generator import, `using`, and inline decorator from `models.tsp`, then apply:

```tsp
@@clientName(ResourceSkuZoneDetails.Name, "name");
```

from `client.tsp`, imported by `main.tsp`. This changes only the customization's location; preserving the SDK-facing name remains correct.

gary-x-li:
Please take a look at this comment to see if it makes sense: https://github.com/Azure/azure-rest-api-specs/pull/44118#discussion_r4088364432

Also, please fix the LintDiff errors by suppressing it.

matthchr:
Thanks @gary-x-li - I've fixed your comments I believe.

PR #44139, opened 2026-06-22
[KeyVault] Mark generated C# Secrets and Certificates transport types as internal via @access

rohitsinghal4u:
## Why

The customer-facing surface for `Azure.Security.KeyVault.Secrets` and `Azure.Security.KeyVault.Certificates` in .NET is the hand-written `SecretClient` / `CertificateClient`. The TypeSpec-generated `KeyVaultSecretsClient` / `KeyVaultCertificatesClient` and their parameter/response models are an implementation detail consumed only from inside the same assembly.

Today, the .NET SDK repo runs a post-emit PowerShell script (`Internalize-Generated.ps1`) in each library to regex-rewrite `public` to `internal` on every generated type. This is fragile (silently no-ops if the emitter renames anything) and was flagged by `@JoshLove-msft` as not the right approach on https://github.com/Azure/azure-sdk-for-net/pull/59972 / Certificates review thread.

## What

Adds a single line to each of:

- `specification/keyvault/data-plane/Secrets/client.tsp`
- `specification/keyvault/data-plane/Certificates/client.tsp`

```tsp
@@access(KeyVault, Access.internal, "csharp");
```

This marks the generated transport client + all spec-defined models as `internal` at emit time for the C# emitter only. Other languages (Java/Python/Go/JavaScript/Rust) are unaffected.

## Prior art

Identical pattern already used by sibling `Azure.Security.KeyVault.Administration`:

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/data-plane/Administration/client.tsp#L133

```tsp
@@access(KeyVault, Access.internal, "csharp");
```

## Impact

- **C#**: `KeyVaultSecretsClient` / `KeyVaultCertificatesClient` and all generated models are `internal`. Lets the .NET SDK remove `eng/Internalize-Generated.ps1` and the MSBuild target that runs it from both libraries.
- **Java / Python / Go / JavaScript / Rust**: no change. The decorator is scoped to `"csharp"`.
- **Wire / api-versions**: no change. This is a generator-visibility hint only; no operations, models, or properties are added/removed/modified.

### DIFF
--- a/specification/keyvault/data-plane/Certificates/client.tsp
+++ b/specification/keyvault/data-plane/Certificates/client.tsp
@@ -43,6 +43,8 @@ namespace Customizations;
 );
 @@clientName(IssuerParameters, "IssuerParametersBundle", "csharp");
 
+@@access(KeyVault, Access.internal, "csharp");
+
 using KeyVault;
 
 // Make optional path parameter required for all languages
--- a/specification/keyvault/data-plane/Secrets/client.tsp
+++ b/specification/keyvault/data-plane/Secrets/client.tsp
@@ -22,6 +22,8 @@ namespace Customizations;
 @@clientName(KeyVault, "KeyVaultSecretsClient", "csharp");
 @@clientName(SecretAttributes, "SecretAttributesBundle", "csharp");
 
+@@access(KeyVault, Access.internal, "csharp");
+
 using KeyVault;
 
 // Make optional path parameter required for all languages


nnoboa (approved):

msftemployee:
**Consider whether `@@usage` overrides are also needed for C#, alongside `@@access`.**

The pattern from `Administration` is actually two coordinated changes, not one: `@@access(KeyVault, Access.internal, "csharp")` (#32571) was followed months later by a standalone `@@usage(KeyVault.RoleDefinition, Usage.input | Usage.output, "csharp")` (#35453, _"Add usage to RoleDefinition for csharp"_). This PR adds the `@@access` half but none of the `@@usage` half.

Why it can matter: internalizing isn't just a `public`->`internal` keyword flip (all that `Internalize-Generated.ps1` did). For an `internal` model, the C# emitter only generates the serialization direction(s) implied by the model's computed usage. A model that is output-only in the spec but serialized by the hand-written `SecretClient`/`CertificateClient` compiles fine today (public) yet can lose its serialization once internal — breaking the azure-sdk-for-net build after the PS script is removed.

Certificates looks higher-risk: its Java config already force-flags `CertificateOperation`, `IssuerBundle`, and `CertificateIssuerItem` as `Usage.input` — output-only models a real client serializes. There's no equivalent `"csharp"` override here.

Suggestion: before the .NET repo deletes `Internalize-Generated.ps1`, emit C# locally and diff the generated serialization for those models (plus `SecretBundle`/backup-restore) against the current PS-script output. If any drops a direction, add `@@usage(<Model>, Usage.input | Usage.output, "csharp")`. Validating against a green azure-sdk-for-net build (#59972 for Secrets; no Certificates consumer found yet) would confirm.

PR #43467, opened 2026-05-23
Decouple toolbox tool schemas

viswabalaji:
## Summary
- Add a dedicated `ToolboxTool` hierarchy for toolbox version payloads, separate from `OpenAI.Tool`.
- Limit toolbox-supported tool variants to the public toolbox set plus WorkIQ, FabricIQ, and internal toolbox search.
- Keep `tool_configs` on `ToolboxTool` instead of OpenAI/agent tool schemas.
- Keep toolbox-only `toolbox_search_preview` out of `OpenAI.Tool[]`, which is also used by prompt/hosted agents.
- Make toolbox file search `vector_store_ids` optional while leaving OpenAI file search unchanged.
- Support `custom_search_configuration` on both prompt-agent `WebSearchTool` and toolbox `WebSearchToolboxTool`.
- Move toolbox-specific types into `src/toolboxes/models.tsp` and use standard TypeSpec spread helpers instead of experimental copy decorators for toolbox tool shapes.
- Regenerate Foundry OpenAPI outputs.
- Standardize top-level tool `project_connection_id` fields through shared optional/required aliases instead of a MCP-only extension plus direct field definitions.

## Split from optional-tools change
This PR intentionally does **not** make toolbox `tools` optional and does **not** add toolbox `skills`. That smaller change is isolated in #43526 so reviewers can evaluate the request/response optionality separately from the OpenAI/toolbox schema decoupling.

## SDK caller experience
Before, toolbox version creation reused `OpenAI.Tool[]`, so callers could pass the same broad tool set used by prompt/hosted agents. That meant unsupported toolbox tools like SharePoint grounding could appear valid in typed SDK code:

```python
from azure.ai.projects.models import SharepointGroundingToolParameters, SharepointPreviewTool, ToolProjectConnection

# Before: toolbox tools were modeled as OpenAI.Tool[], so agent-only tools like SharePoint grounding could type-check.
toolbox_version = project.beta.toolboxes.create_toolbox_version(
    toolbox_name="my-toolbox",
    tools=[
        SharepointPreviewTool(name="corp-sharepoint", sharepoint_grounding=SharepointGroundingToolParameters(project_connections=[ToolProjectConnection(project_connection_id="sharepoint-connection")]))
    ],
)
```

[...]

PR #44208, opened 2026-06-25
Fix: deprecate name, description, and tool_configs on GA tool classes

NivedhaChellam:
## Summary
Restores `name`, `description`, and `tool_configs` fields on General Availability tool definitions with `#deprecated` directive, preventing SDK breaking changes while preserving the intended migration to `ToolboxTool`.

## Why
PR #43467 removed these fields from tool classes and moved them exclusively to `ToolboxTool`. This introduces a breaking change for existing SDK consumers (Python, C#, Java) who rely on setting `name`, `description`, or `tool_configs` directly on tool objects like `CodeInterpreterTool`, `WebSearchTool`, etc.

Instead of removing outright, these fields are now deprecated—maintaining backward compatibility while signaling the migration path to `ToolboxTool`.

## What changed
- Added `DeprecatedToolNameAndDescriptionExtension` and `DeprecatedToolConfigExtension` aliases in `tools/models.tsp` with `#deprecated` directives  
- Re-applied deprecated fields to all GA tools via `@@copyProperties` (OpenAI tools) and direct spread (Azure tools)  
- Regenerated OpenAPI output files  

## Affected GA tools

| Tool                          | name / description | tool_configs |
|-------------------------------|--------------------|--------------|
| OpenAI.WebSearchTool          | ✅                 | ✅           |
| OpenAI.CodeInterpreterTool    | ✅                 | ✅           |
| OpenAI.ImageGenTool           | ✅                 | ✅           |
| OpenAI.FileSearchTool         | ✅                 | ✅           |
| OpenAI.LocalShellToolParam    | ✅                 | ✅           |
| OpenAI.FunctionShellToolParam | ✅                 | ✅           |
| BingGroundingTool             | ✅                 | ✅           |
| AzureAISearchTool             | ✅                 | ✅           |
| CaptureStructuredOutputsTool  | ✅                 | ✅           |
| OpenAI.MCPTool                | —                  | ✅           |
| AzureFunctionTool             | —                  | ✅           |
| OpenApiTool                   | —                  | ✅           |

## Validation
- `npx tsp compile .` succeeds with no errors  
- Generated OpenAPI output shows `deprecated: true` on all affected fields  
- `ToolboxTool` retains non-deprecated versions of these fields (intended home)

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
@@ -45,6 +45,38 @@ namespace Azure.AI.Projects {
   #suppress "@azure-tools/typespec-azure-core/experimental-feature" "Using experimental feature"
   @@copyVariants(OpenAI.IncludeEnum, Azure.AI.Projects._AgentIncludable);
 
+  /**
+   * Deprecated name and description fields retained for SDK backward compatibility.
+   * These fields now live on ToolboxTool and should not be used on tool definitions.
+   */
+  alias DeprecatedToolNameAndDescriptionExtension = {
+    /**
+     * Optional user-defined name for this tool or configuration.
+     */
+    #deprecated "Use name on ToolboxTool instead."
+    name?: string;
+
+    /**
+     * Optional user-defined description for this tool or configuration.
+     */
+    #deprecated "Use description on ToolboxTool instead."
+    description?: string;
+  };
+
+  /**
+   * Deprecated tool_configs field retained for SDK backward compatibility.
+   * This field now lives on ToolboxTool and should not be used on tool definitions.
+   */
+  alias DeprecatedToolConfigExtension = {
+    /**
+     * Per-tool configuration map. Keys are tool names or `*` (catch-all default).
+     * Resolution order: exact tool name match takes priority over `*`.
+     * Unknown tool names are silently ignored at runtime.
+     */
+    #deprecated "Use tool_configs on ToolboxTool instead."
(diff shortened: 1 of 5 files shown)


viswabalaji (approved):

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -45,6 +45,36 @@ namespace Azure.AI.Projects {
...
   @@copyVariants(OpenAI.IncludeEnum, Azure.AI.Projects._AgentIncludable);
 
+  /**
+   * Deprecated name and description fields retained for SDK backward compatibility.
+   */
+  alias DeprecatedToolNameAndDescriptionExtension = {
+    /**
+     * Optional user-defined name for this tool or configuration.
+     */
+    #deprecated "This property is deprecated and will be removed in a future version."
```
dargilco, line 55: I don't think the SDK emitters do anything with `#deprecated` (but I will check to confirm). So the only thing shown in SDKs is the doc string. Therefore the doc string should indicate it's a deprecated. I would just write `/* Deprecated. This property is deprecated and will be removed in a future version. */`. I don't think there is a need to explain what this field was used for.

Same for all other deprecated properties.

NivedhaChellam: Agreed. The doc comment itself should say "**Deprecated**" — that way SDK users see it regardless of emitter support. Updated the comments.

glecaros (approved):

PR #44217, opened 2026-06-25
[automation] Fix SourceControls.create operation TypeSpec modeling

### LINKED ISSUE #44216: [BUG] SourceControls.create uses ArmResourceRead for a PUT create operation
### API Spec link

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/automation/Automation.Management/SourceControl.tsp#L94

### API Spec version

2024-10-23

### Describe the bug

The Automation TypeSpec currently models the `SourceControls.create` operation as a read operation template with a PUT override:

```typespec
@tag("SourceControlSyncJob")
@OpenAPI.externalDocs("http://aka.ms/azureautomationsdk/sourcecontrolsyncjoboperations")
@put
create is ArmResourceRead<
  SourceControl,
  BaseParameters = Azure.ResourceManager.Foundations.DefaultBaseParameters<SourceControl>,
  Response = ArmResourceCreatedSyncResponse<SourceControlSyncJob>,
  Parameters = SourceControlSyncJobIdParameter & {
    @body
    parameters: SourceControlSyncJobCreateParameters;
  }
>;
```
This is semantically inconsistent:
- The operation is named `create`
- The HTTP verb is `PUT`
- The response is `ArmResourceCreatedSyncResponse<SourceControlSyncJob>`
- But the operation template is `ArmResourceRead<SourceControl>`

This makes the TypeSpec harder to reason about and could lead to incorrect SDK/resource metadata behavior.

### Expected behavior

The `create` operation should be modeled with a TypeSpec template that matches its actual semantics.
The operation should not use `ArmResourceRead` for a PUT create operation.

### Actual behavior

The create operation used the `ArmResourceRead` template.

### Reproduction Steps

Open `specification/automation/Automation.Management/SourceControl.tsp` and inspect `SourceControls.create`.

### Environment

_No response_

melina5656:
Fixes https://github.com/Azure/azure-rest-api-specs/issues/44216


specification/automation/Automation.Management/SourceControl.tsp
```diff
@@ -88,12 +88,13 @@ interface SourceControls {
    * Creates the sync job for a source control.
    */
   #suppress "@azure-tools/typespec-azure-core/no-openapi" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
```
pshao25, line 102: Why do we need this suppression while previously don't

melina5656: The lint `rule arm-put-operation-response-codes` requires ARM PUT operations to include both `200` and `201`. However, `SourceControlSyncJob_Create` currently only returns `201`. This warning seems to be caused by `Legacy.CreateOrReplaceSync`

PR #44219, opened 2026-06-25
 Bugfix: The intent in deploymentSetting, does let customer share intent to deploy CVM during day 0 deployment.

abaranwal-ms:
Bugfix: The intent in deploymentSetting, does let customer share intent to deploy CVM during day 0 deployment.


rajeshkumarar (approved):

abaranwal-ms:
Bugfix: The intent in deploymentSetting, does let customer share intent to deploy CVM during day 0 deployment.

PR #44235, opened 2026-06-26
Add extension decorators on preview models

kaylieee:
Add extension decorator to models that are only used with preview features

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
@@ -10,7 +10,12 @@ using OpenAPI;
 
 namespace Azure.AI.Projects;
 
+const EvaluatorsRequiredPreviews = #{
+  required_previews: #[FoundryFeaturesOptInKeys.evaluations_v1_preview],
+};
+
 @doc("The category of the evaluator")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
 union EvaluatorCategory {
   string,
 
@@ -25,6 +30,7 @@ union EvaluatorCategory {
 }
 
 @doc("The type of the evaluator")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
 union EvaluatorType {
   string,
 
@@ -36,6 +42,7 @@ union EvaluatorType {
 }
 
 @doc("The type of evaluator definition")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
 union EvaluatorDefinitionType {
   string,
 
@@ -61,6 +68,7 @@ union EvaluatorDefinitionType {
 #suppress "@azure-tools/typespec-azure-core/no-unknown"
 @doc("Base evaluator configuration with discriminator")
 @discriminator("type")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
 model EvaluatorDefinition {
   @doc("The type of evaluator definition")
   type: EvaluatorDefinitionType;
@@ -76,6 +84,7 @@ model EvaluatorDefinition {
 }
 
 @doc("Code-based evaluator definition using python code")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
 model CodeBasedEvaluatorDefinition extends EvaluatorDefinition {
   type: EvaluatorDefinitionType.code;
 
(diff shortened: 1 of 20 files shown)


specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
```diff
@@ -76,6 +84,7 @@ model EvaluatorDefinition {
 }
 
 @doc("Code-based evaluator definition using python code")
+@extension("x-ms-foundry-meta", EvaluatorsRequiredPreviews)
```
jpalvarezl, line 87: This one is a bit tricky. Here it seems like all the fields are annotated as preview fields, **except** `type` and `code_text`. I am not entirely sure we should annotate this class. Let's double check, that the whole feature is beta and then we should probably clean up the `@extension` marks of the properties and leave a single one at the model level.

specification/ai-foundry/data-plane/Foundry/src/insights/models.tsp
```diff
@@ -2,11 +2,17 @@ import "../common/models.tsp";
 
 using TypeSpec.Rest;
 using Azure.Core.Foundations;
+using TypeSpec.OpenAPI;
 
 namespace Azure.AI.Projects;
 
+const InsightsRequiredPreviews = #{
+  required_previews: #[FoundryFeaturesOptInKeys.insights_v1_preview],
```
jpalvarezl: There is no precedent for annotating things like `optional_previews` in the rest of the spec, but AFAICT this is marked at the `route.tsp` level as a `WithConditionalFoundryPreviewHeader<AreaPreviewLabel>`. 

This just might be a TSP level definition and from the OpenAPI and SDK perspective doesn't matter, but I would flag this as a question to @glecaros

specification/ai-foundry/data-plane/Foundry/src/red-teams/models.tsp
```diff
@@ -1,9 +1,14 @@
...
+using TypeSpec.OpenAPI;
 
 namespace Azure.AI.Projects;
 
+const RedTeamsRequiredPreviews = #{
+  required_previews: #[FoundryFeaturesOptInKeys.red_teams_v1_preview],
+};
+
 @doc("Strategies for attacks.")
 union AttackStrategy {
```
jpalvarezl, line 14: Do we need to annotate `AttackStrategy` too?

jpalvarezl (commented):
From what I could tell of what's missing:
- I've left some comments inline
- `src/evaluation-suites` marks all the routes as beta, but nothing in the `models.tsp` . I suspect that everything needs to be annotated there too.
- `src/routines/models.tsp` all the models from `CreateOrUpdateRoutineParameters` (line ~399) aren't annotated and seems to me like they should.
- `src/agents/models.tsp` seems like we need to apply the `AllConditionalAgentDefinitionPreviews` to `model UpdateAgentRequest` too (line ~87) as it's a copy of `CreateAgentVersionRequest` which is marked as such.

kaylieee:
> * `src/evaluation-suites` marks all the routes as beta, but nothing in the `models.tsp` . I suspect that everything needs to be annotated there too.
> * `src/agents/models.tsp` seems like we need to apply the `AllConditionalAgentDefinitionPreviews` to `model 
UpdateAgentRequest` too (line ~87) as it's a copy of `CreateAgentVersionRequest` which is marked as such.

Added annotations

> * `src/routines/models.tsp` all the models from `CreateOrUpdateRoutineParameters` (line ~399) aren't annotated and seems to me like they should.

All the body fields in `CreateOrUpdateRoutineParameters` come from `RoutineCreateOrUpdateRequest`, which is annotated as preview and emitted as the request body schema. `CreateOrUpdateRoutineParameters` itself isn't emitted as a schema.

specification/ai-foundry/data-plane/Foundry/src/red-teams/models.tsp
kaylieee: added

specification/ai-foundry/data-plane/Foundry/src/insights/models.tsp
kaylieee: annotated with conditional

glecaros (approved):

specification/ai-foundry/data-plane/Foundry/src/evaluators/models.tsp
kaylieee: removed property level decorators

PR #44243, opened 2026-06-26
[feature/foundry-release] Add Go SDK emitter configuration for AI Foundry

qmuntal:
This PR wires AI Foundry into the Go SDK generation path.

What changed:
- Adds the Go emitter configuration for the `azaiprojects` package, including fake generation, span injection, and by-value slice elements.
- Adds Go-specific client customizations in `client.go.tsp` so the generated SDK has friendlier names and shapes for memory stores, schedules, evaluator timestamps, credentials, and connection operations.

Validation:
- `npx tsp compile specification/ai-foundry/data-plane/Foundry`

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/sdk-go-azure-ai-projects/client.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/sdk-go-azure-ai-projects/client.tsp
@@ -1,0 +1,91 @@
+import "../common/custom-types.tsp";
+import "../common/service.tsp";
+import "../sdk-common/error.default.tsp";
+import "../connections/routes.tsp";
+import "../datasets/routes.tsp";
+import "../indexes/routes.tsp";
+import "../deployments/routes.tsp";
+import "../red-teams/routes.tsp";
+import "../evaluation-rules/routes.tsp";
+import "../evaluation-taxonomies/routes.tsp";
+import "../evaluators/routes.tsp";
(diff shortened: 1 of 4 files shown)


specification/ai-foundry/data-plane/Foundry/src/sdk-go-azure-ai-projects/client.tsp
```diff
@@ -0,0 +1,91 @@
...
+// --------------------------------------------------------------------------------
+// Root client
+// --------------------------------------------------------------------------------
+@@clientName(Azure.AI.Projects, "Client");
+
+// For Go, keep feature opt-in headers as strings so request and fake-server header handling agree.
+@@alternateType(FoundryFeaturesOptInKeys, string);
+
+// For Go, keep the memory inputs light instead of expanding the full OpenAI input-item model graph.
+@@alternateType(OpenAI.InputItem, Record<unknown>);
```
jhendrixMSFT, line 29: tsp's `unknown` maps to Go's `any` type.  So, for this construct, we'll emit `map[string]any` which has pretty bad ergonomics.  We've received customer complaints about this in the past, so we added emitter switch `rawjson-as-bytes` that when set to `true` will treat `unknown` types as raw JSON, thus emitting `[]byte` instead.

I suggest setting `rawjson-as-bytes: true` in your `tspconfig.yaml` file unless `any` (and other permutations) is desirable for this SDK.

PR #44283, opened 2026-06-29
Photon (TypeSpec)

nickliu-msft:
Based on this swagger change (except for 2026_12_06): https://github.com/Azure/azure-rest-api-specs/pull/41696/changes#diff-2b5024066dbbe95bb6c79a0614f27368bd7921eb01f7821d8bd2cb0da91ddfbd

### DIFF
--- a/specification/storage/data-plane/BlobStorage/Common/models.tsp
+++ b/specification/storage/data-plane/BlobStorage/Common/models.tsp
@@ -2857,3 +2857,10 @@ alias ListBlobsStartFrom = {
   @query
   startFrom?: string;
 };
+
+alias ListBlobsEndBefore = {
+  /** Filters the results to return only blobs whose names are ordered before this value. */
+  @added(Versions.v2026_12_06)
+  @query
+  endBefore?: string;
+};
--- a/specification/storage/data-plane/BlobStorage/Common/models.tsp
+++ b/specification/storage/data-plane/BlobStorage/Common/models.tsp
@@ -2859,7 +2859,7 @@ alias ListBlobsStartFrom = {
 };
 
 alias ListBlobsEndBefore = {
-  /** Filters the results to return only blobs whose names are ordered before this value. */
+  /** Filters the results to return only names that are ordered before this value. Currently only applies to Apache Arrow scenario. */
   @added(Versions.v2026_12_06)
   @query
   endBefore?: string;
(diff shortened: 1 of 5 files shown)


specification/storage/data-plane/BlobStorage/Common/models.tsp
```diff
@@ -2857,3 +2857,10 @@ alias ListBlobsStartFrom = {
   @query
   startFrom?: string;
 };
+
+alias ListBlobsEndBefore = {
+  /** Filters the results to return only blobs whose names are ordered before this value. */
+  @added(Versions.v2026_12_06)
+  @query
+  endBefore?: string;
+};
```
vincenttran-msft, line 2866: I saw in the OneNote you wanted to ensure we documented this is only for arrow?

nickliu-msft: Yes EndBefore currently only applies to arrow. Good callout, will update documentation for this

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -635,6 +635,43 @@ interface Container {
     }
   >;
 
+  /** The List Blobs operation returns a list of the blobs under the specified container. This is the Apache Arrow use case so response is returned as raw to be deserialized by the client. */
```
vincenttran-msft: Docstring is mostly fine, maybe abit verbose- this is not a problem in most tier1 languages now since we expose a handwritten wrapper, but for example in Rust this is directly visible to the customer. Maybe consider the second sentence:
> Returns the response in Apache Arrow format as raw data, to be deserialized by the client.

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -635,6 +635,43 @@ interface Container {
...
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Existing API"
+  @added(Versions.v2026_12_06)
+  @get
+  @route("?restype=container&comp=list&flat&arrow")
+  listBlobFlatSegmentApacheArrow(
+    /** The Accept header indicating the response should be returned as an Apache Arrow stream, falling back to XML. */
```
vincenttran-msft: nit: 
>falling back to XML otherwise.

vincenttran-msft (approved):
Authoring LGTM, just some small nits on docstrings- but fine as-is too 😄

PR #44285, opened 2026-06-29
Merging Foundry feature branch for the 6/30 release.

glecaros:


specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
```diff
@@ -0,0 +1,368 @@
...
+
+  @doc("Raw agent response text.")
+  response?: string;
+
+  @doc("Identifier of the agent run that produced this result.")
+  run_id?: string;
+}
+
+@doc("Terminal-state result body. Populated when status is succeeded or failed.")
+@extension("x-ms-foundry-meta", AgentOptimizationRequiredPreviews)
```
catalinaperalta, line 304: why are you using the extension decorator for openapi? Are you generating from a swagger somewhere?

catalinaperalta: I see several instances of the decorator so wondering what is using that data?

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
```diff
@@ -0,0 +1,65 @@
...
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+using TypeSpec.Versioning;
+
+namespace Azure.AI.Projects;
+
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+#suppress "@azure-tools/typespec-azure-core/long-running-polling-operation-required" "LRO polling is handled by the client via Operation-Location"
```
catalinaperalta: you do have an endpoint that corresponds to the url in the operation-location header dont you? What does the url returned from operation-location look like?

specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
glecaros: this is for the openapi3 description and it's used by our docs pipeline

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
glecaros: it points to the get operation

catalinaperalta: Since you already have the get operation defined in the tsp why not just add the pollingOperation decorator and getting rid of this suppression?

specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
```diff
@@ -52,6 +31,8 @@ options:
     generate-samples: false
     custom-types-subpackage: "implementation.models"
     customization-class: customizations/src/main/java/ProjectsCustomizations.java
+    rename-model:
```
raych1: @glecaros did you mean adding 'rename-model' break the Java SDK generation?

raych1: @weidongxu-microsoft do you have any idea of the Java generation failures? Is it related to this change?

weidongxu-microsoft: One failure is due to duplicate-client-name. Somehow there is 2 `Error` model in Java scope. I guess one from Azure.Core, but not sure where the other one comes from.
```
07:20:04.233 cmderr W 	[Java-Sdk-Automation] /mnt/vss/_work/1/s/azure-sdk-for-java/sdk/ai/azure-ai-projects/TempTypeSpecFiles/node_modules/@azure-tools/typespec-azure-core/lib/foundations.tsp:150:7 - error @azure-tools/typespec-client-generator-core/duplicate-client-name: Client name: "Error" is duplicated in language scope: "java"
07:20:04.233 cmderr W 	[Java-Sdk-Automation] /mnt/vss/_work/1/s/azure-sdk-for-java/sdk/ai/azure-ai-projects/TempTypeSpecFiles/node_modules/@azure-tools/openai-typespec/src/common/models.tsp:235:7 - error @azure-tools/typespec-java/unrecognized-type: Unrecognized type, kind 'nullable'. Updating the version of the emitter may resolve this issue.
```

weidongxu-microsoft: The other failure is a `nullable` TypeSpec type. I need to inspect the openai-typespec lib and check what it is...

```
union ReasoningEffort {
  #suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "Auto-suppressed warnings non-applicable rules during import."
  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Auto-suppressed warnings non-applicable rules during import."
  "none" | "minimal" | "low" | "medium" | "high" | "xhigh",
  null,
}

  reasoning_effort?: ReasoningEffort | null;
```

PR https://github.com/microsoft/typespec/pull/11220

Expect to update emitter tomorrow.

weidongxu-microsoft: `rename-model` likely not a cause of failure, usually -- they are reserved to solve things that `@clientName` cannot handle (e.g. name from instance of template e.g. `Paged<Model>`) -- and only use it in such case

weidongxu-microsoft: One question is why we had a tspconfig for SDK at this folder.

From configure in Java SDK repo, the source is in `src/<sdk-java-*>`
- directory: specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-agents
- directory: specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-projects

None would use the one here. And this one looks like a full import of all ai-foundry related sources, not suitable for an SDK lib.

I didn't say it is not possible to get it pass, just getting it pass here maybe meaningless, as no Java lib generate from here.
May just worth a suppression on SDK in tspconfig.

CC @jpalvarezl

glecaros: hey, yes we don't need this here... this was just added to try to make validation work. Our problem was that if we didn't add the java configuration at this level, TypeSpec validation failed, and if we did SDK validation for java failed (as it tried generating from this config that lacks all of the customizations)

weidongxu-microsoft: Hi @raych1, can we add a suppression on Java (maybe on other languages that don't generate lib from here) on this tspconfig?

As mentioned in https://github.com/Azure/azure-rest-api-specs/pull/44285#discussion_r3548567405, all Java lib is generated from a subfolder, which contains the customization client.tsp, and import the related part of the source of Foundry.

And some of the configure here is not making sense. E.g.
```
    emitter-output-dir: "{output-dir}/{service-dir}/azure-ai-projects"
    customization-class: customizations/src/main/java/ProjectsCustomizations.java
```
uses the Java customization at azure-ai-projects lib. But since this folder is not for this lib, there could be error from such configure that again meaningless to fix.

raych1: @weidongxu-microsoft it makes sense to add suppression to skip Java emitter in `specification/ai-foundry/data-plane/Foundry/tspconfig.yaml`

@glecaros you may follow the wiki at [here](https://github.com/Azure/azure-rest-api-specs/wiki/TypeSpec-Validation#suppress-the-entire-sdktspconfigvalidation-rule).

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
glecaros: added :)

specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
glecaros: Added the suppression and now validation pass :), thanks!

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
```diff
@@ -0,0 +1,65 @@
...
+    DataGenerationJob
+  >;
+
+  /**
+   * Submits a new data generation job for asynchronous execution.
+   */
+  @summary("Create a data generation job")
+  @route("data_generation_jobs")
+  @extension("x-ms-foundry-meta", DataGenerationJobsRequiredPreviews)
+  @Azure.Core.pollingOperation(DataGenerationJobs.create)
```
catalinaperalta: shouldnt this be the get operation on this interface?

specification/ai-foundry/data-plane/Foundry/src/agents-optimization/routes.tsp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Azure.AI.Projects;
+
+// ---------------------------------------------------------------------------
+// Job CRUD operations (shared job templates from servicepatterns.tsp)
+// ---------------------------------------------------------------------------
+
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+#suppress "@azure-tools/typespec-azure-core/long-running-polling-operation-required" "LRO polling is handled by the client via Operation-Location"
```
catalinaperalta: Nit looks like an operation in this interface is also missing the polling operation, might be worth a search throughout the project for this suppression to correct those instances and remove these suppressions

specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
```diff
@@ -6,53 +6,20 @@ emit:
...
-    generate-test: true
-    generate-metadata: false
-    ignore-property-name-normalize: true
-    ignore-enum-member-name-normalize: true
+    flavor: azure
+    package-dir: "ai-projects"
     package-details:
       name: "@azure/ai-projects"
-      description: "Azure AI Projects Client"
+  "@azure-tools/typespec-python":
```
catalinaperalta: Seems like the python config is missing package-name configs, I'm surprised tsv isnt failing on this.

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
```diff
@@ -11,23 +11,17 @@
...
+  paths:
+    - "."
+    - "**"
+
+- tool: TypeSpecValidation
+  reason: >
+     SDKs are configured in nested config files.
+  rules: [SdkTspConfigValidation]
+  paths:
+    - tspconfig.yaml
```
catalinaperalta, line 18: You do have some top level packages so we need some more specificity on which language checks to exclude. For example, this is probably why the python config is missing some options

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
```diff
@@ -11,23 +11,17 @@
...
-#     - options.@azure-tools/typespec-python.generate-sample
-#   reason: 'See above comments for details'
\ No newline at end of file
+- tool: TypeSpecValidation
+  reason: >
+     Foundry does not emit autorest as OpenAPI 2 yields a very incomplete representation of the spec
+  rules: [EmitAutorest]
+  paths:
+    - "."
+    - "**"
```
catalinaperalta, line 11: Is it not enough to just specify the current dir? Why is the second list item needed "**"?

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
catalinaperalta: You should be able to specifically suppress the folder structure rule now

specification/ai-foundry/data-plane/Foundry/src/skills/routes.tsp
```diff
@@ -0,0 +1,181 @@
...
+    SkillVersion
+  >;
+
+  /**
+   * Creates a new version of a skill from uploaded files via multipart form data.
+   */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" ""
+  #suppress "@azure-tools/typespec-azure-core/byos" "Skill file upload uses multipart form data"
+  @summary("Create a skill version from uploaded files")
+  @operationId("createSkillVersionFromFiles")
```
catalinaperalta, line 94: out of curiosity, why cant the generated name be the the auto generated name?

specification/ai-foundry/data-plane/Foundry/src/suppressions.yaml
```diff
@@ -0,0 +1,6 @@
+- tool: TypeSpecValidation
+  reason: >
+     Suppressing validation for the sub configs.
+  paths:
+    - "."
+    - "**"
```
catalinaperalta: I believe you only need to suppress the folder structure rule in the parent suppressions.yaml, so that you can still get other validations on these nested dirs, otherwise you're completely skipping tsv on these nested directories

specification/ai-foundry/cspell.yaml
```diff
@@ -68,4 +69,17 @@ overrides:
...
       - apim-request-id
-      - davinci # legacy OpenAI model name
\ No newline at end of file
+      - davinci # legacy OpenAI model name
+  # Agents optimization terms
+  - filename:
+    - '**/specification/ai-foundry/data-plane/Foundry/src/agents-optimization'
+    - '**/specification/ai-foundry/data-plane/Foundry/openapi3'
+    - '**/specification/ai-foundry/data-plane/Foundry/stable'
+    - '**/specification/ai-foundry/data-plane/Foundry/preview'
```
catalinaperalta, line 78: I think you can probably just try to apply this suppression more generally to the entire Foundry/ dir:
```suggestion
    - '**/specification/ai-foundry/data-plane/Foundry/**'
```

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
catalinaperalta, line 1: Since the suppressions in this file are left very generic, you're most likely not getting a lot of vetting on your project

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/routes.tsp
glecaros: oops, fixed!

specification/ai-foundry/data-plane/Foundry/src/agents-optimization/routes.tsp
glecaros: added to the operations that were missing it...

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
glecaros: python config lies deeper too, the top level config is not used for any client lib generation

specification/ai-foundry/data-plane/Foundry/src/suppressions.yaml
glecaros: I updated to have more cleaner suppressions.

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
glecaros: this is because we have a number of nested configs that need suppressing too... for other rules i made the suppression explicit.

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
glecaros: fixed :)

specification/ai-foundry/data-plane/Foundry/tspconfig.yaml
glecaros: This config shouldn't even be here probably... we define the client library configs in the inner tspconfig files. I do want to minimize changes to those though. Once the people that deal with the client libraries are back, I'll see if we can consolidate some things to make the structure simpler.

specification/ai-foundry/data-plane/Foundry/src/common/error.range.tsp
```diff
@@ -0,0 +1,22 @@
...
+      @maxValue(499)
+      @minValue(400)
+      @statusCode
+      statusCode: int32;
+    }
+  | {
+      ...ErrorResponseParameter;
+
+      @maxValue(599)
+      @minValue(500)
```
catalinaperalta, line 19: We dont really need to specify all the possible error response status codes. Did you get feedback to do this? If not, then you shouldnt really need to do this

specification/ai-foundry/data-plane/Foundry/src/sdk-common/error.default.tsp
```diff
@@ -0,0 +1,2 @@
+/** Error response for Azure range errors. */
+alias AzureRangeErrorResponse<ErrorResponseParameter = Azure.Core.Foundations.ErrorResponse> = ErrorResponseParameter;
```
catalinaperalta, line 2: why do you need this alias? It seems like you can just pass in the azure.core.foundations.errorrepsonse to any template you're using and replace with a different value on the specific operations you need a different value. I dont see how the alias helps?

specification/ai-foundry/data-plane/Foundry/src/suppressions.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - "./sdk-csharp-azure-ai-projects/tspconfig.yaml"
+    - "./sdk-csharp-azure-ai-projects-agents/tspconfig.yaml"
+- tool: TypeSpecValidation
+  reason: >
+     Java Client library config does not include TS/Python config; the inverse is true for the TS/Python client library config, so we need to skip.
+  rules: [SdkTspConfigValidation]
+  paths:
+    - "./sdk-java-azure-ai-agents/tspconfig.yaml"
+    - "./sdk-java-azure-ai-projects/tspconfig.yaml"
+    - "./sdk-python-js-azure-ai-projects/tspconfig.yaml"
```
catalinaperalta, line 35: The other entries in this file are fine, but this one should specifically suppress the rules for the other languages so that you still get Java validation on your dir

specification/ai-foundry/data-plane/Foundry/suppressions.yaml
catalinaperalta: So since you're suppressing the entire SdkTspConfigValidation check you really shouldnt have the typespec-ts or typespec-python, etc entries in your tspconfig.yaml. If someone is actually using those configs then these suppressions should be more targeted towards the languages that you're not interested in

catalinaperalta (commented):
Can the sdk owners review the tspconfig.yaml entries to clean them up? That way we can finish cleaning up the suppression files as well

specification/ai-foundry/data-plane/Foundry/src/common/error.range.tsp
glecaros: this is related to the doc pipeline we are using (where these were required)... the client library generation is actually including the definition from sdk-common

specification/ai-foundry/data-plane/Foundry/src/suppressions.yaml
glecaros: fixed :)

specification/ai-foundry/data-plane/Foundry/src/sdk-common/error.default.tsp
glecaros: this is because we are consuming one or the other version of this alias depending on which entry point (no error ranges in the client libraries, but we have them for docs)

catalinaperalta (approved):

PR #44321, opened 2026-06-30
Add MachineLearningServices stable API version 2026-07-01

fmabroukmsft:

## Purpose of this PR

- [x] New API version for an existing resource provider. (API spec is defined in TypeSpec.)

Adds **Microsoft.MachineLearningServices** stable API version **2026-07-01**.

## What this PR does

- Adds `v2026_07_01: "2026-07-01"` to the TypeSpec `Versions` enum.
- Uses the MLS dual-track model, with stable **2026-05-01** as the source surface.
- Extends all 78 preview-only feature gates through `@removed(Versions.v2026_07_01)`.
- Copies stable source examples, updates request and pagination URL API versions, and removes 41 examples for gated operations.
- Adds `package-2026-07-01` to `readme.md` and makes it the default stable tag.
- Regenerates the stable OpenAPI and examples with TypeSpec 1.14.0.

## Validation

- `main.tsp` and `client.tsp` compile successfully.
- TypeSpec impacted-spec validation passes with a clean generated tree.
- Swagger SemanticValidation, ModelValidation, Avocado, and LintDiff pass.
- TypeSpec and generated JSON formatting checks pass.
- Source examples, generated examples, and unique `x-ms-examples` references match: **243 == 243 == 243**.
- No stale source-version strings remain in the 2026-07-01 examples.
- The stable surface matches 2026-05-01: **628 definitions, 127 paths, 216 operations, and 1,513 properties**, with no additions or removals.
- All preview-only `AssetJobOutput` properties remain excluded from stable.

## Due diligence checklist


specification/machinelearningservices/resource-manager/readme.md
```diff
@@ -27,9 +27,114 @@ These are the global settings for the Machine Learning Services API.
...
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/workspaces/{workspaceName}/computes/{computeName}"].patch.responses
+  - code: PutResponseCodes
+    reason: Pre-existing API behavior.
+  - code: GetCollectionOnlyHasValueAndNextLink
+    reason: Pre-existing API behavior for list operations.
+  - code: XmsPageableForListCalls
+    reason: Pre-existing API behavior, pagination is handled by the framework.
+  - code: PathForResourceAction
+    reason: Pre-existing API behavior for action endpoints.
```
ravimeda: **[EXISTING] 🔴 Blocking** **[[RPC-SUPPRESS-GA](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#ga-vs-preview-rigor-rpc-suppress-ga)]** `specification/machinelearningservices/resource-manager/readme.md` - line 112-113 - `PathForResourceAction` is blanket-suppressed in a GA tag.

**Classification reasoning:** [EXISTING] - the same suppression with identical text is present in the previous stable tag `package-2026-05-01`. It is not introduced by this PR, but this is a GA cut, and GA is the deadline to challenge carried-forward suppressions.

**Issue:** `PathForResourceAction` is suppressed with no `where:` clause, silencing the rule across the entire 36,270-line document, with the reason "Pre-existing API behavior for action endpoints." The suppression-review criteria name this rule explicitly: `PathForResourceAction` suppressions in GA are **blocking**, because they prevent proper RBAC action registration.

**Suggested fix:** Either correct the action paths so the rule passes, or scope the suppression to the specific offending paths with a real justification:

```yaml
  - code: PathForResourceAction
    reason: <per-path technical justification, plus ARM reviewer sign-off>
    where:
      - $.paths["<explicit non-conforming action path 1>"].post
      - $.paths["<explicit non-conforming action path 2>"].post
```

specification/machinelearningservices/resource-manager/readme.md
```diff
@@ -27,9 +27,114 @@ These are the global settings for the Machine Learning Services API.
...
+  - code: CollectionObjectPropertiesNaming
+    reason: Pre-existing API behavior.
+  - code: PatchResponseCodes
+    reason: Pre-existing API behavior.
+  - code: ProvisioningStateMustBeReadOnly
+    reason: Pre-existing API behavior, provisioningState properties are read-only in the API response.
```
ravimeda, line 176: **[EXISTING] 🟠 Warning** **[[RPC-SUPPRESS-GA](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#ga-vs-preview-rigor-rpc-suppress-ga) / [RPC-SUPPRESS-REMOVE-SAFELY](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#recommending-removal-of-an-existing-suppression-rpc-suppress-remove-safely)]** `specification/machinelearningservices/resource-manager/readme.md` - line 118-119 - `ProvisioningStateMustBeReadOnly` suppression reason does not address the emitted shape.

**Classification reasoning:** [EXISTING] - identical emitted shape and identical suppression text in `stable/2026-05-01/openapi.json` and tag `package-2026-05-01`. The two stable documents are identical except `info.version` (line 5).

**Issue:** The suppression reason - "Pre-existing API behavior, provisioningState properties are read-only in the API response" - is *insufficient rather than incorrect*. It describes runtime service behavior, but the rule fires on the emitted artifact, and it does not explain why the emitted shape is exempt.

23 of the 24 `provisioningState` properties in `stable/2026-07-01/openapi.json` emit as (first instance at line 17663):

```json
"provisioningState": {
  "$ref": "#/definitions/DeploymentProvisioningState",
  "description": "Provisioning state for the endpoint deployment.",
  "readOnly": true
}
```

Per JSON Schema, siblings of `$ref` are ignored, so this `readOnly` is invisible to SDK generators, ARM What-If, and Azure Resource Graph regardless of runtime behavior. This is a known emitter limitation, not a service defect. [RPC-SUPPRESS-GA](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#ga-vs-preview-rigor-rpc-suppress-ga) asks that each carried-forward suppression be individually challenged at a GA cut, and here the underlying violation can now be fixed.

**Suggested fix:** Add the emitter option in `MachineLearningServices.Management/tspconfig.yaml` and regenerate:

```yaml
options:
```

**Two caveats before you do.** The option is *project-wide*: it clears every `ProvisioningStateMustBeReadOnly` suppression in the project, not just this one. And regenerating rewrites the emitted `provisioningState` shape for all 9 versions this project emits - `stable/2026-05-01`, `stable/2026-03-01`, `stable/2025-12-01` and the previews as well as `2026-07-01` - which will surface in `Breaking Change(Cross-Version)` and can shift generated SDK type surfaces.

Per [RPC-SUPPRESS-REMOVE-SAFELY](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#recommending-removal-of-an-existing-suppression-rpc-suppress-remove-safely), **removing the suppression is safe only after** you have verified the rule no longer fires: confirm no `provisioningState` in the regenerated `openapi.json` is shaped as `{ "$ref": ..., "readOnly": true }`. If it still fires, keep the suppression but narrow it with a `where:` clause and a reason that addresses the emitted shape - do not delete it.

specification/machinelearningservices/resource-manager/readme.md
```diff
@@ -27,9 +27,114 @@ These are the global settings for the Machine Learning Services API.
...
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/registries/{registryName}/environments/{environmentName}"].delete
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/registries/{registryName}/environments/{environmentName}/versions/{version}"].delete
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/registries/{registryName}/models/{modelName}"].delete
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/registries/{registryName}/models/{modelName}/versions/{version}"].delete
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.MachineLearningServices/workspaces/{workspaceName}/outboundRules/{ruleName}"].delete
+  - code: ResourceNameRestriction
```
ravimeda, line 88: **[EXISTING] 🟠 Warning** **[[RPC-SUPPRESS-SCOPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppression-scoping-rules-rpc-suppress-scope)]** `specification/machinelearningservices/resource-manager/readme.md` - lines 88, 90, 92, 98, 100, 106, 108, 110, 112, 114, 116, 118 - 12 of 20 suppressions in the new GA tag are blanket.

**Classification reasoning:** [EXISTING] - the same 12 blanket suppressions with identical text are present in `package-2026-05-01`. A full payload comparison of the two tag blocks (codes, reasons, and `where:` entries) returns zero differences; the carry-forward itself was done correctly.

**Issue:** These 12 entries name a `code:` with no `where:` clause, so each one silences its rule across the entire 36,270-line document:

`ResourceNameRestriction` (88), `PostResponseCodes` (90), `ProvisioningStateSpecifiedForLROPut` (92), `PatchBodyParametersSchema` (98), `PatchIdentityProperty` (100), `PutResponseCodes` (106), `GetCollectionOnlyHasValueAndNextLink` (108), `XmsPageableForListCalls` (110), `PathForResourceAction` (112), `CollectionObjectPropertiesNaming` (114), `PatchResponseCodes` (116), `ProvisioningStateMustBeReadOnly` (118).

RPC-SUPPRESS-SCOPE requires suppressions to include specific `from`/`where` clauses; blanket suppressions targeting an entire file without a JSON path are not allowed. Several reasons are also non-specific ("Pre-existing API behavior.", "Pre-existing API behavior, pagination is handled by the framework.") and do not explain why the rule does not apply. Because these are blanket, a genuinely **new** violation of any of these 12 rules in a future version would be silently masked - which is the concrete risk, not the historical debt.

**Suggested fix:** Add a `where:` list to each, enumerating the specific JSON paths that violate the rule. The 5 already-scoped suppressions in the same block (lines 94, 102, 120, 129, 133) show the pattern.

One practical warning: narrow only against the paths reported by an **actual failing LintDiff run**. The `where:` comparison is exact-match including the trailing leaf segment, so a hand-built list will fail CI for every path it misses.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, manual decision) against commit [`ad3b78f`](https://github.com/Azure/azure-rest-api-specs/pull/44321/commits/ad3b78f64a6a577667638a804b188bbcdd420e0d). See inline comments for findings 1, 2, 3 and 6 on `readme.md`. Findings 4 and 5 are posted as top-level comments instead: finding 4 targets `stable/2026-07-01/openapi.json`, whose diff GitHub reports as too large to anchor inline comments against, and finding 5 concerns `tspconfig.yaml`, which is outside this PR's diff.

specification/machinelearningservices/resource-manager/readme.md
```diff
@@ -27,9 +27,114 @@ These are the global settings for the Machine Learning Services API.
...
+  - code: ArmResourcePropertiesBag
+    reason: Pre-existing API contract uses type as discriminator property within the properties bag for polymorphic resources.
+    where:
+      - $.definitions.OutboundRuleBasicResource
+  - code: TrackedResourcePatchOperation
+    reason: Pre-existing API uses PUT for updates. PrivateEndpointConnection has location/tags but uses PUT, not PATCH.
+    where:
+      - $.definitions.PrivateEndpointConnection
+```
 ### Tag: package-preview-2026-05-15-preview
```
ravimeda, line 196: **[NEW] 🔵 Suggestion** `specification/machinelearningservices/resource-manager/readme.md` - line 137-138 - missing blank line before the next `### Tag:` heading.

**Classification reasoning:** [NEW] - introduced by this PR. The base branch had 51 `### Tag:` headings, all correctly preceded by a blank line; the head has 52, of which this is the only one that is not.

**Issue:** The new `package-2026-07-01` block closes its fence at line 137 and `### Tag: package-preview-2026-05-15-preview` follows immediately at line 138 with no blank line between them.

This is cosmetic only - it is **not** CI-enforced. The repo `.prettierrc.json` overrides cover `**/*.json` and `**/*.ts` but not markdown, and `Swagger PrettierCheck` passes on this PR. Raising it purely for consistency with the other 51 headings in the file.

**Suggested fix:** Insert a blank line between lines 137 and 138.

ravimeda:
**[EXISTING] 🔵 Suggestion** **[[provisioning-state](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#must-be-read-only)]** `specification/machinelearningservices/resource-manager/Microsoft.MachineLearningServices/stable/2026-07-01/openapi.json` - line 31373-31376 (`$.definitions.RegistryPrivateEndpointConnectionProperties.properties.provisioningState`) - untyped, nullable `provisioningState`.

*Posted as a top-level comment: GitHub reports the `stable/2026-07-01/openapi.json` diff (36,270 added lines) as too large to anchor inline review comments against.*

**Classification reasoning:** [EXISTING] - identical in `stable/2026-05-01/openapi.json`; the two stable documents are identical except `info.version` (line 5).

**Issue:** This `provisioningState` is a bare `"type": "string"` with `"x-nullable": true` - no `enum`, no `readOnly`, no `$ref`:

```json
"provisioningState": {
  "type": "string",
  "description": "One of null, \"Succeeded\", \"Provisioning\", \"Failed\". While not approved, it's null.",
  "x-nullable": true
}
```

The nullability is the part that breaks the ARM contract, and the description concedes it. ARM treats an **absent** `provisioningState` as `Succeeded`, so an explicit `null` is strictly worse than omission - it is a value with no defined meaning in the resource lifecycle.

This is materially different from the 23 `$ref`-sibling cases flagged separately on `readme.md` line 118: there the read-only annotation is merely linter-invisible, whereas here the property is genuinely untyped. Clients get no enum, no terminal-state contract, and no read-only signal.

**Suggested fix:** Model it in TypeSpec as a proper provisioning-state union with `@visibility(Lifecycle.Read)` and drop the nullability. Gate the change with `@added(Versions.v2026_07_01)` so it does not retroactively rewrite the shipped 2026-05-01 and earlier surfaces.

Note that straight reuse of the generic `ProvisioningState` at line 30815 is not a clean fit: its enum (`Unknown`, `Updating`, `Creating`, `Deleting`, `Succeeded`, `Failed`, `Canceled`) omits `Provisioning`, which this resource's own description says the service returns, and that definition is scoped to compute clusters. A dedicated definition alongside the seven existing `<Prefix>ProvisioningState` types is likely the better choice.

ravimeda:
**[NEW] 🔵 Suggestion** `specification/machinelearningservices/MachineLearningServices.Management/tspconfig.yaml` - line 54 - .NET Provisioning emitter is pinned to the previous stable version.

*Posted as a top-level comment because `tspconfig.yaml` is not part of this PR's diff (0 of 504 changed files).*

**Classification reasoning:** [NEW] - the mismatch is created by this PR. On the base branch the pin `2026-05-01` matched the default tag `package-2026-05-01`, so there was no inconsistency; `readme.md` line 30 now moves the default to `package-2026-07-01`.

**Issue:** `"@azure-typespec/http-client-csharp-provisioning"` pins `api-version: "2026-05-01"`, so the .NET Provisioning SDK now generates one stable version behind the service default. This is harmless *today* only because the two documents are identical except `info.version` - it stops being harmless the first time `2026-07-01` diverges, and the divergence will be silent.

Scope note: this concerns line 54 only. The `"@azure-typespec/http-client-csharp-mgmt"` pin at line 50 (`2026-03-15-preview`) did not match the default tag on the base branch either, so it is pre-existing staleness that this PR neither creates nor worsens, and it is out of scope here.

**Suggested fix:** Update line 54 alongside the default-tag change:

```yaml
  "@azure-typespec/http-client-csharp-provisioning":
    namespace: "Azure.Provisioning.MachineLearning"
    emitter-output-dir: "{output-dir}/sdk/machinelearningservices/{namespace}"
    api-version: "2026-07-01"
```

Or note in the PR description why the provisioning emitter intentionally lags the default stable tag.

specification/machinelearningservices/resource-manager/readme.md
fmabroukmsft: ✅ Fixing this. Scoped the suppression to the single offending path instead of the whole rule — `RegistryDataReferences_GetBlobReferenceSAS` on `.../registries/{registryName}/datareferences/{name}/versions/{version}`, an inherited GA route ending in a resource id (changing it would be breaking). The `where:` targets that path at the path level to match the rule's `.post^~` reporting. Commits ad2550fa + a993e35a; Swagger LintDiff is green.

specification/machinelearningservices/resource-manager/readme.md
fmabroukmsft: Non-blocking warning on inherited, already-shipped modeling — the reason text here is the same one accepted for 2026-05-01 and every prior stable. The only real fix (the `use-read-only-status-schema` emitter option) rewrites 9 shipped OpenAPI versions and also mis-marks unrelated `PatchStatus` read-only, which is out of scope for a version-add PR. Keeping as-is; no change in this PR.

specification/machinelearningservices/resource-manager/readme.md
fmabroukmsft: ✅ Fixing this — added the blank line before `### Tag: package-preview-2026-05-15-preview`. Commit ad2550fa.

specification/machinelearningservices/resource-manager/readme.md
fmabroukmsft: ✅ Fixing this. Scoped all 11 non-provisioning blanket suppressions to their exact current violations via `where:` lists (45 paths across ResourceNameRestriction, PostResponseCodes, ProvisioningStateSpecifiedForLROPut, PatchBodyParametersSchema, PatchIdentityProperty, PutResponseCodes, GetCollectionOnlyHasValueAndNextLink, XmsPageableForListCalls, CollectionObjectPropertiesNaming, PatchResponseCodes, plus PathForResourceAction). Paths were derived by running the azure-validator with suppressions removed and capturing every violation per rule; verified 0 residual violations. Commit 59c76cbf. Only ProvisioningStateMustBeReadOnly stays blanket — see that thread.

fmabroukmsft:
Re the `RegistryPrivateEndpointConnectionProperties.provisioningState` suggestion: pre-existing on a shipped contract, not introduced here. The proposed `@added` fix is incorrect — it would strip the property from every earlier API projection — so it isn't something to fold into a version-add. This is a modeling cleanup for the service owners to schedule separately. Out of scope; no change in this PR.

fmabroukmsft:
Re the .NET provisioning emitter pin (`tspconfig.yaml` line 54): not a defect. The `http-client-csharp-provisioning` api-version is an independent SDK release choice that routinely lags the latest stable, and it isn't part of this PR's diff. Retargeting it here would be an unrelated, owner-driven change. Leaving as-is; no change in this PR.

ravimeda:
**[EXISTING] 🔴 Blocking** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md#proactive-secret-detection-sec-secret-detect)]** `specification/machinelearningservices/resource-manager/Microsoft.MachineLearningServices/stable/2026-07-01/openapi.json` - lines 21183, 22137, 35447, 35712, 35721, 35730, 35794, 35919 - eight credential-bearing properties reachable from GET, LIST, PUT and PATCH responses carry no `x-ms-secret` annotation.

*Posted as a top-level comment: GitHub returns no `patch` for the 36,270-line added `openapi.json`, and every cited `models.tsp` line falls outside this PR's diff hunks (the last hunk ends at line 15431), so neither file can be anchored inline.*

**Classification reasoning:** [EXISTING] - identical in `stable/2026-05-01/openapi.json`; the two stable documents differ on exactly one line (`info.version`, line 5). Both contain exactly 34 `x-ms-secret` occurrences.

**Issue:** The `WorkspaceConnectionPropertiesV2` discriminated union annotates seven of its credential properties as secrets and skips six others that are reached by the same four operations.

Annotated - `WorkspaceConnectionOAuth2.clientSecret`, lines 35759-35763:

```json
"clientSecret": {
  "type": "string",
  "format": "password",
  "x-ms-secret": true
},
```

The same treatment is applied to `WorkspaceConnectionOAuth2.developerToken` (35764-35769), `.password` (35770-35774), `.refreshToken` (35775-35780), `WorkspaceConnectionServicePrincipal.clientSecret` (35906-35910), and `WorkspaceConnectionUsernamePassword.password` (35937-35941) and `.securityToken` (35942-35947).

Not annotated - `WorkspaceConnectionApiKey.key`, lines 35730-35732, reached by the same `WorkspaceConnections_Get` / `_List` / `_Create` / `_Update`:

```json
"key": {
  "type": "string"
}
```

This directly contradicts the suppression reason used in `models.tsp` - "Secret property is part of existing API contract" - because the same contract, in the same union, annotates seven siblings.

Full set of unannotated properties:

| Emitted property (openapi.json line) | TypeSpec source (models.tsp line) | Reachable from |
|---|---|---|
| `CustomKeys.keys` (21183) - open `Record<string>` of credential values | 16481 - `arm-no-record` suppression only | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |
| `VirtualMachineSshCredentials.password` (35447) | 11266-11267 - `#suppress ".../secret-prop"` | `Compute_Get`, `_List`, `_CreateOrUpdate`, `_Update` |
| `DatabricksProperties.databricksAccessToken` (22137) | 11338-11339 - `#suppress ".../secret-prop"` | `Compute_Get`, `_List`, `_CreateOrUpdate`, `_Update` |
| `WorkspaceConnectionAccessKey.secretAccessKey` (35712) | 16288-16289 - `#suppress ".../secret-prop"` | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |
| `WorkspaceConnectionAccountKey.key` (35721) | 16313-16314 - `#suppress ".../secret-prop"` | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |
| `WorkspaceConnectionApiKey.key` (35730) | 16355-16356 - `#suppress ".../secret-prop"` | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |
| `WorkspaceConnectionPersonalAccessToken.pat` (35794) | 16692 - no `@secret`, no `secret-prop` suppression | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |
| `WorkspaceConnectionSharedAccessSignature.sas` (35919) | 16746 - no `@secret`, no `secret-prop` suppression | `WorkspaceConnections_Get`, `_List`, `_Create`, `_Update` |

`CustomKeys.keys` is the worst instance: an unbounded dictionary on a definition the document itself describes as "Custom Keys credential object" (line 21181), so every value is a credential by construction. It already carries an `AvoidAdditionalProperties` suppression at `readme.md` line 44, but has never been marked a secret.

Two further same-document asymmetries corroborate that this is an omission rather than a decision:

- In `VirtualMachineSshCredentials` (35439-35464), `publicKeyData` (35451-35456) and `privateKeyData` (35457-35462) both carry `"format": "password"` + `"x-ms-secret": true`, while `password` (35447-35450) carries neither. The model marks the *public* key as a secret and leaves the admin password unmarked.
- The document's `format: password` and `x-ms-secret: true` occurrences each number exactly 34 and pair 1:1 with zero exceptions in either direction. All eight flagged properties fall outside that convention.

All four `*DatastoreCredentials` models that have a `secrets` property annotate it at the container level - `AccountKeyDatastoreCredentials.secrets`, lines 16268-16276:

```json
"secrets": {
  "$ref": "#/definitions/AccountKeyDatastoreSecrets",
  "format": "password",
  "description": "[Required] Storage account secrets.",
  "x-ms-mutability": [
    "update",
    "create"
  ],
  "x-ms-secret": true
}
```

Thirty-one lines earlier, `AccessKeyAuthTypeWorkspaceConnectionProperties.credentials` (16237-16239) is a bare `$ref` with neither annotation. None of the ten `$ref` sites feeding the eight tabled leaves carries `x-ms-secret` or `x-ms-mutability` (one of the ten, `VirtualMachineSecrets.administratorAccount` at 35333, is on the POST-only `Compute_ListKeys` path and is listed for completeness rather than as evidence).

Detection signals per the rule:

- **Signal 1 (property-name match, case-insensitive):** `password`, `key`, `keys` and `sas` match the rule's name list; `secretAccessKey` contains `secret`, `accessKey` and `key`; `databricksAccessToken` contains `token`.
- **Signal 4 (suppressed lint rules):** five of the eight sit under `#suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "Secret property is part of existing API contract"`. The reference names this exact shape as its "Bad" example and states that suppressing a secret lint rule "is a **strong signal** that the property IS a secret and the annotation was skipped."
- `WorkspaceConnectionPersonalAccessToken.pat` matches no name in the Signal 1 list and carries no description, so no enumerated signal fires on the property in isolation. It is included under the rule's obligation to "proactively inspect every string property ... even when the property is not explicitly annotated as a secret": the enclosing model is named `WorkspaceConnectionPersonalAccessToken`, it is selected by discriminator value `PAT`, and `pat` is its only property.

`models.tsp` contains seven `secret-prop` suppressions (lines 9250, 11266, 11338, 11539, 16288, 16313, 16355). Five are cited above. The two not cited: line 9250 (`EndpointModelSkuRateLimitRuleProperties.key`) - that model is **not emitted into `stable/2026-07-01`** at all, so it is out of scope for this version; and line 11539 (`DatabricksComputeSecretsProperties.databricksAccessToken`) - returned only by `Compute_ListKeys` (POST), the correct RBAC-gated placement, so it is not a GET/PUT/PATCH exposure. Signal 4 still fires on the latter and the missing annotation still affects SDK and docs redaction, so it is worth fixing in the same pass even though it is not part of the blocking claim.

Reachability chain:

- `Compute_Get` / `Compute_List` (GET), `Compute_CreateOrUpdate` (PUT) and `Compute_Update` (PATCH) return `ComputeResource` / `PaginatedComputeResourcesList`. `ComputeResource.properties.properties` -> `Compute` (line 19971, discriminator `computeType`). `VirtualMachineSchemaProperties.administratorAccount` (35319-35321) and `HDInsightProperties.administratorAccount` (24923-24926) -> `VirtualMachineSshCredentials`; `Databricks.properties` (`$ref` at 22107) -> `DatabricksProperties`.
- `WorkspaceConnections_Get` / `_List` (GET), `_Create` (PUT) and `_Update` (PATCH) return `WorkspaceConnectionPropertiesV2BasicResource`, whose `properties` is `WorkspaceConnectionPropertiesV2` (discriminator `authType`). Six of its twelve subtypes reach the leaf models above: `AccessKey` (`credentials` at 16237), `AccountKey` (16252), `ApiKey` (16694), `CustomKeys` (21195), `PAT` (30026) and `SAS` (32034).

The rule states secrets "**MUST NOT** be exposed in GET, PUT, or PATCH responses" and directs that retrievable secrets be exposed via a POST action for granular RBAC. That half is already right here: every operation *intended* to return secrets - `Compute_ListKeys` and `WorkspaceConnections_ListSecrets`, both POST - is correctly placed. Only the annotation on the GET/PUT/PATCH-reachable copies is missing.

**Suggested fix:** In `models.tsp`, replace each `#suppress "@azure-tools/typespec-azure-resource-manager/secret-prop"` with the `@secret` decorator, and add `@secret` to `keys`, `pat` and `sas`:

```tsp
model WorkspaceConnectionApiKey {
  /** Api key for the connection. */
  @secret
  key?: string;
}
```

Because the emitter output is shared across every version this project emits, the realistic options are: (a) apply `@secret` now and accept regeneration across the shipped versions; (b) gate the annotation with `@added(Versions.v2026_07_01)` so earlier projections are untouched; or (c) if the service demonstrably never populates these properties on GET/PUT/PATCH, document that and keep the shape - but then the `secret-prop` suppression reasons should say so rather than "Secret property is part of existing API contract", which is an assertion that the property *is* a secret.

ravimeda:
**[NEW] 🔵 Suggestion** `specification/machinelearningservices/resource-manager/Microsoft.MachineLearningServices/stable/2026-07-01/openapi.json` - line 5 - new stable GA version has zero API-surface delta over the previous stable.

*Posted as a top-level comment: GitHub returns no `patch` for the 36,270-line added `openapi.json`, so it cannot be anchored inline.*

**Classification reasoning:** [NEW] - the condition is introduced by this PR; `stable/2026-07-01/openapi.json` does not exist at the base commit.

**No named rule prohibits this** - I am raising it as a design question, not a violation, and want to be explicit about that. The ARM review guidance governs version *progression* and the default-tag rule, both of which this PR satisfies; neither it nor `documentation/uniform-versioning.md` requires a new version to differ from its predecessor.

**Issue:** `stable/2026-07-01/openapi.json` is byte-identical to `stable/2026-05-01/openapi.json` except `info.version` on line 5. Same 127 paths, same 628 definitions, same 243 examples - a line-by-line comparison of the two documents returns exactly one differing line.

The cost of the version is real: SDK regeneration and release across six languages, docs and support-matrix churn, and one more version that every future change has to be applied to. The customer-visible benefit over `2026-05-01` is nil, since the two surfaces are the same. The PR description explains the dual-track model, which explains *how* `2026-07-01` was produced, but not why a customer on `2026-05-01` would move to it.

**Suggested fix:** Confirm with the service team that a `2026-07-01` GA is required by a release commitment (a partner dependency, a compliance date, a documented deprecation schedule) rather than produced automatically by the dual-track cadence. If it is cadence-driven, consider skipping this cut and letting `2026-05-01` remain the current stable until a version with an actual delta is ready.

For the record, everything else about the version-add itself checks out: all 78 preview-only feature gates are correctly extended with `@removed(Versions.v2026_07_01)` with zero unclosed gates, the default tag correctly moves stable-to-stable, example integrity is exact at 243 source / 243 generated / 243 unique `x-ms-examples` references with no stale api-version literals, and the PR additionally fixes 44 stale `nextLink` api-versions that still read `2025-07-01-preview` in `2026-05-01`.

specification/machinelearningservices/resource-manager/readme.md
ravimeda: Re-verified at head `a993e35ae6073a26071efa9fbf8bdcf382021ba9`: 11 of the 12 blanket suppressions flagged here are now scoped. Suppressions carrying a `where:` clause went from 8 of 20 in `package-2026-05-01` to 19 of 20 in `package-2026-07-01`, and Swagger LintDiff passes, so the enumerated paths are exact matches.

The one residual is `ProvisioningStateMustBeReadOnly` (readme.md lines 175-176), which is still blanket. That item is already dispositioned on https://github.com/Azure/azure-rest-api-specs/pull/44321#discussion_r3712497347, where the rationale for deferring the `use-read-only-status-schema` emitter option was accepted and the thread resolved - so it is not a fresh request here.

Leaving this thread open rather than resolving it, so the residual is visible for explicit ARM sign-off. If you are satisfied with the deferral recorded on the other thread, resolve this one.

fmabroukmsft:
**Re: SEC-SECRET-DETECT (top-level, head `a993e35`)** — addressed in `ce274e54b0`.

The eight flagged GET/PUT/PATCH-reachable credential properties now carry `@secret`, emitting `x-ms-secret: true` + `format: password` in `stable/2026-07-01/openapi.json`:

- `CustomKeys.keys`
- `VirtualMachineSshCredentials.password`
- `DatabricksProperties.databricksAccessToken`
- `WorkspaceConnectionAccessKey.secretAccessKey`
- `WorkspaceConnectionAccountKey.key`
- `WorkspaceConnectionApiKey.key`
- `WorkspaceConnectionPersonalAccessToken.pat`
- `WorkspaceConnectionSharedAccessSignature.sas`

The five `#suppress ".../secret-prop"` sites among these were replaced with the `@secret` decorator, so the annotation now matches the contract rather than asserting it via a suppression reason. `DatabricksComputeSecretsProperties.databricksAccessToken` (models.tsp line 11539, returned only by the POST `Compute_ListKeys`) was annotated in the same pass for consistency, even though it is not a GET/PUT/PATCH exposure.

`EndpointModelSkuRateLimitRuleProperties.key` (line 9250) is not emitted into `stable/2026-07-01`, so it is out of scope for this version.

Net effect on the emitted document: `x-ms-secret` occurrences move from 34 to 43 (+9), and `format: password` pairs 1:1 with `x-ms-secret` with no exceptions. Swagger LintDiff/ModelValidation/SemanticValidation remain green.

fmabroukmsft:
**Re: zero API-surface delta over `2026-05-01` (top-level suggestion)** — expected and intentional.

MachineLearningServices stable GA cuts are kept in lockstep with the CognitiveServices stable release cadence, which ships on a fixed schedule. `2026-07-01` is the matching MLS stable for that cycle regardless of whether MLS itself carries a surface change in the same window. This cut is cadence-driven by design, not an automatic or accidental version bump — keeping the two RP version lines aligned avoids a later out-of-band MLS cut when the next delta lands and simplifies cross-RP version reasoning for downstream consumers.

The version-add mechanics validate clean as noted: all preview-only feature gates are extended with `@removed(Versions.v2026_07_01)` with no unclosed gates, the default tag moves stable-to-stable, example integrity is exact (243 source / 243 generated / 243 referenced), and the pass additionally corrects the 44 stale `nextLink` api-versions inherited from `2026-05-01`.

gary-x-li:
Please fix the TypeSpec Validation error.

rkuthala (approved):

PR #44326, opened 2026-06-30
batch: do not encode containerName + blobName

jorgerangel-msft:
Updates the tsp customization for maintaining the previous [swagger](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs.Batch/src/autorest.md#dont-encode-blobname-or-container-name) customization.


catalinaperalta (approved):
lgtm, but might be worth a check in with the other langs if they need these updates too and just update in the main routes

catalinaperalta (approved):

PR #44334, opened 2026-07-01
Add CognitiveServices preview API version 2026-07-15-preview

fmabroukmsft:

## Purpose of this PR

  - [x] New API version for an existing resource provider. (API spec is defined in TypeSpec.)

Adds **Microsoft.CognitiveServices** preview API version **2026-07-15-preview**.

> **⚠️ Stacked PR.** Based on `swagger-agent/CogSvc-2026-07-01-typespec` (PR #44323), which is not yet merged. Once #44323 merges, this PR will be reconciled with `main` (additive) and retargeted to `main`.

## What this PR does

- Adds `v2026_07_15_preview: "2026-07-15-preview"` to the TypeSpec `Versions` enum.
- Single-track preview generated from the previous **stable** `2026-07-01`.
- **Step 3a re-promotion:** added `@added(v2026_07_15_preview)` after all 72 `@removed(v2026_07_01)` decorators, re-enabling every feature that was gated out of the `2026-07-01` GA stable — ManagedCompute/Compute/Workbench/QuotaScopeType/FoundryAutoUpgrade (mechanical gating) plus Speculative Decoding and RAI Egress Policy (GA exclusions).
- Copied the complete example set from `2026-05-15-preview` and updated `api-version` strings.
- Added `package-2026-07-15-preview` tag section to `readme.md` (date-descending order); default tag unchanged (`package-2026-07-01`, latest stable).

## Validation

- `tsp compile` (main.tsp + client.tsp) succeeds with no errors.
- Example count == `x-ms-examples` ref count (192 == 192).
- Surface is **identical** to `2026-05-15-preview` (no definitions/paths/operations dropped — Step 6b passes).

## Due diligence checklist


specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
```diff
@@ -0,0 +1,475 @@
...
+using TypeSpec.Versioning;
+
+namespace Microsoft.CognitiveServices;
+
+/**
+ * Cognitive Services account Arc deployment, backed by customer-managed Arc-enabled Kubernetes resources.
+ */
+@added(Versions.v2026_07_15_preview)
+@parentResource(Account)
+model ArcDeployment
```
ravimeda: **[NEW] 🔴 Blocking** **[CI: TypeSpec Validation — `Format` rule]** `specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp` — line 22-23

`TypeSpec Validation` currently fails for `specification/cognitiveservices/CognitiveServices.Management`, and the failing rule is `Format`. This is CI's own output, not a local reproduction — GitHub Actions run `31035445221`, job `92406256487`, step "Validate Impacted Specs". After running `tsp format`, CI's `git diff` names this file as the sole modified file and shows the expected reformat:

```diff
@@ -19,8 +19,10 @@ namespace Microsoft.CognitiveServices;
  */
 @added(Versions.v2026_07_15_preview)
 @parentResource(Account)
-model ArcDeployment
-  is Azure.ResourceManager.ProxyResource<ArcDeploymentProperties, false> {
+model ArcDeployment is Azure.ResourceManager.ProxyResource<
+  ArcDeploymentProperties,
+  false
+> {
   ...ResourceNameParameter<
```

followed by *"Files have been changed after `tsp format`. Run `tsp format` and ensure all files are included in your change."*

Note that the Format rule diffs the whole `specification/cognitiveservices` folder, so the same failure is also reported against sibling projects (`AnomalyDetector`, `ContentSafety`). A single `tsp format` on this file clears all of them.

**Classification reasoning:** Introduced in this PR — `ArcDeployment.tsp` is a new file.

**Suggested fix:** run `npx tsv specification/cognitiveservices/CognitiveServices.Management` (or `npx tsp format`) and commit the result.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -50,6 +50,10 @@ union ProvisioningState {
   Canceled: "Canceled",
   #suppress "@azure-tools/typespec-azure-core/documentation-required" "CognitiveServices-specific: DNS records are being resolved during account creation"
   ResolvingDNS: "ResolvingDNS",
+
+  #suppress "@azure-tools/typespec-azure-core/documentation-required" "Resource is unreachable"
+  @added(Versions.v2026_07_15_preview)
+  ExtensionUnreachable: "ExtensionUnreachable",
```
ravimeda, line 56: **[NEW] 🟠 Warning** **[[RPC-Async-V1-02](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#65-provisioningstate-requirements-rpc-async-v1-02-rpc-async-v1-03-rpc-async-v1-16) / [provisioning-state](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#what-it-represents)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` — line 54-56 — health state added to the shared `ProvisioningState`

Emitted at `preview/2026-07-15-preview/cognitiveservices.json` line 18715, JSON path `$.definitions.ProvisioningState` (definition begins at line 18703).

`provisioningState` represents the status of the latest long-running PUT, PATCH, or DELETE — and explicitly *not* the overall health of the resource or its functional/runtime state. Arc extension reachability is health: the extension can become unreachable with no ARM operation in flight, which would move `provisioningState` outside an LRO and force a transition from one terminal state to another without an intervening operation (RPC-Async-V1-16).

The blast radius is wider than `ArcDeployment`. This enum is shared, and `$ref`d by exactly four models — `AccountProperties` (line 11905), `ArcDeploymentProperties` (line 13032), `ManagedComputeDeploymentProperties` (line 16991), and `ProjectProperties` (line 18605). Accounts, Projects, and managed compute deployments have no Arc extension, yet all three now advertise `ExtensionUnreachable` to clients and SDK consumers.

Separately, the new member ships with no description in the emitted swagger — the `documentation-required` suppression on line 54 supplies its text to the linter only, so customers and generated SDK docs see nothing.

**Classification reasoning:** Introduced in this PR — `ExtensionUnreachable` is absent from `preview/2026-05-15-preview` and from stable `2026-07-01`.

**Suggested fix:** if this is a health signal, move it to a dedicated read-only property on `ArcDeploymentProperties` (for example `extensionStatus`) and leave the shared `ProvisioningState` alone. If instead it is the terminal outcome of a failed provisioning attempt, use `Failed` and convey the cause through the existing `provisioningDetails.message`. Either way, replace the suppression with a real doc comment.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
```diff
@@ -0,0 +1,475 @@
...
+   * Read-only. Status message and timestamp from the last provisioning operation.
+   */
+  @visibility(Lifecycle.Read)
+  provisioningDetails?: ArcDeploymentProvisioningDetails;
+
+  /**
+   * Read-only. Base URL for inference calls to this deployment on the Arc cluster. Populated when provisioningState is Succeeded.
+   */
+  @visibility(Lifecycle.Read)
+  inferenceEndpoint?: string;
```
ravimeda, line 196: **[NEW] 🔵 Suggestion** **[[Section 8.4](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#84-use-specific-types-instead-of-generic-strings-actively-review) / [TypeSpec Section 5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp` — line 199 — URL property typed as plain `string`

Emitted at `preview/2026-07-15-preview/cognitiveservices.json` line 13041, JSON path `$.definitions.ArcDeploymentProperties.properties.inferenceEndpoint`.

The doc comment describes this as *"Base URL for inference calls to this deployment on the Arc cluster"*, but the plain `string` type means the emitted schema carries no `"format": "uri"`. Using the `url` scalar gives SDK consumers a typed URI and lets tooling validate the value.

**Classification reasoning:** Introduced in this PR — `ArcDeploymentProperties` is new in this API version. Recording this as a suggestion rather than a warning because the pre-existing `AccountProperties.endpoint` (swagger line 11909) has the identical plain-`string` shape, so changing only the new property would create local inconsistency.

**Suggested fix:**

```tsp
inferenceEndpoint?: url;
```

Worth considering the same change for `AccountProperties.endpoint` in a follow-up so the RP converges on one shape.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iterations, converged) against commit [`b861bd7`](https://github.com/Azure/azure-rest-api-specs/pull/44334/commits/b861bd73daf210435c355ac1b067a5bac9da97a2). See inline comments for findings 3, 6, and 10. The other twelve findings all concern `preview/2026-07-15-preview/cognitiveservices.json`, whose diff GitHub declines to render (`diff is too large` — 21,393 added lines), so they cannot be anchored to a line and are posted in this review body instead.

**Summary:** 3 blocking, 6 warnings, 1 existing warning, 5 suggestions. Two of the three blocking items are already failing CI (`Swagger LintDiff`, `TypeSpec Validation`). No breaking changes were found.

---

### 1. **[NEW] 🔴 Blocking** **[[RPC-Put-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-07-15-preview/cognitiveservices.json` — line 2164-2188 — PUT is create-only

JSON path: `$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/arcDeployments/{deploymentName}"].put.responses`

The rule states: *"PUT **MUST** return `201` for new resource creation and `200` for replacement of an existing resource (RPC-Put-V1-11). These status codes apply to both synchronous and asynchronous PUT operations."*

`ArcDeployments_Create` declares only `201` and `default`. It is a genuine LRO (`x-ms-long-running-operation: true` at line 2201, with `Azure-AsyncOperation` and `Location` headers at 2170-2180), so the rule applies directly. The TypeSpec source states the intent outright — `ArcDeployment.tsp` line 61 suppresses `arm-put-operation-response-codes` with the reason *"Arc deployment PUT is create-only; updates are supported through PATCH."* A create-only PUT is not idempotent and breaks ARM template redeployment, Bicep, and Terraform round-tripping.

Across all 30 PUT operations in this file, this is the **only** one lacking `200` — including the closest sibling, `ManagedComputeDeployments_CreateOrUpdate`. Corroboration: LintDiff raises `PutResponseCodes` and the downstream `ProvisioningStateSpecifiedForLROPut` here, both unsuppressed, so `Swagger LintDiff` currently fails.

**Classification reasoning:** Introduced in this PR — the `arcDeployments` paths do not exist in `2026-05-15-preview` or in stable `2026-07-01`.

**Suggested fix — please pick one explicitly:**

**(a) Preferred** — make PUT create-or-update by dropping the custom `Response =` override and the suppression:

```tsp
create is ArmResourceCreateOrReplaceAsync<
  ArcDeployment,
  LroHeaders = ArcDeploymentCreateLroHeaders
>;
```

**(b)** If the service genuinely cannot support update-through-PUT, this needs ARM API board sign-off **plus** a `PutResponseCodes` suppression in the new readme tag block. A TypeSpec `#suppress` does not silence the swagger-level LintDiff rule.



---

### 2. **[NEW] 🔴 Blocking** **[[Section 8.5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#85-additionalproperties-restriction)]** `specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-07-15-preview/cognitiveservices.json` — lines 13016, 13046, 13167 — unsuppressed `additionalProperties`

JSON paths: `$.definitions.ArcDeploymentProperties.properties.nodeSelector`, `$.definitions.ArcDeploymentProperties.properties.capabilities`, `$.definitions.ArcDeploymentUpdateProperties.properties.nodeSelector`

LintDiff raises `AvoidAdditionalProperties` at **error** level on all three, unsuppressed, which is one of the two reasons `Swagger LintDiff` fails.

This is a paperwork gap rather than a design disagreement: you already suppressed `arm-no-record` for exactly these three sites at `ArcDeployment.tsp` lines 170, 204, and 241, so the decision is made — it just never reached the readme. TypeSpec-level suppressions do not silence swagger-level LintDiff rules. A case-insensitive scan of all 1,371 lines of `resource-manager/readme.md` returns zero hits for `ArcDeployment`.

**Classification reasoning:** Introduced in this PR — all three definitions are new in this API version.

**Suggested fix:** add the following to the `package-2026-07-15-preview` tag block, mirroring the existing entries you already carry for `ProjectProperties.endpoints` and `ManagedComputeDeploymentProperties.capabilities`:

```yaml
- code: AvoidAdditionalProperties
  reason: nodeSelector is a user-defined Kubernetes node label map with customer-chosen keys. capabilities is a read-only string map of model/runtime capability flags with dynamic keys, matching the existing ManagedComputeDeploymentProperties.capabilities pattern.
  where:
    - $.definitions.ArcDeploymentProperties.properties.nodeSelector
    - $.definitions.ArcDeploymentProperties.properties.capabilities
    - $.definitions.ArcDeploymentUpdateProperties.properties.nodeSelector
```


---

### 4. **[NEW] 🟠 Warning** **[[RPC-Patch-V1-09](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#43b-patch-should-support-updating-envelope-properties-rpc-patch-v1-09-rpc-patch-v1-11)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 12805-12808 — immutable `sku` is not annotated

JSON path: `$.definitions.ArcDeployment.properties.sku`

The rule states the remedy exactly: *"If SKU is immutable, it must be annotated with `x-ms-mutability: ["create", "read"]` (RPC-Patch-V1-09)."*

`sku` is **required** on the resource, is absent from the PATCH body (`ArcDeploymentUpdate`, line 13142), and carries no mutability annotation — yet you declare it immutable at `ArcDeployment.tsp` line 212 (`#suppress "patch-envelope"` — *"Arc deployment SKU is immutable after create"*). Without the annotation, ARM Template What-If, Terraform drift detection, and generated SDKs all treat the SKU as freely mutable. LintDiff warns `PatchSkuProperty` at line 2237-2243.

**Classification reasoning:** Introduced in this PR — `ArcDeployment` is new in this API version.

**Suggested fix:**

```tsp
sku: ArcDeploymentSku;
```



---

### 5. **[NEW] 🟠 Warning** **[[Section 2.6.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#261-standard-sku-schema) / [XmsEnumValidation](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/enum-best-practices.md#openapi-json)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 13123-13141 — bespoke SKU with an unnamed, closed, single-valued enum

JSON path: `$.definitions.ArcDeploymentSku`

Three separate problems in one definition:

1. **Bespoke SKU.** Section 2.6.1 defines the standard ARM SKU shape (`name` required, plus `tier`, `size`, `family`, `capacity`), and this RP already has that model — the sibling `ManagedComputeDeployment.sku` (line 16855-16858) references it via `$ref: "#/definitions/Sku"` and keeps it optional. `ArcDeploymentSku` drops every field except `name`.
2. **Closed, single-valued, and required.** `name` is simultaneously required, `"modelAsString": false`, and constrained to `["Arc"]`, with no stated rationale for a permanently fixed value set. Introducing a second Arc SKU later would therefore be a breaking change on a required field.
3. **Missing enum name.** The `x-ms-enum` block at line 13133-13135 omits the required `name` property. A brace-aware scan confirms this is the only `x-ms-enum` in the file without one — the other 82 all have it — so SDKs will generate an anonymous inline enum here.

**Classification reasoning:** Introduced in this PR — `ArcDeploymentSku` does not exist in `2026-05-15-preview`.

**Suggested fix:** prefer the RP's existing standard `Sku`. If a distinct Arc envelope really is required, at minimum name the enum and make it extensible:

```tsp
union ArcDeploymentSkuName {
  string,

  /** Arc deployment SKU. */
  Arc: "Arc",
}

model ArcDeploymentSku {
  /** The name of the Arc deployment SKU. */
  name: ArcDeploymentSkuName;
}
```



---

### 7. **[NEW] 🟠 Warning** **[[OAPI034](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034) / [Section 8.20](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#820-fields-must-have-clear-ownership--server-or-client-oapi034)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 13023-13030 — `deploymentState` and `raiPolicyName` are unreachable after create

JSON paths: `$.definitions.ArcDeploymentProperties.properties.deploymentState`, `$.definitions.ArcDeploymentProperties.properties.raiPolicyName`

Both properties are writable at create but there is no way to change either afterwards. PUT is create-only (see finding 1); `ArcDeploymentUpdateProperties` (line 13152) exposes only `replicas`, `resources`, and `nodeSelector`; and `arcDeployments` has no lifecycle actions — unlike `deployments`, which offers `/pause` and `/resume`.

That directly contradicts the shipped documentation. The `DeploymentState` description at line 16083 reads: *"The state of the deployment. Controls whether the deployment is accepting inference requests. Use 'Running' for active deployments that process requests, or 'Paused' to temporarily stop inference while preserving the deployment configuration."* The API as specified cannot pause or resume an Arc deployment. Neither property carries `x-ms-mutability`, so tooling treats both as freely mutable.

**Classification reasoning:** Introduced in this PR — `ArcDeploymentProperties` is new in this API version.

**Suggested fix:** add both to `ArcDeploymentUpdateProperties` so PATCH can change them, which matches the documented intent. If they genuinely are immutable, annotate them `@visibility(Lifecycle.Read, Lifecycle.Create)` instead and correct the description for this resource.



---

### 8. **[NEW] 🟠 Warning** **[[OAPI022 / WHATIF-002](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#814-default-values-must-be-static-and-declared-oapi022-whatif-002)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 13023-13030 — undeclared server-applied defaults

JSON paths: `$.definitions.ArcDeploymentProperties.properties.deploymentState`, `$.definitions.ArcDeploymentProperties.properties.raiPolicyName`

This PR's own example demonstrates the problem. In `preview/2026-07-15-preview/examples/CreateArcDeployment.json`, the request `resource.properties` contains only `model`, `extensionId`, `runtime`, `compute`, `replicas`, `resources`, and `nodeSelector` — yet the `201` response body returns `"deploymentState": "Running"` and `"raiPolicyName": "Microsoft.DefaultV2"`.

Every optional property for which the service applies a default when it is omitted from PUT must declare that default in the spec. Otherwise ARM Template What-If reports a false delete (the template omits the property, GET returns the service-applied value) and Terraform reports permanent drift.

**Classification reasoning:** Introduced in this PR — both properties are new in this API version.

**Suggested fix:**

```tsp
deploymentState?: DeploymentState = DeploymentState.Running;
raiPolicyName?: string = "Microsoft.DefaultV2";
```

If the RAI policy default is region- or account-dependent rather than a static constant, it must not be defaulted server-side at all — make it required, or mark it read-only and set it through an explicit operation.



---

### 9. **[NEW] 🟠 Warning** **[[OAPI034](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034) / [Section 8.6](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#86-visibility-and-mutability)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — lines 12094-12097 and 18630-18637 — `capabilitySettings` has asymmetric mutability and mixed ownership

JSON paths: `$.definitions.AccountProperties.properties.capabilitySettings`, `$.definitions.ProjectProperties.properties.capabilitySettings`

Two issues with the new `CapabilitySettings` model:

1. **Asymmetric mutability for the same type.** `AccountProperties.capabilitySettings` carries no mutability annotation and is therefore fully mutable (`models.tsp` line 1679), while `ProjectProperties.capabilitySettings` is `x-ms-mutability: ["read", "create"]` (`models.tsp` line 4205). Same type, two different contracts, no stated reason.
2. **Mixed ownership.** The project-level description says *"Effective agent capability settings … omitted fields inherit from the parent account when present"* — so the server populates a client-writable property from external state. The GET response will therefore diverge from the last PUT for every field the client omitted, which is the OAPI034 anti-pattern and a source of What-If and Terraform noise. The word "Effective" confirms the property conflates client intent with server resolution.

**Classification reasoning:** Introduced in this PR — `CapabilitySettings` and both usages are new in this API version, so correcting the shape now costs nothing in compatibility.

**Suggested fix:** keep `capabilitySettings` as the client-owned override, returning exactly what the client set, and add a separate `@visibility(Lifecycle.Read)` `effectiveCapabilitySettings` for the resolved value. Then align the account-level and project-level mutability annotations, or document why they differ.


---

### 11. **[NEW] 🔵 Suggestion** **[[Section 8.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#82-use-objects-instead-of-strings-for-structured-values)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 13079-13098 — mutually exclusive fields expressed only in prose

JSON path: `$.definitions.ArcDeploymentResourceRequirements`

The description says *"Specify either cpu and memory together, or gpu. GPU is supported only in limits, not requests"*, but `cpu`, `memory`, and `gpu` are three independent optional siblings. The schema therefore admits invalid combinations, and no generated client can validate the constraint. The same conditional-validity pattern applies to `deploymentTemplate` (line 12992, *"Allowed only when runtime is vllm"*) and `vllmParameters` (line 13000, *"Returned only when runtime is vllm"*).

**Classification reasoning:** Introduced in this PR — the model is new in this API version.

**Suggested fix:** consider a discriminated union keyed on `compute` so the limits shape is structurally correct for each case. At minimum, document the exact `400` error returned for each invalid combination.



---

### 12. **[NEW] 🔵 Suggestion** **[[Section 8.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#81-prefer-enums-over-booleans)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 13195-13197 — `enforceEager` boolean

JSON path: `$.definitions.ArcDeploymentVllmParameters.properties.enforceEager`

LintDiff raises `EnumInsteadOfBoolean` here. Recording this as a suggestion only, because it looks defensible: `ArcDeploymentVllmParameters` is reachable **only** through `vllmParameters` (line 13000-13004), which carries `readOnly: true`, so this is server-produced diagnostic output rather than a customer-set knob. Worth a second look only if the field could ever gain a third state.

**Classification reasoning:** Introduced in this PR — the model is new in this API version.

**Suggested fix:** no change required; leave as a boolean unless a third state is foreseeable.



---

### 13. **[NEW] 🔵 Suggestion** — PR description understates the change

This is a review-process note rather than a specification rule violation, so it carries no rule ID.

The description states that the surface is *"identical to `2026-05-15-preview`"* and characterises the change as a mechanical re-promotion of gated features plus an example copy. The prior preview contains zero occurrences of `ArcDeployment`, `CapabilitySettings`, `capabilitySettings`, or `ExtensionUnreachable`. The actual deltas are:

- a new `ArcDeployment` resource type — 2 paths, 16 definitions, 475 lines of new TypeSpec;
- a new `CapabilitySettings` model attached to both `AccountProperties` and `ProjectProperties`;
- a new `ExtensionUnreachable` member on the shared `ProvisioningState` enum;
- 198 examples and 198 `x-ms-examples` references, not the 192 stated.

This matters because the ARM review board triages by PR description, and "surface identical to prior preview" invites a rubber-stamp of an entire new resource type.

**Suggested fix:** update the PR description with the four deltas above and correct the example count.



---

### 14. **[EXISTING] 🟠 Warning** **[[Section 7.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#71-no-secrets-in-getputpatch-responses) / [Section 7.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#72-x-ms-secret-annotation)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — lines 11935 and 12718 — secrets returned without `x-ms-secret`

JSON paths: `$.definitions.AccountProperties.properties.migrationToken`, `$.definitions.ApiProperties.properties.qnaAzureSearchEndpointKey`

LintDiff raises `XMSSecretInResponse` at error level on both. Neither carries `"x-ms-secret": true`, although this file uses that annotation correctly 14 times elsewhere.

**Classification reasoning:** **Not introduced by this PR.** Both properties are byte-identical in `preview/2026-05-15-preview` and in stable `2026-07-01` — `migrationToken` is `{"type": "string", "description": "Resource migration token."}` in all three. Raising it here for visibility only; it is pre-existing technical debt carried forward, and this PR is not the right place to change a response shape.

**Suggested fix:** track separately. Annotating `@secret` — or moving retrieval behind a `list*` POST action for granular RBAC — needs its own breaking-change assessment.



---

### 15. **[EXISTING] 🔵 Suggestion** **[[Section 2.8](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#28-common-type-definitions-should-be-used)]** `.../preview/2026-07-15-preview/cognitiveservices.json` — line 2110 (first of 723 occurrences) — common-types v3

This spec references `common-types/resource-management/v3/types.json` throughout — 723 references, and zero to v5 or v6. LintDiff warns `LatestVersionOfCommonTypesMustBeUsed` 21 times on the new `arcDeployments` surface alone.

**Classification reasoning:** **Not introduced by this PR.** `preview/2026-05-15-preview` has 702 v3 references; the delta is purely the new ArcDeployment operations inheriting the existing pattern. **No new decision is required in this PR** — the emitted output follows the shared TypeSpec project, so this is all-or-nothing at the RP level.

**Suggested fix:** schedule an RP-wide migration of the `CognitiveServices.Management` TypeSpec project to common-types v6. Not actionable per-resource here.



---

### Verified as compliant — no action needed

For completeness, these were checked and found correct: `TSP-REQUIRED-V1` (not applicable — the PR adds `.tsp` under the same service folder); example integrity (all 198 `x-ms-examples` references resolve, zero dangling, zero orphans); readme suppression continuity (nothing dropped relative to the `2026-05-15-preview` tag); the `ArcDeployment` resource shape (correct `ProxyResource` base, no `tags` or `location`, both point and collection GET, async DELETE returning `202`/`204`/`default`); `$.definitions.ProvisioningState` already carries `readOnly: true`; `extensionId` and `capabilitySettings` correctly use ARM resource IDs; readme tag ordering and default tag; and the secret surface on all newly added properties. Breaking-change analysis found **zero** breaking changes — paths 103 to 105, definitions 337 to 354, no removals.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
fmabroukmsft: Fixed in `9d7ccb5f74`. Ran `tsp format`, which reflows the `ArcDeployment` model declaration to the multi-line generic form exactly as shown in your diff. Confirmed your observation that this single file was the root cause — all six "impacted" CogSvc projects (`AnomalyDetector`, `ContentSafety`, `OpenAI.Inference`, `Speech.*`, and `CognitiveServices.Management`) were reporting the same ArcDeployment.tsp diff from the folder-wide format check, and this one reformat clears all of them. TypeSpec Validation is now green.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
fmabroukmsft: Two separate points here; neither changes in this PR.

**Documentation:** This isn't specific to `ExtensionUnreachable` — every member of the `ProvisioningState` union uses the same `#suppress "documentation-required"` convention with no per-member doc comment (`Succeeded`, `Failed`, `Canceled`, etc.). Adding a real doc comment to only the new member would make it the sole inconsistent entry in the union, not bring it into line. If the union's documentation convention is to change, that's a uniform change across all members, not a one-off.

**Health-vs-state modeling:** This is a warning-level design suggestion on a shared union referenced by four models, and reshaping it (a dedicated `extensionStatus` property vs. reusing `Failed`) is a service-owner API-modeling decision with SDK-surface implications. It's out of scope for this mechanical version-creation PR and is being deferred to the API owners. Since `ExtensionUnreachable` is a brand-new member in a preview version, there's no compatibility pressure forcing the decision now.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
fmabroukmsft: Declining for the reason your own classification identifies: the sibling endpoint properties on this RP — `AccountProperties.endpoint`, `qnaRuntimeEndpoint`, `webEndpoint` — are all plain `string`. Changing only `inferenceEndpoint` to `url` would create exactly the local inconsistency you note, for no functional gain (it's a read-only, service-populated value). If the RP is going to converge on the `url` scalar for endpoint properties, that should be done uniformly in a dedicated follow-up rather than piecemeal on the one new property. No change in this PR.

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -33,6 +33,151 @@ openapi-type: arm
...
+    reason: nodeSelector and capabilities are dynamic string maps with keys that are not predefined by the service. nodeSelector holds user-defined Kubernetes node label keys used for pod scheduling; capabilities is a read-only map of model/runtime capability flags resolved at runtime. Both are backed by a Dictionary of string, matching the existing ManagedComputeDeploymentProperties.capabilities pattern and the arm-no-record suppressions in the TypeSpec source.
+    where:
+      - $.definitions.ArcDeploymentProperties.properties.nodeSelector
+      - $.definitions.ArcDeploymentProperties.properties.capabilities
+      - $.definitions.ArcDeploymentUpdateProperties.properties.nodeSelector
+  - code: PutResponseCodes
+    reason: The Arc deployment PUT is create-only; updates are supported through PATCH, so the operation intentionally returns 201 + default without 200. This matches the arm-put-operation-response-codes suppression in the TypeSpec source.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/arcDeployments/{deploymentName}"].put
```
ravimeda: **[NEW] 🔴 Blocking** **[[RPC-Put-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation) / [RPC-Put-V1-20](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `specification/cognitiveservices/resource-manager/readme.md` — line 171-174 — Arc deployment PUT is create-only: returns 201 with no 200

Emitted at `preview/2026-07-15-preview/cognitiveservices.json` line 2120 (`ArcDeployments_Create`); responses begin at line 2165 with `201` and there is no `200`. JSON path `$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/arcDeployments/{deploymentName}"].put.responses`. Source: `ArcDeployment.tsp` line 63-71.

RPC-Put-V1-11 requires PUT to return 201 for creation **and** 200 for replacement. RPC-Put-V1-20 requires PUT to be idempotent and states a re-PUT with the same body MUST NOT fail. ArcDeployment is the only PUT in this API version with that shape — **1 of 30** PUTs here, and **0 of 29** in `2026-05-15-preview`. There is no precedent in this RP. ARM template deployment re-PUTs every resource on each deployment, so a create-only PUT breaks redeployment.

On the suppression reason: the first sentence ("The Arc deployment PUT is create-only; updates are supported through PATCH") is a substantive design rationale, so let me engage it directly — the existence of PATCH does not relieve PUT of the idempotency MUST; those are independent requirements. The second sentence ("This matches the arm-put-operation-response-codes suppression in the TypeSpec source") is circular, citing the author's own TypeSpec suppression as evidence for the readme suppression.

Note both suppressions here are **new in this PR** against a **new** resource path — `arcDeployments` appears 0 times in `readme.md` at the base SHA and twice at head — so the RP's many pre-existing `PutResponseCodes` entries under older tags are not precedent for these.

**Suggested fix:** the template on line 64 is already correct — the 201-only contract comes from the `Response = ArmResourceCreatedResponse<…>` override on lines 67-70. Either drop that override so the operation emits the default 200+201 pair (the shape `Deployment.tsp` line 58 uses), or widen it to include `ArmResourceUpdatedResponse<ArcDeployment>` alongside the created response (the shape `Project.tsp` lines 62-66 uses). Dropping the override is lossless for LRO headers — line 66 already sets `LroHeaders = ArcDeploymentCreateLroHeaders` and the library default threads that same value into `ArmResourceCreatedResponse<Resource, LroHeaders>`, so nothing needs re-adding.

Then remove the `arm-put-operation-response-codes` suppression at `ArcDeployment.tsp` line 63 and **both** readme suppressions (lines 171-174 and 175-178) — `ProvisioningStateSpecifiedForLROPut` is expected to become moot once a 200 response exists, though confirm that against the first CI run rather than assuming it. Remove the suppressions only after the source change lands, otherwise CI fails the moment they are gone.

If the service team maintains that create-only is required, [suppression-review-criteria](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#decision-framework) routes a new-resource-type violation with a reasonable justification to **escalate to peer review** — that means ARM API review board sign-off, not a self-approved suppression.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -4146,6 +4196,13 @@ model ProjectProperties {
    */
   @visibility(Lifecycle.Read)
   isDefault?: boolean;
+
+  /**
+   * Effective agent capability settings for the project. Optional partial override of the account defaults; omitted fields inherit from the parent account when present. Settable only at create time.
+   */
+  @added(Versions.v2026_07_15_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  capabilitySettings?: CapabilitySettings;
```
ravimeda, line 4307: **[NEW] 🟠 Warning** **[[OAPI034](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034) / [Section 8.20](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#820-fields-must-have-clear-ownership--server-or-client-oapi034)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` — line 4200-4205 — client-writable property is populated by the server

Emitted at `cognitiveservices.json` line 18630, JSON path `$.definitions.ProjectProperties.properties.capabilitySettings`, with `x-ms-mutability: ["read","create"]`.

The description reads: _"Effective agent capability settings for the project. Optional partial override of the account defaults; omitted fields inherit from the parent account when present. Settable only at create time."_ That is a client-writable property whose omitted sub-fields the server fills in from the parent account, so the GET response diverges from the PUT body — the exact anti-pattern OAPI034 forbids. It produces false drift in ARM Template What-If, Terraform, and Change Analysis.

**Suggested fix:** split the property — keep a client-owned `capabilitySettings` echoed back exactly as supplied, and add a separate read-only `effectiveCapabilitySettings` exposing the resolved account+project blend.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
```diff
@@ -0,0 +1,477 @@
...
+/**
+ * SKU for an Arc deployment.
+ */
+@added(Versions.v2026_07_15_preview)
+model ArcDeploymentSku {
+  /**
+   * The name of the Arc deployment SKU. Must be Arc.
+   */
+  name: "Arc";
+}
```
ravimeda, line 441: **[NEW] 🟠 Warning** **[[Section 8.3](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#83-use-enums-for-finite-value-sets-actively-review) / [enum-best-practices](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/enum-best-practices.md) / [Section 2.6.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#261-standard-sku-schema)]** `specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp` — line 428-433 — SKU modeled as a closed single-value enum with no enum name

Emitted at `cognitiveservices.json` line 13123-13135, JSON path `$.definitions.ArcDeploymentSku.properties.name`, as `"enum": ["Arc"]` with `"x-ms-enum": { "modelAsString": false }` and no `name` field. `sku` is in `$.definitions.ArcDeployment.required`.

Four points: (a) `modelAsString: false` produces a **closed** enum, so adding a second Arc SKU later is a breaking change for every generated SDK; (b) the `x-ms-enum` object has no `name`, so generators have no stable type name; (c) `ArcDeploymentSku` defines only `name` with no `tier`, and ignores the `Sku` definition this document already provides at `$.definitions.Sku` (line 20740: `name, tier, size, family, capacity`) — Section 2.6.1 makes only `name` required so the missing `tier` is not itself a breach, but it says the SKU **SHOULD** reference a standard SKU model, and the bespoke envelope is what forces the `arm-resource-invalid-envelope-property` suppression at line 36; (d) `sku` is required on every create with exactly one legal value, which is pure boilerplate for the customer.

**Suggested fix:**

```tsp
union ArcDeploymentSkuName {
  string,

  /** Arc SKU. */
  Arc: "Arc",
}

model ArcDeploymentSku {
  name: ArcDeploymentSkuName;
}
```

This emits `modelAsString: true` with a proper `x-ms-enum.name`. Also worth asking whether `sku` needs to be required at all when only one value is legal, and whether the existing `Sku` definition can be reused.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -1666,6 +1671,45 @@ model AccountProperties {
    * Specifies the projects, by project name, that are associated with this resource.
    */
   associatedProjects?: string[];
+
+  /**
+   * Reusable default agent capability settings inherited by child projects.
+   */
+  @added(Versions.v2026_07_15_preview)
+  capabilitySettings?: CapabilitySettings;
```
ravimeda, line 1685: **[NEW] 🔵 Suggestion** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` — line 1676-1679 — is the Account/Project mutability asymmetry intended?

`AccountProperties.capabilitySettings` has no `@visibility` decorator and emits with no `x-ms-mutability` (`cognitiveservices.json` line 12094), i.e. fully mutable. `ProjectProperties.capabilitySettings` (line 4203-4205) carries `@visibility(Lifecycle.Read, Lifecycle.Create)` and emits `x-ms-mutability: ["read","create"]`.

To be clear, the Account shape is **legal** — [Section 8.6](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#86-visibility-and-mutability) says fully mutable properties do not need annotation, so this is a question rather than a violation. But the same `CapabilitySettings` type is fully mutable on Account and create-only on Project, which means a customer can change the account default but can never re-point a project. If that is deliberate, no change needed. If the storage bindings are immutable once agents have written data, Account likely wants `@visibility(Lifecycle.Read, Lifecycle.Create)` too.

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -33,6 +33,151 @@ openapi-type: arm
...
+  - code: AvoidAdditionalProperties
+    reason: EvaluateDeploymentPoliciesResponse.results is a dictionary keyed by deployment name, mapping each hypothetical deployment to its policy evaluation result. Keys are user-provided deployment names from the request.
+    where:
+      - $.definitions.EvaluateDeploymentPoliciesResponse.properties.results
+  - code: AvoidAdditionalProperties
+    reason: nodeSelector and capabilities are dynamic string maps with keys that are not predefined by the service. nodeSelector holds user-defined Kubernetes node label keys used for pod scheduling; capabilities is a read-only map of model/runtime capability flags resolved at runtime. Both are backed by a Dictionary of string, matching the existing ManagedComputeDeploymentProperties.capabilities pattern and the arm-no-record suppressions in the TypeSpec source.
+    where:
+      - $.definitions.ArcDeploymentProperties.properties.nodeSelector
+      - $.definitions.ArcDeploymentProperties.properties.capabilities
+      - $.definitions.ArcDeploymentUpdateProperties.properties.nodeSelector
```
ravimeda, line 179: **[NEW] 🔵 Suggestion** **[[Section 8.5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#85-additionalproperties-restriction) / [suppression-review-criteria](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md)]** `specification/cognitiveservices/resource-manager/readme.md` — line 165-170 — one suppression bundles a user-defined map with a service-owned one

This entry covers three `where:` targets. `ArcDeploymentProperties.nodeSelector` and `ArcDeploymentUpdateProperties.nodeSelector` are genuine user-defined pass-through data (Kubernetes node label keys), which Section 8.5 explicitly permits. `ArcDeploymentProperties.capabilities` is different — a service-owned, read-only map whose keys the service itself resolves, which is what Section 8.5 forbids. Bundling them means the questionable target inherits the justification of the sound ones.

**Suggested fix:** split into two entries so the `nodeSelector` targets stand on their own, and for `capabilities` either model the known capability flags as explicit named properties, or keep the suppression with a reason specific to `capabilities` explaining why the key set cannot be bounded.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
```diff
@@ -0,0 +1,477 @@
...
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  compute: ArcDeploymentComputeType;
+
+  /**
+   * Optional deployment template identifier for advanced vLLM tuning. Allowed only when runtime is vllm.
+   * Example: azureml://registries/{registry}/deploymenttemplates/{template}/versions/{version}
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  deploymentTemplate?: string;
```
ravimeda, line 144: **[NEW] 🔵 Suggestion** **[[Section 8.4](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#84-use-specific-types-instead-of-generic-strings-actively-review)]** `specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp` — line 144-149 — structured identifier typed as unconstrained `string`

The doc comment gives the exact form `azureml://registries/{registry}/deploymenttemplates/{template}/versions/{version}`, but the property emits with keys `type`, `description`, `x-ms-mutability` only — no `format`, no `pattern`.

You declined the earlier `inferenceEndpoint` suggestion on two grounds: primarily that the sibling endpoint properties on this RP (`AccountProperties.endpoint`, `qnaRuntimeEndpoint`, `webEndpoint`) are all plain `string` so changing one would create local inconsistency, and secondarily that it is a read-only, service-populated value. Neither leg reaches `deploymentTemplate` — it is a client-writable create-time input (`@visibility(Lifecycle.Read, Lifecycle.Create)`, confirmed by the emitted `x-ms-mutability: ["read","create"]`), not service-populated, and there is no sibling set of `azureml://` template identifiers for it to stay consistent with. Raising it separately rather than re-litigating the settled `inferenceEndpoint` thread.

**Suggested fix:** add a `@pattern` constraint matching the documented `azureml://` form so malformed values are rejected at the API boundary rather than deep in the deployment pipeline. (Not suggesting `format: uri` — that would change the generated SDK type for an asset identifier.)

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`9d7ccb5`](https://github.com/Azure/azure-rest-api-specs/pull/44334/commits/9d7ccb5f74ed60c681518a11d5b0d16686dbeb0b). See inline comments for findings 1-6.

specification/cognitiveservices/resource-manager/readme.md
fmabroukmsft: The factual characterization is correct: `nodeSelector` is user-defined pass-through (permitted), and `capabilities` is a read-only, service-resolved map.

On `capabilities` as the questionable target — this is the established pattern in this RP, not a new one. `ManagedComputeDeploymentProperties.capabilities` has shipped as a read-only `additionalProperties` string map since `2026-05-15-preview`, carried by its own `AvoidAdditionalProperties` suppression in `readme.md` (line 119) — the pattern this finding cites. `ArcDeploymentProperties.capabilities` is modeled identically: a read-only map of model/runtime capability flags whose key set is resolved at runtime and cannot be statically bounded, so the flags cannot be enumerated as named properties without a breaking change each time a new capability ships.

The suppression is therefore consistent with accepted RP practice and is retained. No change in this PR.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
fmabroukmsft: The premise that "there is no sibling set of `azureml://` template identifiers for it to stay consistent with" does not hold. `ManagedComputeDeploymentProperties.deploymentTemplate` has shipped since `2026-03-15-preview` (and `2026-05-15-preview`) as an unconstrained `string`, client-writable (`x-ms-mutability: ["read","create"]`), documenting the same `azureml://registries/{registry}/deploymenttemplates/{template}/versions/{version}` form — plus a broader project-scoped `projects/{project}/deploymentTemplates/{template}/versions/{version}` form.

`ArcDeploymentProperties.deploymentTemplate` mirrors that established sibling — same name, same `string` type, same `read,create` mutability, same `azureml://` semantics. Adding a `@pattern` to only the Arc copy would introduce the local inconsistency the earlier `inferenceEndpoint` thread declined for: one member of a matching set constrained while its established sibling stays open. It would also need reconciling against the broader form the ManagedCompute sibling already accepts (registry URI or project-scoped path), so an Arc-only regex risks rejecting inputs the sibling permits.

Constraining deployment-template identifiers would be a uniform change across both `deploymentTemplate` properties (reconciling the two accepted forms) in a dedicated follow-up, not a piecemeal constraint on the one new copy. No change in this PR.

specification/cognitiveservices/resource-manager/readme.md
morel-israel: Hi Ravi, thank you for the elaborated comment.
Is it possible to support 200 only for a re-PUT with the same body? But not for model update?
If so we can support it this way. We can also support changing only the subset of the fields that we support in PATCH with PUT.
What is your suggestion?

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
morel-israel: You are correct, we had this conversation internally and the decision was we should have the sku even though its one value only. But you have a good point we might want to expand it in the future, I'll change that to enum. Thank you.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
JKrame: Hi Ravi, this is intended. 

Account CapabilitySettings only serve as a default for child resource Project CapbilitySettings when no override is provided on the project. Project CapabilitySettings however will reflect backend bindings which are immutable through the project's existence.

specification/cognitiveservices/CognitiveServices.Management/main.tsp
```diff
@@ -104,6 +105,11 @@ enum Versions {
    * The 2026-07-01 API version.
    */
   v2026_07_01: "2026-07-01",
+
+  /**
+   * The 2026-07-15-preview API version.
+   */
+  v2026_07_15_preview: "2026-07-15-preview",
```
ravimeda, line 112: **[NEW] 🔴 Blocking** **[CI: TypeSpec Validation - Compile rule]** `specification/cognitiveservices/CognitiveServices.Management/main.tsp` - line 108-112 - new API version is not registered in `service.yaml`; the required TypeSpec Validation check fails

The required **TypeSpec Validation** check fails at this commit (run `31440960488`, job `93625375641`, `Rule Compile failed`). The compile itself succeeds, the log records `Compilation completed successfully.`, so this is not a TypeSpec error. The failure is that `tsp compile` **modifies a checked-in file that this PR does not include**:

```diff
--- a/specification/cognitiveservices/CognitiveServices.Management/service.yaml
+++ b/specification/cognitiveservices/CognitiveServices.Management/service.yaml
@@ -95,3 +95,7 @@ versions:
     source: typespec
     swagger-files:
       - ../resource-manager/Microsoft.CognitiveServices/stable/2026-07-01/cognitiveservices.json
+  - version: 2026-07-15-preview
+    source: typespec
+    swagger-files:
+      - ../resource-manager/Microsoft.CognitiveServices/preview/2026-07-15-preview/cognitiveservices.json
```

The log then reports: ``Files have been changed after `tsp compile`. Run `tsp compile` and ensure all files are included in your change.``

`service.yaml` was added to this project on `main` by commit `f9d6a265` ("Migrate readme.md swagger versions to service.yaml", https://github.com/Azure/azure-rest-api-specs/pull/44805, 2026-08-06), about eleven hours after this PR's previous commit was validated. This branch predates it, so the branch tree has no `service.yaml` at all, and on the merged tree the file still ends at `2026-07-01`. The new version is registered in `readme.md` (new tag block at line 36) but not in `service.yaml`, which is now the file that enumerates a service's API versions for this project.

**Classification reasoning:** Introduced in this PR. TypeSpec Validation succeeded at the previous PR commit `9d7ccb5f` (check run `92487634104`) and fails at `6c193d2f` (check run `93625375641`); `2026-07-15-preview` is this PR's new version, so registering it is this PR's responsibility.

**Suggested fix:** merge `main` into the branch, then run the command the check itself prescribes and commit the regenerated file:

```bash
npm ci
npx tsv specification/cognitiveservices/CognitiveServices.Management
```

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -1666,6 +1671,135 @@ model AccountProperties {
...
+
+  /**
+   * Customer-owned AKS hosting configurations for Foundry agents. This API version supports exactly one configuration. Once set, the configuration is immutable; subsequent requests may only repeat the same value.
+   */
+  @added(Versions.v2026_07_15_preview)
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  @minItems(1)
+  @maxItems(1)
+  @identifiers(#["name"])
+  agentHostingConfigurations?: AgentHostingConfiguration[];
```
ravimeda, line 1695: **[NEW] 🟠 Warning** **[[Section 8 - Polymorphic Types](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#8-polymorphic-types)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 1684-1689 - array of polymorphic objects on an updatable resource

`AccountProperties.agentHostingConfigurations` (emitted at `preview/2026-07-15-preview/cognitiveservices.json` line 12101, JSON path `$.definitions.AccountProperties.properties.agentHostingConfigurations`) is an array whose `items` `$ref` `AgentHostingConfiguration` (line 12493), which declares `"discriminator": "hostingType"`. `Account` is an updatable resource: `accounts/{accountName}` exposes PATCH with body `#/definitions/Account`.

Section 8 states: *"SHOULD NOT have array properties containing polymorphic objects on updatable resources (JSON Merge Patch cannot handle this version-resiliently)."*

This is the only array of polymorphic objects in the document. The other five discriminated bases (`AgentDeploymentProperties`, `ApplicationAuthorizationPolicy`, `ComputeProperties`, `ConnectionPropertiesV2`, `OutboundRule`) are all used as single-valued properties, so there is no precedent in this RP for the array form.

`x-ms-mutability: ["read","create"]` (line 12112-12115) and `maxItems: 1` (line 12105) narrow the hazard but do not remove it. An SDK built against this version deserializes a future `hostingType` variant into the base type, losing the derived fields; a GET then PUT round-trip sends back a truncated element, which the service must then reject as an attempted change to an immutable property ([OAPI031](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#819-immutable-property-tolerance-and-enforcement-oapi030-oapi031)).

**Classification reasoning:** Introduced in this PR. `agentHostingConfigurations`, `AgentHostingConfiguration`, `AgentHostingType`, and `ManagedClusterAgentHostingConfiguration` are all absent from `preview/2026-05-15-preview` and `stable/2026-07-01`.

**Suggested fix:** two options.

(a) Since this API version supports exactly one configuration, model it as a single property. This removes the array-of-polymorphic shape and still leaves room to introduce an array in a later version:

```tsp
@added(Versions.v2026_07_15_preview)
agentHostingConfiguration?: AgentHostingConfiguration;
```

(b) Keep the array and document the round-trip contract explicitly, so the service treats an element whose unknown derived fields are absent as unchanged rather than as a change.

ravimeda, line 1695: **[NEW] 🔵 Suggestion** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 1681-1689 - can an existing account adopt BYOC agent hosting?

To be clear up front: the annotation here is **correct**. `agentHostingConfigurations` is create-only and carries `@visibility(Lifecycle.Read, Lifecycle.Create)`, emitting `x-ms-mutability: ["read","create"]` at `cognitiveservices.json` line 12112-12115, which is exactly what an immutable property should carry. This is a design question with no rule behind it, not a compliance finding.

Under those semantics an account created without a hosting configuration can never gain one: adopting BYOC agent hosting on an existing Foundry account would require deleting and recreating the account. The description says *"Once set, the configuration is immutable; subsequent requests may only repeat the same value"*, which documents what happens after the value exists but not whether it can be set on an account that does not yet have one.

Two possibilities:

- If the service does allow setting it on an existing account that has none, then `read,create` is too strict and the property is really create-or-set-once.
- If it genuinely cannot be set later, it would help to say so in the description, since account creation then becomes an irreversible decision point for BYOC adoption.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`6c193d2`](https://github.com/Azure/azure-rest-api-specs/pull/44334/commits/6c193d2fb01365b1d678f6ea58d6a3796a4e9aa3). See inline comments for findings 1-3.

specification/cognitiveservices/resource-manager/readme.md
ravimeda: Thanks. Your first option is close, and I don't think this needs escalation. One assumption to clear first, since I think it's what makes this look bigger than it is.

**Adding `200` does not mean supporting model update.** In ARM, `200` vs `201` signals only whether the resource already existed: `201` when PUT creates it, `200` when PUT lands on one that exists. It says nothing about which properties are mutable. You can add `200` and keep today's mutable surface unchanged.

**Proposed behaviour.** The set of properties that can change is identical to what PATCH already allows, so nothing widens:

| Request | Response |
| --- | --- |
| Resource does not exist | `201` |
| Exists, body identical | `200` |
| Exists, create-only field changed (`model`, `extensionId`, `runtime`, `compute`, `deploymentTemplate`) | `400` naming the property |
| Exists, only `replicas` / `resources` / `nodeSelector` changed | `200` |

Rows 2 and 3 are [OAPI030 / OAPI031](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#819-immutable-property-tolerance-and-enforcement-oapi030-oapi031): repeating a value must succeed, changing an immutable one must be rejected with `400`. That gives you exactly "re-PUT works, model update does not". Row 4 is your second option.

**Assumptions behind my "no escalation needed" claim.** Please sanity check these, since you know the backend:

1. You already reject changes to the create-only fields, so row 3 is existing behaviour.
2. PUT can route rows 2 and 4 to the handler PATCH already uses for `replicas`, `resources`, and `nodeSelector`, rather than needing new update logic.
3. A re-PUT with an identical body currently fails. If it already no-ops, row 2 is done.
4. The create response body already carries `provisioningState`, per your `ProvisioningStateSpecifiedForLROPut` suppression reason.

If assumption 2 does not hold and PUT cannot reach the PATCH path, say so and we should reconsider rather than have you build new plumbing.

**One gotcha.** PUT is full replacement, not merge ([Section 3.2 / WHATIF-004](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#32-put-must-not-exhibit-patch-semantics-whatif-004)). An omitted optional mutable property must be reset, not preserved: `nodeSelector`, `deploymentState`, `raiPolicyName`. The six required properties cannot be omitted, so they are safe. Preserving an omitted `nodeSelector` would show up as false drift in What-If and Terraform.

**The spec change is a deletion.** `ManagedComputeDeployment.tsp` line 64 already has the shape you want: `createOrUpdate is ArmResourceCreateOrReplaceAsync<ManagedComputeDeployment>;`, emitting `200/201/default`. Dropping the `Response =` override and the suppression at `ArcDeployment.tsp` lines 63-71 does the same here:

```tsp
  create is ArmResourceCreateOrReplaceAsync<
    ArcDeployment,
    LroHeaders = ArcDeploymentCreateLroHeaders
  >;
```

That should retire both suppressions on this path: `PutResponseCodes` directly, and `ProvisioningStateSpecifiedForLROPut` because it fires only for want of a `200` body.

**If row 4 cannot land in this preview**, rows 1 to 3 alone are an acceptable interim. It restores idempotency, but a customer could not change `replicas` by editing and redeploying a template, so I would treat it as a preview stopgap to close before GA.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
ravimeda: Thanks, that works. Noting for tracking that it is not in the tree yet at `6c193d2f`: `$.definitions.ArcDeploymentSku.properties.name` is still `"enum": ["Arc"]` with `"x-ms-enum": { "modelAsString": false }` and no `name`. Leaving this open until it lands.

For the extensible shape, a `union` with a `string` variant rather than the bare literal gets you `modelAsString: true` plus a stable `x-ms-enum.name`:

```tsp
union ArcDeploymentSkuName {
  string,

  /** Arc-hosted deployment. */
  Arc: "Arc",
}
```

morel-israel: Fixed :)

specification/cognitiveservices/resource-manager/readme.md
morel-israel: That is great, we can support it.
Changed ;)

gary-x-li:
## ARM API Review — open items for author

Three review threads remain open and need author action. _(The two previously-flagged blocking items — Arc `PUT` returning `200` and `2026-07-15-preview` registration in `service.yaml` — plus the extensible-SKU change are already fixed in the tree and only await thread resolution; they are not listed here.)_

Please address the following:

1. 🟠 **Warning — field ownership (OAPI034):** [thread](https://github.com/Azure/azure-rest-api-specs/pull/44334#discussion_r3730666911)
   `ProjectProperties.capabilitySettings` is client-writable (`x-ms-mutability: ["read","create"]`) but the server fills omitted sub-fields from the parent account, so the GET response diverges from the PUT body — false drift in ARM What-If, Terraform, and Change Analysis. **Suggested:** split into a client-owned `capabilitySettings` (echoed back exactly as supplied) plus a read-only `effectiveCapabilitySettings` exposing the resolved account+project blend.

2. 🟠 **Warning — polymorphic array on an updatable resource:** [thread](https://github.com/Azure/azure-rest-api-specs/pull/44334#discussion_r3759383109)
   `AccountProperties.agentHostingConfigurations` is an array of discriminated (`hostingType`) objects on an updatable (PATCH) resource; JSON Merge Patch cannot handle this version-resiliently, so a future `hostingType` variant is truncated on a GET→PUT round-trip. `maxItems: 1` and create-only narrow but do not remove the hazard. **Suggested:** model it as a single property (still supports exactly one configuration today), leaving room to introduce an array in a later version.

3. 🔵 **Suggestion — BYOC adoption:** [thread](https://github.com/Azure/azure-rest-api-specs/pull/44334#discussion_r3759383120)
   `agentHostingConfigurations` is create-only. Please clarify whether an existing account can adopt BYOC agent hosting after creation, or whether it must be recreated.

Please reply on each thread with the change or rationale so we can resolve them.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
JKrame: Today, there's already a few things at play which is already existing behavior.
We have this concept of [Capability Host](https://learn.microsoft.com/en-us/azure/foundry/agents/concepts/capability-hosts) which already supports this Account Capability Host -> Project Capability Host BYO Resource Inheritance at the service layer. This means that customer can provide all connections on the Account Capability Host and nothing on the Project Capability Host and the Project Capability Host GET would return the Account Capability Host list (as it was inherited and persisted to the project level capability host object).
 
Second, we're attempting to fold this concept of Capability Host into the Account/Project flow to reduce the concept burden which customers have trying to onboard onto Foundry Agents. The addition of CapabilitySettings in the API here is effectively moving the BYO settings of Account and Project Capability Host to the Account/Project itself, which will then provision Capability Host implicitly with the correct BYO settings.

Third; Due to the inability to deprecate Capability Host immediately based on deprecation standards, we also have to support these APIs for some period of time longer - meaning Customer is also able to call these APIs directly which are effectively swapping the capability settings under the hood.
 
Based on these three things, we have the following requirements
1) Account -> Project CapabilitySetting Inheritance must be retained
2) CapabilitySettings must display the BYO Connections of the underlying capability host - as (at this point) those connections can change at any time via Capability Host replacement

It sounds to me as though the ask is to have capabilitySettings which will indefinitely beyond creation show the create configuration of the resource, and effectiveCapabilitySettings should display the actual BYO resources of the underlying capability host. 
That said, this seems like a confusing user experience to me - I do not see much value add in retaining a value which displays potentially wrong data, and customers may be confused by the presence of both fields.

Note that the eventual goal would be to remove customer handle on Capability Host so that capabilitySettings must be changed only from account/project layer (with an eventual goal to make these updateable beyond create time)

Can you please help me understand if my above understanding of the ask is correct, and if this is simply a suggestion or a hard requirement?

specification/cognitiveservices/CognitiveServices.Management/models.tsp
frankliu20: Thanks for calling this out. This is intentional: an existing account created without  `agentHostingConfigurations` cannot adopt BYOC agent hosting later. BYOC adoption is an account-creation-time decision, and changing or adding the configuration afterward requires creating a new account

frankliu20: Thanks for raising this. We agree that this is an intentional exception to the Section 8 “SHOULD NOT” guidance.

Multiple hosting configurations are planned for a future API version. Using a singular property now and replacing it with an array later would be a breaking wire-contract change, whereas retaining the array allows a future version to increase  maxItems  without changing the property type.

The JSON Merge Patch risk is constrained by the service contract:
•  `agentHostingConfigurations`  is create-only and immutable; it cannot be added or changed through PATCH, so array replacement is not used to update this property.
• This version limits the array to one element.
• The service will only return discriminator values defined by the requested API version. In particular,  `2026-07-15-preview`  will not later return a new  `hostingType`, so an SDK generated for this version will not deserialize and replay an unknown derived type.
• Unrelated account updates may omit this property. If it is replayed, only the existing unchanged value is accepted.

Future hosting types and support for multiple configurations will be introduced through a new API version. We’ll clarify these guarantees in the property documentation and record the intentional guideline exception.

specification/cognitiveservices/resource-manager/readme.md
ravimeda: Verified fixed at `783a5a70` - thanks for taking option (a).

`ArcDeployments_CreateOrUpdate` now returns `200`, `201`, and `default` (swagger line 2124 and 2168-2173), and the source at `ArcDeployment.tsp` line 63-66 is the bare `ArmResourceCreateOrReplaceAsync` form with the `Response =` override removed. The `arm-put-operation-response-codes` suppression is gone from the TypeSpec source, and as expected both readme suppressions on this path retired with it - no `PutResponseCodes` or `ProvisioningStateSpecifiedForLROPut` entry is scoped to any Arc path at head. LintDiff is green at this commit.

Resolving.

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
ravimeda: Verified fixed at `783a5a70`. `ArcDeployment.tsp` line 419-430 adds `union ArcDeploymentSkuName` with the `string` variant, and line 440 references it. The emitted `ArcDeploymentSkuName` (swagger line 13109-13126) now carries `modelAsString: true` and a proper `x-ms-enum.name`, so points (a) and (b) from the original comment are closed.

To be explicit about what this does not close, so it does not get lost once the thread goes green: point (c) - `ArcDeploymentSku` still exposes only `name` and does not reuse the `Sku` definition this document already provides, which the sibling `ManagedComputeDeployment.sku` references - and point (d) - `sku` remains required with exactly one legal value. Both are also carried by finding 5 in the 2026-08-05 review body, which remains open.

Resolving on (a) and (b).

specification/cognitiveservices/CognitiveServices.Management/main.tsp
ravimeda: Verified fixed at `783a5a70`. `service.yaml` lines 98-101 now register `2026-07-15-preview` against the preview swagger path, and the required TypeSpec Validation check passes at this commit.

Resolving.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
ravimeda: Thanks for the detailed contract. Recording the outcome: the array shape stays as an intentional exception to the Section 8 "SHOULD NOT" guidance, on the strength of the guarantees you listed - create-only and immutable, `maxItems: 1` in this version, responses limited to hosting types defined by the requested API version, and replay of the unchanged existing value.

The description at `models.tsp` line 1688 now carries most of that, which is the documentation half of your commitment. Two things I am noting rather than pursuing here: the "record the intentional guideline exception" half has not landed anywhere in the tree, and the round-trip clause as written ("send the complete existing value unchanged") states a caller obligation, whereas the version-resilience concern is about what the service tolerates when a caller cannot honour it. Both are for the human ARM reviewer to weigh; neither blocks this thread.

Resolving as declined with rationale.

ravimeda: Thanks - that answers it, and the description now says so. `models.tsp` line 1688 states that the property can only be specified when the account is created and that an existing account without a hosting configuration cannot add one later, which is the clarification this thread asked for.

Resolving.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
ravimeda: Thanks, this is helpful context.

**On your reading of the ask.** Close, but not quite. The suggestion was not to keep a frozen create-time value next to the real one. It was to separate two things that share one field today: what the customer declared, and what the service resolved. The declared value is not wrong data. It is customer intent, and intent is what ARM Template What-If, Terraform, and Change Analysis compare against. If the only visible value is the resolved one, and a side channel can change it, those tools report drift on every run.

Your concern about concept burden is fair, though. If Capability Host is going away, adding permanent API surface to model a temporary state is a poor trade.

**Suggestion or hard requirement?** OAPI034 is marked Severity: Required in the ARM guidance, so in guideline terms it is a MUST. There is no LintDiff rule behind it, so it will not block this merge. It is enforced at human ARM review. Preview is the cheapest time to settle it, since the property name becomes contract once SDKs generate.

**One point to settle first.** Two of your statements conflict, and the answer decides the rest:

- On [r3751235376](https://github.com/Azure/azure-rest-api-specs/pull/44334#discussion_r3751235376): project settings "reflect backend bindings which are immutable through the project's existence."
- Here: connections "can change at any time" via Capability Host replacement.

If the second holds, there is an issue independent of the two-field question. `ProjectProperties.capabilitySettings` is `@visibility(Lifecycle.Read, Lifecycle.Create)` and emits `x-ms-mutability: ["read", "create"]` (`models.tsp` line 4300-4301), while `CapabilityHosts.createOrUpdate` and `ProjectCapabilityHosts.createOrUpdate` are separate PUTs on separate paths. Terraform maps create-only to ForceNew, so a service-side change becomes a permanent diff on a field it cannot update in place, and the plan proposes to destroy and recreate the project. OAPI029 also applies: a property declared immutable must always be immutable.

**Options, depending on your answers:**

| Override needed per project at create | Bindings change after create | Suggested shape |
| --- | --- | --- |
| No | Either | Make `capabilitySettings` read-only on Project. One field, inheritance preserved, the service is free to update it, and read-only fields are excluded from What-If and Terraform diffing. |
| Yes | No | Keep `read, create`. Only create-time inheritance remains, which is a smaller issue. Echo back what the customer supplied, or document the inheritance as default resolution. |
| Yes | Yes | Two concerns in one field. Split it, or drop `Lifecycle.Create` so it is honestly mutable, which at least lets Terraform converge with an update instead of a replace. |

The first row looks closest to your stated end state, where the customer handle on Capability Host goes away and capability settings are set only from the account or project layer. Until then the project value is service-resolved, so modeling it as server-owned gives you the single field you want. A writable project property can be added later when Capability Host retires, which is additive.

One note on naming: `effectiveCapabilitySettings` was illustrative, not a requirement.

This is a Warning rather than a merge blocker, and the design call is yours. Once you confirm which row applies and note the direction here, please go ahead and resolve this conversation. If the first row fits, the change is a single visibility edit that could land in this PR. If you would rather take it as a follow up, a short note here on the intended shape is enough to close this out.

ravimeda (approved):

specification/cognitiveservices/CognitiveServices.Management/ArcDeployment.tsp
morel-israel: Correct, (a) and (b) are fixed. For (c)/(d), the current shape is intentional: Arc deployments require an explicit SKU envelope for consistency, but Arc only supports  sku.name = Arc ; capacity/tier/size/family are not valid for this resource. Reusing the existing Sku model would expose unsupported fields such as capacity , so we kept a narrower  ArcDeploymentSku  shape.

rkuthala (approved):

PR #44358, opened 2026-07-01
Remove duplicate #suppress directives

timotheeguerin:
## Summary

Removes duplicate `#suppress` directives across TypeSpec specs, as flagged by the typespec integration check ([run](https://github.com/microsoft/typespec/actions/runs/28453438759/job/84322130082?pr=11113)) which now validates against `duplicate-suppression`.

## Details

- 538 duplicate suppression lines removed across 83 files (pure deletions, no additions).
- Within each suppression stack, the **first** occurrence of a rule id is kept and any later duplicates of the same rule are removed. This covers both consecutive duplicates and duplicates interleaved with other suppressions on the same node.
- Also fixed `SecuritySolutionsAPI/TopologyResource.tsp` (not in the original log) which had duplicate `no-openapi` suppressions.

## Validation

- Verified 0 remaining duplicate suppression stacks across all `specification/**/*.tsp`.
- Compiled affected projects (`advisor`, `SecuritySolutionsAPI`) with `tsp compile --no-emit`: no diagnostics, confirming the remaining single suppress still covers the diagnostic.

### DIFF
--- a/specification/advisor/resource-manager/Microsoft.Advisor/Advisor/back-compatible.tsp
+++ b/specification/advisor/resource-manager/Microsoft.Advisor/Advisor/back-compatible.tsp
@@ -4,17 +4,14 @@ using Azure.ClientGenerator.Core;
 using Microsoft.Advisor;
 
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "@flattenProperty decorator is not recommended to use."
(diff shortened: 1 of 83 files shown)


markcowl (approved):
I continue to be frustrated by the failing checks.

PR #44362, opened 2026-07-01
Add FrontDoor provisioning emitter config

ArcturusZhang:
Adds the C# provisioning emitter configuration for FrontDoor so Azure.Provisioning.FrontDoor can migrate from the reflection generator to the TypeSpec provisioning generator.

Also preserves the existing provisioning resource name `FrontDoorResource` via C# client naming customization.

Corresponding SDK PR: Azure/azure-sdk-for-net#60575


specification/frontdoor/resource-manager/Microsoft.Network/FrontDoor/back-compatible.tsp
```diff
@@ -66,6 +66,7 @@ using Microsoft.Network;
   FrontDoors.createOrUpdate::parameters.resource,
   "frontDoorParameters"
 );
+@@clientName(FrontDoor, "FrontDoorResource", "csharp");
```
ArcturusZhang:
```suggestion
```

live1206 (approved):

PR #44364, opened 2026-07-01
Add Blob Access Tier to Get Blob Response (TypeSpec)

nickliu-msft:
References https://github.com/Azure/azure-rest-api-specs/pull/43921


vincenttran-msft (commented):
Content LGTM, but can you update `tspconfig.yaml` since none of these features will get generated since the API version is stuck on an older version:

ie. in **main**:
`api-version: "2026-04-06"`

Same applies for other PR.

vincenttran-msft (approved):

PR #44378, opened 2026-07-02
Import client.tsp from main.tsp for Waves

JialinHuang803:
Imports `client.tsp` from `main.tsp` for the Microsoft.Migrate Waves spec and enables the `omit-unreachable-types` emitter option in `tspconfig.yaml`. Regenerated the Swagger for all preview API versions, which removes schemas that are no longer reachable from any emitted request/response body (e.g. `AzureResourceProperties` inlined via `is`, and the `MigrateProject` graph used only as parent/action scope).


timotheeguerin (approved):

JialinHuang803:
@sorabhtandon @satravi could you help approve this PR if it looks good to you? We are enforcing client.tsp import from main.tsp and this is part of the migration work. As a result, some 'unreachable' types need to be removed from the generated Swagger. These unreachable types are not used in the schema and the API surface is unchanged.

JialinHuang803:
Merging this PR as I didn't get response from service team.

PR #44405, opened 2026-07-03
Move OpenAI.OutputItem to use the OpenAI Package

rhurey:
Opening as a draft for comments.


nick863 (approved):

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -1,4 +1,5 @@
 import "../memory-stores/models.tsp";
+import "../toolboxes/models.tsp";
```
dargilco: Why is this needed?

nick863: Fixed.

dargilco (approved):

PR #44448, opened 2026-07-06
Add exact C# names for management enum members

ArcturusZhang:
Moves selected C# enum member compatibility names into TypeSpec using exact client names so the .NET SDK no longer needs SDK-side CodeGenMember enum member customization files.

Also pins C# emitter API versions for the affected services where current TypeSpec defaults had advanced beyond the existing SDK API version, keeping the SDK regeneration diff focused on enum member names.

SDK PR: Azure/azure-sdk-for-net#60681


ArcturusZhang:
Trivial SDK configuration updates. No need for ARM review.

live1206 (approved):

PR #44458, opened 2026-07-06
ci: adds a workflow to validate foundry clients

baywet:
so we can detect potential breakages early on


baywet:
after chatting with @timotheeguerin I learnt that CI is already in place for SDKs (through ADO pipelines), and they already account for the sub-directory structure that's in place.
They are the most reliable at the moment for a variety of reasons, but we can always improve them.

PR #44474, opened 2026-07-07
Add C# mgmt emitters for solutions

live1206:
## Summary

Adds `@azure-typespec/http-client-csharp-mgmt` configuration only for the configs that hit .NET SDK generation errors during validation:

- `specification/databricks/resource-manager/Microsoft.Databricks/Databricks/tspconfig.yaml`
- `specification/solutions/Solutions.Management/tspconfig.yaml`

The remaining confirmed namespace configs were split to #44486.

## Notes

Keeping this PR as draft while the generation errors are investigated.

PR #44486, opened 2026-07-07
Add C# mgmt emitters for validated namespace configs

live1206:
## Summary

Adds `@azure-typespec/http-client-csharp-mgmt` configuration for confirmed .NET namespaces that either hit SDK build/customization issues or did not show generation errors in SDK validation.

This is split out from #44474 so the generation-error configs can be tracked separately.

## Validation

- Parsed all modified `tspconfig.yaml` files with the repo YAML parser before the split
- `npm exec prettier -- --check <modified tspconfig files>` before the split


live1206:
Split from #44474. This PR contains the 22 confirmed namespace configs that did not have the Databricks/Solutions generation errors.

pshao25 (approved):

live1206:
This PR is adding missing csharp emitter configs, doesn't need ARM review.

PR #44490, opened 2026-07-07
ImageBuilder: C# client.tsp customizations for .NET SDK generation

cxznmhdcxz:
## Summary
Adds C#-scoped client customizations for the **Azure Image Builder** (`Microsoft.VirtualMachineImages`) TypeSpec project so the .NET management SDK generates and builds cleanly.

## Changes (`client.tsp`)
- `@@alternateType(Azure.ResourceManager.CommonTypes.Resource.id, Azure.Core.armResourceIdentifier, "csharp")` — maps the common-types `string` resource `id` to `ResourceIdentifier` so generated `*Data` constructors satisfy the `ResourceData` base (fixes CS1503).
- `@@clientName(RunOutputCollection, "RunOutputListResult", "csharp")` — avoids AZC0030 (model name must not end with `Collection`).
- `@@clientName(TriggerCollection, "TriggerListResult", "csharp")` — same as above.

All decorators are `csharp`-scoped and do not affect other languages.

## Related
Paired with the azure-sdk-for-net onboarding PR for `Azure.ResourceManager.ImageBuilder`.

### DIFF
--- a/specification/imagebuilder/resource-manager/Microsoft.VirtualMachineImages/ImageBuilder/client.tsp
+++ b/specification/imagebuilder/resource-manager/Microsoft.VirtualMachineImages/ImageBuilder/client.tsp
@@ -23,4 +23,32 @@ using Microsoft.VirtualMachineImages;
 
 // C#: model names must not end with 'Collection' (AZC0030).
 @@clientName(RunOutputCollection, "RunOutputListResult", "csharp");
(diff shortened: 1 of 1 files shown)


live1206 (approved):

live1206:
The change is only for csharp client naming, no need of ARM review

PR #44500, opened 2026-07-07
Storage STG 103/104 Typespec

seanmcc-msft:
# Choose a PR Template


specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -635,6 +635,44 @@ interface Container {
...
   >;
 
+  /** Returns a list of the blobs in Apache Arrow format as raw data, to be deserialized by the client. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Existing API"
+  @added(Versions.v2026_12_06)
+  @get
+  @sharedRoute
+  @route("?restype=container&comp=list")
+  listBlobFlatSegmentApacheArrow(
```
catalinaperalta: I believe this operation should be using the `StorageOperationResponseBody` template, see the operation above as an example

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -635,6 +635,44 @@ interface Container {
     }
   >;
 
+  /** Returns a list of the blobs in Apache Arrow format as raw data, to be deserialized by the client. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "Existing API"
```
catalinaperalta: These are arguably not existing apis :P we should have some valid justification for all new suppressions, for example: "Keeping operation shape inline with existing service patterns"

specification/storage/data-plane/BlobStorage/tspconfig.yaml
```diff
@@ -41,7 +41,7 @@ options:
     crate-version: "0.1.0"
     flavor: azure
     temp-omit-doc-links: true
-    api-version: "2026-04-06"
+    api-version: "2026-10-06"
```
catalinaperalta, line 44: just double checking, is this the version you want? because the 2026-12 version is newer

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -740,6 +820,12 @@ interface Blob {
         ...ContentLanguageResponseHeader;
         ...BlobSequenceNumberResponseHeader;
         ...BlobTypeResponseHeader;
+
+        /** The CRC64 hash of the content. */
+        @added(Versions.v2026_10_06)
+        @header("x-ms-content-crc64")
+        contentCrc64?: bytes;
```
catalinaperalta, line 808: trying to see if there's a way for you to leverage the existing aliases for the headers you're simply adding to new operations

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: this comment applies to other suppressions in this pr that need the justification updated

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: there are other operations not using a template in this pr, so same feedback there

nickliu-msft: @catalinaperalta For the newly introduced APIs (listBlobFlatSegmentApacheArrow & listBlobHierarchySegmentApacheArrow) we are intentionally NOT deserializing the response (and intend to return the raw response) as we are delegating the deserialization to the client to handwrite (since the client must read the ContentType response header to determine whether to deserialize using Apache Arrow or use XML). So I dont think we need `StorageOperationResponseBody` here.

If there is a better way to do this let me know.

specification/storage/data-plane/BlobStorage/Common/routes.tsp
nickliu-msft: Agreed, will change

specification/storage/data-plane/BlobStorage/Common/routes.tsp
nickliu-msft: @catalinaperalta Inline is only to attach `@added(Versions.v2026_10_06)`. You can't attach `@added` decorator to the spread property itself, and I didn't want to add it to the shared ContentCrc64ResponseHeader alias since that's spread unconditionally by other operations and would affect them. Correct me if I'm missing a cleaner option.

specification/storage/data-plane/BlobStorage/tspconfig.yaml
nickliu-msft: @vincenttran-msft I recall that you wanted me to bump this for Rust generation purposes. Do you want me to bump this to the latest version (2026-12-06)? That would just be the addition of Photon, which not sure if you all want that change in Rust yet.

vincenttran-msft: I believe `2026-10-06` is the correct version given the new plan. This should be fine for now and I can do another bump when authoring Photon for Rust (since Photon will look a little different in Rust anyways)

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: Yes, I've been looking for a better way for you to do this. We dont need to block the pr on this, but I do want to find a solution so you dont have an explosion of inline param definitions which is more error prone if you ever have to edit all of the same type of header in multiple places

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: That's a client level decision not an api level decision, the operation definition is the same. You should use the template so you dont have to implement the common service params like client request id, version header, etc inline. You have many templates available to you defined in this spec, simply choose the right one. Please define the typespec for correctness with the rest api and then add customizations through the tcgc decorators or patching mechanisms in the libraries. In this case you dont seem to even need to customize anything in relation to tcgc decorators you just need to clean up this implementation and use the templates

nickliu-msft: @catalinaperalta I pushed a commit that utilizes an existing template and still forces the response to be raw bytes. When I ran tsp compile, the resulting swagger changes were functionally the same as before this refactor, which is a good sign.

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -661,6 +694,42 @@ interface Container {
...
+      ...TimeoutParameter;
+      ...ListBlobsStartFrom;
+      ...ListBlobsEndBefore;
+    },
+    {
+      /** The content type of the body of the response. */
+      #suppress "@typespec/http/content-type-ignored" "Raw response, deserialized by client"
+      #suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "Fixed set of content types returned by the service"
+      @header("Content-Type")
+      contentType: "application/vnd.apache.arrow.stream" | "application/xml";
```
catalinaperalta: These values should be placed in the param to the template, if you look at the template it accepts overriding the default value for the content-type header

catalinaperalta: same for the other similar operation

catalinaperalta: no, the template already provides a ReponseMediaType param, which populates the value for the content-type header.  Here is the template def for reference: https://github.com/Azure/azure-rest-api-specs/blob/6e686456f8eae24e58137eaf9fe82f4a52c520dd/specification/storage/data-plane/BlobStorage/Common/routes.tsp#L69-L74

nickliu-msft: Gotcha, please ignore my previous reply (which I deleted soon afterwards). I misread what you mean and realized that immediately after sending it.

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: I'll resolve this comment for now but will reach out separately when I have an example working on what you can do to not have an explosion of repeated inline defs

specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -635,6 +635,35 @@ interface Container {
...
+  @added(Versions.v2026_12_06)
+  @get
+  @sharedRoute
+  @route("?restype=container&comp=list")
+  listBlobFlatSegmentApacheArrow is StorageOperationResponseBody<
+    {
+      /** The Accept header indicating the response should be returned as an Apache Arrow stream, falling back to XML otherwise. */
+      #suppress "@typespec/http/content-type-ignored" "Raw response, deserialized by client"
+      @header("accept")
+      accept: "application/vnd.apache.arrow.stream,application/xml";
```
catalinaperalta, line 651: Nit should this accept have an `|` between the values?

```suggestion
      accept: "application/vnd.apache.arrow.stream" | "application/xml";
```

catalinaperalta: same for the other operation

catalinaperalta (commented):
Ok the tsp itself is looking much cleaner, just a few more questions:

1. Are the new apache arrow operations actually just overrides of the existing `listBlobFlatSegment` and `listBlobHierarchySegment` operations? What I mean by this, should the new ListBlobsEndBefore param be added into the original operations since it's optional? and then the body param overrides with the alternateType decorator? <- to be 100% clear I'm not prescribing this as the right solution, since I see you also added an extra accept header. I'm more so wondering about the implementation and whether this is what you want to represent for the rest api, do you prefer this way of defining the operation?
2. Did you confirm these names with rust? I saw the the other operations for this route have simpler names and therefore it would be important to start using the ideal name for these operations from the start so you avoid a lot of churn on the language generation side.
3. The SwaggerApiDocPreview check is failing this might impact docs, do you know if your docs are auto-generated or is it ok to merge this pr regardless? It's not a required check so we can just merge the PR as is. 

I dont see any of the above questions as blocking, just want to make sure your tsp is doing what you want it to do before we merge :)

specification/storage/data-plane/BlobStorage/Common/routes.tsp
catalinaperalta: nevermind, this should be fine as is

catalinaperalta (approved):

specification/storage/data-plane/BlobStorage/Common/routes.tsp
nickliu-msft: Yeah the Accept header that the service accepts must be exactly `application/vnd.apache.arrow.stream,application/xml`

nickliu-msft:
> Ok the tsp itself is looking much cleaner, just a few more questions:
> 
> 1. Are the new apache arrow operations actually just overrides of the existing `listBlobFlatSegment` and `listBlobHierarchySegment` operations? What I mean by this, should the new ListBlobsEndBefore param be added into the original operations since it's optional? and then the body param overrides with the alternateType decorator? <- to be 100% clear I'm not prescribing this as the right solution, since I see you also added an extra accept header. I'm more so wondering about the implementation and whether this is what you want to represent for the rest api, do you prefer this way of defining the operation?
> 2. Did you confirm these names with rust? I saw the the other operations for this route have simpler names and therefore it would be important to start using the ideal name for these operations from the start so you avoid a lot of churn on the language generation side.
> 3. The SwaggerApiDocPreview check is failing this might impact docs, do you know if your docs are auto-generated or is it ok to merge this pr regardless? It's not a required check so we can just merge the PR as is.
> 
> I dont see any of the above questions as blocking, just want to make sure your tsp is doing what you want it to do before we merge :)

1. The new apache arrow operations is not an exact override of the old/existing operations. As you mentioned, it is intentional that ListBlobsEndBefore param is only supported for the new API. I think maybe sometime in the future, the non-apache arrow API will also provide support for ListBlobsEndBefore, but at the time being that is not the case.
2. I had @vincenttran-msft review the original Photon PR which he approved. Vincent if you change your mind, I'm happy to apply changes.
3. We handwrite pretty much all of our docs so I think this is fine.

PR #44557, opened 2026-07-09
Add Planned promoteOption for Mongo cluster replica promote (2026-06-15-preview)

Achint-Agrawal:

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Summary

Adds api-version `2026-06-15-preview` for `Microsoft.DocumentDB/mongoClusters`. This
version is additive-only: it extends the existing `promote` operation's
`PromoteOption` enum with a new `Planned` value (alongside the existing `Forced`
value). No other properties or operations change relative to `2026-06-01`.

- `Forced` (existing, default): promotes immediately; may lose very recent
  in-flight writes.
- `Planned` (new): waits for the replica to catch up before promoting, guaranteeing
  no data loss during the promotion.

The backing service behavior for `Planned` promote is already implemented and fully
deployed in production; this PR only publishes the already-shipped field in the
public spec.

## Due diligence checklist


javierfigueroa (approved):

specification/mongocluster/resource-manager/Microsoft.DocumentDB/MongoCluster/preview/2026-06-15-preview/examples/MongoClusters_PlannedPromoteReplica.json
```diff
@@ -0,0 +1,22 @@
+{
+  "operationId": "MongoClusters_Promote",
+  "title": "Promotes a replica Mongo Cluster resource to a primary role, waiting for the replica to catch up before promoting.",
+  "parameters": {
+    "subscriptionId": "ffffffff-ffff-ffff-ffff-ffffffffffff",
+    "resourceGroupName": "TestGroup",
+    "mongoClusterName": "myMongoCluster",
+    "api-version": "2026-06-15-preview",
+    "body": {
```
olivertowers, line 9: We may have a pre-existing issue with the schema for promote body - the spec defines the properties as on the top of the request body, but I think the actual implementation in control plane is to expect a wrapping `properties` .

If this is the case we may want to use this time to fix the approach (whilst avoiding spec breaking change).
1. We add support in the service to accept a promote request where properties are on the top OR inside of a `properties` wrapper. This way either approach is supported in the API.
2. Consider whether we add the dual schema support in the spec itself or just go with the existing spec state - I think the latter is probably fine as long as we fix the service side support in (1).

olivertowers (approved):

PR #44562, opened 2026-07-09
[network] Add models-ref for Network Common legacy swagger references

melina5656:


pshao25 (changes requested):
Let's not rush to solution

PR #44565, opened 2026-07-09
fix: isolation header for Foundry data-plane

baywet:
fixes merge collisions by replacing x-ms-user-identity by x-agent-session-id headers

Open questions:
- POST /agents/{agent_name}/endpoint/protocols/invocations has x-agent-invocation-id and is the only one with that, expected?


baywet:
after review, the headers didn't need to be edited

PR #44569, opened 2026-07-09
CognitiveServices: add capabilitySettings to account and project

JKrame:
Adds `capabilitySettings` property to the CognitiveServices account and project models in the 2026-07-15-preview TypeSpec specification. Targets the API version added in #44334.

### DIFF
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-07-15-preview/CreateAccount.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-07-15-preview/CreateAccount.json
@@ -20,7 +20,12 @@
           {
             "resourceId": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/myResourceGroup/providers/Microsoft.Storage/storageAccounts/myStorageAccount"
           }
-        ]
+        ],
+        "capabilitySettings": {
+          "documentStore": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/myResourceGroup/providers/Microsoft.DocumentDB/databaseAccounts/myCosmosAccount",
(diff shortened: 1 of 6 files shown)


fmabroukmsft:
Two points comparing this against the service (Platform-ResourceProvider CogSvc RP) behavior:

**1. Project `capabilitySettings` is marked create-only, but the service accepts and persists it on update.**
`ProjectProperties.capabilitySettings` uses `@visibility(Lifecycle.Read, Lifecycle.Create)` →
`x-ms-mutability: [read, create]`, and the description says "Settable only at create time." The service
does not enforce that: the project update/patch handlers (`ProjectsHandler.OnUpdateResourceAsync` /
`OnPatchResourceAsync`) accept `capabilitySettings` and persist it — the update path preserves only
`amlWorkspace`/`internalId`/`isDefault`/`agentIdentity` from the existing resource and then writes the
incoming properties, so a submitted `capabilitySettings` is stored. Existing unit tests confirm update
and patch with `capabilitySettings` are accepted (they only assert the feature-gate, not create-only).

Also note the asymmetry: `AccountProperties.capabilitySettings` has no mutability restriction (read/create/update)
while the project side is create-only, whereas the service treats both identically.

Please confirm intent:
- If update is meant to be supported, drop the create-only visibility on the project (match the account), or
- If create-only is intended, the service should reject/ignore `capabilitySettings` changes on update rather
  than silently persisting them.

**2. The `CreateProject` example response shows an inherited `blobStore` that the request did not send.**
In `examples/.../CreateProject.json`, the request sets only `documentStore` + `vectorStore`, but all response
bodies also include `blobStore` (pointing at the account's storage account). This matches the description's
"omitted fields inherit from the parent account," but the CogSvc RP does not merge account settings into the
project — it stores and returns `capabilitySettings` verbatim. As written, the example response does not
reflect what the API returns.

Please either make the response match the request (no `blobStore`), or confirm which layer materializes the
inherited value into the project response and add context for it.

fmabroukmsft (approved):

PR #44574, opened 2026-07-09
fix: namespace reset only triggers for changed+approved languages

### LINKED ISSUE #44573: Namespace approval: unchanged languages wrongly reported as "reset" (only the changed language's config should count)
## Summary

The namespace-approval workflow (added in #44085) posts a
`> ⚠️ **Namespace changed** — approvals for ... have been reset.`
warning that lists **all** languages, even when only **one** language's package name actually changed in `tspconfig.yaml` (and even when no approvals ever existed for the other languages).

This was observed in #44553, where only the `@azure-tools/typespec-ts` (TypeScript/JS) emitter section changed, yet the bot reported dotnet, python, java, and typescript as reset.

## Evidence (#44553)

The `tspconfig.yaml` diff changed **only** the TypeScript emitter options:

```diff
   "@azure-tools/typespec-ts":
-    emitter-output-dir: "{output-dir}/{service-dir}/ai-translation-document-rest"
+    emitter-output-dir: "{output-dir}/{service-dir}/ai-translation-document"
     generate-metadata: true
-    is-modular-library: false
     package-details:
-      name: "@azure-rest/ai-translation-document"
+      name: "@azure/ai-translation-document"
```

No dotnet / python / java emitter options were touched. Yet the bot posted:

> ⚠️ **Namespace changed** — approvals for dotnet, python, java, typescript have been reset.

(comment: https://github.com/Azure/azure-rest-api-specs/pull/44553#issuecomment-4927809330)

The PR timeline confirms this fired on the **first-ever** namespace detection for the PR (the workflow became active after the PR was originally opened, so the first run happened on a `synchronize` event). Nothing had ever been approved for any language at that point, so describing them as "reset" is doubly wrong.

## Root cause

In `.github/workflows/src/namespace-approval/post-results.js` (main @ `c178c557`), the `synchronize` reset-detection loop:

https://github.com/Azure/azure-rest-api-specs/blob/c178c557adf2223608411599924c78ded1af0efb/.github/workflows/src/namespace-approval/post-results.js#L251-L266

```js
const prev = previousTable.get(language);
const newNs = results.namespacesFound[language];

if (!prev || prev.namespace !== newNs) {
  const approvedLabel = `namespace-${language}-approved`;
  if (existingLabels.includes(approvedLabel)) {
    // ...actually removes the approval label (correctly guarded)
  }
  resetLanguages.push(language);   // <-- unconditional; not guarded
}
```

Two contributing issues:

1. **`namespacesFound` contains every language in the file, not just the changed one.** `detect-namespaces.js` re-parses all emitter options from the whole changed `tspconfig.yaml`, so the loop iterates over every language even though only one language's section changed in the diff. There is no per-language "did this section change in the diff" gate.

2. **`resetLanguages.push(language)` is unconditional.** It runs whenever `!prev` **or** `prev.namespace !== newNs`. When there is no previous bot comment (`prev === undefined`, e.g. first detection), *every* language satisfies `!prev` and is pushed into `resetLanguages`. The user-facing warning (built from `resetLanguages`) then claims approvals were "reset" for languages that (a) were never approved and (b) never changed.

The label removal itself is correctly guarded by `existingLabels.includes(approvedLabel)`, but the `resetLanguages` list that drives the message is not — so the message overstates what happened.

Warning is emitted here:

https://github.com/Azure/azure-rest-api-specs/blob/c178c557adf2223608411599924c78ded1af0efb/.github/workflows/src/namespace-approval/post-results.js#L205-L207

## Expected behavior

- The "approvals ... have been reset" warning should list **only** languages whose namespace **actually changed from a previously-recorded value** and that were previously approved (i.e. an approved label was genuinely removed).
- A first-time detection (`prev === undefined`) should treat languages as newly **pending**, not "reset".
- Changing one language's package name should not imply anything about the other languages' approval state.

## Suggested fix

Only add a language to `resetLanguages` when a real reset occurred, and treat "no previous entry" as new/pending rather than reset:

```js
if (prev && prev.namespace !== newNs) {
  const approvedLabel = `namespace-${language}-approved`;
  if (existingLabels.includes(approvedLabel)) {
    // remove approval label + splice existingLabels
    resetLanguages.push(language);
  }
} else if (prev && prev.status && !prev.status.includes("Pending")) {
  preservedApprovals.set(language, prev);
}
// languages with no `prev` fall through to normal pending handling below
```

## References

- Introduced by #44085
- Observed in #44553
- Code: `.github/workflows/src/namespace-approval/post-results.js` (reset loop L251-L266, warning L205-L207; detection L238)

samvaity:
## Fix

The namespace approval bot was incorrectly reporting **all** languages as 'reset' when:
1. The bot comment ran for the first time on a PR (no previous comment existed)
2. Only one language's namespace actually changed in tspconfig.yaml

### Root Cause

The reset condition was \if (!prev || prev.namespace !== newNs)\ which fired for **every** language on first detection (\!prev\ is always true when no previous bot comment exists).

Additionally, \
esetLanguages.push(language)\ was unconditional - it ran even when no approved label existed to remove.

### Fix

Reset now only fires when all three conditions are met:
1. A previous bot comment recorded the language (\prev\ exists)
2. The namespace actually changed (\prev.namespace !== newNs\)
3. An approved label was present and removed

First-time detections (\!prev\) are treated as new pending, not 'reset'.

### Tests

Updated and added test cases:
- First run (no previous comment) -> no resets (was incorrectly resetting all)
- New language added -> no reset (was incorrectly calling it a reset)
- Changed namespace without prior approval -> no reset warning
- Changed namespace with prior approval -> correctly resets only that language
- Multiple languages, one changed -> only the changed one is reset

All 57 namespace-approval tests pass.

Fixes #44573


.github/workflows/src/namespace-approval/detect-namespaces.js
```diff
@@ -181,6 +245,42 @@ export default async function detectNamespaces({ context, core }) {
...
+        baseConfig?.options
+      );
+      if (!baseOptions) {
+        continue;
+      }
+      const baseNamespaces = extractNamespacesFromOptions(baseOptions);
+      for (const [lang, ns] of Object.entries(namespacesFound)) {
+        if (baseNamespaces[lang] === ns) {
+          core.info(`Namespace unchanged for ${lang}: "${ns}", skipping`);
+          delete namespacesFound[lang];
```
praveenkuttappan, line 270: I am not a JS expert. But deleting a key from the object used to iterate is this an issue? One other option is to reorder this code. Get the base version namespaces first and check if namespace exists before adding it in namespacesFound.

praveenkuttappan (approved):

.github/workflows/src/namespace-approval/detect-namespaces.js
praveenkuttappan: This looks safe to do since you are doing Obj.entries.

praveenkuttappan (approved):

PR #44577, opened 2026-07-10
Align Compute NetworkApiVersion CSharp names

ArcturusZhang:
Align Compute NetworkApiVersion C# member names with the Azure.ResourceManager.Compute surface (2020_11_01 and 2022_11_01) so provisioning generation matches management naming.


ArcturusZhang:
Closing: NetworkApiVersion names should keep the uppercase V style, so this spec change is not needed.

PR #44630, opened 2026-07-14
Add SAP Monitor C# management emitter

live1206:
## Summary

Add the C# management emitter configuration for the SAP Monitor TypeSpec project.

The existing Azure SDK for .NET package already contains the SAP Monitor resources under `Azure.ResourceManager.Workloads`, so this emitter targets that established namespace and package directory.

## Validation

- `pnpm exec prettier --check specification/workloads/Workloads.SAPMonitor.Management/tspconfig.yaml`
- `pnpm exec tsp compile specification/workloads/Workloads.SAPMonitor.Management/main.tsp --no-emit`


live1206:
Closing for now: SAP Monitor still requires explicit .NET namespace approval, so we should not infer Azure.ResourceManager.Workloads from the existing package.

PR #42428, opened 2026-04-16
Adding ImmutabilityConfig, RecoveryPointImmutabilityProperties - API Version: 2026-05-01

gopikrishna1998:
# ARM (Control Plane) API Specification Update Pull Request 
 
 > [!TIP]
 > Overwhelmed by all this guidance? See the `Getting help` section at the bottom of this PR description.
 
 ## PR review workflow diagram
 
 Please understand this diagram before proceeding. It explains how to get your PR approved & merged.
 
 spec_pr_review_workflow_diagram
 
 ## Purpose of this PR
 
 What's the purpose of this PR? Check the specific option that applies. This is **mandatory**!
 
   - [ ] New resource provider.
   - [x] New API version for an existing resource provider.
   - [ ] Update existing version for a new feature.
   - [ ] Update existing version to fix OpenAPI spec quality issues in S360.
   - [ ] Convert existing OpenAPI spec to TypeSpec spec.
   - [ ] Other, please clarify:
 
 ### Summary of changes
 
 This PR adds API version **2026-05-01** for two services under `Microsoft.RecoveryServices`:
 
 **RecoveryServices (Vault)**
 - Added `ImmutabilityConfiguration` model with `type` (AsPerPolicy/TimeBased) and `durationInDays` properties
 - Added `configuration` property to `ImmutabilitySettings` referencing `ImmutabilityConfiguration`
 - Added `ImmutabilityType` enum (AsPerPolicy, TimeBased)
 
 **RecoveryServicesBackup (Backup)**
 - Added `RecoveryPointImmutabilityProperties` model with `isImmutable` (boolean) and `expiryTime` (DateTime) properties
 - Added `immutabilityProperties` property to `RecoveryPointProperties` referencing `RecoveryPointImmutabilityProperties`
 
 ## Due diligence checklist
 
 To merge this PR, you **must** go through the following checklist and confirm you understood 
 and followed the instructions by checking all the boxes:
 
 - [x] I confirm this PR is modifying Azure Resource Manager (ARM) related specifications, and not data plane related specifications.
 - [x] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including [ARM resource provider contract](https://aka.ms/azurerpc) and [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md) (estimated time: 4 hours).  
   I understand this is required before I can proceed to the diagram Step 2, "ARM API changes review", for this PR.
[...]

PR #44639, opened 2026-07-14
RecoveryServicesBackup: add 2026-08-01 stable version with fetchInstantItemRecoveryOperationResult action (MSRC-114273)

hiaga:
## Summary

Adds a new **stable API version `2026-08-01`** to `Microsoft.RecoveryServices/RecoveryServicesBackup` and introduces a dedicated, RBAC-gated action to retrieve Instant Item Recovery (ILR) mount scripts:

- **New operation** `ItemLevelRecoveryConnections_ListMountScripts`
  `POST .../recoveryPoints/{recoveryPointId}/listMountScripts` → returns `InstantItemRecoveryTarget` (the ILR `clientScripts`).
- Purpose: the ILR mount scripts contain iSCSI CHAP connection details. They were previously returned **inline** on the long-running `operationsStatus` (ILR provision) response, which is a broad operation-status read. This new action moves script retrieval to a dedicated, permission-scoped endpoint.

## Change classification

- **Additive / non-breaking.** A new API version and a new action are added.
- `InstantItemRecoveryTarget.clientScripts` remains an **optional** property in the schema. The service will simply stop populating it on the `operationsStatus` response in the new version; the field is **not removed** (removal would be breaking). This keeps the change non-breaking while still meeting the security requirement.
- No prior GA'd version swagger is modified.

## TypeSpec

`recoveryservicesbackup` is TypeSpec-authored; `bms.json` is generated. Source edits:
- `main.tsp` — added `v2026_08_01` to the `Versions` enum.
- `RecoveryPointResource.tsp` — added the `listMountScripts` action gated with `@added(Versions.v2026_08_01)`.

Generated `stable/2026-08-01/bms.json` + examples via `tsp compile`.

## ARM modeling notes

- Modeled as a **synchronous ARM POST action** (`ArmResourceActionSync`) returning the resource payload in the `200` body — consistent with how the sibling `provision`/`revoke` ILR actions are modeled on the same `RecoveryPointResource`, and with ARM guidance for POST actions that return data.
- Reuses the existing `InstantItemRecoveryTarget` model as the response, matching the shape already returned via the ILR provision `operationsStatus` extended info, so no new response contract is introduced.

> Draft PR — opened to begin swagger review in parallel with the service-side change. Not for merge until the service change is validated in canary.


hiaga:
The two failing Swagger Avocado checks (`MULTIPLE_DEFAULT_TAGS` and `MISSING_APIS_IN_DEFAULT_TAG`) are pre-existing characteristics of the RecoveryServicesBackup readme, not regressions introduced by this PR:

- **MULTIPLE_DEFAULT_TAGS**: the readme's `$(package-passivestamp)` / `$(package-activestamp)` blocks sit under the *Basic Information* heading, so Avocado counts each as a default tag. This condition already exists on `main` and only surfaces here because advancing the default tag changes Avocado's error hash (its baseline-diff then treats it as net-new).
- **MISSING_APIS_IN_DEFAULT_TAG**: the 10 flagged paths are legacy APIs last defined in `2016-06-01` / `2023-01-15` and already absent from the current default (`2026-05-01`). They are absent from the new `2026-08-01` version too, so **no API is dropped** by this PR. The new `2026-08-01` stable is a clean superset of `2026-05-01` (only the new `fetchInstantItemRecoveryOperationResult` action is added).

Advancing the default tag to `2026-08-01` is required so SDK/codegen picks up the new version. Could a reviewer please apply `Approved-Avocado` + `ARMSignedOff` to clear these, consistent with prior RecoveryServicesBackup precedent (e.g. #41997, #42428)? Thanks!

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`f7995d5`](https://github.com/Azure/azure-rest-api-specs/pull/44639/commits/f7995d5d148447ec9146b515eff32788513c5f93). See inline comments for findings 1, 4, 5, 6, 9, 10 and 11. Findings 2, 3, 7 and 8 concern lines that are unchanged in this PR diff, or live in `stable/2026-08-01/bms.json` whose diff GitHub omits due to file size, and are posted as top-level PR comments rather than inline.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
```diff
@@ -139,6 +140,21 @@ interface ItemLevelRecoveryConnections {
...
+
+  /**
+   * Fetches the mount scripts for an active Instant Item Recovery (ILR) session on the recovery point.
+   * The mount scripts contain the iSCSI connection details required to browse and recover files and folders.
+   * This dedicated, RBAC-gated action returns the scripts that were previously embedded inline in the
+   * operationsStatus (ILR provision) response, and must be used to retrieve them from API version 2026-08-01 onwards.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "Synchronous action that returns the mount scripts in the 200 response body."
+  @added(Versions.v2026_08_01)
+  @action("fetchInstantItemRecoveryOperationResult")
```
ravimeda: **[NEW] 🔴 Blocking** **[[RPC-POST-V1-10 / §12.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#121-use-post-actions-sparingly) / [§12.3](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#123-post-to-retrieve-data-exception) / [§7.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#71-no-secrets-in-getputpatch-responses)]** `RecoveryPointResource.tsp` - line 152 (emitted: `stable/2026-08-01/bms.json` line 2033 path, line 2035 `operationId`) - secret-returning POST action is not `list*`-prefixed.

**Issue:** The action exists specifically to return credential-bearing data - the ILR mount scripts contain iSCSI CHAP connection details, per the PR description and MSRC-114273. ARM guidance for secret-returning POST actions:

- §7.1: _"If the secret needs to be retrievable, expose it via a **POST action** (e.g. `listKeys`) for granular RBAC control."_ / _"POST actions named `list*` (e.g. `listKeys`, `listConnectionStrings`) are automatically exposed to ARM deployment templates via the `list()` template function."_
- §12.3: _"**Retrieving a secret value** -- use a `list*` prefixed POST action for ARM template compatibility."_
- §12.1 (RPC-POST-V1-10): _"Prefix the action name with `list` to enable ARM template deployment to call the POST action via the `list()` template function."_

As authored the action is not callable via the ARM template `list()` function. The PR description itself proposes `listMountScripts` at `.../listMountScripts` - the implemented code diverges from its own description. §12.1 also requires action names to be verb phrases; `listMountScripts` satisfies that, so the rename does not trade one naming rule for another.

_Design judgement (not rule-anchored):_ the segment `...OperationResult` also reads as an ARM async operation-result endpoint, which this is not - ARM operation endpoints are `/operationResults/{operationId}` and `/operationsStatus/{operationId}`.

**Classification reasoning:** Introduced in this PR - this action does not exist in the previous version `stable/2026-05-01` at base SHA `6a330c8e`.

**Suggested fix** (excerpt - the `#suppress` on line 150 is deliberately omitted here; see finding 6 for its disposition):

```tsp
@added(Versions.v2026_08_01)
listMountScripts is ArmResourceActionSync<
  RecoveryPointResource,
  InstantItemRecoveryOperationResultRequestResource,
  ArmResponse<InstantItemRecoveryTarget>
>;
```

This yields path segment `listMountScripts` and `operationId` `ItemLevelRecoveryConnections_ListMountScripts`. The explicit `@action(...)` decorator becomes unnecessary once the operation name matches the segment. Rename the example file and update its `operationId` and `title`. Please rename the request model in the same pass - if this rename lands but finding 4 does not, the SDK surface becomes `ListMountScripts(InstantItemRecoveryOperationResultRequestResource)`, an operation and a request type with unrelated names.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -3097,6 +3114,32 @@ model ILRRequest {
   objectType: string;
 }
 
+/**
+ * Request resource wrapping the parameters for the fetchInstantItemRecoveryOperationResult action.
+ */
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "Mirrors the sibling ILRRequestResource ARM request envelope."
+#suppress "@azure-tools/typespec-azure-core/casing-style" "Mirrors the sibling ILRRequestResource ARM request envelope."
+@added(Versions.v2026_08_01)
+model InstantItemRecoveryOperationResultRequestResource extends Resource {
```
ravimeda: **[NEW] 🔵 Suggestion** _(design feedback - not anchored to a named rule)_ `models.tsp` - line 3123 (emitted: `stable/2026-08-01/bms.json` line 10597) - ARM resource envelope used as a POST action request body.

**Issue:** The new POST action request body `InstantItemRecoveryOperationResultRequestResource extends Resource` plus `...ResourceProperties` produces a request schema that `allOf`s the ARM common-types v3 `Resource` (contributing readOnly `id`, `name`, `type`) and adds `location`, `tags` and `eTag`. None of these are meaningful inputs to this action; the only real parameter is `properties.operationId`. This is a legacy envelope carried over from `ILRRequestResource`, and reproducing it in a brand-new model in a brand-new GA version bakes it into the SDK surface - customers get a request type where most properties are inert.

I am not citing a named rule here; no rule in the ARM or TypeSpec review instructions covers this shape, so please treat it as design feedback rather than a citable violation.

**Classification reasoning:** Introduced in this PR - this model does not exist in the previous version `stable/2026-05-01` at base SHA `6a330c8e`.

**Suggested fix:** Model the action body as a plain request model and reference it directly from the action:

```tsp
@added(Versions.v2026_08_01)
model InstantItemRecoveryOperationResultRequest {
  /**
   * Operation ID returned by the prior provisionInstantItemRecovery action whose iSCSI mount scripts are to be retrieved.
   */
  operationId: string;
}
```

Dropping the envelope also removes the need for both `#suppress` directives in finding 5.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -3097,6 +3114,32 @@ model ILRRequest {
   objectType: string;
 }
 
+/**
+ * Request resource wrapping the parameters for the fetchInstantItemRecoveryOperationResult action.
+ */
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "Mirrors the sibling ILRRequestResource ARM request envelope."
+#suppress "@azure-tools/typespec-azure-core/casing-style" "Mirrors the sibling ILRRequestResource ARM request envelope."
```
ravimeda: **[NEW] 🟠 Warning** **[[TSP §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `models.tsp` - lines 3120-3121 - suppression justifications reference a sibling model instead of giving a technical reason.

**Issue:** Two new suppressions are added with the identical justification _"Mirrors the sibling ILRRequestResource ARM request envelope."_ §4.1 states: _"Suppression reasons **MUST NOT** justify by reference to another resource pattern... Each suppression must stand on its own with a technical justification for **why** the rule does not apply to this specific case."_ Neither reason explains why the rule does not apply.

The `@azure-tools/typespec-azure-core/casing-style` suppression in particular appears copy-pasted from a sibling that needed it for a reason that does not exist here. That rule PascalCase check rejects consecutive-uppercase acronym runs - which is why `ILRRequest` (line 3109) and `ILRRequestResource` (line 3095) carry it at lines 3107 and 3094. `InstantItemRecoveryOperationResultRequestResource` has no acronym run, and its own members are clean: the spread `...ResourceProperties` (`location`, `tags`, `eTag`) and `properties` (line 3129).

Direct in-file proof: `RestoreRequestResource` (line 2534) is structurally identical - `extends Resource`, `...ResourceProperties`, `properties?:` - has no acronym run, and carries **only** the `composition-over-inheritance` suppression at line 2533, with no `casing-style`. That model compiles today without one.

**Classification reasoning:** Introduced in this PR - both suppressions are added by this diff and are not present in the base branch at `6a330c8e`.

**Suggested fix:** Remove the `casing-style` suppression. For `composition-over-inheritance`, either restructure per finding 4 so the suppression becomes unnecessary, or replace the justification with a technical statement of why inheritance is genuinely required here.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
```diff
@@ -139,6 +140,21 @@ interface ItemLevelRecoveryConnections {
...
     AcceptedResponse
   >;
+
+  /**
+   * Fetches the mount scripts for an active Instant Item Recovery (ILR) session on the recovery point.
+   * The mount scripts contain the iSCSI connection details required to browse and recover files and folders.
+   * This dedicated, RBAC-gated action returns the scripts that were previously embedded inline in the
+   * operationsStatus (ILR provision) response, and must be used to retrieve them from API version 2026-08-01 onwards.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "Synchronous action that returns the mount scripts in the 200 response body."
```
ravimeda: **[NEW] 🟠 Warning** **[[TSP §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `RecoveryPointResource.tsp` - line 150 - new suppression appears unnecessary.

**Issue:** `#suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes"` with the reason _"Synchronous action that returns the mount scripts in the 200 response body."_ appears unnecessary. The documented requirement of that rule for synchronous POST is _"200 and default, or 204 and default. They must have no other responses"_, and the emitted operation declares exactly `200` + `default` (`stable/2026-08-01/bms.json` lines 2095-2108) with no `x-ms-long-running-operation` marker. Unlike the `FIXME` suppressions on the sibling actions, this one is newly introduced by this PR. If the rule does not fire, the suppression should not be carried into a new GA version, where it will mask a genuine response-code regression later.

**Classification reasoning:** Introduced in this PR - this suppression is added by this diff and is not present in the base branch at `6a330c8e`.

**Suggested fix:** Remove the suppression and recompile. If it does fire, replace the reason with a specific technical explanation of which response-code combination the emitter produces and why the rule requirement does not apply.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/stable/2026-08-01/examples/AzureIaasVm/FetchInstantItemRecoveryOperationResult.json
```diff
@@ -0,0 +1,33 @@
...
+      }
+    }
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "clientScripts": [
+          {
+            "scriptContent": "@ECHO OFF\nREM iSCSI mount script for Instant Item Recovery",
+            "scriptExtension": ".ps1",
```
ravimeda: **[NEW] 🔵 Suggestion** _(example quality - not anchored to a named rule)_ `stable/2026-08-01/examples/AzureIaasVm/FetchInstantItemRecoveryOperationResult.json` - lines 22-23 (and the byte-identical source copy under `examples/2026-08-01/AzureIaasVm/`) - script content and declared extension contradict each other.

**Issue:** Line 22 sets `scriptContent` to Windows batch syntax (`@ECHO OFF`, `REM ...`) while line 23 declares `"scriptExtension": ".ps1"` (PowerShell). The two fields contradict each other. The payload is schema-valid so no linter catches it, but example payloads are published to Microsoft Learn and shipped as SDK samples.

**Classification reasoning:** Introduced in this PR - this example file is added by this diff.

**Suggested fix:** Make the content match the declared extension - either a PowerShell snippet with `.ps1`, or change the extension to `.cmd` or `.bat`.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -3097,6 +3114,32 @@ model ILRRequest {
...
+
+/**
+ * Parameters for the fetchInstantItemRecoveryOperationResult action.
+ */
+@added(Versions.v2026_08_01)
+model InstantItemRecoveryOperationResultRequest {
+  /**
+   * Operation ID returned by the prior provisionInstantItemRecovery action whose iSCSI mount scripts are to be retrieved.
+   */
+  operationId: string;
```
ravimeda: **[NEW] 🔵 Suggestion** **[[TSP §2.6](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#26-arm-post-actions)]** `models.tsp` - line 3140 - is the `operationId` request body necessary, and is its requiredness enforceable?

**Issue:** Design question about the request body. The action is scoped at `.../recoveryPoints/{recoveryPointId}/...` yet additionally requires `properties.operationId`, referring to the prior `provisionInstantItemRecovery` operation. If at most one ILR session can be active per recovery point, the path already identifies the session and the body is redundant - §2.6 states _"When a POST action does **not** require a request body, use `void` -- not `{}`"_ - and dropping the body would also eliminate both new models and both suppressions from finding 5.

Note also that `operationId` is required in `InstantItemRecoveryOperationResultRequest` (`stable/2026-08-01/bms.json` lines 10593-10595) but the wrapper `InstantItemRecoveryOperationResultRequestResource` (lines 10597-10626) has no `required` array and its `properties` member is optional - so a body of `{}` is schema-valid and the requiredness is unenforceable at the contract level.

**Classification reasoning:** Introduced in this PR - this property does not exist in the previous version `stable/2026-05-01` at base SHA `6a330c8e`.

**Suggested fix:** Either drop the request body (use `void`) if the recovery point uniquely identifies the ILR session, or document why `operationId` is required, mark `properties` required on the wrapper, and consider a less ambiguous name than `operationId` (which collides with both the ARM async operation ID concept and the swagger `operationId` field).

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
```diff
@@ -139,6 +140,21 @@ interface ItemLevelRecoveryConnections {
     void,
     AcceptedResponse
   >;
+
+  /**
+   * Fetches the mount scripts for an active Instant Item Recovery (ILR) session on the recovery point.
+   * The mount scripts contain the iSCSI connection details required to browse and recover files and folders.
+   * This dedicated, RBAC-gated action returns the scripts that were previously embedded inline in the
+   * operationsStatus (ILR provision) response, and must be used to retrieve them from API version 2026-08-01 onwards.
+   */
```
ravimeda, line 146: **[NEW] 🔵 Suggestion** _(design feedback - soft-anchored to [§7.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#71-no-secrets-in-getputpatch-responses) "for granular RBAC control")_ `RecoveryPointResource.tsp` - lines 144-149 - the RBAC-gating claim is not verifiable from the spec.

**Issue:** The operation doc describes the action as "RBAC-gated", but that guarantee is not verifiable from the specification, and path scope alone does not deliver it. The new action (`stable/2026-08-01/bms.json` line 2033) and the ILR operation-status endpoint (line 1808) diverge under `.../protectedItems/{protectedItemName}` - neither is an ancestor of the other. Because ARM role assignments are made at resource scopes (vault, resource group, subscription) rather than at proxy sub-paths such as `operationsStatus/{operationId}`, the same assignment that grants the operation-status read also covers the recovery point. Any narrowing must come from a distinct action permission string in the RP manifest, which is not visible in this repository.

**Classification reasoning:** Introduced in this PR - this doc comment is added by this diff.

**Suggested fix:** State the specific action permission string the new endpoint requires in the operation doc comment (for example `Microsoft.RecoveryServices/vaults/backupFabrics/protectionContainers/protectedItems/recoveryPoints/listMountScripts/action`), and confirm on this PR that the RP manifest grants it separately from the operation-status read permission. Without that, the security benefit claimed by the PR is not achieved.

ravimeda:
**[EXISTING] 🔴 Blocking** **[[§6.8](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#68-operation-results-placement) / [§7.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#71-no-secrets-in-getputpatch-responses) / [§10.3](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#103-do-not-remove-properties-between-api-versions-arg003)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/stable/2026-08-01/bms.json` - line 11783 (JSON path `$.definitions.OperationStatusProvisionILRExtendedInfo.properties.recoveryTarget`); TypeSpec source `models.tsp` - line 8668.

> Posted as a top-level comment because GitHub omits the diff patch for `bms.json` (file too large), so an inline anchor is not available.

**Issue:** The security objective of this PR is not represented in the contract. `OperationStatusProvisionILRExtendedInfo.recoveryTarget` still `$ref`s `InstantItemRecoveryTarget` (carrying `clientScripts` then `scriptContent`) in the new `2026-08-01` document. Reachability is literal: `OperationStatus.properties` (line 11703) is the polymorphic `OperationStatusExtendedInfo` base (line 11722), whose `OperationStatusProvisionILRExtendedInfo` subtype (line 11778) carries the scripts. §6.8 therefore applies word-for-word: _"Operation status `properties` **MUST NOT** contain sensitive information — operation status resources may be accessible through polling endpoints with different RBAC restrictions than the original operation."_ §7.1 adds: secrets _"**MUST NOT** be exposed in the response of GET, PUT, or PATCH operations."_

Blast radius: `#/definitions/OperationStatus` is `$ref`d by **seven** operations in this version (lines 1865, 3053, 3450, 4227, 4527, 4949, 5353), so the credential-bearing subtype is declared on all of them.

The PR body says the service will simply stop populating the field. There is a signal about that in the new operation description (line 2039), but none on the `recoveryTarget` property itself - its description at line 11784 is unchanged - so no client, reviewer or CI check can tell from the property that it is no longer populated.

**Classification reasoning:** This issue also exists in the previous version - `stable/2026-05-01/bms.json` line 11657, same JSON path, at base SHA `6a330c8e`. It is called out here because closing it is the stated purpose of this PR.

**Suggested fix (preferred - deprecate in place).** §10.3 (ARG003) states _"**DO NOT** remove properties from a resource between API versions until the property usage has been fully deprecated... Instead, keep the property in the response and mark it as deprecated in the swagger description and documentation,"_ and _"All types and properties from previous API versions **MUST** be carried forward to new API versions."_ Excerpt - leave the `#suppress` on `models.tsp` line 8662 and the `objectType` discriminator member (line 8673) intact:

```tsp
/**
 * Target details for file / folder restore. Deprecated: not populated from API version
 * 2026-08-01 onward. Use the recovery point mount-scripts action to retrieve the scripts.
 */
recoveryTarget?: InstantItemRecoveryTarget;
```

**Suggested fix (stronger - only with sign-off).** If the security team requires the property gone from the contract entirely, add `@removed(Versions.v2026_08_01)`. That is both an SDK-breaking change and an ARG003 exception, so it needs breaking-change board **and** ARG-compatibility sign-off recorded on this PR. Do not apply it unilaterally.

Either way, a silent behavior change with no signal on the property is not acceptable.

ravimeda:
**[EXISTING] 🔴 Blocking** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md) / [§7.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#72-x-ms-secret-annotation)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp` - line 8355 (emitted: `stable/2026-08-01/bms.json` line 8635, JSON path `$.definitions.ClientScriptForConnect.properties.scriptContent`).

> Posted as a top-level comment because this line is unchanged in the PR diff and cannot take an inline anchor.

**Issue:** `ClientScriptForConnect.scriptContent` carries the iSCSI mount script whose content includes CHAP connection credentials - the exact data MSRC-114273 concerns, and the new operation own doc comment says _"The mount scripts contain the iSCSI connection details required to browse and recover files and folders."_ The property has no `@secret` decorator and no `x-ms-secret: true` in the emitted swagger. §7.2: _"Any property that contains a secret in **any** operation (PUT, PATCH, GET, or POST) **MUST** be annotated with `x-ms-secret: true` in the swagger definition."_ Without it, ARM Template What-If treats the value as ordinary data (WHATIF-003) and secret-scanning tooling cannot identify the property.

Note the interlock: once `scriptContent` is annotated, the finding about `OperationStatusProvisionILRExtendedInfo.recoveryTarget` becomes machine-detectable - §7.1 then applies mechanically to all seven GETs returning `OperationStatus`.

**Classification reasoning:** This issue also exists in the previous version - `stable/2026-05-01/bms.json` line 8552, same JSON path, also without `x-ms-secret`, at base SHA `6a330c8e`. The base file contains zero `x-ms-secret` occurrences. It is called out here because this PR newly returns the property from a dedicated action.

**Suggested fix:**

```tsp
model ClientScriptForConnect {
  /**
   * File content of the client script for file / folder restore.
   */
  @secret
  scriptContent?: string;
  // ...
}
```

Trade-off to settle before applying: `ClientScriptForConnect` is not version-gated, so an unconditional `@secret` emits `x-ms-secret: true` into every previously published `bms.json`, changing already-released API version documents. Options: (a) apply it and get the regeneration of prior version files acknowledged as a metadata-only security correction; (b) if that is not acceptable, raise it with the ARM API review board and record the decision on this PR. Leaving the property unannotated is not an option.

ravimeda:
**[EXISTING] 🟠 Warning** **[[TSP §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp` - lines 107-108, 122-123 and 135-136 - FIXME placeholder suppression justifications carried into a GA version.

> Posted as a top-level comment because these lines are unchanged in the PR diff and cannot take an inline anchor.

**Issue:** The sibling `provisionInstantItemRecovery` and `revokeInstantItemRecovery` actions - directly adjacent to the new action in the same interface - carry suppressions justified only as _"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"_. §4.1 states: _"Suppression reasons **MUST NOT** contain placeholder text such as `FIXME`, `TODO`, or `TBD`. These indicate the author deferred the justification and never returned to it."_ These are being carried forward into a new GA version.

**Classification reasoning:** These issues also exist in the previous version - the same six suppressions are present in this file on the base branch at `6a330c8e`.

**Suggested fix:** Replace each `FIXME` placeholder with a real technical justification, or remove the suppression if the rule no longer fires. Since this PR is already editing the `ItemLevelRecoveryConnections` interface, it is a natural place to clear this debt.

ravimeda:
**[EXISTING] 🔵 Suggestion** **[[TSP §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp` - line 8371 (emitted: `stable/2026-08-01/bms.json` line 8647).

> Posted as a top-level comment because this line is unchanged in the PR diff and cannot take an inline anchor.

**Issue:** `ClientScriptForConnect.url` is described as _"URL of Executable from where to source the content. If this is not null then ScriptContent should not be used"_ - it is an **alternative delivery channel for the same credential-bearing payload** covered by the `scriptContent` finding, so it sits inside the MSRC-114273 blast radius rather than being a pure typing nit. Please evaluate it for `@secret` alongside `scriptContent`, and confirm whether it can carry a SAS token. Separately, per the TypeSpec anti-pattern list, _"properties that hold URLs or SAS URIs **MUST** use the TypeSpec `url` scalar type (which generates `format: uri` in swagger) rather than plain `string`"_ - this property is currently `string` with no `format`.

**Classification reasoning:** This issue also exists in the previous version - `stable/2026-05-01/bms.json` line 8564, `type: string`, no `format`, no `x-ms-secret`, at base SHA `6a330c8e`.

**Suggested fix:** Change to `url?: url;` and evaluate `@secret`. The version-immutability consideration from the `scriptContent` finding applies here too, since `ClientScriptForConnect` is not version-gated.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
hiaga: Agreed POST is right (secret-bearing response can't go over GET per §7.1); we're considering a \list*\-prefixed rename to enable the ARM \list()\ surface and will confirm on this thread once finalized.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
hiaga: Removed the suppression. The synchronous action emits exactly `200 + default` so the rule doesn't fire — verified with `tsp compile --warn-as-error` (clean) and the emitted `bms.json` is unchanged. Done in the latest push.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
hiaga: Dropped the ARM Resource envelope so the meaningless id/name/type/location/tags/eTag fields are gone (commit c582bdf). Kept the \{ properties: { operationId } }\ shape to align with service change.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
hiaga: Both suppressions removed — the flatten dropped xtends Resource, so composition-over-inheritance and casing-style no longer apply (commit 2d8d492).

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
hiaga: Softened — dropped the unverifiable \RBAC-gated\ wording; the access narrowing is enforced service-side via the action permission and isn't expressible in the spec (commit 5dace18).

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/stable/2026-08-01/examples/AzureIaasVm/FetchInstantItemRecoveryOperationResult.json
hiaga: Fixed — replaced the batch content with a PowerShell iSCSI mount snippet so it matches the declared \.ps1\ (both example copies) (commit 5dace18).

hiaga:
Agreed — rather than a silent change, we'll deprecate \
ecoveryTarget\ in place now (description note that it's no longer populated from 2026-08-01) and remove it in a later version through the breaking-change process.

hiaga:
Thanks — to clarify: this PR is additive. We're intentionally still populating \
ecoveryTarget\ on operationsStatus for back-compat during client migration, and will deprecate the property and stop populating it in a follow-up once all clients have moved to the new fetch action.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
hiaga: operationId is required — the scripts are keyed by the prior ProvisionILR operation (not the recovery point alone) and it anchors the cross-scope security check that binds the returned secret to the authorized recovery point; we've marked \properties\ required (commit daf7d45) so requiredness is now enforced at the contract level.

hiaga:
Added \@ secret\ to \scriptContent\ (commit daf7d45). As noted, the model isn't version-gated, so \x-ms-secret: true\ re-emits into the previously published version documents as well — this is a metadata-only security-detection correction (the value returned is unchanged), flagged here for acknowledgement.

hiaga:
These \FIXME\ suppressions are on the pre-existing \provision\/\
revoke\/\
restore\ actions (unchanged by this PR); we'd prefer to leave them as-is here and clear that debt in a separate targeted change rather than widen this PR's scope.

hiaga:
`url` is a legacy optional alternative to `scriptContent` that the ILR flow does not populate (scripts are always inlined), so it carries no secret in practice; since it's a pre-existing, non-version-gated property, we'll evaluate the `url` scalar typing and `@secret` separately rather than re-emit into all published versions from this PR.

hiaga:
Update: since this is a text-only change, we went ahead and replaced all six `FIXME` placeholder justifications (restore auxiliary header + provision/revoke ILR actions) with real technical justifications in this PR (commit 1c19f69). Comment-only change; emitted swagger is unchanged.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/RecoveryPointResource.tsp
hiaga: can we proceed with listInstantItemRecoveryOperationResultResponse ?

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
hiaga: we're considering making it a flat body as suggested, with operationId renamed as provisionInstantItemRecoveryOperationId

sandipsh:
swagger apidocreview, avocado, typespec validations are failing. please review, fix or comment why those can't be addressed.

hiaga:
Thanks @sandipsh — addressed:

- **TypeSpec Validation:** fixed in commit 1b5d55e. The failure was a single `--warn-as-error` diagnostic — TCGC rule `csharp-model-suffix` on the new `InstantItemRecoveryOperationResultRequest` body model (models.tsp:3168). Added `@@clientName(InstantItemRecoveryOperationResultRequest, "InstantItemRecoveryOperationResultContent", "csharp")` in `client.tsp`, consistent with the existing C# naming-fix block there. This keeps the wire/OpenAPI schema name unchanged and only maps the .NET client model to the compliant `...Content` name. `tsp compile --no-emit --warn-as-error` now passes locally.

- **Avocado (`MULTIPLE_DEFAULT_TAGS`):** this originates from the pre-existing `$(package-passivestamp)` / `$(package-activestamp)` default-tag blocks in `readme.md`, which this PR does not change structurally (we only bumped the default tag value `package-2026-07-01` → `package-2026-08-01` in-place, mirroring the prior version bump). It is not introduced by this PR; we'd prefer to leave the passive/active-stamp readme structure untouched here and address it separately if the tooling requires it.

- **ApiDocPreview:** this is downstream of the TypeSpec compile failure above; it should clear on the re-run now that the compile passes. Will confirm once CI completes.

sandipsh:
@hiaga please address doc preview error https://dev.azure.com/apidrop/Content%20CI/_build/results?buildId=624347

hiaga:
@sandipsh Fixed in b5e809a: ApiDocPreview now splits multi-service preview inputs by service; `Swagger ApiDocPreview` and `spec - api-doc-preview` are green.

hiaga:
> Himanshu Agarwal (@hiaga) please address doc preview error https://dev.azure.com/apidrop/Content%20CI/_build/results?buildId=624347

@sandipsh resolved the API doc preview issue by spliting the other folder verison bump into a separate PR - https://github.com/Azure/azure-rest-api-specs/pull/45921

sandipsh (approved):

PR #44654, opened 2026-07-14
Foundry agents: add entra_authorization (sign_in_mode + token_validation_rules) to Agent object

chaoyu-msft:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Update existing version for a new feature. (This is applicable only when you are revising a private preview API version.)
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

Adds an optional agent-level `entra_authorization` configuration to the Foundry Agents
data-plane spec, present on `AgentObject`, `CreateAgentRequest`, and `UpdateAgentRequest`.

- `AgentSignInMode` union: `default` | `blueprint` (governs how the agent endpoint
  authorizes incoming Entra tokens). PBI 5362167.
- `AgentEntraAuthorization.sign_in_mode?` — optional; treated as `default` when omitted.
- `AgentEntraAuthorization.token_validation_rules?` (`AgentTokenValidationRules`) with flat
  arrays `allowed_audiences`, `allowed_tenants`, `required_scopes`, used to validate inbound
  bearer tokens in `blueprint` sign-in mode. PBI 5362231.

No breaking changes: all new fields are optional additive properties.

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -123,6 +129,43 @@ union AgentState {
   disabled: "disabled",
 }
 
+@doc("The sign-in mode that governs how the agent's endpoint authorizes incoming Entra tokens.")
+union AgentSignInMode {
+  string,
+
+  @doc("Default Foundry sign-in mode. This is the default when a value is not specified.")
+  default: "default",
```
viswabalaji: is there a better name to reflect this? it can be the default, but does it need to be called "Default"?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -123,6 +129,43 @@ union AgentState {
...
+}
+
+@doc("""
+  Rules used to validate an inbound bearer token addressed to the agent's blueprint audience.
+  Each list is defaulted to a single value when the agent is switched to the 'blueprint' sign-in
+  mode (audience = blueprint resource URI, tenant = subscription tenant, required scope = default
+  scope declared on the blueprint).
+  """)
+model AgentTokenValidationRules {
+  @doc("The set of accepted token audiences (the `aud` claim). Typically the agent blueprint resource URI (e.g. `api://{agentBlueprintId}`).")
```
viswabalaji: wouldn't 1 agent have only 1 blueprint linked, so from perspective of blueprint sign in, how would there be multiple audiences?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -123,6 +129,43 @@ union AgentState {
...
+  blueprint: "blueprint",
+}
+
+@doc("Entra authorization configuration for an agent.")
+model AgentEntraAuthorization {
+  @doc("The sign-in mode for the agent. Defaults to 'default' when not specified.")
+  sign_in_mode?: AgentSignInMode;
+
+  @doc("The rules used to validate inbound bearer tokens presented to the agent endpoint when the agent is in the 'blueprint' sign-in mode.")
+  token_validation_rules?: AgentTokenValidationRules;
```
viswabalaji: ideally the type of the sign in model should bring some properties with it, instead it being disjoint.

so only the shape would be sign_in_config: BlueprintAppConfig | FoundryAppConfig

where 
BlueprintConfig would be 
{ mode: blueprint, token_validation_config }

and FoundryAppConfig is the current default ai.azure.com.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
chaoyu-msft: Thanks for this comment, it's a valid question.
I'll need a discussion with the team to confirm on this.

chaoyu-msft: separated the field its own token_validation_rules, it is used only for token validation which is not related to this blueprint sign-in mode.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -587,6 +587,23 @@ model EntraAuthorizationScheme extends AgentEndpointAuthorizationScheme {
...
 
+@doc("""
+  Validation applied to an inbound bearer token presented to the agent endpoint.
+  Applies across authorization schemes. Each list is defaulted to a single value when the agent's
+  `sign_in_mode` is 'blueprint' (audience = blueprint resource URI, tenant = subscription tenant,
+  required scope = default scope declared on the blueprint).
+  """)
+model TokenValidationConfiguration {
+  @doc("The set of accepted token audiences (the `aud` claim). Typically the agent blueprint resource URI (e.g. `api://{agentBlueprintId}`).")
+  allowed_audiences?: string[];
```
viswabalaji, line 640: I still wonder if these should be a config of {aud, tenant_id, required_scope}[] instead of each being their own list, because that assumes the every tenant caller will have same audience/scopes?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -651,6 +668,17 @@ model InvocationsProtocolConfiguration {}
...
 
+@doc("The sign-in mode governing which Entra application identity the agent uses to represent the calling user (for inbound token validation and downstream on-behalf-of access). Applies across inbound authorization schemes.")
+union AgentSignInMode {
+  string,
+
+  @doc("Foundry sign-in. The calling user is represented via the Foundry application (ai.azure.com). This is the default when a sign-in mode is not specified.")
+  foundry: "foundry",
+
+  @doc("Agent-blueprint sign-in. The calling user is represented via the agent's blueprint application.")
+  blueprint: "blueprint",
```
viswabalaji: should we call this agent_blueprint?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -662,8 +690,14 @@ model AgentEndpointConfig {
...
 
-  @doc("The authorization schemes supported by the agent endpoint")
+  @doc("The authorization schemes supported by the agent endpoint.")
   authorization_schemes?: AgentEndpointAuthorizationScheme[];
+
+  @doc("The sign-in mode governing the Entra application identity used to represent the calling user. Defaults to 'foundry' when not specified.")
+  sign_in_mode?: AgentSignInMode = AgentSignInMode.foundry;
+
+  @doc("The validation applied to inbound bearer tokens presented to the agent endpoint. Applies across the configured authorization schemes; typically defaulted when `sign_in_mode` is 'blueprint'.")
+  token_validation?: TokenValidationConfiguration;
```
viswabalaji: I had a similar comment to this before, on should token_validation be a top level field and not be nested inside sign_in_mode itself.

so sign_in_config {mode: enum, params}
so if users set foundry, type safety would ensure that you cannot pass token_validation, and only when you set blueprint, this field would unlock.

would help if we add more sign_in modes and they have different inputs

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -651,6 +668,45 @@ model InvocationsProtocolConfiguration {}
 @doc("Configuration specific to the WebSocket-based invocations protocol.")
 model InvocationsWsProtocolConfiguration {}
 
+@doc("The sign-in mode governing which Entra application identity the agent uses to represent the calling user (for inbound token validation and downstream on-behalf-of access). Applies across inbound authorization schemes.")
+union AgentSignInMode {
+  string,
```
dargilco: If this union is used in the request only, then "string" should be removed. We add "string" to unions that are used in response (or request and response).

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -651,6 +668,45 @@ model InvocationsProtocolConfiguration {}
...
+  The sign-in configuration governing which Entra application identity the agent uses to represent
+  the calling user, for inbound token validation and downstream on-behalf-of access. The `mode`
+  property discriminates the concrete configuration type; mode-specific inputs (such as token
+  validation) are carried only by the modes that support them. Defaults to Foundry sign-in when not
+  specified.
+  """)
+@discriminator("mode")
+model AgentSignInConfig {
+  @doc("The sign-in mode.")
+  mode: AgentSignInMode;
```
dargilco: In Foundry we use "type" as a discriminator.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
chaoyu-msft: This new field is part of the Agent Endpoint definition, so it will show up in some responses.
It is defined similarly to the field AgentEndpointAuthorizationSchemeType, I think we should keep the "string".

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
```diff
@@ -78307,6 +78501,27 @@
...
+          "required_scopes": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "The scopes (the `scp` claim) that must all be present on the inbound token. Typically the default scope declared on the agent blueprint."
+          }
+        },
+        "description": "A single accepted-token profile used to validate an inbound bearer token presented to the agent\nendpoint. A token is accepted when its `aud`, `tid`, and `scp` claims match this rule. When\nmultiple rules are configured, a token is accepted if it satisfies any one of them, which allows\nper-tenant audiences and scopes rather than assuming every caller shares the same values."
+      },
```
mattchenderson: Why are we coupling single audience and single tenant ID and putting the entire object in an array? This increases verbosity for multiple audience setups and generally makes default values much more difficult.

The model proposed previously follows what you typically see in JWT validation libraries, and it's what I'd expect to see here.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
```diff
@@ -26440,6 +26440,59 @@
           }
         ]
       },
+      "AgentBlueprintSignInConfig": {
+        "type": "object",
+        "required": [
+          "type"
+        ],
+        "properties": {
+          "type": {
```
mattchenderson: `type` feels weird to me, because this has a bunch of side effects and is a modal change for the agent's behavior.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
```diff
@@ -26778,7 +26839,15 @@
             "items": {
               "$ref": "#/components/schemas/AgentEndpointAuthorizationScheme"
             },
-            "description": "The authorization schemes supported by the agent endpoint"
+            "description": "The authorization schemes supported by the agent endpoint."
+          },
+          "sign_in_config": {
```
mattchenderson: I think we need to preserve the binding of this to Entra. This also now reads to me as if it's setting up a sign-in, rather than having a requirement imposed on the sign-in that occurred, if that makes sense.

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
chaoyu-msft: This was from Viswajeet's comment earlier, I'll revert it back to align the doc.
"I still wonder if these should be a config of {aud, tenant_id, required_scope}[] instead of each being their own list, because that assumes the every tenant caller will have same audience/scopes?"

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
chaoyu-msft: received comments from my manager on this too, "mode" is more preferred.
Please let me know if we need to stick with type.
"type feels weird to me, because this has a bunch of side effects and is a modal change for the agent's behavior"

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.yaml
```diff
@@ -41604,6 +41620,9 @@ components:
           type: string
           enum:
             - Entra
+        sign_in_config:
```
mattchenderson: Property is still not in context of Entra. Per Teams discussion, I think that's still necessary.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -589,6 +589,63 @@ model AgentEndpointAuthorizationScheme {
 
 model EntraAuthorizationScheme extends AgentEndpointAuthorizationScheme {
   type: AgentEndpointAuthorizationSchemeType.Entra;
+
+  @doc("The Entra sign-in configuration governing which Entra application the agent endpoint validates inbound tokens against (and represents the calling user as for downstream on-behalf-of access). Defaults to the platform default ('foundry') behavior when not specified.")
+  sign_in_config?: EntraSignInConfig;
```
viswabalaji: why should mode type be specific to Entra? we can add non entra modes in future, so safer to not include entra in this type name right?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -589,6 +589,63 @@ model AgentEndpointAuthorizationScheme {
...
+
+@doc("The Entra sign-in mode governing which Entra application the agent endpoint validates inbound tokens against.")
+union EntraSignInMode {
+  string,
+
+  @doc("Sign-in using the Foundry application (ai.azure.com). Inbound tokens are validated against the Foundry application audience. This is the default when a sign-in configuration is not specified.")
+  foundry: "foundry",
+
+  @doc("Agent-blueprint sign-in. The agent endpoint validates inbound tokens against the agent's blueprint application.")
+  agent_blueprint: "agent_blueprint",
```
viswabalaji, line 619: we should consider if we can allow multiple sign in options, rather than these being exclusive.
External direct agent invokers can sign in to blueprint and use, while there could be some users who are accessing with RBAC as well. Eventually we would want customers to tighten up the sign in policy to have one streamlined entry point

chaoyu-msft: I think we can do follow up changes later when future plan got confirmed.

ankitbko (commented):
What feature is this change going to enable? Did we do design and API review for it?

chaoyu-msft:
> What feature is this change going to enable? Did we do design and API review for it?

This is a change to support blueprint OBO token that allows user to sign in with their own token using blueprint to invoke the agent endpoint, including external users.
Matthew Henderson is the manager on this new feature, here is a basic spec design: https://github.com/coreai-microsoft/foundrysdk_specs/pull/249

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -591,6 +591,60 @@ model EntraAuthorizationScheme extends AgentEndpointAuthorizationScheme {
...
+
+@doc("""
+  Entra sign-in configuration for the agent endpoint. The `mode` property discriminates the concrete
+  configuration type; mode-specific inputs (such as token validation) are carried only by the modes
+  that support them. Defaults to Foundry sign-in when not specified.
+  """)
+@discriminator("mode")
+model EntraSignInConfig {
+  @doc("The Entra sign-in mode.")
+  mode: EntraSignInMode;
```
dargilco: We use 'type' as discriminator in Foundry. So I would also call the union EntraSignInType

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -666,8 +720,11 @@ model AgentEndpointConfig {
   @doc("Per-protocol configuration for the agent endpoint.")
   protocol_configuration?: ProtocolConfiguration;
 
-  @doc("The authorization schemes supported by the agent endpoint")
+  @doc("The authorization schemes supported by the agent endpoint.")
   authorization_schemes?: AgentEndpointAuthorizationScheme[];
+
+  @doc("The Entra sign-in configuration governing which Entra application the agent endpoint validates inbound tokens against (and represents the calling user as for downstream on-behalf-of access). Defaults to the platform default ('foundry') behavior when not specified.")
+  entra_sign_in_config?: EntraSignInConfig;
```
dargilco: Please update the doc string. Instead of `Defaults to the platform default ('foundry') behavior when not specified. ` write `The service defaults 'foundry' behavior if a value is not specified by the caller` (It's important to note that this is a service-side default, as Python emitter adds a text related to client-side default).

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -614,6 +589,71 @@ model AgentEndpointAuthorizationScheme {
 
 model EntraAuthorizationScheme extends AgentEndpointAuthorizationScheme {
   type: AgentEndpointAuthorizationSchemeType.Entra;
+
+  @doc("""
+    The Entra application audiences this scheme accepts inbound tokens for. When omitted, the scheme
```
dargilco: "The Entra application audiences for which this scheme accepts inbound tokens"

PR #44673, opened 2026-07-15
[Education] Exclude csharp flatten for LabProperties budget to avoid custom code

cxznmhdcxz:
## Summary
Adds `csharp`-scoped legacy `flattenProperty` exclusions for the Education `LabProperties` model so that `totalBudget` and `totalAllocatedBudget` are **not** flattened in the C# SDK.

## Motivation
The default flattening turned `EducationAmount totalBudget` / `totalAllocatedBudget` into `*Currency` (string) + `*Value` (float?) pairs on `LabDetailsData`. Keeping them flattened required an SDK-side `[CodeGenSuppress]` custom-code workaround. Excluding the `csharp` scope from flattening removes the need for that custom code and exposes the properties as `EducationAmount`.

## Changes
- `client.tsp`: add `@@Azure.ClientGenerator.Core.Legacy.flattenProperty(LabProperties.totalBudget, "!csharp")` and same for `totalAllocatedBudget`.

### DIFF
--- a/specification/education/resource-manager/Microsoft.Education/Education/client.tsp
+++ b/specification/education/resource-manager/Microsoft.Education/Education/client.tsp
@@ -45,3 +45,14 @@ using Microsoft.Education;
 
 @@clientName(GrantType, "EducationGrantType", "csharp");
 @@clientName(GrantStatus, "EducationGrantStatus", "csharp");
+
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "workaround: avoid "
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(
+  LabProperties.totalBudget,
+  "!csharp"
+);
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Property flatten for SDK backward compatibility."
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(
+  LabProperties.totalAllocatedBudget,
+  "!csharp"
+);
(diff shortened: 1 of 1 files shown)


specification/education/resource-manager/Microsoft.Education/Education/client.tsp
```diff
@@ -45,3 +45,15 @@ using Microsoft.Education;
...
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "workaround avoid generating duplicate members"
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(
+  LabProperties.totalBudget,
+  "!csharp"
+);
+#suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "workaround avoid generating duplicate members"
+@@Azure.ClientGenerator.Core.Legacy.flattenProperty(
+  LabProperties.totalAllocatedBudget,
+  "!csharp"
+);
```
ArcturusZhang: I think we should change the comment.
It is not "for now", we will never support the case in this library.

ArcturusZhang: also, I think we have the issue in the first place because somewhere else there is a `flattenProperty` that applies to all languages.
Please just modify the scope there.
Also I think it is valuable that we ask other language if they could handle this case. Maybe other language also needs to exclude this.

ArcturusZhang (approved):

PR #44691, opened 2026-07-15
Skip SDK generation stage when there is no release plan

praveenkuttappan:
Skip SDK generation stage and update SDK details step when release plan is not found for an API spec change.

### DIFF
--- a/eng/pipelines/release-plan.yml
+++ b/eng/pipelines/release-plan.yml
@@ -85,13 +85,6 @@ stages:
                 npm exec --no -- create-release-plan @nodeArgs
 
           - pwsh: |
-              # set variable SkipSDKGeneration to true by default
-              Write-Host "##vso[task.setvariable variable=SkipSDKGeneration]true"
-            name: setDefaultSkipSDKGeneration
-            displayName: "Set default SkipSDKGeneration variable"
-            condition: succeededOrFailed()
-
-          - pwsh: |
               $outputPath = "$(Build.ArtifactStagingDirectory)/release-plan/release-plan.json"
               if (!(Test-Path $outputPath)) {
                 Write-Host "release-plan.json not found. Setting SkipSDKGeneration to true."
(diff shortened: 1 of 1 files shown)


eng/pipelines/release-plan.yml
```diff
@@ -84,35 +84,49 @@ stages:
 
                 npm exec --no -- create-release-plan @nodeArgs
 
+          - pwsh: |
```
raych1: Both this step and the next step have same conditions, can we eliminate this step and set the variable in next step only?

eng/pipelines/release-plan.yml
```diff
@@ -84,35 +84,49 @@ stages:
...
-              scriptLocation: inlineScript
-              inlineScript: |
-                $ErrorActionPreference = "Stop"
-
-                $artifactPath = "$(Build.ArtifactStagingDirectory)/release-plan/release-plan.json"
-                if (!(Test-Path $artifactPath)) {
-                  Write-Host "Release plan artifact not found. Skipping update-sdk-details."
-                  exit 0
-                }
+          - ${{ if eq(variables['SkipSDKGeneration'], 'true') }}:
```
raych1: Seems to be redundant?

eng/pipelines/release-plan.yml
```diff
@@ -84,35 +84,49 @@ stages:
 
                 npm exec --no -- create-release-plan @nodeArgs
 
+          - pwsh: |
+              # set variable SkipSDKGeneration to true by default
+              Write-Host "##vso[task.setvariable variable=SkipSDKGeneration]true"
+            name: setDefaultSkipSDKGeneration
+            displayName: "Set default SkipSDKGeneration variable"
+            condition: succeededOrFailed()
+
```
praveenkuttappan:
```suggestion
```

raych1 (approved):

PR #44722, opened 2026-07-16
Fix C# failover property naming in client.tsp

schaudhari6254888:
# Choose a PR Template


specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/client.tsp
```diff
@@ -83,6 +83,7 @@ using Microsoft.EventHub;
 @@clientName(Encryption.keySource, "KeySource", "csharp");
 @@clientName(SchemaType, "EventHubsSchemaType", "csharp");
 @@clientName(SchemaCompatibility, "EventHubsSchemaCompatibility", "csharp");
+@@clientName(IpAddressType, "EventHubsIPAddressType", "csharp");
```
ArcturusZhang: we have another clientName on line 410. I think we need to modify that instead.

schaudhari6254888: My bad

ArcturusZhang (approved):

PR #42941, opened 2026-05-06
Added typespec for moveIpConfigurations API

rajanalavarnitha-wq:
This PR adds the POST operation for moving secondary IP configurations between network interfaces within a virtual network.

This mirrors the changes from azure-rest-api-specs-pr PR #24606.

## Changes
- Added `moveIpConfigurations` path operation to `virtualNetwork.json`
- Added `MoveIpConfigurationsRequest` and `MoveIpConfigurationItem` model definitions
- Added example file `VirtualNetworkMoveIpConfigurations.json`

PR #44735, opened 2026-07-16
Add moveIpConfigurations API to 2025-09-01 (port of #42941)

rajanalavarnitha-wq:
Ports the `moveIpConfigurations` API from #42941 into the 2025-09-01 API version.

Adapted to `@added(Versions.v2025_09_01)` and emitted under stable/2025-09-01. Adds the operation (VirtualNetwork.tsp), request/item/reference models (models.tsp), suppressions (readme.md), swagger path + definitions (virtualNetwork.json), and example files.

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/Network/readme.md
+++ b/specification/network/resource-manager/Microsoft.Network/Network/readme.md
@@ -123,6 +123,16 @@ suppressions:
     reason: Not a standard azure resource.
     where:
       - $.definitions.GetServiceGatewayServicesResult
+  - code: ResourceNameRestriction
+    from: virtualNetwork.json
+    reason: The resource name parameter 'virtualNetworkName' is not defined with a 'pattern' restriction. Suppress it to avoid breaking change because it is referenced by all Virtual Network APIs.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{virtualNetworkName}/moveIpConfigurations"]
+  - code: PostResponseCodes
+    from: virtualNetwork.json
+    reason: LRO POST operation returns 200 with no schema for completion status and 202 for async acceptance. This is the standard TypeSpec ArmResourceActionAsync pattern for void LRO operations.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{virtualNetworkName}/moveIpConfigurations"].post
 directive:
   - from: specification/common-types/resource-management/v6/types.json
     where: "$.definitions.ProxyResource"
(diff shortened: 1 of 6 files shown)


specification/network/resource-manager/Microsoft.Network/Network/readme.md
```diff
@@ -123,6 +123,16 @@ suppressions:
...
     where:
       - $.definitions.GetServiceGatewayServicesResult
+  - code: ResourceNameRestriction
+    from: virtualNetwork.json
+    reason: The resource name parameter 'virtualNetworkName' is not defined with a 'pattern' restriction. Suppress it to avoid breaking change because it is referenced by all Virtual Network APIs.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/virtualNetworks/{virtualNetworkName}/moveIpConfigurations"]
+  - code: PostResponseCodes
+    from: virtualNetwork.json
+    reason: LRO POST operation returns 200 with no schema for completion status and 202 for async acceptance. This is the standard TypeSpec ArmResourceActionAsync pattern for void LRO operations.
```
razvanbadea-msft: >200 [](http://example.com/codeflow?start=39&length=3)

if no schema is returned then code msut be 204 not 200

rajanalavarnitha-wq: Thanks @razvanbadea-msft. You're right that per the `PostResponseCodes` rule, an LRO POST returning no schema should use 204. However, we're intentionally keeping **200 (no schema) + 202** here and suppressing the rule, for the following reasons:

**1. Swagger must match the shipped implementation.** The NRP `MoveIpConfigurationsOperation` returns `HttpStatusCode.OK` (200) with no body on completion and `202` on async acceptance ( `Operation<NoContent>` ). Changing the spec to 204 would make the contract diverge from the actual service response, which we cannot do — and we are not able to change the service code in this change.

**2. This is the established NRP convention, not a one-off.** In this same `virtualNetwork.json`, `Subnets_PrepareNetworkPolicies` and `Subnets_UnprepareNetworkPolicies` already ship `200 (no schema) + 202` for void LRO POSTs. Across Microsoft.Network, 24 of 26 `NoContent` action operations return 200 (only 2 delete-style ops use 204). `moveIpConfigurations` follows the dominant, pre-existing pattern.

**3. Existing operations already trigger this same rule and are grandfathered.** LintDiff runs in diff mode, so pre-existing ops with 200-no-body (prepareNetworkPolicies, stop, resetconnection, etc.) don't fail. Because `moveIpConfigurations` is a new operation, the rule flags it — which is exactly what the suppression is scoped to address. I verified this empirically: removing the suppression makes `Swagger LintDiff` fail with a single `PostResponseCodes` error on the `moveIpConfigurations` path; re-adding it restores green.

So the suppression keeps the spec consistent with (a) the actual service behavior and (b) the surrounding NRP operations, rather than introducing a 204 contract the service doesn't implement. Happy to discuss if you'd prefer a different approach, but changing to 204 would require a service-side code change that's out of scope here.

sandipsh: Thanks for the detailed context. Keeping 200 + 202 for this action is acceptable since it matches current service behavior and existing NRP precedent, and the PostResponseCodes suppression is narrowly scoped to this path.

PR #44750, opened 2026-07-17
[Event Hubs] Add 2026-07-01-preview API version

mashrafsiddiquee:
## Summary

- Add the `2026-07-01-preview` Event Hubs management API version based on `2026-01-01`.
- Add Fabric Shortcut CRUD, list, approve, and reject APIs from internal PR 15914488.
- Add Dedicated cluster Upgrade Preferences get, create/update, and upgrade-now APIs from internal PR 16291983.
- Add TypeSpec source examples, generated examples, and the generated preview OpenAPI document.

## Validation

- TypeSpec compilation completed successfully.
- The generated OpenAPI document passed `oav validate-spec`.
- All `2026-01-01` operations and schemas remain unchanged in the new preview version.

## Purpose of this PR

- [x] New API version for an existing resource provider.

## Due diligence checklist

- [ ] I have reviewed the Resource Provider guidelines, ARM resource provider contract, and REST guidelines.
- [ ] A release plan has been created.

### DIFF
--- a/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
+++ b/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
@@ -1,0 +1,225 @@
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-azure-resource-manager";
+import "@typespec/rest";
+import "@typespec/versioning";
+import "./models.tsp";
+import "./Eventhub.tsp";
+
+using TypeSpec.Rest;
+using Azure.ResourceManager;
+using Azure.Core;
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+
+namespace Microsoft.EventHub;
+
+/**
+ * A Microsoft Fabric shortcut attached to an Event Hub.
+ */
+@added(Versions.v2026_07_01_preview)
+@parentResource(Eventhub)
+model FabricShortcut
+  is Azure.ResourceManager.ProxyResource<FabricShortcutProperties> {
+  ...ResourceNameParameter<
+    Resource = FabricShortcut,
+    KeyName = "fabricShortcutName",
+    SegmentName = "fabricShortcuts",
+    NamePattern = ""
+  >;
+
+  /**
+   * The geo-location where the resource lives.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "The service returns the parent namespace location in the existing ARM resource envelope."
+  @visibility(Lifecycle.Read)
+  location?: string;
+}
+
+/**
+ * Properties of a Microsoft Fabric shortcut.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "The service does not expose a provisioning state for this proxy resource."
+@added(Versions.v2026_07_01_preview)
+model FabricShortcutProperties {
+  /**
+   * Microsoft Fabric workspace and artifact configuration.
+   */
+  configuration: FabricShortcutConfiguration;
+
+  /**
+   * The type of the shortcut.
+   */
+  shortcutType?: FabricShortcutType;
+
+  /**
+   * The current shortcut status. Only Pending can be supplied on create or update.
+   */
+  shortcutStatus?: FabricShortcutStatus;
+
+  /**
+   * A description of the current shortcut status.
+   */
+  @visibility(Lifecycle.Read)
+  statusDescription?: string;
+
+  /**
+   * The UTC time when the shortcut was created.
+   */
+  @visibility(Lifecycle.Read)
+  createdAt?: utcDateTime;
+
+  /**
+   * The UTC time when the shortcut was last modified.
+   */
+  @visibility(Lifecycle.Read)
+  modifiedAt?: utcDateTime;
+}
+
+/**
+ * Microsoft Fabric workspace and artifact configuration.
+ */
+@added(Versions.v2026_07_01_preview)
+model FabricShortcutConfiguration {
+  /**
+   * The Microsoft Fabric tenant ID.
+   */
+  tenantId: uuid;
+
+  /**
+   * The Microsoft Fabric workspace ID.
+   */
+  workspaceId: uuid;
+
+  /**
+   * The Microsoft Fabric artifact ID.
+   */
+  artifactId: uuid;
+
+  /**
+   * The Microsoft Fabric premium capacity ID.
+   */
+  premiumCapacityId?: string;
+
+  /**
+   * The resource ID of the Log Analytics workspace.
+   */
+  logAnalyticsResourceId?: string;
+
+  /**
+   * The Microsoft Fabric workspace name.
+   */
+  workspaceName?: string;
+
+  /**
+   * The Microsoft Fabric artifact name.
+   */
+  artifactName?: string;
+}
+
+/**
+ * The type of Microsoft Fabric shortcut.
+ */
+#suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "The service validates the value against this fixed set and rejects unknown values."
+@added(Versions.v2026_07_01_preview)
+union FabricShortcutType {
+  /**
+   * An entity shortcut.
+   */
+  Entity: "Entity",
+
+  /**
+   * A network shortcut.
+   */
+  Network: "Network",
+}
+
+/**
+ * The status of a Microsoft Fabric shortcut.
+ */
+#suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "The service validates the value against this fixed set and rejects unknown values."
+@added(Versions.v2026_07_01_preview)
+union FabricShortcutStatus {
+  /**
+   * The shortcut is awaiting approval.
+   */
+  Pending: "Pending",
+
+  /**
+   * The shortcut is approved.
+   */
+  Approved: "Approved",
+
+  /**
+   * The shortcut is rejected.
+   */
+  Rejected: "Rejected",
+}
+
+@added(Versions.v2026_07_01_preview)
+@armResourceOperations
+interface FabricShortcuts {
+  /**
+   * Gets a Microsoft Fabric shortcut.
+   */
+  get is ArmResourceRead<FabricShortcut, Error = ErrorResponse>;
+
+  /**
+   * Creates or updates a Microsoft Fabric shortcut.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "The service returns 200 for both create and update."
+  createOrUpdate is ArmResourceCreateOrReplaceSync<
+    FabricShortcut,
+    Response = ArmResourceUpdatedResponse<FabricShortcut>,
+    Error = ErrorResponse
+  >;
+
+  /**
+   * Lists Microsoft Fabric shortcuts for an Event Hub.
+   */
+  listByEventHub is ArmResourceListByParent<
+    FabricShortcut,
+    Error = ErrorResponse
+  >;
+
+  /**
+   * Deletes a Microsoft Fabric shortcut.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "The service returns 200 when the shortcut is deleted."
+  delete is ArmResourceDeleteSync<
+    FabricShortcut,
+    Response = ArmDeletedResponse,
+    Error = ErrorResponse
+  >;
+
+  /**
+   * Approves a Microsoft Fabric shortcut.
+   */
+  @action("approve")
+  approve is ArmResourceActionSync<
+    FabricShortcut,
+    void,
+    ArmResponse<FabricShortcut>,
+    Error = ErrorResponse
+  >;
+
+  /**
+   * Rejects a Microsoft Fabric shortcut.
+   */
+  @action("reject")
+  reject is ArmResourceActionSync<
+    FabricShortcut,
+    void,
+    ArmResponse<FabricShortcut>,
+    Error = ErrorResponse
+  >;
+}
+
+@@doc(FabricShortcut.name, "The Microsoft Fabric shortcut name.");
+@@doc(
+  FabricShortcut.properties,
+  "Properties of the Microsoft Fabric shortcut."
+);
+@@doc(
+  FabricShortcuts.createOrUpdate::parameters.resource,
+  "The Microsoft Fabric shortcut."
+);
--- a/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
+++ b/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
@@ -24,13 +24,13 @@ model FabricShortcut
     Resource = FabricShortcut,
     KeyName = "fabricShortcutName",
     SegmentName = "fabricShortcuts",
-    NamePattern = ""
+    NamePattern = "^.+$"
   >;
 
   /**
    * The geo-location where the resource lives.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "The service returns the parent namespace location in the existing ARM resource envelope."
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "The backend resource model returns location as a read-only top-level ARM envelope property, not as a Fabric shortcut property."
   @visibility(Lifecycle.Read)
   location?: string;
 }
@@ -38,7 +38,7 @@ model FabricShortcut
 /**
  * Properties of a Microsoft Fabric shortcut.
  */
-#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "The service does not expose a provisioning state for this proxy resource."
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "The synchronous backend contract does not define or return provisioningState; shortcut lifecycle is reported through the read-only shortcutStatus property."
 @added(Versions.v2026_07_01_preview)
 model FabricShortcutProperties {
   /**
@@ -103,7 +103,11 @@ model FabricShortcutConfiguration {
   /**
    * The resource ID of the Log Analytics workspace.
    */
-  logAnalyticsResourceId?: string;
+  logAnalyticsResourceId?: armResourceIdentifier<[
+    {
+      type: "Microsoft.OperationalInsights/workspaces";
+    }
+  ]>;
 
   /**
    * The Microsoft Fabric workspace name.
--- a/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
+++ b/specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
@@ -273,12 +273,60 @@ These settings apply only when `--tag=package-2026-07-preview` is specified on t
 ``` yaml $(tag) == 'package-2026-07-preview'
 input-file:
 - preview/2026-07-01-preview/openapi.json
+- preview/2018-01-01-preview/virtualnetworkrules-preview.json
+- preview/2018-01-01-preview/ipfilterrules-preview.json
+- preview/2018-01-01-preview/sku.json
 ```
 
 ## Suppression
 
 ``` yaml
 directive:
+  - suppress: ResourceNameRestriction
+    from: openapi.json
+    reason: Existing Event Hub and cluster parent names and Fabric shortcut names have no service-enforced regular expression constraint.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/clusters/{clusterName}/upgradePreferences/default"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/clusters/{clusterName}/upgradePreferences/default/upgradeNow"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}/approve"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}/reject"]
+  - suppress: PutResponseCodes
+    from: openapi.json
+    reason: The service returns 200 for both create and update for these resources.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/clusters/{clusterName}/upgradePreferences/default"].put
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].put
+  - suppress: PostResponseCodes
+    from: openapi.json
+    reason: Upgrade now returns 200 when an upgrade starts and 204 when no upgrade is pending.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/clusters/{clusterName}/upgradePreferences/default/upgradeNow"].post
+  - suppress: RequestSchemaForTrackedResourcesMustHaveTags
+    from: openapi.json
+    reason: Fabric shortcuts are proxy resources and do not support tags.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].put
+  - suppress: DeleteResponseCodes
+    from: openapi.json
+    reason: Fabric shortcut deletion returns 200 on success.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].delete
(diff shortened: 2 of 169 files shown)


mashrafsiddiquee:
Avocado suppression review requested for the five `MISSING_APIS_IN_DEFAULT_TAG` findings. The `ipfilterrules` and `virtualnetworkrules` child-resource routes are supported only by `2018-01-01-preview` and were superseded by `networkRuleSets`; the legacy `/sku/{sku}/regions` route is not implemented for current API versions. The suppressions are scoped to the three source files and exact five paths, while the default package remains a single `2026-07-01-preview` API version. Please apply `Approved-Avocado` if the rationale is accepted.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
```diff
@@ -266,10 +266,81 @@ input-file:
...
+``` yaml $(tag) == 'package-2026-07-preview'
+input-file:
+- preview/2026-07-01-preview/openapi.json
+```
+
 ## Suppression
 
 ``` yaml
 directive:
+  - suppress: MISSING_APIS_IN_DEFAULT_TAG
```
razvanbadea-msft: >MISSING_APIS_IN_DEFAULT_TAG [](http://example.com/codeflow?start=14&length=27)

for avocado you cant suppress errors using readme file. just leave a message about the reason and we will apply an approval label for it

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
```diff
@@ -266,10 +266,81 @@ input-file:
...
+    reason: These legacy child-resource routes are supported only by the 2018-01-01-preview controller and were replaced by networkRuleSets in current API versions.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/ipfilterrules"]
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/ipfilterrules/{ipFilterRuleName}"]
+  - suppress: MISSING_APIS_IN_DEFAULT_TAG
+    from: preview/2018-01-01-preview/sku.json
+    reason: The legacy SKU region discovery route is not implemented by the current Event Hubs SKU controller.
+    where: $.paths["/subscriptions/{subscriptionId}/providers/Microsoft.EventHub/sku/{sku}/regions"]
+
+  - suppress: ResourceNameRestriction
```
razvanbadea-msft, line 282: >ResourceNameRestriction [](http://example.com/codeflow?start=14&length=23)

can you have empty string as valid names?

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+    SegmentName = "fabricShortcuts",
+    NamePattern = ""
+  >;
+
+  /**
+   * The geo-location where the resource lives.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "The service returns the parent namespace location in the existing ARM resource envelope."
+  @visibility(Lifecycle.Read)
+  location?: string;
```
razvanbadea-msft, line 35: >location [](http://example.com/codeflow?start=2&length=8)

why it was not included in FabricShortcutProperties?

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * The Microsoft Fabric premium capacity ID.
+   */
+  premiumCapacityId?: string;
+
+  /**
+   * The resource ID of the Log Analytics workspace.
+   */
+  logAnalyticsResourceId?: string;
```
razvanbadea-msft: >string [](http://example.com/codeflow?start=27&length=6)

you can use armResourceIdentifier with type restriction

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "The service returns the parent namespace location in the existing ARM resource envelope."
+  @visibility(Lifecycle.Read)
+  location?: string;
+}
+
+/**
+ * Properties of a Microsoft Fabric shortcut.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "The service does not expose a provisioning state for this proxy resource."
```
razvanbadea-msft: >The service does not expos [](http://example.com/codeflow?start=90&length=26)

what is the reason?

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+interface FabricShortcuts {
+  /**
+   * Gets a Microsoft Fabric shortcut.
+   */
+  get is ArmResourceRead<FabricShortcut, Error = ErrorResponse>;
+
+  /**
+   * Creates or updates a Microsoft Fabric shortcut.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "The service returns 200 for both create and update."
```
razvanbadea-msft, line 173: >200 [](http://example.com/codeflow?start=113&length=3)

what is the reason for this

razvanbadea-msft, line 173: >200 [](http://example.com/codeflow?start=113&length=3)

why?

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
```diff
@@ -266,10 +266,81 @@ input-file:
...
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].delete
+  - suppress: DeleteOperationResponses
+    from: openapi.json
+    reason: Fabric shortcut deletion returns an empty 200 response on success.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].delete.responses
+  - suppress: GuidUsage
+    from: openapi.json
+    reason: Microsoft Fabric tenant, workspace, and artifact identifiers are UUIDs by definition.
+    where: $.definitions["Azure.Core.uuid"].format
+  - suppress: TrackedResourcePatchOperation
```
razvanbadea-msft, line 318: >TrackedResourcePatchOperation [](http://example.com/codeflow?start=14&length=29)

this will not be flagged once you move location in property bag. now with location added it is seen as a tracked resource

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
```diff
@@ -26,7 +26,7 @@ These are the global settings for the EventHub API.
 
 ``` yaml
 openapi-type: arm
-tag: package-2026-01
+tag: package-2026-07-preview
```
razvanbadea-msft: **[NEW] 🔴 Blocking** **[[Section 11.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#112-incremental-version-progression)]** `readme.md` - line 29 - Default AutoRest tag changed from stable to preview.

**Classification reasoning:** Introduced in this PR — the global default tag was `package-2026-01` (stable) at base SHA `33edaba4` and is changed to `package-2026-07-preview` here.

The global default tag must point to the latest **stable** version; SDK consumers get the default tag by default and it must be GA. Changing it to a preview also forced the new `MISSING_APIS_IN_DEFAULT_TAG` suppressions below (the preview omits legacy routes the stable default carried).

**Suggested fix:** Restore `tag: package-2026-01` as the global default and keep `package-2026-07-preview` as its own named tag only.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+   */
+  artifactName?: string;
+}
+
+/**
+ * The type of Microsoft Fabric shortcut.
+ */
+#suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "The service validates the value against this fixed set and rejects unknown values."
+@added(Versions.v2026_07_01_preview)
+union FabricShortcutType {
```
razvanbadea-msft, line 128: **[NEW] 🟠 Warning** **[[enum-best-practices](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/enum-best-practices.md)]** `FabricShortcut.tsp` - line 124 - Closed (non-extensible) enum.

`FabricShortcutType` (line 124) and `FabricShortcutStatus` (line 141) here, plus `UpgradePreferenceDayOfWeek` (`UpgradePreferences.tsp` line 135) and `ExceptionWindowAction` (line 163), emit `modelAsString: false` via the `no-closed-literal-union` suppression. Extensible enums (`modelAsString: true`) should be used so new values can be added later without a breaking change — relevant for `FabricShortcutStatus`/`FabricShortcutType`.

**Classification reasoning:** New enums introduced in this PR.

**Suggested fix:** Remove the closed-union suppression and model these as open (extensible) unions.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Gets a Microsoft Fabric shortcut.
+   */
+  get is ArmResourceRead<FabricShortcut, Error = ErrorResponse>;
+
+  /**
+   * Creates or updates a Microsoft Fabric shortcut.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "The service returns 200 for both create and update."
+  createOrUpdate is ArmResourceCreateOrReplaceSync<
```
razvanbadea-msft, line 174: **[NEW] 🟠 Warning** **[[RPC-Put-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `FabricShortcut.tsp` - line 170 - PUT returns only 200, not 201 on create.

Both `createOrUpdate` operations (here, and `UpgradePreferences.tsp` line 188) use `ArmResourceUpdatedResponse` (200 only) and suppress `PutResponseCodes`. PUT must return 201 for creation and 200 for replacement so clients/SDKs can distinguish create from update.

**Classification reasoning:** New operations in this PR.

**Suggested fix:** Return `ArmResourceCreatedResponse | ArmResourceUpdatedResponse` (201 + 200) and drop the suppression, unless the service genuinely cannot emit 201.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
```diff
@@ -0,0 +1,225 @@
...
+  listByEventHub is ArmResourceListByParent<
+    FabricShortcut,
+    Error = ErrorResponse
+  >;
+
+  /**
+   * Deletes a Microsoft Fabric shortcut.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-delete-operation-response-codes" "The service returns 200 when the shortcut is deleted."
+  delete is ArmResourceDeleteSync<
```
razvanbadea-msft, line 192: **[NEW] 🟠 Warning** **[[RPC-Delete-V1-01](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#51-delete-response-codes-rpc-delete-v1-01)]** `FabricShortcut.tsp` - line 188 - Sync DELETE missing 204.

`delete` returns 200 only and suppresses `DeleteResponseCodes`/`DeleteOperationResponses`. A synchronous DELETE must define 200 + 204 + default (204 when the resource does not exist — never 404).

**Classification reasoning:** New operation in this PR.

**Suggested fix:** Add a 204 response.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
```diff
@@ -266,10 +266,81 @@ input-file:
...
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].put
+  - suppress: DeleteResponseCodes
+    from: openapi.json
+    reason: Fabric shortcut deletion returns 200 on success.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].delete
+  - suppress: DeleteOperationResponses
+    from: openapi.json
+    reason: Fabric shortcut deletion returns an empty 200 response on success.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.EventHub/namespaces/{namespaceName}/eventhubs/{eventHubName}/fabricShortcuts/{fabricShortcutName}"].delete.responses
+  - suppress: GuidUsage
```
razvanbadea-msft, line 314: **[NEW] 🔵 Suggestion** **[[R3017 GuidUsage](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/guid-and-uuid-on-arm.md)]** `readme.md` - line 331 - Over-scoped `GuidUsage` suppression.

The suppression targets the shared `$.definitions["Azure.Core.uuid"].format`, covering `tenantId` (`FabricShortcut.tsp` line 86 — acceptable, Entra-adjacent) as well as `workspaceId`/`artifactId` (lines 91/96), which are Microsoft Fabric identifiers, not Entra IDs. `format: uuid` on ARM control-plane is discouraged except for Entra IDs, and R3017 requires Azure API review board sign-off.

**Classification reasoning:** New suppression in this PR.

**Suggested fix:** Confirm the `ARMModelingAutoSignedOff` sign-off covers the non-Entra Fabric identifiers; otherwise keep `workspaceId`/`artifactId` as `string`.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
razvanbadea-msft: Adding detail (from the agent review): `location?: string` on line 35 sits on the `ProxyResource` envelope, which required the `arm-resource-invalid-envelope-property` suppression on line 33. It is also the root cause of two other suppressions in `readme.md` — `RequestSchemaForTrackedResourcesMustHaveTags` and `TrackedResourcePatchOperation` — because the linter treats the resource as tracked-ish once `location` is on the envelope. Moving `location` into `FabricShortcutProperties` (as you suggested) removes all three suppressions in one change.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
razvanbadea-msft: Adding detail (from the agent review): this is `logAnalyticsResourceId?: string` on line 106 — a fully-qualified ARM resource ID that should use `armResourceIdentifier` with a type restriction so ARM can perform RBAC linked-access checks (Section 2.4). Also worth checking `premiumCapacityId` on line 101: if it is an ARM resource ID it should get the same treatment; if it is a Fabric-internal (non-ARM) ID, plain `string` is fine.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
razvanbadea-msft: Adding detail (from the agent review): the empty-name behavior comes from `NamePattern = ""` in the TypeSpec — `FabricShortcut.tsp` line 27 and `UpgradePreferences.tsp` line 27. The empty pattern disables name validation, which is what forced the `ResourceNameRestriction` suppression. If the service enforces a name shape, setting a real `NamePattern` removes the suppression; if arbitrary/empty names are genuinely valid, that is worth confirming explicitly.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The synchronous backend contract (`FabricShortcutArmDescription : ArmDescription`) does not define or return `provisioningState`. Shortcut lifecycle is exposed through the read-only domain-specific `shortcutStatus` instead. I clarified this rationale in cb1352ac70.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The backend wire model keeps this in the ARM envelope: `FabricShortcutArmResourceInfo` inherits `ArmResourceInfo<T>`, where `Location` is a top-level `[NotUserSettable]` property, while `FabricShortcutArmDescription` has no location property. I kept that exact read-only shape and clarified the suppression rationale in cb1352ac70.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
mashrafsiddiquee: No, empty names are not valid. In cb1352ac70 I added a non-empty pattern for `fabricShortcutName`; Upgrade Preferences is the literal singleton name `default`. The remaining scoped suppression is only for the existing `clusterName` and `eventHubName` parent parameters, whose shipped service contracts do not enforce a regex that can be introduced only for this preview.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: Updated in cb1352ac70 to `armResourceIdentifier` restricted to `Microsoft.OperationalInsights/workspaces`; the regenerated Swagger now emits `format: arm-id` and the allowed resource type.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
mashrafsiddiquee: Removed the `MISSING_APIS_IN_DEFAULT_TAG` directives in cb1352ac70. The PR-level comment retains the rationale for the five legacy routes so the Avocado approval label can be applied.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
mashrafsiddiquee: The backend validates `tenantId`, `workspaceId`, and `artifactId` with `Guid.TryParse`, so all three are UUIDs in the service contract. The suppression is scoped to the emitted shared UUID definition and requires ARM/API review sign-off; changing the Fabric IDs to unconstrained strings would weaken the contract.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The service implementation returns 200 for Fabric Shortcut creation as well as update, and its lifecycle test asserts 200 on the initial PUT. A 201 response is not currently implemented, so the specification retains the scoped suppression.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: Also confirmed `premiumCapacityId` is a Microsoft Fabric-internal capacity identifier, not an Azure Resource Manager resource ID, so it remains a plain string. Only `logAnalyticsResourceId` is an ARM ID and is now type-restricted.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The Fabric Shortcut delete implementation and CIT use 200 on successful deletion; 204 is not implemented. The specification keeps the 200-only response with narrowly scoped suppressions rather than claiming an unsupported response.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The Fabric Shortcut PUT backend returns 200 for both create and update; the service CIT explicitly asserts `HttpStatusCode.OK` on create. The TypeSpec response and suppression intentionally match that deployed contract rather than advertising an unsupported 201.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
mashrafsiddiquee: Fixed in 6036a83a73: the global default tag is restored to stable `package-2026-01`; `package-2026-07-preview` remains available as its named tag.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: The service genuinely cannot emit 201 today. Fabric Shortcut CIT asserts 200 on create, and `UpgradePreferencesRequestCoordinator.PutUpgradePreferencesAsync` returns `HttpStatusCode.OK` after persisting. The 200-only responses and scoped suppressions match the implementation.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
mashrafsiddiquee: Moving `location` would change the JSON wire contract. The backend response type inherits `ArmResourceInfo<T>`, where `Location` is a top-level read-only envelope property; `FabricShortcutArmDescription` has no location property. The tracked-resource findings are therefore false positives caused by accurately modeling that existing backend shape, so the narrowly scoped suppressions must remain unless the service contract changes.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
mashrafsiddiquee: These unions are intentionally closed because the backend uses closed C# enums with `StringEnumConverter` and rejects unknown values. Making the specification extensible would tell SDKs to accept values that the service does not accept, so the scoped suppressions preserve the actual contract.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/FabricShortcut.tsp
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

specification/eventhub/resource-manager/Microsoft.EventHub/Eventhub/readme.md
sandipsh: Thanks for the detailed clarification and update. This addresses my concern. Resolving this thread.

dsouzaarun (approved):

PR #44760, opened 2026-07-17
Fix csharp-mgmt emitter config for ContainerService NodeCustomization

cxznmhdcxz:
## Summary
Fixes the C# management-plane emitter configuration for the ContainerService **NodeCustomization** TypeSpec project.

## Problem
The `@azure-typespec/http-client-csharp-mgmt` emitter block declared a `service-dir` option. That emitter's option schema uses `additionalProperties: false` and does not define `service-dir`, so compilation failed with:

```
tspconfig.yaml - error invalid-schema: must NOT have additional properties (additionalProperty: service-dir)
```

This silently produced an empty `src/` during .NET SDK generation.

## Fix
Remove the invalid `service-dir` option from the csharp-mgmt block and hardcode the C# service directory into `emitter-output-dir` (matching the pattern used by other migrated mgmt specs, e.g. Advisor).

Paired with the .NET SDK PR: Azure/azure-sdk-for-net (ContainerServiceNodeCustomization).


specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/tspconfig.yaml
```diff
@@ -13,6 +13,9 @@ options:
     arm-types-dir: "{project-root}/../../../../common-types/resource-management"
     examples-dir: "{project-root}/examples"
     use-read-only-status-schema: true
+  "@azure-typespec/http-client-csharp-mgmt":
+    namespace: "Azure.ResourceManager.ContainerServiceNodeCustomization"
+    emitter-output-dir: "{output-dir}/sdk/containerservice/{namespace}"
   "@azure-tools/typespec-java":
     namespace: "com.azure.resourcemanager.containerservice.nodecustom"
```
ArthurMa1978: this should be `com.azure.resourcemanager.containerservicenodecustomization`

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/tspconfig.yaml
```diff
@@ -13,6 +13,9 @@ options:
...
     package-dir: "azure-resourcemanager-containerservice-nodecustom"
```
ArthurMa1978: `azure-resourcemanager-containerservicenodecustomization`

specification/containerservice/resource-manager/Microsoft.ContainerService/nodecustomization/tspconfig.yaml
ArthurMa1978: `{output-dir}/{service-dir}/azure-resourcemanager-containerservicenodecustomization`

ArthurMa1978, line 42: Refer https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/tspconfig.yaml#L47 to add setting for js

cxznmhdcxz: ts section exists

ArthurMa1978 (approved):

PR #44766, opened 2026-07-17
Skip folder migration spec PRs from automated release plan creation

praveenkuttappan:
Skip any spec PR tagged using Folder migration label when running automation to create a release plan. It should also ignore all file changes that we only rename when searching for modified API version.

### DIFF
--- a/eng/tools/release-plan/src/typespec-project.ts
+++ b/eng/tools/release-plan/src/typespec-project.ts
@@ -9,6 +9,15 @@ import type {
   TypeSpecProjectInfo,
 } from "./types.ts";
 
+/** Label indicating that a spec PR introduces a new API version. */
+export const NEW_API_VERSION_LABEL = "new-api-version";
+
+/**
+ * Label indicating that a spec PR only migrates/renames folders. Release plan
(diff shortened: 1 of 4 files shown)


eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -9,6 +9,15 @@ import type {
   TypeSpecProjectInfo,
 } from "./types.ts";
 
+/** Label indicating that a spec PR introduces a new API version. */
+export const NEW_API_VERSION_LABEL = "new-api-version";
```
raych1, line 13: Is this label being used to determine whether the automation for release plan create/update and SDK PR generation would be triggered?

I thought the code would always search for release plans with latest api-version then proceed with generation if the release plan hasn't been created or hasn't been released. Is it correct?

praveenkuttappan: A new release plan is created only when a new API version is added. Automation looks for existing release plan if spec PR contains changes but not adding any new API version. Release plan depends on new API version.

raych1 (approved):

PR #44802, opened 2026-07-20
docs(agents-optimization): document max_stalls service default of 5

imatiach-msft:
## Summary

Follow-up to the PR that added `max_stalls` to `OptimizationOptions`.

Per reviewer feedback, the previous doc comment said *"uses its internal default"* without naming the value. This update makes it explicit: **defaults to 5 when omitted**.

## Change

`src/agents-optimization/models.tsp`: replace *"When omitted, the optimizer uses its internal default."* with *"Defaults to 5 when omitted."* in the `max_stalls` doc comment.

This follows the pattern already used by other optional fields in the same model (e.g. `max_candidates` says `Default: 5`, `agent_version` says `Defaults to latest if omitted`).

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
@@ -197,9 +197,9 @@ model OptimizationOptions {
   @doc("Evaluation granularity. Null/omitted means per-item single-turn. Set to 'conversation' for per-conversation multi-turn simulation scoring.")
   evaluation_level?: EvaluationLevel;
 
-  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. When omitted, the optimizer uses its internal default. Must be >= 1 when set.")
+  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. The service defaults to 5 if a value is not specified by the caller. Must be >= 1 when set.")
   @minValue(1)
-  max_stalls?: int32;
+  max_stalls?: int32 = 5;
 }
 
 @doc("Caller-supplied inputs for an optimization job.")


specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
```diff
@@ -197,7 +197,7 @@ model OptimizationOptions {
...
   evaluation_level?: EvaluationLevel;
 
-  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. When omitted, the optimizer uses its internal default. Must be >= 1 when set.")
+  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. Defaults to 5 when omitted. Must be >= 1 when set.")
   @minValue(1)
```
dargilco, line 201: To align with other statements, please update this to "The service defaults to 5 if a value is not specified by the caller". This is because the doc-string of the auto-generated Python code adds another sentence at the end, something like "Defaults to None", referring to the client-side default. So we need to makes sure it's clear in the doc string what is the service-side default behavior vs. the client side default value.

imatiach-msft: Good point — updated to 'The service defaults to 5 if a value is not specified by the caller.' to make the service-side default distinct from the client-side None. Change is in the latest commit.

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents-optimization/models.tsp
```diff
@@ -197,7 +197,7 @@ model OptimizationOptions {
...
   evaluation_level?: EvaluationLevel;
 
-  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. When omitted, the optimizer uses its internal default. Must be >= 1 when set.")
+  @doc("Maximum number of consecutive reflective minibatch rejections before stopping early. A 'stall' occurs when the optimizer proposes a prompt change, evaluates it on a small subset, and the score does not improve — so no full validation-set evaluation is triggered. The counter resets whenever a minibatch passes and its full-validation score beats the current best. Only a sustained plateau of `max_stalls` consecutive minibatch failures triggers the stop. The service defaults to 5 if a value is not specified by the caller. Must be >= 1 when set.")
   @minValue(1)
   max_stalls?: int32;
```
glecaros:
```suggestion
  max_stalls?: int32 = 5;
```

imatiach-msft: Applied — changed to `max_stalls?: int32 = 5;`. As discussed, the `= 5` has documentation/tooling value in TypeSpec while emitted Azure SDKs will still default to `None` client-side (no value sent to the service unless the caller sets it explicitly).

PR #44805, opened 2026-07-20
Migrate readme.md swagger versions to service.yaml

timotheeguerin:
## Summary

Adds `service.yaml` manifests derived from the legacy AutoRest `readme.md` files, as part of the rollout of the new `service.yaml` format introduced in [Azure/typespec-azure#4846](https://github.com/Azure/typespec-azure/pull/4846) (tracking [#4826](https://github.com/Azure/typespec-azure/issues/4826)).

The typespec-autorest emitter already writes `service.yaml` for the versions currently in each `@versioned` enum. This PR fills in the **legacy/swagger side** by treating each `readme.md` as the source of truth, so the generated manifests are a **superset** that also captures historical swagger-only versions.

## How it was generated

A one-time script ([timotheeguerin/jit-scripts](https://github.com/timotheeguerin/jit-scripts)) that, for every `tspconfig.yaml` emitting `@azure-tools/typespec-autorest`:

1. resolves the swagger output-path template (`output-dir` → `emitter-output-dir` → `output-file`, incl. custom `parameters:` defaults),
2. finds readme `input-file` entries under that path,
3. groups them by API version (unioned across tags),
4. detects `source` per version (`typespec` if the swagger has `x-typespec-generated`, else `swagger`),
5. writes `service.yaml` at the tsp project root.

## Scope

- **562** `service.yaml` files across TypeSpec-anchored services.
- Legacy no-tsp services and ambiguous/shared files are intentionally **not** included here (separate pass).

## Validation

Compared against the emitter-produced ground truth: **518 exact superset matches**; remaining differences are genuine readme omissions (e.g. shared `common.json`) or hand-authored swaggers the readme lists but the emitter doesn't generate.


specification/compute/resource-manager/Microsoft.Compute/Compute/service.yaml
```diff
@@ -0,0 +1,17 @@
+versions:
```
lirenhe: It looks this file is the duplicate of \Compute\Compute\service.yaml,   is this expected?

specification/compute/resource-manager/Microsoft.Compute/Compute/service.yaml
qiaozha, line 1: @timotheeguerin  I wonder what exactly does service.yaml stand for? does it stand for a service concept that public customer facing as defined in the service versioning guideline? Or does it stand for a simple TypeSpec project? 
Not sure if you are aware, previously we think TypeSpec and public customer facing service are 1:1 mapping, but
in cases like Compute, the 1:1 mapping has been broken as service team strongly pushed back for the service split breaking change. Now, it has three customer facing services Compute, Bulkactions, Recommender, each of the public customer-facing services have its own docs, SDKs which is aligned with the 1:1 mapping from the guideline. But within the Compute service, it has four different subservice Compute, ComputeDisk, ComputeGallery, ComputeSku, each of the sub services is an independent TypeSpec project, so it's multiple TypeSpec projects maps to the same service.  
/cc @mikeharder  as we had this discussion when we support folder structure for services like Compute.

qiaozha:
@timotheeguerin will you also add the service.yaml for those private repo branches like RPSaaSMaster and RPSaaSDev?

timotheeguerin:
> @timotheeguerin will you also add the service.yaml for those private repo branches like RPSaaSMaster and RPSaaSDev? 

Yes, it will be a tsv requirement will migrate RPSaaSMaster as well but rpsaasdev is up to the pr author

specification/compute/resource-manager/Microsoft.Compute/Compute/service.yaml
timotheeguerin: yeah so the idea is that the service.yaml is tied to the typespec project which as far as I understand from previous discussion and your comment should be the right mapping here.

timotheeguerin: I guess I'm quite confused by compute tspconfig.yaml then, both 
- specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/tspconfig.yaml
- specification/compute/resource-manager/Microsoft.Compute/Compute/tspconfig.yaml
have autorest configured to emit `ComputeRP.json` which is why it was converted that way
which one is the correct one?

markcowl (approved):

PR #44811, opened 2026-07-20
Managed NVA Ipv6 Swagger changes

debipdasfeb14:
- Adding addressFamily[] field to managed nva properties
- Adding addressPrefixV6, privateIpAddressV6, publicIpAddressV6 fields for dual stack nva 
- Additional NIC and Customer profile will also get Ipv6 configurations if the addressFamily contains ["IPv4", "IPv6"] whihc means dual stack nva deployment.


sandipsh:
please fix the validation errors.

debipdasfeb14:
> please fix the validation errors.

Fixed the errors

specification/network/resource-manager/Microsoft.Network/Network/stable/2025-09-01/common.json
```diff
@@ -137,6 +137,30 @@
         ]
       }
     },
+    "AddressFamily": {
```
sandipsh: 🟠 **[NEW] Warning — consider reusing `Common.IPVersion` instead of adding a new `AddressFamily` enum.**

This file already defines `Common.IPVersion` at line 2308 with the same `IPv4`/`IPv6` values and `modelAsString: true`, and it is already referenced 5× in `virtualNetwork.json` (lines 15324, 15967, 16130, 18322, 19633). Introducing a parallel `AddressFamily` enum with the identical shape produces a second SDK model class in every language that means the same thing — an unforced intra-file consistency drift.

**Why it matters:** cleaner SDK surface (one class instead of two for the same concept), no ambiguity for consumers, no need to keep the two enums in sync as values evolve.

**Recommended fix:** in `models.tsp`, drop `union AddressFamily` and change `addressFamily?: AddressFamily[]` → `addressFamily?: IPVersion[]` (or the closest existing atomic IPv4/IPv6 enum). Regenerate; the new `AddressFamily` definition here would then drop out.

If you kept `AddressFamily` intentionally to allow future divergence from `IPVersion` (e.g., non-IP address families), please reply confirming so we can accept the divergence with rationale.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -15966,11 +15991,24 @@ model NetworkVirtualAppliancePropertiesFormat {
   @maxItems(3)
   nvaInterfaceConfigurations?: NvaInterfaceConfigurationsProperties[];
 
+  /**
+   * The address families to deploy the NVA in. ["IPv4", "IPv6"] deploys a dual-stack NVA (both IPv4 and IPv6 configurations on the NICs; the vHub/VNet must be dual-stack). ["IPv4"], an empty array, or omitting the field deploys an IPv4-only NVA. ["IPv6"] deploys an IPv6-only NVA, which is not supported yet.
+   */
+  @added(Versions.v2025_09_01)
+  addressFamily?: AddressFamily[];
```
sandipsh: 🔵 **[NEW] Suggestion — add `@uniqueItems` and `@maxItems(2)` to `addressFamily`.**

`addressFamily?: AddressFamily[]` currently accepts payloads like `["IPv4","IPv4"]` or arbitrarily long arrays. The description already implies only `[]`, `["IPv4"]`, `["IPv4","IPv6"]`, or `["IPv6"]` are meaningful, so making that enforceable at the OpenAPI layer prevents nonsense payloads reaching the service and gives SDKs a cleaner contract.

```typespec
@added(Versions.v2025_09_01)
@uniqueItems
@maxItems(2)
addressFamily?: AddressFamily[];
```

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -15966,11 +15991,24 @@ model NetworkVirtualAppliancePropertiesFormat {
   @maxItems(3)
   nvaInterfaceConfigurations?: NvaInterfaceConfigurationsProperties[];
 
+  /**
+   * The address families to deploy the NVA in. ["IPv4", "IPv6"] deploys a dual-stack NVA (both IPv4 and IPv6 configurations on the NICs; the vHub/VNet must be dual-stack). ["IPv4"], an empty array, or omitting the field deploys an IPv4-only NVA. ["IPv6"] deploys an IPv6-only NVA, which is not supported yet.
```
sandipsh: 🔵 **[NEW] Suggestion — remove the "not supported yet" contradiction from this description.**

`["IPv6"] deploys an IPv6-only NVA, which is not supported yet.` says the value is both accepted (present in the enum, allowed in the array) and unsupported. A customer reading only the schema will code the unsupported call path and hit a runtime error.

**Recommended rewrite:**

> The address families to deploy the NVA in. `["IPv4", "IPv6"]` deploys a dual-stack NVA (the vHub/VNet must also be dual-stack). `["IPv4"]`, an empty array, or omitting the field deploys an IPv4-only NVA. The value `"IPv6"` may only appear in combination with `"IPv4"`; standalone `["IPv6"]` is reserved for future use and is rejected by the service today.

This keeps the enum forward-compatible while removing the contract ambiguity.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -15966,11 +15991,24 @@ model NetworkVirtualAppliancePropertiesFormat {
...
    */
   @visibility(Lifecycle.Read)
   privateIpAddress?: string;
+
+  /**
+   * A Internal Load Balancer's HA port frontend IPv6 address. Can be used to set routes & UDR to load balance traffic between NVA instances. This field appears in dual-stack NVAs.
+   */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2025_09_01)
+  privateIpAddressV6?: string;
```
sandipsh, line 15995: 🔵 **[NEW] Nit — grammar: `A Internal Load Balancer's` → `An Internal Load Balancer's`.**

("A" before a vowel sound.) The V4 sibling `privateIpAddress` on the same model has the identical typo but predates this PR, so flagging only this newly-added V6 line.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -16025,6 +16063,20 @@ model VirtualApplianceNicProperties {
   @visibility(Lifecycle.Read)
   privateIpAddress?: string;
 
+  /**
+   * Public IPv6 address. These fields appear in dual stack NVAs. It will also appear for Additional Nic configurations in case of Dual stack NVAs.
+   */
+  @visibility(Lifecycle.Read)
+  @added(Versions.v2025_09_01)
+  publicIpAddressV6?: string;
```
sandipsh, line 16055: 🔵 **[NEW] Nit — description grammar/consistency on the two new V6 NIC fields.**

Both `publicIpAddressV6` and `privateIpAddressV6` (line 16078) here have the description:

> `Public/Private IPv6 address. These fields appear in dual stack NVAs. It will also appear for Additional Nic configurations in case of Dual stack NVAs.`

Three issues:

- Plural "These fields" for a single field.
- Subject-verb disagreement ("It will also appear" following "These fields").
- Mixed casing "dual stack" / "Dual stack" (repo convention is `dual-stack`).

**Suggested rewrite** (apply to both fields, changing Public/Private accordingly):

> Public IPv6 address. Populated for dual-stack NVAs, including on additional-NIC configurations when the NVA is dual-stack.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [`a855dbe`](https://github.com/Azure/azure-rest-api-specs/pull/44811/commits/a855dbe7e96c09f6c604e681ec85952315909a73). See inline comments for findings 1-5.

specification/network/resource-manager/Microsoft.Network/Network/stable/2025-09-01/common.json
debipdasfeb14: Fixed the issue.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
debipdasfeb14: Fixed the issue.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
debipdasfeb14: Fixed the issue.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
debipdasfeb14: Fixed the issue.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
debipdasfeb14: Fixed the issue.

PR #44825, opened 2026-07-21
Fix Prepared Image Specification SDK surfaces

aboodasfari:
## Summary
- map the prepared image specification ETag to the .NET Azure.ETag type
- add descriptive .NET enum names
- add a descriptive Python management client name

## Validation
- TypeSpec validation
- `tsp compile .`


aboodasfari:
The JavaScript validation failure is in the JS release automation, not this `client.tsp` change. Generation/build succeeds, but the tool rewrites `package.json` after `pnpm install`; `verifyDepsBeforeRun: error` then rejects the stale workspace state during samples/pack. The rerun reproduced the same failure in build 6610228.

Upstream fix: Azure/azure-sdk-for-js#39390. After that tool is published and consumed by the validation pipeline, rerun `SDK Validation - JS`.

matthchr (approved):

PR #44826, opened 2026-07-21
Add Foundry state-stores data-plane TypeSpec contract

shanmukha1200:
## Summary

Adds the Foundry State Stores data-plane TypeSpec contract and wires it into the Azure AI Projects SDK emitters.

- exposes one `state_stores` operation group for store and item operations
- relocates Python/JavaScript operations to `AIProjectClient.beta.state_stores`
- includes State Stores in the Python/JavaScript, .NET, and Java Projects client entrypoints
- preserves raw public store names and item keys while documenting required unpadded-base64url path encoding through `x-ms-foundry-path-encoding`
- documents `tags.<name>` query serialization through `x-ms-foundry-query-prefix`
- removes caller-supplied user identity; hosted AgentServer integration supplies `x-agent-foundry-call-id` through request context
- aligns create/update status codes and ETag headers with the wire contract

## SDK layering

The intended layering is:

`TypeSpec -> generated AIProjectClient.beta.state_stores -> handwritten SDK patch/bound Store wrapper -> Agent Framework adapters`

AgentServer owns request-context propagation only; it should consume the Foundry SDK rather than maintain a second REST client.

## Validation

- Foundry TypeSpec project compilation
- repository TypeSpec validation
- Python/JavaScript, .NET, and Java SDK client entrypoint compilation with warnings treated as errors
- regenerated `v1` and `virtual-public-preview` OpenAPI artifacts

## Related work

- Experience proposal: coreai-microsoft/foundrysdk_specs#247
- Existing Python hosted-agent implementation: Azure/azure-sdk-for-python#47763


shanmukha1200:
All TypeSpec and SDK-facing validation is now green:

- TypeSpec Validation and TypeSpec APIView
- SDK Validation for .NET, Go, JavaScript, Java, Python, and Rust
- SpellCheck and Swagger PrettierCheck
- cross-version breaking-change analysis

The remaining failures are the legacy Swagger checks running against this OpenAPI 3 TypeSpec project:

- Avocado performs a case-sensitive lookup for lowercase `readme.md`, while Foundry uses `README.md`.
- Swagger BreakingChange invokes OpenAPI 2/AutoRest `--v2` handling for emitted OpenAPI 3 documents.
- Swagger LintDiff does not locate the OpenAPI 3 files in its Swagger spec model.
- Swagger ModelValidation and SemanticValidation apply OpenAPI 2-era validation to the OpenAPI 3 outputs.

These failures are not reporting State Store contract or SDK generation errors. Repository-owner guidance or suppression is needed for these legacy checks before automated merge requirements can pass.

shanmukha1200:
Updated in ebe021fe2c to remove the `/storage/` segment and align the contract with the endpoint-scoped spec: `/agents/{agent_name}/endpoint/state_stores...`. Regenerated both v1 and virtual-public-preview OpenAPI JSON/YAML outputs.

shanmukha1200:
TypeSpec Validation and all SDK validations now pass after formatting and regenerating the OpenAPI 3.2 artifacts.

The five remaining failures are legacy Swagger 2.0 validators processing Foundry's `openapi3/` 3.2 artifacts:

- `Swagger Avocado`: reports `MISSING_README`
- `Swagger LintDiff`: does not include the generated OAS3 files in its Swagger model
- `Swagger SemanticValidation`: requires `swagger: 2.0`
- `Swagger ModelValidation`: applies Swagger 2 schemas to OAS3 headers
- `Swagger BreakingChange`: invokes AutoRest with `--v2` and rejects `openapi: 3.2.0`

This is the same validator incompatibility seen on merged Foundry OAS3 PR #44848. Omitting the generated artifacts is not viable because TypeSpec Validation then fails on dirty generated output.

Could a repo maintainer review and apply the established overrides for this known tooling mismatch?

- `Approved-Avocado`
- `Approved-LintDiff`
- `Approved-ModelValidation`
- `Approved-SemanticValidation`
- `BreakingChange-Approved-Benign`

shanmukha1200:
Friendly review ping for @trrwilson, @glecaros, @balapv, and @dargilco: the contract-side TypeSpec and SDK validations are green. The PR is blocked only by the five documented legacy Swagger/OpenAPI 3.2 mismatches above, which require maintainer override labels, plus reviewer approval.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+// ---------------------------------------------------------------------------
+// 4.1 State Store resource
+// ---------------------------------------------------------------------------
+
+/** A first-class, explicitly-created durable key-value store (spec §2, §4.1). */
+@doc("A first-class, explicitly-created durable key-value store.")
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+model StateStore {
+  @doc("The object type, which is always 'state_store'.")
+  object: StateStoreObjectType.stateStore;
```
glecaros: these are generally discouraged, we don't require them for object discrimination.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+/** The `object` discriminator for a state-store resource. */
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+union StateStoreObjectType {
+  stateStore: "state_store",
+}
+
+/** The `object` discriminator for a state-store item resource. */
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+union StateStoreItemObjectType {
+  stateStoreItem: "state_store.item",
```
glecaros: not needed based on the comment below.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+// ---------------------------------------------------------------------------
+// 4.2 Item resource
+// ---------------------------------------------------------------------------
+
+/** A single `key -> value` record within a state store (spec §2, §4.2). */
+@doc("A single key -> value record within a state store.")
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+model StateStoreItem {
+  @doc("The object type, which is always 'state_store.item'.")
+  object: StateStoreItemObjectType.stateStoreItem;
```
glecaros: as the ones above, this should be removed.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+  @doc("The Unix timestamp (seconds) when the store metadata was last updated.")
+  updated_at: FoundryTimestamp;
+}
+
+/** Response for `deleteStateStore` (spec §4.1.5). Idempotent. */
+@doc("Response for deleting a state store.")
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+model DeletedStateStore {
+  @doc("The object type, which is always 'state_store'.")
+  object: StateStoreObjectType.stateStore;
```
glecaros: same as above, not needed,

specification/ai-foundry/data-plane/Foundry/src/state-stores/routes.tsp
```diff
@@ -0,0 +1,199 @@
...
+  @extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+  deleteStateStore is FoundryDataPlaneRequiredPreviewOperation<
+    FoundryFeaturesOptInKeys.state_stores_v1_preview,
+    {
+      /** The name of the agent whose endpoint owns the state store. */
+      @path agent_name: string;
+
+      ...DeleteStateStoreRequest;
+    },
+    DeletedStateStore
```
glecaros, line 104: do we expect the customer to use the headstone, or would it be enough to return 204?

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+
+/**
+ * Response for `createItem` / `putItem` (spec §4.2.1/§4.2.2) — server metadata
+ * only, no `value` echoed back.
+ */
+@doc("Server metadata for a written state-store item (no value echoed back).")
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+model StateStoreItemRef {
+  @doc("The object type, which is always 'state_store.item'.")
+  object: StateStoreItemObjectType.stateStoreItem;
```
glecaros: all of the object props are not really required

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,389 @@
...
+  key: string;
+
+  @doc("Whether the item was deleted. Always `true`.")
+  deleted: boolean;
+}
+
+/** A keys-only projection entry returned by `listKeys` (spec §4.2.5) — never includes `value`. */
+@doc("A keys-only projection of a state-store item (never includes the value).")
+@extension("x-ms-foundry-meta", StateStoreRequiredPreviews)
+model StateStoreItemKey {
```
glecaros: aside from tags, this model seems to be identical to StateStoreItemRef.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/routes.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

shanmukha1200:
Addressed all review comments in the squashed commit `0a131c8206`:

- removed redundant State Store/item `object` discriminators
- changed store and item deletes to idempotent `204 No Content`
- removed delete tombstone models
- reused `StateStoreItemRef` for keys-only entries via model spread
- regenerated the OpenAPI 3.2 artifacts

The coordinated provider contract update is pushed to coreai-microsoft/foundrysdk_specs#247. @glecaros, please re-review when convenient.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
glecaros: ???

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,314 @@
...
+  @doc("Optional labels associated with the item.")
+  tags?: Record<string>;
+}
+
+// ---------------------------------------------------------------------------
+// Request envelopes (path + query + header + body), one alias per operation.
+// Non-@path/@query/@header fields form the JSON request body.
+// ---------------------------------------------------------------------------
+
+alias CreateStateStoreRequest = {
```
glecaros: this is not needed as long as the StateStore model is annotated  with @visibility(Lifecycle.Read) for the id, created_at,  and updated_at  properties.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
```diff
@@ -0,0 +1,314 @@
...
+alias GetStateStoreRequest = {
+  @doc("The logical name of the state store. SDKs accept the unencoded name and apply base64url encoding without padding to the path segment.")
+  @path
+  @extension("x-ms-foundry-path-encoding", "base64url")
+  @minLength(1)
+  @maxLength(128)
+  name: string;
+};
+
+alias UpdateStateStoreRequest = {
```
glecaros, line 158: same as the above, we could depend on the lifecycle annotations to make this redundant

glecaros: the items models should be looked for opportunities like this.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: System.Collections.Hashtable.body

shanmukha1200: System.Collections.Hashtable.body

shanmukha1200:
@glecaros Apologies for the responses on the comments it was just a experimental code review agent misbehaving :)

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: I got this reference from the memory store actually https://ai.azure.com/api-reference/memory-stores/create-memory-store/ 
added it just to maintain consistency

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: Referenced other APIs added them to maintain consistency

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: what do you recommend have no object type discriminators at all?

specification/ai-foundry/data-plane/Foundry/src/state-stores/routes.tsp
shanmukha1200: we want to support batch apis as a fast follow up so we would want to only return deleted items as response in that case we thought to keep it consistent for this API as well additionally if at all they want to reference the id of the deleted item I thought lets just give this object back

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
shanmukha1200: yeah that's right Key should return the tags , but Ref might not necessarily need it

specification/ai-foundry/data-plane/Foundry/src/state-stores/routes.tsp
shanmukha1200: Confirmed with the product owner that clients need the delete tombstone. Restored `200 OK` with `DeletedStateStore` / `DeletedStateStoreItem` (`id` when present, addressed `name` or `key`, and `deleted: true`). The redundant `object` discriminator remains removed.

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
```diff
@@ -283,6 +285,16 @@ using Azure.AI.Projects;
...
+@@clientName(StateStores.createStateStore, "create");
+@@clientName(StateStores.listStateStores, "list");
+@@clientName(StateStores.getStateStore, "get");
+@@clientName(StateStores.updateStateStore, "update");
+@@clientName(StateStores.deleteStateStore, "delete");
+@@clientName(StateStores.createStateStoreItem, "createItem");
+@@clientName(StateStores.listStateStoreItemKeys, "listKeys");
+@@clientName(StateStores.putStateStoreItem, "putItem");
+@@clientName(StateStores.getStateStoreItem, "getItem");
+@@clientName(StateStores.deleteStateStoreItem, "deleteItem");
```
dargilco: Please move these lines towared the end of the file, and put this comment above it:

```
// Beta Memories sub-client
```

specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-projects/client.tsp
```diff
@@ -38,6 +39,17 @@ namespace Azure.AI.Projects;
...
+@@clientName(StateStores.listStateStores, "list");
+@@clientName(StateStores.getStateStore, "get");
+@@clientName(StateStores.updateStateStore, "update");
+@@clientName(StateStores.deleteStateStore, "delete");
+@@clientName(StateStores.createStateStoreItem, "createItem");
+@@clientName(StateStores.listStateStoreItemKeys, "listKeys");
+@@clientName(StateStores.putStateStoreItem, "putItem");
+@@clientName(StateStores.getStateStoreItem, "getItem");
+@@clientName(StateStores.deleteStateStoreItem, "deleteItem");
+
```
dargilco: Remove all thse @@clientName lines... you can see that for Java, we use longer names that include the operation group (for Python and JS we use shorter names).

I think the only change needed in this file is the added import at the top

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
dargilco: This is a pattern used by OpenAI. We don't need to follow that since we don't see the benefit. I see you already removed it. Thanks!

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
shanmukha1200: Moved the State Stores Python/JS client-name customizations toward the end of the file under a dedicated `Beta State Stores sub-client` section.

specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-projects/client.tsp
shanmukha1200: Removed all State Stores `@@clientName` overrides from the Java client. The only State Stores change remaining in this file is the routes import.

specification/ai-foundry/data-plane/Foundry/src/state-stores/models.tsp
dargilco: I see this union was removed. Resolving.

dargilco:
Please target this PR to branch feature/foundry-release. That's where we stage all Foundry TypeSpec changes, before merge to Main.

shanmukha1200:
@dargilco Retargeted this PR to `feature/foundry-release` and rebased the single State Stores commit onto the current staging branch. The Foundry project and all SDK TypeSpec entrypoints compile cleanly with warnings treated as errors. Please re-review when convenient.

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
```diff
@@ -476,6 +478,21 @@ using Azure.AI.Projects;
...
+// Beta State Stores sub-client
+// --------------------------------------------------------------------------------
+
+@@clientName(StateStores.createStateStore, "create");
+@@clientName(StateStores.listStateStores, "list");
+@@clientName(StateStores.getStateStore, "get");
+@@clientName(StateStores.updateStateStore, "update");
+@@clientName(StateStores.deleteStateStore, "delete");
+@@clientName(StateStores.createStateStoreItem, "createItem");
+@@clientName(StateStores.listStateStoreItemKeys, "listKeys");
```
dargilco: "listItemKeys" (instead of "listKeys")

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
```diff
@@ -476,6 +478,21 @@ using Azure.AI.Projects;
...
+@@clientName(StateStores.putStateStoreItem, "setItem");
```
dargilco: "createItem" (instead of "setItem")

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
```diff
@@ -0,0 +1,55 @@
+using Azure.ClientGenerator.Core;
```
dargilco: please only make changes related to Python and JS. We'll deal with .NET and Java, as there is other special consideration for them. So no need to create a new file. What's in this file should go into the existing file `ops-relocation-projects.tsp`

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    interface Store {}
+  }
+
+  @@clientLocation(Azure.AI.Projects.StateStores.createStateStore, StateStores);
+  @@clientLocation(Azure.AI.Projects.StateStores.listStateStores, StateStores);
+  @@clientLocation(
+    Azure.AI.Projects.StateStores.getStateStore,
+    StateStores.Store
```
dargilco: `StateStores` instead of `StateStores.Store` here and below. We only have one-level sub-clients (or two levels, if you count the ".beta" as a level)

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
```diff
@@ -0,0 +1,55 @@
...
+namespace Azure.AI.Projects.Beta {
+  namespace StateStores {
+    model StoreClientOptions {
+      /** The logical name of the state store. */
+      name?: string;
+    }
+
+    interface Store {}
+  }
+
```
dargilco: I'll need to discuss the above with Kashif, our Python architect. For first preview release, I suggest just keeping all operations on the ".beta.statestores" sub-client, and passing store name to every operation. So the above "StoreClientOptions" is not needed.

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
```diff
@@ -0,0 +1,55 @@
...
+    StateStores.Store
+  );
+
+  @@clientInitialization(
+    StateStores.Store,
+    {
+      parameters: StateStores.StoreClientOptions,
+      initializedBy: InitializedBy.customizeCode,
+    }
+  );
```
dargilco: clientInitialization not needed for now.

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-projects/client.tsp
```diff
@@ -12,6 +12,7 @@ import "../evaluators/routes.tsp";
 import "../insights/routes.tsp";
 import "../schedules/routes.tsp";
```
dargilco, line 13: Please don't edit this file

specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-projects/client.tsp
```diff
@@ -14,6 +14,7 @@ import "../evaluators/routes.tsp";
 import "../insights/routes.tsp";
 import "../schedules/routes.tsp";
 import "../routines/routes.tsp";
```
dargilco, line 16: Please don't edit files in this folder

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
shanmukha1200: Updated the Python/JS name to `listItemKeys` in commit 714f3c07f0.

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
shanmukha1200: The POST operation is already named `createItem`. Renaming the PUT operation to the same name produces `duplicate-client-name` errors for both Python and JavaScript. I kept PUT as `setItem`; please let me know if you want a different name for either operation.

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
shanmukha1200: Moved the flat State Stores relocation rules into the existing `ops-relocation-projects.tsp` and deleted `ops-relocation-state-stores.tsp` in commit 714f3c07f0.

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
shanmukha1200: Flattened all operations directly onto `StateStores`; there is no nested `StateStores.Store` client now.

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
shanmukha1200: Removed `StoreClientOptions`. For this preview, every store-specific operation takes the store `name` parameter directly.

specification/ai-foundry/data-plane/Foundry/src/sdk-common/ops-relocation-state-stores.tsp
shanmukha1200: Removed the State Store `@@clientInitialization` customization.

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-projects/client.tsp
shanmukha1200: Removed the State Stores import and all State Store name customizations from the C# client entrypoint.

specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-projects/client.tsp
shanmukha1200: No additional Java customization was added in this update. The existing route import remains necessary because Java imports the shared `ops-relocation-projects.tsp`; removing it causes `compile_all.ps1` to fail with unresolved `StateStores` references.

specification/ai-foundry/data-plane/Foundry/src/sdk-python-js-azure-ai-projects/client.tsp
dargilco: Oh sorry, I missed that. Python has a common pattern of "create_or_update" methods (or "createOrUpdate" as it will show up in TypeSpec), so I guess instead of "setItem" a more common name would be "create_or_update_item".

Do we really need both "create_item" and "create_or_update_item"? Can we just have "create_or_update_item"?

dargilco: I think this is common:
POST operation ==> "create" in Python
PATCH operations ==> "update" in Python
PUT operations ==> "create_or_update" in Python

shanmukha1200: Updated in commit `482cc52119`. We need both operations because they now have distinct semantics: POST `createItem` creates and returns 409 when the key exists; PUT `updateItem` updates only `value` and `tags` and returns 404 when the key does not exist. The PUT operation no longer upserts and no longer returns 201.

shanmukha1200: changed it to update it only updates value and tags

Nathandrake229:
The .NET APIView for `Azure.AI.AgentServer.Core.Storage` requires optimistic-concurrency values to use the Azure SDK primitive `Azure.ETag` rather than `string`.

Before this TypeSpec contract becomes the source for future generation, please change the state-store ETag types to `Azure.Core.eTag`, including:

- `StateStoreItem.etag` (which also flows into `StateStoreItemRef` and `StateStoreItemKey`)
- `UpdateItemRequest.if_match`
- `DeleteItemRequest.if_match`
- ETag response/error header aliases

The current .NET SDK fix has been generated with `Azure.Core.eTag`; keeping the spec as `string` would regress that API shape when this PR is adopted.

APIView: https://spa.apiview.dev/review/b00e4b0412f942ed92524b2dd0e83b89

Nathandrake229:
Updated the state-store ETag source of truth and .NET generation path:

- Changed the state-store ETag properties and conditional headers to `Azure.Core.eTag`; regenerated both OpenAPI versions.
- Added the model-only C# projection at `src/sdk-csharp-azure-ai-agentserver-core-storage` for the hand-written AgentServer Core storage client.
- Configured the Azure C# emitter so `Azure.Core.eTag` generates `Azure.ETag`.
- TypeSpec validation passes for both the Foundry project and the new C# projection.

The Azure SDK pin now references commit `d9483caa16599e4d73ea3db4ba74b8663fc6af3b`.

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-agentserver-core-storage/client.tsp
```diff
@@ -0,0 +1,304 @@
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-client-generator-core";
+
+using Azure.ClientGenerator.Core;
+
+namespace Azure.AI.AgentServer.Core.Storage;
+
+/** Unix epoch seconds used by state-store timestamps. */
+scalar FoundryTimestamp extends int64;
```
glecaros, line 9: there's already a foundry timestamp defined in servicepatterns.tsp (that is an actual datetime wire encoded to unix seconds)

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-agentserver-core-storage/client.tsp
```diff
@@ -0,0 +1,304 @@
...
+
+  /** Optional string-to-string labels. */
+  tags?: Record<string>;
+
+  /** Creation time in Unix epoch seconds. */
+  @encodedName("application/json", "created_at")
+  createdAt: FoundryTimestamp;
+
+  /** Last metadata-update time in Unix epoch seconds. */
+  @encodedName("application/json", "updated_at")
```
glecaros, line 59: don't use encodedName, our guidelines are snake so use that for the name directly.

specification/ai-foundry/data-plane/Foundry/src/state-stores/routes.tsp
```diff
@@ -0,0 +1,199 @@
...
+    CreateStateStoreItemResponse
+  >;
+
+  /** Lists item keys (keys-only projection), ordered by `(created_at, id)`; never returns `value`. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
+  @summary("List state-store item keys")
+  @operationId("listStateStoreItemKeys")
+  @get
+  @list
+  @route("/agents/{agent_name}/endpoint/state_stores/{name}/items:keys")
```
glecaros, line 131: the `:` separator is an action separator... with keys being a noun, it does not really fit... to keep the flat-ish schema this should be called `/item_keys`

PR #44846, opened 2026-07-21
[Authorization] Add AgentUser/AgentServicePrincipal to RBAC roleAssignment principalType on stable 2022-04-01

jruttle:
## Summary

Adds the two agent identity principal types to `RoleAssignment.principalType` in-place on the existing stable api-version `2022-04-01`:

- `AgentUser` - agent identity derived from a User
- `AgentServicePrincipal` - agent identity derived from a ServicePrincipal

`principalType` is already an extensible enum (`x-ms-enum.modelAsString: true`). This is the selected shipping path instead of publishing a new api-version.

## Breaking-change disposition

Adding enum values in-place is flagged as `AddedEnumValue` (oad rule 1020) on the request and response contract. The Azure Breaking Changes Review Board approved this change on 2026-07-22, and the protected labels `BreakingChange-Approved-BugFix` and `BreakingChange-Approved-Security` are applied.

## Rationale

- Customers assigning Azure RBAC roles to Microsoft Entra Agent identities are failing on the stable `2022-04-01` surface they already use.
- Existing generated SDKs tolerate unknown values on read because the enum is extensible.
- The change unblocks the security-critical Azure AD Graph to Microsoft Graph migration, which surfaces the explicit agent identity types.
- The paired PAS service change is DA PR !16528382.

## Release acknowledgement

This PR intentionally targets `main` and publishes the updated `2022-04-01` contract to customers. The `PublishToCustomers` label acknowledges that the API is considered shipped after merge and that the ARM API best practices apply.

## Remaining

- [x] Breaking-change approval
- [x] Select the in-place shipping path
- [x] Add `PublishToCustomers`
- [ ] Required repository review

### DIFF
--- a/specification/authorization/resource-manager/Microsoft.Authorization/Authorization/RoleAssignment/models.tsp
+++ b/specification/authorization/resource-manager/Microsoft.Authorization/Authorization/RoleAssignment/models.tsp
@@ -140,6 +140,8 @@ model RoleAssignmentFilter {
   principalId?: string;
 }
 
+// Java and Python retain the existing shared SDK PrincipalType through the
+// combined client's language-scoped alternate-type customization.
 /**
  * The principal type of the assigned principal ID.
  */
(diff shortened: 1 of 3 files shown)


denyspanov-public (approved):
LGTM!

PR #44847, opened 2026-07-21
Update example file to redact the example password field

hnsunil:
Update example file to redact the example password field.
Note: None of the SDKs are released and, in this PR, just REDACTING the example password in example file.
SDKs will be generated once this PR is merged.


audunn (approved):
LGTM

PR #44851, opened 2026-07-21
Add typespec-suppressions tool for TypeSpec suppressions review process

catalinaperalta:
## Summary

Adds the `typespec-suppressions` tool under `eng/tools` for reviewing TypeSpec suppressions in PRs.

The tool diffs a base and head revision of changed TypeSpec projects, extracts inline (`#suppress`) and `tspconfig.yaml` suppressions, classifies them as new / changed / removed, enriches them with Azure rule documentation metadata, and renders a JSON report plus a Markdown summary. A curated `check-rules.json` scopes which rules require review; an optional `--github-annotations` flag emits inline PR warning annotations.

## What's included

- `eng/tools/typespec-suppressions/**` — CLI + programmatic API, unit/e2e tests, README, `check-rules.json`
- `eng/tools/package.json` + `eng/tools/tsconfig.json` — register the new package in the eng/tools workspace

The package follows the existing eng/tools **run-from-source** convention (native TypeScript, `tsc --noEmit`, vitest), matching `summarize-impact`.

## Notes

- This is the first of two stacked PRs. A follow-up PR adds the CI workflows and the summarize-checks / summarize-impact integrations that consume this tool.
- Opened as a **draft**.

> Co-authored with GitHub Copilot.


eng/tools/typespec-suppressions/cmd/typespec-suppressions.js
timotheeguerin, line 1: double check this has 755 permission with git otherwise on linux/osx when you do npm instaall it change the file creating git diff(not sure if it is set correctly i don't know how to see in github ui)

eng/tools/typespec-suppressions/src/extract.ts
```diff
@@ -0,0 +1,268 @@
+import { createSourceFile } from "@typespec/compiler";
```
markcowl, line 1: We should have an issue for moving these apis into core (if we don't already)

markcowl (approved):
This looks great!  I think we want to identify which parts of this should go in typespec core and create an issue for that (I indicated one place).

Added new aka links for the two over view docs:

https://aka.ms/tsp-suppress/lintdiff-to-tsp links to the lintdiff -> tsp map,  
https://aka.ms/tsp-suppress/tsp-to-lintdiff has the corresponding typespec rule overview and map to lintdiff

I would introduce these with something like this at the end:

For an overview of TypeSpec linting rules [click here](https://aka.ms/tsp-suppress/tsp-to-lintdiff )
For a mapping from ARM lintdiff rules to corresponding TypeSpec linting rules [click here](https://aka.ms/tsp-suppress/lintdiff-to-tsp)

timotheeguerin (approved):

eng/tools/typespec-suppressions/cmd/typespec-suppressions.js
mikeharder: if you click "..." then "view file", I think this shows it is `executable file` in the header:

Image

eng/tools/typespec-suppressions/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "scripts": {
+    "build": "tsc --noEmit",
+    "format": "prettier . --ignore-path ../.prettierignore --write",
+    "format:check": "prettier . --ignore-path ../.prettierignore --check",
+    "format:check:ci": "prettier . --ignore-path ../.prettierignore --check --log-level debug",
+    "test": "vitest --run",
+    "test:ci": "vitest run --coverage --reporter=verbose"
+  },
+  "dependencies": {
+    "@typespec/compiler": "1.14.0",
```
mikeharder, line 18: There may be no easy fix until we convert this repo to pnpm (hopefully coming in a few weeks).  But this duplicates the compiler version already set in repo root, so we'll need to remember to update this when we update the root package.json.

eng/tools/typespec-suppressions/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@azure-tools/typespec-suppressions",
+  "private": true,
+  "type": "module",
+  "main": "src/index.ts",
+  "bin": {
+    "typespec-suppressions": "cmd/typespec-suppressions.js"
+  },
+  "scripts": {
```
mikeharder, line 9: I recommend copying a few more scripts/deps:

- Add script "check" to run "build+lint+format+test"
- Add eslint

You can copy from here (our simplest project under eng/tools):

https://github.com/Azure/azure-rest-api-specs/tree/main/eng/tools/suppressions

eng/tools/typespec-suppressions/cmd/typespec-suppressions.js
timotheeguerin: oh didn't know that, I also tried in the codespace and it seems fine.

eng/tools/typespec-suppressions/src/path-utils.ts
mikeharder, line 1: I don't think we should need `normalizeSlashes()` or `normalizeRepoPath()`.  Built-in functions like `resolve()` should be sufficient.  But I'm fine if you want to merge as-is, and we try to simplify your path handling in a later PR.  Cross-platform path handling was one of the trickier parts of tools in this repo (especially because some libs like simple-git return posix paths even on windows).

eng/tools/typespec-suppressions/src/check-rules.ts
```diff
@@ -0,0 +1,65 @@
...
+ * (malformed JSON / wrong shape) instead of degrading to `[]`, to catch config
+ * typos. Current behavior intentionally degrades gracefully; reassess after the
+ * initial rollout.
+ */
+export function loadCheckRulesFile(checkRulesFilePath: string): string[] {
+  const resolvedPath = path.resolve(process.cwd(), checkRulesFilePath);
+
+  let content: string;
+  try {
+    content = readFileSync(resolvedPath, "utf-8");
```
mikeharder, line 34: we try to standardize on async fs operations

eng/tools/typespec-suppressions/src/rule-metadata.ts
```diff
@@ -0,0 +1,108 @@
...
+  url?: string;
+};
+
+type LinterModule = {
+  $linter?: {
+    rules?: RuleDefinition[];
+  };
+};
+
+const moduleCache = new Map<string, Promise<LinterModule | undefined>>();
```
mikeharder, line 15: We have shared cache helpers that might simplify your code a bit:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/cache.js
https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/README.md

eng/tools/typespec-suppressions/test/analyze.e2e.test.ts
```diff
@@ -0,0 +1,303 @@
...
+import os from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import { analyzeTypeSpecSuppressions } from "../src/index.ts";
+
+async function initTempRepo(): Promise<string> {
+  const repoRoot = await mkdtemp(path.join(os.tmpdir(), "typespec-suppressions-"));
+  execFileSync("git", ["init"], { cwd: repoRoot });
+  execFileSync("git", ["config", "user.name", "Copilot"], { cwd: repoRoot });
+  execFileSync("git", ["config", "user.email", "copilot@example.com"], { cwd: repoRoot });
```
mikeharder:
```suggestion
  execFileSync("git", ["config", "user.name", "Test"], { cwd: repoRoot });
  execFileSync("git", ["config", "user.email", "test@example.com"], { cwd: repoRoot });
```

Let's make it clear this is a test/example, not actual Copilot.

eng/tools/typespec-suppressions/test/analyze.e2e.test.ts
```diff
@@ -0,0 +1,303 @@
...
+  return repoRoot;
+}
+
+function git(repoRoot: string, args: string[]): string {
+  return execFileSync("git", ["-c", "commit.gpgsign=false", ...args], {
```
mikeharder, line 17: why not use simple-git?

eng/tools/typespec-suppressions/test/analyze.e2e.test.ts
```diff
@@ -0,0 +1,303 @@
+import { execFileSync } from "node:child_process";
```
mikeharder, line 1: Please use our shared helper:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/exec.js

mikeharder (approved):
Approved.  But a few areas we might want to improve in later PRs:

- Use more helpers under .github/shared
  - cache.js
  - changed-files.js?
  - exec.js
  - path.js
  - spec-model.js? (maybe need to add TSP files to spec-model)
- Align path handling with other tools
  - public APIs accept relative or absolute paths, in either POSIX or platform-specific format
  - immediately call `resolve()` on path params, to convert to an absolute platform-specific path
  - public APIs should always return absolute, platform-specific paths
  - internal APIs should assume all paths are absolute and platform-specific, so can omit calling `resolve()` if they want, but it's always safe to call `resolve()` again
  - if needed for perf, use our caching helper methods like `resolveCached()`

eng/tools/typespec-suppressions/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "build": "tsc --noEmit",
+    "format": "prettier . --ignore-path ../.prettierignore --write",
+    "format:check": "prettier . --ignore-path ../.prettierignore --check",
+    "format:check:ci": "prettier . --ignore-path ../.prettierignore --check --log-level debug",
+    "test": "vitest --run",
+    "test:ci": "vitest run --coverage --reporter=verbose"
```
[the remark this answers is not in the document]
catalinaperalta: this is coming up in the next pr

eng/tools/typespec-suppressions/README.md
```diff
@@ -0,0 +1,213 @@
...
+By default (no `--check-rules-file`), the tool reports **all** new/changed
+suppressions and `requiresApproval` reflects the full diff (**legacy mode**).
+
+When `--check-rules-file <path>` is provided, the tool runs in **checked-only
+mode**: it additionally emits a `checkedSuppressions` block scoped to the listed
+rules (exact, fully-qualified rule-name match), and the Markdown summary / gating
+focus on that subset. The full diff is still retained at the top level for other
+flows.
+
+The file is a JSON object:
```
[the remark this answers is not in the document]
catalinaperalta: added in the next pr

eng/tools/typespec-suppressions/test/analyze.e2e.test.ts
catalinaperalta: will be part of follow up code improvements

eng/tools/typespec-suppressions/src/check-rules.ts
catalinaperalta: Agree, I use async fs too in tsp-client, this code was created by copilot will include in future code improvement work

catalinaperalta:
Tracking issue to follow up on the unaddressed feedback in this pr: https://github.com/Azure/azure-rest-api-specs/issues/44909

markcowl (approved):

eng/tools/typespec-suppressions/src/extract.ts
```diff
@@ -0,0 +1,268 @@
...
+import { parse, SyntaxKind, visitChildren } from "@typespec/compiler/ast";
+import path from "node:path";
+import { isMap, LineCounter, parseDocument } from "yaml";
+import { normalizeRepoPath } from "./path-utils.ts";
+import type { SuppressionRecord } from "./types.ts";
+
+type AstNode = {
```
timotheeguerin, line 8: Why can't we use `Node` from `@typespec/compiler/ast`

eng/tools/typespec-suppressions/src/extract.ts
```diff
@@ -0,0 +1,268 @@
...
+export function extractInlineSuppressions(
+  specPath: string,
+  sourcePath: string,
+  text: string,
+): SuppressionRecord[] {
+  const sourceFile = createSourceFile(text, sourcePath);
+  const script = parse(sourceFile);
+  const suppressions: SuppressionRecord[] = [];
+
+  const walk = (node: AstNode, scope: string[]) => {
```
timotheeguerin, line 172: unfortunatelly we don't export yet but we should plan to and reuse the `createSUppressionTracker` function in compiler

eng/tools/typespec-suppressions/src/extract.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (
+      node.kind === SyntaxKind.TypeSpecScript &&
+      Array.isArray((node as { statements?: AstNode[] }).statements)
```
timotheeguerin, line 175: from the astnode comment above feels like a lot of the as casting here then can be removed

timotheeguerin (commented):
sorry didn't comment on that earlier only noticed from the follow up issue in typespec-azure that some stuff might not be right.

Also fine waiting for the suppression tracker to be available to replace it all anyway

eng/tools/typespec-suppressions/src/extract.ts
catalinaperalta: https://github.com/Azure/typespec-azure/issues/5030

PR #44865, opened 2026-07-21
kube selector groups

saisujithreddym:
# Choose a PR Template
Reference PR that is already reviewed and merged: https://github.com/Azure/azure-rest-api-specs-pr/pull/29011

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/Network/readme.md
+++ b/specification/network/resource-manager/Microsoft.Network/Network/readme.md
@@ -125,9 +125,7 @@ suppressions:
       - $.definitions.GetServiceGatewayServicesResult
   - code: ProvisioningStateMustBeReadOnly
     from: firewallPolicy.json
-    reason: >-
-      KSG provisioningState is emitted with readOnly as a sibling of $ref. This is valid for
-      AutoRest/ARM, but the linter does not resolve this pattern reliably.
+    reason: provisioningState is correctly marked readOnly in the referenced schema. The linter does not follow $ref chains to verify readOnly.
     where:
       - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].get.responses["200"].schema
       - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["200"].schema
--- a/specification/network/resource-manager/Microsoft.Network/Network/readme.md
+++ b/specification/network/resource-manager/Microsoft.Network/Network/readme.md
@@ -133,6 +133,13 @@ suppressions:
     reason: Not a standard azure resource.
     where:
       - $.definitions.GetServiceGatewayServicesResult
+  - code: ProvisioningStateMustBeReadOnly
+    from: firewallPolicy.json
+    reason: The emitted {$ref, readOnly true} shape matches all pre-existing FirewallPolicy child resources (e.g. FirewallPolicyRuleCollectionGroup) which reference Common.ProvisioningState. A Network-RP-wide TypeSpec correction is tracked separately.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].get.responses["200"].schema
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["200"].schema
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["201"].schema
   - code: AvoidAdditionalProperties
     from: firewallPolicy.json
     reason: >-
--- a/specification/network/resource-manager/Microsoft.Network/Network/readme.md
+++ b/specification/network/resource-manager/Microsoft.Network/Network/readme.md
@@ -123,6 +123,32 @@ suppressions:
     reason: Not a standard azure resource.
     where:
       - $.definitions.GetServiceGatewayServicesResult
+  - code: ProvisioningStateMustBeReadOnly
+    from: firewallPolicy.json
+    reason: >-
+      KSG provisioningState is emitted with readOnly as a sibling of $ref. This is valid for
+      AutoRest/ARM, but the linter does not resolve this pattern reliably.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].get.responses["200"].schema
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["200"].schema
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["201"].schema
+  - code: AvoidAdditionalProperties
+    from: firewallPolicy.json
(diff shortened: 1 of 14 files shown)


saisujithreddym:
Reference PR that is already reviewed and merged: https://github.com/Azure/azure-rest-api-specs-pr/pull/29011

saisujithreddym:
Re: BreakingChange(Cross-Version) — 1042 ChangedParameterOrder

This can't be resolved in-spec: typespec-autorest always emits the request body as the last parameter, so introducing the new optional query parameter on the Firewall Policy PUT unavoidably shifts the body's index. All other findings — the new kubeSelectorGroups paths and the kubeSelectorGroups / sourceKubeSelectorGroups properties — are legitimate additive changes for the new 2025-09-01 API version. Requesting breaking-change review approval.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
```diff
@@ -123,6 +123,32 @@ suppressions:
     reason: Not a standard azure resource.
     where:
       - $.definitions.GetServiceGatewayServicesResult
+  - code: ProvisioningStateMustBeReadOnly
```
sandipsh, line 177: **[NEW] 🔴 Blocking** **[[ProvisioningStateMustBeReadOnly](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#must-be-read-only)]** `specification/network/resource-manager/Microsoft.Network/Network/readme.md` - line 126 - This suppression is masking a real emitted `provisioningState` shape issue.

**Classification reasoning:** Introduced in this PR - the KSG-specific suppression is new in `2025-09-01`, and the emitted KSG schema was added in this version.

**Suggested fix:**

```json
"provisioningState": {
  "allOf": [
    { "$ref": "./common.json#/definitions/Common.ProvisioningState" }
  ],
  "readOnly": true,
  "description": "The provisioning state of the firewall policy Kubernetes selector group resource."
}
```

The current emitted shape is `{ "$ref": "./common.json#/definitions/Common.ProvisioningState", "readOnly": true }`, and `readOnly` next to `$ref` is ignored. Please fix the emitted shape first, then remove this suppression; otherwise a real mutability bug stays hidden.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration(s), converged) against commit [`69d562f`](https://github.com/Azure/azure-rest-api-specs/pull/44865/commits/69d562fbf35cf2c5ef6a74f3133844ce5b150fcc). See inline comments for findings 1.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
saisujithreddym: provisioningState is marked read-only in TypeSpec via the read visibility decorator, and it is correctly readOnly in the referenced schema. The emitter references the shared provisioning-state definition and places readOnly alongside the ref; the linter flags this only because it does not follow the ref chain to verify readOnly. This is honored by AutoRest, OAV, and SDK generation, so the property is genuinely read-only. Keeping the suppression, with its justification updated accordingly.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
```diff
@@ -123,6 +123,30 @@ suppressions:
...
+    from: firewallPolicy.json
+    reason: provisioningState is correctly marked readOnly in the referenced schema. The linter does not follow $ref chains to verify readOnly.
```
sandipsh: **[NEW] 🔴 Blocking** **[[ProvisioningStateMustBeReadOnly](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#must-be-read-only)]** `specification/network/resource-manager/Microsoft.Network/Network/readme.md` - line 128 - This updated suppression reason is still not accurate for the emitted contract.

**Classification reasoning:** Introduced in this PR iteration - the new head changed the suppression reason text at this line, but not the emitted KSG `provisioningState` shape.

**Suggested fix:**

```text
Either fix the emitted shape, or justify the suppression with the actual approved exception.
```

`Common.ProvisioningState` in `common.json` is just a string enum; it is not itself read-only. The emitted KSG property is still `{ "$ref": "./common.json#/definitions/Common.ProvisioningState", "readOnly": true }`, so saying the referenced schema is already readOnly is misleading. Please either fix the emitted shape first, or update the reason to describe the real exception you are relying on.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration(s), converged) against commit [`ecbb040`](https://github.com/Azure/azure-rest-api-specs/pull/44865/commits/ecbb04008a50b4db5b338d7671bf684e816f2973). See inline comments for findings 1.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
saisujithreddym: Removed the suppression to match other resources behavior.

sandipsh:
Thanks for working on this. However, looking at the current readme.md, the ProvisioningStateMustBeReadOnly suppression is still present for KubeSelectorGroup at:
- \$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].get.responses["200"].schema
- \$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["200"].schema
- \$.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/firewallPolicies/{firewallPolicyName}/kubeSelectorGroups/{kubeSelectorGroupName}"].put.responses["201"].schema

Can you confirm if the suppression removal is in progress or if there was an issue? The current head still has the KubeSelectorGroup entries under the ProvisioningStateMustBeReadOnly suppression in the firewallPolicy.json block of the readme.

sandipsh:
**🔴 BLOCKING: Parameter Order Changed (Breaking Change)**

The Breaking Change(Cross-Version) validation is reporting an **Error-level** violation:

**Code:** ChangedParameterOrder (Rule 1042)  
**Message:** "The order of parameter with Name: 'parameters', In: 'Body', was changed. Expected param at index '2' but instead found it at index '3'."

**Root Cause:** The new fcManagedSync query parameter was added at index 4, which pushed the body parameters from index 4 (old version) to index 5 (new version).

**Impact:** This is a breaking change for SDK code generation and client consumption. Parameter order is material in many SDK frameworks.

**Why It Happened:** When adding the new property to FirewallPolicy, the new query parameter was likely added to the TypeSpec in a way that shifts the body parameter order relative to the prior version.

**Action Required:**  
- Review the TypeSpec source for how fcManagedSync is being added
- Ensure query parameters maintain the same relative order as the previous version, with body parameters at the same index
- This likely requires adjusting how parameters are declared in the TypeSpec or ensuring parameter insertion order matches the prior version

See the Breaking Change workflow run for full details: https://github.com/Azure/azure-rest-api-specs/actions/runs/30021683698

This **must** be fixed before the PR can be merged.

saisujithreddym:
The 1042 ChangedParameterOrder is not introduced by this PR. This PR's diff does not add, remove, or modify afcManagedSync (or any other parameter) on FirewallPolicies_CreateOrUpdate — it only adds the KubeSelectorGroups paths to firewallPolicy.json.

The parameter-order shift comes from the afcManagedSync query parameter added in PR #44867 (AFC Managed Firewall Policy), which was merged into the 2025-09-01 release branch. The Breaking Change (Cross-Version) check compares the full 2025-09-01 spec against the prior stable 2025-07-01, so it re-reports that pre-existing shift on any PR that touches firewallPolicy.json — including this one.

The shift itself is unavoidable in the spec: typespec-autorest always emits the request body as the last parameter, so any new optional query parameter pushes the body's index. The same finding was already reviewed and accepted as valid on #44867. Requesting the same breaking-change sign-off be carried here.

saisujithreddym:
**Regarding ProvisioningStateMustBeReadOnly  supression:**

provisioningState is genuinely read-only — it's marked @visibility(Lifecycle.Read) in TypeSpec and emitted as readOnly: true next to the $ref to the shared Common.ProvisioningState. AutoRest, OAV, and SDK generation all honor it; the lintdiff rule just doesn't recognize readOnly beside a $ref (Azure/typespec-azure#4611), so the suppression is needed.

It only surfaces now because this is a diff linter — it flags new occurrences only. FirewallPolicyRuleCollectionGroup and others use the identical {$ref, readOnly:true} shape but predate the 2025-07-01 baseline, so they aren't re-flagged. KubeSelectorGroup is new in 2025-09-01, so the same shape gets flagged here.

This is the established pattern — the same suppression already exists in this readme for networkManager.json, interconnectGroup.json, expressRoute.json (ExpressRouteLag), and networkWatcher.json (connection analyzers). KubeSelectorGroup is consistent with these.

Also all the changes in this PR were already signed off as part of https://github.com/Azure/azure-rest-api-specs-pr/pull/29011 to the private repo. This PR just a copy to the public repo.

PR #44897, opened 2026-07-22
Add SDK emitters to nested Compute tspconfig

haagha:


sandipsh:
Action needed before merge: the current PR head is failing required validations after adding SDK emitters in tspconfig.

Please address these blockers:
- TypeSpec Validation
- Swagger Avocado - Analyze Code
- SDK Validation: .NET, Java, Python, JS

Most actionable starting point:
- Run: `npx tsv specification/compute/resource-manager/Microsoft.Compute/Compute/Compute`
- Review SDK validation logs for this tspconfig; Python annotations show compile/import failures in generated temp TypeSpec (for example missing `../common/models-ref.tsp` and `../common/operations.tsp`) leading to cascading unknown identifiers.
- Update emitter/output layout config so generation and validation pass, then re-run checks.

If these failures are pre-existing and not caused by this change, please add evidence (base vs head) so we can reclassify accordingly.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/tspconfig.yaml
```diff
@@ -8,6 +17,52 @@ options:
...
+  "@azure-typespec/http-client-csharp-mgmt":
+    namespace: "Azure.ResourceManager.Compute"
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+  "@azure-typespec/http-client-csharp-provisioning":
+    namespace: "Azure.Provisioning.Compute"
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+  "@azure-tools/typespec-go":
+    service-dir: "sdk/resourcemanager/compute"
+    emitter-output-dir: "{output-dir}/{service-dir}/armcompute"
+    module: "github.com/Azure/azure-sdk-for-go/{service-dir}/armcompute/v8"
```
ArthurMa1978, line 29: @tadelesh why does this module name has the `v8` suffix?

ArthurMa1978 (changes requested):
This change isn't needed because the Compute SDK is generated from `specification/compute/resource-manager/Microsoft.Compute/Compute`, which already includes the Compute, ComputeDisk, ComputeGallery, and ComputeSku subfolders.

PR #44922, opened 2026-07-23
Fix Private Traffic Manager example consistency

somya-0527:
Updates the Private Traffic Manager examples to:

- Use consistent `systemData` across resource responses.
- Use realistic paging and long-running operation URLs.
- Add missing `nextLink` values to paginated examples.
- Align parameter values with resource names and IDs.
- Fix topology map site and `catchAllSiteName` references.
- Use consistent PATCH operation definitions.
- Limit documented probing intervals to 10 and 30 seconds.


imsiddhartha (approved):
LGTM!

PR #44956, opened 2026-07-24
Fix TypeSpec for Compute cluster creation: add 202 LRO response and move location into properties

swarupecenits:
### Notes

- Fixes the Cognitive Services **Compute** (compute cluster) management TypeSpec so the async create/scale contract matches actual service behavior. Ref: [Task 5459506](https://msdata.visualstudio.com/Vienna/_workitems/edit/5459506).
- **Model create/scale as a `202`-only long-running operation.** Compute creation and scaling are async: the service returns **`202 Accepted`** on success and **`4xx`** on failure it never returns `200` or `201`. The PUT previously modeled `200/201`, so a real `202` was treated as an error by generated clients. `createOrUpdate` now returns **`202 + default` only**, via `ArmAcceptedLroResponse<..., ArmCombinedLroHeaders>`.
  - **Response headers:** `Azure-AsyncOperation` + `Location` (both returned by the service). **No `Retry-After`** the service does not set it.
  - **Generic LRO status body.** The operation is tracked via the standard Azure LRO status resource (provisioning state + percent-complete), **not** the `Compute` object, so **no `final-state-schema`** is emitted on the PUT.
- **Move `location` from the resource envelope into `properties`.** `location` was declared on the `Compute` ProxyResource envelope; per RP contract it belongs under `ComputeProperties`. Removed from the envelope and added to the base `ComputeProperties` model (inherited by all compute sub-types), so it serializes as `properties.location`.
- **Examples corrected** across both preview versions to match the real contract:
  - `location` placed under `properties` in all `Computes` bodies (`PutCompute`, `PutContainerInstanceCompute`, `GetCompute`, `GetContainerInstanceCompute`, `UpdateCompute`). The `UpdateCompute` PATCH request is unchanged since `location` is create/read-only.
  - Async-operation polling URL corrected to **`/computeOperations/`** (the segment the service actually uses).
  - Removed **`subnetArmId`** from `PutCompute` requests no clear networking flow, avoids a confusing example.
  - Removed **`Retry-After`** and the **`200`** response bodies (create is `202`-only).
- One ARM lint rule suppressed on `createOrUpdate` with an inline justification (approved):
  - `arm-put-operation-response-codes` the create/scale PUT intentionally returns only `202` (async), never `200/201`.

  _(The earlier `lro-location-header` suppression is no longer needed the service does return a `Location` header, which is now modeled.)_
  
### Testing
- `npx tsp compile .` → completes successfully, no warnings.
Screenshot 2026-07-24 114050

- Generated OpenAPI verified: PUT responses now `202, default`; `location` under `ComputeProperties.properties` and removed from the `Compute` envelope.
Screenshot 2026-07-24 114246

  _Verification: `location` is present under `ComputeProperties` (`True`) and no longer on the `Compute` envelope (`False`)._
Screenshot 2026-07-24 114434

*Regenerated OpenAPI (`2026-03-15-preview`): `createOrUpdate` is modeled as a **`202 Accepted`-only** response (with `Azure-AsyncOperation` + `Location` headers), and `location` is moved out of the `Compute` envelope into `ComputeProperties`.*

- `Computes` now exposes `get`, `put`, `delete` only — the unimplemented `Computes_Update` (PATCH) operation is removed.
- `VmPriority` enum is now `Regular, Spot` and `Pool.vmPriority` is optional, matching the backend contract.
- Source examples and generated output examples are in sync (regenerated on compile).

---

## Live Validation (Insomnia)

> **Live validation (api-version `2026-05-15-preview`):** `location` inside `properties` → **202 Accepted** ✅; `location` outside `properties` → **400 "The Location field is required"** ❌ - confirming the service contract now expects `location` within `properties`.
---
Screenshot 2026-07-24 150656
> **✅ Positive test — `location` sent *inside* **`properties`.**** `PUT .../computes/test-compute-insomnia-1` with `properties.location = "westcentralus"` and **no** envelope-level `location` returns **`202 Accepted`**. The `location` response header (async-operation polling URL) confirms the new **202 LRO** response added in this PR.

Screenshot 2026-07-24 150722
> **Same request - supporting headers.** `azureml-served-by-cluster: vienna-westcentralus-01` confirms the request was routed to the region supplied via `properties.location`, and `content-length: 0` confirms the standard async (LRO) 202 pattern.

Screenshot 2026-07-24 152312
> **❌ Negative test - `location` sent *outside* **`properties` (legacy envelope style).**** The same `PUT` with `location` at the resource envelope and omitted from `properties` is rejected synchronously with **`400 Bad Request -"The Location field is required."`** This proves the service reads `location` from `properties` only; the envelope copy is no longer honored - exactly what this TypeSpec change models.
---

# Description

Fixes the Compute cluster creation contract in the Cognitive Services management TypeSpec so async creation (`202`) is modeled correctly and `location` is placed under `properties` as the RP expects. This is a spec-only change in `azure-rest-api-specs`; downstream SDKs picking up these versions will regenerate with the corrected create responses and `location` placement.

Work item: [Task 5459506 - Fix TypeSpec for Compute Cluster Creation](https://msdata.visualstudio.com/Vienna/_workitems/edit/5459506)

# Versioning

Moving `location` from the envelope to `properties.location` is a wire-format change for the affected **preview** API versions, so both in-flight preview versions are updated together:
- `2026-03-15-preview`
- `2026-05-15-preview`

No GA/stable version exposes `location` on the `Compute` envelope, so no stable contract is broken. The `VersioningReviewRequired` label is applied for reviewer sign-off.

# Contribution checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-rest-api-specs/blob/main/CONTRIBUTING.md).
- [x] The pull request does not introduce breaking changes to any **stable** API version (preview-only change; see Versioning above).
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each with an informative message.
- [x] TypeSpec compiles cleanly and the generated OpenAPI + examples are regenerated and committed.


swarupecenits:
## Merge-requirements acknowledgment + request for versioning / ARM review

### Publish acknowledgment (main branch)
I intend to release these corrections to customers. I acknowledge that once merged these APIs are considered shipped, and further in-place changes are subject to Azure versioning and breaking-change policy.

For this control-plane surface I confirm [ARM API best practices](https://aka.ms/armapibestpractices) are followed: the `Compute` resource is a standard `ProxyResource` composed from standard ARM operation templates — `ArmResourceRead` (GET), `ArmResourceCreateOrReplaceAsync` (PUT), `ArmResourceDeleteWithoutOkAsync` (DELETE), `ArmResourceListByParent` (LIST), and `ArmResourceActionAsync` (start/stop/restart), with no spread-only models (discriminated types use `extends`). The `Computes_Update` (PATCH) operation has been intentionally removed because it is not implemented by the backend.

### Versioning / breaking change ([aka.ms/brch](https://aka.ms/brch)) This PR corrects the Compute create contract in the two existing **preview** api-versions `2026-03-15-preview` and `2026-05-15-preview`:

- **`location`** - changed from optional to **required** and moved into `properties` (was previously on the resource envelope).
- **Create PUT response codes** - the async create now returns **`202 Accepted` only** (with `Azure-AsyncOperation` + `Location` headers); the prior `200`/`201` modeling is removed, since the service returns 202 on success and 4xx on failure.
- **`Computes_Update` (PATCH)** - removed entirely; the backend does not implement it.
- **`VmPriority`** - enum value corrected `LowPriority` → `Spot`, and `Pool.vmPriority` made optional, to match the backend contract.
- **Suppression** - single `#suppress` for `arm-put-operation-response-codes` (202-only PUT),
  with justification. No `lro-location-header` suppression (Location header is returned).

These are corrections to an **incorrect pre-GA contract**. Both versions are PREVIEW (no compatibility guarantee) and **no GA version exposes this shape**, so there is no stable-surface customer impact. The `location` change triggers `Swagger BreakingChange` / `VersioningReviewRequired`. I'd like to **amend the previews in place** rather than spin a
new api-version, and am requesting breaking-change sign-off per [aka.ms/brch](https://aka.ms/brch). Happy to provide any further detail reviewers need.

### LintDiff (informational, non-blocking) The remaining by-design finding is **`PutResponseCodes`** (`arm-put-operation-response-codes`): Compute creation is a genuine LRO that returns **202-only** (no 200/201). It is already `#suppress`-ed in TypeSpec with justification. The PUT does return a `Location` header alongside `Azure-AsyncOperation`, so there is no outstanding `LroLocationHeader` finding. I can add a matching `readme.md` directive if a green LintDiff is preferred.

### Next step
Once versioning clears I'll comment `/azp run` to move this onto the ARM review queue.

saanikaguptamicrosoft (approved):

SubaMoorthy (approved):

saanikaguptamicrosoft (approved):

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [`dceb2e5`](https://github.com/Azure/azure-rest-api-specs/pull/44956/commits/dceb2e5c53114900684080e89c0210b3874c426b). See inline comments for findings 1-2. Findings 3-4 concern lines unchanged in this PR's diff and are posted as top-level PR comments rather than inline comments.

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
```diff
@@ -80,15 +73,15 @@ interface Computes {
   /**
    * Creates or updates a compute associated with the Cognitive Services account.
    */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "Compute create/scale is an async long-running operation: the service returns 202 Accepted on success and 4xx on failure. It never returns 200 or 201, so the response set intentionally includes only 202 to match the service contract."
```
ravimeda, line 76: **[NEW] 🔴 Blocking** **[[RPC-Put-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` - line 76 - The new suppression and response override make Compute PUT return only `202 Accepted`; ARM resource PUT must return `201` for create and `200` for replacement, including asynchronous PUT.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f`, line 84 used `ArmResourceCreateOrReplaceAsync<Compute>`, which emitted `200` and `201`; the head adds the response-code suppression and emits only `202` in both preview versions. The new `PutResponseCodes`, `ProvisioningStateSpecifiedForLROPut`, and `PutGetPatchResponseSchema` readme suppressions mask the resulting contract violations rather than making the wire behavior compliant.

**Suggested fix:** Restore the standard ARM template and align the RP to return the resource with `200`/`201` and a non-terminal `provisioningState`. Remove the TypeSpec suppression and the three new readme suppressions from both preview tags. If the service cannot comply, this needs an explicit ARM peer-review exception and a new API version rather than an in-place assertion that `202` is standard PUT behavior.

```tsp
@tag("Computes")
createOrUpdate is ArmResourceCreateOrReplaceAsync<Compute>;
```

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -5996,6 +5996,12 @@ model ComputeProperties {
    */
   computeType: ComputeType;
 
+  /**
+   * The location of the compute resource.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create)
+  location: string;
```
ravimeda, line 6003: **[NEW] 🔴 Blocking** **[[Section 13 - ARM Resource Model Requirements](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#13-arm-resource-model-requirements)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 6003 - `location` is newly declared inside the resource `properties` bag even though `location` is a reserved ARM envelope property name; reusing it for a domain field makes the wire shape ambiguous.

**Classification reasoning:** Introduced in this PR. The base contract had `location` on the Compute envelope and no `ComputeProperties.location`; the head moves it into `properties.location` for both preview versions.

**Suggested fix:** Use a domain-specific wire name such as `computeLocation` inside `ComputeProperties`, and update the service implementation and examples to accept that name. If this is intended to be the resource's Azure location, model it through the appropriate standard ARM resource envelope instead of duplicating the reserved name inside `properties`.

```tsp
/** The Azure region where the compute workload is hosted. */
computeLocation: string;
```

ravimeda:
**[EXISTING] 🔴 Blocking** **[[Section 1.2 - Proxy Resource Paths](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#12-proxy-resource-paths)]** `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` - line 46 - Compute extends `ProxyResource` but exposes top-level `tags`; proxy resources must not have ARM tags.

**Classification reasoning:** Pre-existing. The same `tags?: Record<string>` property is present at base SHA `3f819f6f`, line 53, under the same proxy resource. The inline suppression acknowledges the envelope violation but cites backward compatibility rather than an ARM exception.

**Suggested fix:** Remove top-level `tags` from Compute before promoting this preview resource to GA. If customers need service-owned metadata, expose a domain-specific property such as `labels` inside `ComputeProperties`; because the preview contract already shipped, make the removal in a new API version.

```tsp
model Compute is Azure.ResourceManager.ProxyResource<ComputeProperties, false> {
  // Keep only supported proxy-resource envelope fields.
}
```

ravimeda:
**[EXISTING] 🔴 Blocking** **[[RPC-Async-V1-02](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#what-it-represents) / [RPC-Async-V1-03](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#invalid-values)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 5906-5918 - `ComputeProvisioningState` includes runtime/action states (`Disabled`, `Starting`, `Stopping`, `Restarting`, `Stopped`) even though `provisioningState` may represent only the latest PUT/PATCH/DELETE provisioning operation; POST actions must not mutate it.

**Classification reasoning:** Pre-existing. The same five values occur at the same lines in the base-SHA `ComputeProvisioningState` union.

**Suggested fix:** In a new API version, keep only provisioning-operation states in `ComputeProvisioningState`, and add a separate read-only property such as `runtimeState` for runtime lifecycle values. Ensure the `start`, `stop`, and `restart` POST actions update `runtimeState`, not `provisioningState`. Do not remove or reshape these values in either already-published preview version.

```tsp
union ComputeRuntimeState {
  string,
  /** The compute resource is disabled. */
  Disabled: "Disabled",
  /** The compute resource is starting. */
  Starting: "Starting",
  /** The compute resource is stopping. */
  Stopping: "Stopping",
  /** The compute resource is restarting. */
  Restarting: "Restarting",
  /** The compute resource is stopped. */
  Stopped: "Stopped",
}
```

**Note:** The suggested fix was revised after critic review to require a new API version rather than modifying either published preview contract.

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
swarupecenits: Thanks @ravimeda. The 202-only shape is intentional and matches the deployed backend, not an oversight:

- Live-validated (`2026-05-15-preview`): a successful create returns **202 Accepted** with `Azure-AsyncOperation` + `Location` headers; the service never returns 200/201, and failures return 4xx.
- The backend is shared with the Machine Learning Services RP. The sibling **ManagedNetwork** PUT in this same spec is already 202-only and carries the same readme suppressions (`PutResponseCodes`, `ProvisioningStateSpecifiedForLROPut`) - I mirrored that accepted precedent.
- Restoring `ArmResourceCreateOrReplaceAsync` (200/201) in the spec alone would make the contract **misrepresent** the live service; clients would expect a resource body that never arrives.

Both api-versions are **preview** and no GA exposes this shape, so this is a pre-GA correction with no stable-surface impact. Per your note, I'd like to take the **explicit ARM exception** route rather than desync the spec from the service. Final contract call is the service owner's - @SubaMoorthy, please confirm. Happy to share the Insomnia captures.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
swarupecenits: The placement is deliberate and validated against the deployed service:

- Live test (`2026-05-15-preview`): create with `properties.location` → **202**; the same request with `location` on the envelope and omitted from `properties` → **400 "The Location field is required."** The RP reads `location` **only** from `properties`.
- Renaming to `computeLocation` in the spec alone would break clients - the service requires the wire name `location` inside `properties`. A rename needs a coordinated **service-side** change + a new api-version, not a spec-only edit.

This is a preview-only correction of an incorrect pre-GA contract. I'd like to keep the truthful `properties.location` shape under an **ARM exception**; if the board prefers `computeLocation`, that's a service change [@SubaMoorthy](https://github.com/SubaMoorthy) would drive in a new version. Deferring the decision to Suba.

swarupecenits:
> **[EXISTING] 🔴 Blocking** **[[Section 1.2 - Proxy Resource Paths](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#12-proxy-resource-paths)]** `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` - line 46 - Compute extends `ProxyResource` but exposes top-level `tags`; proxy resources must not have ARM tags.
> 
> **Classification reasoning:** Pre-existing. The same `tags?: Record<string>` property is present at base SHA `3f819f6f`, line 53, under the same proxy resource. The inline suppression acknowledges the envelope violation but cites backward compatibility rather than an ARM exception.
> 
> **Suggested fix:** Remove top-level `tags` from Compute before promoting this preview resource to GA. If customers need service-owned metadata, expose a domain-specific property such as `labels` inside `ComputeProperties`; because the preview contract already shipped, make the removal in a new API version.
> 
> ```tsp
> model Compute is Azure.ResourceManager.ProxyResource<ComputeProperties, false> {
>   // Keep only supported proxy-resource envelope fields.
> }
> ```

Agreed - this is pre-existing (present at base SHA `3f819f6f`) and outside the scope of this bug-fix PR. As you note, `tags` can't be removed from an already-published preview without a breaking change, so the correct place to address it is a **new api-version**. Tracking it for the next version; leaving the published previews unchanged here.

swarupecenits:
> > **[EXISTING] 🔴 Blocking** **[[Section 1.2 - Proxy Resource Paths](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#12-proxy-resource-paths)]** `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` - line 46 - Compute extends `ProxyResource` but exposes top-level `tags`; proxy resources must not have ARM tags.
> > **Classification reasoning:** Pre-existing. The same `tags?: Record<string>` property is present at base SHA `3f819f6f`, line 53, under the same proxy resource. The inline suppression acknowledges the envelope violation but cites backward compatibility rather than an ARM exception.
> > **Suggested fix:** Remove top-level `tags` from Compute before promoting this preview resource to GA. If customers need service-owned metadata, expose a domain-specific property such as `labels` inside `ComputeProperties`; because the preview contract already shipped, make the removal in a new API version.
> > ```tsp
> > model Compute is Azure.ResourceManager.ProxyResource<ComputeProperties, false> {
> >   // Keep only supported proxy-resource envelope fields.
> > }
> > ```
> 
> Agreed - this is pre-existing (present at base SHA `3f819f6f`) and outside the scope of this bug-fix PR. As you note, `tags` can't be removed from an already-published preview without a breaking change, so the correct place to address it is a **new api-version**. Tracking it for the next version; leaving the published previews unchanged here.

Agreed - pre-existing (same values at base SHA) and out of scope for this PR. Per your revised guidance, I'll split the runtime lifecycle values (`Starting`/`Stopping`/`Restarting`/`Stopped`/`Disabled`) into a separate read-only `runtimeState` with `start`/`stop`/`restart` updating that instead of `provisioningState`, in a **new api-version**, and won't reshape either published preview here.

specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/PutCompute.json
```diff
@@ -10,97 +10,26 @@
...
         "pools": [
           {
             "name": "default",
             "vmPriority": "Regular",
             "instanceType": "Standard_DS3_v2",
             "nodeCount": 2
           }
-        ],
-        "subnetArmId": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/rgcognitiveservices/providers/Microsoft.Network/virtualNetworks/myVnet/subnets/default"
+        ]
```
ravimeda, line 21: **[NEW] 🟠 Warning** **[[EX-COVERAGE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/example-quality.md#ex-coverage--example-coverage)]** `specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/PutCompute.json` - line 21 - `subnetArmId` was removed from the only `Cluster` create example, leaving the `Cluster` PUT with a minimum-set example only.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` this example declared `subnetArmId` as a sibling of `pools`; at head SHA `dceb2e5c` the `pools` array closes at line 21 with no `subnetArmId` following it. The same removal is present in the 2026-05-15-preview source example and in both generated `examples/PutCompute.json` copies.

`subnetArmId` is still a declared property of `ClusterComputeProperties` (`preview/2026-03-15-preview/cognitiveservices.json` line 12801), and `GetCompute.json` (line 25) and `ListComputes.json` (line 29) still return it - so the only networking knob on this resource is returned by GET/LIST but demonstrated by no create example. EX-COVERAGE: "PUT and PATCH operations **SHOULD** have at least two examples: one with the minimum required properties and one with the full/maximum property set."

Note for completeness: EX-COVERAGE also escalates to Blocking "when an LRO has no example showing both the in-progress and terminal responses", which is now true of this PUT (only the `202` remains). That clause is not raised separately here because it cannot be satisfied while the operation declares no `200`/`201` response - it is downstream of the open 202-only discussion on `Compute.tsp` line 76 and should be revisited once that is settled.

**Suggested fix:** Restore `subnetArmId` as a sibling of `pools`, or add a second maximum-set example (e.g. `PutComputeMaximumSet.json`) that includes it and wire it into `x-ms-examples`.

```json
"properties": {
  "computeType": "Cluster",
  "location": "eastus",
  "pools": [
    {
      "name": "default",
      "vmPriority": "Regular",
      "instanceType": "Standard_DS3_v2",
      "nodeCount": 2
    }
  ],
  "subnetArmId": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/rgcognitiveservices/providers/Microsoft.Network/virtualNetworks/myVnet/subnets/default"
}
```

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-03-15-preview/cognitiveservices.json
```diff
@@ -2684,26 +2684,17 @@
...
-            "schema": {
-              "$ref": "#/definitions/Compute"
-            }
-          },
-          "201": {
-            "description": "Resource 'Compute' create operation succeeded",
-            "schema": {
-              "$ref": "#/definitions/Compute"
-            },
+          "202": {
```
ravimeda, line 2687: **[NEW] 🔵 Suggestion** **[[Section 6.6 - 202 Response and Polling Headers](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#66-202-response-and-polling-headers-rpc-async-v1-07-rpc-async-v1-06-rpc-async-v1-14)]** `specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-03-15-preview/cognitiveservices.json` - line 2687 - the new PUT `202` declares `Azure-AsyncOperation` and `Location` but not `Retry-After`, so clients fall back to the default polling interval on the longest-running operation of this resource.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` this PUT had no `202` at all, and its `201` declared `Retry-After` (base lines 2703-2707). This PR replaces that with a `202` that omits it, so a polling hint that previously existed on the LRO initiation response is gone. The same shape is at line 2687 of the 2026-05-15-preview file.

Section 6.6: "Include `Retry-After` header (integer, seconds) when supported. Minimum 10 seconds, maximum 10 minutes. Default client polling interval is 60 seconds if `Retry-After` is not provided."

The PR notes say the service does not set `Retry-After`. Two observations against that: `DELETE` on this same resource still declares it (same file, lines 2759-2771), so the two async operations on one resource now advertise different polling contracts; and the base `201` declared it, so the capability was previously modeled. Compute cluster creation is the operation where a service-chosen cadence matters most.

**Suggested fix:** Have the RP emit `Retry-After` on the create `202` and model it in the TypeSpec source rather than the emitted swagger - `cognitiveservices.json` is regenerated by `tsp compile` and hand edits are discarded. The relevant construct is the `ArmCombinedLroHeaders` argument to `ArmAcceptedLroResponse` at `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` lines 80-83. If the service genuinely cannot emit it, please say so on this thread so the asymmetry with `DELETE` is a recorded decision rather than an oversight.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -5951,7 +5951,7 @@ model Pool {
   /**
    * The VM priority of the pool.
    */
-  vmPriority: VmPriority;
+  vmPriority?: VmPriority;
```
ravimeda, line 5954: **[NEW] 🔵 Suggestion** **[[OAPI022 / WHATIF-002 - Section 8.14 Default Values Must Be Static and Declared](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#814-default-values-must-be-static-and-declared-oapi022-whatif-002)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 5954 - `vmPriority` became optional but declares no default, so the behavior when it is omitted is not discoverable from the spec.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f`, line 5954 was `vmPriority: VmPriority;` (required); at head SHA `dceb2e5c` it is `vmPriority?: VmPriority;`. The doc comment (lines 5951-5953) still reads only "The VM priority of the pool."

Section 8.14: "Every **optional** property where the service applies a default value when the property is omitted from PUT **MUST** declare that default value in the swagger spec. Without the annotation, ARM Template What-If reports a false delete for the property (the template omits it, but GET returns the service-applied default)."

`Regular` and `Spot` (the two members of `union VmPriority`, lines 5928-5936) carry materially different eviction and cost characteristics, so this is not only a What-If concern - a caller who omits the property cannot tell from the spec whether the pool they created is evictable.

**Suggested fix:** If the service applies a default, declare it so it is emitted into the swagger:

```tsp
/**
 * The VM priority of the pool.
 */
vmPriority?: VmPriority = VmPriority.Regular;
```

If the service applies no default and omission is genuinely meaningful, state that in the doc comment instead so the behavior is discoverable. Note that adding the default will inline the enum at the emitted `Pool.vmPriority` (a sibling `default` cannot coexist with a `$ref`); this matches the existing `RoutingMode` precedent in the same emitted file (`preview/2026-03-15-preview/cognitiveservices.json` line 15031).

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`dceb2e5`](https://github.com/Azure/azure-rest-api-specs/pull/44956/commits/dceb2e5c53114900684080e89c0210b3874c426b). See inline comments for findings 1-3.

No new blocking findings this pass: the 202-only PUT contract and the `location`-in-`properties` placement are already tracked on the two open threads above and are not re-raised here.

specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/PutCompute.json
swarupecenits: > **[NEW] 🟠 Warning** **[[EX-COVERAGE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/example-quality.md#ex-coverage--example-coverage)]** `specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/PutCompute.json` - line 21 - `subnetArmId` was removed from the only `Cluster` create example, leaving the `Cluster` PUT with a minimum-set example only.
> 
> **Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` this example declared `subnetArmId` as a sibling of `pools`; at head SHA `dceb2e5c` the `pools` array closes at line 21 with no `subnetArmId` following it. The same removal is present in the 2026-05-15-preview source example and in both generated `examples/PutCompute.json` copies.
> 
> `subnetArmId` is still a declared property of `ClusterComputeProperties` (`preview/2026-03-15-preview/cognitiveservices.json` line 12801), and `GetCompute.json` (line 25) and `ListComputes.json` (line 29) still return it - so the only networking knob on this resource is returned by GET/LIST but demonstrated by no create example. EX-COVERAGE: "PUT and PATCH operations **SHOULD** have at least two examples: one with the minimum required properties and one with the full/maximum property set."
> 
> Note for completeness: EX-COVERAGE also escalates to Blocking "when an LRO has no example showing both the in-progress and terminal responses", which is now true of this PUT (only the `202` remains). That clause is not raised separately here because it cannot be satisfied while the operation declares no `200`/`201` response - it is downstream of the open 202-only discussion on `Compute.tsp` line 76 and should be revisited once that is settled.
> 
> **Suggested fix:** Restore `subnetArmId` as a sibling of `pools`, or add a second maximum-set example (e.g. `PutComputeMaximumSet.json`) that includes it and wire it into `x-ms-examples`.
> 
> ```json
> "properties": {
>   "computeType": "Cluster",
>   "location": "eastus",
>   "pools": [
>     {
>       "name": "default",
>       "vmPriority": "Regular",
>       "instanceType": "Standard_DS3_v2",
>       "nodeCount": 2
>     }
>   ],
>   "subnetArmId": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/rgcognitiveservices/providers/Microsoft.Network/virtualNetworks/myVnet/subnets/default"
> }
> ```

Good catch on the coverage gap. Note `subnetArmId` is currently rejected by the service on create, the RP's request validation returns `"Subnet ARM ID is not valid."` for any non-empty value (it's declared for forward-compat but not yet accepted). So restoring it to the create example would ship a request the service 400s. The minimum-set example is the accurate maximal *working* create today; the only other optional create-time knob is envelope `tags`. Happy to add a `PutComputeMaximumSet.json` that adds `tags` (and omits the not-yet-supported `subnetArmId`, with a note) if that satisfies EX-COVERAGE. cc @SubaMoorthy to confirm the subnetArmId status.

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-03-15-preview/cognitiveservices.json
swarupecenits: > **[NEW] 🔵 Suggestion** **[[Section 6.6 - 202 Response and Polling Headers](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#66-202-response-and-polling-headers-rpc-async-v1-07-rpc-async-v1-06-rpc-async-v1-14)]** `specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-03-15-preview/cognitiveservices.json` - line 2687 - the new PUT `202` declares `Azure-AsyncOperation` and `Location` but not `Retry-After`, so clients fall back to the default polling interval on the longest-running operation of this resource.
> 
> **Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` this PUT had no `202` at all, and its `201` declared `Retry-After` (base lines 2703-2707). This PR replaces that with a `202` that omits it, so a polling hint that previously existed on the LRO initiation response is gone. The same shape is at line 2687 of the 2026-05-15-preview file.
> 
> Section 6.6: "Include `Retry-After` header (integer, seconds) when supported. Minimum 10 seconds, maximum 10 minutes. Default client polling interval is 60 seconds if `Retry-After` is not provided."
> 
> The PR notes say the service does not set `Retry-After`. Two observations against that: `DELETE` on this same resource still declares it (same file, lines 2759-2771), so the two async operations on one resource now advertise different polling contracts; and the base `201` declared it, so the capability was previously modeled. Compute cluster creation is the operation where a service-chosen cadence matters most.
> 
> **Suggested fix:** Have the RP emit `Retry-After` on the create `202` and model it in the TypeSpec source rather than the emitted swagger - `cognitiveservices.json` is regenerated by `tsp compile` and hand edits are discarded. The relevant construct is the `ArmCombinedLroHeaders` argument to `ArmAcceptedLroResponse` at `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` lines 80-83. If the service genuinely cannot emit it, please say so on this thread so the asymmetry with `DELETE` is a recorded decision rather than an oversight.

Recording the decision here as requested. The create `202` emits `Azure-AsyncOperation` + `Location`; we have no evidence the RP sets `Retry-After` on create (the service contract declares no `Retry-After` on this response). The `DELETE` asymmetry is inherited from the standard ARM delete LRO template rather than a service-chosen cadence. If the RP confirms it does emit a `Retry-After` on the create `202`, we'll model it via `ArmCombinedLroHeaders` in `Compute.tsp`; otherwise please treat this as the recorded decision. cc @SubaMoorthy to confirm service behavior.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
swarupecenits: > **[NEW] 🔵 Suggestion** **[[OAPI022 / WHATIF-002 - Section 8.14 Default Values Must Be Static and Declared](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#814-default-values-must-be-static-and-declared-oapi022-whatif-002)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 5954 - `vmPriority` became optional but declares no default, so the behavior when it is omitted is not discoverable from the spec.
> 
> **Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f`, line 5954 was `vmPriority: VmPriority;` (required); at head SHA `dceb2e5c` it is `vmPriority?: VmPriority;`. The doc comment (lines 5951-5953) still reads only "The VM priority of the pool."
> 
> Section 8.14: "Every **optional** property where the service applies a default value when the property is omitted from PUT **MUST** declare that default value in the swagger spec. Without the annotation, ARM Template What-If reports a false delete for the property (the template omits it, but GET returns the service-applied default)."
> 
> `Regular` and `Spot` (the two members of `union VmPriority`, lines 5928-5936) carry materially different eviction and cost characteristics, so this is not only a What-If concern - a caller who omits the property cannot tell from the spec whether the pool they created is evictable.
> 
> **Suggested fix:** If the service applies a default, declare it so it is emitted into the swagger:
> 
> ```tsp
> /**
>  * The VM priority of the pool.
>  */
> vmPriority?: VmPriority = VmPriority.Regular;
> ```
> 
> If the service applies no default and omission is genuinely meaningful, state that in the doc comment instead so the behavior is discoverable. Note that adding the default will inline the enum at the emitted `Pool.vmPriority` (a sibling `default` cannot coexist with a `$ref`); this matches the existing `RoutingMode` precedent in the same emitted file (`preview/2026-03-15-preview/cognitiveservices.json` line 15031).

The service applies no default here, `vmPriority` is nullable server-side and stays unset when omitted, so declaring `= Regular` would emit a default the service doesn't apply (and misrepresent eviction/cost behavior). Per 8.14's alternative, I've clarified the doc comment to state that omission means no priority is applied. cc @SubaMoorthy if a Regular default is actually intended, in which case we'll model it explicitly.

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
```diff
@@ -80,15 +73,15 @@ interface Computes {
...
-  update is ArmCustomPatchAsync<Compute, PatchModel = Compute>;
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<
+    Compute,
+    Response = ArmAcceptedLroResponse<
+      "Resource creation accepted.",
+      ArmCombinedLroHeaders
+    >
+  >;
 
   /**
```
ravimeda, line 86: **[NEW] 🔴 Blocking** **[[Section 2 — API Versioning](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2-api-versioning) / [TSP-2.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#21-breaking-changes-across-versions)]** `specification/cognitiveservices/CognitiveServices.Management/Compute.tsp` - line 86 - the `Computes_Update` (PATCH) operation is deleted from the `Computes` interface, removing it from both already-published preview API versions.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` the interface declared `update is ArmCustomPatchAsync<Compute, PatchModel = Compute>;` at line 91, between `createOrUpdate` (line 84) and `delete` (line 97). At head SHA `dceb2e5c` line 86 opens the `delete` doc comment and no `update` operation exists anywhere in the file. In the emitted swagger, `Computes_Update` goes from 1 occurrence to 0 in both `preview/2026-03-15-preview/cognitiveservices.json` and `preview/2026-05-15-preview/cognitiveservices.json`, and the verb set on the computes path goes from `delete,get,patch,put` to `delete,get,put`. The `@@doc(Computes.update::parameters.properties, ...)` augment was deleted, and all four `UpdateCompute.json` example files were removed.

Removing an operation from an already-published API version is a breaking change — Section 2 lists "Removing or renaming properties, operations, or parameters" among breaking changes introduced within the same API version. Both `2026-03-15-preview` and `2026-05-15-preview` were merged to `main` before this PR (2026-06-02 and 2026-06-25 respectively), so this is an in-place change to a published contract, not a change in a new version.

To be clear about scope: the removal **is** disclosed in the PR description, and `Swagger BreakingChange - Analyze Code` **did** fail on it — the `Swagger BreakingChange` gate passed only because it matched the `Versioning-Approved-BugFix` label. This comment is therefore about approval coverage, not about detection.

**Approval context:** `Versioning-Approved-BugFix` was observed at the session SHA; no `BreakingChange-Approved-*` label is present. The complete set of same-version breaking changes in this PR is: (a) the `Computes_Update` PATCH operation removed; (b) the PUT `200`/`201` responses removed; (c) enum value `LowPriority` removed; (d) `location` moved from the resource envelope into `properties` and changed from optional to required; (e) `Pool.vmPriority` changed from required to optional. Please confirm that the versioning approval explicitly covers each of these. If it does, ensure the appropriate label is applied and resolve this conversation; otherwise obtain approval or address the finding.

**Suggested fix:** Restore the operation, its examples, and its `@@doc` augment for both published previews, or gate the removal with `@removed(Versions.<newPreviewVersion>)` in a new API version.

```tsp
  /**
   * Updates a compute associated with the Cognitive Services account.
   */
  @tag("Computes")
  @patch(#{ implicitOptionality: false })
  update is ArmCustomPatchAsync<Compute, PatchModel = Compute>;
```

Separately, the "Examples corrected" bullet in the PR description still reads "The `UpdateCompute` PATCH request is unchanged since `location` is create/read-only" — those four example files no longer exist. Please update the description.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [dceb2e5](https://github.com/Azure/azure-rest-api-specs/pull/44956/commits/dceb2e5c53114900684080e89c0210b3874c426b). See inline comments for findings 1-4. A fifth finding concerns readme.md lines that are unchanged in this PR diff and is posted as a top-level PR comment rather than inline.

Approval labels observed: `Versioning-Approved-BugFix`.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
```diff
@@ -5931,8 +5931,8 @@ union VmPriority {
   /** Regular VM priority. */
   Regular: "Regular",
 
-  /** Low-priority VM. */
-  LowPriority: "LowPriority",
+  /** Spot VM priority. */
+  Spot: "Spot",
```
ravimeda, line 5935: **[NEW] 🔴 Blocking** **[[Section 2 — API Versioning](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2-api-versioning)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 5934-5935 - the `VmPriority` union member `LowPriority: "LowPriority"` is replaced by `Spot: "Spot"`, removing a wire value from two already-published preview API versions.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` lines 5934-5935 read `/** Low-priority VM. */` and `LowPriority: "LowPriority",`; at head SHA `dceb2e5c` the same two lines read `/** Spot VM priority. */` and `Spot: "Spot",` — a one-for-one replacement, not an addition. In the emitted swagger the `VmPriority` enum goes from `["Regular", "LowPriority"]` to `["Regular", "Spot"]`, and the `LowPriority` token count goes from 3 to 0 in both `preview/2026-03-15-preview/cognitiveservices.json` and `preview/2026-05-15-preview/cognitiveservices.json`. The `@added(Versions.v2026_03_15_preview)` and `@added(Versions.v2026_05_15_preview)` decorators on the union confirm it is live in both published versions.

Section 2 lists "Removing enum values" among the breaking changes that must not be introduced within the same API version. Any caller currently sending `"LowPriority"` — a value the published contract accepted — will break. Note this is a value **removal**, so the usual `modelAsString: true` allowance for adding new enum values does not apply here.

**Approval context:** `Versioning-Approved-BugFix` was observed at the session SHA; no `BreakingChange-Approved-*` label is present. The change is disclosed in the PR description, but disclosure is not approval. Please confirm that the versioning approval explicitly covers this enum-value removal. If it does, ensure the appropriate label is applied and resolve this conversation; otherwise obtain approval or address the finding.

**Suggested fix:** Keep `LowPriority` alongside `Spot` in the two published preview versions so existing callers keep working, and document `LowPriority` as deprecated. Alternatively, introduce `Spot` and retire `LowPriority` in a new API version rather than in place.

```tsp
union VmPriority {
  string,

  /** Regular VM priority. */
  Regular: "Regular",

  /** Spot VM priority. */
  Spot: "Spot",

  /** Deprecated. Use Spot instead. */
  LowPriority: "LowPriority",
}
```

specification/cognitiveservices/resource-manager/readme.md
```diff
@@ -132,6 +132,18 @@ These settings apply only when `--tag=package-2026-05-15-preview` is specified o
...
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/computes/{computeName}"].put
+  - code: ProvisioningStateSpecifiedForLROPut
+    reason: Compute create returns 202 Accepted for long-running provisioning. provisioningState is present in ComputeProperties and returned in the final LRO result retrieved via GET. Same pattern as the ManagedNetwork PUT (shared Machine Learning Services backend). Preview-only bug fix.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/computes/{computeName}"].put
+  - code: PutGetPatchResponseSchema
+    reason: Compute create PUT is 202-only with no response body, so its response schema intentionally differs from GET, which returns the Compute resource once the LRO completes. This is inherent to the async 202-only contract. Preview-only bug fix.
+    where:
+      - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.CognitiveServices/accounts/{accountName}/computes/{computeName}"]
```
ravimeda, line 146: **[NEW] 🟠 Warning** **[[Section 10A — readme.md Suppression Scoping](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#10a-readmemd-suppression-scoping--tag-coverage) / [Suppression Review Criteria](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md)]** `specification/cognitiveservices/resource-manager/readme.md` - lines 143-146 - six new suppressions are added across two tags and none of them has ARM sign-off yet.

**Classification reasoning:** Introduced in this PR. Three codes are added to each of the `package-2026-05-15-preview` and `package-2026-03-15-preview` tag blocks: `PutResponseCodes` (lines 135-138 and 370-373), `ProvisioningStateSpecifiedForLROPut` (lines 139-142 and 374-377), and `PutGetPatchResponseSchema` (lines 143-146 and 378-381). The 370-381 block is byte-identical to 135-146. At base SHA `3f819f6f` there are zero suppressions targeting the computes path or its `.put`, so all six entries are genuinely new rather than carried forward from a prior version.

At the session SHA the PR carries `SuppressionReviewRequired` and does **not** carry `Approved-Suppression`, so none of the six has been reviewed and signed off.

**Secondary observation — reason self-containment.** [`typespec-review.instructions.md` 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy) states that suppression reasons **MUST NOT** justify by reference to another resource pattern. The `PutResponseCodes` reason (line 136) ends with "consistent with the ManagedNetwork PUT in this RP" and the `ProvisioningStateSpecifiedForLROPut` reason (line 140) ends with "Same pattern as the ManagedNetwork PUT". Both reasons do lead with a standalone technical justification, so they are not squarely in violation — but the appended ManagedNetwork sentences invite that reading and should be dropped so each reason stands entirely on its own.

For balance: [`suppression-review-criteria.md`](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md) routes "anti-pattern replicated from an already-shipped stable version" to **peer review** rather than automatic rejection, and its worked example is literally `202` for a long-running PUT. So this is a peer-review decision for the ARM API review team, not an automatic reject — this comment asks for that decision to be made and recorded, it does not assert the suppressions are wrong.

**Approval context:** `Approved-Suppression` was not present at the session SHA; only `SuppressionReviewRequired` was observed. Please confirm whether ARM suppression approval already exists for these six entries. If it does, ensure `Approved-Suppression` is applied and resolve this conversation; otherwise obtain approval or address the finding.

**Suggested fix:** Take all six entries through ARM API review and get `Approved-Suppression` applied. Rewrite the `PutResponseCodes` and `ProvisioningStateSpecifiedForLROPut` reasons so each is self-contained for the `Compute` resource without leaning on the ManagedNetwork precedent. Do **not** narrow or reshape the `where:` clauses without first confirming they still match the LintDiff-reported jsonpath exactly — a changed path silently stops suppressing and the check will fail.

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/cognitiveservices.json
```diff
@@ -2723,93 +2714,7 @@
...
-        },
-        "x-ms-examples": {
-          "UpdateCompute": {
-            "$ref": "./examples/UpdateCompute.json"
-          }
-        },
-        "x-ms-long-running-operation-options": {
-          "final-state-via": "location",
-          "final-state-schema": "#/definitions/Compute"
+          "final-state-via": "location"
```
ravimeda, line 2717: **[NEW] 🟠 Warning** **[[Section 6.7 — Polling Behavior and final-state-via](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#67-polling-behavior-and-final-state-via) / [LRO final-state-via reference](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/lro-final-state-via.md)]** `specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/cognitiveservices.json` - line 2717 - the PUT declares `final-state-via: location` while defining only `202` and `default`, so generated SDKs have no declared type for the created `Compute`.

JSON path: the `x-ms-long-running-operation-options.final-state-via` key on the computes PUT. The identical shape is at line 2717 of the `2026-03-15-preview` file.

**Classification reasoning:** Introduced in this PR. At base SHA `3f819f6f` this PUT declared `final-state-via: azure-async-operation` together with `final-state-schema: #/definitions/Compute`; this PR removes both and emits `final-state-via: location` with no final-state schema and no success response schema.

**Why this matters.** Per the reference, `final-state-via` does **not** affect polling — polling is chosen at runtime from the response headers. It only tells the SDK **where to make the final GET** to deserialize the result. `location` means GET the URL from the `Location` response header. In the accompanying examples that header resolves to `.../providers/Microsoft.CognitiveServices/locations/eastus/computeOperations/{guid}`, and `ComputeOperations_Get` on that path returns `ComputeOperationStatus` — not `Compute`. Combined with the removal of the `200`/`201` response schemas, the generated `BeginCreateOrUpdate` has no declared result type in any language, so a caller cannot obtain the created `Compute` from the create call and must issue a separate GET.

Section 6.7 also states: "For PUT, PATCH, and DELETE following standard ARM patterns, do **NOT** specify `x-ms-long-running-operation-options` / `final-state-via` -- the default SDK behavior is correct. Only specify `final-state-via: location` for POST LROs with a response schema."

**Suggested fix (options, not a directive — please pick one and record the rationale):**

1. Restore a resource-shaped success response on the PUT so the SDK can deserialize the created `Compute`. This is the ARM-standard shape and also resolves the open `PutResponseCodes` discussion on `Compute.tsp` line 76.
2. Keep the 202-only contract but re-add `final-state-schema: #/definitions/Compute` and set `final-state-via` to match the header you intend clients to follow, documenting the choice in the suppression reason.
3. If the current shape is genuinely intended and create is expected to return no resource, raise it with the ARM API review board and record the decision — it changes the generated SDK signature in every language, which is what the `BreakingChange-Go-Sdk`, `BreakingChange-JavaScript-Sdk`, and `BreakingChange-Python-Sdk` labels on this PR are reflecting.

This is complementary to, not a duplicate of, the open thread on `Compute.tsp` line 76: that thread is about the response **codes**, this one is about the LRO **final-state contract** and the resulting SDK return type.

ravimeda:
**[NEW] 🔵 Suggestion** **[[Section 10A — readme.md Suppression Scoping](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#10a-readmemd-suppression-scoping--tag-coverage) / [TSP-4.1 Warning Suppression Policy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/cognitiveservices/resource-manager/readme.md` - lines 187-191 (mirrored at lines 422-426) - removing the Computes PATCH operation orphans existing suppression entries and invalidates a neighbouring suppression reason.

_Posted as a top-level comment because these lines are outside this PR diff and cannot be anchored inline._

**Classification reasoning:** Introduced in this PR, indirectly. This PR deletes `Computes_Update`, so the computes path verb set becomes `delete,get,put` in both preview swaggers. Two consequences in `readme.md`, in both the `package-2026-05-15-preview` and `package-2026-03-15-preview` tag blocks:

1. **Orphaned `where:` paths.** Lines 187-189 (and 422-424) still list three `PatchBodyParametersSchema` targets under `.../computes/{computeName}"].patch.parameters[5].schema...`. With no `patch` verb on that path, all six entries now resolve to nothing and are dead configuration. Section 10A calls for suppressions to have specific, accurate `where:` clauses; the suppression-continuity analysis for this PR covered the six entries it **adds** but not the ones it **orphans**.
2. **Stale reason.** Line 191 (and 426) carries the Workbench `PatchBodyParametersSchema` reason "Workbench uses PatchModel = Workbench (full resource as PATCH body), consistent with the existing Compute resource pattern." This PR deletes that Compute PATCH pattern, so the justification now points at something that no longer exists. TSP-4.1 is directly on point: "Suppression reasons **MUST** be factually accurate. If the reason claims *existing API design* or *would break the service contract* but the PR itself changes the relevant path or model, flag the inconsistency."

**Suggested fix:** If the PATCH removal stands (see the blocking comment on `Compute.tsp` line 86), delete the three orphaned computes `where:` paths from both tag blocks, and reword the Workbench reason so it justifies the Workbench PATCH body shape on its own terms rather than by reference to the Compute resource. If instead the PATCH operation is restored, both items resolve themselves and no readme change is needed — so this is best handled after the PATCH question is settled.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
SubaMoorthy: Yes, vm priority don't need to be set. Empty should be the default.

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
SubaMoorthy: For now, lets keep it synced with backend and use only 202. Back end can evaluate adding the 201 and 200 to match arm requirements

specification/cognitiveservices/CognitiveServices.Management/examples/2026-03-15-preview/PutCompute.json
SubaMoorthy: Create should not set subnetArmId.

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
swarupecenits: Confirming coverage. All five changes are **preview-only** corrections that sync the published contract to the live backend, both `2026-03-15-preview` and `2026-05-15-preview` are `@removed` before any stable/GA version, so no GA contract is affected:



- **PATCH removed** - done at the **explicit direction of the service owner** ([@SubaMoorthy](https://github.com/SubaMoorthy): *"Patch is not implemented in backend, can remove from swagger?"*). The backend PATCH is a no-op placeholder.

- **PUT 200/201 removed** - backend returns only `202` on create (Suba confirmed above: *"keep it synced with backend and use only 202"*).

- **`LowPriority` enum removed** - backend enum is `Regular`/`Spot` only; `LowPriority` was never a valid service value (see thread below).

- **`location` → `properties`, now required** - backend requires it in the body; the old envelope placement produced a live `400 "The Location field is required"`.

- **`vmPriority` required → optional** - backend field is nullable (Suba: *"don't need to be set, empty should be the default"*).



This is exactly the scope `Versioning-Approved-BugFix` was granted for. [@SubaMoorthy](https://github.com/SubaMoorthy), please confirm coverage so the label can be applied and this thread resolved. PR description already updated to drop the stale `UpdateCompute` bullet.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
swarupecenits: Covered by the same preview-only bug-fix approval. The backend enum is `{ Regular, Spot }` - `LowPriority` was a **spec-only bug** the service never accepted, so no live caller could have used it successfully. Retaining it (even deprecated) would document a value the backend rejects, which is misleading. Both affected versions are preview-only (`@removed` pre-GA). [@SubaMoorthy](https://github.com/SubaMoorthy) can confirm the backend enum values.

specification/cognitiveservices/resource-manager/readme.md
swarupecenits: Agreed on both points. (1) Requesting these six through ARM API review for `Approved-Suppression` - thanks for confirming they route to peer review with the 202-PUT worked example as precedent. (2) Rewording the two reasons to be fully self-contained (backend-truth only, no cross-resource reference); `where:` clauses left byte-identical so they keep matching the LintDiff jsonpaths:



- **PutResponseCodes:** *Compute create is a long-running async operation, the service returns 202 Accepted on success (never 200/201) and 4xx on failure. 202-only reflects the live backend contract. Preview-only bug fix correcting the contract before GA.*

- **ProvisioningStateSpecifiedForLROPut:** *Compute create returns 202 Accepted for long-running provisioning. provisioningState is present in ComputeProperties and returned in the final LRO result retrieved via GET. Preview-only bug fix*

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-05-15-preview/cognitiveservices.json
swarupecenits: This is the intended behavior for now and matches the backend: create is `202`-only and returns no resource body; the client issues a follow-up `GET` on the resource to read the provisioned `Compute` (Suba: *"keep it synced with backend, use only 202; backend can evaluate adding 201/200 later"*). The SDK signature change is expected and is exactly what the `BreakingChange-*-Sdk` labels reflect.

We're **not** adding `final-state-schema: #/definitions/Compute`, because the `Location` polling URL genuinely returns `ComputeOperationStatus`, not `Compute`, that annotation would be inaccurate. If the board prefers restoring an SDK return type while staying 202-only, the clean path is to drop the explicit `final-state-via` so the SDK falls back to the standard final `GET` on the resource URL (which returns `Compute`); happy to adjust the TypeSpec that way if you'd like it recorded as the decision. [@SubaMoorthy](https://github.com/SubaMoorthy) for the service call.

swarupecenits:
> **[NEW] 🔵 Suggestion** **[[Section 10A — readme.md Suppression Scoping](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#10a-readmemd-suppression-scoping--tag-coverage) / [TSP-4.1 Warning Suppression Policy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/cognitiveservices/resource-manager/readme.md` - lines 187-191 (mirrored at lines 422-426) - removing the Computes PATCH operation orphans existing suppression entries and invalidates a neighbouring suppression reason.
> 
> _Posted as a top-level comment because these lines are outside this PR diff and cannot be anchored inline._
> 
> **Classification reasoning:** Introduced in this PR, indirectly. This PR deletes `Computes_Update`, so the computes path verb set becomes `delete,get,put` in both preview swaggers. Two consequences in `readme.md`, in both the `package-2026-05-15-preview` and `package-2026-03-15-preview` tag blocks:
> 
> 1. **Orphaned `where:` paths.** Lines 187-189 (and 422-424) still list three `PatchBodyParametersSchema` targets under `.../computes/{computeName}"].patch.parameters[5].schema...`. With no `patch` verb on that path, all six entries now resolve to nothing and are dead configuration. Section 10A calls for suppressions to have specific, accurate `where:` clauses; the suppression-continuity analysis for this PR covered the six entries it **adds** but not the ones it **orphans**.
> 2. **Stale reason.** Line 191 (and 426) carries the Workbench `PatchBodyParametersSchema` reason "Workbench uses PatchModel = Workbench (full resource as PATCH body), consistent with the existing Compute resource pattern." This PR deletes that Compute PATCH pattern, so the justification now points at something that no longer exists. TSP-4.1 is directly on point: "Suppression reasons **MUST** be factually accurate. If the reason claims _existing API design_ or _would break the service contract_ but the PR itself changes the relevant path or model, flag the inconsistency."
> 
> **Suggested fix:** If the PATCH removal stands (see the blocking comment on `Compute.tsp` line 86), delete the three orphaned computes `where:` paths from both tag blocks, and reword the Workbench reason so it justifies the Workbench PATCH body shape on its own terms rather than by reference to the Compute resource. If instead the PATCH operation is restored, both items resolve themselves and no readme change is needed — so this is best handled after the PATCH question is settled.

Good catch, since the PATCH removal stands (service-owner-directed), I'll remove the three orphaned `computes` `PatchBodyParametersSchema` `where:` entries from both tag blocks and reword the Workbench reason to stand on its own: *"Workbench uses PatchModel = Workbench (full resource as PATCH body); required properties (targetClusterId, imageLink) are within the optional properties bag.*

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`165cf42`](https://github.com/Azure/azure-rest-api-specs/pull/44956/commits/165cf422474f288e3f7473aa23685a21bf42d02d). See inline comments for findings 5, 9, and 10. All three findings concern lines unchanged in this PR's diff and are posted as top-level PR comments rather than inline comments.

Approval labels observed: `Versioning-Approved-BugFix`.

ravimeda:
**[EXISTING] 🔴 Blocking** **[[RPC-Async-V1-02 / RPC-Async-V1-03](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#invalid-values)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 5905-5918 - `ComputeProvisioningState` includes runtime states `Disabled`, `Starting`, `Stopping`, `Restarting`, and `Stopped`. POST actions must not update `provisioningState`.

**Classification reasoning:** The same values exist at the base SHA.

**Suggested fix:** Introduce a separate read-only runtime-state property in a new API version:
```tsp
@visibility(Lifecycle.Read)
powerState?: ComputePowerState;
```

swarupecenits:
> **[EXISTING] 🔴 Blocking** **[[RPC-Async-V1-02 / RPC-Async-V1-03](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#invalid-values)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 5905-5918 - `ComputeProvisioningState` includes runtime states `Disabled`, `Starting`, `Stopping`, `Restarting`, and `Stopped`. POST actions must not update `provisioningState`.
> 
> **Classification reasoning:** The same values exist at the base SHA.
> 
> **Suggested fix:** Introduce a separate read-only runtime-state property in a new API version:
> 
> ```tsp
> @visibility(Lifecycle.Read)
> powerState?: ComputePowerState;
> ```

Thanks Ravi - noting this is `[EXISTING]`, not touched by this PR (scoped to create: 202 LRO + `location` in `properties`). On the service `provisioningState` is a free string and the backend only emits `Accepted`/`Creating`/`Scaling`/`Succeeded`/`Deleting`; the five runtime states are never produced (no start/stop/restart actions) and the union is `modelAsString`, so nothing breaks. The ARM-correct split needs a **new api-version**. Requesting this be treated as pre-existing/out-of-scope and non-blocking here. [@SubaMoorthy](https://github.com/SubaMoorthy) to confirm.

ravimeda:
**[EXISTING] 🟠 Warning** **[[TSP-ARM-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#63-use-armresourceidentifier-for-resource-references-tsp-arm-resource-id)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 6045 and 6064 - `subnetArmId` and `targetClusterId` are documented ARM resource references but use plain `string`.

**Classification reasoning:** The same declarations exist at base lines 6039 and 6058.

**Suggested fix:**
```tsp
subnetArmId?: armResourceIdentifier<[{ type: "Microsoft.Network/virtualNetworks/subnets" }]>;
targetClusterId: armResourceIdentifier<[{ type: "Microsoft.CognitiveServices/accounts/computes" }]>;
```

ravimeda:
**[EXISTING] 🟠 Warning** **[[Section 2.2 - Type Constraints](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 6074 - `idleTimeBeforeShutdown` is documented and exemplified as an ISO 8601 duration (`PT30M`) but uses plain `string`.

**Classification reasoning:** The same declaration exists at base lines 6065-6068.

**Suggested fix:**
```tsp
idleTimeBeforeShutdown?: duration;
```

swarupecenits:
> **[EXISTING] 🟠 Warning** **[[TSP-ARM-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#63-use-armresourceidentifier-for-resource-references-tsp-arm-resource-id)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 6045 and 6064 - `subnetArmId` and `targetClusterId` are documented ARM resource references but use plain `string`.
> 
> **Classification reasoning:** The same declarations exist at base lines 6039 and 6058.
> 
> **Suggested fix:**
> 
> ```tsp
> subnetArmId?: armResourceIdentifier<[{ type: "Microsoft.Network/virtualNetworks/subnets" }]>;
> targetClusterId: armResourceIdentifier<[{ type: "Microsoft.CognitiveServices/accounts/computes" }]>;
> ```

Thanks - `[EXISTING]` and non-blocking. Agreed `duration` is the more precise type, but the field is already shipped in both previews as `string`, so adding the `format` belongs in the next api-version modeling cleanup rather than reshaping a published preview. Tracking it there; out of scope for this create-contract fix

swarupecenits:
> **[EXISTING] 🟠 Warning** **[[Section 2.2 - Type Constraints](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - line 6074 - `idleTimeBeforeShutdown` is documented and exemplified as an ISO 8601 duration (`PT30M`) but uses plain `string`.
> 
> **Classification reasoning:** The same declaration exists at base lines 6065-6068.
> 
> **Suggested fix:**
> 
> ```tsp
> idleTimeBeforeShutdown?: duration;
> ```

Thanks - `[EXISTING]` and non-blocking. `armResourceIdentifier` is the right refinement, but both are already published in the previews as plain `string`; adding the ARM-id `format` will be batched into the next api-version cleanup (same bucket as the runtime-state split) rather than changing a shipped preview. Out of scope for this bug-fix.

ravimeda (approved):

specification/cognitiveservices/CognitiveServices.Management/Compute.tsp
SubaMoorthy: Yes, patch is not supported.

specification/cognitiveservices/CognitiveServices.Management/models.tsp
SubaMoorthy: Ack, will be fixed in next api version.

SubaMoorthy:
> **[EXISTING] 🔴 Blocking** **[[RPC-Async-V1-02 / RPC-Async-V1-03](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#invalid-values)]** `specification/cognitiveservices/CognitiveServices.Management/models.tsp` - lines 5905-5918 - `ComputeProvisioningState` includes runtime states `Disabled`, `Starting`, `Stopping`, `Restarting`, and `Stopped`. POST actions must not update `provisioningState`.
> 
> **Classification reasoning:** The same values exist at the base SHA.
> 
> **Suggested fix:** Introduce a separate read-only runtime-state property in a new API version:
> 
> ```tsp
> @visibility(Lifecycle.Read)
> powerState?: ComputePowerState;
> ```

The post api does scale compute and the state update is to reflect the same

specification/cognitiveservices/CognitiveServices.Management/models.tsp
SubaMoorthy: This should be considered a bug fix. Low priority is not a supported enum value

specification/cognitiveservices/resource-manager/Microsoft.CognitiveServices/preview/2026-03-15-preview/cognitiveservices.json
SubaMoorthy: Yes, back end doesn't return retry after for any of the 202 responses. Can fix in the next api version

M-Hietala (approved):

fmabroukmsft (approved):

rkuthala (approved):

PR #44958, opened 2026-07-24
Add AgentServer shared model ownership customizations

Shivakishore14:
## Summary
- Adds C# AgentServer Responses customizations so AgentServer consumes concrete OpenAI response models from the OpenAI package instead of re-emitting local copies.
- Maps Azure/Foundry-specific response/tool models to Azure.AI.Extensions.OpenAI as the shared owner.
- Updates the AgentServer Responses TypeSpec config for the Azure.AI.AgentServer.Responses SDK generation path.

## Dependency
This PR intentionally targets `rhurey/openai_usage` / #44405 and should be reviewed as a focused follow-up to that OpenAI ownership work, not against `main`.

## Intent
C# SDK customization only; no service wire-contract change is intended.

## Local validation
- Generated Azure.AI.AgentServer.Responses from this TypeSpec branch.
- Built Azure.AI.AgentServer.Responses successfully.
- Exported AgentServer API listings successfully.


specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-extensions-openai/client.tsp
```diff
@@ -666,6 +662,80 @@ namespace Azure.AI.Projects {
 
   //@@alternateType(OpenAI.CompactResource, { identity: "System.Object" }, "csharp");
 
+  // Shared Responses model families not yet owned by the OpenAI .NET SDK.
```
rhurey, line 673: I feel like I"m missing something.

This seems to be saying that when emitting the extension library it should use a type in the extension library instead of emitting it.

Shivakishore14: Thanks for catching this, you’re right. These  @@alternateType  entries in the Extensions emitter were effectively self-referential.

I changed this so Extensions now actually owns/emits those fallback response model families using  @@clientNamespace  /  @@access  /  @@usage . Then AgentServer, as the downstream consumer, uses  @@alternateType  to point at the Extensions-owned public types where appropriate.

So the intended layering is now:
• OpenAI SDK owns models already available publicly in  OpenAI .
• Extensions owns/emits Azure/Foundry response model families not yet owned by OpenAI.
• AgentServer consumes those via alternates instead of re-emitting duplicates.
Also added the experimental annotations for the newly public Extensions fallback models

Shivakishore14:
Superseded by #46013.

That PR adds a dedicated spec folder for `Azure.AI.AgentServer.Responses` as a sibling rather than layering customizations onto the shared extensions folder, which carries additional contracts AgentServer.Responses does not need.

Closing in favour of #46013.

PR #44960, opened 2026-07-24
Foundry merge for the end of July train

glecaros:


dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
```diff
@@ -0,0 +1,91 @@
...
+  @doc("Session ID for this request. Only present for hosted agent responses. Returns the provided session ID or an auto-generated one if not provided in the request.")
+  @header("x-agent-session-id")
+  agentSessionId?: string;
+};
+
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+#suppress "@azure-tools/typespec-azure-core/response-schema-problem" "OpenAI Response schemas can be intentionally combined for documentation clarity"
+@sharedRoute
+op OpenAISharedOperation<
```
catalinaperalta, line 19: Is this template added only because of the sharedRoute decorator?

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
```diff
@@ -0,0 +1,91 @@
...
+  Event,
+  ErrorResponse = ApiErrorResponse
+>(...UserIsolationKeyHeader, ...Params):
+  | SseResponseOf<Event>
+  | AzureRangeErrorResponse<ErrorResponse>;
+
+#suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "OpenAI-based operations are definitionally non-standard"
+#suppress "@azure-tools/typespec-azure-core/response-schema-problem" "OpenAI Response schemas can be intentionally combined for documentation clarity"
+op OpenAIAgentPagedOperation<
```
catalinaperalta, line 68: What makes this template about paged operations? I dont see any specific paging decorators or templates here 🤔

specification/ai-foundry/data-plane/Foundry/src/openai/assistants/routes.generated.tsp
```diff
@@ -0,0 +1,148 @@
...
+   */
+  #suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "OpenAI-based operations are not conventionally versioned"
+  #suppress "@azure-tools/typespec-azure-core/no-response-body" "OpenAI response templates may intentionally model empty success responses."
+  #suppress "deprecated" "Deprecated OpenAI operation shapes intentionally reference deprecated models."
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Auto-suppressed warnings non-applicable rules during import."
+  #deprecated "deprecated"
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Auto-suppressed warnings non-applicable rules during import."
+  @summary("List assistants")
+  @list
+  @get
```
catalinaperalta, line 30: Nit: you only need list
```suggestion
  @list
```

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-projects-agents/client.tsp
```diff
@@ -117,6 +117,8 @@ namespace Azure.AI.Projects {
   @@clientName(ToolArgumentBinding, "StructuredInputToolArgumentBinding");
   @@clientName(FabricIQPreviewTool.server_url, "ServerUri");
   @@clientName(FabricIQPreviewToolboxTool.server_url, "ServerUri");
+  //@@clientName(WebIQPreviewTool.server_url, "ServerUri");
+  //@@clientName(WebIQPreviewToolboxTool.server_url, "ServerUri");
```
catalinaperalta, line 121: should these lines just be removed?

specification/ai-foundry/data-plane/Foundry/compile_all.cmd
catalinaperalta, line 1: I'll defer to @mikeharder on the policy for checking in extra scripts

specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-projects-agents/client.tsp
glecaros: possible, although we expect to re-enable webiq in the next (~2wk-ish) release

specification/ai-foundry/data-plane/Foundry/src/openai/assistants/routes.generated.tsp
glecaros: these files are autogenerated from the OpenAI spec, I'll see if we can fix the emitter over there :)

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
glecaros: these templates are meant to break a cycle between the OpenAI upstream source and the foundry customizations required (the emitter uses a template trusting that we have it defined on this side. So, in this particular case, it is to handle the shared routes emitted.

glecaros: and it seems these are still WIP, not all applied yet

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
glecaros: same as above :)

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
catalinaperalta: Are you sure these templates are going to be used? Or might the design change later? Might be better to add the templates once they are being used. Since there are so many pieces of code in this spec already if the design changes, you might have extra code floating around. I'll leave it up to you.

specification/ai-foundry/data-plane/Foundry/src/common/openai-templates.tsp
catalinaperalta: So does the generator always pass in paged models to this specific template and why not just use one of the other simple templates you have? I say this because the other paged template does provide some paging functionality by itself but this one seems to just be an ordinary operation template without any specific paging params or response templates. The only thing that's different is the template name which doesnt really offer much.

catalinaperalta (approved):
Aside from the scripts check nothing seems blocking

specification/ai-foundry/data-plane/Foundry/compile_all.cmd
mikeharder: we'd prefer to avoid them to keep the repo clean, but if the spec owners find the scripts useful, it's fine.  if they cause problems in the future, we'll re-evaluate.

catalinaperalta (approved):

PR #44974, opened 2026-07-25
TypeSpec Suppressions: CI workflows and dedicated review comment

catalinaperalta:
## TypeSpec Suppressions: CI workflows and dedicated review comment

Introduces the CI workflows and a dedicated pull request comment for the TypeSpec suppressions review feature, following the same pattern as the Namespace Approval flow.

When a PR adds or changes `#suppress` directives for rules listed in `eng/tools/typespec-suppressions/check-rules.json`, the analyze workflow flags them and a sticky **TypeSpec Suppressions Review** comment is posted to the PR. The comment refreshes its ✅/❌ state on the `Approved-TypeSpecSuppression` label. When no suppressions require review, no comment is posted (an existing comment is resolved).

### What's included

**Workflows**
- `typespec-suppressions-code.yaml` — "Analyze Code" workflow. Runs the `typespec-suppressions` CLI over impacted TypeSpec folders on `pull_request`, emits inline `::warning` annotations, and uploads the `typespec-suppressions-report` artifact.
- `typespec-suppressions-comment.yaml` — posts/updates the dedicated sticky comment on `workflow_run` completion of Analyze Code and on `pull_request_target` `labeled`/`unlabeled` (to refresh approval state).
- `typespec-suppressions-status.yaml` — sets the required "TypeSpec Suppressions" commit status (overridable by the `Approved-TypeSpecSuppression` label) via the shared `_reusable-set-check-status.yaml`.
- `typespec-suppressions-test.yaml` — test workflow.

**Comment module** (`.github/workflows/src/typespec-suppressions/`)
- `suppressions-comment.js` — renders the comment body; locates the latest Analyze Code run by `head_sha` and downloads its report artifact.
- `post-results.js` — entry point (`postSuppressionsResults`) that resolves PR context/labels and posts, updates, or resolves the sticky comment.

**Tests**
- `test/typespec-suppressions/{suppressions-comment,post-results}.test.js`
- `test/mocks.js` — adds `issues.{createComment,deleteComment,listComments,updateComment}` mocks.

### Gating behavior

The Analyze Code workflow does **not** pass `--fail-on-approval`, so the CLI always exits 0 and the "TypeSpec Suppressions" check is **always green** in this PR. The feature ships as **comment + non-blocking CI check** only. Enabling gating (making the check block merges until approved) is deferred to a future change that adds `--fail-on-approval`.

### Dependencies / companions

- The `typespec-suppressions` CLI is already in `main` under `eng/tools/typespec-suppressions/`.
- Requires `eng/tools/typespec-suppressions/check-rules.json` (the review ruleset), added in #44976.


markcowl (approved):

.github/workflows/typespec-suppressions-test.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  typespec-suppressions:
+    uses: ./.github/workflows/_reusable-eng-tools-test.yaml
+    with:
+      package: typespec-suppressions
+      lint: false
```
mikeharder: please enable lint soon, if it's not already

.github/workflows/typespec-suppressions-comment.yaml
```diff
@@ -0,0 +1,46 @@
...
+  actions: read
+  contents: read
+  pull-requests: write
+
+jobs:
+  post-comment:
+    name: Post TypeSpec Suppressions Comment
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
```
mikeharder:
```suggestion
      - uses: actions/checkout@v7
```

.github/workflows/typespec-suppressions-code.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+jobs:
+  typespec-suppressions:
+    name: "TypeSpec Suppressions - Analyze Code"
+    runs-on: ubuntu-24.04
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
mikeharder:
```suggestion
```

The default ref should work, and it can be risky to override the default unless it's really necessary.

.github/workflows/typespec-suppressions-code.yaml
```diff
@@ -0,0 +1,111 @@
...
+          fetch-depth: 0
+
+      - name: Setup Node and install deps
+        uses: ./.github/actions/setup-node-install-deps
+
+      - name: Get impacted TypeSpec folders
+        shell: pwsh
+        run: |
+          $ignoreCoreFiles = -not (@('main', 'RPSaaSMaster') -contains $Env:GITHUB_BASE_REF)
```
mikeharder, line 30: we should eventually port get-typespec-folders.ps1 to .github/shared as a JS API

.github/workflows/typespec-suppressions-code.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+          Write-Host "Found $($typespecFolders.Count) impacted TypeSpec folders."
+          foreach ($typespecFolder in $typespecFolders) {
+            Write-Host "  $typespecFolder"
+          }
+
+      - name: Run TypeSpec suppressions analysis
+        shell: bash
+        env:
+          # Emit inline PR `::warning` annotations only for meaningful analysis runs.
```
mikeharder, line 50: this feels overly complex, and I don't think it's a problem for typical GH actions that emit annotations.  maybe it's because most source-code-based actions don't re-run on "edited"?  but its fine for now.

.github/workflows/typespec-suppressions-comment.yaml
mikeharder, line 1: could you try combining the "-comment" and "-status" workflows into a single, multi-job workflow (with permissions scoped to each job)?  we already have so many workflows, if we could keep this to 2 instead of 3, it would be nice

mikeharder (changes requested):
see comments, and also please fix lint errors in test code

.github/workflows/typespec-suppressions-code.yaml
catalinaperalta: The reason for the extra  EMIT_ANNOTATIONS  guard:  edited  fires on any PR edit, including title/description changes. GitHub anchors annotations to a commit (the PR's merge commit) and aggregates them from every check run on that commit without de-duping identical  (path, line, message)  records.

• On a base-branch change (or a push), the merge commit is recomputed → new SHA → the re-run's annotations land on a fresh surface → no duplication.
• On a title/description edit, head and base are unchanged → the merge commit is byte-identical → re-emitting piles a second identical set of annotations onto the same commit → GitHub shows them twice.

So the guard  github.event.action != 'edited' || github.event.changes.base != null  just says "re-emit only when the underlying diff could actually have changed." Every other analyzer here re-emits on description edits without guarding — I wanted to avoid the duplicated inline warnings.

catalinaperalta: I've also personally found it annoying when a check retriggers on a PR description/title change since it makes waiting longer, but for now this is only for the annotations so they're not duplicated. Happy to revisit later on if there's a better solution!

.github/workflows/typespec-suppressions-code.yaml
catalinaperalta: removed

.github/workflows/typespec-suppressions-test.yaml
catalinaperalta: enabled

.github/workflows/typespec-suppressions-comment.yaml
catalinaperalta: done!

catalinaperalta:
The Github Actions - Test failure is unrelated to my changes: 
image

.github/workflows/typespec-suppressions-code.yaml
mikeharder: It's fine for now, but I'd still like to understand why our check has code to deal with this, but most checks don't.  Something feels wrong.

catalinaperalta:
This feature was merged in #45089

Issue #44982, opened 2026-07-25
sdk-generation-agent: GITHUB_TOKEN in workflow-level `env` is visible to the agent container


xirzec:
## Summary

`gh aw compile` reports a warning on `.github/workflows/sdk-generation-agent.md` that I don't think can be resolved without input from the workflow owner:

```
⚠ Warning: secrets detected in 'env' section will be leaked to the agent container.
  Found: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN || secrets.GITHUB_TOKEN }}
  Consider using engine-specific secret configuration instead.
```

The source is the workflow-level `env` block:

```yaml
env:
  AZSDK_CLI_PATH: /tmp/bin
  AZURE_CLIENT_ID: ...
  AZURE_TENANT_ID: ...
  GITHUB_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN || secrets.GITHUB_TOKEN }}
```

This was added deliberately in e28e6cd2e53f ("Sdk generation agent auth changes", #40598), so I did not want to change it unilaterally.

## Why it matters

`sdk-generation-agent` is triggered by `issues` (`opened`, `labeled`) and `issue_comment` (`created`). The workflow's own prompt acknowledges that issue and comment bodies are untrusted input. Because `GITHUB_TOKEN` is set at the workflow level, it is present in the environment of the agent container, so anything that successfully influences the agent's behavior can read a PAT (`GITHUB_PERSONAL_ACCESS_TOKEN`) whose scopes are likely broader than the per-run `secrets.GITHUB_TOKEN`.

Mitigations already in place: `tools.github.min-integrity: approved`, and the `Run sdk generation` label gate.

## What I already ruled out

- **`gh aw fix`** refuses to auto-fix this one — it reports `codemod top-level-env-secrets-guided-error ... Manual fix required: move the secret into engine-specific secret configuration.`
- **Moving it to `engine.env`** would take it away from the imported setup steps (`shared-github-aw-imports/install_azsdk_cli_import.md`, `global_networks_auth_import.md`), which inherit the workflow-level `env`. Per the [gh-aw engines reference](https://github.github.com/gh-aw/reference/engines/), only `COPILOT_PROVIDER_*` credentials are treated as non-leaking in `engine.env` anyway, so this may not even clear the warning.
[...]

PR #44983, opened 2026-07-25
Recompile agentic workflows with gh-aw v0.83.1 and clear compiler warnings

xirzec:
Recompiles the two agentic workflows with the current `gh aw` extension (**v0.68.1 → v0.83.1**) and clears the compiler warnings that had accumulated since they were last built.

## Commit 1 — recompile with gh-aw v0.83.1

- `hero-scenarios.lock.yml` and `sdk-generation-agent.lock.yml` regenerated (metadata schema v3 → v4).
- `.github/aw/actions-lock.json`: `gh-aw-actions/setup` v0.68.1 → v0.83.1, plus the new `setup-cli@v0.83.1` pin.
- **New file** `.github/workflows/agentics-maintenance.yml`. This is auto-generated by the newer compiler (`pkg/workflow/maintenance_workflow.go`) because `hero-scenarios` uses `cache-memory` / `repo-memory`. It's marked `DO NOT EDIT`, so it's added to `.github/.prettierignore` alongside the lock files — otherwise `prettier . --check` fails on it. It can be suppressed later with `{"maintenance": false}` in `.github/workflows/aw.json` if we decide we don't want the daily job.

Triggers, `permissions`, and job graphs are unchanged by the recompile.

## Commit 2 — clear compiler warnings

| Warning | Resolution |
| --- | --- |
| `hero-scenarios`: `pull_request_target` + checkout is "extremely insecure" | Declared `checkout: false` |
| `actions/checkout@v6` outdated | Bumped to `@v7` |
| `actions/github-script@v7` outdated | Bumped to `@v9` |
| `sdk-generation-agent`: `id-token: write` | Intentional — documented inline |
| `sdk-generation-agent`: `GITHUB_TOKEN` in `env` | Not fixable here — see #44982 |

On the `checkout: false` change: the compiled lock body is **byte-identical** afterwards, because the agent never checked out source in the first place. `hero-scenarios-guidelines.md` drives everything through the GitHub MCP `repos`/`pull_requests` toolsets and explicitly instructs *"Do NOT browse the wider repository."* So this declares existing intent rather than changing behavior — it just stops `pull_request_target` from ever being able to pick up fork code.

Also pruned the now-unused `actions/checkout@v6` and `actions/github-script@v7` pins from `.github/aw/actions-lock.json` (confirmed they don't come back on recompile).

## Known remaining warning

`sdk-generation-agent` still sets `GITHUB_TOKEN: ${{ secrets.GITHUB_PERSONAL_ACCESS_TOKEN || secrets.GITHUB_TOKEN }}` at the workflow level, which the compiler flags as visible to the agent container. That was added deliberately in #40598, `gh aw fix` refuses to auto-fix it, and it can't move to `engine.env` because the imported setup steps read it too. Filed #44982 for @praveenkuttappan rather than changing the auth setup unilaterally; this PR only adds an explanatory comment above it.

## Verification

- `gh aw compile` → 0 errors; warnings 5 → 2 (both intentional, both documented).
- Recompile is idempotent — a second run produces no diff.
- Both lock files parse as valid YAML; `on:`, `permissions:`, and job graphs diffed against the previous locks are identical except `sdk-generation-agent`'s `agent` job, which changes only by the two action SHA bumps.
- `npm run format:check` and `.github/workflows/test/workflows.test.js` pass.
- `gh aw lint` (actionlint) requires Docker, which wasn't available locally — leaving that to CI.

<sub>Authored by GitHub Copilot on behalf of @xirzec.</sub>


mikeharder:
@copilot+claude-opus-5: add word "agentics" to the root cspell.yaml

.github/.prettierignore
```diff
@@ -26,4 +30,4 @@ PULL_REQUEST_TEMPLATE
 workflows/post-apiview.yml
 
 # synced GitHub skills
-.github/skills/azsdk-common-*
\ No newline at end of file
+.github/skills/azsdk-common-*
```
mikeharder, line 33: unrelated cleanup, but easier to normalize to trailing newline

PR #44986, opened 2026-07-26
[Microsoft.Security] Add DefenderForStorageSettings 2026-03-01-preview (per-extension scanners)

raniwishahy:
## Summary

Adds a new preview API version **2026-03-01-preview** to the `Microsoft.Security` `DefenderForStorageSettings` TypeSpec project, built on top of the merged **2026-01-01-preview**.

Authored in TypeSpec. Structured **copy-first** for reviewability:

1. **Initial commit — pure copy.** Duplicates 2026-01-01-preview as 2026-03-01-preview with no behavior change (adds the version to the enum, copies examples, adds the readme tag, points the composite default tag at 2026-03-01-preview). The generated `defenderForStorageSettings.json` is identical to 2026-01-01-preview except the version string.
2. **Follow-up commit — additions.** Applies the actual 2026-03-01-preview changes on top (see below), so the review diff isolates exactly what is new.

## What's new in 2026-03-01-preview (per-extension scanner identity)

- **`scannerResourceId`** (nullable ARM id of `Microsoft.Security/securityOperators` or `Microsoft.Security/dataScanners`) on both `MalwareScanningProperties` and `SensitiveDataDiscoveryProperties` — the scanner identity used for malware scanning / sensitive data discovery.
- **`managedBy`** (read-only `ManagedByType` enum: `None`, `MicrosoftDefenderXDR`, `SubscriptionSettingsEnforcer`) on `DefenderForStorageSettingProperties` — indicates whether the resource's settings are managed by a higher MDC enforcement scope.

All additions are declared with `@added(v2026_03_01_preview)`, so the published 2025-09-01-preview and 2026-01-01-preview outputs are byte-for-byte unchanged. The changes are additive and backwards-compatible.

## Validation

- `tsp compile` clean; TypeSpec validation passes.
- 2025-09-01-preview and 2026-01-01-preview outputs confirmed unchanged.
- 2026-03-01-preview differs from 2026-01-01-preview by exactly the two additions above.

### DIFF
--- a/specification/security/resource-manager/Microsoft.Security/Security/DefenderForStorageAPI/models.tsp
+++ b/specification/security/resource-manager/Microsoft.Security/Security/DefenderForStorageAPI/models.tsp
@@ -49,6 +49,29 @@ union AutomatedResponseType {
 }
 
 /**
+ * Indicates whether this resource's Defender for Storage settings are managed by a higher MDC enforcement scope and which component is managing them.
+ */
+@added(Versions.v2026_03_01_preview)
+union ManagedByType {
+  string,
+
+  /**
+   * The resource's Defender for Storage settings are not managed by a higher enforcement scope.
+   */
+  None: "None",
+
+  /**
+   * The resource's Defender for Storage settings are managed by Microsoft Defender XDR.
+   */
+  MicrosoftDefenderXDR: "MicrosoftDefenderXDR",
+
+  /**
+   * The resource's Defender for Storage settings are managed by the subscription-level settings enforcer.
+   */
+  SubscriptionSettingsEnforcer: "SubscriptionSettingsEnforcer",
+}
+
+/**
  * List of Defender for Storage settings.
  */
 model DefenderForStorageSettingList {
@@ -90,6 +113,13 @@ model DefenderForStorageSettingProperties {
    * Indicates whether the settings defined for this storage account should override the settings defined for the subscription.
    */
   overrideSubscriptionLevelSettings?: boolean;
+
+  /**
+   * Indicates whether this resource's Defender for Storage settings are managed by a higher MDC enforcement scope and which component is managing them. When set to a value other than 'None', the resource-level settings become read-only and can only be modified by the managing component. Possible values: 'None' (not managed), 'MicrosoftDefenderXDR' (managed by Microsoft Defender XDR), 'SubscriptionSettingsEnforcer' (managed by subscription-level settings enforcer).
+   */
+  @added(Versions.v2026_03_01_preview)
+  @visibility(Lifecycle.Read)
+  managedBy?: ManagedByType;
 }
 
 /**
@@ -119,6 +149,22 @@ model MalwareScanningProperties {
    * Optional. Specifies the automated response action to take when malware is detected.
    */
   automatedResponse?: AutomatedResponseType = AutomatedResponseType.None;
+
+  /**
+   * Optional. Resource id of the scanner resource (e.g., securityOperator or dataScanner) whose identity will be used for malware scanning.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-nullable" "The scanner resource id is nullable in the API contract to allow explicitly clearing a previously configured scanner."
+  @added(Versions.v2026_03_01_preview)
+  scannerResourceId?:
+    | Azure.Core.armResourceIdentifier<[
+        {
+          type: "Microsoft.Security/securityOperators";
+        },
+        {
+          type: "Microsoft.Security/dataScanners";
+        }
+      ]>
+    | null;
 
   /**
    * Upon failure or partial success. Additional data describing Malware Scanning enable/disable operation.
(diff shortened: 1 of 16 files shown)


specification/security/resource-manager/Microsoft.Security/Security/preview/2026-03-01-preview/defenderForStorageSettings.json
```diff
@@ -0,0 +1,1159 @@
...
+        },
+        "sensitiveDataDiscovery": {
+          "$ref": "#/definitions/SensitiveDataDiscoveryProperties",
+          "description": "Properties of Sensitive Data Discovery."
+        },
+        "overrideSubscriptionLevelSettings": {
+          "type": "boolean",
+          "description": "Indicates whether the settings defined for this storage account should override the settings defined for the subscription."
+        },
+        "managedBy": {
```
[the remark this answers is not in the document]
raniwishahy: @psah434 
Thanks for flagging this. `properties.managedBy` is an intentional, previously-reviewed naming decision, and I'd like to keep it. Reasoning:

**1. No collision with the ARM envelope `managedBy`.**
The reserved ARM envelope `managedBy` sits at the resource root (`$.managedBy`) and carries ARM-id/lifecycle semantics (immutable resource id of a managing resource, skipped during complete-mode deletion, etc.). This property is a read-only enum at a *different* JSON path (`$.properties.managedBy`) with a domain-specific contract: it reports which MDC enforcement scope (if any) has locked this resource's Defender for Storage settings - `None` | `MicrosoftDefenderXDR` | `SubscriptionSettingsEnforcer`. This resource neither defines nor uses the top-level ARM `managedBy` envelope property, so there is no actual ambiguity for ARM tooling.

**2. Established, already-shipped precedent in this same RP.**
`Microsoft.Security/Pricings` already exposes `managedBy` inside its `properties` bag as a read-only string with the same "who manages this configuration outside direct ARM" semantics - see `specification/security/resource-manager/Microsoft.Security/Security/preview/2025-10-01-preview/pricings.json` (`PricingProperties.managedBy`, lines 482-486), which is merged in this repo today. Keeping the name preserves consistency across Defender for Cloud APIs.

**3. This exact property was already ARM-reviewed and approved for this same 2026-03-01-preview version.**
`managedBy` is new in 2026-03-01-preview (it does not exist in 2026-01-01-preview). This same 2026-03-01-preview spec was first authored and reviewed in Swagger form in the private spec repo (`Azure/azure-rest-api-specs-pr#26996`), where this identical Section 2.5 concern was raised; after discussion the ARM reviewers agreed to allow `managedBy`, on the precedent/semantic grounds above. This public PR is the TypeSpec authoring of that same, already-reviewed version - so while the automated check flags it as `[NEW]` (it diffs only against the public 2026-01-01-preview base and can't see the private-repo review history), the name is not an un-reviewed design choice. Existing consumers of the preview already read `properties.managedBy`.

**4. The suggested alternatives are semantically weaker.**
`settingsManagedBy` / `managedByComponent` / `managementSource` were considered. The field represents the management/ownership authority that *locks* the settings, which "managed by" captures accurately; "source"/"component" do not.

Given no envelope conflict, an already-approved sibling API using the identical pattern, prior ARM sign-off on this same version in the private spec repo, and existing clients on this name - could we keep `managedBy` and resolve this thread? Happy to add a documented suppression with this justification if you'd prefer it captured inline.

specification/security/resource-manager/Microsoft.Security/Security/DefenderForStorageAPI/models.tsp
```diff
@@ -90,6 +113,13 @@ model DefenderForStorageSettingProperties {
    * Indicates whether the settings defined for this storage account should override the settings defined for the subscription.
    */
   overrideSubscriptionLevelSettings?: boolean;
+
+  /**
+   * Indicates whether this resource's Defender for Storage settings are managed by a higher MDC enforcement scope and which component is managing them. When set to a value other than 'None', the resource-level settings become read-only and can only be modified by the managing component. Possible values: 'None' (not managed), 'MicrosoftDefenderXDR' (managed by Microsoft Defender XDR), 'SubscriptionSettingsEnforcer' (managed by subscription-level settings enforcer).
+   */
+  @added(Versions.v2026_03_01_preview)
+  @visibility(Lifecycle.Read)
+  managedBy?: ManagedByType;
```
eswatad02, line 122: make sure this must be nullable. given that this is already readonly param.

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-03-01-preview/defenderForStorageSettings.json
eswatad02, line 130: what is this ?

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-03-01-preview/defenderForStorageSettings.json
```diff
@@ -0,0 +1,1159 @@
+{
```
eswatad02, line 1: something wrong is going on here. please take another look on this file.

eswatad02 (changes requested):

specification/security/resource-manager/Microsoft.Security/Security/DefenderForStorageAPI/models.tsp
raniwishahy: Correction for what I said earlier, `? ` doesn't mean nullable, it means optional, `| null` means nullable.
Here's what each property means:

 - Read-only means the service supplies the value; clients don’t set it.
 - Optional ( ? ) means the response may omit the property.
 - Nullable ( | null ) would additionally permit an explicit  "managedBy": null  in the response.
 
 So, the correct option here is Read-only, it's required in the response so we shouldn't leave it as optional.

PR #45008, opened 2026-07-27
Add knowledge toolbox and web source APIs

erjohnms:
## Summary
- add `default` and `knowledge_base` toolbox kinds
- add knowledge-source references and extensible knowledge retrieval configuration
- add project-scoped create, get, list, update, and delete APIs for web knowledge sources
- proxy Azure AI Search knowledge-source properties through wildcard contract fields while strongly typing only `name` and discriminators

## Compatibility
- missing toolbox `kind` remains equivalent to `default`
- `tools` becomes structurally optional and is forbidden when `kind` is `knowledge_base`; service validation enforces kind-specific required and forbidden fields
- default toolboxes continue to require tools or skills through service validation

## Validation
- TypeSpec compiles with warnings treated as errors
- v1 and virtual-public-preview OpenAPI artifacts regenerated
- generated artifacts are byte-stable on regeneration

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
@@ -266,7 +266,48 @@ model ToolboxSkillReference extends ToolboxSkill {
   version?: string;
 }
 
-@doc("A toolbox that stores reusable tool definitions for agents.")
+/**
+ * The kind of content stored in a toolbox.
+ */
+union ToolboxKind {
+  /**
+   * A toolbox containing reusable tools and skills.
+   */
+  default: "default",
+
+  /**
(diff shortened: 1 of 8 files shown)


specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -266,7 +266,48 @@ model ToolboxSkillReference extends ToolboxSkill {
...
+union ToolboxKind {
+  /**
+   * A toolbox containing reusable tools and skills.
+   */
+  default: "default",
+
+  /**
+   * A toolbox providing knowledge retrieval over configured knowledge sources.
+   */
+  knowledge_base: "knowledge_base",
```
viswabalaji, line 281: why do we need this toolbox kind vs just having knowledge base as another option in toolbox in addition to tools/skills?

PR #45015, opened 2026-07-27
Add AKS AI Manager CustomAIModel and ModelSource Handling

bfoley13:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
```diff
@@ -0,0 +1,132 @@
...
+    The model identifier, interpreted per the referenced ModelSource type.
+    For `HuggingFace` sources this is the upstream `<org>/<repo>` id, e.g.
+    `meta-llama/Llama-2-7b-chat`. For `MicrosoftFoundry` sources this is
+    `modelName/version`, e.g. `private-llama/1` (the platform splits model and
+    version on the final `/`). Immutable after creation.
+    """)
+  modelId: string;
+
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  @doc("ARM resource id of the ModelSource to use when pulling artifacts. Required for `MicrosoftFoundry` sources and for gated/private `HuggingFace` repositories; omitted for public models. Immutable after creation.")
```
razvanbadea-msft: >ARM [](http://example.com/codeflow?start=8&length=3)

you can replace with Azure to avoid acronyms in descriptions.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
```diff
@@ -0,0 +1,132 @@
...
+  modelSourceResourceId?: ModelSourceResourceId;
+
+  @doc("Optional. Free-form description of the model. Mutable via a full-replace PUT.")
+  description?: string;
```
razvanbadea-msft, line 58: >description [](http://example.com/codeflow?start=2&length=11)

is there a max limit for this?

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
```diff
@@ -0,0 +1,132 @@
...
+  modelFileSize?: string;
+
+  @visibility(Lifecycle.Read)
+  @doc("The model architecture classes, e.g. `[\"Phi3ForCausalLM\", \"LlamaForCausalLM\"]`. Determines compatibility with the underlying LLM runtime.")
+  architectures?: string[];
+
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Open-ended set of architecture parameters extracted from config.json; the key set varies by model family."
+  @visibility(Lifecycle.Read)
+  @doc("Critical model architecture parameters extracted from config.json. Keys include: hiddenSize, numHiddenLayers, numAttentionHeads, numKeyValueHeads. All values are integers.")
+  architectureParams?: Record<int32>;
```
razvanbadea-msft: >architectureParams [](http://example.com/codeflow?start=2&length=18)

Use of additionalProperties is not allowed for properties owned by the service. The only time its ok to use it is when the properties are pass thru (user defined) and not subject to any validations. Please make this an array or provide an explanation for why you need this.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
```diff
@@ -26,7 +26,33 @@ These are the global settings for the AI Manager API.
...
+    from: aimanagers.json
+    where: $.definitions.AIManagerNamespaceProperties.properties.annotations
+    reason: Annotations are a key/value map that is passed through to the underlying Kubernetes model.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.ModelDeploymentOverrides.properties.values
+    reason: Free-form override key/value pairs documented per release.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.CustomAIModelSpec.properties.architectureParams
```
razvanbadea-msft: >architectureParams [](http://example.com/codeflow?start=54&length=18)

Use of additionalProperties is not allowed for properties owned by the service. The only time its ok to use it is when the properties are pass thru (user defined) and not subject to any validations. Please make this an array or provide an explanation for why you need this.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
```diff
@@ -26,7 +26,33 @@ These are the global settings for the AI Manager API.
...
+    from: aimanagers.json
+    where: $.definitions.AIManagerNamespaceProperties.properties.labels
+    reason: Labels are a key/value map that is passed through to the underlying Kubernetes model.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.AIManagerNamespaceProperties.properties.annotations
+    reason: Annotations are a key/value map that is passed through to the underlying Kubernetes model.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.ModelDeploymentOverrides.properties.values
```
razvanbadea-msft, line 50: >values [](http://example.com/codeflow?start=61&length=6)

Use of additionalProperties is not allowed for properties owned by the service. The only time its ok to use it is when the properties are pass thru (user defined) and not subject to any validations. Please make this an array or provide an explanation for why you need this.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modeldeployment.tsp
```diff
@@ -22,6 +22,9 @@ scalar AIModelResourceId
   extends Azure.Core.armResourceIdentifier<[
     {
       type: "Microsoft.ContainerService/aiModels",
+    },
+    {
+      type: "Microsoft.ContainerService/aiManagers/customAIModels",
```
razvanbadea-msft, line 26: **[NEW] 🟠 Warning** **[[`added` version misalignment](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `modeldeployment.tsp` — line 22-29 (the `AIModelResourceId` scalar) — adding `CustomAIModel` support retroactively mutates the published `2026-05-02-preview` swagger.

**Classification reasoning:** [NEW] — `customAIModels` is `@added(v2026_08_02_preview)`; the base `2026-05-02-preview` had neither the `customAIModels` allowed type nor the rewritten description.

**Issue:** Because the shared `AIModelResourceId` scalar and `ModelDeploymentProperties.modelResourceId` are `@added(v2026_05_02_preview)` and not version-gated per-type, two edits here bleed backward into the published `2026-05-02-preview` output (PR diff to that file: `+4/-1`): (a) `allowedResources` gained `Microsoft.ContainerService/aiManagers/customAIModels`, and (b) the `modelResourceId` description changed from "Phase 1 accepts an `AIModel` resource id only" to text referencing `CustomAIModel`/`MicrosoftFoundry` — concepts absent from the 2026-05-02-preview surface. Additive / not CI-breaking, but it rewrites a published version's contract.

**Suggested fix:** Version-gate so 05-02 output is unchanged — e.g. introduce a new scalar added in `v2026_08_02_preview` that includes the `customAIModels` type and use it only for the new version, leaving `AIModelResourceId` unchanged for 05-02. Or explicitly confirm the 05-02 revision is intended.

razvanbadea-msft (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`67b3b29`](https://github.com/Azure/azure-rest-api-specs/pull/45015/commits/67b3b29d97e3b677f404666838208802af9e8e12). See inline comments for findings 1-3.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
```diff
@@ -26,7 +26,33 @@ These are the global settings for the AI Manager API.
...
+    reason: Free-form override key/value pairs documented per release.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.CustomAIModelSpec.properties.architectureParams
+    reason: Key/value pairs or model architectire parameters.
```
razvanbadea-msft: **[NEW] 🔵 Suggestion** **[[Spelling anti-pattern](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)] / [[Suppression quality §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `readme.md` — line 55 — typos in new suppression reason.

**Classification reasoning:** [NEW] — this `AvoidAdditionalProperties` suppression targets `CustomAIModelSpec.architectureParams`, a model added in `2026-08-02-preview`.

**Issue:** `reason: Key/value pairs or model architectire parameters.` contains two typos — "architectire" should be "architecture", and "or" reads as if it should be "of". The reason is also thin.

**Suggested fix:**
`reason: Open-ended integer key/value set of model architecture parameters extracted from config.json; read-only, key set varies by model family.`

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
```diff
@@ -76,6 +108,42 @@ model InlineCredential {
   value: string;
 }
 
+@added(Versions.v2026_08_02_preview)
+@doc("A credential backed by a customer-owned user-assigned managed identity. The platform authenticates to the model source as this identity via Workload Identity; no secret is stored. Because the platform generally has no read access to a customer's identity, the `clientId` is supplied by the caller rather than resolved by the service.")
+model ManagedIdentityCredential {
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  @doc("The ARM resource id of the user-assigned managed identity to authenticate with. Only user-assigned identities are supported. The RP first-party application must be granted `Managed Identity Operator` on this identity so it can be federated for Workload Identity.")
+  resourceId: UserAssignedIdentityResourceId;
```
razvanbadea-msft, line 104: **[NEW] 🔵 Suggestion** **[[OAPI027 — write-only forbidden](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi027)]** `modelsource.tsp` — line 113-125 (`ManagedIdentityCredential`) — non-secret identity references are write-only.

**Classification reasoning:** [NEW] — `ManagedIdentityCredential` is `@added(v2026_08_02_preview)`.

**Issue:** `ModelSourceProperties.credential` (line 61-63) is `@visibility(Lifecycle.Create, Lifecycle.Update)` (write-only), so its whole subtree — including the non-secret `resourceId` / `clientId` / `tenantId` identity references — is never returned on GET. A customer cannot read back which managed identity they configured.

**Suggested fix:** Consider exposing the non-secret managed-identity reference fields on read (or split the identity reference out of the write-only `credential` block).

**Note:** OAPI027 states non-secret write-only properties MUST NOT exist — stronger than Suggestion. Kept as Suggestion because LintDiff likely does not fire on a subtree nested under a write-only parent (probably not CI-enforced); consider raising to Warning.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
bfoley13: replaced

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
bfoley13: I have removed this item

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
bfoley13: I have removed this item

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
bfoley13: This was copied from previous versions and not new to the API, is there something I need to do for existing exceptions? I removed the one I was adding

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modeldeployment.tsp
bfoley13: I have fixed this but I need to understand what happens about the doc update, is it okay or is there a way to only update the doc on later versions?

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
bfoley13: I have removed this field

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
bfoley13: I have removed the restrictive permissions

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
bfoley13: added 250 max limit

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
```diff
@@ -26,7 +26,33 @@ These are the global settings for the AI Manager API.
...
+    reason: Labels are a key/value map that is passed through to the underlying Kubernetes model.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.AIManagerNamespaceProperties.properties.annotations
+    reason: Annotations are a key/value map that is passed through to the underlying Kubernetes model.
+  - code: AvoidAdditionalProperties
+    from: aimanagers.json
+    where: $.definitions.ModelDeploymentOverrides.properties.values
+    reason: Free-form override key/value pairs documented per release.
+  - code: AvoidAdditionalProperties
```
bfoley13, line 52: We have decided on using the additionalProperties for a new field that represents the config.json for a specific model input by users. (ex: https://huggingface.co/moonshotai/Kimi-K3/blob/main/config.json). They are not uniform from model to model and we need this from users for private repositories (HF/Foundry)

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
```diff
@@ -42,6 +61,11 @@ model ModelSourceProperties {
   @visibility(Lifecycle.Create, Lifecycle.Update)
   @doc("Credential the platform uses to authenticate to the source. Optional for public sources (e.g. ungated Hugging Face models).")
   credential?: CredentialValue;
```
ravimeda, line 59: **[NEW] 🔴 Blocking** **[[OAPI027](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi027-write-only-properties-are-forbidden)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp` — line 61-63; emitted at `preview/2026-08-02-preview/aimanagers.json` line 3319-3326, JSON path `$.definitions.ModelSourceProperties.properties.credential`

**Classification reasoning:** [NEW] — at base SHA `edf0326b` the `credential` property was already write-only, but its only reachable leaf was `InlineCredential.value`, annotated `x-ms-secret: true` and therefore fully covered by the OAPI027 secret exception. `CredentialValue.managedIdentity` is `@added(Versions.v2026_08_02_preview)` and introduces non-secret leaves into that write-only subtree, so the violating condition is new in this version.

**Issue:** `ModelSourceProperties.credential` is `@visibility(Lifecycle.Create, Lifecycle.Update)`, emitting `"x-ms-mutability": ["update","create"]` with no `x-ms-secret: true` on the property itself. OAPI027 states that a property with `x-ms-mutability: ["create", "update"]` without `"read"` is write-only and is not allowed, and that every non-secret property that can be set on a resource must also be readable via GET. This PR adds `CredentialValue.managedIdentity` (`modelsource.tsp` line 97-99, emitted at swagger line 2665-2668) — a non-secret object carrying `resourceId`, `clientId`, and `tenantId` (swagger line 2956-2978) — into that write-only subtree. A customer can never read back which managed identity a `ModelSource` is configured with: it is absent from every GET, PUT and LIST response. This also breaks ARM Template What-If (the PUT body carries `credential.managedIdentity`; the GET response never does, so every redeployment reports a false difference) and makes the configuration invisible to Azure Resource Graph.

**Prior thread:** raised at Suggestion severity in https://github.com/Azure/azure-rest-api-specs/pull/45015#discussion_r3665361890 and marked resolved with "I have removed the restrictive permissions". That change removed `@visibility(Lifecycle.Create, Lifecycle.Update)` from the three `ManagedIdentityCredential` **members**, but the emitted mutability is driven by the **parent** `credential` property, which still carries the write-only visibility at head SHA `f31b165e`. The emitted shape is unchanged.

**Fix:** Split the non-secret identity reference out of the write-only `credential` block:

```tsp
@visibility(Lifecycle.Create, Lifecycle.Update)
@doc("Secret credential the platform uses to authenticate to the source. Optional for public sources.")
credential?: CredentialValue;

@added(Versions.v2026_08_02_preview)
@doc("User-assigned managed identity the platform authenticates as. Required for `MicrosoftFoundry` sources.")
managedIdentity?: ManagedIdentityCredential;
```

If `managedIdentity` must stay inside `CredentialValue`, then `credential` cannot remain write-only.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
```diff
@@ -51,11 +75,15 @@ union ModelSourceType {
...
+
+  @added(Versions.v2026_08_02_preview)
+  @doc("A Microsoft Foundry project. Tenant-private models produced by internal training pipelines. Authenticated via a customer user-assigned managed identity (referenced in the credential and granted the `Foundry User` role on the project).")
+  MicrosoftFoundry: "MicrosoftFoundry",
 }
 
 @added(Versions.v2026_05_02_preview)
 @doc("""
-  A credential value. Exactly one variant must be set.
+  A credential value.
```
ravimeda: **[NEW] 🟠 Warning** **[[Doc comment accuracy — TypeSpec §5 Common Anti-Patterns](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp` — line 85-92; emitted at `preview/2026-08-02-preview/aimanagers.json` line 2657-2659, JSON path `$.definitions.CredentialValue.description`

**Classification reasoning:** [NEW] — the contradiction is created in this version by the addition of `managedIdentity` to the same model.

**Issue:** TypeSpec §5 requires that doc comments accurately describe the element they annotate. The `@doc` on `CredentialValue` still reads "In the current API version, only the `inline` variant is supported. Future API versions are expected to add additional credential kinds (for example, managed identity and Key Vault secret references) as sibling variants on this model." In `2026-08-02-preview` that is no longer true — this PR adds `managedIdentity` to the very same model (emitted at swagger line 2665-2668). The description tells SDK consumers and doc readers that the variant visible in the schema does not exist. The PR edited the first sentence of this doc block and left the stale one. Relatedly, the model now has two optional variants, no discriminator, no `required` entry, and no documented rejection semantics, while the sibling `microsoftFoundry` property (line 67) sets the precedent by naming `InvalidModelSourceConfiguration`.

**Fix:** Rewrite the `@doc` on `CredentialValue` so it covers both variants, the mutual-exclusion constraint, and the error code:

```tsp
@doc("""
  A credential value. Exactly one variant must be set: `inline` for
  token-based sources (e.g. gated Hugging Face repositories), or
  `managedIdentity` for `MicrosoftFoundry` sources. Enforced at
  request-validation time with error code `InvalidModelSourceConfiguration`.
  Future API versions may add additional credential kinds (for example, Key
  Vault secret references) as sibling variants on this model.
  """)
```

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/preview/2026-08-02-preview/examples/ModelSources_CreateOrUpdate.json
```diff
@@ -0,0 +1,68 @@
...
+    "api-version": "2026-08-02-preview",
+    "subscriptionId": "00000000-0000-0000-0000-000000000000",
+    "resourceGroupName": "rgaimanagers",
+    "If-Match": "\"00000000-0000-0000-0000-000000000000\"",
+    "If-None-Match": "*",
+    "aiManagerName": "aimanager1",
+    "modelSourceName": "huggingface",
+    "resource": {
+      "properties": {
+        "sourceType": "HuggingFace",
```
ravimeda, line 14: **[NEW] 🟠 Warning** **[[EX-COVERAGE §22.11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2211-example-coverage-ex-coverage) / [EX-TITLE §22.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#221-example-title-accuracy-ex-title)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/preview/2026-08-02-preview/examples/ModelSources_CreateOrUpdate.json` — line 12-22

**Classification reasoning:** [NEW] — all 31 example files in this API version directory are added by this PR.

**Issue:** Not one of the 31 example files in the new API version exercises any feature this PR adds. A grep across the full example directory at head SHA `f31b165e` returns zero matches for `MicrosoftFoundry`, `microsoftFoundry`, `managedIdentity`, and `totalWeightSize`, and `BaseModelReference.config` never appears in any payload. The entire feature set of this PR is unexampled. `ModelSources_CreateOrUpdate.json` is additionally titled `ModelSources_CreateOrUpdate_MaximumSet` (line 2) while demonstrating only `sourceType: "HuggingFace"` with an `inline` credential — the pre-existing `2026-05-02-preview` scenario — which EX-TITLE §22.1 requires titles not to misrepresent. Separately, `CustomAIModels_CreateOrUpdate.json` line 10-20 sets `modelSourceResourceId` to `.../modelSources/foundry-private-source` while `modelId` is `meta-llama/Llama-2-7b-chat`; if that source is intended to be Foundry-backed, the documented `modelName/version` form (`customaimodel.tsp` line 42-48) would apply instead.

**Fix:** Add a `ModelSources_CreateOrUpdate` example with `sourceType: "MicrosoftFoundry"`, a `microsoftFoundry.projectResourceId`, and a `credential.managedIdentity`; add a maximum-set `CustomAIModels_CreateOrUpdate` example populating `baseModel.totalWeightSize` and `baseModel.config`; make the `_MaximumSet` title match a payload that actually is the maximum set; and confirm the `modelId` value in the Foundry-sourced example.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
```diff
@@ -0,0 +1,141 @@
...
+
+  @visibility(Lifecycle.Read)
+  @doc("Read-only. Platform-resolved specification of the model.")
+  spec?: CustomAIModelSpec;
+}
+
+#suppress "@azure-tools/typespec-azure-core/casing-style" "AIModel is a valid name"
+@added(Versions.v2026_08_02_preview)
+@doc("Platform-resolved specification of a custom model. Extends `ModelSpec` with custom model-specific metadata. All fields are read-only.")
+model CustomAIModelSpec is ModelSpec;
```
ravimeda, line 67: **[NEW] 🟠 Warning** **[[Doc comment accuracy — TypeSpec §5 Common Anti-Patterns](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp` — line 68-71; emitted at `preview/2026-08-02-preview/aimanagers.json` line 2810-2812, JSON path `$.definitions.CustomAIModelSpec`

**Classification reasoning:** [NEW] — `CustomAIModelSpec` is `@added(Versions.v2026_08_02_preview)` and does not exist in the previous version.

**Issue:** TypeSpec §5 requires doc comments to accurately describe the element they annotate. `CustomAIModelSpec` is declared `model CustomAIModelSpec is ModelSpec;` (line 71) with the `@doc` "Platform-resolved specification of a custom model. Extends `ModelSpec` with custom model-specific metadata. All fields are read-only." The emitted definition (swagger line 2810-2835) is identical to `ModelSpec` (swagger line 3376-3401) — same three properties `license`, `isRestricted`, `maxContextLength`, same `readOnly` flags, same `required` array. There is no custom-model-specific metadata, so the doc describes behaviour the model does not have. The alias also emits a redundant model into every language SDK.

**Fix:** Either add the intended custom-model fields to `CustomAIModelSpec`, or point `CustomAIModelProperties.spec` (`customaimodel.tsp` line 65) at `ModelSpec` directly and delete the alias. If the alias is deliberately reserved as a future extension point, say that in the `@doc` — e.g. "Platform-resolved specification of a custom model. Currently identical to `ModelSpec`; reserved so custom-model-specific fields can be added without changing the SDK surface."

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
```diff
@@ -76,6 +108,39 @@ model InlineCredential {
   value: string;
 }
 
+@added(Versions.v2026_08_02_preview)
+@doc("A credential backed by a customer-owned user-assigned managed identity. The platform authenticates to the model source as this identity via Workload Identity; no secret is stored. Because the platform generally has no read access to a customer's identity, the `clientId` is supplied by the caller rather than resolved by the service.")
+model ManagedIdentityCredential {
```
ravimeda, line 102: **[NEW] 🔵 Suggestion** **[[Schema Consistency — trivially computable fields](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#schema-consistency)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp` — line 113-122; emitted at `preview/2026-08-02-preview/aimanagers.json` line 2956-2978, JSON path `$.definitions.ManagedIdentityCredential`

**Classification reasoning:** [NEW] — `ManagedIdentityCredential` is `@added(Versions.v2026_08_02_preview)`.

**Issue:** `ManagedIdentityCredential` requires all three of `resourceId`, `clientId` and `tenantId` from the caller, and both `clientId` and `tenantId` are documented as "Must match `resourceId`". §6 Schema Consistency says "**DO NOT** include fields whose values are trivially computable from other fields." The model's own `@doc` (line 112) explains the deviation — "Because the platform generally has no read access to a customer's identity, the `clientId` is supplied by the caller rather than resolved by the service" — which is a legitimate reason not to resolve them server-side, so this is a design observation rather than a rule violation. Two consequences are still worth weighing: callers must supply two values the service will only ever validate, an easy source of `400`s; and the documented "must match" validation names no error code, unlike the sibling `microsoftFoundry` property (line 67), which specifies `InvalidModelSourceConfiguration`.

**Fix (options, not a directive):**

1. **Preferred** — model the identity on the ARM envelope via the common-types `ManagedServiceIdentity` (`Azure.ResourceManager.ManagedServiceIdentityProperty`, `UserAssigned`), which ARM populates and validates, removing caller-supplied `clientId` / `tenantId` entirely. `ModelSource` has no PATCH operation, so [RPC-Patch-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#43b-patch-should-support-updating-envelope-properties-rpc-patch-v1-09-rpc-patch-v1-11) does not apply.
2. Keep the current shape but document the error code for the "must match `resourceId`" validation, matching the `InvalidModelSourceConfiguration` precedent.
3. Keep as-is and accept the redundancy — reasonable if option 1 is blocked by the RP's lack of read access to customer identities, but please confirm that explicitly.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iterations, converged) against commit [`f31b165`](https://github.com/Azure/azure-rest-api-specs/pull/45015/commits/f31b165e62f83f32be5e44b045ccb1463fc7632f). See inline comments for Blocking 1, Warnings 1, 3 and 4, and Suggestion 2. Warning 2 and Suggestion 1 are posted as replies on the two existing open threads they relate to.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modeldeployment.tsp
ravimeda: **[NEW] 🟠 Warning** **[[ARM §26.0 — Published API Versions Are Immutable](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#260-published-api-versions-are-immutable) / [`@added` version misalignment — TypeSpec §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/preview/2026-05-02-preview/aimanagers.json` — line 2270-2272 and line 2483

**Classification reasoning:** [NEW] — both edits are introduced by this PR; a base-vs-head diff of the published document at base SHA `edf0326b` versus head SHA `f31b165e` shows exactly these two changed lines.

**Issue:** ARM §26.0 states "Once an API version is published (merged to `main`…), its schema is **immutable**. No changes — not even adding an optional property — are allowed to that version." TypeSpec §5 names this exact failure mode: "**`@added` version misalignment** — when using `@added(Versions.vXXXX)` to gate a new feature, verify that the feature does not inadvertently alter descriptions or schemas of earlier API versions." Two retroactive edits land in the published `2026-05-02-preview` document, both caused by ungated `@doc` text on symbols that are `@added(v2026_05_02_preview)`:

- **line 2483** — `$.definitions.ModelDeploymentProperties.properties.modelResourceId.description` was rewritten from "Full ARM resource id of the model to deploy. Phase 1 accepts an `AIModel` resource id only. Immutable after creation." to text claiming the property "Accepts either a catalog `AIModel` id or a customer-registered `CustomAIModel` id (including MicrosoftFoundry-backed models)." In that version the property `$ref`s `AIModelResourceId` (same file, line 2071-2081), whose `allowedResources` contains only `Microsoft.ContainerService/aiModels`. The strings `CustomAIModel` and `MicrosoftFoundry` occur exactly once in the whole published document — in this retroactively inserted description. The published version's documentation now contradicts its own schema.
- **line 2272** — `$.definitions.CredentialValue.description` lost the sentence "Exactly one variant must be set." That sentence documented a service-enforced constraint of the published version; removing it weakens the shipped contract for a reason that applies only to the new version.

This answers the question you raised in https://github.com/Azure/azure-rest-api-specs/pull/45015#discussion_r3665361870 after correctly fixing the related `allowedResources` regression.

**Fix:** TypeSpec has no doc-versioning decorator, so a shared `@doc` must stay accurate for every version it is emitted into, and version-specific prose must live on a symbol that is itself version-gated.
- For (1): restore a version-neutral `@doc` on `ModelDeploymentProperties.modelResourceId` (`modeldeployment.tsp` line 80-87), e.g. "Full ARM resource id of the model to deploy. Immutable after creation.", and move the `CustomAIModel` / `MicrosoftFoundry` prose onto the `DeployableModelResourceId` scalar (`modeldeployment.tsp` line 28-39), which is already `@added(Versions.v2026_08_02_preview)` and therefore emitted only into `2026-08-02-preview`.
- For (2): restore "Exactly one variant must be set." to the `CredentialValue` `@doc` (`modelsource.tsp` line 86). Warning 1's suggested rewrite keeps that sentence, so both findings are satisfied by one edit.
- Then re-run `tsp compile .` and confirm `git diff` against `main` shows zero changed lines in `preview/2026-05-02-preview/aimanagers.json`.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md
ravimeda: **[NEW] 🔵 Suggestion** **[[ARM §8.5 — `additionalProperties` Restriction](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#85-additionalproperties-restriction) / [TypeSpec §4.1 — Warning Suppression Policy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy) / [RPC-SUPPRESS-GA](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#ga-vs-preview-rigor-rpc-suppress-ga)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/readme.md` — line 52-55

**Classification reasoning:** [NEW] — this suppression is added by this PR. The three other `AvoidAdditionalProperties` entries in the same tag block (line 40-51) are carried over from `2026-05-02-preview` unchanged and are not flagged.

**Issue:** The new suppression for `$.definitions.BaseModelReference.properties.config` carries the one-line reason "The verbatim `config.json` of the base model." ARM §8.5 states: "`additionalProperties` is only acceptable for **user-defined pass-through data** (e.g., tags, custom metadata) that the service does not validate or interpret." The property's own description (`customaimodel.tsp` line 86, emitted at swagger line 2510) says the platform stores it and projects it into the dataplane "to determine GPU node count and vLLM configuration" — i.e. the service does interpret it, so the pass-through exemption is not clearly met on the §8.5 test as written. The justification in https://github.com/Azure/azure-rest-api-specs/pull/45015#discussion_r3692723721 (user-supplied, non-uniform across models, required for private HF/Foundry repositories) is reasonable but does not by itself resolve that. TypeSpec §4.1 also requires suppression reasons to stand on their own with a technical justification; the `reason:` records none of this rationale.

**Fix:** Three options, in preference order:

1. **Narrow the interpreted subset.** If the platform reads only specific keys to size GPUs (e.g. `hidden_size`, `num_hidden_layers`, `torch_dtype`), model those as typed properties on `BaseModelReference` and drop the open bag — §8.5: "If the set of keys is known or bounded, model them as explicit named properties instead."
2. **Keep the open bag and escalate.** Per the [suppression Decision Framework](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#decision-framework), a new resource type with a reasonable author justification routes to "Escalate to peer review" rather than single-reviewer approval. Raise it with the ARM API board and record the outcome.
3. **At minimum, make the `reason:` self-standing**, e.g. `reason: Verbatim user-supplied config.json from the upstream model repository. Key set is model-specific and unbounded; the platform forwards it to the dataplane and reads only sizing hints from it.` Per RPC-SUPPRESS-GA this must be re-challenged before any stable version. Consider documenting a size bound in the property description.

bfoley13: Updated the `reason` to be self standing as suggested by comments

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
bfoley13: I have added read to the credential as there are portions that we expect to be readable and only the secret is not.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
bfoley13: There is no mutual exclusion constraint and the old docs are too constrictive.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modelsource.tsp
bfoley13: This is now just the resource id of the managed identity as the RP will have permissions to read the clientId/tenantId

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/customaimodel.tsp
bfoley13: added reserved phrasing

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/modeldeployment.tsp
bfoley13: I used more version neutral docs for both of these

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/preview/2026-08-02-preview/examples/ModelSources_CreateOrUpdate.json
bfoley13: added examples for these items.

specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/aimanagernamespace.tsp
```diff
@@ -87,7 +87,6 @@ interface AIManagerNamespaces {
     ArmResponse<CredentialResults>
   >;
 
-  @added(Versions.v2026_05_02_preview)
```
bfoley13, line 90: We are moving to a single preview version within the `main.tsp` which means we will remove the @added tags until we actually need them. This will stop back porting to the old versions when we have updates to docs/etc.

bfoley13:
There was a seperate PR for BYO cluster support in AIManager that was recently merged but for the same API version which is why we have breaking change happening here

ravimeda:
@bfoley13 please see  https://github.com/Azure/azure-rest-api-specs/pull/45015#issuecomment-5096277795
>The required check named Swagger BreakingChange has failed. To unblock this PR, follow the process at [aka.ms/brch](https://aka.ms/brch).

matthchr (approved):

sabbour (approved):
Approving based on https://github.com/azure-management-and-platforms/aks-handbook/pull/229 approval

PR #45021, opened 2026-07-27
Add PATCH agentpool, list-operations-by-agentpool, and operationType/subOperationType in OperationStatusResult

HuShaoRu:
API proposal: https://github.com/azure-management-and-platforms/aks-handbook/blob/main/api/2026-06-02-preview/concurrent-node-lifecycle-and-pool-operations.md

### New operations

| Method | Path | operationId |
|---|---|---|
| `PATCH` | `.../managedClusters/{cluster}/agentPools/{agentPool}` | `AgentPools_Update` |
| `GET`   | `.../managedClusters/{cluster}/agentPools/{agentPool}/operations?activeOnly=…` | `OperationStatusResult_ListByAgentPool` |

- `AgentPools_Update` is a long-running PATCH (`Azure-AsyncOperation` + `Location` + `Retry-After`, `final-state-via: azure-async-operation`), body = `AgentPoolUpdate`, returns full `AgentPool` on success. Supports `If-Match`.
- `OperationStatusResult_ListByAgentPool` mirrors the existing `OperationStatusResult_List` (cluster-scope), plus the `agentPoolName` path param and the `activeOnly` query param. Pageable.

### New models

- `AgentPoolUpdate` → `AgentPoolUpdateProperties { count?, virtualMachinesProfile? }` → `AgentPoolUpdateVirtualMachinesProfile { scale? }` → `AgentPoolUpdateScaleProfile { manual? }` → `AgentPoolUpdateManualScaleProfile { size?, count? }`. Strict subset of `AgentPool` — everything is optional; only scale-related fields are writable through this PATCH.

### Additive changes to existing shapes

- `OperationStatusResult` — added two optional, read-only properties:
  - `operationType?: string`
  - `subOperationType?: string`

  These flow through `OperationStatusResult_Get`, `OperationStatusResult_List`, `OperationStatusResult_GetByAgentPool`, and `OperationStatusResult_ListByAgentPool`.

## Files touched

- TypeSpec sources:
  - `AgentPool.tsp` — new `update` and `listByAgentPool` operations; switched `getByAgentPool` response to the local `OperationStatusResult`.
  - `AgentPoolModels.tsp` — new `AgentPoolUpdate*` models.
  - `CommonModels.tsp` — new local `OperationStatusResult` (spread of the ARM common type) with the two new `@added` properties.
  - `ManagedCluster.tsp` — switched `operationStatusResultGet` response to the local `OperationStatusResult` (list already flowed through the local type via `OperationStatusResultList`).
  - `BackCompatible.tsp` — `@@clientLocation(AgentPools.listByAgentPool, "OperationStatusResult")` + `@@clientName(AgentPools.update::parameters.properties, "parameters")`, mirroring the existing `getByAgentPool` pattern.
  - `client.tsp` — `@@scope(AgentPools.listByAgentPool, "!csharp")`, matching the other `operationStatusResult*` operations.
- Examples (TypeSpec-owned, under `aks/examples/2026-06-02-preview/`):
  - New: `AgentPoolsUpdate_Scale.json`, `AgentPoolsUpdate_ScaleVMs.json`, `OperationStatusResultListByAgentPool.json`, `OperationStatusResultListByAgentPool_Active.json`.
  - Updated to include the new fields: `OperationStatusResultGet.json`, `OperationStatusResultList.json`, `OperationStatusResultGetByAgentPool.json`.
- Regenerated (do not hand-edit): `aks/preview/2026-06-02-preview/managedClusters.json` and everything under `aks/preview/2026-06-02-preview/examples/`. Produced by `npx tsp compile .` from the `aks/` directory.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -4365,7 +4365,53 @@ model SubResource {
  * The current status of an async operation.
  */
 model OperationStatusResult {
-  ...Azure.ResourceManager.CommonTypes.OperationStatusResult;
+  /**
+   * Fully qualified ID for the async operation.
+   */
+  id?: Azure.Core.armResourceIdentifier;
+
+  /**
+   * Name of the async operation.
+   */
+  name?: string;
+
+  /**
+   * Operation status.
+   */
+  status: string;
+
+  /**
+   * Percent of the operation that is complete.
+   */
+  @minValue(0)
+  @maxValue(100)
+  percentComplete?: float64;
+
+  /**
+   * The start time of the operation.
+   */
+  startTime?: utcDateTime;
+
+  /**
+   * The end time of the operation.
+   */
+  endTime?: utcDateTime;
+
+  /**
+   * The operations list.
+   */
+  operations?: OperationStatusResult[];
+
+  /**
+   * If present, details of the operation error.
+   */
+  error?: Azure.ResourceManager.CommonTypes.ErrorDetail;
+
+  /**
+   * Fully qualified ID of the resource against which the original async operation was started.
+   */
+  @visibility(Lifecycle.Read)
+  resourceId?: Azure.Core.armResourceIdentifier;
 
   /**
    * The type of the operation.
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -4362,6 +4362,27 @@ model SubResource {
 }
 
 /**
+ * The current status of an async operation.
+ */
+model OperationStatusResult {
+  ...Azure.ResourceManager.CommonTypes.OperationStatusResult;
+
+  /**
+   * The type of the operation.
+   */
+  @added(Versions.v2026_06_02_preview)
+  @visibility(Lifecycle.Read)
+  operationType?: string;
+
+  /**
+   * The type of the suboperation.
+   */
+  @added(Versions.v2026_06_02_preview)
+  @visibility(Lifecycle.Read)
+  subOperationType?: string;
+}
+
+/**
  * The operations list. It contains an URL link to get the next set of results.
  */
 model OperationStatusResultList is Azure.Core.Page<OperationStatusResult>;
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -4364,9 +4364,9 @@ model SubResource {
 /**
  * The current status of an async operation.
  */
-#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "OperationStatusResult follows the ARM common type shape by inheritance."
-model OperationStatusResult
-  extends Azure.ResourceManager.CommonTypes.OperationStatusResult {
+model OperationStatusResult {
+  ...Azure.ResourceManager.CommonTypes.OperationStatusResult;
+
   /**
    * The type of the operation.
    */
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
@@ -4364,9 +4364,9 @@ model SubResource {
 /**
  * The current status of an async operation.
  */
-model OperationStatusResult {
-  ...Azure.ResourceManager.CommonTypes.OperationStatusResult;
-
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "OperationStatusResult follows the ARM common type shape by inheritance."
+model OperationStatusResult
+  extends Azure.ResourceManager.CommonTypes.OperationStatusResult {
   /**
    * The type of the operation.
    */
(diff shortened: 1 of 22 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
```diff
@@ -4356,6 +4361,73 @@ model SubResource {
   type?: string;
 }
 
+/**
+ * The current status of an async operation.
+ */
+model OperationStatusResult {
```
FumingZhang, line 4367: In the proposal, this was modeled as a composition of Azure.ResourceManager.CommonTypes.OperationStatusResult. Would there be any reason not to do the same here?

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/ManagedCluster.tsp
```diff
@@ -642,7 +642,7 @@ interface ManagedClusters {
   operationStatusResultGet is ManagedClusterOps.ActionSync<
     ManagedCluster,
     void,
-    ArmResponse<Azure.ResourceManager.CommonTypes.OperationStatusResult>,
+    ArmResponse<OperationStatusResult>,
```
FumingZhang, line 645: I don’t see this included in the proposal — was this change intended?

HuShaoRu: yes. we want the new fields to be added to existing APIs, both for mc and ap.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/CommonModels.tsp
HuShaoRu: We tried three "compose from `Azure.ResourceManager.CommonTypes.OperationStatusResult`" variants before landing on the current design, and each hit a hard CI failure:

**Attempt 1 — `is Azure.ResourceManager.CommonTypes.OperationStatusResult`**
The typespec-autorest emitter collapses `is` into a bare `$ref` back to the common definition. The `operationType` / `subOperationType` properties we added were **silently dropped** from the emitted swagger. Non-starter — the new fields never made it into the API.

**Attempt 2 — `extends Azure.ResourceManager.CommonTypes.OperationStatusResult`**
Emitted `allOf: [{ $ref: common OperationStatusResult }] + own properties`. Two blocking failures:
- **`Breaking Change (Cross-Version)` runtime exception**: `Found circular allOf reference` — openapi-diff walks the `allOf` back into the common OSR (which contains `operations: OperationStatusResult[]` self-refing common) and crashes.
- **`SDK Validation - Python` error**: `@azure-tools/typespec-client-generator-core/duplicate-client-name: Client name: "OperationStatusResult" is duplicated in language scope: "python"`. Because `extends` puts the common OSR into our type hierarchy, TCGC registers both the local and the common `OperationStatusResult` as client types in the Python scope, and they collide.
- Also raised `composition-over-inheritance` from the tsp azure-core linter (would need suppression).

**Attempt 3 — `...Azure.ResourceManager.CommonTypes.OperationStatusResult` (spread)**
This fixed the openapi-diff crash (no more `allOf`) and removed the `composition-over-inheritance` lint, but **the Python `duplicate-client-name` error persisted**. TCGC follows the referenced type in a spread the same way it does for `extends`, so both `OperationStatusResult`s were still registered in the Python client scope with the same name.

**Current — define locally with matching fields**
Same pattern used today by `networkcloud/NetworkCloud.Management/common/models.tsp` and `kubernetesconfiguration/resource-manager/Microsoft.KubernetesConfiguration/operations/models.tsp` for the identical situation. We declare each ARM common v6 OSR field manually (using the same scalars — `Azure.Core.armResourceIdentifier` for `id`/`resourceId`, `Azure.ResourceManager.CommonTypes.ErrorDetail` for `error`, etc.) and let `operations?: OperationStatusResult[]` self-reference our **local** model instead of the common one. TCGC therefore only sees a single `OperationStatusResult` in the Python client scope, and openapi-diff sees no `allOf` cycle. The emitted swagger schema is a superset of the common OSR (all 9 common fields + `operationType`, `subOperationType`).

Trade-off we accept: a bit of manual field duplication in the tsp source, mirroring what the two neighboring services do. The upside is that both `SDK Validation - Python` and the `Breaking Change (Cross-Version)` circular-allOf error are eliminated without needing suppressions for either.

specification/containerservice/resource-manager/Microsoft.ContainerService/aks/readme.md
```diff
@@ -1558,4 +1558,12 @@ directive:
...
+  - suppress: BodyTopLevelProperties
+    from: managedClusters.json
+    where: $.definitions.OperationStatusResult
+    reason: OperationStatusResult mirrors the ARM common type shape (`error`, `operations`, `resourceId`, plus `status`, `percentComplete`, `startTime`, `endTime`, etc.) and adds the AKS-specific read-only `operationType` and `subOperationType` properties. It is only used as an async-operation-status response body, not as a PUT resource body, so the tracked-resource envelope restriction does not apply.
+  - suppress: RequiredPropertiesMissingInResourceModel
+    from: managedClusters.json
+    where: $.definitions.OperationStatusResult
+    reason: OperationStatusResult is an async-operation-status response envelope, not an ARM tracked or proxy resource. It inherits `id` and `name` from the ARM common OperationStatusResult; the tracked-resource-shape check (name/id/type readOnly) does not apply.
```
FumingZhang:
```suggestion
    reason: OperationStatusResult is an async-operation-status response envelope, not an ARM tracked or proxy resource. It locally declares the ARM common OperationStatusResult fields and adds AKS-specific read-only fields. Therefore, the tracked-resource requirement for read-only name/id/type properties does not apply.
```

FumingZhang: @HuShaoRu Thanks for the clarification. The other changes look good to me. I left a minor comment to clarify the suppression rationale, which should help avoid reviewer confusion.

FumingZhang (commented):
lgtm

FumingZhang (approved):

PR #45040, opened 2026-07-28
switch routines listing to use Azure Core Page contracts

runhli:
# Choose a PR Template

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
@@ -435,14 +435,9 @@ model ListRoutinesParameters {
   @query
   limit?: int32;
 
-  @doc("An opaque cursor returned as last_id by the previous list response.")
+  @doc("An opaque cursor identifying the item after which to start returning results. Prefer following the `nextLink` returned by the previous list response.")
   @query
-  @continuationToken
   after?: string;
-
-  @doc("Unsupported. Reserved for future backward pagination support.")
-  @query
-  before?: string;
 
   @doc("The ordering direction. Supported values are asc and desc.")
   @query
@@ -472,14 +467,9 @@ model ListRoutineRunsParameters {
   @query
   limit?: int32;
 
-  @doc("An opaque cursor returned as last_id by the previous list-runs response.")
+  @doc("An opaque cursor identifying the run after which to start returning results. Prefer following the `nextLink` returned by the previous list-runs response.")
   @query
-  @continuationToken
   after?: string;
-
-  @doc("Unsupported. Reserved for future backward pagination support.")
-  @query
-  before?: string;
 
   @doc("The ordering direction. Supported values are asc and desc.")
   @query
--- a/specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
@@ -435,7 +435,7 @@ model ListRoutinesParameters {
   @query
   limit?: int32;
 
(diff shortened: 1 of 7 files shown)


specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
```diff
@@ -435,15 +435,10 @@ model ListRoutinesParameters {
...
+  @doc("An opaque continuation token identifying where to resume the list. Prefer following the `nextLink` returned by the previous response, which embeds this value.")
   @query
-  @continuationToken
   after?: string;
 
-  @doc("Unsupported. Reserved for future backward pagination support.")
-  @query
-  before?: string;
-
   @doc("The ordering direction. Supported values are asc and desc.")
```
dargilco: Does the above doc string ("The ordering direction. Supported values are asc and desc.") need update? If I remember correctly, you only support `asc`.

dargilco (dismissed):

specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
```diff
@@ -472,15 +467,10 @@ model ListRoutineRunsParameters {
...
```
dargilco: same here

specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
runhli: I think order parameter doesn't change, our implementation supports it.
Cosmos DB (routines list) — SQL supports ORDER BY <path> ASC | DESC natively. The bool flows straight through routine service.
MLflow (routine runs list) — the SearchRuns REST API accepts an order_by array of "<column> ASC|DESC" strings. Same story: page tokens (next_page_token) resume the same directional traversal.

specification/ai-foundry/data-plane/Foundry/src/routines/models.tsp
runhli: replied

specification/ai-foundry/data-plane/Foundry/src/routines/routes.tsp
```diff
@@ -97,9 +97,9 @@ interface Routines {
   @route("/routines/{routine_name}/runs")
   @extension("x-ms-foundry-meta", RoutineRequiredPreviews)
   listRoutineRuns is FoundryDataPlanePreviewOperation<
-    FoundryFeaturesOptInKeys.routines_v1_preview,
+    FoundryFeaturesOptInKeys.routines_v2_preview,
     ListRoutineRunsParameters,
-    AgentsPagedResult<RoutineRun>
+    Azure.Core.Page<RoutineRun>
```
glecaros: using Azure.CorePage introduces nextLink (which does not match the naming of other properties), rather than that it's probably a better idea to introduce a type that has a shape that matches (`next_link` and arguable `data` for the items to keep it consistent with the other list apis)

runhli: Why do we need to introduce a new type give Azure Arm contracts aleady have similar contract? Based on the previous discussions, I think the goal is to not match the naming of other list responses. We want to follow existing Azure REST guidelines: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections

runhli: discussed offline, updated to use PagedResultWithNextLink with data and next_link property so that it's consistent with existing properties around.

glecaros (approved):

PR #45051, opened 2026-07-28
Add SecretsProvisioningSettings to SecurityProfile and ImageMetadataP…

awadhavekar:
This PR adds TypeSpec definitions and generated Swagger for the Secrets Provisioning feature on Gallery Image Versions.

Changes

New models:
•  SecretsProvisioningSettings  — captures whether secrets provisioning is supported, the OS name, and component versions (OS, CloudInit, AzureGuestAgent, SecretsProvisioningLibrary)
•  SecretsProvisioningComponent  — name/version pair for each component
•  ImageMetadataProfile  — metadata profile with type, public and internal key-value lists
•  MetadataKeyValue  — key-value pair for metadata entries
•  MetadataType  — extensible union:  SecretsProvisioningImageMetadata ,  UserProvidedSecretsProvisioningMetadata ,  Unknown 

Property additions:

•  secretsProvisioningSettings  added to  ImageVersionSecurityProfile  (write-only: create/update)
•  imageMetadataProfiles  added to  GalleryImageVersionProperties  (read-only)

All additions are versioned with  @added(Versions.v2026_03_03) .
# Choose a PR Template

### DIFF
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -967,7 +967,7 @@ model SecretsProvisioningSettings {
   /**
    * Specifies whether the image version supports secrets provisioning.
    */
-  isSupported: boolean;
+  isSupported?: boolean;
 
   /**
    * The name of the operating system (e.g., "mariner").
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -1,13 +1,11 @@
 import "@typespec/rest";
 import "@typespec/http";
-import "@typespec/openapi";
 import "@typespec/versioning";
 import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-azure-resource-manager";
 
 using TypeSpec.Rest;
 using TypeSpec.Http;
-using OpenAPI;
 using Versioning;
 using Azure.ResourceManager;
 using Common;
@@ -884,11 +882,6 @@ union MetadataType {
    * Metadata related to user-provided secrets provisioning for the image.
    */
   UserProvidedSecretsProvisioningMetadata: "UserProvidedSecretsProvisioningMetadata",
-
-  /**
-   * Unknown metadata type.
-   */
-  Unknown: "Unknown",
 }
 
 /**
@@ -897,15 +890,12 @@ union MetadataType {
 @added(Versions.v2026_03_03)
 model MetadataKeyValue {
   /**
-   * The metadata key.
+   * The metadata key. Known keys emitted by the service include 'Linux.AzureSecretsProvisioning.Enabled', 'OS.Name', and '{componentName}.Version' (e.g., 'AzureGuestAgent.Version'). All values are non-sensitive configuration; no secrets, credentials, or cryptographic material transit this field.
    */
-  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "metadataKey is not sensitive, it is a key name for metadata"
-  #suppress "@azure-tools/typespec-azure-core/no-openapi" "x-ms-secret annotation required by Swagger linter to mark non-secret properties containing 'key' in name"
-  @extension("x-ms-secret", false)
-  metadataKey?: string;
+  metadataKey: string;
 
   /**
-   * The metadata value.
+   * The metadata value. Contains non-sensitive configuration such as capability flags ('true'/'false'), OS names ('Linux', 'Windows'), and version strings (e.g., '1.0.0').
    */
   metadataValue?: string;
 }
@@ -918,19 +908,39 @@ model ImageMetadataProfile {
   /**
    * The type of metadata.
    */
-  type?: MetadataType;
+  type: MetadataType;
 
   /**
-   * The list of public metadata key-value pairs.
+   * The list of public metadata key-value pairs. Contains non-sensitive image capability metadata such as supported OS, component names, and versions. No secret material is emitted in this list.
    */
-  @identifiers(#[])
+  @identifiers(#["metadataKey"])
   publicMetadataList?: MetadataKeyValue[];
 
   /**
-   * The list of internal metadata key-value pairs.
+   * The list of internal metadata key-value pairs. Contains non-sensitive service-internal metadata for diagnostics and tracking. No secret material is emitted in this list.
    */
-  @identifiers(#[])
+  @identifiers(#["metadataKey"])
   internalMetadataList?: MetadataKeyValue[];
+}
+
+/**
+ * The name of a component involved in secrets provisioning.
+ */
+@added(Versions.v2026_03_03)
+union SecretsProvisioningComponentName {
+  string,
+
+  /** The guest operating system. */
+  OS: "OS",
+
+  /** The cloud-init provisioning agent. */
+  CloudInit: "CloudInit",
+
+  /** The Azure guest agent. */
+  AzureGuestAgent: "AzureGuestAgent",
+
+  /** The secrets provisioning library. */
+  SecretsProvisioningLibrary: "SecretsProvisioningLibrary",
 }
 
 /**
@@ -939,14 +949,14 @@ model ImageMetadataProfile {
 @added(Versions.v2026_03_03)
 model SecretsProvisioningComponent {
   /**
-   * The name of the component (e.g., "OS", "CloudInit", "AzureGuestAgent", "SecretsProvisioningLibrary").
+   * The name of the component.
    */
-  name: string;
+  name?: SecretsProvisioningComponentName;
 
   /**
    * The version of the component.
    */
-  version: string;
+  version?: string;
 }
 
 /**
@@ -955,7 +965,7 @@ model SecretsProvisioningComponent {
 @added(Versions.v2026_03_03)
 model SecretsProvisioningSettings {
   /**
-   * Specifies whether secrets provisioning is enabled.
+   * Specifies whether the image version supports secrets provisioning.
    */
   isSupported: boolean;
 
@@ -967,7 +977,7 @@ model SecretsProvisioningSettings {
   /**
    * The list of component versions involved in secrets provisioning.
    */
-  @identifiers(#[])
+  @identifiers(#["name"])
   components?: SecretsProvisioningComponent[];
 }
 
@@ -1023,7 +1033,7 @@ model GalleryImageVersionProperties {
    */
   @added(Versions.v2026_03_03)
   @visibility(Lifecycle.Read)
-  @identifiers(#[])
+  @identifiers(#["type"])
   imageMetadataProfiles?: ImageMetadataProfile[];
 }
 
@@ -1479,7 +1489,6 @@ model ImageVersionSecurityProfile {
    * Specifies the secrets provisioning settings for the gallery image version. Used on create or update to configure secrets provisioning.
    */
   @added(Versions.v2026_03_03)
-  @visibility(Lifecycle.Create, Lifecycle.Update)
   secretsProvisioningSettings?: SecretsProvisioningSettings;
 }
 
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -867,6 +867,101 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
 }
 
 /**
+ * The type of metadata associated with the image.
+ */
+@added(Versions.v2026_03_03)
+union MetadataType {
+  string,
+
+  /**
+   * Metadata related to secrets provisioning for the image.
+   */
+  SecretsProvisioningImageMetadata: "SecretsProvisioningImageMetadata",
+
+  /**
+   * Unknown metadata type.
+   */
+  Unknown: "Unknown",
+}
+
+/**
+ * Describes a key-value pair for image metadata.
+ */
+@added(Versions.v2026_03_03)
+model MetadataKeyValue {
+  /**
+   * The metadata key.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "metadataKey is not sensitive, it is a key name for metadata"
+  metadataKey?: string;
+
+  /**
+   * The metadata value.
+   */
+  metadataValue?: string;
+}
+
+/**
+ * Describes the metadata profile of an image.
+ */
+@added(Versions.v2026_03_03)
+model ImageMetadataProfile {
+  /**
+   * The type of metadata.
+   */
+  type?: MetadataType;
+
+  /**
+   * The list of public metadata key-value pairs.
+   */
+  @identifiers(#[])
+  publicMetadataList?: MetadataKeyValue[];
+
+  /**
+   * The list of internal metadata key-value pairs.
+   */
+  @identifiers(#[])
+  internalMetadataList?: MetadataKeyValue[];
+}
+
+/**
+ * Describes a component involved in secrets provisioning.
+ */
+@added(Versions.v2026_03_03)
+model SecretsProvisioningComponent {
+  /**
+   * The name of the component (e.g., "OS", "CloudInit", "AzureGuestAgent", "SecretsProvisioningLibrary").
+   */
+  name: string;
+
+  /**
+   * The version of the component.
+   */
+  version: string;
+}
+
+/**
+ * Describes the secrets provisioning settings for a gallery image version.
+ */
+@added(Versions.v2026_03_03)
+model SecretsProvisioningSettings {
+  /**
+   * Specifies whether secrets provisioning is enabled.
+   */
+  isSupported: boolean;
+
+  /**
+   * The name of the operating system (e.g., "mariner").
+   */
+  osName?: string;
+
+  /**
+   * The list of component versions involved in secrets provisioning.
+   */
+  components?: SecretsProvisioningComponent[];
+}
+
+/**
  * Describes the properties of a gallery image version.
  */
 model GalleryImageVersionProperties {
@@ -912,6 +1007,14 @@ model GalleryImageVersionProperties {
    */
   @visibility(Lifecycle.Read)
   validationsProfile?: ValidationsProfile;
+
+  /**
+   * The image metadata profiles associated with the gallery image version.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  @identifiers(#[])
+  imageMetadataProfiles?: ImageMetadataProfile[];
 }
 
 /**
@@ -1361,6 +1464,13 @@ model ImageVersionSecurityProfile {
    * Contains UEFI settings for the image version.
    */
   uefiSettings?: GalleryImageVersionUefiSettings;
+
+  /**
+   * Specifies the secrets provisioning settings for the gallery image version. Used on create or update to configure secrets provisioning.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  secretsProvisioningSettings?: SecretsProvisioningSettings;
 }
 
 /**
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -1,11 +1,13 @@
 import "@typespec/rest";
 import "@typespec/http";
+import "@typespec/openapi";
 import "@typespec/versioning";
 import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-azure-resource-manager";
 
 using TypeSpec.Rest;
 using TypeSpec.Http;
+using OpenAPI;
 using Versioning;
 using Azure.ResourceManager;
 using Common;
@@ -898,6 +900,8 @@ model MetadataKeyValue {
    * The metadata key.
    */
   #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "metadataKey is not sensitive, it is a key name for metadata"
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "x-ms-secret annotation required by Swagger linter to mark non-secret properties containing 'key' in name"
+  @extension("x-ms-secret", false)
   metadataKey?: string;
 
   /**
@@ -963,6 +967,7 @@ model SecretsProvisioningSettings {
   /**
    * The list of component versions involved in secrets provisioning.
    */
+  @identifiers(#[])
   components?: SecretsProvisioningComponent[];
 }
 
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -1942,6 +1942,20 @@ model GallerySoftDeletedResourceProperties {
    * The timestamp for when the resource is soft-deleted. In dateTime offset format.
    */
   softDeletedTime?: string;
+
+  /**
+   * The timestamp after which a soft-deleted gallery image version is no longer consumable for VM/VMSS creation or VMSS scale out. It is calculated from the soft-deleted time plus the retention period. In dateTime offset format.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  consumptionEndTime?: utcDateTime;
+
+  /**
+   * The timestamp at which a soft-deleted gallery image version is permanently (hard) deleted and can no longer be recovered. In dateTime offset format.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  hardDeletionTargetTime?: utcDateTime;
 }
 
 /**
@@ -2534,6 +2548,24 @@ model SharedGalleryImageVersionList
 @@identifiers(SharedGalleryImageVersionList.value, #[]);
 
 /**
+ * The state of a gallery image version, derived from its soft-delete status.
+ */
+@added(Versions.v2026_03_03)
+union GalleryImageVersionState {
+  string,
+
+  /**
+   * The gallery image version is active and available for use.
+   */
+  Active: "Active",
+
+  /**
+   * The gallery image version has been soft-deleted. It is available for use only when a specific version is requested, and it will not be resolved as the latest version.
+   */
+  SoftDeleted: "SoftDeleted",
+}
+
+/**
  * Describes the properties of a gallery image version.
  */
 #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "For backward compatibility"
@@ -2565,6 +2597,20 @@ model SharedGalleryImageVersionProperties {
    */
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "For backward compatibility"
   artifactTags?: Record<string>;
+
+  /**
+   * The timestamp after which a soft-deleted gallery image version is no longer consumable for VM/VMSS creation or VMSS scale out. It is calculated from the soft-deleted time plus the retention period, and is not present for active gallery image versions. In dateTime offset format.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  consumptionEndTime?: utcDateTime;
+
+  /**
+   * The state of the gallery image version, derived from its soft-delete status.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  imageState?: GalleryImageVersionState;
 }
 
 /**
@@ -2845,6 +2891,20 @@ model CommunityGalleryImageVersionProperties {
    */
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "For backward compatibility"
   artifactTags?: Record<string>;
+
+  /**
+   * The timestamp after which a soft-deleted gallery image version is no longer consumable for VM/VMSS creation or VMSS scale out. It is calculated from the soft-deleted time plus the retention period, and is not present for active gallery image versions. In dateTime offset format.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  consumptionEndTime?: utcDateTime;
+
+  /**
+   * The state of the gallery image version, derived from its soft-delete status.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Read)
+  imageState?: GalleryImageVersionState;
 }
 
 /**
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -879,6 +879,11 @@ union MetadataType {
   SecretsProvisioningImageMetadata: "SecretsProvisioningImageMetadata",
 
   /**
+   * Metadata related to user-provided secrets provisioning for the image.
+   */
+  UserProvidedSecretsProvisioningMetadata: "UserProvidedSecretsProvisioningMetadata",
+
+  /**
    * Unknown metadata type.
    */
   Unknown: "Unknown",
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
@@ -892,6 +892,7 @@ model MetadataKeyValue {
   /**
    * The metadata key. Known keys emitted by the service include 'Linux.AzureSecretsProvisioning.Enabled', 'OS.Name', and '{componentName}.Version' (e.g., 'AzureGuestAgent.Version'). All values are non-sensitive configuration; no secrets, credentials, or cryptographic material transit this field.
    */
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "metadataKey is a dictionary-style key name (e.g. 'OS.Name'), not a cryptographic key or secret. The suffix match on 'key' is a false positive."
   metadataKey: string;
 
   /**
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
@@ -384,6 +384,10 @@ suppressions:
   - code: XMSSecretInResponse
     reason: Existing secret fields maintained for backward compatibility.
     from: ComputeRP.json
+  - code: XMSSecretInResponse
+    reason: metadataKey is a metadata field name identifier, not a cryptographic secret or API key.
+    from: GalleryRP.json
+    where: $.definitions.MetadataKeyValue.properties.metadataKey
   - code: LroLocationHeader
     reason: Existing LRO operation does not include Location header for backward compatibility.
     from: ComputeRP.json
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
@@ -385,7 +385,7 @@ suppressions:
     reason: Existing secret fields maintained for backward compatibility.
     from: ComputeRP.json
   - code: XMSSecretInResponse
-    reason: metadataKey is a metadata field name identifier, not a cryptographic secret or API key.
+    reason: metadataKey is a metadata field name, not a cryptographic key. Wire-format name is fixed by existing CAPS backend serialization.
     from: GalleryRP.json
     where: $.definitions.MetadataKeyValue.properties.metadataKey
   - code: LroLocationHeader
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
@@ -7680,10 +7680,7 @@
             "name"
           ]
         }
-      },
-      "required": [
-        "isSupported"
-      ]
+      }
     },
     "ShareInviteStatus": {
       "type": "string",
--- a/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
+++ b/specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
@@ -5136,6 +5136,17 @@
           "additionalProperties": {
             "type": "string"
           }
+        },
+        "consumptionEndTime": {
+          "type": "string",
+          "format": "date-time",
+          "description": "The timestamp after which a soft-deleted gallery image version is no longer consumable for VM/VMSS creation or VMSS scale out. It is calculated from the soft-deleted time plus the retention period, and is not present for active gallery image versions. In dateTime offset format.",
+          "readOnly": true
+        },
+        "imageState": {
+          "$ref": "#/definitions/GalleryImageVersionState",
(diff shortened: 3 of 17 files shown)


haagha (approved):

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+ */
+@added(Versions.v2026_03_03)
+model MetadataKeyValue {
+  /**
+   * The metadata key.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/secret-prop" "metadataKey is not sensitive, it is a key name for metadata"
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "x-ms-secret annotation required by Swagger linter to mark non-secret properties containing 'key' in name"
+  @extension("x-ms-secret", false)
+  metadataKey?: string;
```
ravimeda: **1. [NEW] 🔴 Blocking** **[[TypeSpec Anti-Patterns §5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns) / [TypeSpec §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp` — line 902-905 (also lines 3 and 10)

**Issue:** `@extension("x-ms-secret", false)` is used in TypeSpec source. §5 states: "**`@extension(...)` in TypeSpec source** -- never add `@extension(...)` decorators to TypeSpec source for any reason." Two `#suppress` directives were added to make it compile, and `import "@typespec/openapi"` (line 3) plus `using OpenAPI` (line 10) were added solely to enable this one decorator — they have no other use in the file.

Beyond the rule violation, **the decorator does not do what its suppression reason claims**, which independently violates §4.1 ("Suppression reasons **MUST** be factually accurate"). Verified against both linter implementations at their current sources:

- LintDiff `XMSSecretInResponse` (`packages/rulesets/src/spectral/functions/xms-secret-in-response.ts`) gates on `properties[prpName]["x-ms-secret"] !== true`. `false !== true` evaluates to **true**, so the rule still fires. `x-ms-secret: false` does not satisfy it — only `true` does.
- TypeSpec `secret-prop` (`packages/typespec-azure-resource-manager/src/rules/secret-prop.ts`) gates on `!isSecret(program, property)`. `@extension` does not set the compiler's secret flag, so this rule also still fires.

The PR's own commit sequence confirms it: the `@extension` was added first, and the readme suppression (finding 3) had to be added afterwards in commit `d287e52d` ("Suppress XMSSecretInResponse for MetadataKeyValue.metadataKey"). The suppression is what silences LintDiff; the decorator contributes nothing. Net effect: a forbidden decorator, two unnecessary source suppressions, a meaningless `"x-ms-secret": false` published into the GA swagger (`GalleryRP.json` line 7198), and a false justification.

**Fix:** Rename the two properties so both linters' built-in key/value-pair escape hatch applies, then delete everything added to work around them.

```tsp
/**
 * Describes a key-value pair for image metadata.
 */
@added(Versions.v2026_03_03)
model MetadataKeyValue {
  /**
   * The metadata key.
   */
  key: string;

  /**
   * The metadata value.
   */
  value?: string;
}
```

This clears both linters by two independent mechanisms:

- `XMSSecretInResponse` — `isPotentialSensitiveProperty` requires `lowerName.endsWith(keyword) && !lowerName.startsWith(keyword)`; `"key".startsWith("key")` is true, so the property is never considered sensitive.
- `secret-prop` — `isKeyValuePairKeyProp` returns true when `property.name === "key"`, the model also has `value`, and `properties.size === 2`. All three hold. (This rule's copy of `isPotentialSensitiveProperty` has no `!startsWith` guard, so the key/value escape hatch is the only thing that clears it — which is why the model must keep exactly two properties named `key` and `value`.)

Then remove: both `#suppress` lines (902-903), the `@extension` (904), `import "@typespec/openapi"` (line 3), `using OpenAPI` (line 10), and the readme suppression (finding 3); and update `@identifiers(#[])` per finding 5. Making `key` required also satisfies the identity constraint in finding 5. These models are new in this PR and unshipped, so renaming is not a breaking change.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -1361,6 +1474,13 @@ model ImageVersionSecurityProfile {
    * Contains UEFI settings for the image version.
    */
   uefiSettings?: GalleryImageVersionUefiSettings;
+
+  /**
+   * Specifies the secrets provisioning settings for the gallery image version. Used on create or update to configure secrets provisioning.
+   */
+  @added(Versions.v2026_03_03)
+  @visibility(Lifecycle.Create, Lifecycle.Update)
+  secretsProvisioningSettings?: SecretsProvisioningSettings;
```
ravimeda, line 1493: **2. [NEW] 🔴 Blocking** **[[OAPI027](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi027-write-only-properties-are-forbidden) / [WHATIF-003](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/what-if-preflight-compliance.md#whatif-003-mark-write-only--secret-properties)]** `.../ComputeGallery/models.tsp` — line 1481-1483 / `.../stable/2026-03-03/GalleryRP.json` — line 7181-7187, JSON path `$.definitions.ImageVersionSecurityProfile.properties.secretsProvisioningSettings`

**Issue:** `secretsProvisioningSettings` is declared `@visibility(Lifecycle.Create, Lifecycle.Update)`, emitting `"x-ms-mutability": ["update", "create"]` with no `"read"`. That is a write-only property. OAPI027: "Properties **MUST NOT** be write-only. Every non-secret property that can be set on a resource **MUST** also be readable via GET," with the sole exception being secret properties returned as `null` on read. The TypeSpec form is called out verbatim: "A property with `@visibility(Lifecycle.Create, Lifecycle.Update)` (without `Lifecycle.Read`) is write-only and is **not allowed**."

This property is not a secret and is not annotated as one. Its entire payload is non-sensitive configuration and version metadata: `isSupported: boolean`, `osName?: string`, and `components?: SecretsProvisioningComponent[]` (each a `name`/`version` pair) — `GalleryRP.json` lines 7575-7599. The feature is *named* "secrets provisioning" but the settings object carries no secret material.

Consequences: ARM Template What-If reports a false create on every deployment because the property appears in the template but never in the GET response; ARM Change Analysis produces the same noise; and a caller can never verify what was actually configured.

**Supporting detail (not a separate finding):** the emitted `x-ms-mutability` sits as a sibling of `$ref` (`GalleryRP.json` lines 7182-7187). Per [SCHEMA-REF-SIBLINGS](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#241-ref-must-not-have-sibling-keywords-schema-ref-siblings), keywords alongside `$ref` are ignored by JSON Schema draft-04 compliant tooling — so the write-only intent may not even be reliably expressed. Whichever way tooling resolves it the outcome is wrong: either write-only (OAPI027 violation), or the annotation is silently dropped so the contract says one thing and the spec means another.

**Second-order effect:** because `ImageVersionSecurityProfile` is also reachable from `GalleryList` (finding 4), this write-only property is emitted into the `Galleries_List` and `Galleries_ListByResourceGroup` response schemas — models that are never request bodies. On that path the property can never be written *and* never be read, yet it still surfaces in every generated SDK.

**Fix:**

```tsp
/**
 * Specifies the secrets provisioning settings for the gallery image version.
 */
@added(Versions.v2026_03_03)
secretsProvisioningSettings?: SecretsProvisioningSettings;
```

If some sub-field genuinely must not be readable, split that specific field out and annotate it `@secret` rather than making the whole settings object write-only.

specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
```diff
@@ -384,6 +384,10 @@ suppressions:
   - code: XMSSecretInResponse
     reason: Existing secret fields maintained for backward compatibility.
     from: ComputeRP.json
+  - code: XMSSecretInResponse
+    reason: metadataKey is a metadata field name identifier, not a cryptographic secret or API key.
+    from: GalleryRP.json
+    where: $.definitions.MetadataKeyValue.properties.metadataKey
```
ravimeda, line 390: **3. [NEW] 🔴 Blocking** **[[Suppression Decision Framework](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#decision-framework) / [Never Valid](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppressions-that-are-never-valid)]** `.../Compute/readme.md` — line 387-390

**Issue:** A new `XMSSecretInResponse` suppression is added for a **brand-new** model on a **stable/GA** API version. The decision framework approves a suppression in exactly two cases — a false alarm the linter gets wrong, or a pre-existing violation carried forward. **Neither applies.** The rule fires correctly (verified against its source in finding 1), and the readme diff against merge base `c0977d4a` shows these four lines are new in this PR. That lands on "Push the author to fix the violation if … **3. The approval criteria above are not met.**"

Compounding it, this is a security-rule suppression on a GA contract: "**Security-related rule suppressions** (`secret-prop`, `security-definition-missing`) require extra scrutiny and should almost never be suppressed." `XMSSecretInResponse` is the LintDiff counterpart of exactly the same secret-detection concern that `secret-prop` covers on the TypeSpec side — and this PR suppresses *both* for the same property.

The stated reason is correct on the facts — `metadataKey` genuinely is not a cryptographic secret — but the right way to express that is to name the property so the linters' key/value-pair escape hatch recognises it, not to suppress a security rule on a GA contract. A zero-suppression fix exists and costs two identifier renames.

**Fix:** Apply finding 1's rename, then delete the block:

```yaml
# remove these four lines from readme.md
- code: XMSSecretInResponse
  reason: metadataKey is a metadata field name identifier, not a cryptographic secret or API key.
  from: GalleryRP.json
  where: $.definitions.MetadataKeyValue.properties.metadataKey
```

Removing the suppression is only safe once the rename lands in the same PR — the rename is what stops the rule from firing.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+model ImageMetadataProfile {
+  /**
+   * The type of metadata.
+   */
+  type?: MetadataType;
+
+  /**
+   * The list of public metadata key-value pairs.
+   */
+  @identifiers(#[])
```
ravimeda: **5. [NEW] 🟠 Warning** **[[TSP-ARRAY-IDENTIFIERS](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#65-array-identifiers--x-ms-identifiers-tsp-array-identifiers)]** `.../ComputeGallery/models.tsp` — line 926, line 932, line 970, line 1026

**Issue:** All four new arrays declare `@identifiers(#[])`, emitting `"x-ms-identifiers": []`. §6.5 reserves the empty form for "arrays whose items are ordered result records, primitive values, or otherwise have no stable identity field." Each item type has an obvious identity property:

- line 926 `publicMetadataList?: MetadataKeyValue[]` — identified by the key property
- line 932 `internalMetadataList?: MetadataKeyValue[]` — identified by the key property
- line 970 `components?: SecretsProvisioningComponent[]` — identified by `name`
- line 1026 `imageMetadataProfiles?: ImageMetadataProfile[]` — identified by `type`

Declaring an empty identifier list tells ARM, What-If, and SDK tooling to treat these as positional arrays. For a metadata key-value list that produces incorrect diffing: reordering entries reads as a full replacement.

**Identity properties must be required.** Two of the four candidates are currently optional — `metadataKey?` (line 905) and `type?` (line 921). An identifier that items may omit is not a stable identity. Today every property of `ImageMetadataProfile` is optional and both properties of `MetadataKeyValue` are optional, so `{}` is a valid array element; combined with `@identifiers(#[])`, an array of empty objects is entirely indistinguishable to What-If and Terraform.

Making `type` and `key` required does **not** collide with finding 9. `imageMetadataProfiles` emits `"readOnly": true` (`GalleryRP.json` line 6276), so `ImageMetadataProfile` and `MetadataKeyValue` are response-only and never form part of the PATCH request contract — RPC-Patch-V1-10 does not reach them. `SecretsProvisioningComponent.name` is the opposite case: it is already required and *is* in the PATCH body, which is why finding 9 flags it and this finding does not ask to change it.

**Fix:**

```tsp
@identifiers(#["key"])          // line 926, after the finding-1 rename; `key` made required there
publicMetadataList?: MetadataKeyValue[];

@identifiers(#["key"])          // line 932
internalMetadataList?: MetadataKeyValue[];

@identifiers(#["name"])         // line 970
components?: SecretsProvisioningComponent[];

@identifiers(#["type"])         // line 1026 — also change line 921 to `type: MetadataType;`
imageMetadataProfiles?: ImageMetadataProfile[];
```

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+  publicMetadataList?: MetadataKeyValue[];
+
+  /**
+   * The list of internal metadata key-value pairs.
+   */
+  @identifiers(#[])
+  internalMetadataList?: MetadataKeyValue[];
+}
```
ravimeda, line 925: **6. [NEW] 🟠 Warning** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md#detection-signals)]** `.../ComputeGallery/models.tsp` — line 894-934 / `.../stable/2026-03-03/GalleryRP.json` — line 7129-7205

**Issue:** Two detection signals fire on `MetadataKeyValue`:

- **Signal 1 (property-name match):** `metadataKey` ends with `key`, one of the listed keywords.
- **Signal 4 (suppressed lint rules):** `models.tsp` line 902 carries `#suppress "@azure-tools/typespec-azure-resource-manager/secret-prop"` and `readme.md` line 387-390 carries an `XMSSecretInResponse` suppression, both on this one property. Rule text: "Suppressing a secret lint rule is a **strong signal** that the property IS a secret and the annotation was skipped."

I accept the author's position that the *key name* is not a secret — finding 1 proposes the rename that removes both suppressions and makes that self-evident to the linters. The residual concern is the **value** side, which no rule and no linter currently covers. `publicMetadataList` and `internalMetadataList` are unconstrained `string` → `string` bags returned read-only in GET, LIST, PUT and PATCH responses, explicitly bound to secrets provisioning: `MetadataType` enumerates `SecretsProvisioningImageMetadata` and `UserProvidedSecretsProvisioningMetadata` (`GalleryRP.json` lines 7209-7212). Nothing in the schema constrains what may appear in `metadataValue`, and its name trips no heuristic — it ends with none of `access|credential|secret|password|key|token|auth|connection`, so neither `XMSSecretInResponse` nor `secret-prop` will ever inspect it. On a feature whose entire purpose is provisioning secrets, `metadataValue` is an unbounded and unreviewable channel into GA responses.

Note that `"x-ms-secret": false` is not a documented way to assert non-secrecy — the annotation has no defined negative form in the Azure guidance, and `XMSSecretInResponse` reads it as "not `true`", i.e. identical to absent. So the property is not annotated in any sense the rule recognises; the emitted `false` is inert. That is the second reason it should be removed (the first being finding 1).

Design note (not a rule citation): this is the same shape as the "`Record<unknown>` / `Record<string>` for typed data" anti-pattern — an untyped bag standing in for a known schema. That rule is about `Record<>`/`additionalProperties` and does not literally apply to an array of a typed model, but the failure mode is identical.

**Fix:** Either model the known metadata as typed properties on `ImageMetadataProfile`, or, if the key set must stay open, document the closed set of keys the service emits in each list's `@doc` and confirm in this PR that no secret material — provisioning tokens, bootstrap credentials, KEK/DEK references — can transit `metadataValue`. Any key that can carry sensitive material must be a separate `@secret`-annotated property, not a member of an untyped bag.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+
+/**
+ * Describes a component involved in secrets provisioning.
+ */
+@added(Versions.v2026_03_03)
+model SecretsProvisioningComponent {
+  /**
+   * The name of the component (e.g., "OS", "CloudInit", "AzureGuestAgent", "SecretsProvisioningLibrary").
+   */
+  name: string;
```
ravimeda: **7. [NEW] 🟠 Warning** **[[Enum Candidates](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/enum-best-practices.md#actively-review-string-properties-for-enum-candidates)]** `.../ComputeGallery/models.tsp` — line 941-944 / `.../stable/2026-03-03/GalleryRP.json` — line 7561-7563, JSON path `$.definitions.SecretsProvisioningComponent.properties.name`

**Issue:** `enum-best-practices.md` directs: "**Actively examine every `string` property.** If the property name or description suggests it takes values from a limited or well-known set … **flag it** and recommend modeling it as an extensible enum." The doc comment names the set: `The name of the component (e.g., "OS", "CloudInit", "AzureGuestAgent", "SecretsProvisioningLibrary")`. Leaving it a bare `string` means no SDK constant surface, no client-side validation, and no docs enumeration — and once GA ships, converting it later is a breaking change for strongly typed SDKs.

**Fix:**

```tsp
/**
 * The name of a component involved in secrets provisioning.
 */
@added(Versions.v2026_03_03)
union SecretsProvisioningComponentName {
  string,

  /** The guest operating system. */
  OS: "OS",

  /** The cloud-init provisioning agent. */
  CloudInit: "CloudInit",

  /** The Azure guest agent. */
  AzureGuestAgent: "AzureGuestAgent",

  /** The secrets provisioning library. */
  SecretsProvisioningLibrary: "SecretsProvisioningLibrary",
}
```

and change line 944 to `name: SecretsProvisioningComponentName;`. Apply the same reasoning to `osName` (line 965), whose doc gives `"mariner"` as an example.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+
+/**
+ * Describes the secrets provisioning settings for a gallery image version.
+ */
+@added(Versions.v2026_03_03)
+model SecretsProvisioningSettings {
+  /**
+   * Specifies whether secrets provisioning is enabled.
+   */
+  isSupported: boolean;
```
ravimeda: **8. [NEW] 🟠 Warning** **[[OAPI034](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034-clear-field-ownership)]** `.../ComputeGallery/models.tsp` — line 957-960 / `.../stable/2026-03-03/GalleryRP.json` — line 7579-7581, JSON path `$.definitions.SecretsProvisioningSettings.properties.isSupported`

**Issue:** The property name and its documentation disagree, and the disagreement points at an unresolved ownership question. The name `isSupported` describes a **capability** — something the platform determines about an image. The doc comment says "Specifies whether secrets provisioning is **enabled**" — a **setting** the customer chooses. The property is currently client-writable (create/update, per finding 2) and `required`, which matches "enabled" and contradicts "supported".

OAPI034: "Every property **MUST** have clear ownership — either server-owned (`readOnly`) or client-owned," and "**Client-owned** properties are set by the user. The service **MUST NOT** silently overwrite or modify client-owned field values." As written a reader cannot tell which this is — and if the service computes support from the image contents and overwrites whatever the client sent, that is precisely the OAPI034 anti-pattern and will produce What-If noise on every deployment.

**Fix:** Decide the ownership and make name, doc, and visibility agree:

- Client-owned setting → rename to `enabled` (or `secretsProvisioningEnabled`), keep it writable, and make it readable per finding 2.
- Server-computed capability → keep `isSupported`, mark it `@visibility(Lifecycle.Read)`, and drop it from the required set.
- Both concepts are real → model them as two properties, e.g. `enabled?: boolean` (client-owned) and `@visibility(Lifecycle.Read) isSupported?: boolean` (server-owned), per the OAPI034 worked example.

specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
```diff
@@ -7466,6 +7554,49 @@
...
+          "description": "The list of component versions involved in secrets provisioning.",
+          "items": {
+            "$ref": "#/definitions/SecretsProvisioningComponent"
+          },
+          "x-ms-identifiers": []
+        }
+      },
+      "required": [
+        "isSupported"
+      ]
```
ravimeda: **9. [NEW] 🟠 Warning** **[[RPC-Patch-V1-10](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#41-patch-body-must-not-have-required-properties-rpc-patch-v1-10)]** `.../stable/2026-03-03/GalleryRP.json` — line 7596-7598 and line 7570-7573, JSON paths `$.definitions.SecretsProvisioningSettings.required` and `$.definitions.SecretsProvisioningComponent.required` / `.../ComputeGallery/models.tsp` — line 960, line 944, line 949

**Issue:** RPC-Patch-V1-10: "PATCH request body parameters **MUST NOT** have any properties marked as `required`, **MUST NOT** have `default` values, and **MUST NOT** have `x-ms-mutability` set to only `["create"]`." This PR introduces three new required properties inside the PATCH body graph. The full chain is checkable in the swagger without leaving the file:

> `GalleryImageVersionUpdate` (line 6363) → `properties` (line 6367) → `GalleryImageVersionProperties` (line 6368) → `securityProfile` (line 6257) → `ImageVersionSecurityProfile` → `secretsProvisioningSettings` (line 7181) → `SecretsProvisioningSettings` (`required: ["isSupported"]`, line 7596-7598) → `components[]` → `SecretsProvisioningComponent` (`required: ["name", "version"]`, line 7570-7573).

CI passes only because of the blanket `PatchBodyParametersSchema` suppression at `readme.md` line 345-347, which has **no `where:` scope** (file-wide for `GalleryRP.json`) and whose reason is "PATCH and PUT follow the same behavior and response codes in Compute. Keeping it for legacy reasons." That justification covers legacy shapes; it does not extend to models introduced in this PR. Note also that RPC-Patch-V1-10 is a **contract** requirement — the instruction text names `PatchBodyParametersSchema` only parenthetically — so a green LintDiff run under a blanket suppression is not evidence of compliance.

Practically, a JSON Merge Patch client that wants to change only `osName` is forced to resend `isSupported`, and `"secretsProvisioningSettings": {"osName": "..."}` is a schema-invalid request.

**Fix:** Make all three optional — `isSupported?: boolean;` (line 960), `name?: SecretsProvisioningComponentName;` (line 944), `version?: string;` (line 949) — and enforce the create-time requirements server-side. The `isSupported` case also resolves if finding 8 is fixed by marking it read-only. Do not widen the existing blanket suppression to cover new models.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, manual decision) against commit [`d287e52`](https://github.com/Azure/azure-rest-api-specs/pull/45051/commits/d287e52d0279633547087f4f42c22aa4e3a9e22c). See inline comments for findings 1-3, 5-9, 11-14. Findings 4 and 10 concern lines unchanged in this PR's diff (finding 10 concerns the examples directory, which has no diff line at all) and are posted as top-level PR comments rather than inline.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+   * The list of public metadata key-value pairs.
+   */
+  @identifiers(#[])
+  publicMetadataList?: MetadataKeyValue[];
+
+  /**
+   * The list of internal metadata key-value pairs.
+   */
+  @identifiers(#[])
+  internalMetadataList?: MetadataKeyValue[];
```
ravimeda, line 924: **11. [NEW] 🔵 Suggestion** `.../ComputeGallery/models.tsp` — line 929-933 / `.../stable/2026-03-03/GalleryRP.json` — line 7145-7152 — **`internalMetadataList` publishes an audience-scoped name in a GA contract.** _(No rule anchor — this is a design question, not a rule violation. I searched the instruction files and shared references and found no rule governing "internal" in client-facing property names.)_ The property is read-only, so customers see it in every GET and LIST response, and its pairing with `publicMetadataList` implies a trust or visibility boundary the contract never describes: the description ("The list of internal metadata key-value pairs") does not say who "internal" refers to, what a customer should do with it, or why Microsoft-internal data would appear in a customer response. Either the data is not actually internal, in which case the name misleads and should describe the data (e.g. `platformMetadata`, `serviceMetadata`); or it is genuinely Microsoft-internal, in which case it should not be in the public contract. Worth resolving before GA, since renaming afterwards is breaking.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+
+  /**
+   * Metadata related to user-provided secrets provisioning for the image.
+   */
+  UserProvidedSecretsProvisioningMetadata: "UserProvidedSecretsProvisioningMetadata",
+
+  /**
+   * Unknown metadata type.
+   */
+  Unknown: "Unknown",
```
ravimeda: **12. [NEW] 🔵 Suggestion** `.../ComputeGallery/models.tsp` — line 888-891, line 921 — `MetadataType` includes an explicit `Unknown: "Unknown"` member. The union already includes the `string` base type (line 876) and emits `modelAsString: true`, so unrecognized server values already round-trip safely as raw strings — that is what makes the enum extensible. An explicit `Unknown` sentinel adds a second, ambiguous representation of the same condition: a client cannot tell whether the service classified the metadata as type "Unknown" or simply returned a value the client's SDK does not know. This mirrors the guidance against `NotSpecified` in [provisioning-state.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#invalid-values) ("internal sentinel with no customer-facing purpose"). Compounding it, `type` is optional (line 921), so an entry can also carry *no* type — three ways to say "unclassified". Consider removing the `Unknown` member and making `type` required (which finding 5 needs anyway).

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+  /**
+   * The type of metadata.
+   */
+  type?: MetadataType;
+
+  /**
+   * The list of public metadata key-value pairs.
+   */
+  @identifiers(#[])
+  publicMetadataList?: MetadataKeyValue[];
```
ravimeda, line 918: **13. [NEW] 🔵 Suggestion** `.../ComputeGallery/models.tsp` — line 927, line 933 — [naming-conventions.md — Plural vs. Singular](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/naming-conventions.md#plural-vs-singular) requires array properties to have plural names. `publicMetadataList` and `internalMetadataList` are singular nouns with a `List` suffix; the collection nature is already conveyed by the array type. Consider `publicMetadata` / `internalMetadata` (or `publicMetadataEntries`). `components` (line 971) and `imageMetadataProfiles` (line 1027) already follow the convention.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
```diff
@@ -866,6 +868,109 @@ model GalleryImageUpdate extends UpdateResourceDefinition {
...
+
+/**
+ * Describes the metadata profile of an image.
+ */
+@added(Versions.v2026_03_03)
+model ImageMetadataProfile {
+  /**
+   * The type of metadata.
+   */
+  type?: MetadataType;
```
ravimeda: **14. [NEW] 🔵 Suggestion** `.../ComputeGallery/models.tsp` — line 921 — `ImageMetadataProfile.type` generates an SDK member named `Type`, which collides with language built-ins in several targets (`System.Object.GetType` in .NET, `type` in Python, `Type` in Java reflection contexts) and typically forces emitter-side renames or `@clientName` overrides later. `metadataType` would be unambiguous at every call site. Cheap to change now while the model is unshipped.

ravimeda:
**4. [EXISTING]** **[[RPC-Get-V1-09](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#22-tracked-resource-requirements)]** `.../ComputeGallery/models.tsp` — line 1873-1878 / `.../stable/2026-03-03/GalleryRP.json` — line 6553-6556, JSON path `$.definitions.GalleryList.properties.securityProfile`

**Issue:** `GalleryList` — the paged response for `Galleries_List` and `Galleries_ListByResourceGroup` — carries a third top-level property `securityProfile` referencing `ImageVersionSecurityProfile`, alongside `value` and `nextLink`. RPC-Get-V1-09: "Collection GET responses **MUST** only have `value` and `nextLink` as top-level properties." Sibling list models `SharedGalleryList` and `GalleryImageList` were both verified to have exactly `value, nextLink`; an *image version* security profile on a *list of galleries* is semantically meaningless. This looks like a stray property introduced during TypeSpec conversion. It is silenced today by the blanket `GetCollectionOnlyHasValueAndNextLink` suppression at `readme.md` line 355-357 ("Existing issue from last version").

**Previous version:** Also present at merge base `c0977d4a9fa5c9a95eaa64df887926a48e87fe22` and in shipped stable `stable/2025-12-03/GalleryRP.json` (`value, nextLink, securityProfile` in both). Not introduced by this PR.

**Why it matters here:** this PR adds `secretsProvisioningSettings` to `ImageVersionSecurityProfile`, so the new write-only property (finding 2) is now emitted into the `Galleries_List` and `Galleries_ListByResourceGroup` response schemas. Those are response-only models and the property is write-only, so it is unreachable in both directions — dead surface that still lands in every generated SDK.

**Fix:**

```tsp
/**
 * The List Galleries operation response.
 */
model GalleryList is Azure.Core.Page<Gallery>;
```

Because `securityProfile` has shipped in GA responses, removal must go through the breaking-change approval process. At minimum, file a tracked work item and reference it from the suppression rather than leaving it as undated debt.

ravimeda:
**10. [NEW] 🟠 Warning** **[[EX-COVERAGE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2211-example-coverage-ex-coverage)]** `.../stable/2026-03-03/examples/galleryExamples/`

**Issue:** EX-COVERAGE requires that "PUT and PATCH operations **SHOULD** have at least two examples (minimum + maximum property set)." `GalleryImageVersions_CreateOrUpdate` and `_Update` have many examples between them, but none is a true maximum-property-set example for this API version: **0 of the 23 `GalleryImageVersion_*` example files contain `secretsProvisioningSettings` or `imageMetadataProfiles`**. The PR changes exactly three files (verified against merge base `c0977d4a`) and adds no example.

So the new GA surface ships with no example coverage in public docs or SDK samples. The gap is worst on the read side: `imageMetadataProfiles` is read-only, so `GalleryImageVersion_ListByGalleryImage.json` — the only LIST example, and the only example bound to that operation — is the single place a customer could ever see the shape of `publicMetadataList` and `internalMetadataList`. That is also exactly the open question raised in finding 6, so the example gap and the security question close together.

The service has already established the convention for this case: `GalleryImageVersion_Create_WithVHD_UefiSettings.json` is a dedicated example for the other `securityProfile` sub-feature.

**Fix:** Add examples following the existing naming convention and wire them into the respective `x-ms-examples` maps, using realistic values rather than filler:

- `GalleryImageVersion_Create_WithSecretsProvisioningSettings.json` — maximum-property-set create showing `securityProfile.secretsProvisioningSettings` with real component names and versions.
- `GalleryImageVersion_Get_WithImageMetadataProfiles.json` — read path showing a populated `imageMetadataProfiles`.
- Update `GalleryImageVersion_ListByGalleryImage.json` to include `imageMetadataProfiles` in at least one array element.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: I've removed the  https://github.com/extension  decorator, both  #suppress  directives, the  import "@typespec/openapi" , and  using OpenAPI
Regarding the property rename ( metadataKey  →  key ): the CAPS backend already serializes these fields as  metadataKey  and  metadataValue  in its wire-format responses. Renaming in the spec would create a contract mismatch with the deployed service. Since the wire-format name is fixed by the existing backend, I've kept the  readme.md  LintDiff suppression as the legitimate mechanism for this case, with an updated reason explaining the constraint.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: As OSNames are open ended not validating them against a fixed set

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: Specifies whether the image version supports secrets provisioning, hence keeping isSupported

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: The naming mirrors the existing CAPS data model which has  PublicMetadataList  and  InternalMetadataList  on  ImageMetadataProfile . In this context, "internal" refers to metadata consumed by Azure platform services (e.g., for orchestration and diagnostics) rather than metadata intended for end-customer consumption — it is not confidential, but it's not actionable by the customer either.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: This is the name decided in contract

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: type  matches the CAPS backend serialization ( ImageMetadataProfile.Type  in the data model). Renaming to  metadataType  in the spec would create a wire-format mismatch. Since the union itself is called  MetadataType , using  type  as the property name is consistent with how discriminator fields are typically named in Azure APIs

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: Removed the Create, Update lifecycle

specification/compute/resource-manager/Microsoft.Compute/Compute/readme.md
awadhavekar: Responded in #1

specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
awadhavekar: We need the isSupported flag to be in required as that is the main switch for the feature

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
gary-x-li: "The backend already behaves this way" isn't a sufficient justification here. The spec is the contract and is intended to be reviewed before the implementation is locked in — otherwise the review can't actually influence the design.

I wouldn't block your PR for this; but can you reconsider changing your backend logic? At the very minimum, please keep in mind for future spec reviews.

specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
gary-x-li: > We need the isSupported flag to be in required as that is the main switch for the feature

Thanks — that explains why `isSupported` is important, but it doesn't resolve what RPC-Patch-V1-10 is guarding against, which is JSON Merge Patch semantics rather than the property's domain importance.

`secretsProvisioningSettings` is reachable through the update model (`GalleryImageVersionUpdate` → `properties` → `GalleryImageVersionProperties` → `securityProfile` → `ImageVersionSecurityProfile` → `secretsProvisioningSettings` → `SecretsProvisioningSettings`). With `required: ["isSupported"]` on that sub-object, a client who includes `secretsProvisioningSettings` to change only `osName` is forced by the schema to also resend `isSupported`. That is exactly the partial-update friction the rule exists to prevent — "it's the main switch" is orthogonal to it.

Note also that LintDiff is green here only because of the file-wide `PatchBodyParametersSchema` suppression (`readme.md`, no `where:` scope, reason "legacy reasons"). That blanket legitimately covers pre-existing shapes; it does not extend to a **new** model introduced in this PR, so a passing run isn't evidence of compliance for `SecretsProvisioningSettings`.

The domain requirement ("must be provided when configuring the feature") is real and can be kept without breaking PATCH:

- **Preferred:** make `isSupported` optional in the schema and enforce "required on create" at the service layer (or via a create-specific model). The property stays the semantic switch without breaking mergeability.
- **If the team insists on keeping it `required`:** add an explicit, narrow `where:`-scoped `PatchBodyParametersSchema` suppression targeting `$.definitions.SecretsProvisioningSettings` with a real justification, rather than relying on the legacy blanket — and get it reviewed rather than silently carried.

Leaving this open for your call.

gary-x-li:
Please see my responses to thread 3 and 9.

Also, please fix TypeSpec Validation error.

specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeGallery/models.tsp
awadhavekar: Understood and appreciate the feedback. The  metadataKey / metadataValue  naming is an established convention across multiple asset types in the CAPS platform — it was decided and implemented well before this spec PR. Changing it now would require a coordinated rename across all existing consumers. That said, point taken for future spec reviews — we'll aim to have the spec drive the design from the start. Thank you

awadhavekar: Will keep this in mind in the future, also these names were decided long back and are used for all artifacts. Changing logic in backend is infeasible

specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-03-03/GalleryRP.json
awadhavekar: Have made the change in backend

gary-x-li (approved):

haagha (approved):

PR #45066, opened 2026-07-28
Set release plan based on the latest API version

praveenkuttappan:
Currently automation marks release plan type as preview if any of the changed files is for preview version. This is incorrect. Release plan type should be derived based on the on the final API version detected from the changes in this PR to avoid creating incorrect release plan type when a spec PR contains changes in old preview version and adds a new stable version.

### DIFF
--- a/eng/tools/release-plan/src/typespec-project.ts
+++ b/eng/tools/release-plan/src/typespec-project.ts
@@ -27,8 +27,7 @@ export const FOLDER_MIGRATION_LABEL = "FolderMigrationV2";
  * @param params.repo Repository name
  * @param params.workspace Absolute path to workspace root
  * @param params.octokit Octokit instance for GitHub API calls
- * @returns TypeSpec project info with path and API version, or null if no spec changes, no projects found, or multiple projects found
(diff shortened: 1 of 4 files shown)


praveenkuttappan:
@helen229 FYI..

eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -172,7 +174,10 @@ export async function getTypeSpecProjectInfoFromCommit(params: {
   );
 
   if (versionResult.apiVersions.length === 0) {
-    throw new Error("No API version found in modified files or TypeSpec project path/content.");
+    console.log(
```
samvaity, line 177: console.log  on the "no API version" paths - is that consistent with how other skip conditions are logged?

samvaity (approved):
Lgtm

eng/tools/release-plan/src/typespec-project.ts
praveenkuttappan: yeah, it's a silent completion.

raych1 (approved):

PR #45079, opened 2026-07-28
Add an agentic data-plane API reviewer (label-gated, merges dormant)

xirzec:
> _This pull request was authored by Copilot on behalf of @xirzec._

## This merges dormant

The reviewer **does not run unless someone asks it to.** It is gated on the `data-plane-api-review-needed` label and never fires unprompted. Merging it changes the behaviour of exactly zero existing pull requests.

Security posture, unchanged from the original design:

- `checkout: false` — the agent reads PR files through the GitHub MCP toolsets, never from a working tree
- **no mutating GitHub tools** — the tool allowlist is read-only, and `safe-outputs` is the only write channel
- `min-integrity: approved`
- label-gated on `pull_request_target`, with a pre-step that removes the trigger label

Read this as *safe to merge dormant*, not *proven reviewer*. The evidence below is deliberately stated in both directions.

---

## Where the numbers actually stand

Two different criteria have been conflated in earlier discussion. They are both true and they mean different things.

| Criterion | Status |
| --- | --- |
| **Phase-2 promotion gate** — zero blocking-severity false positives on the true-negative suite | **Passing.** 0 across 27 TN trials, on each of the last two full runs. |
| **Runner hard-fail rule** — *any* TN trial failure blocks the run | **Failing.** 19/27 trials pass (70.4%). |

The promotion gate is the one that governs whether the reviewer may be rolled out more broadly. The hard-fail rule is a stricter internal tripwire that counts warning- and suggestion-severity output too.

**The 70.4% figure is noise-dominated at this sample size and should not be read as a precise measurement.** With 27 trials, a single trial is 3.7 points. This run had three stimuli improve and three regress, cancelling exactly. Nothing in this data distinguishes 70.4% from 75%.

### Detection is validated

- **Positives: 12/13.** `resource-modeling`, `naming-and-docs`, `error-design`, `visibility-and-secrets` and `lro-and-paging` suites all at **100%**.
- Every mechanical grader that asserts a finding matched a *real* finding in finding position — no grader passes on a vacuous match.

### Format instability now causes more trial failures than judgment

This is the most important qualification on the numbers, and it is not a judgment problem.

Roughly **1 trial in 20** fails because the report's *presentation* deviates, not because the review is wrong. It migrates between stimuli rather than sticking to one.

The worked example is `stable-version-breaking-changes`, which fell from 3/3 to 1/3 in the latest run. The agent found **all four** breaking changes — property removed, type changed `int32`→`float64`, open union closed with a member dropped, optional request property made required — explained each correctly, and assigned Blocking severity appropriately. It then wrote `### 🚫 Blocking` instead of `### 🔴 Blocking` and used a numbered list instead of the bracketed finding form. Both mechanical graders failed on syntax while the LLM judge passed the content.

The graders have since been made tolerant of glyph and position variation. The bracketed rule ID is deliberately **not** loosened: the "rules considered and declined" table writes plain bold IDs without brackets, so the bracket is the only thing separating a raised finding from a declined one. That means the failure mode above is *reduced, not eliminated* — a report that drops brackets entirely will still fail, and that is the correct trade.

---

## What this PR contains

- **A data-plane reviewer agent** (`.github/agents/data-plane-api-reviewer.agent.md`) plus a false-positive-defence critic, following the existing ARM reviewer/critic pattern.
- **Data-plane reference files** under the existing `azure-api-review` skill — resource modeling, LRO/paging, error design, naming and docs, visibility and secrets, design decisions, report format, and a **linter interlock** that tells the agent which guidelines CI already enforces so it does not double-report.
- **A label-gated gh-aw workflow**, dormant until the label is applied.
- **An eval harness**: 23 fixtures, 48 mechanical graders, 7 eval files, including a true-negative suite that measures false positives directly.
- **CI alignment checks** (`data-plane-review-alignment.yaml`) that fail the build when the linter pin, the model pins, the report-format contract, or the graders drift apart — plus a job that compiles every fixture.

## Reviewers should look at this file specifically

**`.github/skills/azure-api-review/references/enum-best-practices.md` is shared with the ARM reviewer, and this PR changes it.** The change restores an `unless` clause that the upstream Guidelines actually contain and this file had flattened into an unconditional requirement. It corrects real behaviour for the ARM reviewer as well as the data-plane one. If you review nothing else here, review that.

## Honest notes on methodology

- **Eleven instrumentation defects were found during development, and every one was in the measurement apparatus rather than in the reviewer.** Four were in graders that had never been executed; one was a rubric that YAML silently parsed as a map, so a grader crash was scored as a stimulus failure. The standing lesson is recorded in the eval README: when a measured run disagrees with expectation, suspect the instrument first.
- **The fixture corpus has been through independent review** by an agent with no stake in it. That review found the whole corpus had never compiled — every fixture carried a `@useDependency` on a namespace removed from `typespec-azure-core`. All 23 now compile, and CI enforces it.
- **Phase 0 is a dark launch**, and it requires merging first. The runner hard-fail rate is a phase-2 promotion criterion, not a merge criterion.

## What is still open

- The true-negative corpus is small (5 distinct TN fixtures) and two open fixture questions are with independent review.
- Format instability, as described above.
- No production traffic has been reviewed yet. Phase 0 exists to fix that.


.github/agents/data-plane-api-reviewer.agent.md
```diff
@@ -0,0 +1,379 @@
...
+# `create_pull_request_review`, `remove_labels`, `update_review_comment`,
+# or any other mutating GitHub tool to this list.
+tools:
+  - agent
+  # GitHub read-only:
+  - github/get_file_contents
+  - github/get_pull_request
+  - github/get_review_comments
+  - github/list_pull_request_files
+  - github/search_code
```
heaths, line 19: That's a good idea I should incorporate into my custom agent for the Rust SDK. Copilot just suggested `github/*` but I like the read-only aspect here.

.github/agents/data-plane-api-reviewer.agent.md
```diff
@@ -0,0 +1,379 @@
...
+
+You review **data-plane TypeSpec** against the
+[Azure REST API Guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md)
+in the areas that deterministic lint cannot reach.
+
+## Persona and calibration (read this first, every time)
+
+You are an experienced Azure API reviewer. You are also, in the unattended
+case, a bot posting into a service team's pull request, and that changes the
+economics:
```
heaths: Overall: too wordy. I suggest my approach of distilling these to just what the agent thinks it needs but remain human reviewable. This is burning a lot of tokens and much of these words/punctuation is just fluff anyway. And as we've seen with more context, it can actually lead to worse results.

.github/agents/data-plane-api-reviewer.agent.md
```diff
@@ -0,0 +1,379 @@
...
+  result and you must be willing to produce it. Do not manufacture findings to
+  demonstrate effort.
+- **Every finding anchors to a line and a section.** If you cannot cite a
+  specific `file:line` in the diff _and_ a specific Guidelines section or a
+  rule ID from the data-plane references, you do not have a finding. You may
+  have a question -- ask it, at Suggestion severity, or drop it.
+- **When in doubt, ask rather than assert.** A question costs the author ten
+  seconds. A wrong assertion costs them a rebuttal and costs you their trust.
+
+This is a deliberate deviation from `arm-api-reviewer.agent.md`'s "Be critical.
```
heaths: I don't think - but I don't know - it's a good idea to reference another agent here. That most likely would cause it to be read in - at least the frontmatter - and may produce conflicting instructions.

.github/skills/azure-api-review/SKILL.md
```diff
@@ -70,21 +117,114 @@ files, and reference files are designed and maintained:
...
-3. **Low maintenance overhead.** The files are designed so that
+1. **Low maintenance overhead.** The files are designed so that
    maintainers spend minimal effort. Updates are needed only when
    upstream guidance (RPC contract, Azure REST API Guidelines, ARM
    wiki) changes. When that happens, a maintainer refreshes the
    affected instruction/reference files, updates the `Upstream
 alignment` date, and the change propagates through all
    cross-references automatically.
 
+## Normative strength and documented rationale
```
heaths, line 153: Another good idea I should clarify in my custom agent.

.github/skills/azure-api-review/references/data-plane-lro-and-paging.md
```diff
@@ -0,0 +1,160 @@
...
+`@azure-tools/typespec-azure-core/long-running-polling-operation-required`
+enforces that an LRO _has_ a linked polling operation. It does not judge whether
+the status monitor is usable, whether terminal states are complete, or whether
+paging is consistent. That is this file.
+
+---
+
+## LRO
+
+### DP-LRO-01: Status monitor completeness
```
heaths, line 30: There are `@lroSucceeded` and `@lroFailed` (or something like that) decorators as well for custom `@lroStatus` monitors you should document. This just came up in a review I did this morning. Ideally, they didn't even define their own (grandfathered LROs - we need to revisit grandfathering or this will never significantly improve), but since they did, they need to use these decorators as well.

.github/cspell.yaml
```diff
@@ -105,3 +105,11 @@ overrides:
...
     words:
       - vally
+
+  - filename: "**/.github/workflows/data-plane-review-alignment.yaml"
+    words:
+      - vally
+
+  - filename: "**/.github/workflows/src/data-plane-review-alignment.js"
+    words:
+      - vally
```
heaths, line 115: SDK repos are starting to adopt directory-or-descendant-specific CSpell configuration to avoid these global files needing more review. If you put a `.cspell.json` in, say, `.github/workflows` that inherits from this file, you can go to town with exemptions without requiring a global file reviewer.

.github/agents/data-plane-api-reviewer.agent.md
xirzec: (Replying as Copilot, the agent working this branch, on behalf of @xirzec.)

Adopted. The file is 32% smaller by characters (22,066 -> 14,930), 379 -> 317 lines, with no operative instruction removed.

Most of the reduction turned out to be structural rather than editorial, which I think supports your point more than mine:

- three Prettier-padded tables whose whitespace exceeded their content
- a step-summary table that restated the eight step headings immediately below it
- `../skills/azure-api-review/references/` repeated twelve times, now stated once
- maintainer rationale that explained decisions to humans instead of instructing the agent

Worth flagging the tension honestly rather than pretending it away: our own eval runs showed richer guidance raising AIU 28-33%, and detection held at 10/10 across both. So the added context was doing *something*. Your argument is that it is a cost and a quality risk regardless, and since the distillation removed padding rather than instructions, both things can be true. If detection regresses on the next measured run I will say so rather than quietly attributing it elsewhere.

.github/agents/data-plane-api-reviewer.agent.md
xirzec: (Replying as Copilot, the agent working this branch.)

Adopted, and I checked the mechanism rather than assuming, because it is not quite the one you guessed — the conclusion lands in the same place.

**Frontmatter is loaded regardless.** The runtime enumerates every `.github/agents/*.agent.md` and exposes each one's `name` and `description` as an available agent. I can see all four (both ARM agents, both data-plane agents) in my own agent list in this session, and none of them is referenced from anything I loaded. So naming a file in prose adds no frontmatter that was not already there.

**What the reference does add is an invitation to read the body.** `arm-api-reviewer.agent.md` is 168.9 KB. An agent that sees it named and pathed mid-review may reasonably open it, and that is both a large token cost and a genuine source of conflicting instructions — its persona is "be critical, assume every spec has issues until proven otherwise", which is close to the opposite of the calibration this reviewer needs.

All three references are gone. The deliberate critic dispatch stays, since that is a subagent call rather than an accidental read, and the critic's frontmatter is scoped to false-positive defence only.

Also worth noting for your own agent: the `description` field of every agent file is in context whether you want it or not, so it is worth writing as though the model will read it — which is roughly how I have treated it here.

.github/skills/azure-api-review/references/data-plane-lro-and-paging.md
xirzec: (Replying as Copilot, the agent working this branch.)

Adopted, and your "or something like that" was worth checking — the full family is `@lroStatus`, `@lroSucceeded`, `@lroFailed`, `@lroCanceled`, `@lroResult`, `@lroErrorResult`, all in `Azure.Core`.

Then compiling four probe specs changed the rule substantially, and I think this is the more useful half of the answer:

| probe | result |
| --- | --- |
| bespoke status vocabulary, nothing marked | `lro-status-missing` (**error**) + `polling-operation-no-status-monitor` |
| bespoke vocabulary, `@lroSucceeded`/`@lroFailed` applied | compiles clean |
| literal `"Succeeded"`/`"Failed"`, nothing marked | compiles clean |
| `Canceled` unmarked, no `@lroResult`/`@lroErrorResult` | compiles clean |

So a service **cannot** ship a custom status monitor without either the decorators or the literal `Succeeded`/`Failed` values — the linter stops it at error severity. My first draft of this rule told the agent to flag exactly that, which would have been a double-report of a failed build, and precisely the interlock violation this design exists to prevent.

What is genuinely uncovered is the residue, and that is what `DP-LRO-05` now covers, at Suggestion:

- `@lroCanceled` on an unmarked cancellation state — compiles clean, and a generated poller then treats cancellation as an unrecognised non-terminal state, so it hangs on a cancelled operation
- `@lroResult` / `@lroErrorResult`, so neither the result nor the error is typed for the SDK

Never Blocking, since a grandfathered monitor is usually breaking to change — which matches the situation you described.

I have also added rows for `polling-operation-no-status-monitor` and `lro-status-missing` to `data-plane-linter-rule-coverage.md`. The first had no row, and that file's own rule is that a rule with no row is agent-owned — so the agent would have duplicated a CI warning that one of our own fixtures actually trips.

On grandfathering more broadly: agreed, and out of scope here. This reviewer can at least make the residue visible on the specs that already have it.

.github/cspell.yaml
xirzec: (Replying as Copilot, the agent working this branch.)

Partly adopted — I reduced the exemptions rather than relocating them, because the relocation does not buy what it buys in the SDK repos. Two things I checked:

**1. This file already is the scoped, inheriting config.** `.github/cspell.yaml` starts with `import: ../cspell.yaml`, and it already carries three identical per-file `vally` overrides for the ARM eval suite (`:97`, `:101`, `:105`). So the pattern you are describing is in place here; my entries were following it.

**2. A config under `.github/workflows/` would still need the same reviewer.** `protected-files.yaml:48` matches with PowerShell `-like` against `.github/*`, and PowerShell wildcards match across `/`. So `.github/workflows/.cspell.json` is just as protected as `.github/cspell.yaml`. The only exempted subtree is `.github/skills/(?!azsdk-common-).+` (`:52`) — and my remaining entries are not there.

What I did instead was remove the need for two of the three. The SpellCheck failure was `NOSUCHRULE` and `vally` in a test file: the first was an invented rule ID in a fixture, now `DP-ABSENT-01`, which reads better anyway; the second was in a comment and is now "the eval harness". Both gone rather than exempted.

The two that remain are unavoidable — there `vally` is a *path segment*, `.github/skills/evals/data-plane-api-reviewer/vally/`, following the ARM suite's directory layout.

Genuinely useful correction though, and it prompted the audit that fixed the CI failure. If the scoped-config pattern is worth having repo-wide, the blocker is `protected-files.yaml`'s glob rather than the cspell side, and that seems worth raising separately.

heaths (approved):

PR #45103, opened 2026-07-29
[AI Foundry][Data Plane] Add session_configuration.idle_timeout_seconds to HostedAgentDefinition

Deeksharma:

## Summary

Model the optional `session_configuration.idle_timeout_seconds` field on `HostedAgentDefinition` in TypeSpec so the service-side hand-written customization can be removed once regenerated. Ungated, so it applies to both `v1` and `virtual-public-preview`. `idle_timeout_seconds` is `int32`, range `[60, 3600]`, defaulting to the 900s server default when unset. Additive, non-breaking change.

## API Info: The Basics

- [ ] a private preview
- [x] a public preview
- [ ] GA release

### Change Scope

* Updated paths:
  * `specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp` — new `SessionConfiguration` model + optional `HostedAgentDefinition.session_configuration` property
  * Regenerated OpenAPI3 outputs for `v1` and `virtual-public-preview` (`.json` + `.yaml`)

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -370,6 +370,19 @@ model ContainerConfiguration {
   image: string;
 }
 
+@doc("Session defaults applied to sessions created for a hosted agent version.")
+model SessionConfiguration {
+  @doc("""
+    The idle duration, in seconds, before a session's sandbox is suspended. Optional — when
+    unset, the server default of 900 seconds is used. Must be between 300 and 3600 seconds
+    (inclusive).
+    """)
+  @minValue(300)
+  @maxValue(3600)
+  @example(300)
+  idle_timeout_seconds?: int32 = 900;
+}
+
 @doc("Code-based deployment configuration for a hosted agent.")
 model CodeConfiguration {
   @doc("The runtime identifier for code execution (e.g., 'python_3_11', 'python_3_12', 'python_3_13').")
@@ -459,6 +472,10 @@ model HostedAgentDefinition extends AgentDefinition {
 
   @doc("Optional customer-supplied telemetry configuration for exporting container logs, traces, and metrics.")
   telemetry_config?: TelemetryConfig;
+
+  @doc("Optional session defaults (for example, the idle timeout) applied to sessions created for this agent version.")
+  @example(#{ idle_timeout_seconds: 300 })
+  session_configuration?: SessionConfiguration;
 }
 
 @doc("The container app agent definition.")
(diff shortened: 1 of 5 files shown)


Deeksharma:

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -357,6 +357,19 @@ model ContainerConfiguration {
...
+model SessionConfiguration {
+  @doc("""
+    The idle duration, in seconds, before a session's sandbox is suspended. Optional — when
+    unset, the server default (currently 900 seconds) is used. Must be between 60 and 3600
+    seconds (inclusive).
+    """)
+  @minValue(60)
+  @maxValue(3600)
+  @example(300)
+  idle_timeout_seconds?: int32;
```
dargilco: Please explicitly specify the default in the above line by adding `= 900`.

Deeksharma: Done

dargilco (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -393,6 +393,19 @@ model ContainerConfiguration {
...
+model SessionConfiguration {
+  @doc("""
+    The idle duration, in seconds, before a session's sandbox is suspended. Optional — when
+    unset, the server default of 900 seconds is used. Must be between 300 and 3600 seconds
+    (inclusive).
+    """)
+  @minValue(300)
+  @maxValue(3600)
+  @example(300)
+  idle_timeout_seconds?: int32 = 900;
```
glecaros:
```suggestion
  @encode(DurationKnownEncoding.seconds)
  idle_timeout_seconds?: duration = duration.fromISO("PT900S");
```

dargilco: Agree. In Python code this means it will be generated as type `datetime.timedelta` instead of just `int`.

Deeksharma: Thanks! Applied in 4b6fe5bb7c9 — `idle_timeout_seconds` is now a `duration` with `@encode(DurationKnownEncoding.seconds, int32)`.

Note I adjusted the suggestion slightly so it compiles and stays wire-compatible:
- Added the `int32` target to `@encode` (without it, the compiler serializes the duration as an ISO-8601 string rather than an integer).
- Converted `@minValue`/`@maxValue`/`@example` to duration values (e.g. `duration.fromISO("PT300S")`), since those decorators require a value assignable to `duration`.

Verified by compiling: the emitted OpenAPI is byte-identical to before — `type: integer, format: int32, minimum: 300, maximum: 3600, default: 900, examples: [300]` — so there's no wire change, and client libraries now get the idiomatic duration representation.

glecaros (approved):

PR #45104, opened 2026-07-29
Fix .NET emitter-output-dir in aimanager tspconfig

PugDeveloper:
Fixed .NET `emitter-output-dir` to `{output-dir}/{service-dir}/{namespace}` so it uses `service-dir` like the other languages.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/tspconfig.yaml
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/tspconfig.yaml
@@ -24,7 +24,7 @@ options:
     flavor: azure
   "@azure-typespec/http-client-csharp-mgmt":
     namespace: "Azure.ResourceManager.ContainerServiceAIManager"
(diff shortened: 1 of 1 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aimanager/tspconfig.yaml
```diff
@@ -24,7 +24,7 @@ options:
     flavor: azure
   "@azure-typespec/http-client-csharp-mgmt":
     namespace: "Azure.ResourceManager.ContainerServiceAIManager"
-    emitter-output-dir: "{output-dir}/sdk/aimanager/{namespace}"
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
```
purpleran7, line 27: we're using the same pattern as fleet, DSG,

emitter-output-dir: "{output-dir}/sdk/fleet/{namespace}"
emitter-output-dir: "{output-dir}/sdk/containerservicesafeguards/{namespace}"

PR #45110, opened 2026-07-29
Automation to create a release plan to use TypeSpec metadata emitter

praveenkuttappan:
Run TypeSpec metadata emitter to identify the API version and create a release plan using this version as the base instead of using REGEX parsing the changed files to identify the version.

New logic:

1. Automation run TypeSpec metadata emitter to identify API version when a spec PR is merged that either has new-api-version label or updates a. tsp or tspconfig.yaml
2. Searches for an existing release plan for the same release type based on the API version (preview or stable)
3. Create a release plan if a release plan does not exist.


eng/tools/release-plan/src/index.ts
```diff
@@ -59,8 +87,6 @@ export async function main(): Promise<void> {
       });
 
       resolvedPrNumber = args.prNumber;
-
-      hasNewApiVersionLabel = labels.includes(NEW_API_VERSION_LABEL);
     } else {
       const commitSha = args.commitSha as string;
```
raych1, line 92: Curious - what's the case by passing commitSha?

eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -367,6 +351,202 @@ export function findTspConfigDir(relativeFilePath: string, workspace: string): s
...
+ * @throws Error if tsp compile fails or if main.tsp and client.tsp don't exist
+ */
+export function runTypeSpecMetadataEmitter(tspProjectPath: string): TypeSpecMetadataResult {
+  const mainTspPath = join(tspProjectPath, "main.tsp");
+  const clientTspPath = join(tspProjectPath, "client.tsp");
+
+  let compileCommand: string;
+  if (existsSync(mainTspPath)) {
+    compileCommand =
+      'npm exec -- tsp compile . --emit @azure-tools/typespec-metadata --output-dir "./tsp-output" --option @azure-tools/typespec-metadata.format="json"';
```
samvaity: The `compileCommand` string is built here but never used. The actual execution below uses `spawnSync` with a separate args array. Dead code?

eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -367,6 +351,202 @@ export function findTspConfigDir(relativeFilePath: string, workspace: string): s
...
+      "compile",
+      existsSync(mainTspPath) ? "." : "./client.tsp",
+      "--emit",
+      "@azure-tools/typespec-metadata",
+      "--output-dir",
+      "./tsp-output",
+      "--option",
+      "@azure-tools/typespec-metadata.format=json",
+    ],
+    {
```
samvaity: `console.log(result)` dumps the entire `SpawnSyncReturns` object (stdout, stderr, pid, signal, etc.) into CI logs. Should this be `console.log(result.stdout)` or removed?

eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -367,6 +351,202 @@ export function findTspConfigDir(relativeFilePath: string, workspace: string): s
...
+  if (!metadata.languages || typeof metadata.languages !== "object") {
+    throw new Error("TypeSpec metadata does not contain 'languages' object");
+  }
+
+  const versionAndTypeMap = new Map<string, Set<string>>();
+
+  for (const [, langConfigs] of Object.entries(metadata.languages)) {
+    if (!Array.isArray(langConfigs)) {
+      continue;
+    }
```
samvaity, line 372: This throws when multiple API versions are found across languages. The azsdk-cli version (PR #[16624](https://github.com/Azure/azure-sdk-tools/pull/16624/changes)) instead picks the first. Should these align? Otherwise the same project could work in CLI but fail here.

eng/tools/release-plan/src/index.ts
```diff
@@ -112,7 +159,7 @@ export async function main(): Promise<void> {
         testReleasePlan: args.testReleasePlan,
```
samvaity, line 167: `hasNewApiVersionLabel || isTspConfigChanged` means any `tspconfig.yaml` edit (even non-version changes like updating emitter options or output dirs) would trigger release plan creation. Could that produce spurious release plans for config-only changes?

eng/tools/release-plan/src/typespec-project.ts
```diff
@@ -367,6 +351,202 @@ export function findTspConfigDir(relativeFilePath: string, workspace: string): s
...
 }
 
+/**
+ * Runs TypeSpec metadata emitter to generate metadata JSON output.
+ * Checks for main.tsp first, then falls back to client.tsp.
+ * @param tspProjectPath Absolute path to TypeSpec project directory
+ * @returns Metadata result with apiVersion and sdkType
+ * @throws Error if tsp compile fails or if main.tsp and client.tsp don't exist
+ */
+export function runTypeSpecMetadataEmitter(tspProjectPath: string): TypeSpecMetadataResult {
```
raych1: Can you re-use the shared code https://github.com/Azure/azure-rest-api-specs/blob/aec845ea557d1bd76efb66729d6407acf15dafc3/.github/shared/src/typespec-metadata.js#L40

eng/tools/release-plan/src/index.ts
praveenkuttappan: It will still not create if a release plan already exists for that TypeSpec project path and API type/api version.

eng/tools/release-plan/src/index.ts
praveenkuttappan: This to support manually running the pipeline against a commit sha and also to support testing the pipeline without merging a spec PR.

eng/tools/release-plan/src/typespec-project.ts
praveenkuttappan: Only to show the the command in logs. I can remove it.

eng/tools/release-plan/src/index.ts
samvaity: That handles the duplicate case, but what about a first-time tspconfig.yaml edit that only changes emitter options (no version bump)? There's no existing plan to dedupe against, so it would create one. Maybe that's acceptable since the emitter would just pick up the current version anyway - just wanted to make sure that's expected.

eng/tools/release-plan/src/typespec-project.ts
praveenkuttappan: updated to use first version to march azsdk agent

eng/tools/release-plan/src/typespec-project.ts
praveenkuttappan: Updated to reuse the common code.

eng/tools/release-plan/src/index.ts
```diff
@@ -80,9 +106,30 @@ export async function main(): Promise<void> {
         process.exit(0);
       }
 
+      hasNewApiVersionLabel = commitResult.hasNewApiVersionLabel;
+
+      // Check if commit contains TypeSpec files (.tsp or tspconfig.yaml)
+      const commitFiles = await getCommitChangedFiles({
```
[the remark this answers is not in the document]
praveenkuttappan: Fixed

eng/tools/release-plan/src/index.ts
```diff
@@ -157,13 +204,37 @@ export async function main(): Promise<void> {
   } catch (error) {
     const message = error instanceof Error ? error.message : String(error);
     console.error(`release-plan tool failed: ${message}`);
+
+    // Try to post error comment on PR if available and PR has new-api-version label
+    if (resolvedPrNumber && projectInfo && (hasNewApiVersionLabel || isTspConfigChanged)) {
+      try {
+        await postReleasePlanErrorComment({
```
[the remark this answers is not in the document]
praveenkuttappan: Fixed

eng/tools/release-plan/src/pr-comment.ts
```diff
@@ -117,3 +117,68 @@ export function buildReleaseplanCommentBody(params: CommentBodyParams): string {
...
+  /** TypeSpec project path */
+  tspProjectPath?: string;
+}
+
+/**
+ * Post an error comment on a GitHub PR indicating release plan creation failed.
+ * @param params - Error comment parameters
+ * @throws Error if GitHub API call fails
+ */
+export async function postReleasePlanErrorComment(params: ErrorCommentParams): Promise<void> {
```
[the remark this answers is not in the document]
praveenkuttappan: Fixed

eng/tools/release-plan/src/index.ts
```diff
@@ -112,8 +167,9 @@ export async function main(): Promise<void> {
         testReleasePlan: args.testReleasePlan,
       },
       createAzdskRunner(),
-      hasNewApiVersionLabel,
+      hasNewApiVersionLabel || isTspConfigChanged,
```
raych1, line 170: Nit: it would be good to document the difference between this condition and line #132 for better understanding.

raych1 (approved):

eng/tools/release-plan/src/index.ts
praveenkuttappan: I will file an issue to remove tspfile condition based on how this goes.

PR #45113, opened 2026-07-29
Add Foundry trace PII redaction option

Hassona97:
## Summary
- Add optional `pii_redaction` boolean to trace data generation options only, defaulting to enabled when omitted.
- Rename the task generation data generation mode to `simulation_seed`.
- Update generated Foundry OpenAPI JSON/YAML outputs for `v1` and `virtual-public-preview`.

## Validation
- `azsdk_run_typespec_validation` succeeded for `specification/ai-foundry/data-plane/Foundry`.
- Focused OpenAPI checks confirmed `pii_redaction` is trace-only and `simulation_seed` replaces `task_generation`.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
@@ -25,8 +25,8 @@ union DataGenerationJobType {
   @doc("Tool calling conversation between user and agent.")
   tool_use: "tool_use",
 
-  @doc("Task generation for evaluation scenarios.")
-  task_generation: "task_generation",
+  @doc("Simulation seed for evaluation scenarios.")
+  simulation_seed: "simulation_seed",
 }
 
 @doc("LLM model options for data generation jobs.")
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -223,13 +223,16 @@ model SimpleQnADataGenerationJobOptions extends DataGenerationJobOptions {
 model TracesDataGenerationJobOptions extends DataGenerationJobOptions {
   @doc("The data generation job type, which is Traces for this model.")
   type: "traces";
+
+  @doc("Whether to redact personally identifiable information from traces. When omitted or set to true, PII is redacted. Set to false to opt out of redaction.")
+  pii_redaction?: boolean = true;
```
shivamMg: rename to `redact_private_content`

shivamMg (dismissed):

aprilk-ms (approved):

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -25,11 +25,10 @@ union DataGenerationJobType {
   @doc("Tool calling conversation between user and agent.")
   tool_use: "tool_use",
 
-  @doc("Task generation for evaluation scenarios.")
-  task_generation: "task_generation",
+  @doc("Simulation seed for evaluation scenarios.")
+  simulation_seed: "simulation_seed",
```
dargilco, line 29: This is breaking change. Do you plan to support both old and new properties in order to no break existing customers? (if there are any). We usually define v2 of the preview (`data_generation_jobs_v2_preview`) when there are breaking changes, where the TypeSpec (and therefore emitted SDKs and REST API ref docs) only document the v2 behavior, while the service supports both.

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
```diff
@@ -222,22 +218,23 @@ model SimpleQnADataGenerationJobOptions extends DataGenerationJobOptions {
...
 
-@doc("The options for a task generation data generation job. Use with multiturn evaluation scenarios and with prompt, file, or agent sources. Generated dataset rows include fields such as `id`, `category`, `test_case_description`, and `desired_num_turns`.")
+@doc("The options for a simulation seed data generation job. Use with multiturn evaluation scenarios and with prompt, file, or agent sources. Generated dataset rows include fields such as `id`, `category`, `test_case_description`, and `desired_num_turns`.")
 @extension("x-ms-foundry-meta", DataGenerationJobsRequiredPreviews)
-model TaskGenerationDataGenerationJobOptions extends DataGenerationJobOptions {
-  @doc("The data generation job type, which is TaskGeneration for this model.")
-  type: "task_generation";
+model SimulationSeedDataGenerationJobOptions extends DataGenerationJobOptions {
+  @doc("The data generation job type, which is SimulationSeed for this model.")
+  type: "simulation_seed";
```
dargilco: Use `DataGenerationJobType.simulation_seed` instead of the hard-coded string

specification/ai-foundry/data-plane/Foundry/src/data_generation_jobs/models.tsp
Hassona97: This feature is not released yet, so there is no customer impact. That's why the decision for renaming was an easy one.

dargilco: We released SDKs with "task_generation", and I'm assuming it's also in the latest REST API ref-code. Are you saying the service never supported it? So this should not have been in the TypeSpec to begin with...

Anyway, if you are saying the service never supported it, the change it good.

aprilk-ms: Yes, it is a new feature.

dargilco (approved):

PR #44951, opened 2026-07-24
Add scheduledEvents schema for 2026-06-01

jdbencardinop:
## Summary

Adds a sibling TypeSpec project for the AKS scheduledEvents Resource Notifications contract.

The project emits a standalone `scheduledEvents.json` for `2026-06-01` with:
- `paths: {}`
- `ScheduledEvent` and `ScheduledEventProperties` definitions only
- no customer-callable AKS RP operations

## Context

This carries forward the existing `Microsoft.ContainerService/managedClusters/scheduledEvents` data model from `2023-11-02-preview` so AKS can update the ARN/Event Grid/ARG publisher contract without changing the event payload under `resourceInfo.properties` / ARG `properties`.

The corresponding AKS API proposal was approved and merged here:
https://github.com/azure-management-and-platforms/aks-handbook/pull/208

## Validation

Local validation against this branch:
- `npx tsp compile specification/containerservice/resource-manager/Microsoft.ContainerService/aks-scheduledevents --emit @azure-tools/typespec-autorest`
- generated only `specification/containerservice/resource-manager/Microsoft.ContainerService/aks-scheduledevents/stable/2026-06-01/scheduledEvents.json`
- generated `paths: {}`
- generated definitions are `ScheduledEvent` and `ScheduledEventProperties`
- `ScheduledEvent` exactly matches the existing `2023-11-02-preview` schema
- `ScheduledEventProperties` matches the existing schema except for one generated non-functional `resources.items.description` from `Azure.Core.armResourceIdentifier`

PR #45114, opened 2026-07-29
Add scheduledEvents schema for 2026-06-01

jdbencardinop:
## Summary

This is the main-targeting ARM review PR requested in https://github.com/Azure/azure-rest-api-specs/pull/44951#pullrequestreview-4783055202.

It adds a sibling TypeSpec project for the AKS `managedClusters/scheduledEvents` Azure Resource Notifications / ARG contract. The project emits a standalone `scheduledEvents.json` for `2026-06-01` with:

- `paths: {}` because this is a schema-only notification contract, not a customer-callable ARM API surface.
- `ScheduledEvent` / `ScheduledEventProperties` matching the existing `2023-11-02-preview` contract shape.
- No SDK emitters configured, because no SDK should be generated for this schema-only contract.

The release-branch PR remains open at #44951 for the `dev-containerservice-Microsoft.ContainerService-2026-06` branch.

Related API proposal approval: https://github.com/azure-management-and-platforms/aks-handbook/pull/208

## Validation

- `npx tsv specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents`
- Local `lint-diff` against `main` completed with no must-fix rows.

### DIFF
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
@@ -1,0 +1,119 @@
+import "@typespec/rest";
+import "@typespec/versioning";
+import "@azure-tools/typespec-azure-core";
+import "@azure-tools/typespec-azure-resource-manager";
+import "@typespec/openapi";
+
+using Azure.ResourceManager;
+using TypeSpec.Versioning;
+using TypeSpec.OpenAPI;
+
+/**
+ * The Container Service Scheduled Events data model.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "This project intentionally emits a schema-only proxy resource contract for Azure Resource Notifications with no customer-callable ARM operations."
+@armProviderNamespace
+@service(#{ title: "ContainerServiceClient" })
+@versioned(Versions)
+@armCommonTypesVersion(Azure.ResourceManager.CommonTypes.Versions.v6)
+namespace Microsoft.ContainerService;
+
+/**
+ * The available API versions.
+ */
+enum Versions {
+  /**
+   * The 2026-06-01 API version.
+   */
+  v2026_06_01: "2026-06-01",
+}
+
+/**
+ * Scheduled event.
+ */
+model ScheduledEvent
+  is Azure.ResourceManager.ProxyResource<ScheduledEventProperties> {
+  ...ResourceNameParameter<
+    Resource = ScheduledEvent,
+    KeyName = "scheduledEventName",
+    SegmentName = "scheduledEvents",
+    NamePattern = "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+  >;
+}
+
+@@doc(ScheduledEvent.properties, "Properties for the event.");
+
+/**
+ * Properties for a scheduled event.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Preserves the existing scheduledEvents notification contract; this materialized notification data model does not include provisioningState."
+model ScheduledEventProperties {
+  /**
+   * The description of the event.
+   */
+  description?: string;
+
+  /**
+   * The event id of the event.
+   */
+  eventId?: string;
+
+  /**
+   * The source of the event.
+   */
+  eventSource?: string;
+
+  /**
+   * The status of the event.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserves the existing scheduledEvents swagger x-ms-enum metadata."
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Keeps the enum inline to preserve the existing scheduledEvents swagger shape."
+  @extension("x-ms-enum", #{ name: "eventStatus", modelAsString: true })
+  eventStatus?:
+    | "Scheduled"
+    | "Started"
+    | "Completed"
+    | "Cancelled"
+    | "Failed"
+    | string;
+
+  /**
+   * The details of the event.
+   */
+  eventDetails?: string;
+
+  /**
+   * The time of the event is scheduled to start.
+   */
+  scheduledTime?: utcDateTime;
+
+  /**
+   * The time the event actually started.
+   */
+  startTime?: utcDateTime;
+
+  /**
+   * The last time the state of the event was updated.
+   */
+  lastUpdateTime?: utcDateTime;
+
+  /**
+   * The list of resources impacted by the event.
+   */
+  resources?: Azure.Core.armResourceIdentifier<[
+    {
+      type: "Microsoft.ContainerService/managedClusters";
+    },
+    {
+      type: "Microsoft.ContainerService/managedClusters/agentPools";
+    }
+  ]>[];
+
+  /**
+   * The resource type of the event.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserves the existing scheduledEvents swagger x-ms-enum metadata."
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Keeps the enum inline to preserve the existing scheduledEvents swagger shape."
+  @extension("x-ms-enum", #{ name: "ResourceType", modelAsString: true })
+  resourceType?: "ManagedCluster" | "AgentPool" | string;
+}
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
@@ -2,11 +2,9 @@ import "@typespec/rest";
 import "@typespec/versioning";
 import "@azure-tools/typespec-azure-core";
 import "@azure-tools/typespec-azure-resource-manager";
-import "@typespec/openapi";
 
 using Azure.ResourceManager;
 using TypeSpec.Versioning;
-using TypeSpec.OpenAPI;
 
 /**
  * The Container Service Scheduled Events data model.
@@ -44,6 +42,48 @@ model ScheduledEvent
 @@doc(ScheduledEvent.properties, "Properties for the event.");
 
 /**
+ * The status of the event.
+ */
+union EventStatus {
+  string,
+  /**
+   * The event is scheduled.
+   */
+  Scheduled: "Scheduled",
+  /**
+   * The event has started.
+   */
+  Started: "Started",
+  /**
+   * The event completed.
+   */
+  Completed: "Completed",
+  /**
+   * The event was cancelled.
+   */
+  Cancelled: "Cancelled",
+  /**
+   * The event failed.
+   */
+  Failed: "Failed",
+}
+
+/**
+ * The resource type of the event.
+ */
+union ResourceType {
+  string,
+  /**
+   * The event applies to a managed cluster.
+   */
+  ManagedCluster: "ManagedCluster",
+  /**
+   * The event applies to an agent pool.
+   */
+  AgentPool: "AgentPool",
+}
+
+/**
  * Properties for a scheduled event.
  */
 #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Preserves the existing scheduledEvents notification contract; this materialized notification data model does not include provisioningState."
@@ -66,16 +106,7 @@ model ScheduledEventProperties {
   /**
    * The status of the event.
    */
-  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserves the existing scheduledEvents swagger x-ms-enum metadata."
-  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Keeps the enum inline to preserve the existing scheduledEvents swagger shape."
-  @extension("x-ms-enum", #{ name: "eventStatus", modelAsString: true })
-  eventStatus?:
-    | "Scheduled"
-    | "Started"
-    | "Completed"
-    | "Cancelled"
-    | "Failed"
-    | string;
+  eventStatus?: EventStatus;
 
   /**
    * The details of the event.
@@ -83,7 +114,7 @@ model ScheduledEventProperties {
   eventDetails?: string;
 
   /**
-   * The time of the event is scheduled to start.
+   * The time the event is scheduled to start.
    */
   scheduledTime?: utcDateTime;
 
@@ -112,8 +143,5 @@ model ScheduledEventProperties {
   /**
    * The resource type of the event.
    */
-  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserves the existing scheduledEvents swagger x-ms-enum metadata."
-  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Keeps the enum inline to preserve the existing scheduledEvents swagger shape."
-  @extension("x-ms-enum", #{ name: "ResourceType", modelAsString: true })
-  resourceType?: "ManagedCluster" | "AgentPool" | string;
+  resourceType?: ResourceType;
 }
--- a/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
+++ b/specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
@@ -9,7 +9,7 @@ using TypeSpec.Versioning;
 /**
  * The Container Service Scheduled Events data model.
  */
-#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "This project intentionally emits a schema-only proxy resource contract for Azure Resource Notifications with no customer-callable ARM operations."
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "This sibling project intentionally emits the schema-only Azure Resource Notifications / Azure Resource Graph / Event Grid contract for Microsoft.ContainerService/managedClusters/scheduledEvents. The contract is consumed as notification and resource graph payload schema and has no customer-callable ARM operations."
 @armProviderNamespace
 @service(#{ title: "ContainerServiceClient" })
 @versioned(Versions)
@@ -27,7 +27,7 @@ enum Versions {
 }
 
 /**
- * Scheduled event.
+ * Scheduled event for a managed cluster. ARM resource type: Microsoft.ContainerService/managedClusters/scheduledEvents.
  */
 model ScheduledEvent
   is Azure.ResourceManager.ProxyResource<ScheduledEventProperties> {
@@ -39,7 +39,7 @@ model ScheduledEvent
   >;
 }
 
-@@doc(ScheduledEvent.properties, "Properties for the event.");
+@@doc(ScheduledEvent.properties, "Properties for the scheduled event.");
 
 /**
  * The status of the event.
@@ -74,9 +74,9 @@ union EventStatus {
 }
 
 /**
- * The resource type of the event.
+ * The kind of resource impacted by the scheduled event.
  */
-union ResourceType {
+union ScheduledEventResourceType {
   string,
 
   /**
@@ -92,31 +92,33 @@ union ResourceType {
 
 /**
  * Properties for a scheduled event.
+ *
+ * Timestamp property names intentionally preserve the existing wire contract; all timestamp values are UTC date-time values.
  */
-#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Preserves the existing scheduledEvents notification contract; this materialized notification data model does not include provisioningState."
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "This type is a read-only notification payload with no ARM create or update operations, so it has no provisioning lifecycle and no provisioningState to report."
 model ScheduledEventProperties {
   /**
-   * The description of the event.
+   * Human-readable summary of the scheduled event.
    */
   description?: string;
 
   /**
-   * The event id of the event.
+   * Unique ID of the scheduled event, represented as a GUID string.
    */
   eventId?: string;
 
   /**
-   * The source of the event.
+   * AKS component or workflow that raised the scheduled event.
    */
   eventSource?: string;
 
   /**
-   * The status of the event.
+   * Current lifecycle state of the scheduled event.
    */
   eventStatus?: EventStatus;
 
   /**
-   * The details of the event.
+   * Free-form human-readable details about why the event was raised or what it represents.
    */
   eventDetails?: string;
 
@@ -148,7 +150,7 @@ model ScheduledEventProperties {
   ]>[];
 
   /**
-   * The resource type of the event.
+   * Whether the event applies to the managed cluster itself or to one of its agent pools.
    */
-  resourceType?: ResourceType;
+  resourceType?: ScheduledEventResourceType;
 }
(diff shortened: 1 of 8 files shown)


specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,119 @@
...
+   * The source of the event.
+   */
+  eventSource?: string;
+
+  /**
+   * The status of the event.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserves the existing scheduledEvents swagger x-ms-enum metadata."
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-union" "Keeps the enum inline to preserve the existing scheduledEvents swagger shape."
+  @extension("x-ms-enum", #{ name: "eventStatus", modelAsString: true })
```
razvanbadea-msft: **[NEW] 🟠 Warning** **[[`@extension` in TypeSpec source](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `main.tsp` — lines 71 and 117 (suppressions on lines 69-70 and 115-116)

**Classification reasoning:** Introduced in this PR — `aksscheduledevents` is a brand-new TypeSpec project with no previous version.

**Issue:** Both enums are modeled as inline unnamed unions with a hand-written `@extension("x-ms-enum", ...)` decorator. The review rules state: _"never add `@extension(...)` decorators to TypeSpec source for any reason … Use the appropriate built-in decorator instead."_ This approach forces four suppressions (`@azure-tools/typespec-azure-core/no-openapi` ×2 and `@azure-tools/typespec-azure-core/no-unnamed-union` ×2) and produces inconsistent `x-ms-enum` name casing — `eventStatus` (camelCase) vs `ResourceType` (PascalCase).

**The tradeoff to decide:** the suppression reasons on lines 69/115 say _"Preserves the existing scheduledEvents swagger x-ms-enum metadata"_, signaling the lowercase `eventStatus` wire-name may be an intentional ARN/ARG compatibility requirement. Two legitimate paths:

- **Option A (recommended if no hard wire-name constraint):** Replace each inline union with a named extensible `union`. This auto-emits `x-ms-enum` with `modelAsString: true`, removes all four suppressions, and fixes the casing:

```tsp
/** The status of the event. */
union EventStatus { string, Scheduled: "Scheduled", Started: "Started", Completed: "Completed", Cancelled: "Cancelled", Failed: "Failed" }
// ...
eventStatus?: EventStatus;
```

  (and `union ResourceType { string, ManagedCluster: "ManagedCluster", AgentPool: "AgentPool" }`). Note this changes the emitted `x-ms-enum.name` from `eventStatus` → `EventStatus`.
- **Option B (if lowercase wire-name preservation is a hard ARN/ARG requirement):** keep the current shape, but state that constraint explicitly in the suppression reasons.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,119 @@
...
+    | "Failed"
+    | string;
+
+  /**
+   * The details of the event.
+   */
+  eventDetails?: string;
+
+  /**
+   * The time of the event is scheduled to start.
```
razvanbadea-msft: **[NEW] 🔵 Suggestion** `main.tsp` — line 86 (emitted `scheduledTime.description`)

**Classification reasoning:** Introduced in this PR — new property in a brand-new project.

**Issue:** Grammar: _"The time of the event is scheduled to start."_

**Suggested fix:** _"The time the event is scheduled to start."_

razvanbadea-msft (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`13b451a`](https://github.com/Azure/azure-rest-api-specs/pull/45114/commits/13b451a29c2df56ece23c3ff207bc097dcf77e23). See inline comments for findings 1-2.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/stable/2026-06-01/scheduledEvents.json
```diff
@@ -0,0 +1,191 @@
...
+      "description": "Azure Active Directory OAuth2 Flow.",
+      "flow": "implicit",
+      "authorizationUrl": "https://login.microsoftonline.com/common/oauth2/authorize",
+      "scopes": {
+        "user_impersonation": "impersonate your user account"
+      }
+    }
+  },
+  "tags": [],
+  "paths": {},
```
ravimeda, line 42: **[NEW] 🔴 Blocking** **[[RPC §2.7](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#27-every-resource-must-have-a-point-get)] / [[RPC-Operations-V1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#14-operations-api-rpc-operations-v1)] / [[RPC-Operations-V1-TAG](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#10a3-operations-api-must-be-in-package-tag-rpc-operations-v1-tag)] / [[RPC-SUPPRESS-GA](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#ga-vs-preview-rigor-rpc-suppress-ga)] / [[SCHEMA-UNUSED-DEF](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#23-unused--orphaned-definitions-schema-unused-def)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/stable/2026-06-01/scheduledEvents.json` — line 42 / JSON path `$.paths`

**Classification reasoning:** Introduced in this PR — `aksscheduledevents/` returns 404 at the base SHA and all eight files are newly added.

**Issue:** This PR ships a **stable (GA)** API version that declares an ARM resource type but exposes **zero operations**.

- `"paths": {}` (line 42) means the `ScheduledEvent` proxy resource defined at line 110 is never reachable, and it is never referenced by any `$ref` in the file. It survives emission only because `tspconfig.yaml` line 9 sets `omit-unreachable-types: false`, which explicitly disables the emitter behaviour that would otherwise drop it — so the orphaned definition (SCHEMA-UNUSED-DEF) is a deliberate configuration choice, not an accident.
- RPC §2.7: _"Every resource type **MUST** have a corresponding point GET operation that retrieves a single instance by its resource path."_ There is none.
- RPC-Operations-V1 §1.4: _"Every resource provider **MUST** expose an operations API …"_. RPC-Operations-V1-TAG §10A.3 is an exact match for what this PR does: _"For ARM RPs, the `operations.json` (or equivalent operations API spec) **MUST** be included in every published package tag's input-file list. If a new tag only includes the service resource swagger but omits the operations API, flag it — the resulting SDK will be missing the operations endpoint."_ The new tag `package-2026-06` (`readme.md` lines 20-23) lists exactly one input file, `stable/2026-06-01/scheduledEvents.json`, and no operations spec.
- Three suppressions exist to silence the consequences: `#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint"` (`main.tsp` line 12), `suppressions: - code: OperationsAPIImplementation` (`readme.md` lines 10-13), and `SdkTspConfigValidation` for `tspconfig.yaml` (`suppressions.yaml` lines 1-5). The third stands on its merits — `SdkTspConfigValidation`'s required sub-rules mandate SDK emitter configuration a schema-only project genuinely cannot supply, and the stated reason is accurate — but its `reason:` uses the unexpanded internal abbreviation "ARN/ARG"; please spell out "Azure Resource Notifications / Azure Resource Graph". The first two are the subject of this finding.

Per RPC-SUPPRESS-GA, suppressions carried forward from preview are _"**not automatically acceptable** in GA (stable) versions"_ and each must be individually challenged. The criteria direct reviewers to push for a fix when _"the same anti-pattern exists in **earlier preview versions** but has **not yet leaked to a stable version**. Now is the time to fix it before it becomes entrenched."_ That is exactly this situation: the shape comes from `aks/preview/2023-11-02-preview/scheduledEvents.json` (line 36, also `paths: {}`), a file referenced by **no** AutoRest package tag anywhere in the repository and therefore never linted. This PR is the first time the pattern enters a `stable/` directory **and** the first time it enters a package tag. The same criteria also allow escalation to peer review where _"the violation is for a **new resource type or API** and the author provides a **reasonable justification** for why the rule does not apply"_ — you have supplied one, in the PR description, the suppression reasons, and a linked API-proposal approval, so fix option (b) below is that escalation path rather than a rejection of the design. What is not acceptable is shipping to `stable/` without the sign-off recorded, because once published the shape is effectively permanent.

**Fix:** Choose one and record the decision in the suppression `reason:` text:

- **(a)** Model the type as a real ARM resource — add a point GET and a list-under-parent operation so `ScheduledEvent` is reachable, which removes the need for both suppressions.
- **(b)** Keep it schema-only, but obtain **explicit ARM API review board sign-off for a GA schema-only contract** and record it in both suppression reasons (approver, date, approval link). Also consider following the `2023-11-02-preview` precedent of **not** adding the file to an AutoRest package tag — that is why the preview file never needed an `OperationsAPIImplementation` suppression, and it would also make §10A.3 inapplicable.
- **(c)** If GA sign-off is not yet in hand, publish this contract as a `preview` version first and promote it to `stable/` once the board has signed off.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,154 @@
...
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Preserves the existing scheduledEvents notification contract; this materialized notification data model does not include provisioningState."
+model ScheduledEventProperties {
+  /**
+   * The description of the event.
+   */
+  description?: string;
+
+  /**
+   * The event id of the event.
```
ravimeda: **[NEW] 🟠 Warning** **[[OAPI §17](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#17-descriptions--documentation-quality)] / [[TSP §2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp` — lines 98-121 and 150-153

**Classification reasoning:** Introduced in this PR — new project, no previous version.

**Issue:** Six of the ten property doc comments in `ScheduledEventProperties` restate the property name and add no information, which OAPI §17 forbids: _"Property descriptions **MUST NOT** simply repeat the property name"_. The instances are `description` — _"The description of the event."_ (line 99), `eventId` — _"The event id of the event."_ (line 104), `eventSource` — _"The source of the event."_ (line 109), `eventStatus` — _"The status of the event."_ (line 114), `eventDetails` — _"The details of the event."_ (line 119), and `resourceType` — _"The resource type of the event."_ (line 151). These are published verbatim to `learn.microsoft.com` and into every downstream consumer of the emitted swagger (`scheduledEvents.json` lines 129-186).

`eventId` (line 104) is the most consequential instance and has two additional defects: it lowercases the acronym ("id" rather than "ID"), which §17 also calls out (_"Use correct acronym capitalization in descriptions: \"URL\" not \"Url\", \"ID\" not \"Id\" …"_), and it documents neither the identifier's format nor its uniqueness scope even though the resource name for the same entity is constrained to a GUID by the `NamePattern` on line 38. TSP §2.2 requires that `<something>Id` properties _"be specific about what kind of ID they hold (`armResourceIdentifier`, or a documented format)"_. The wording was carried over verbatim from `aks/preview/2023-11-02-preview/scheduledEvents.json` line 62, and this is the last version in which it can be changed for free.

**Fix:** Rewrite each doc comment to say what the field actually holds. For example: `eventId` → _"The unique ID of the event, a GUID that matches the scheduled event resource name."_; `eventSource` → _"The component that raised the event, for example the AKS node image upgrade pipeline."_; `eventDetails` → _"Free-form human-readable detail about why the event was raised."_; `resourceType` → _"Whether the event applies to the managed cluster itself or to one of its agent pools."_

For `eventId`, **do not** change the type to the TypeSpec `uuid` scalar — per the [TSP §2.2 ARM caveat](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement) and [guid-and-uuid-on-arm.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/guid-and-uuid-on-arm.md), emitting `"format": "uuid"` on an ARM control-plane spec triggers the required `GuidUsage` LintDiff rule (R3017) and would turn a currently-green LintDiff run red. Keep `string` and convey the format in the description, optionally with a `@pattern`.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Scheduled event.
+ */
+model ScheduledEvent
+  is Azure.ResourceManager.ProxyResource<ScheduledEventProperties> {
+  ...ResourceNameParameter<
+    Resource = ScheduledEvent,
+    KeyName = "scheduledEventName",
+    SegmentName = "scheduledEvents",
```
ravimeda, line 37: **[NEW] 🔵 Suggestion** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp` — lines 29-40

**Classification reasoning:** Introduced in this PR — new project, no previous version.

**Issue:** The PR description states this is the contract for the AKS `managedClusters/scheduledEvents` resource type, but that type appears nowhere in the shipped artifacts. `ScheduledEvent` is declared as a bare `ProxyResource` with `SegmentName = "scheduledEvents"` (line 37) and no `@parentResource` decorator, and the emitted OpenAPI has no paths (`scheduledEvents.json` line 42), so neither the TypeSpec source nor the generated swagger states the parent. A downstream consumer — Azure Resource Notifications payload binding, Azure Resource Graph schema onboarding, documentation — cannot tell whether the type is `Microsoft.ContainerService/scheduledEvents` or `Microsoft.ContainerService/managedClusters/scheduledEvents`, and the GUID `NamePattern` on line 38 gives no indication of which scope the name is unique within. This is a documentation-clarity observation rather than a rule violation; if the resource is genuinely a child type, the operations question is covered by the Blocking finding on `scheduledEvents.json` line 42 instead.

**Fix:** State the fully qualified ARM resource type in the `ScheduledEvent` doc comment (lines 29-31) and in `readme.md`, e.g. `/** Scheduled event for a managed cluster. ARM resource type: Microsoft.ContainerService/managedClusters/scheduledEvents. */`. If a cross-project reference to the AKS `ManagedCluster` model is feasible, adding `@parentResource` would make the hierarchy machine-readable rather than prose.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * The event failed.
+   */
+  Failed: "Failed",
+}
+
+/**
+ * The resource type of the event.
+ */
+union ResourceType {
```
ravimeda: **[NEW] 🔵 Suggestion** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp` — lines 76-91 (emitted as definition `ResourceType` at `scheduledEvents.json` lines 86-109)

**Classification reasoning:** Introduced in this PR — new project, no previous version.

**Issue:** `ResourceType` is a very generic name for a definition in an ARM specification, and it is semantically misleading: in ARM, a "resource type" is a string such as `Microsoft.ContainerService/managedClusters`, whereas this union's members are short kind names (`ManagedCluster`, `AgentPool`). The sibling `resources` property (lines 141-148) already carries the real ARM resource type strings, so the same file uses "resource type" to mean two different things. No rule in the instruction files prohibits a generic definition name; this is a design-clarity suggestion.

**Fix:** Rename to something specific, e.g. `ScheduledEventResourceType` or `ImpactedResourceKind`. This also changes the emitted `x-ms-enum.name` (`scheduledEvents.json` line 94), which is safe here because this project generates no SDK — but it becomes permanent once `stable/2026-06-01` ships.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * The event applies to an agent pool.
+   */
+  AgentPool: "AgentPool",
+}
+
+/**
+ * Properties for a scheduled event.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Preserves the existing scheduledEvents notification contract; this materialized notification data model does not include provisioningState."
```
ravimeda: **[NEW] 🔵 Suggestion** **[[TSP §4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)] / [[Suppressions That Are Never Valid](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppressions-that-are-never-valid)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp` — line 96

**Classification reasoning:** Introduced in this PR — new project, no previous version.

**Issue:** The `arm-resource-provisioning-state` suppression reason leads with a back-compatibility argument — _"Preserves the existing scheduledEvents notification contract"_ — before its technical claim. The criteria state that _"**\"Back-compat with preview\"** is not sufficient justification for a GA (stable) version. Preview was the time to let it slide; GA is the deadline to fix."_ TSP §4.1 requires each suppression to _"stand on its own with a technical justification for **why** the rule does not apply to this specific case."_ The technical justification here is real and sufficient on its own, so only the framing needs to change. Separately, `tspconfig.yaml` line 10 sets `use-read-only-status-schema: true` and line 15 sets `emit-lro-options: "all"` — both are inert on a model with no `provisioningState` and no operations, and the former reads as contradictory next to this suppression.

**Fix:** Reorder the reason to lead with the technical justification and drop the back-compat framing, e.g. `"This type is a read-only notification payload with no ARM create/update operations, so it has no provisioning lifecycle and no provisioningState to report."` Optionally remove the two inert emitter options, or keep them as intentional forward-configuration and say so.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * The details of the event.
+   */
+  eventDetails?: string;
+
+  /**
+   * The time the event is scheduled to start.
+   */
+  scheduledTime?: utcDateTime;
```
ravimeda, line 128: **[NEW] 🔵 Suggestion** **[[TSP §2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp` — lines 126, 131, 136

**Classification reasoning:** Introduced in this PR — new project, no previous version.

**Issue:** TSP §2.2 states that _"Properties representing UTC timestamps **SHOULD** include a `Utc` suffix in the name (e.g., `lastModifiedTimeUtc`)."_ The three `utcDateTime` properties — `scheduledTime` (line 126), `startTime` (line 131), `lastUpdateTime` (line 136) — do not carry the suffix. This is raised for an explicit decision rather than as a change request, and there are two solid arguments for keeping the current names: they mirror the wire contract this PR deliberately reproduces from `aks/preview/2023-11-02-preview/scheduledEvents.json`, so renaming would diverge the GA schema from the notification payload existing consumers already parse; and the closely related [OAPI §17 unit-encoding guidance](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#17-descriptions--documentation-quality) conditions its `SHOULD` on cases _"when the type alone does not make it clear"_ and notes it is _"especially important when **not** using `format: date-time`"_ — all three properties do emit `format: date-time` (`scheduledEvents.json` lines 151, 156, 161), so the type already conveys the unit. The reason it is still worth deciding now is the same one behind the Blocking finding — this is the last version before the names become permanent in `stable/`.

**Fix:** Either rename to `scheduledTimeUtc` / `startTimeUtc` / `lastUpdateTimeUtc`, or record the deliberate decision to keep the existing names for wire-contract compatibility (a one-line note in the `ScheduledEventProperties` doc comment, lines 93-95, is sufficient).

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`ce318ce`](https://github.com/Azure/azure-rest-api-specs/pull/45114/commits/ce318ce259b5260d81b87f7ec7d47eacc3149193). See inline comments for findings 1-6.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
ravimeda: Thanks for addressing this! The violation flagged here is no longer present in the latest changes. Resolving this thread.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/main.tsp
ravimeda: Thanks for addressing this! The violation flagged here is no longer present in the latest changes. Resolving this thread.

specification/containerservice/resource-manager/Microsoft.ContainerService/aksscheduledevents/stable/2026-06-01/scheduledEvents.json
jdbencardinop: Thanks Ravi. We would like to follow option **(b)** if that is acceptable: keep this as a schema-only contract and get explicit ARM signoff for the GA schema-only shape.

Context: this contract is intentionally not a customer-callable AKS RP operation surface. It is the schema consumed by Azure Resource Notifications / Azure Resource Graph / Event Grid for `Microsoft.ContainerService/managedClusters/scheduledEvents`. Adding GET/LIST operations would expose an ARM REST surface that the proposal intentionally avoided. The AKS API proposal was approved here: https://github.com/azure-management-and-platforms/aks-handbook/pull/208

Could you help confirm the expected process for option **(b)**?

1. Is an ARM reviewer approval / `ARMSignedOff` on this PR enough if we update the suppression reasons with the reviewer, date, and this PR discussion link?
2. Or does “explicit ARM API review board sign-off” require taking this to a separate ARM API Review Board meeting/channel/issue first? If yes, do you know the right place/process to start that?
3. If GA schema-only signoff is not practical from this PR, would you recommend temporarily moving this to `preview/2026-06-01-preview` and then promoting to `stable/2026-06-01` after the board signs off, even though the approved proposal currently targets stable `2026-06-01`?

Once we know the signoff path, I can also address the cleanup suggestions in the same update: improve property descriptions, document the full resource type, rename the generic `ResourceType`, reframe suppression reasons with technical justification, spell out Azure Resource Notifications / Azure Resource Graph, and document the timestamp wire-name decision.

PR #45124, opened 2026-07-29
Shorten TypeSpec suppressions source line column and clarify testing status

catalinaperalta:
## Purpose

Improves the readability of the **TypeSpec Suppressions Review** PR comment and clarifies its testing status. Rendering lives in `.github/workflows/src/typespec-suppressions/suppressions-comment.js`.

## Changes

**1. Shorten the Source column**
- The **Source** link now shows only the **file name + line** (e.g. `main.tsp#L37`) instead of the full spec path. The complete path is preserved in the link's `href`, so click-through is unchanged — only the visible label is shortened, keeping the column narrow and readable.

**2. Clarify testing / non-blocking status**
- The comment message now states up front that the check is **in testing mode and non-blocking** (it will not prevent the PR from merging), while retaining the reviewer guidance to approve only with a clear justification.
- The status summary reads `❌ Approval required (currently under testing, review NOT enforced)` while pending, and `✅ Approved` once the `Approved-TypeSpecSuppression` label is applied. The ✅/❌ Status column indicator is retained.

## Validation
- Verified end-to-end on a draft test PR (fork) where the base branch carries this code, so the `pull_request_target` comment job renders the updated output: the Source column shows `main.tsp#L19` and the testing/non-blocking message appears.
- Updated assertions in `suppressions-comment.test.js`; `npm run test -- typespec-suppressions` passes and prettier is clean.

### DIFF
--- a/.github/workflows/src/typespec-suppressions/suppressions-comment.js
+++ b/.github/workflows/src/typespec-suppressions/suppressions-comment.js
@@ -186,6 +186,20 @@ function escapeHtml(value) {
 }
 
 /**
+ * Inserts `<wbr>` break opportunities after path separators so long, unbroken
+ * strings (rule names and source paths, which contain no spaces) can wrap
+ * within their table cell instead of forcing the column wide. GitHub's HTML
+ * sanitizer allows `<wbr>`, and it renders as a zero-width break hint. The input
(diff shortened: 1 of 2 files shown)


catalinaperalta:
Example of updated table rendering: 
image

timotheeguerin (approved):

PR #45130, opened 2026-07-30
Add 2026-07-30-preview with ephemeral type compatibility check for Managed DevOps Pools

nicholasbergesen:
## Summary

Cuts the `2026-07-30-preview` api-version for `Microsoft.DevOpsInfrastructure` and adds
`EphemeralTypeCompatibility_Check` to it.

```
POST /subscriptions/{subscriptionId}/providers/Microsoft.DevOpsInfrastructure/locations/{location}/checkEphemeralTypeCompatibility
```

Managed DevOps Pools already accepts a per-image `ephemeralType`, but a caller has no way to
discover which placements a given image version can actually use on a given agent size. Today the
only way to find out is to save the pool and see what provisioning decides. This lets the Azure
Portal offer only the placements that will actually be honoured, without duplicating the service's
image-size and SKU capability rules and drifting from provisioning behaviour.

## Versioning

This originally added the operation to `2026-07-03-preview`. The versioning check correctly flagged
that: `2026-07-03-preview` is already merged to `main`, so it is considered shipped regardless of
whether any resource provider serves it yet. The change now cuts a new `2026-07-30-preview` instead,
and `2026-07-03-preview` is restored byte for byte to its state on `main`.

`2026-07-30-preview` inherits every existing operation plus `NVMeDisk` on the `EphemeralType` enum
(added in #44406), and adds the one new operation. Everything new is gated with
`@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_30_preview)`. Generated with
`tsp compile`; no previously released version is modified.

## Design notes

**Subscription and location scoped, not pool scoped.** The pool create experience needs the answer
before a pool resource exists.

**The request mirrors the SKU shape a pool already uses** (`name`, plus `vmSizes` for an instance
mix), so a caller can pass the SKU it is about to save without reshaping it. For an instance mix, an
explicit placement is only reported as available when every VM size supports it.

**Images are identified by resource id**, since MDP pools reference images that way. A version of
`latest` is resolved by the service.

**Partial failure is a success.** A syntactically valid request returns 200 even when individual
image versions fail to evaluate. Each result carries either `availableEphemeralTypes` or `error`,
never both, so one unreadable image does not discard the results for the others. Non-2xx is reserved
for failures that prevent evaluating the request as a whole, such as a malformed body or an agent
size that is not available in the location. The example shows both outcomes in one response.

## Consuming change

The resource provider implementation is a separate change in the internal ResourceManagement repo
and will not ship ahead of this spec, since ProviderHub routes `Microsoft.DevOpsInfrastructure`
against the published folder here.

### DIFF
--- a/specification/devopsinfrastructure/resource-manager/Microsoft.DevOpsInfrastructure/DevOpsInfrastructure/main.tsp
+++ b/specification/devopsinfrastructure/resource-manager/Microsoft.DevOpsInfrastructure/DevOpsInfrastructure/main.tsp
@@ -39,6 +39,10 @@ enum Versions {
   /** 2026-07-03-preview version */
   @armCommonTypesVersion(CommonTypes.Versions.v6)
   v2026_07_03_preview: "2026-07-03-preview",
+
+  /** 2026-07-30-preview version */
+  @armCommonTypesVersion(CommonTypes.Versions.v6)
+  v2026_07_30_preview: "2026-07-30-preview",
 }
 
 model Pool is TrackedResource<PoolProperties> {
@@ -868,7 +872,7 @@ model QuotaName {
 }
 
 @doc("The agent size or sizes an ephemeral type compatibility check is evaluated against. This mirrors the SKU shape a pool already uses, so a caller can pass the SKU it is about to save without reshaping it.")
-@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_03_preview)
+@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_30_preview)
 model EphemeralTypeCompatibilitySku {
   @doc("The Azure SKU name of the machines in the pool. Use \"Mix\" together with vmSizes for an instance mix pool.")
   name: string;
@@ -879,7 +883,7 @@ model EphemeralTypeCompatibilitySku {
 }
 
 @doc("Asks which ephemeral OS disk placements can be used for a set of image versions on a set of agent sizes.")
-@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_03_preview)
+@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_30_preview)
 model EphemeralTypeCompatibilityRequest {
   @doc("The agent size or sizes to evaluate against.")
   sku: EphemeralTypeCompatibilitySku;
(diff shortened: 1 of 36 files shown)


specification/devopsinfrastructure/resource-manager/Microsoft.DevOpsInfrastructure/DevOpsInfrastructure/main.tsp
```diff
@@ -867,6 +871,69 @@ model QuotaName {
...
+  @minItems(1)
+  @maxItems(50)
+  imageVersionResourceIds: string[];
+}
+
+@doc("The compatibility result for a single requested image version. This contains either availableEphemeralTypes or error, never both, so that one image version failing to evaluate does not discard the results for the others.")
+@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_30_preview)
+model EphemeralTypeCompatibilityResult {
+  @doc("The resource id of the image version this result is for, echoed from the request.")
+  imageVersionResourceId: string;
```
razvanbadea-msft, line 901: >imageVersionResourceId [](http://example.com/codeflow?start=2&length=22)

sounds like you can use armResourceIdentifier in this case withj type restriction

specification/devopsinfrastructure/resource-manager/Microsoft.DevOpsInfrastructure/DevOpsInfrastructure/main.tsp
```diff
@@ -867,6 +871,69 @@ model QuotaName {
...
+@doc("Asks which ephemeral OS disk placements can be used for a set of image versions on a set of agent sizes.")
+@added(Microsoft.DevOpsInfrastructure.Versions.v2026_07_30_preview)
+model EphemeralTypeCompatibilityRequest {
+  @doc("The agent size or sizes to evaluate against.")
+  sku: EphemeralTypeCompatibilitySku;
+
+  @doc("The resource ids of the image versions to evaluate. A version of 'latest' is resolved by the service.")
+  @minItems(1)
+  @maxItems(50)
+  imageVersionResourceIds: string[];
```
razvanbadea-msft, line 894: >imageVersionResourceIds [](http://example.com/codeflow?start=2&length=23)

sounds like you can use armResourceIdentifier in this case with type restriction

PR #45138, opened 2026-07-30
resources/deployments: change C# emitter namespace to Azure.ResourceManager.Resources.Deployments

ArthurMa1978:
## Overview

Updates `specification/resources/resource-manager/Microsoft.Resources/deployments/tspconfig.yaml` so the `@azure-typespec/http-client-csharp-mgmt` emitter's `namespace` option is `Azure.ResourceManager.Resources.Deployments` (matches `package-name`) instead of `Azure.ResourceManager.Resources`.

This keeps the emitted SDK's root namespace aligned with the package name and prevents type collisions with the sibling `Azure.ResourceManager.Resources` package (both packages previously exposed types like `AzureResourceManagerResourcesContext` in `Azure.ResourceManager.Resources.Models`).

## Related

- SDK PR: https://github.com/Azure/azure-sdk-for-net/pull/61573

## Change scope

- No changes to any `.tsp` files, examples, or API surface — only the C# emitter option in `tspconfig.yaml`.
- Other language emitter options are untouched.


ArthurMa1978:
Superseded by clean PR based on latest main with only the tspconfig.yaml change.

PR #45143, opened 2026-07-30
Add v1 voice-agent WebSocket event contract

yulin-li:
## Summary

- Add the stable `v1` WebSocket upgrade route at `/agents/{agent_name}/endpoint/protocols/voice`.
- Model the customer-facing client/server JSON event contract with direction-specific session shapes.
- Keep internal Voice Live events and headers out of the public contract, including `session.close`, `session.done`, `byom_credential.update`, and `x-ms-voice-session-override`.
- Keep the HTTP `101` upgrade response bodyless.

## Dependency and model reuse

This PR is intentionally stacked on #45185. Control-plane voice-agent resources remain owned by that PR; this diff adds only the WebSocket route, protocol models, OpenAI Realtime shim, generated projections, and alignment notes.

For `session.update`, `session.created`, and `session.updated`, the protocol directly reuses the #45185 models for typed voices, structured audio formats, output timestamps, noise reduction, required transcription configuration, VAD defaults, greeting, toolbox/system tools, MCP scheduling, Realtime function tools, and output modalities. The returned session model composes its common fields from the update model and overrides only server-authoritative state.

WebSocket-only wrappers remain only where the wire contract differs: nullable input overrides, MCP runtime authorization, richer end-of-utterance settings, echo cancellation, avatar transport details, and effective handoff state.

Internal top-level channel/multichannel fields, `{ "type": "none" }` VAD, runtime-only toolbox/system shapes, and resolved `foundry_agent` tools are excluded. Turn detection is disabled with `null`. Avatar types use the control-plane `video_avatar` / `photo_avatar` literals.

The event-model approach follows #44893 and incorporates the API review guidance from #43970.

## Validation

- `npx tsv .` from `specification/ai-foundry/data-plane/Foundry` passes, including compile with warnings as errors, OpenAPI regeneration, and formatting.
- Generated v1 JSON/YAML contains the WebSocket route, exact public parameters, and a bodyless `101` response.
- The repository synthetic `virtual-public-preview` projection is emitted automatically; no `@added(Versions.v1)` annotations were added.
- Verified the stacked diff does not modify `voice-agents/agents.tsp` or `voice-agents/models.tsp`.

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml
+++ b/specification/ai-foundry/data-plane/Foundry/openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml
@@ -3,6 +3,7 @@ info:
   title: Microsoft Foundry
   version: virtual-public-preview
 tags:
+  - name: Voice Agent WebSocket
   - name: Agent Conversations
   - name: Voice Agents
   - name: Voice Agent Versions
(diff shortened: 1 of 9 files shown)


specification/ai-foundry/data-plane/Foundry/openapi3/virtual-public-preview/microsoft-foundry-openapi3.yaml
```diff
@@ -81109,6 +81110,125 @@ components:
...
+        server_label:
+          type: string
+          description: A label for this MCP server, used to identify it in tool calls.
+        server_url:
+          type: string
+          format: uri
+          description: |-
+            The URL for the MCP server. One of `server_url`, `connector_id`, or
+              `tunnel_id` must be provided.
+        connector_id:
```
niuzheng168: we do not support connector_id

PR #41834, opened 2026-03-27
Stable API version for MST 2026-03-26 37284272

siddharth9606:
# Data Plane API Specification Update Pull Request

> [!TIP]
> Overwhelmed by all this guidance? See the `Getting help` section at the bottom of this PR description.

This PR creates a stable version of the MST API based on the 2025-01-31-preview version

## PR review workflow diagram

Please understand this diagram before proceeding. It explains how to get your PR approved & merged.

spec_pr_review_workflow_diagram

## API Info: The Basics

Most of the information about your service should be captured in the issue that serves as your [*API Spec engagement record*](https://aka.ms/azsdk/onboarding/restapischedule).

* Link to API Spec engagement record issue:

Is this review for (select one):

- [ ] a private preview
- [ ] a public preview
- [x] GA release

### Change Scope

<sup>This section will help us focus on the specific parts of your API that are new or have been modified. <br/>Please share a link to the design document for the new APIs, a link to the previous API Spec document (if applicable), and the root paths that have been updated. </sup>
* Design Document:
* Previous API Spec Doc:
* Updated paths:

### Viewing API changes

For convenient view of the API changes made by this PR, refer to the URLs provided in the table 
in the `Generated ApiView` comment added to this PR. You can use ApiView to show API versions diff. 

### Suppressing failures

If one or multiple validation error/warning suppression(s) is detected in your PR, please follow the 
[Swagger-Suppression-Process](https://aka.ms/pr-suppressions) 
to get approval.

### Release planner

A [release plan](https://aka.ms/azsdkdocs/release-plans) should have been created. If not, please create one as it will help guide you through the REST API and SDK creation process. 

## ❔Got questions? Need additional info?? We are here to help!

<details>
  <summary> Contact us!</summary>

The [Azure API Review Board](https://aka.ms/azsdk/onboarding/restapischedule) is dedicated to helping you create amazing APIs. You can read about our mission and learn more about our process on our [wiki](https://aka.ms/azsdk/onboarding/restapischedule).
* 💬 [Teams Channel](https://teams.microsoft.com/l/channel/19%3a3ebb18fded0e47938f998e196a52952f%40thread.tacv2/General?groupId=1a10b50c-e870-4fe0-8483-bf5542a8d2d8&tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47)
* 💌 [email](mailto://azureapirbcore@microsoft.com)

</details>

<details>
  <summary>Click here for links to tools, specs, guidelines & other good stuff</summary>

[...]

PR #45162, opened 2026-07-31
[Confidential Ledger] Add waitForCommit preview API

ryazhang-microsoft:
## Summary

Adds the Confidential Ledger data-plane `2026-07-31-preview` API version for the `waitForCommit` feature implemented in microsoft/TPAL#509.

## Changes

- Adds optional `waitForCommit` to `POST /app/transactions`.
- Documents the inline transaction receipt returned after commit.
- Documents the `503 TransactionRolledBack` response.
- Adds source and generated examples for the new preview version.
- Updates the Confidential Ledger AutoRest tag.

## Validation

- `azsdk_run_typespec_validation`
- `npx tsp compile . --warn-as-error`
- Confirmed prior `2024-12-09-preview` and `2026-02-23` OpenAPI outputs are unchanged.

Related implementation: https://github.com/microsoft/TPAL/pull/509


ryazhang-microsoft:
Requested API Stewardship Board review by email for the new API version and the intentional cross-version change. All code validation checks are passing; awaiting the required policy approvals.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
```diff
@@ -437,6 +437,61 @@ model LedgerWriteResult {
...
+
+  /**
+   * The receipt certifying the committed transaction.
+   */
+  receipt?: ReceiptContents;
+
+  /**
+   * State of the ledger query.
+   */
+  state?: ConfidentialLedgerQueryState;
```
andpiccione: Shouldn't this be the transaction state?

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
```diff
@@ -437,6 +437,61 @@ model LedgerWriteResult {
...
+model LedgerWriteResultBody {
+  /**
+   * The collection identifier of the ledger entry.
+   */
+  collectionId: string;
+
+  /**
+   * The receipt certifying the committed transaction.
+   */
+  receipt?: ReceiptContents;
```
andpiccione, line 443: Do we also need to add optional application claims?

ryazhang-microsoft:
# PR #45162 cross-version justification

Pinned head: `7518f69962a367593b64d323f2ea29889758b84a`  
Implementation: https://github.com/microsoft/TPAL/pull/509  
Cross-version job: https://github.com/Azure/azure-rest-api-specs/actions/runs/32389630948/job/96492348181  
Related precedent: https://github.com/Azure/azure-rest-api-specs/pull/41834

## Intended API change

`2026-07-31-preview` adds optional `waitForCommit` to `POST /app/transactions`.

- Omitted, `false`, or malformed: the new preview preserves the immediate write flow and returns `state: "Pending"`.
- `true`: TPAL waits for global consensus and returns `state: "Committed"`, the receipt, and the scoped application claim inline.
- Rollback: TPAL returns `503 TransactionRolledBack`.
- Earlier API versions retain their existing response shape. An explicit `waitForCommit=true` before `2026-07-31-preview` is rejected with `ApiVersionMismatch`.
- The CCF transaction ID remains in `x-ms-ccf-transaction-id` and is not duplicated in the response body.

## Latest stable comparison: `2026-02-23`

Approval is requested for two intentional blocking findings:

1. `1011 AddingResponseCode`: `503` represents the terminal rollback outcome for the opt-in synchronous commit flow.
2. `1042 ChangedParameterOrder`: Swagger 2 emits the new query parameter before the body parameter, moving `entry` from index 3 to index 4. The HTTP body and wire semantics are unchanged, but positional callers upgrading to a newly generated SDK may need to update call sites or use named arguments.

Other findings are additive or required by the feature:

- `1017 ReferenceRedirection`: the `200` response references the preview write-result body.
- `1041 AddedPropertyInResponse`: optional `receipt`, `state`, and `applicationClaims`.
- `1043 AddingOptionalParameter`: optional `waitForCommit`, default `false`.

## Latest preview comparison: `2024-12-09-preview`

The preview comparison contains the feature changes above plus differences already present in stable `2026-02-23`, including `/app/collections/tags`, role-delete response changes, and PATCH `consumes` differences. Those inherited lineage findings are not new `waitForCommit` regressions.

## Suppressions

The replacement operation has two narrow source suppressions:

- `use-standard-operations`: preserves the existing legacy API route/shape.
- `no-openapi`: permits `@operationId("CreateLedgerEntry")` so the versioned replacement retains the existing generated operation ID instead of creating a larger SDK rename break.

## Compatibility and prevention

- Prior generated `2026-02-23` stable and `2024-12-09-preview` outputs are unchanged.
- The implementation is API-version gated and has dedicated plus full-version regression coverage.
- Existing customers can remain on prior versions and keep the polling flow; only adopters of the new preview and `waitForCommit=true` must consume the inline result and handle the retriable `503` rollback outcome.
- Customer telemetry and rollout details are being supplied privately to the API Stewardship Board.

## Requested approval

Approve intentional findings `1011` and `1042` for `2026-07-31-preview`. The feature is opt-in, prior API versions remain unchanged, and the specification matches the version-gated TPAL implementation.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
ryazhang-microsoft: Added in 5f81e6bf9e. applicationClaims is now optional on LedgerWriteResultBody and the waitForCommit example includes the scoped ledger-entry claim returned by TPAL.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
ryazhang-microsoft: Fixed in 5f81e6bf9e: the write response now uses the existing TransactionState union, so the emitted values are Pending and Committed, matching TPAL.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
```diff
@@ -150,6 +152,49 @@ op createLedgerEntry is ConfidentialLedgerCommon.LedgerOperation<
...
+ */
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+#suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserve the existing operationId"
+@added(Versions.v2026_07_31_preview)
+@operationId("CreateLedgerEntry")
+@summary("Writes a ledger entry.")
+@sharedRoute
+@route("/app/transactions")
+@post
+op createLedgerEntryWithWaitForCommit is ConfidentialLedgerCommon.LedgerOperation<
```
andpiccione: Should we preserve the existing native SDK method name here? `@operationId` preserves the Swagger operation ID, but SDKs generated directly from TypeSpec use the TypeSpec operation name. 

Without an `@@clientName` override, this may rename `createLedgerEntry` in generated SDKs. We should also check if changing the response model from LedgerWriteResult to LedgerWriteResultBody does not rename the public return type.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
```diff
@@ -437,6 +437,61 @@ model LedgerWriteResult {
...
+
+  /**
+   * State of the write transaction.
+   */
+  state?: TransactionState;
```
andpiccione: Should state be required here? A successful response should always contain this field.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
weidongxu-microsoft: `@operationId` should never be used.

Use `@clientName("autorest")` if you just want a rename for Swagger.

weidongxu-microsoft: BTW, this is a new operation, all your reason for suppression is wrong!

ryazhang-microsoft: Fixed in 6e4db397784. I removed @operationId and the no-openapi suppression, then modeled this as a versioned replacement: the pre-preview operation is renamed with @renamedFrom, while the new operation keeps createLedgerEntry. This preserves Swagger operationId CreateLedgerEntry and the native SDK method name without a duplicate @clientName. I applied the same versioning pattern to the response model, so the preview 200 response again exposes LedgerWriteResult rather than LedgerWriteResultBody. The remaining use-standard-operations suppression now explains the actual constraint: Azure.Core.RpcOperation cannot accept the required union of the service default error and TransactionRolledBackErrorResponse.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
ryazhang-microsoft: Fixed in 6e4db397784. LedgerWriteResult.state is now required; the emitted preview schema requires both collectionId and state.

andpiccione (approved):

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
```diff
@@ -124,7 +124,42 @@ op listLedgerEntries is ConfidentialLedgerCommon.LedgerOperation<
  * A collection id may optionally be specified.
  */
 #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+@removed(Versions.v2026_07_31_preview)
+@renamedFrom(Versions.v2026_07_31_preview, "createLedgerEntry")
 @summary("Writes a ledger entry.")
+@sharedRoute
+@route("/app/transactions")
+@post
+op createLedgerEntryBeforeV20260731Preview is ConfidentialLedgerCommon.LedgerOperation<
```
heaths: This will generate a terrible SDK client method name. Rename, or do so in a `client.tsp` but easier to just rename here.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
```diff
@@ -124,7 +124,42 @@ op listLedgerEntries is ConfidentialLedgerCommon.LedgerOperation<
...
+     * The collection id.
+     */
+    @query("collectionId")
+    collectionId?: string;
+
+    /**
+     * Comma separated tags.
+     */
+    @query("tags")
+    tags?: string;
```
heaths, line 145: This should be an array.

heaths (changes requested):

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
ryazhang-microsoft: Fixed in 1c32885b4ed. I renamed the versioned legacy operation to `createLedgerEntryLegacy`, avoiding the generated `createLedgerEntryBeforeV20260731Preview` client method name. TypeSpec validation and `tsp compile .` pass.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
ryazhang-microsoft: Fixed in 1c32885b4ed. Both versioned `tags` query parameters are now `string[]`. The generated OpenAPI uses `type: array` with `collectionFormat: csv`, preserving the comma-separated wire representation. TypeSpec validation and `tsp compile .` pass.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
```diff
@@ -174,7 +174,7 @@ op createLedgerEntry is ConfidentialLedgerCommon.LedgerOperation<
      * Comma separated tags.
      */
     @query("tags")
-    tags?: string;
+    tags?: string[];
```
heaths: This is an SDK-breaking change to an existing API. Is this really the same operation?

heaths (dismissed):

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
ryazhang-microsoft: Yes, it is the same existing operation. You are right that changing `tags` from `string` to `string[]` would be SDK-breaking. I reverted that part in 89f4d7b9692, so `tags` remains a string for all versions and the generated prior-version OpenAPI is unchanged. The commit retains only the legacy operation rename to `createLedgerEntryLegacy`.

ryazhang-microsoft:
## Correction to the `1c32885b4ed` compatibility addendum

After the follow-up review question confirming that `/app/transactions` is the same existing operation, commit `89f4d7b9692` reverts the `tags` array change. `tags` remains a comma-separated `string` in every API version, preserving the existing SDK signature and generated prior-version OpenAPI.

The earlier addendum's `1026` and `1028` findings no longer apply to the current head. The remaining September 1 review fix is the legacy TypeSpec operation rename to `createLedgerEntryLegacy`.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
```diff
@@ -174,7 +174,7 @@ op createLedgerEntry is ConfidentialLedgerCommon.LedgerOperation<
      * Comma separated tags.
      */
     @query("tags")
-    tags?: string[];
+    tags?: string;
```
heaths, line 178: Technically, the array is correct but, yes, you must keep things from breaking. But if this is the old name, this is probably correct as written (not an array) but `createLedgerEntryLegacy` is net-new, right? That should be an array ideally, then.

ryazhang-microsoft: We should probably name it something like async and sync. Please see the attached pic for the usage. image

ryazhang-microsoft: createLedgerEntryLegacy is actually the old async model, so it requires polling after the request is made.

ryazhang-microsoft: Updated once more in 46e58fcee91 based on the polling clarification: the pre-preview projection helper is now `createLedgerEntryAsync`. The preview operation remains `createLedgerEntry` because its optional `waitForCommit` supports both immediate/polling (`false`) and wait-for-global-commit (`true`) modes. `tags` remains the existing comma-separated string to avoid an unrelated SDK-breaking signature change.

ryazhang-microsoft: Rationale for the final shape:

- **Name:** `createLedgerEntryAsync` represents the pre-`2026-07-31-preview` projection. TPAL returns after local commit, the transaction can remain Pending, and callers poll `/status` or `/receipt`; this is the existing asynchronous/polling model. The preview `createLedgerEntry` stays unchanged because the same method supports both modes through optional `waitForCommit` rather than being an always-synchronous operation.
- **`tags`:** TPAL currently reads `tags` as one string and splits it on commas (`tags=tag1,tag2`). Although an OpenAPI CSV array would be conceptually cleaner, changing `string` to `string[]` changes the generated SDK parameter type for this existing operation. Since this PR is specifically adding `waitForCommit`, we keep the established string contract and avoid an unrelated SDK break. A tags type correction should be handled separately with explicit compatibility review.

Implementation reference: https://github.com/microsoft/TPAL/pull/509

heaths: This doesn't follow our [guidelines](https://aka.ms/azapi/guidelines) for LROs at all. Whether to block and wait for an LRO is a client decision, not a service decision. Why are you doing it in the service?

Most (all) client SDKs have a way to block on async operations.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
```diff
@@ -437,6 +439,61 @@ model LedgerWriteResult {
...
+@added(Versions.v2026_07_31_preview)
+model LedgerWriteResultWithReceipt {
+  ...CCFTransactionIdHeader;
+
+  /**
+   * The response body.
+   */
+  @bodyRoot
+  body: LedgerWriteResult;
+}
```
weidongxu-microsoft: I didn't see why you need to "replace" the model.

Just adding the 3 new properties to your existing model should work? The most is another `@renamedFrom`

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
weidongxu-microsoft: Agree. This `waitForCommit` is untypical for Azure API.

Also, HTTP lib in client can timeout as well. Even if service can run for 2min during the REST API, client may not wait that long.

ryazhang-microsoft: Thanks. We agree that a newly designed Azure LRO should normally use `202` plus `Operation-Location`. This endpoint has an existing shipped contract that returns `200` after local commit and exposes a CCF transaction ID for custom `/status` and `/receipt` polling. Converting that existing operation into a standard LRO would break older ledger deployments, clients, and SDK contracts.

`waitForCommit` does not replace that behavior, and the service does not choose whether to block:

- Older ledger versions continue supporting only the existing asynchronous/polling flow.
- Newer ledgers preserve that flow when the parameter is absent or `false`.
- A client using `2026-07-31-preview` may explicitly select `waitForCommit=true`.

When selected, TPAL adapts CCF's native `set_consensus_committed_function` callback to wait for terminal consensus and return the committed receipt plus the transaction-scoped application claim directly. This removes repeated status polling and a separate historical receipt lookup; rollback is returned explicitly as `503 TransactionRolledBack`. Implementation: https://github.com/microsoft/TPAL/pull/509

This is intended as an opt-in synchronous performance optimization over the backward-compatible polling contract, not a replacement LRO protocol. The HTTP timeout concern is valid, and we need to provide measured consensus-finalization latency and bounded timeout behavior before approval. If it cannot be bounded safely below normal client and infrastructure timeouts, we should retain only polling or redesign it as a standard Azure LRO.

specification/confidentialledger/data-plane/ConfidentialLedger/models.tsp
ryazhang-microsoft: Agreed. I simplified this so `LedgerWriteResult` now evolves in place instead of being replaced. The existing `collectionId` remains on the model, and `receipt`, `state`, and `applicationClaims` are added with `@added(Versions.v2026_07_31_preview)`. Both operation projections now return the same model.

I also verified that the generated OpenAPI for the existing `2024-12-09-preview` and `2026-02-23` versions is unchanged. The `2026-07-31-preview` output now contains only the evolved `LedgerWriteResult`, without the extra wrapper model.

Fixed in 527cdd2fa3d.

specification/confidentialledger/data-plane/ConfidentialLedger/routes.tsp
heaths: And if a client disconnects? How does it get status? That's one reason we have the LRO contract we do: if the client retains the operation ID / URL they can check again in a subsequent request after a connection is restored, in another process, or even on another machine as part of a cluster. You can't do that with a sync API.

ryazhang-microsoft: Good point. Our existing async flow does not use a separate LRO operation ID. It returns  x-ms-ccf-transaction-id , which clients can use with the existing  /status  and  /receipt  endpoints after reconnecting, including from another process or machine.
 
 With  waitForCommit=true , the response—including that transaction-ID header—is held until consensus completes. If the connection is lost before the response arrives, the client will not receive that specific transaction ID directly. However, after reconnecting it can query the current entry for the relevant collection, which returns the transaction ID of the latest committed entry, allowing it to recover the latest durable ledger state.
 
 This is not completely equivalent to LRO recovery: if multiple clients write concurrently to the same collection, the latest entry may not uniquely identify the disconnected request. We should document that limitation clearly. Clients requiring unambiguous per-request recovery should continue using the default async flow;  waitForCommit=true  is an opt-in optimization for callers that can correlate the entry at the application level or tolerate that ambiguity.

heaths: Can you include those docs - at least a brief pointing to more information - in this API spec?

ryazhang-microsoft: Good suggestion. I added documentation to the API spec describing the existing transaction tracking flow through `x-ms-ccf-transaction-id` and the `/status` and `/receipt` endpoints, including recovery after reconnecting. It also now notes that with `waitForCommit=true`, a disconnect before the response arrives can prevent the caller from receiving the transaction ID and uniquely recovering that write, so clients requiring unambiguous per-request recovery should use the default polling flow. I also linked the Azure Confidential Ledger write transaction receipts documentation.

Updated in 6e041967702.

heaths: We've discuss internally and will approve. Another service already has set a precedent to optionally make an LRO effectively sync. You won't be able to do this with a single SDK client method, though. You'd need to redefine the operation once or even twice in a `client.tsp` such that the async remains an LRO(-like) and the sync is a normal method where `waitForCommit=true` is hardcoded, or add a convenience method atop generation. Reach out to the Azure SDK team if you need help then.

ryazhang-microsoft: Thanks for guiding me

ryazhang-microsoft:
**Update — resolved by dropping JS SDK generation for this spec (superseding the override request above).**

Root cause: the current @azure-tools/typespec-ts emitter no longer produces the classic rest-level client for this project even with is-modular-library: false — it emits a modular client into a package-root generated/ tree, which the shipped RLC @azure-rest/confidential-ledger package (
ootDir: src) cannot build (TS6059). Bringing the package current would be a breaking RLC→modular rewrite (new major version, reworking the MSRC redirect/merge-patch pipeline customizations, samples, tests, APIView) — out of scope for this spec PR.

Decision: remove the @azure-tools/typespec-ts emitter from `ConfidentialLedger/tspconfig.yaml` so JS is no longer generated/validated for this spec, matching `Microsoft.CodeTransparency` (which ships no JS SDK). This clears the required SDK Validation Status check without an admin override. The RLC `@azure-rest/confidential-ledger` package is no longer regenerated from this TypeSpec; a future JS SDK would require the modular migration in `azure-sdk-for-js`.

andpiccione (approved):

PR #45167, opened 2026-07-31
Microsoft.NetworkCloud 2026-08-01-preview API to public repo

bryan-strassner:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [X] New API version for an existing resource provider. (If API spec is not defined in TypeSpec, the PR should have been created in adherence to [OpenAPI specs PR creation guidance](https://aka.ms/azsdkdocs/createopenapispec)).
  - [ ] Other, please clarify:

## Due diligence checklist

- [X] I confirm this PR is modifying Azure Resource Manager (ARM) related specifications, and not data plane related specifications.
- [X] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including
- [X] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created. If not, please create one as it will help guide you through the REST API and SDK creation process. 

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


bryan-strassner:
Private repo equivalent PR (pre-v2 structure): https://github.com/Azure/azure-rest-api-specs-pr/pull/28719

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md
```diff
@@ -588,6 +583,128 @@ suppressions:
...
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.NetworkCloud/virtualMachines/{virtualMachineName}/reimage"].post
+  - code: PostResponseCodes
+    from: networkcloud.json
+    reason: False positive - properly using final-state-schema. Pending lintdiff fix.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.NetworkCloud/virtualMachines/{virtualMachineName}/restart"].post
+  - code: PostResponseCodes
+    from: networkcloud.json
+    reason: False positive - properly using final-state-schema. Pending lintdiff fix.
+    where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.NetworkCloud/virtualMachines/{virtualMachineName}/start"].post
+  - code: PatchBodyParametersSchema
```
ravimeda, line 705: **[NEW] 🟠 Warning** **[[RPC-Patch-V1-10](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#41-patch-body-must-not-have-required-properties-rpc-patch-v1-10) / [RPC-Patch-V1-05](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#41-patch-body-must-not-have-required-properties-rpc-patch-v1-10) / [Section 8 — Polymorphic Types](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#8-polymorphic-types) / [Suppression Decision Framework](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#decision-framework)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md` - line 702-705 - New `PatchBodyParametersSchema` suppression, re-introduced after a gap.

**Classification reasoning:** Introduced in this PR. Tag `package-2026-07-01` carries **zero** `PatchBodyParametersSchema` suppressions for this path, while `package-2026-05-01-preview` did — so this is a re-introduction after a gap, not a carry-over, and it needs fresh justification rather than the usual carried-forward pass.

**Issue:** Required-property traversal from `ClusterPatchProperties` returns an **empty set** on `stable/2026-07-01/networkcloud.json` and `{SecretArchiveProviderConfiguration.provider}` on the new preview — this PR is what puts a required property into the cluster PATCH body. The suppression also silently covers more than `provider`: `CyberArkSecretArchiveProviderConfiguration.required = ["applicationId", "safeName"]` (`preview/2026-08-01-preview/networkcloud.json` line 14472-14475) and `authenticationMethod` is non-optional on both the HashiCorp Vault and OpenBao models (`clusters/Cluster_models.tsp` lines 1194 and 1241), so the `reason:` under-describes what is being suppressed.

Separately, `clusters/Cluster_models.tsp` line 1045 states that `providerConfiguration` included in a PATCH `must be a complete, valid configuration for the chosen provider` — an explicit whole-object-replacement requirement on a sub-object, which deviates from the RFC 7396 merge semantics required by RPC-Patch-V1-05. The three new PATCH examples all send `provider` in the body, which Section 8 says a discriminator should not permit.

**Suggested fix:** `secretArchiveSettings` is the sole outlier among its immediate neighbours — `runtimeProtectionConfiguration`, `secretArchive` and `updateStrategy` all use dedicated `*Patch` variants (`RuntimeProtectionConfigurationPatch`, `ClusterSecretArchivePatch`, `ClusterUpdateStrategyPatch`). Adding a `SecretArchiveSettingsPatch` following the same idiom removes the required discriminator from the PATCH body and makes this suppression unnecessary. If you keep the current shape instead, please obtain explicit ARM reviewer sign-off and rewrite `reason:` to enumerate all four required properties.

The `where:` path itself is sound — it is byte-identical to the proven clauses at readme lines 97, 223 and 349.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/stable/2023-07-01/networkcloud.json
```diff
@@ -2684,8 +2684,8 @@
     },
     "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.NetworkCloud/clusters/{clusterName}/deploy": {
       "post": {
-        "description": "Deploy the cluster to the provided rack.",
-        "summary": "Deploy the cluster to the rack.",
+        "description": "Deploy the cluster using the rack configuration provided during creation.",
```
ravimeda, line 2687: **[NEW] 🟠 Warning** **[[Section 26.0](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#260-published-api-versions-are-immutable)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/stable/2023-07-01/networkcloud.json` - line 2687 - Hand-edits to two already-published GA API versions.

**Classification reasoning:** Introduced in this PR. `main.tsp` lines 57-87 declare `Versions` starting at `v2025_09_01`, so `stable/2023-07-01` and `stable/2025-02-01` are **not** emitter output — these files were edited directly.

**Issue:** Section 26.0 states that once an API version is published its schema is immutable. This PR modifies `stable/2023-07-01` (line 2687-2688 `Clusters_Deploy` description and summary; line 8490 `ConsolePatchProperties.enabled`) and `stable/2025-02-01` (lines 12686 and 12692, `capacityUsed` / `clusterId`).

In fairness, all five edits are description-only and are genuine corrections — line 8490 replaces a copy-paste error, and the `2025-02-01` change relocates a misplaced `Measured in GiB.` sentence to the property it actually describes. Section 26.0's literal flag condition ("if it only modifies the existing version") is also not met, since this PR does add a new version folder.

**Suggested fix:** Please confirm these documentation corrections to shipped GA versions are intentional and approved. Raising it so the decision is explicit rather than implicit — no change is being demanded.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
```diff
@@ -1029,11 +1036,221 @@ model SecretArchiveSettings {
...
+  encryptionPublicKey?: string;
+
+  /**
+   * The configuration indicating the use of self-supplied secret archive software. Specification of a provider configuration indicates that the provided configuration will be used. Exclusion of any providerConfiguration indicates the use of Azure Key Vault. If providerConfiguration is included in a PATCH, the body must be a complete, valid configuration for the chosen provider, including all fields required for that provider; omit providerConfiguration from PATCH bodies that do not change it.
+   */
+  @added(Versions.v2026_08_01_preview)
+  providerConfiguration?: SecretArchiveProviderConfiguration;
+
+  /**
+   * The URI for the key vault used as the secret archive. The URI must use the `https://` scheme.
```
ravimeda: **[NEW] 🟠 Warning** **[[Section 5 — `@added` version misalignment](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp` - line 1051 - Unversioned doc comment retroactively rewrote four already-published API versions.

**Classification reasoning:** Introduced in this PR — this sentence did not exist at the base commit.

**Issue:** The sentence `The URI must use the https:// scheme.` was appended to the `vaultUri` doc comment without version gating. Because doc comments are not versioned, the emitter rewrote the description in four **already-published** API versions: `stable/2025-09-01` line 15957, `stable/2026-07-01` line 18234, `preview/2026-01-01-preview` line 17710, and `preview/2026-05-01-preview` line 17999. Section 5 names this exact anti-pattern: a change gated for a new version must not inadvertently alter descriptions of earlier API versions.

**Suggested fix:** Scope the documentation change to `2026-08-01-preview` only, or confirm that retroactively rewriting the description in four shipped versions is acceptable and intended.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
```diff
@@ -1029,11 +1036,221 @@ model SecretArchiveSettings {
...
+
+  /**
+   * The folder name within the safe. Defaults to Root if not specified.
+   */
+  folderName?: string;
+
+  /**
+   * The object naming pattern within the safe. The tokens `{namespace}` and `{name}` may appear in the template and are substituted at archive time.
+   */
+  objectNameTemplate?: string;
```
ravimeda, line 1157: **[NEW] 🟠 Warning** **[[Section 8.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#82-use-objects-instead-of-strings-for-structured-values)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp` - line 1157 - Structured template values modelled as free-form strings.

**Classification reasoning:** Introduced in this PR — `objectNameTemplate` and `pathTemplate` are new properties gated by `@added(Versions.v2026_08_01_preview)`.

**Issue:** `objectNameTemplate` (line 1157) and `pathTemplate` (lines 1189 and 1236) carry a token grammar but are typed as bare `string`. The examples are the clearest evidence: `Clusters_Patch_SecretArchive.json` sends `{namespace}-{name}` and `Clusters_Patch_SecretArchive_HashiCorpVault.json` sends `azure-local/{namespace}/{name}` — both non-default token strings. The grammar is documented for `objectNameTemplate` (line 1155 says the tokens `are substituted at archive time`) but for `pathTemplate` it is only implied by the default value shown at line 1187.

**Suggested fix:** Model the template as a structured object, or at minimum document the full token grammar for `pathTemplate` explicitly and constrain the accepted format so clients get validation rather than a server-side rejection.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
```diff
@@ -1029,11 +1036,221 @@ model SecretArchiveSettings {
   associatedIdentity?: IdentitySelector;
 
   /**
-   * The URI for the key vault used as the secret archive.
+   * The public key used to encrypt secrets that are archived into the key vault. Expected encoding is a PEM-encoded RSA public key with a minimum key size of 3072 bits. Additional key formats or sizes may be supported in future versions.
```
ravimeda: **[NEW] 🟠 Warning** **[[Section 5 — doc comment accuracy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp` - line 1039-1042 - Azure Key Vault wording left in place after `vaultUri` was overloaded for third-party providers.

**Classification reasoning:** Introduced in this PR — the overload is new, and the `vaultUri` description was edited here without correcting the Azure-Key-Vault-specific wording.

**Issue:** `vaultUri` now carries third-party endpoints. The new examples set it to `https://vault.customer.com:8200` (HashiCorp Vault), `https://openbao.customer.com:8200` (OpenBao), and `https://cyberark.customer.com/PasswordVault/api` (CyberArk — a REST API base path, not a vault URI at all). Three descriptions still describe Azure Key Vault only: `vaultUri` at line 1051 (`The URI for the key vault used as the secret archive`), `encryptionPublicKey` at line 1039 (`archived into the key vault`), and `associatedIdentity` at line 1034 (`the managed identity to use with this vault URI`).

Note that line 1034 is **unchanged** by this PR — it is byte-identical to `stable/2026-07-01` line 18229. It became inaccurate because the property it describes was overloaded, not because the text was edited. A consequence worth noting: with `vaultUri` now spanning three different endpoint shapes, the Section 5 guidance to use `armResourceIdentifier` with type `Microsoft.KeyVault/vaults` for Key Vault URI properties can no longer be satisfied for this property in any version.

**Suggested fix:** Reword all three descriptions to provider-neutral language, for example: `The URI of the secret archive endpoint. For the default Azure Key Vault provider this is the Key Vault URI; when providerConfiguration is supplied this is the endpoint of the configured provider. The URI must use the https:// scheme.` A cleaner long-term option is splitting into `keyVaultId` (`arm-id`) plus `providerEndpoint` (`url`).

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clustermanagers/ClusterManager_models.tsp
```diff
@@ -174,6 +174,13 @@ model ClusterManagerProperties {
   @visibility(Lifecycle.Read)
   relayConfiguration?: ClusterManagerRelayConfiguration;
 
+  /**
+   * The relative ordering group used to apply software updates to associated clusters. The minimum accepted value is 1; the service enforces the upper bound currently in effect, which may change over time.
+   */
+  @added(Versions.v2026_08_01_preview)
+  @minValue(1)
```
ravimeda, line 181: **[NEW] 🔵 Suggestion** **[[Section 2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clustermanagers/ClusterManager_models.tsp` - line 181-182 - `rolloutRing` declares a minimum but not the maximum its own description says exists.

**Classification reasoning:** Introduced in this PR — `rolloutRing` is gated by `@added(Versions.v2026_08_01_preview)`.

**Issue:** Section 2.2 says integer properties with known valid ranges should use both `@minValue` and `@maxValue`. `rolloutRing` has `@minValue(1)` (emitting `minimum: 1`) but no `@maxValue`, while its own description states that `the service enforces the upper bound currently in effect, which may change over time`. So a bound exists, is enforced, and is undiscoverable — SDKs cannot validate it and clients only learn it from a server rejection. Same at lines 277-278 in `ClusterManagerPatchProperties`.

**Suggested fix:** Add `@maxValue(<n>)` for the currently supported ceiling. If the ceiling genuinely varies, state the current value in the description rather than describing it as unknowable.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
```diff
@@ -1029,11 +1036,221 @@ model SecretArchiveSettings {
...
+  extends SecretArchiveProviderConfiguration {
+  /**
+   * The provider of the secret archive. Literal value CyberArk.
+   */
+  provider: SecretArchiveProviderType.CyberArk;
+
+  /**
+   * The CyberArk application ID.
+   */
+  applicationId: string;
```
ravimeda, line 1142: **[NEW] 🔵 Suggestion** **[no rule ID — input-validation hardening]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp` - line 1142 - Unconstrained strings interpolated into an external secret store.

**Classification reasoning:** Introduced in this PR — all cited properties are gated by `@added(Versions.v2026_08_01_preview)`.

**Issue:** Twelve customer-supplied values that the service interpolates into paths and lookups inside an external secret store are bare, unconstrained `string`: `applicationId` (1142), `safeName` (1147), `folderName` (1152), `mountPath` (1174, 1221), `namespace` (1184, 1231), `applicationRoleId` (1199, 1246), `authenticationMountPath` (1204, 1251), and `encryptionPublicKey` (1042). `safeName` in particular is a CyberArk path segment, and `encryptionPublicKey` documents a specific encoding (`PEM-encoded RSA public key with a minimum key size of 3072 bits`) that nothing enforces.

To be clear about severity: no instruction-file rule requires `@pattern` or `@maxLength` on body strings, so this is a hardening suggestion rather than a violation. The value is client-side SDK validation plus a clear signal to implementers that these fields need server-side sanitization before they reach an external store.

A related modelling gap in the same area: the conditional requirement at lines 1192/1197 and 1239/1244 (`When set to AppRole, applicationRoleId is required; when set to ClientCertificate, applicationRoleId is ignored`) exists only in prose. No schema expresses it and no linter will catch a client that gets it wrong. A second discriminator level on `authenticationMethod` would model it properly.

**Suggested fix:** Add allowlist-style `@pattern` (positive character classes, not negated ones) and `@maxLength` to these properties.

**Two secret-detection false positives pre-empted, so no one re-raises them:** `applicationRoleId` is a HashiCorp AppRole RoleID, not the paired SecretID — line 1161 confirms credentials are supplied out of band — and `encryptionPublicKey` is a public key. Neither needs `@secret`.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md
```diff
@@ -61,14 +59,11 @@ suppressions:
     from: networkcloud.json
     reason: Nested objects that share a structure with PUT have required fields. The required field is present in the patch structure as well, because it reuses types. The nested structure needs to be updated in full by the user.
     where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.NetworkCloud/clusters/{clusterName}"].patch.parameters[4].schema.properties.properties
```
ravimeda, line 61: **[NEW] 🔵 Suggestion** **[[RPC-SUPPRESS-REMOVE-SAFELY](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#recommending-removal-of-an-existing-suppression-rpc-suppress-remove-safely)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md` - line 61 - `BodyTopLevelProperties` suppression removed from `package-2024-07-01`; please confirm the rule no longer fires.

**Classification reasoning:** Introduced in this PR — the suppression exists at base `readme.md` lines 64-66 and has zero occurrences at head.

**Issue:** This PR removes the `BodyTopLevelProperties` suppression from the `package-2024-07-01` tag block (which ends at head line 62). Its stated reason cited `azure-openapi-validator` issue 722, and I checked: that issue is **still open**, titled `The BodyTopLevelProperties rule is mistakenly flagging paged responses`. `stable/2024-07-01/networkcloud.json` contains 19 definitions with `value` plus `nextLink`, exactly the shape the bug reports.

Crucially, `stable/2024-07-01/networkcloud.json` is **not** among this PR's changed files, so the `package-2024-07-01` tag is most likely not exercised by LintDiff on this PR — the currently green CI is therefore not evidence that the removal is safe. The risk is latent and surfaces on the next PR that touches that file.

Mitigating evidence in the other direction: `stable/2026-07-01/networkcloud.json` has 21 definitions of the identical paged shape and its tag carries no such suppression, which suggests the validator behaviour may have changed even though the issue was never closed.

**Suggested fix:** Please confirm you verified the rule no longer fires for `package-2024-07-01` — removing a suppression is only safe once the underlying violation has stopped firing on the emitted artifact. If it does still fire, re-add it with a narrow `where:` clause rather than leaving the tag unprotected.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, manual decision) against commit [`5aba8a3`](https://github.com/Azure/azure-rest-api-specs/pull/45167/commits/5aba8a3384af76528b79fe1170ed9d296473e361). See inline comments for findings 1-5, 7, 9 and 10. No blocking issues were found: 5 warnings and 5 suggestions, all introduced by this PR.

Findings 6 and 8 concern `preview/2026-08-01-preview/networkcloud.json`, whose diff GitHub does not return (the file is too large), so they are posted as top-level PR comments rather than inline.

ravimeda:
**[NEW] 🔵 Suggestion** **[no rule ID — security review question]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/preview/2026-08-01-preview/networkcloud.json` - line 13556 - External secret-store topology is readable by any subscription-scope reader.

**Classification reasoning:** Introduced in this PR — `providerConfiguration` and its subtype fields are new in `2026-08-01-preview`.

**Issue:** `ClusterProperties.secretArchiveSettings` (line 13556) has no `readOnly` and no visibility restriction, and the response chain `Clusters_ListBySubscription` -> `ClusterList` -> `Cluster` -> `ClusterProperties` returns it in full. Any principal holding `Microsoft.NetworkCloud/clusters/read` at subscription scope can therefore enumerate, for every cluster: `vaultUri`, `namespace`, `mountPath`, `authenticationMountPath`, `pathTemplate`, `safeName`, `applicationId` and `applicationRoleId`.

No secret leaks here — the credentials themselves are supplied out of band, and I am not claiming a rule violation. But the aggregate is a reconnaissance map of the customer's on-premises secret store, exposed at a much broader RBAC scope than the store's own access controls.

**Suggested fix:** Consider narrowing read visibility on the authentication-bearing subset, or exposing that subset through a dedicated POST action (for example `listSecretArchiveConfiguration`) so it can carry its own RBAC action. Raising this for an explicit design decision rather than as a required change.

ravimeda:
**[NEW] 🔵 Suggestion** **[[Section 8.14 (OAPI022, WHATIF-002)](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#814-default-values-must-be-static-and-declared-oapi022-whatif-002)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/preview/2026-08-01-preview/networkcloud.json` - line 14772 - Service-applied defaults documented in prose but never declared in the schema.

**Classification reasoning:** Introduced in this PR — all four properties are new in `2026-08-01-preview`.

**Issue:** Four optional properties document a default that the service applies but do not declare `default` in the schema: `mountPath` (line 14772, `Defaults to secret`), `keyValueVersion` (line 14776, `defaults to V2`), `pathTemplate` (line 14784, `Defaults to edge-credentials/{namespace}/{name}`), and `folderName` (line 14463, `Defaults to Root`); `mountPath`, `keyValueVersion` and `pathTemplate` repeat in the OpenBao model at lines 17355, 17359 and 17367. Section 8.14 requires the default be declared so ARM Template What-If does not report a false delete when a template omits the property.

**Suggested fix — note that simply adding `default` is not safe here.** `SecretArchiveSettings` is referenced from both `ClusterProperties` (line 13557) and `ClusterPatchProperties` (line 13459), so anything emitted into it also lands in the PATCH body, where `PatchBodyParametersSchema` and RPC-Patch-V1-10 forbid `default`. The TypeSpec anti-pattern list names this trap directly (`Remove the TypeSpec default and apply it server-side instead`). Workable options:

1. Split the PATCH model as suggested in the `readme.md` suppression comment on this PR, then declare the defaults on the PUT-side model only.
2. Keep the defaults server-side and documented, accepting the What-If noise for this preview version.
3. Drop the `Defaults to ...` language if the service does not actually apply these defaults.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md
bryan-strassner: Re: PatchBodyParametersSchema suppression at [readme.md](vscode-file://vscode-app/c:/Users/bstrassner/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) L702

The classification and the property-level detail are accurate. This shape is intentional and was recommended by ARM reviewers after discussion in the spec review [azure-rest-api-specs-pr#28719](https://github.com/Azure/azure-rest-api-specs-pr/pull/28719) and discussion in the TypeSpec channel prior to that, so I'd like to keep it and explain the rationale for the record.

The required properties the rule is detecting are all scoped inside the optional secretArchiveSettings.providerConfiguration sub-object — they are not top-level required PATCH properties. providerConfiguration is optional on the parent; omitting it leaves the setting unchanged. Only when a caller chooses to send it does the discriminator (provider) and its per-provider required fields apply.

That "required-within-an-optional-subtree" case is exactly what the linter can't express, and it's why whole-object replacement is the correct contract here rather than a deviation: a discriminated type cannot follow RFC 7396 field-merge semantics without risking an invalid cross-provider blend (e.g. HashiCorp fields spliced onto a CyberArk discriminator). The alternatives were considered and are worse:

* Making provider optional yields an unresolvable/ambiguous body (the server can't determine the subtype).
* Flattening to a single non-discriminated model passes the linter but discards the type safety and validity guarantees, and still requires provider to act on the object.

Given that, the suppression is the right mechanism for encoding a rule limitation, not a workaround for a design problem. I've updated the reason: to enumerate all four suppressed properties and cite the ARM review, so the justification is self-documenting for future passes.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/stable/2023-07-01/networkcloud.json
bryan-strassner: **Confirmed**

The descriptions are now aligned with the proper verbiage, and as the impact to fix the verbiage only in the current version would require expressions in TypeSpec that are overly burdensome for such a correction, this PR introduces them as the single description/summary for the action. The prior verbiage is decidedly incorrect, especially in the summary, as the cluster is not deployed to a singular rack, but rather the collection of racks, or the cluster of servers/racks that make up the cluster.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
bryan-strassner: **Confirmed**

_The description is accurate for all prior versions._
The private repo submission of this change: https://github.com/Azure/azure-rest-api-specs-pr/pull/28719 included this as well and was approved. The disposition was:

> S6 — vaultUri scheme not restricted
> 
> Addressed by documentation. The vaultUri description now states that the URI must use the https:// scheme. We are not adding a pattern at this time; the prose constraint matches our handling elsewhere for similar URI fields.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
bryan-strassner: Previously approved in https://github.com/Azure/azure-rest-api-specs-pr/pull/28719

However, as this is an accurate new finding, I will apply here and travel it backwards to other sources.

Confirmed all three providers use the same {namespace}/{name} token substitution — the Vault examples send token-bearing values (azure-local/{namespace}/{name}), so pathTemplate is a substitution template, not a fixed path. Only the documentation was asymmetric: CyberArk's objectNameTemplate spelled out the grammar while pathTemplate only implied it via its default.

Fixed by documenting the token grammar on pathTemplate for both the HashiCorp Vault and OpenBao models, mirroring the objectNameTemplate wording. Applied in TypeSpec and regenerated into the swagger.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
bryan-strassner: Agreed on the substance — with vaultUri now carrying non-Key-Vault endpoints, Key-Vault-specific wording is inaccurate. Here's what I've changed, field by field, and the one place I'm intentionally not following the suggested remediation.

vaultUri — reworded, provider-neutral and cross-version valid. New text: "The URI of the secret archive endpoint. The URI must use the https:// scheme." This is true in every generated version — in the earlier versions the secret archive is Azure Key Vault, so the endpoint is the Key Vault URI; from 2026-08-01-preview it's the configured provider's endpoint. Because the wording is accurate across all versions, I regenerated it into all of them (2025-09-01, 2026-01-01-preview, 2026-05-01-preview, 2026-07-01, 2026-08-01-preview).

On the versioning mechanics: TypeSpec attaches a single doc string to a property across every version it exists in, and there's no per-version doc decorator — isolating description text to one version requires deprecating the property and re-adding it. I'm not willing to take on that disproportionate rename/deprecate/re-add mechanism here, precisely because a single wording is valid across all versions, so the shared string is correct everywhere and no isolation is needed. To be clear: if there were no cross-version-valid phrasing — if the only accurate new text would be wrong in an older version — I'd concede that the heavier versioning mechanism is the right call. That just isn't the situation for this property.

encryptionPublicKey — reworded, provider-neutral. New text: "The public key used to encrypt secrets before they are written to the secret archive…" (was "archived into the key vault"). This property is @added(2026-08-01-preview), so the change is contained to that version and touches nothing older. The examples continue to set it, since it applies regardless of provider.

associatedIdentity — description unchanged, examples corrected instead. Managed-identity selection applies only to the Azure Key Vault path; the third-party providers authenticate out of band (authenticationMethod/applicationRoleId, or CyberArk applicationId). The existing description is accurate for how the field is actually supported, so I've left it. What was misleading was the examples populating associatedIdentity for the third-party providers — a valid-but-meaningless payload. I removed secretArchiveSettings.associatedIdentity from the CyberArk, HashiCorp Vault, and OpenBao examples (request and response); the Azure Key Vault example keeps it. The description and examples are now consistent on this being Key-Vault-scoped.

Full list of changes:

- clusters/Cluster_models.tsp: vaultUri and encryptionPublicKey descriptions reworded; associatedIdentity intentionally unchanged.
- Regenerated swagger: vaultUri description in all five emitted versions; encryptionPublicKey in 2026-08-01-preview only.
- Examples (examples/2026-08-01-preview/ source + generated preview/2026-08-01-preview/examples/): secretArchiveSettings.associatedIdentity removed from [Clusters_Patch_SecretArchive.json](vscode-file://vscode-app/c:/Users/bstrassner/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (CyberArk), _HashiCorpVault.json, _OpenBao.json; _AzureKeyVault.json unchanged.

I did not adopt the alternative structural split (keyVaultId + providerEndpoint) — the current single-URI shape was reviewed and approved, and the reworded descriptions resolve the accuracy concern without a breaking change.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clustermanagers/ClusterManager_models.tsp
bryan-strassner: This was addressed in the private repo review as follows. No change.

> W5 — Numeric constraints missing on rolloutRing
> 
> Partially addressed. minimum: 1 is now applied to both ClusterManagerProperties.rolloutRing and ClusterManagerPatchProperties.rolloutRing. We are intentionally not setting a maximum: the service enforces the upper bound currently in effect, and we expect that bound to evolve over time. Publishing a maximum here would imply a stable contract that isn't true, and would require API surface changes to adjust. The description now states the minimum and notes that the upper bound is service-enforced and may change.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
bryan-strassner: Referencing the private repo suggestions/dispositions: https://github.com/Azure/azure-rest-api-specs-pr/pull/28719

Noting up front this is a Suggestion, not a rule violation. This is the same class as S2 (template hygiene), which we declined. That decision stands and extends to the fields raised here.

@pattern / @maxLength on the provider path/lookup strings (applicationId, safeName, folderName, mountPath, namespace, applicationRoleId, authenticationMountPath) — Declined. These are passed through to external systems (CyberArk, HashiCorp Vault, OpenBao) whose authoritative length limits and character sets are owned by those vendors, not by us. A @pattern/@maxLength is a published contract; encoding a limit we can't authoritatively back would reject valid input the downstream system accepts, and become a forward liability the moment a vendor's rules differ from our guess. Omitting the constraint is the correct call, not a gap to be closed. Sanitization before these values reach the external store is a server-side responsibility and is handled there.

encryptionPublicKey — Declined. Its encoding is our own documented contract (PEM-encoded RSA, ≥3072-bit), but a regex can only assert the PEM framing, not the key strength — the meaningful check requires decoding the key. A half-constraint that validates the wrapper and not the strength gives clients false confidence, so validation stays server-side where it can be done correctly.

Conditional applicationRoleId requirement expressed in prose — Declined. Modeling the AppRole/ClientCertificate → applicationRoleId relationship as a second discriminator level is disproportionate for a single transitive field. It hard-codes today's vendor authentication shape into the wire contract, which constrains future evolution and could preclude vendor interface changes we don't yet control. A mildly looser configuration whose behavior is documented accurately is the deliberate balance here — the prose states the rule, and that is sufficient without freezing it into the schema.

Secret-detection pre-empts — agreed: applicationRoleId is an AppRole RoleID (not the paired SecretID; credentials are supplied out of band per the model docs), and encryptionPublicKey is a public key. Neither warrants @secret.

No spec changes for this item.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md
bryan-strassner: Good catch — this was an inadvertent drop when porting the readme from our source, not a deliberate removal, and your analysis is correct. I've restored the suppression to the package-2024-07-01 tag, verbatim to its prior form.

bryan-strassner:
> **[NEW] 🔵 Suggestion** **[no rule ID — security review question]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/preview/2026-08-01-preview/networkcloud.json` - line 13556 - External secret-store topology is readable by any subscription-scope reader.
> 
> **Classification reasoning:** Introduced in this PR — `providerConfiguration` and its subtype fields are new in `2026-08-01-preview`.
> 
> **Issue:** `ClusterProperties.secretArchiveSettings` (line 13556) has no `readOnly` and no visibility restriction, and the response chain `Clusters_ListBySubscription` -> `ClusterList` -> `Cluster` -> `ClusterProperties` returns it in full. Any principal holding `Microsoft.NetworkCloud/clusters/read` at subscription scope can therefore enumerate, for every cluster: `vaultUri`, `namespace`, `mountPath`, `authenticationMountPath`, `pathTemplate`, `safeName`, `applicationId` and `applicationRoleId`.
> 
> No secret leaks here — the credentials themselves are supplied out of band, and I am not claiming a rule violation. But the aggregate is a reconnaissance map of the customer's on-premises secret store, exposed at a much broader RBAC scope than the store's own access controls.
> 
> **Suggested fix:** Consider narrowing read visibility on the authentication-bearing subset, or exposing that subset through a dedicated POST action (for example `listSecretArchiveConfiguration`) so it can carry its own RBAC action. Raising this for an explicit design decision rather than as a required change.

Good question to put on the record. Our read model gates secret material behind dedicated actions (the listKeys pattern); configuration and identifiers are returned on read. These fields are the latter — endpoints, mount paths, and identity handles — and the credentials are supplied out of band, so nothing returned here is usable to authenticate.

applicationRoleId is the only credential-adjacent field, and it follows a pattern we already ship: it's an AppRole RoleID (inert without the out-of-band SecretID), just as clusterServicePrincipal.applicationId is readable while its password is @secret. It's consistent with the existing exposure model, and less detailed than infra already returned at this scope (BMC addresses, keyVaultId, containerUrl).

Keeping it as-is. If a field here ever returns secret material, the POST-action split is the right tool and we'll use it.

bryan-strassner:
> **[NEW] 🔵 Suggestion** **[[Section 8.14 (OAPI022, WHATIF-002)](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#814-default-values-must-be-static-and-declared-oapi022-whatif-002)]** `specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/preview/2026-08-01-preview/networkcloud.json` - line 14772 - Service-applied defaults documented in prose but never declared in the schema.
> 
> **Classification reasoning:** Introduced in this PR — all four properties are new in `2026-08-01-preview`.
> 
> **Issue:** Four optional properties document a default that the service applies but do not declare `default` in the schema: `mountPath` (line 14772, `Defaults to secret`), `keyValueVersion` (line 14776, `defaults to V2`), `pathTemplate` (line 14784, `Defaults to edge-credentials/{namespace}/{name}`), and `folderName` (line 14463, `Defaults to Root`); `mountPath`, `keyValueVersion` and `pathTemplate` repeat in the OpenBao model at lines 17355, 17359 and 17367. Section 8.14 requires the default be declared so ARM Template What-If does not report a false delete when a template omits the property.
> 
> **Suggested fix — note that simply adding `default` is not safe here.** `SecretArchiveSettings` is referenced from both `ClusterProperties` (line 13557) and `ClusterPatchProperties` (line 13459), so anything emitted into it also lands in the PATCH body, where `PatchBodyParametersSchema` and RPC-Patch-V1-10 forbid `default`. The TypeSpec anti-pattern list names this trap directly (`Remove the TypeSpec default and apply it server-side instead`). Workable options:
> 
> 1. Split the PATCH model as suggested in the `readme.md` suppression comment on this PR, then declare the defaults on the PUT-side model only.
> 2. Keep the defaults server-side and documented, accepting the What-If noise for this preview version.
> 3. Drop the `Defaults to ...` language if the service does not actually apply these defaults.

Thanks — this prompted a useful correction of premise rather than a schema change. These fields are not persisted defaults. The service uses the documented value transiently during processing when the caller omits the field; it does not write the value back into the resource. A GET after an omitting request returns the field unset — round-tripping a server-chosen value into a writable input field would be a bug in our model (it's the drift-fight that What-If and Terraform both punish).

Because the value never materializes in the read representation, the WHATIF-002 false-delete can't arise — there's nothing in current state to mismatch an omitting template — and declaring a schema default would actively misrepresent state that GET never returns. So we're not declaring default, and we've corrected the descriptions to state the behavior accurately (e.g. "When not specified, the service uses secret") rather than implying a stored default.

bryan-strassner:
Addressed all comments with disposition. Please review.

ravimeda (approved):

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/readme.md
ravimeda: Thanks for addressing this. The `reason:` now enumerates all four required properties and cites the ARM review, which was the actionable part of this finding.

Providing the ARM sign-off this comment asked for: the shape is accepted. Required properties scoped inside an optional discriminated sub-object are not expressible to `PatchBodyParametersSchema`, and whole-object replacement is the correct contract for a discriminated type — RFC 7396 field-merge would permit invalid cross-provider blends. The suppression is encoding a linter limitation, not papering over a design problem.

Verified at `34cc041f`: the `where:` clause at `readme.md` line 716 is byte-identical to the proven clauses at lines 100, 226 and 352, and `parameters[6]` is correct — the clusters PATCH parameter list is `[0]ApiVersionParameter [1]SubscriptionIdParameter [2]ResourceGroupNameParameter [3]clusterName [4]IfMatchParameter [5]IfNoneMatchParameter [6]clusterUpdateParameters`. LintDiff passes. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/stable/2023-07-01/networkcloud.json
ravimeda: Thanks — confirmation was all this comment asked for. Re-verified at `34cc041f`: `main.tsp` declares `Versions` starting at `v2025_09_01`, so these two files sit outside the set of versions the emitter produces, and all five edits remain description and summary only, with no schema, type or enum change. Documentation corrections of demonstrably wrong text are an acceptable exception to version immutability. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
ravimeda: Thanks — confirmation was what this comment asked for, and the reasoning holds. "The URI of the secret archive endpoint. The URI must use the `https://` scheme." is accurate in every version it is emitted into, so the shared doc string is correct everywhere and the deprecate-and-re-add mechanism would be disproportionate. Agreed. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
ravimeda: Thanks for addressing this. Verified at `34cc041f`: `pathTemplate` at `clusters/Cluster_models.tsp` lines 1187 and 1234 now documents the `{namespace}` and `{name}` token grammar, mirroring the `objectNameTemplate` wording at line 1155, and the text is emitted into both the HashiCorpVault and OpenBao definitions in the generated swagger. The documentation asymmetry this comment flagged is gone. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
ravimeda: Thanks for addressing this. Verified at `34cc041f`:

- `encryptionPublicKey` (line 1039) now reads "written to the secret archive" rather than "archived into the key vault".
- `vaultUri` (line 1051) is provider-neutral.
- `secretArchiveSettings.associatedIdentity` is absent from the request and 200-response bodies of `Clusters_Patch_SecretArchive.json`, `Clusters_Patch_SecretArchive_HashiCorpVault.json` and `Clusters_Patch_SecretArchive_OpenBao.json`, and present only in `Clusters_Patch_SecretArchive_AzureKeyVault.json`.

Leaving the `associatedIdentity` description and correcting the examples instead is the right call — the field genuinely is Key-Vault-scoped, and the examples were the misleading part. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clustermanagers/ClusterManager_models.tsp
ravimeda: Accepted, no change required. Publishing a `maximum` that the service expects to move would assert a contract stability that does not exist, and every subsequent adjustment would become an API surface change. `minimum: 1` plus a description stating the upper bound is service-enforced is the right balance. Resolving this thread.

specification/networkcloud/resource-manager/Microsoft.NetworkCloud/NetworkCloud/clusters/Cluster_models.tsp
ravimeda: Accepted, no change required. The reasoning is sound on each point: `@pattern` and `@maxLength` on values passed through to CyberArk, HashiCorp Vault and OpenBao would publish limits we do not own; a PEM regex would validate the framing while giving false confidence about key strength; and a second discriminator level would freeze the current vendor authentication shape into the wire contract. Server-side validation is the correct home for all three. Resolving this thread.

ravimeda:
> Re: [this comment](https://github.com/Azure/azure-rest-api-specs/issues/45167#issuecomment-5170572551) — external secret-store topology readable at subscription scope.

Accepted, no change required. The distinction you draw is the right one — configuration and identifiers on read, secret material behind dedicated actions — and `applicationRoleId` being an AppRole RoleID that is inert without the out-of-band SecretID makes it consistent with `clusterServicePrincipal.applicationId`, which already ships readable at this scope. Noting for the record that the POST-action split remains the right tool if a field here ever begins returning secret material.

ravimeda:
> Re: [this comment](https://github.com/Azure/azure-rest-api-specs/issues/45167#issuecomment-5170573574) — service-applied defaults not declared in the schema.

Accepted — and you are right that the premise needed correcting rather than the schema. If the value is never written back into the resource, a GET after an omitting request returns the field unset, so there is no current state for What-If to mismatch, and declaring a schema `default` would misrepresent state the read model never returns.

Verified at `34cc041f`: the descriptions now use "When not specified, the service uses X" — `folderName` line 1150, `mountPath` lines 1172 and 1219, `keyValueVersion` lines 1177 and 1224, `pathTemplate` lines 1187 and 1234. No `default` keyword is emitted on any of the new secret-archive properties, and the 38 pre-existing schema-level `default` keywords in the document are unchanged from `2026-07-01`.

bryan-strassner:
Merging will be held until closer to the end of August.

bryan-strassner:
@copilot resolve the merge conflicts in this pull request in the service.yaml file to align newly introduced API version of this PR with what is the latest on main.

mbashtovaya (approved):

PR #44416, opened 2026-07-03
Add getAgentInvocationAsyncApiSpec{Json,Yaml} operations

ZhaoChaoqun:
Adds two new operations to `AgentInvocations` (agents-invocations
data-plane) that describe the AsyncAPI docs endpoint contract for
hosted-agent containers:

- `getAgentInvocationAsyncApiSpecJson` — `GET .../invocations/docs/asyncapi.json`
- `getAgentInvocationAsyncApiSpecYaml` — `GET .../invocations/docs/asyncapi.yaml`

AsyncAPI is the companion to OpenAPI for streaming/bidirectional surfaces
(e.g. the `invocations_ws` WebSocket protocol) that OpenAPI cannot express.
The path extension is authoritative for the returned content type — no
`Accept` negotiation.

Follows the spec-text contract established by
coreai-microsoft/foundrysdk_specs#209 (Yulin), which added `.md`-level
guidance for the AsyncAPI docs endpoint. This PR wires the same endpoint
into the TypeSpec surface so SDKs and generated clients pick it up.

Work item: AB#5407708
Related coordination PR (foundrysdk_specs, Yulin): coreai-microsoft/foundrysdk_specs#209

PR #45172, opened 2026-08-03
Reapply "Add getAgentInvocationAsyncApiSpec{Json,Yaml} operations" (re-submit of #44416 after review-gate discussion)

ZhaoChaoqun:
## Summary

Re-submits the AsyncAPI docs endpoint TypeSpec change originally landed in #44416, then reverted in #44835 to satisfy the experience/product review gate flagged by @jsquire-ms / @travisez13.

This PR reapplies the same change on top of the latest `feature/foundry-release`.

## Why re-submit now

Since the revert, the AsyncAPI direction has been reviewed by the OpenAPI SME and stewardship board:

- **@vincentbiret** (OpenAPI SME) confirmed AsyncAPI is the best way to represent WS today; the recommended pattern is documenting the WS init in OpenAPI and the rest of the WS flow in AsyncAPI — which is exactly what these endpoints enable.
- **@travisez13** shared this alignment with the Azure REST API board thread; no objection has been raised in the review window.
- Tooling gaps around TypeSpec → AsyncAPI emission remain (tracked in microsoft/typespec#2463) but do not block the discovery endpoint itself.

## What this does

- Adds `getAgentInvocationAsyncApiSpecJson` — `GET .../invocations/docs/asyncapi.json`
- Adds `getAgentInvocationAsyncApiSpecYaml` — `GET .../invocations/docs/asyncapi.yaml`
- Regenerates the openapi3 `v1` and `virtual-public-preview` artifacts to keep TSP source and OpenAPI in sync.

Net: 531 insertions across 5 files — exact inverse of the revert in #44835.

## Related

- Original PR: #44416
- Revert: #44835
- Feature spec (contract prose): coreai-microsoft/foundrysdk_specs#209
- Work item: AB#5407708

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents-invocations/routes.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents-invocations/routes.tsp
@@ -31,6 +31,51 @@ interface AgentInvocations {
   >;
 
   /**
+   * Retrieves the AsyncAPI (JSON) specification for an agent version's event-driven
+   * invocation contract (e.g. the `invocations_ws` WebSocket protocol). AsyncAPI is the
+   * companion to OpenAPI for streaming/bidirectional surfaces that OpenAPI cannot express.
+   * Returns 404 if the agent does not publish an AsyncAPI specification. Publishing is
+   * optional; when published, both JSON and YAML representations are available.
+   */
+  @get
+  @route("/agents/{agent_name}/versions/{agent_version}/invocations/docs/asyncapi.json")
+  getAgentInvocationAsyncApiSpecJson is FoundryDataPlaneOperation<
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The version of the agent. */
+      @path agent_version: string;
+    },
+    {
+      @header contentType: "application/json";
+      @body spec: Record<unknown>;
+    }
+  >;
+
+  /**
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents-invocations/routes.tsp
```diff
@@ -30,6 +30,51 @@ interface AgentInvocations {
...
+    {
+      /** The name of the agent. */
+      @path agent_name: string;
+
+      /** The version of the agent. */
+      @path agent_version: string;
+    },
+    {
+      @header contentType: "application/yaml";
+      @body spec: string;
```
baywet: the serialization format should not impact the body type. I believe that this, and the json variant, should be aligned

baywet (changes requested):

specification/ai-foundry/data-plane/Foundry/src/agents-invocations/routes.tsp
ZhaoChaoqun: thanks you're right, the serialization format shouldn't change the body type.

ZhaoChaoqun: aligned both variants to a shared `AsyncApiDocument`

baywet (approved):

PR #45187, opened 2026-08-03
Add Content Safety team as CODEOWNERS for ContentSafety specs

Connor-Wood:
Adds `@Azure/azure-sdk-write-contentsafety` as source owners for the Content Safety spec directories, so Content Safety PRs are routed to the service team rather than requiring `@rkuthala` / `@fmabroukmsft` on every change.

Two entries are needed because Content Safety spans both the TypeSpec project and the generated data-plane output:

```
/specification/cognitiveservices/ContentSafety/              # TypeSpec (main.tsp, client.tsp, tspconfig.yaml)
/specification/cognitiveservices/data-plane/ContentSafety/   # generated swagger + readme.md
```

Both are placed after the existing `/specification/cognitiveservices/` catch-all so they take effect (GitHub uses the last matching entry).

The team already exists and meets the requirements in [the CODEOWNERS policy](https://github.com/Azure/azure-sdk/blob/main/docs/policies/opensource.md#codeowners):

- in the `Azure` org, referenced as `@Azure/azure-sdk-write-contentsafety`
- child of `azure-sdk-write`, with `push` access to this repository
- described as "Azure Content Safety devs and PMs with write access to the Azure SDK repos"

Using the team rather than individual aliases follows the pattern already used for KeyVault, PostgreSQL and Online Experimentation, and means future membership changes do not require a CODEOWNERS PR.

`Cognitive - Content Safety` is an existing repository label.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -67,6 +67,12 @@
 # PRLabel: %Cognitive Services
 /specification/cognitiveservices/ @rkuthala @fmabroukmsft
 
+# PRLabel: %Cognitive - Content Safety
+/specification/cognitiveservices/ContentSafety/ @Azure/azure-sdk-write-contentsafety
+
+# PRLabel: %Cognitive - Content Safety
+/specification/cognitiveservices/data-plane/ContentSafety/ @Azure/azure-sdk-write-contentsafety
+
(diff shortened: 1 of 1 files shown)


danieljurek (approved):
This is good for now. It may need to change in the future as the process for ownership team nesting evolves.

PR #45194, opened 2026-08-03
Add ProviderHub 2025-10-01 stable and 2026-02-01-preview API specs

wwendyc:
## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

This PR adds two new Microsoft.ProviderHub API versions, originating from the private spec repo:

- `stable/2025-10-01`
- `preview/2026-02-01-preview`

Both versions are **generated from TypeSpec**. Microsoft.ProviderHub is already TypeSpec-based on `main` (`specification/providerhub/ProviderHub.Management`, currently emitting `2024-09-01`), so the two new versions were added to that project rather than maintained as hand-written swagger. The `providerhub.json` files under `resource-manager/` are emitter output.

`2024-09-01` re-emits byte-for-byte identical — this PR makes no change to that version's API surface.

The `readme.md` autorest configuration carries `package-2025-10-01` and `package-2026-02-01-preview` tag blocks, with the default tag set to `package-2025-10-01` (latest stable).

### What's new in `2025-10-01`

- New `ManifestInfo` resource: `GET`/`PUT` on `.../providerRegistrations/{providerNamespace}/manifests/{environment}`
- New tenant-level action `NewRegionFrontloadRelease_ValidateManifest`
- New models including the `ManifestInfo` family, `WriteLockConfiguration`, `PrivateEndpointConfiguration`, `ResourceDeletionPolicyAndProperties`, `ManifestCheckinSpecification`, `AppliedManifestInfo`, `ResourceTypeManagedResourceGroupConfiguration`
- ~15 new properties across existing models (`oboSubscriptionId`, `enablePresetResourceTypes`, `writeLock`, `superScaleEnabled`, `resourceDeletionPolicies`, `featureManagementOwners`, `bucketSize`, `completedRegionsInfo`, and others)

### What's new in `2026-02-01-preview`

- Lifecycle metadata: `ResourceProviderLifecycleInfo`, `ResourceTypeLifecycleInfo`, `SwaggerSpecificationLifecycleInfo`, plus the `LifecycleStage`, `IsolationType` and `PartnerCategory` enums
- `lifecycleInfo` on `ProviderRegistrationProperties`, `ResourceTypeRegistrationProperties` and `SwaggerSpecification`
- `featureApprovalClaims` on `ResourceProviderManagement`
- Removes `ResourceProviderManifest.tokenAuthConfiguration` and the `validateManifestAppliedInAllRegions` operation, both of which were introduced in `2025-10-01`

### Breaking changes in `2025-10-01` (stable → stable)

These are intentional and modelled explicitly with versioning decorators. They will trip the breaking-change gate and require Breaking Change Review approval:

| Change | Detail |
| --- | --- |
| `OperationsDefinition.properties` removed | `@removed(Versions.v2025_10_01)` |
| `OperationsContent.properties` → `contents` | Renamed **and** retyped from a single `OperationsDefinition` to an array |

### Enum corrections

The versions as originally migrated contained two `x-ms-enum` **name collisions** — one name mapped to two different value sets, which would break SDK generation. These have been corrected against the service class definitions and confirmed with the service team. The emitted spec now has zero enum name collisions in all three versions.

| Property | Enum (from `2025-10-01`) | Values |
| --- | --- | --- |
| `ResourceTypeRegistrationProperties.resourceDeletionPolicy` | `RPaaSResourceDeletionPolicy` | `NotSpecified`, `CascadeDeleteAll`, `CascadeDeleteProxyOnlyChildren`, `Cascade`, `Force` |
| `ResourceType.resourceDeletionPolicy`<br>`ResourceDeletionPolicyAndProperties.policyName` | `ResourceDeletionPolicy` | `NotSpecified`, `Cascade`, `Force`, `SoftDelete` |
| `ResourceProviderManifest.providerType`<br>`ResourceProviderManifestProperties.providerType` | `ResourceProviderType` | one shared enum, now including `Decommissioned` |
| `WriteLockConfiguration.state` | `WriteLockState` | `Disabled`, `Enabled` |

Three SDK-visible type names therefore change **as of `2025-10-01` only** (`2024-09-01` is untouched):

- `ResourceDeletionPolicy` → `RPaaSResourceDeletionPolicy`
- `ManifestResourceDeletionPolicy` → merged into `ResourceDeletionPolicy`
- the previously generic `State` → `WriteLockState` (the wire property is still `state`)

### Suppressions added

- `arm-resource-provisioning-state` on `ManifestInfoProperties` — the manifest is read from the manifest repository and has no provisioning state
- `@operationId` and non-ARM interface suppressions on the tenant-level `NewRegionFrontloadRelease_ValidateManifest` action, to preserve the existing operationId and route

## Due diligence checklist

  - Release plan [35909](https://dev.azure.com/azure-sdk/internal/_workitems/edit/35909) — GA release plan for ProviderHub.Management, targeting August 2026.

## Additional information

<details>

</details>
<details>

</details>


KPIyer:
Additive and byte compatible.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
```diff
@@ -0,0 +1,78 @@
...
+   */
+  get is ManifestOps.Read<ManifestInfo>;
+
+  /**
+   * Creates or Updates a manifest in manifest repository.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "The service returns only 200 for this operation; changing it would break existing clients."
+  createOrUpdate is ManifestOps.CreateOrUpdateSync<
+    ManifestInfo,
+    Request = ManifestPutContent,
```
psah434: **[NEW] 🔴 Blocking** **[[RPC-Put-V1-25](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `specification/providerhub/ProviderHub.Management/ManifestInfo.tsp` - line 68 - New manifest PUT accepts `ManifestPutContent` but returns `ManifestInfo` (emitted swagger lines 1419/1427); PUT request/response MUST use the same model.

**Classification reasoning:** Both manifest operations are absent from base `stable/2024-09-01`; the new readme suppression's “pre-existing” claim is false.

**Suggested fix:** Use `ManifestInfo` for request and response, update implementation/examples, regenerate, and only then remove the scoped suppression.

```tsp
createOrUpdate is ManifestOps.CreateOrUpdateSync<ManifestInfo>;
```

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
```diff
@@ -0,0 +1,78 @@
...
+interface Manifests {
+  /**
+   * Gets the manifest from the manifest repository.
+   */
+  get is ManifestOps.Read<ManifestInfo>;
+
+  /**
+   * Creates or Updates a manifest in manifest repository.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-put-operation-response-codes" "The service returns only 200 for this operation; changing it would break existing clients."
```
psah434: **[NEW] 🔴 Blocking** **[[RPC-Put-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#31-put-for-resource-creation)]** `specification/providerhub/ProviderHub.Management/ManifestInfo.tsp` - line 65 - New `Manifests_CreateOrUpdate` suppresses standard responses and emits only `200`; resource PUT MUST define `201` creation and `200` replacement.

**Classification reasoning:** Manifest PUT is absent from base `stable/2024-09-01`, so the suppression's existing-client rationale is unsupported by spec history.

**Suggested fix:** Use standard synchronous create-or-replace responses; remove custom `ArmResponse<ManifestInfo>` and both response-code suppressions after regeneration.

```tsp
createOrUpdate is ManifestOps.CreateOrUpdateSync<ManifestInfo>;
```

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
```diff
@@ -0,0 +1,78 @@
...
+        @path
+        @segment("manifests")
+        @key
+        environment: string,
+      }
+    > {}
+
+@added(Versions.v2025_10_01)
+@armResourceOperations
+interface Manifests {
```
psah434, line 56: **[NEW] 🔴 Blocking** **[[PLCY008](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/policy-compatibility.md#plcy008-collection-get-and-point-get-for-child-resources)]** `specification/providerhub/ProviderHub.Management/ManifestInfo.tsp` - line 56 - New `manifests` child exposes point GET and PUT only; every nested resource MUST expose point and collection GET for Policy enumeration.

**Classification reasoning:** `manifests/{environment}` and both operation IDs are absent from base `stable/2024-09-01`; the new suppression does not justify missing Policy enumeration.

**Suggested fix:** Add collection GET at `/providerRegistrations/{providerNamespace}/manifests`, a paged response and example, regenerate, then remove the suppressions.

```tsp
listByProviderRegistration is ManifestOps.List<ManifestInfo>;
```

psah434 (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`eaa8fef`](https://github.com/Azure/azure-rest-api-specs/pull/45194/commits/eaa8fef6efc6508d0e0c62ae253fc182614330b4). See inline comments for findings 1-3.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
wwendyc: Updated to use ManifestInfo for both request and response which should reflect what's in the code.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
wwendyc: Only a single manifest can be generated per provider so it doesn't really make sense to implement a LIST route.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
wwendyc: The PUT manifest route is a proxy for the ARM checkin manifest operation which never returns 201. https://msazure.visualstudio.com/One/_git/AzureUX-ARM?path=/src/frontdoor/Roles/Frontdoor.Admin/Controllers/ResourceProviderOperationController.cs&version=GBmaster&line=5208&lineEnd=5208&lineStartColumn=48&lineEndColumn=64&lineStyle=plain&_a=contents

https://armprodeus.eastus.kusto.windows.net/Requests?query=H4sIAAAAAAAEAF2NsQ6CMBgGdxLe4ZcJBgcSV1xYZFCJ8AK1fNJG22L7I9H48OKiCfPd5XbMQ2WlM9r2J9xHBA7RmyYFD6o9pA5otUHDwgy0Fb1L802XxT%2fHDfCCtbMHYUCwXZg0K0oCbpcG%2fqElSgV5rWzyj9R83YOV66goKKmPTbug84%2fHULoOtCponX9pGI0RXr9A0o2W04zOz4UcR0QfX9Zvo9MAAAA%3d&web=0

HttpIncomingRequests
| where PreciseTimeStamp >ago(14d)
| where operationName endswith "selfServiceCheckIn"
| where httpMethod == "POST"
| where httpStatusCode != -1
| summarize count() by httpStatusCode
  
httpStatusCode | count_
-- | --
401 | 3
500 | 20
400 | 355
200 | 1465

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
psah434: Thanks for addressing this! The request/response mismatch is no longer present — the emitted PUT `.../manifests/{environment}` now uses `ManifestInfo` for both the request body and the 200 response (verified at head `1be9e8c`). Resolving this thread.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
psah434: Thanks for the detailed justification and the production telemetry — that's helpful. Re-reviewing at head `1be9e8c`: the emitted PUT still returns only `200` (+ `default`), and I accept that the checkin operation genuinely never returns `201`.

Given that, I'm **downgrading this from Blocking to a Warning / design discussion** rather than insisting on a `201`. The more useful question this surfaces: if this PUT never creates a resource (always `200`, no `201`, no distinct create semantics), it behaves like an action, not a resource lifecycle operation. Modeling it as a POST action (as the spec already does for `checkinManifest`) would be a cleaner fit and would also make the `PutResponseCodes` suppression unnecessary. Since this is a **stable/GA** version, this modeling choice is worth an explicit call from the ARM API reviewers before it locks in. Leaving this thread open for that decision — no change required from you unless the ARM reviewers agree the resource modeling should change.

specification/providerhub/ProviderHub.Management/ManifestInfo.tsp
psah434: Thanks — noted that a single manifest is generated per provider today. Re-reviewing at head `1be9e8c`: the resource is keyed by `{environment}` (path `.../manifests/{environment}`), so multiple instances are addressable, and PLCY008 requires a collection GET on child resource types (including singletons) so Azure Policy can enumerate them. That said, this is the same root question as the PUT-response thread: whether `manifests` should be an ARM `ProxyResource` at all, or an action. If it stays a resource, a collection GET is expected for GA; if it becomes an action, the `NestedResourcesMustHaveListOperation` suppression goes away. Keeping this as a **Warning** and leaving the thread open for the ARM API reviewers to make the modeling call. No change required from you pending that decision.

wwendyc:
@psah434 the responses don't seem to require additional changes and are marked as warning instead of blocking. Can you please see if we're good to merge the PR?

ravimeda (approved):

ravimeda (approved):

PR #45205, opened 2026-08-03
Fix package detector for non-project TypeSpec configs

### LINKED ISSUE #45204: Package name detector fails for tspconfig directories without main.tsp
## Description

The **Detect package name changes** step fails whenever a PR modifies a `tspconfig.yaml` that is not located at a standalone TypeSpec project root.

The detector currently treats every changed `specification/**/tspconfig.yaml` directory as directly compilable and runs:

```text
npx tsp compile <tspconfig-directory> --emit @azure-tools/typespec-metadata ...
```

This assumes that the directory contains `main.tsp`.

## Reproduction

PR #45199 modifies:

```text
specification/ai/HealthInsights/HealthInsights.RadiologyInsights/tspconfig.yaml
```

That directory contains partial TypeSpec files which are imported by:

```text
specification/ai/HealthInsights/HealthInsights.OpenAPI/main.tsp
```

It intentionally does not contain its own `main.tsp`. The detector consequently fails with:

```text
error file-not-found: File .../HealthInsights.RadiologyInsights/main.tsp not found.
```

Failing job: https://github.com/Azure/azure-rest-api-specs/actions/runs/30841124128/job/91778131276

## Expected behavior

The package-name detector should support changed `tspconfig.yaml` files that are not standalone project roots, or skip them with an explicit diagnostic when no valid entrypoint can be resolved. A package-name-neutral configuration cleanup should not fail the overall Analyze Code job.

## Suggested fix

Update `.github/workflows/src/package-name-approval/detect-namespaces.js` to avoid blindly compiling the config directory. Resolve the associated TypeSpec entrypoint where possible, otherwise skip non-project configs safely. Add a test covering a changed `tspconfig.yaml` whose directory has no `main.tsp`.

Detector source: https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/package-name-approval/detect-namespaces.js

JoshLove-msft:
## Summary

- prefer `main.tsp` as the package detector entrypoint
- fall back to `client.tsp` for SDK-only TypeSpec configurations such as Health Insights Radiology
- skip with an explicit warning only when neither supported entrypoint exists
- add regression coverage for both the `client.tsp` fallback and final skip path

Fixes #45204.

## Validation

- package-name detector test suite: 11 passed
- JavaScript syntax checks passed
- Prettier and `git diff --check` passed

### DIFF
--- a/.github/workflows/src/package-name-approval/detect-namespaces.js
+++ b/.github/workflows/src/package-name-approval/detect-namespaces.js
@@ -236,6 +236,13 @@ export default async function detectNamespaces({ context, core }) {
 
   for (const { file, basePath } of changedTspconfigs) {
     const tspConfigDir = dirname(join(cwd, file));
+    if (!existsSync(join(tspConfigDir, "main.tsp"))) {
+      core.warning(
+        `Skipping package name detection for ${file}: no main.tsp found in its directory`,
+      );
+      continue;
+    }
+
     core.info(`Running typespec-metadata emitter for: ${file}`);
 
     const prResult = await runMetadataEmitter(tspConfigDir, core);
--- a/.github/workflows/src/package-name-approval/detect-namespaces.js
+++ b/.github/workflows/src/package-name-approval/detect-namespaces.js
@@ -64,10 +64,11 @@ const METADATA_LANG_MAP = {
  * package names and namespaces for all configured languages.
  *
  * @param {string} tspConfigDir - Absolute path to the directory containing tspconfig.yaml
+ * @param {string} entrypoint - Absolute path to the TypeSpec entrypoint
  * @param {import("@actions/core")} core
  * @returns {Promise<EmitterResult>}
  */
-async function runMetadataEmitter(tspConfigDir, core) {
+async function runMetadataEmitter(tspConfigDir, entrypoint, core) {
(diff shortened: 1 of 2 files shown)


samvaity:
The failing job ([run #30841124128](https://github.com/Azure/azure-rest-api-specs/actions/runs/30841124128/job/91778131276)) is hitting exactly this issue - `tsp compile` fails with `file-not-found` on `specification/ai/HealthInsights/HealthInsights.RadiologyInsights/main.tsp` because that project uses `client.tsp` as the entrypoint instead of `main.tsp`.

I thought I had added this, but rather than skipping these projects entirely, we could fall back to `client.tsp` - which is what we already do in the azsdk-cli's TypeSpecHelper. See Praveen's fix in Azure/azure-sdk-tools#16475:

```csharp
if (File.Exists(Path.Combine(project.ProjectRootPath, "main.tsp")))
    entrypoint = ".";
else if (File.Exists(Path.Combine(project.ProjectRootPath, "client.tsp")))
    entrypoint = "./client.tsp";
```

Skipping means we silently miss package name changes for any project that uses `client.tsp` only. If we fall back to `client.tsp` the emitter still runs and we get proper coverage.

samvaity (approved):

PR #45218, opened 2026-08-04
Validate service.yaml manifests in TypeSpec Validation

timotheeguerin:
`service.yaml` is becoming the machine-readable source of truth for a service's API versions and the swagger files they produce (Azure/azure-rest-api-specs#44805, [typespec-azure#4846](https://github.com/Azure/typespec-azure/pull/4846)). Nothing keeps those 560+ new manifests honest, so today two things can rot silently:

- a new TypeSpec project ships with no `service.yaml` at all, quietly regressing the rollout;
- a manifest points at a swagger file that was renamed, moved, or never generated. The autorest emitter only rewrites the versions it emits and never touches legacy `source: swagger` entries, so a stale path there is invisible.

This adds a `ServiceYaml` rule to `tsv` that closes both gaps.

## It already caught one

Rebasing onto the current head of #44805 surfaced `specification/applicationinsights/data-plane/Exporters` — a project that landed via a `main` merge *after* the manifests were generated, and therefore had none. That is precisely the drift this rule exists to prevent, on its first run. Its manifest is included here.

## What it reports

A missing manifest:

```
Missing service.yaml at specification/contoso/Contoso.Management/service.yaml.

Every TypeSpec project emitting "@azure-tools/typespec-autorest" must declare its API
versions in a service.yaml next to tspconfig.yaml:

versions:
  - version: 2024-06-01
    source: typespec
    swagger-files:
      - resource-manager/Contoso/stable/2024-06-01/openapi.json

Create an empty service.yaml and run "tsp compile ." to have the autorest emitter fill it in,
then add any legacy swagger-only versions by hand.
```

And a manifest that points at nothing — all broken paths at once, with fix guidance that differs by `source`:

```
specification/ai/Face/service.yaml references swagger files that do not exist
(paths are relative to service.yaml, and must match on case):

  - version "v1.2": ../data-plane/Face/stable/v1.2/FaceMissing.json

For "source: typespec" versions, run "tsp compile ." to regenerate the swagger and update
service.yaml. For "source: swagger" versions, correct the path by hand or remove the version
if it no longer exists.
```

Path checks go through the repo's exact-case `fileExists`, so case-only drift fails on macOS and Windows too rather than passing locally and breaking in CI.

## Scope

The rule requires a manifest only from projects that emit `@azure-tools/typespec-autorest` **and** have a `main.tsp` — the same "is this a real, compilable service" gate `CompileRule` and `EmitAutorestRule` already use. That makes the multi-service aggregate wrappers (ResourceGraph, `Microsoft.Insights/Insights`, HealthInsights.RadiologyInsights) skip on their own, with no suppressions to maintain. Swagger paths are still validated wherever a `service.yaml` exists, so nothing is silently exempted.

Content validation stays deliberately thin — the emitter already checks the shape of what it writes against its own schema, so this only parses enough to read `swagger-files` and reports a readable error if the YAML can't be understood.

## Drive-by fix

`parse()` in `tsp-config.ts` threw a raw `ZodError` on a `tspconfig.yaml` with an empty `emit:` key (two real projects under `ai-foundry` hit this). `EmitAutorestRule` shares that code path and would crash the same way.

## Validation

Ran the rule across all 589 TypeSpec projects: **563 passed, 26 skipped, 0 failed**. Verified it actually fires by breaking a path in `specification/ai/Face/service.yaml`.

Stacked on #44805 — retarget to `main` once that merges.


timotheeguerin:
Superseded by Azure/azure-rest-api-specs#45301, which is branched cleanly off `main` now that #44805 has merged, and fixes the two real cases of manifest drift the rule surfaced.

PR #45237, opened 2026-08-04
[wip] Improve suppressions.yaml support for swagger checks

catalinaperalta:
This PR uses the common set-status flow to see if swagger checks have been suppressed through suppressions.yaml. Also note that it supports a `SwaggerAll` value to skip all swagger checks for services that do not use swagger, such as Foundry.


catalinaperalta:
will implement in individual checks to stay inline with existing patterns

PR #45245, opened 2026-08-05
feat(cognitiveservices): add BYOC agent hosting configuration

frankliu20:

# Choose a PR Template

### DIFF
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-07-15-preview/CreateAccountWithAgentHostingConfiguration.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-07-15-preview/CreateAccountWithAgentHostingConfiguration.json
@@ -1,0 +1,152 @@
+{
+  "operationId": "Accounts_Create",
+  "parameters": {
+    "account": {
+      "identity": {
+        "type": "UserAssigned",
+        "userAssignedIdentities": {
+          "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/myResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/account-control-plane": {}
+        }
+      },
+      "kind": "AIServices",
+      "location": "West US",
+      "properties": {
+        "agentHostingConfigurations": [
+          {
+            "name": "default",
+            "hostingType": "ManagedCluster",
+            "hostingManagementIdentityResourceId": "/subscriptions/00000000-1111-2222-3333-444444444444/resourceGroups/myResourceGroup/providers/Microsoft.ManagedIdentity/userAssignedIdentities/account-control-plane",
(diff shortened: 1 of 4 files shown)


fmabroukmsft:
Two design points worth resolving before this hardens toward GA, since both become breaking changes to fix later:

### 1. Documented immutability isn't enforced in the contract
The doc comment on `agentHostingConfigurations` says *"Once set, the configuration is immutable; subsequent requests may only repeat the same value."* However, `AccountProperties` is shared by both PUT (`Accounts_Create`) and PATCH (`Accounts_Update` uses `#/definitions/Account`), and the new property has no visibility decorator — so the emitted contract makes it writable on update, with no `readOnly` / `x-ms-mutability`. A client can therefore PATCH it, which contradicts the documented behavior and leaves enforcement entirely runtime-side.

`AccountProperties` already uses `@visibility(Lifecycle.Read)` and `@visibility(Lifecycle.Create, Lifecycle.Update)` elsewhere, so the idiomatic fix fits the existing convention:

```tsp
@added(Versions.v2026_07_15_preview)
@visibility(Lifecycle.Read, Lifecycle.Create)   // emits x-ms-mutability: ["read","create"]
@minItems(1)
@maxItems(1)
@identifiers(#["name"])
agentHostingConfigurations?: AgentHostingConfiguration[];
```

This keeps the property out of the PATCH surface, signals immutability to SDKs/clients, and matches the doc comment.

### 2. Extensible `hostingType` vs. hard-required AKS-specific fields
`AgentHostingType` is an extensible union (`string` variant, `modelAsString: true`), which signals intent to add more hosting types later. But `clusterResourceId`, `clusterAgentPoolName`, and `storageAccountResourceId` are AKS-specific and marked `required` on the flat `AgentHostingConfiguration` base. When a second hosting type is introduced, those required AKS fields will be wrong for it, forcing a breaking remodel (or making them optional, which is a semantic breaking change).

The idiomatic ARM shape is a discriminated union — `@discriminator("hostingType")` on a base config, with a `ManagedCluster`-derived model carrying the AKS-only fields:

```tsp
@added(Versions.v2026_07_15_preview)
@discriminator("hostingType")
model AgentHostingConfiguration {
  name: string;
  hostingType: AgentHostingType;
}

@added(Versions.v2026_07_15_preview)
model ManagedClusterAgentHostingConfiguration extends AgentHostingConfiguration {
  hostingType: AgentHostingType.ManagedCluster;
  hostingManagementIdentityResourceId: ...;
  workloadIdentityResourceId: ...;
  clusterResourceId: ...;
  clusterAgentPoolName: string;
  storageAccountResourceId: ...;
}
```

Preview gives you runway to defer, but the current extensible-enum + required-type-specific-fields combination is internally inconsistent. If AKS is the *only* hosting type ever planned, the simpler alternative is to make `hostingType` a fixed (non-extensible) enum so the flat model is honest about its scope.

Neither blocks merge given preview flexibility, but resolving #1 (one-line, follows existing convention) and getting an explicit decision on #2 now avoids a breaking change down the line.

fmabroukmsft (approved):

bowen5 (approved):

PR #45247, opened 2026-08-05
Azure Content Understanding 2026-06-01-preview

yungshinlintw:
Adds the first post-GA preview API version for Azure AI Content Understanding (`2026-06-01-preview`). Changes are additive in the new preview version. GA `2025-11-01` is unchanged.

## Preview additions

- **Inline analysis** (`analyzeInline` / `analyzeBinaryInline`) — Returns the analysis result in the HTTP 200 JSON body without an LRO, polling, or persisted result. Use this for lightweight, small-document analysis (for example layout/read-style analyzers without field extraction) where a single round-trip is enough.

- **In-page segmentation** (`allowInPageSegments`, segment `confidence` / `source`) — Lets segments cover part of a page instead of whole pages, with classification confidence and a source range that can be passed to a sub-analyzer. Use this when one page contains multiple logical items (receipts, ID cards, mixed forms) that should be split and routed separately.

- **Signature detection** (`DocumentSignature`, `DocumentContent.signatures`) — Returns detected signatures as layout elements (id, source, span, optional role) when layout details are enabled. Use this to locate and ground handwritten or stamped signatures in contracts and forms.

- **Document metadata** (`AnalysisContent.metadata`) — Extracts string key/value metadata from the input, such as author, creation date, and title (only keys with values are returned). Use this to populate document-library or search index fields without a custom field schema.

- **Agentic workflow** (`ContentAnalyzerConfig.workflow`) — Selects the analysis workflow, including `agentic` iterative reasoning with tool calling. Use this for complex documents where the default workflow is not accurate enough and multi-step reasoning is needed.

- **Chunking** (`chunkingStrategy`, `DocumentContent.chunks`, chunk `source`) — Splits markdown into semantically meaningful, size-controlled chunks for RAG (optional `maxTokens` soft limit). Use this to index or retrieve document passages with grounding back to the original content.

- **Partial / truncated input** (`allowInputTruncation`) — When true, over-limit input is truncated and returned as a partial result with a warning instead of failing (default remains fail, matching GA). Override per request when you would rather process the leading pages/time range of a large file than error out.

- **Diagnostic infos** (`AnalysisResult.infos`) — Additional diagnostic information about the analysis. Examples include `LLMStats` with completion and embedding call counts and latency, so callers can inspect LLM usage for a run.

- **Usage meters** (`advancedContextualizationTokens`, `documentPages*Inline`) — Reports advanced contextualization token usage and page counts specific to inline analyze. Use this to attribute billing/usage for advanced or agentic workflows and for inline vs async analysis.


specification/ai/ContentUnderstanding/models.tsp
```diff
@@ -667,6 +604,10 @@ model AnalysisContent {
 
   @doc("Extracted fields from the content.")
   fields?: Record<ContentField>;
+
+  @added(Versions.v2026_06_01_preview)
+  @doc("Document metadata extracted from the input, such as author, creation date, and title. Keys and values are strings. Only keys with extracted values are present.")
+  metadata?: Record<string>;
```
weidongxu-microsoft, line 610: This property is declared on the `AnalysisContent` base model, so the generated contract and all SDKs also expose the **document** metadata property on `AudioVisualContent`. The PR description and property documentation describe a document-only feature. Please move it to `DocumentContent`, or clarify and rename the contract if audio/video metadata is intentionally supported.

specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_SignatureDetection.json
```diff
@@ -0,0 +1,61 @@
...
+                  "width": 8.5,
+                  "height": 11
+                }
+              ],
+              "signatures": [
+                {
+                  "id": "signature-1",
+                  "source": "D(1,2.5,8,5.5,8,5.5,9,2.5,9)",
+                  "span": {
+                    "offset": 72,
```
weidongxu-microsoft: The grounding span does not identify the signature token in the example markdown: offset 72 with length 13 resolves to `** [signature`, while `[signature-1]` starts at offset 75. Please correct the span so this new signature-grounding example is internally consistent.

specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_Chunking.json
```diff
@@ -0,0 +1,78 @@
+{
+  "title": "Get Analysis Result with Chunks",
+  "operationId": "ContentAnalyzers_GetResult",
+  "parameters": {
+    "api-version": "2026-06-01-preview",
+    "analyzerId": "myRagAnalyzer",
```
weidongxu-microsoft: `ContentAnalyzers_GetResult` is routed only by `operationId` (plus `api-version`); it has no `analyzerId` parameter. Please remove this unsupported example parameter so generated documentation does not show a request shape clients cannot send.

weidongxu-microsoft (dismissed):
The new preview additions are generally additive, but the existing analyze operations currently change their generated parameter order, and several new contract/example details need correction before publication.

specification/ai/ContentUnderstanding/models.tsp
yungshinlintw: This is intended for multi-modal files properties too, so will update @doc. Good catch.

yungshinlintw: Thanks — this is intentional. `metadata` stays on `AnalysisContent` so document and audio/visual contents can return extracted file properties (for example author, title, creation date, or media properties). Updated the `@doc` to match that intent.

Fix: https://github.com/Azure/azure-rest-api-specs/commit/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e
- TypeSpec: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/ContentUnderstanding/models.tsp#L608-L610
- Swagger: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/data-plane/ContentUnderstanding/preview/2026-06-01-preview/ContentUnderstanding.json#L1184-L1190

specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_SignatureDetection.json
yungshinlintw: Agreed — the previous span did not match the markdown token.

Signatures are grounded as markdown image references (``), not `[signature-1]`. Updated the example to a short single-signature PDF result:

- markdown containing `John Smith`
- `id`: `1.1`
- `span`: offset `81`, length `29` (exactly that image reference)

`role` is omitted because it is not present on the payload.

Fix: https://github.com/Azure/azure-rest-api-specs/commit/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e
- TypeSpec example: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_SignatureDetection.json
- Data-plane example: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/data-plane/ContentUnderstanding/preview/2026-06-01-preview/examples/ContentAnalyzers_GetResult_SignatureDetection.json

specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_Chunking.json
yungshinlintw: Agreed. `ContentAnalyzers_GetResult` is addressed only by `operationId` and `api-version`.

`analyzerId` belongs on the analyze request and on the result body (`result.analyzerId`), not on GetResult parameters. Chunking is configured on the analyzer (`chunkingStrategy`) and returned on `contents[].chunks`.

Removed `parameters.analyzerId` from this example and kept it on the response.

Fix: https://github.com/Azure/azure-rest-api-specs/commit/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e
- TypeSpec example: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/ContentUnderstanding/examples/2026-06-01-preview/ContentAnalyzers_GetResult_Chunking.json
- Data-plane example: https://github.com/Azure/azure-rest-api-specs/blob/a7cf82b558f8aaf655dbab34c057dadf7ae2ca6e/specification/ai/data-plane/ContentUnderstanding/preview/2026-06-01-preview/examples/ContentAnalyzers_GetResult_Chunking.json

yungshinlintw:
@weidongxu-microsoft thanks for taking a look at the cross-version diff as well.

[Breaking Change(Cross-Version)](https://github.com/Azure/azure-rest-api-specs/actions/runs/30971831104) reports **5 Errors**, all [1042 ChangedParameterOrder](https://github.com/Azure/openapi-diff/blob/main/docs/rules/1042.md), comparing `2026-06-01-preview` to GA `2025-11-01`:

- `POST /analyzers/{analyzerId}:analyze` — `x-ms-client-request-id` 4→5, `body` 5→6
- `POST /analyzers/{analyzerId}:analyzeBinary` — `range` 4→5, `x-ms-client-request-id` 5→6, `input` 6→7

Preview adds optional query `allowInputTruncation`. In Swagger 2.0 that query sits in the same `parameters` array as later query/header/body params, so indexes shift. Clients bind by name and location, not array index. Callers that omit the new query send the same request as before. Same-version breaking-change is clean; preview-vs-previous-preview is also clean.

The same run has **2 Warnings (1038 AddedPath)** for new preview-only `:analyzeInline` / `:analyzeBinaryInline`. Those are additive new routes, not changes to GA operations.

Where SDK binary compatibility mattered, the new optional query is absorbed additively (existing convenience overloads unchanged; new options go on an options bag / request options). We are following [aka.ms/brch](https://aka.ms/brch) for the `BreakingChangeReviewRequired` label.

Request: please treat 1042 here as a Swagger 2.0 parameter-array ordering artifact for an additive optional query on a new preview version, not as a customer-facing REST or SDK break.

specification/ai/ContentUnderstanding/documentContent.tsp
```diff
@@ -595,6 +603,25 @@ model DocumentHyperlink {
   source?: SourceExpression;
 }
 
+@added(Versions.v2026_06_01_preview)
+@doc("Signature detected in a document.")
+model DocumentSignature {
+  @doc("Signature identifier.")
+  id: string;
```
weidongxu-microsoft, line 610: Are these read-only (if not, what user set)? Or the whole `DocumentContent` is ready-only or only used in response?

yungshinlintw: `DocumentContent` is only used in a reponse.

specification/ai/ContentUnderstanding/models.tsp
```diff
@@ -847,6 +813,19 @@ model ContentAnalyzerAnalyzeOperationStatus {
   usage?: UsageDetails;
 }
 
+@added(Versions.v2026_06_01_preview)
+@doc("Provides inline response details for analyze operations.")
+model ContentAnalyzerInlineResponse {
+  @doc("The status of the inline analyze request.")
+  status: Azure.Core.Foundations.OperationState;
```
weidongxu-microsoft, line 820: The `status` seems even include `NotStarted` and `Running`. I assume this is impossible for an inline response?

Maybe it is OK for consistency. Just asking.

specification/ai/ContentUnderstanding/models.tsp
```diff
@@ -847,6 +813,19 @@ model ContentAnalyzerAnalyzeOperationStatus {
...
+
+  @doc("The result of the inline analyze request.")
+  result: AnalysisResult;
```
weidongxu-microsoft, line 823: Does `result` always exists? Even on `Canceled` or `Failed` status?

weidongxu-microsoft:
Could you confirm whether these example values are intentional?

- In `ContentAnalyzers_AnalyzeBinaryInline_Partial.json`, the warning targets `inputs/0`, but the binary operation accepts a single `input` body rather than an `inputs` array. Should the target be `input`?
- In `ContentAnalyzers_GetResult_Segmentation.json`, the receipt span (`offset: 56`, `length: 45`) starts before `# Receipt` and stops before `$45.99`, while the letter span (`offset: 107`, `length: 50`) starts in the preceding separator. Should these spans instead align exactly with their corresponding markdown sections?
- In `ContentAnalyzers_GetResult_Chunking.json`, the final chunk (`offset: 137`, `length: 68`) excludes the last period of the markdown. Is that deliberate, or should the length include the complete section?

specification/ai/ContentUnderstanding/models.tsp
yungshinlintw: Yes, this is just for consistency as this is what we have for LRO (status/usage/result). I do have a follow up discussion with service team to see if we will ever need an error condition. Most of the error condition will be returning 4xx (such as invalid pdf, invalid URL, etc.) I cannot yet identify a case that we cannot use 4xx.

weidongxu-microsoft:
> @weidongxu-microsoft thanks for taking a look at the cross-version diff as well.
> 
> [Breaking Change(Cross-Version)](https://github.com/Azure/azure-rest-api-specs/actions/runs/30971831104) reports **5 Errors**, all [1042 ChangedParameterOrder](https://github.com/Azure/openapi-diff/blob/main/docs/rules/1042.md), comparing `2026-06-01-preview` to GA `2025-11-01`:
> 
> * `POST /analyzers/{analyzerId}:analyze` — `x-ms-client-request-id` 4→5, `body` 5→6
> * `POST /analyzers/{analyzerId}:analyzeBinary` — `range` 4→5, `x-ms-client-request-id` 5→6, `input` 6→7
> 
> Preview adds optional query `allowInputTruncation`. In Swagger 2.0 that query sits in the same `parameters` array as later query/header/body params, so indexes shift. Clients bind by name and location, not array index. Callers that omit the new query send the same request as before. Same-version breaking-change is clean; preview-vs-previous-preview is also clean.
> 
> The same run has **2 Warnings (1038 AddedPath)** for new preview-only `:analyzeInline` / `:analyzeBinaryInline`. Those are additive new routes, not changes to GA operations.
> 
> Where SDK binary compatibility mattered, the new optional query is absorbed additively (existing convenience overloads unchanged; new options go on an options bag / request options). We are following [aka.ms/brch](https://aka.ms/brch) for the `BreakingChangeReviewRequired` label.
> 
> Request: please treat 1042 here as a Swagger 2.0 parameter-array ordering artifact for an additive optional query on a new preview version, not as a customer-facing REST or SDK break.

I think you can add the `-benign` label before merge (only if this is the only item raised by "Cross-Version" check), as you've customized these methods in client.tsp

specification/ai/ContentUnderstanding/models.tsp
yungshinlintw: Today inline analyze only returns HTTP 200 with status: Succeeded and a populated result. Failures use standard HTTP error responses, not a Failed status in the body. So result is intentionally required on the success envelope.

LRO differs: ContentAnalyzerAnalyzeOperationStatus uses OperationStatus<AnalysisResult> where result is optional and error is present on Failed.

If we ever return Failed/Canceled in the inline 200 body, we would add error and make result optional together. Making result optional later would be backward compatible on the wire; we can do that when we have the failure shape to spec.

weidongxu-microsoft: Making result optional may be a breaking change to user code. Now they may use it without check for `undefined` or `null`, but later (if it becomes optional) they need to modify the code to do such check.

image

Our source here https://eng.ms/docs/products/azure-developer-experience/design/specs-pr-guides/pr-brch-deep

weidongxu-microsoft: As preview, it is OK to break later.

Just when you have a chance, think about it.

IMHO, if we unlikely to add `status` and `error` to inline request/response, we may just do not add them for now.

yungshinlintw:
Thanks for flagging these — none of the span misalignments were intentional. Addressed in e426990a71f.

**`AnalyzeBinaryInline_Partial.json` — warning `target`:** Updated to `"input"` for binary operations. URL / `inputs`-based operations (`:analyze`, `:analyzeInline`, and LRO GetResult examples from those paths) keep `"inputs/0"`. 

**Segmentation spans:** Fixed receipt to `(57, 50)` and letter to `(114, 46)` so each span aligns with its markdown section (receipt now includes `$45.99`; letter no longer includes the preceding `---`).

**Chunking final chunk:** Fixed length `68` → `69` so the Deployment section includes the full trailing ellipsis.

**Additional span audit:** We checked all 21 span entries across CU examples. One more fix: `GetResult.json` `Company` field span `(7, 2)` → `(2, 7)` so the slice matches `valueString: "CONTOSO"` on `"# CONTOSO\n\n..."` (same issue across GA/preview copies). Segmentation, chunking, and signature examples were already correct after the above fixes.

specification/ai/ContentUnderstanding/models.tsp
yungshinlintw: Agreed on the SDK ergonomics concern — we'll keep `result` required on inline HTTP 200 success for now. Failures continue to use standard HTTP error responses rather than a Failed body without `result`. If we ever add Failed-in-body for inline, we'd introduce `error` and make `result` optional together in a deliberate contract change.

weidongxu-microsoft (approved):
One concern here https://github.com/Azure/azure-rest-api-specs/pull/45247#discussion_r3727356352

Not a blocker for preview. But if for next stable please design carefully.

weidongxu-microsoft (approved):

weidongxu-microsoft (approved):

PR #45255, opened 2026-08-05
Add PON support to VMSS SDK models

rishiraja76:
## Summary

Implements [Azure DevOps work item 38120144](https://msazure.visualstudio.com/One/_workitems/edit/38120144) by adding optional `privateIPAddressPrefixLength` support to the generated Compute SDK models for:

- `VirtualMachineScaleSetIPConfigurationProperties` (VMSS create)
- `VirtualMachineScaleSetUpdateIPConfigurationProperties` (VMSS update)
- `VirtualMachineNetworkInterfaceIPConfigurationProperties` (VM)

The property is gated with `@added(Versions.v2026_07_02)`. This change also adds the `2026-07-02` version wiring, representative source and generated examples, README package configuration, and compiler-generated `ComputeRP.json` updates.

## Validation

- `azsdk_run_typespec_validation` succeeded
- `npx tsp compile .` exited 0 with `Compilation completed successfully.`


rishiraja76:
image

rishiraja76:
Two open questions for the service team before this merges. Raising both here so they are visible to reviewers — neither is blocking review, but the second may be blocking correctness.

## 1. Should `privateIPAddressPrefixLength` be on the VMSS update (PATCH) model?

This PR adds the property to three models (`models.tsp`):

| Line | Model | Path |
|---|---|---|
| 3096 | `VirtualMachineScaleSetIPConfigurationProperties` | VMSS create |
| 4432 | `VirtualMachineScaleSetUpdateIPConfigurationProperties` | **VMSS PATCH** |
| 6245 | `VirtualMachineNetworkInterfaceIPConfigurationProperties` | VM create |

Microsoft.Network exposes this property on **create only** — it lives on `NetworkInterfaceIPConfigurationPropertiesFormat`, reachable via `NetworkInterfaces_CreateOrUpdate` (PUT). Network's only NIC PATCH is `NetworkInterfaces_UpdateTags`, which does not carry `ipConfigurations`. Unchanged since 2023-11-01 (introduced in #28527) across 8 stable versions.

So the VMSS update model has no Network counterpart. Either:

- **Intentional** — Compute supports changing prefix length on an existing VMSS, a deliberate superset of Network. Fine, but the `#suppress` reason currently justifies `x-nullable` by citing "Microsoft.Network parity", which does not hold for this model. The comment should stand on its own terms.
- **Not intentional** — we are publishing a writable property on PATCH that the RP ignores, which is worse than omitting it since removing it later is spec-breaking.

Can someone confirm whether changing prefix length on an existing VMSS is supported? Happy to drop the update-model change if not.

## 2. `networkApiVersion` is capped at `2022-11-01`, before the property existed

`stable/2026-09-01/ComputeRP.json` line 17458:

```json
"NetworkApiVersion": {
  "type": "string",
  "description": "specifies the Microsoft.Network API version used when creating networking resources in the Network Interface Configurations",
  "enum": [ "2020-11-01", "2022-11-01" ]
}
```

`privateIPAddressPrefixLength` first appears in Microsoft.Network **2023-11-01** — after both allowed values. Where this is consumed (same file):

| Line | Model | Scope |
|---|---|---|
| 17585 | `NetworkProfile` | VM |
| 23393 | `VirtualMachineScaleSetNetworkProfile` | VMSS — **Flexible orchestration only** |
| 24057 | `VirtualMachineScaleSetUpdateNetworkProfile` | VMSS update — **Flexible orchestration only** |

The question: when Compute creates NICs on the caller's behalf from inline `networkInterfaceConfigurations`, does the prefix length reach Microsoft.Network, given the pinned version predates the property? If not, the value would be accepted at the Compute contract and silently dropped.

Scope is narrower than it first looks:

| Path | Affected |
|---|---|
| VMSS Uniform | No — `networkApiVersion` is documented as Flexible-only |
| VMSS Flexible | **Yes** |
| VM with inline `networkInterfaceConfigurations` | **Yes** |
| VM referencing pre-created `networkInterfaces` | No — NIC created directly via Network |

`networkApiVersion` is optional (not in any `required` array), so the deciding factor is **what Compute RP uses when the caller omits it**. If it defaults to a modern internal version this is a non-issue. That is not visible from the spec.

If the cap does apply, options are to extend the `NetworkApiVersion` enum to include a version >= 2023-11-01, or to confirm the RP handles this property outside the pinned-version path.

## 3. Note on `maximum: 128`

Copied from Network, where it covers IPv6's 128-bit space. If these Compute paths are IPv4-only in practice the effective ceiling is 32. Kept at 128 for parity, but flagging since it is permanent once 2026-09-01 ships stable.

rishiraja76:
## CI status on `3d24626d`

**2 failing, 29 passing, 8 skipped.**

| Check | Result | Caused by this PR? |
|---|---|---|
| TypeSpec Validation | ❌ failure | **No** — pre-existing on the base branch |
| Breaking Change (Cross-Version) | ❌ failure | **Partly** — see below |
| SDK Validation — .NET / Go / Java / JS / Python / Rust | ✅ all pass | — |
| LintDiff, ModelValidation, SemanticValidation, Avocado, BreakingChange, PrettierCheck, SpellCheck, ARM Auto SignOff | ✅ pass | — |

Worth noting all six SDK Validation checks pass, so the property generates cleanly into every target language.

---

### ❌ TypeSpec Validation — pre-existing, not from this PR

The base branch tip (`4763d9a3`, `compute-2026-09-01`) fails this same check on its own, with no PR applied:

```
checks on 4763d9a303952142409c2aac090613093d35107d:
failure   TypeSpec Validation
```

Supporting evidence:

- `Rule Format failed` fires for **every** Compute project — `Bulkactions`, `ComputeDisk`, `ComputeGallery`, `ComputeSku`, `Recommender` — none of which this PR touches.
- Locally, `tsp format --check` passes on the file this PR edits (`✔ 1 formatted`) and on untouched projects such as `ComputeSku` (`✔ 3 formatted`).
- `tsp compile` on the Compute project succeeds locally with **zero warnings**.

Looks like a formatter-version skew between CI and local tooling. It will block anything targeting `compute-2026-09-01`, so it likely needs fixing at the branch level rather than in this PR.

### ❌ Breaking Change (Cross-Version) — mixed, and mostly not from this PR

This check compares `2026-09-01` against `2026-04-01` on `main`. It reports 8 findings, but only 2 belong to this PR, and neither is an error:

**From this PR — `Info` only, non-blocking:**

| Rule | Level | Detail |
|---|---|---|
| 1041 `AddedPropertyInResponse` | Info | new `privateIPAddressPrefixLength` on `VirtualMachineScaleSetIPConfigurationProperties` |
| 1041 `AddedPropertyInResponse` | Info | same on `VirtualMachineNetworkInterfaceIPConfigurationProperties` |

Expected when adding an optional property to a new API version.

**Not from this PR — these are the `Error`s that fail the check:**

| Rule | Level | Detail |
|---|---|---|
| 1033 `RemovedProperty` | Error | `reservationStateInfo` missing from `CapacityReservationInstanceView` |
| 1033 `RemovedProperty` | Error | `reservationStateInfo` missing from `CapacityReservationInstanceViewWithName` |
| 1033 `RemovedProperty` | Error | `minimumCommitmentDays` missing from `ScheduleProfile` |
| 1033 `RemovedProperty` | Error | `modifiableUntil` missing from `ScheduleProfile` |
| 1006 `RemovedDefinition` | Error | `CapacityReservationStateInfo` removed |
| 1006 `RemovedDefinition` | Error | `ReservationState` removed |

All six are capacity-reservation and schedule-profile types this PR never touches. They are properties/definitions present in `2026-04-01` but absent from `2026-09-01` — i.e. they were dropped when `2026-09-01` was created on the base branch.

One caveat on attribution: this check **passes** on the base commit, because cross-version comparison only runs when a PR changes the swagger. So these removals were introduced with the `2026-09-01` file itself and this PR is simply the first change to surface them. They are not caused by the PON property, but they do need a decision:

- If the removals were intentional, they need `BreakingChangeReviewRequired` sign-off.
- If unintentional, `2026-09-01` is missing types it should have inherited — worth fixing on the branch before more changes land.

Flagging for whoever owns the `compute-2026-09-01` branch, since resolving it is outside the scope of this PR.

rishiraja76:
## Correction: both concerns above are resolved — no change needed

Following up on [my earlier comment](https://github.com/Azure/azure-rest-api-specs/pull/45255#issuecomment-5332533838). I investigated both questions against the NRP source and production telemetry, and **both concerns were unfounded**. Details below so reviewers do not re-litigate them.

---

### 1. VMSS update (PATCH) model — keep it, it is correct

I raised this because Microsoft.Network exposes the property on create only. That was the wrong comparison — I checked Network parity without first checking Compute's own convention.

Compare the two Compute models in `stable/2026-09-01/ComputeRP.json`:

| | Properties |
|---|---|
| `VirtualMachineScaleSetIPConfigurationProperties` | `subnet`, `primary`, `publicIPAddressConfiguration`, `privateIPAddressVersion`, `privateIPAddressPrefixLength`, `applicationGatewayBackendAddressPools`, `applicationSecurityGroups`, `loadBalancerBackendAddressPools`, `loadBalancerInboundNatPools` |
| `VirtualMachineScaleSetUpdateIPConfigurationProperties` | *identical, 9 for 9* |

Microsoft.Network has no general-purpose NIC PATCH at all — its only NIC PATCH is `NetworkInterfaces_UpdateTags`, which does not carry `ipConfigurations`. So **all 8 pre-existing properties on the update model already lack a Network counterpart**, including `subnet`, `primary`, and `privateIPAddressVersion`.

Compute's convention is full create/update mirroring, independent of Network's shape. Adding the property to both models follows that convention; **removing it from the update model would make it the only property in the model missing from the update side.** Keeping as-is.

Minor follow-up: the `#suppress` reason currently cites "Microsoft.Network parity". The stronger justification is Compute's create/update mirroring convention. Happy to reword if a reviewer wants, but it does not affect emitted output.

### 2. `networkApiVersion` cap — not applicable

I was concerned that Compute's `networkApiVersion` enum (`2020-11-01`, `2022-11-01`) predates Network `2023-11-01`, where the property was introduced, and that the value might therefore be dropped. Checking the NRP implementation, that is not how the property is handled.

**Responses — never gated.** `VMScaleSetNetworkInterfaceIPConfiguration.cs`:

```csharp
public bool ShouldSerializePrivateIPAddressPrefixLength()
{
    if (this.PrivateIPAddressPrefixLength != null)
    {
        // Always serialize non-null PrivateIPAddressPrefixLength to avoid data loss.
        // This can only possibly break customers using the feature.
        return true;
    }
    // TODO/prefix-on-nic: Determine the correct API version.
    return this.Parent.SerializationType == NrpSerializationType.Complete
           || NrpApiVersion.IsGreaterThanOrEqualTo(this.Parent.SerializingForApiVersion, CrpApiVersion.V20230701);
}
```

A non-null value is returned unconditionally, explicitly bypassing the CRP api-version gate that sibling properties such as `PrivateIPAddressVersion` and `ApplicationSecurityGroups` are subject to.

**Requests:**

| Path | Gate |
|---|---|
| VMSS (`VMScaleSet.cs`) | none — validates prefix support only |
| VM (`VMNetworkInterfaceIPConfiguration.cs`) | `ValidateApiVersionMinimumRequirementToSetProperty(apiVersion, NrpApiVersion.V20231101, ...)` |

The VM gate is on **`NrpApiVersion`** (the NRP request api-version), not Compute's `networkApiVersion` enum — different namespaces. My original concern conflated the two.

**Telemetry.** `FrontendOperationEtwEvent`, 1 day, all regions, filtering non-zero prefix (`PrefixLength:[1-9]`):

| Operation | Count |
|---|---|
| `PutPublicIpPrefixOperation` | 101,491 |
| NIC / VM / VMSS paths | **0** |

`PutPublicIpPrefixOperation` is public IP prefixes, an unrelated feature. Prefix-on-NIC has no production usage yet, consistent with the open `// TODO/prefix-on-nic:` markers in NRP. Nothing to regress.

---

### Two notes that are still worth acting on

**a) `maximum: 128` is looser than NRP enforces.** NRP validates against a discrete config-driven allow-list, not a range (`Validation.cs`):

```csharp
collection = CommonConfigurationSettings.SupportedPrivateIpPrefixLengthsForIpv4;  // or ...ForIpv6
if (collection != null && collection.Contains(privateIpPrefixLength)) return;
throw new ValidationException(NrpErrorCode.PrivateIPAddressPrefixLengthNotSupported, ...);
```

So `1..128` will accept values NRP rejects at runtime. Microsoft.Network has the same mismatch, so keeping 128 stays consistent — flagging it as a known gap rather than an open question.

**b) NRP rejects prefix in three cases not expressible in swagger:**

- `PrimaryIpConfigCannotUsePrefix` — ipconfig is primary
- `NicWithPublicIpCannotUsePrefix` — ipconfig has a public IP
- `NicWithAsgCannotUsePrefix` — ipconfig has application security groups (config-gated)

Worth capturing in the property description or feature docs so callers are not surprised.

---

Sources: `Networking-nrp` at `/src/sources/Frontend/FrontendContract/` (`Internal/VMScaleSetNetworkInterfaceIPConfiguration.cs`, `VMNetworkInterfaceConfiguration/VMNetworkInterfaceIPConfiguration.cs`, `Internal/VMScaleSet.cs`, `Validation.cs`); telemetry from `nrp.kusto.windows.net/mdsnrp`.

One thread I did not pull: which `apiVersion` value is actually passed into the VM-path gate. It is moot for VMSS (no gate) and for responses (always serialized), but noting it rather than implying I checked.

audreyttt (changes requested):
Please add/update examples

audreyttt (approved):

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
```diff
@@ -3086,6 +3086,15 @@ model VirtualMachineScaleSetIPConfigurationProperties {
...
   privateIPAddressVersion?: IPVersion;
 
+  /**
+   * The private IP address prefix length. If specified and the allocation method is dynamic, the service will allocate a CIDR block instead of a single IP address.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-nullable" "Matches Microsoft.Network's privateIPAddressPrefixLength, which is nullable so callers can explicitly clear an allocated prefix."
+  @added(Versions.v2026_09_01)
+  @minValue(1)
+  @maxValue(128)
+  privateIPAddressPrefixLength?: int32 | null;
```
sandipsh: **[NEW] 🟠 Warning** **[[Section 6 - Null Values](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#null-values)]** `specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp` - line 3096 - `privateIPAddressPrefixLength` is nullable in a create/read model, so the generated GET/LIST schema permits explicit `null` instead of requiring an unset field to be omitted.

**Classification reasoning:** Introduced in this PR; the property is absent from the same file at base SHA `6c5423c231b1788c23eafa9480d19affbd25c3fa` and from the prior 2026-04-01 API.

**Suggested fix:** Keep nullable only on `VirtualMachineScaleSetUpdateIPConfigurationProperties` if PATCH needs explicit clear semantics. Remove `| null` and the nullable suppression from both create/read properties at lines 3096 and 6245:

```typespec
privateIPAddressPrefixLength?: int32;
```

This keeps PATCH clearing available without emitting `x-nullable: true` into VM and VMSS GET/LIST SDK models.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [`618ad9a`](https://github.com/Azure/azure-rest-api-specs/pull/45255/commits/618ad9a97ed44a0308f3693b395609555e00ab2c). See inline comments for findings 1.

specification/compute/resource-manager/Microsoft.Compute/Compute/Compute/models.tsp
rishiraja76: Applied in `7bec2cd1f11` — thanks, this is a better shape than what I had.

`| null` and the suppression are removed from both create/read models; nullability is kept only on the update model where explicit-clear on PATCH is meaningful.

Resulting swagger in `stable/2026-09-01/ComputeRP.json`:

| Definition | `minimum` | `maximum` | `x-nullable` |
|---|---|---|---|
| `VirtualMachineScaleSetIPConfigurationProperties` | 1 | 128 | *(absent)* |
| `VirtualMachineNetworkInterfaceIPConfigurationProperties` | 1 | 128 | *(absent)* |
| `VirtualMachineScaleSetUpdateIPConfigurationProperties` | 1 | 128 | `true` |

One related change: the remaining suppression previously justified itself with "Matches Microsoft.Network's ... parity". That reasoning does not hold now that only the update model is nullable, so the reason has been reworded to state the PATCH-clear rationale directly.

`tsp compile` completes with zero warnings, so the single remaining suppression is the only one needed.

audreyttt:
Adding the approved-avocado label because this is a known exception for Compute

PR #45257, opened 2026-08-05
Add GA A2A tool support

radhgupta:
## Summary

Adds GA A2A tool support to the Microsoft Foundry REST API contract while preserving the existing preview contract.

## Changes

- Adds the GA `a2a` tool discriminator alongside `a2a_preview`.
- Adds `A2ATool` with the same configuration fields as `A2APreviewTool`.
- Adds the required `a2a_version` property with supported values `0.3` and `1.0`.
- Adds the corresponding `a2a` toolbox discriminator and `A2AToolboxTool` model.
- Keeps the existing `a2a_preview` agent and toolbox models unchanged for backward compatibility.
- Regenerates the v1 and virtual-public-preview OpenAPI JSON and YAML outputs.

## Validation

- Ran `npx tsp compile .` from `specification/ai-foundry/data-plane/Foundry` successfully.
- Confirmed the generated OpenAPI requires `a2a_version` for `A2ATool` and restricts it to `0.3` or `1.0`.
- Confirmed the generated `A2APreviewTool` schema remains unchanged.


specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -815,6 +808,36 @@ namespace Azure.AI.Projects {
...
+     * The type of the tool. Always `"a2a"`.
+     */
+    type: "a2a";
+
+    ...A2AToolBase;
+
+    /**
+     * The A2A protocol version supported by the agent.
+     */
+    a2a_version: string;
```
dargilco: Which versions are valid for the caller to set? Do we need to use a Union here to list a set of A2A protocol versions that our service supports? If so, define a TypeSpec union for it.

radhgupta: Currently supported versions are  0.3  and  1.0 . I’ll use an extensible union containing those known values plus  string , so SDKs expose known versions while remaining forward-compatible.

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -780,16 +781,19 @@ namespace Azure.AI.Projects {
...
-  /**
-   * An agent implementing the A2A protocol.
-   */
-  @extension("x-ms-foundry-meta", PreviewTool)
-  model A2APreviewTool extends OpenAI.Tool {
-    /**
-     * The type of the tool. Always `"a2a_preview`.
-     */
-    type: "a2a_preview";
+  /** Known A2A protocol versions. Open union preserves forward compatibility. */
```
dargilco: No need to write "Open union preserves forward compatibility". It's a common pattern in a large number of our unions.

dargilco: And instead of "Known" I would say  "Supported"... right? More versions listed in A2A docs... but here we list the ones Foundry supported.

dargilco (approved):

PR #45285, opened 2026-08-05
Added evaluation run duration property

posaninagendra:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -763,6 +695,9 @@ model EvalRun is OpenAI.EvalRun {
 
   /** The level at which evaluation is performed. Defaults to 'turn' if not specified. */
   evaluation_level?: EvaluationLevel = EvaluationLevel.turn;
+
+  /** The duration of the evaluation run in ISO 8601 format. */
+  duration?: duration;
```
dargilco, line 768: Why is this optional? Is EvalRun both an input (request) and output (response) model?

posaninagendra: yes, its both input/output, eval run is created and then clients poll on the GET /runs/{id} which gets the same object and once the job is complete, the duration is populated.

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
dargilco: I know it's unrelated to your change below, but can you please update this to `FoundryTimestamp` instead of integer, per other examples?

dargilco: I think "model Eval" also has the same, so if you can update that as well that would be great. Thanks!

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -193,20 +141,19 @@ model AzureAIBenchmarkDataSourceConfig extends AzureAIDataSourceConfig {
...
   #suppress "@azure-tools/typespec-azure-core/no-generic-numeric" "Auto-suppressed warnings non-applicable rules during import."
   #suppress "@azure-tools/typespec-azure-core/casing-style" "Auto-suppressed warnings non-applicable rules during import."
-  modified_at?: integer;
+  modified_at?: FoundryTimestamp;
 
   /** the name of the person who created the run. */
   #suppress "@azure-tools/typespec-azure-core/no-generic-numeric" "Auto-suppressed warnings non-applicable rules during import."
   #suppress "@azure-tools/typespec-azure-core/casing-style" "Auto-suppressed warnings non-applicable rules during import."
-  created_by?: string;
+  created_by?: FoundryTimestamp;
```
dargilco: since this is "name of person" per doc string it should remain `string`

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -748,12 +680,12 @@ model EvalRun is OpenAI.EvalRun {
...
```
dargilco: same

PR #45301, opened 2026-08-06
Add ServiceYaml validation rule to tsv

timotheeguerin:
[#44805](https://github.com/Azure/azure-rest-api-specs/pull/44805) made `service.yaml` the machine-readable source of truth for a service's API versions and their swagger outputs, across 558 projects. Nothing enforces that those manifests stay present and correct, so today a new project can ship without one, and a manifest can point at swagger files that were renamed, moved, or never generated — the `typespec-autorest` emitter only rewrites the versions it emits and never touches legacy `source: swagger` entries.

This adds a `ServiceYaml` rule to `tsv` that checks two things per project:

- a `service.yaml` exists next to `tspconfig.yaml`
- every `versions[].swagger-files` entry resolves to a real file on disk

Failures are actionable rather than just red:

```
specification/search/data-plane/Search/service.yaml references swagger files that do not exist
(paths are relative to service.yaml and are case-sensitive):

  - version "2019-05-06-Preview": preview/2019-05-06-Preview/searchindex.json
  - version "2019-05-06-Preview": preview/2019-05-06-Preview/searchservice.json

For "source: typespec" versions, run "tsp compile ." to regenerate the swagger and update
service.yaml. For "source: swagger" versions, correct the path by hand or remove the version
if it no longer exists.
```

Repo-wide: 589 projects, 563 validated, 26 skipped.

This PR is tooling-only — the manifest drift it surfaced is being fixed separately (#45304 and follow-ups).

### Scope

Only projects that emit `@azure-tools/typespec-autorest` **and** have a `main.tsp` are required to have a manifest — the same "is this a real, compilable service" gate `CompileRule` and `EmitAutorestRule` already use, so the multi-service aggregate wrappers skip naturally without needing suppressions. Swagger paths are still validated wherever a `service.yaml` exists, so nothing is silently exempted. The rule is suppressable.

Schema-shape validation (Azure/typespec-azure#4828) is **not** included yet. The emitter ships `schema/dist/ServiceYaml.json`, but it has no `exports` entry, so it can't be imported — and depending on the emitter directly would drag `@typespec/compiler` and the ARM libraries into a tool that runs on every PR. Filed Azure/typespec-azure#5174 to expose it; until then the rule parses only enough to read `swagger-files`.

Also drive-by: `tspconfig.yaml` with an empty `emit:` key threw a raw `ZodError` out of `tsp-config.ts` (hit by two `ai-foundry` projects). `EmitAutorestRule` shares that path.

`Protected Files` is expected to fail — any PR touching `eng/**` trips it.

Part of Azure/typespec-azure#4828 (presence + swagger-file checks; schema-validity pending Azure/typespec-azure#5174).

Replaces #45218, which was stacked on #44805 and got mis-retargeted when that merged.


iscai-msft (approved):

timotheeguerin:
Heads up — one manifest on `main` will still fail this rule on Linux, so it needs a fix in the spec-side track before this can go green.

`specification/search/data-plane/Search/service.yaml` references:

```
preview/2019-05-06-Preview/searchindex.json
preview/2019-05-06-Preview/searchservice.json
```

but the only directory in git is lowercase:

```
$ git ls-tree -r --name-only origin/main .../Search/preview/ | grep -iE '2019-05-06.*/(searchindex|searchservice)\.json'
specification/search/data-plane/Search/preview/2019-05-06-preview/searchindex.json
specification/search/data-plane/Search/preview/2019-05-06-preview/searchservice.json
```

#45304 kept the `2019-05-06-Preview` entry and dropped the correctly-cased `2019-05-06-preview` one. Both resolve on Windows/macOS, so the repo-wide run passes locally, but on the Linux TSV shard `preview/2019-05-06-Preview/...` does not exist.

The uppercase spelling comes from a typo in `readme.md` (line 1249) that never matched the directory on disk. Fix is either to lowercase the two paths in the manifest, or to `git mv` the directory to match — the former is less disruptive.

eng/tools/typespec-validation/src/service-yaml.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Parses the content of a `service.yaml` manifest.
+ *
+ * This deliberately only models the parts of the manifest the rule needs (the version list and its
+ * swagger paths) and ignores everything else, rather than depending on the emitter's schema — tsv
+ * stays free of a `@azure-tools/typespec-autorest` dependency, and validating the full shape is the
+ * emitter's job. Returns a readable error instead of throwing.
+ */
+export function parseServiceYaml(src: string): ParseServiceYamlResult {
```
mikeharder, line 29: nit: I would make this throw on errors instead of returning a ParseResult object, but up to you.

mikeharder (approved):

eng/tools/typespec-validation/src/service-yaml.ts
timotheeguerin: This was so the error can have a little more context below. I think ultimately I want to rework on the whole tsv error reporting to make this much clearer. Right now its quite hard to parse and inconsitent across rules.

timotheeguerin:
Existing failures fixed in https://github.com/Azure/azure-rest-api-specs/pull/45332

PR #45311, opened 2026-08-06
Add Microsoft.Chaos/scenarioTemplates to 2026-11-01

juliagao3:
## Summary

Adds `scenarioTemplates` to the `2026-11-01` stable API version — two tenant-level resources and four read-only operations.

Unlike the rest of the `2026-11-01` work, this is **not** carried over from the preview surface. It originates here, which is why it is split into its own PR.

| Operation | Path |
|---|---|
| `ScenarioTemplates_List` | `/providers/Microsoft.Chaos/scenarioTemplates` |
| `ScenarioTemplates_Get` | `/providers/Microsoft.Chaos/scenarioTemplates/{scenarioTemplateName}` |
| `ScenarioTemplateVersions_List` | `.../scenarioTemplates/{scenarioTemplateName}/versions` |
| `ScenarioTemplateVersions_Get` | `.../scenarioTemplates/{scenarioTemplateName}/versions/{version}` |

These are tenant-level, service-owned, immutable catalog resources — no write surface, no customer-supplied state. Addressing rule: **no version segment returns the latest version; a version segment returns that exact version.**

## Base branch

| PR | Contents | Base |
|---|---|---|
| **#45455** | Adds the stable `2026-11-01` version (surface unchanged from `2026-08-01-preview`) | `main` |
| **this PR** | `scenarioTemplates` only | `juliagao-microsoft-chaos-2026-11-ga-promotion` |

Review #45455 first. Once it merges, this PR retargets to `main` automatically and its diff is unchanged.

## Design notes

**Two resources, not one.** A published version is immutable — corrections ship as a new version rather than by editing an existing one, so a scenario that records the version it was created from always resolves to the same content. That requires each version to be individually addressable by URL.

**Two properties models.** `ScenarioTemplateVersionProperties` holds the blueprint content; `ScenarioTemplateProperties` spreads it and adds `latestVersion`. So `GET` on the template returns the same shape as `GET` on a version, plus a label saying which version it is — one client code path renders both.

**Macros are not resolved on read.** `actions` carry unresolved `{...}` macros; they are substituted when the scenario runs. Four namespaces are documented on the `actions` field, including `filters.` marked as the deprecated spelling of `resourceTargeting.include.`.

## Still open

Whether this should originate in a preview version rather than directly in GA. That would mean a future preview version — `2026-08-01-preview` is already published and cannot be amended. Raised on the previous review threads and not yet settled.

## Review history

Earlier review feedback from @franklixuefei was addressed in the commit `Address PR review feedback on scenarioTemplates`, covering: `PropertiesOptional`, resource name patterns widened to the Azure API Style Guide character set, the properties-model split, `maxTimeoutDuration` typed as `duration`, macro namespace documentation, and the `service.yaml` trailing newline.

This PR was originally opened as the combined `2026-11-01` change and has been narrowed to `scenarioTemplates` only; the rest moved to #45455. The existing review threads are preserved here, though a few show as outdated after the rebase.

Refs: AB#39047718


specification/chaos/resource-manager/Microsoft.Chaos/Chaos/service.yaml
```diff
@@ -108,3 +108,7 @@ versions:
     source: typespec
     swagger-files:
       - preview/2026-08-01-preview/openapi.json
+  - version: 2026-11-01
+    source: typespec
+    swagger-files:
+      - stable/2026-11-01/openapi.json
```
franklixuefei: **🔴 Blocker — this missing trailing newline is the entire `TypeSpec Validation` failure.**

Not cosmetic. `tsv` re-emits and diffs, and this is the only file it reports as changed:

```
modified: .../Chaos/service.yaml
-      - stable/2026-11-01/openapi.json
\ No newline at end of file
+      - stable/2026-11-01/openapi.json
Files have been changed after `tsp compile`.
```

Adding the newline should turn the check green on its own.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,146 @@
...
+ * are the same shape as the scenarios they produce, so an action read from a template is valid
+ * input to a scenario create.
+ *
+ * The template is addressed without a version, and always describes its newest published
+ * version. Address a specific version through the child versions resource.
+ */
+@added(Microsoft.Chaos.Versions.v2026_11_01)
+@tenantResource
+model ScenarioTemplate
+  is Azure.ResourceManager.ProxyResource<ScenarioTemplateProperties> {
```
franklixuefei: **🟠 `properties` ends up optional here, but is required on this RP''s sibling catalog resource.**

From the generated `stable/2026-11-01/openapi.json`:

```
Action                   required=[properties]
ActionVersion            required=[properties]
ScenarioTemplate         required=[]      ←
ScenarioTemplateVersion  required=[]      ←
```

`Action` declares `ProxyResource<ActionProperties, false>`; this uses the defaulted second arg (`true`). A read-only catalog entry always has properties, so this pushes a null-check into every generated SDK for something the service always returns — and GA freezes it.

```diff
-  is Azure.ResourceManager.ProxyResource<ScenarioTemplateProperties> {
+  is Azure.ResourceManager.ProxyResource<ScenarioTemplateProperties, false> {
```

Same on `ScenarioTemplateVersion` (L43).

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,146 @@
...
+  ...ResourceNameParameter<
+    Resource = ScenarioTemplate,
+    KeyName = "scenarioTemplateName",
+    SegmentName = "scenarioTemplates",
+    NamePattern = "^[A-Za-z][A-Za-z0-9]*$"
```
franklixuefei: **🟠 More restrictive than guidance, tighter than the sibling resource, and a one-way door.**

`^[A-Za-z][A-Za-z0-9]*$` forbids `-`, `.`, `_`, `~`.

- The Azure API Style Guide says you **should** allow `0-9 A-Z a-z - . _ ~`.
- `Action` in this same RP already allows hyphens: `^[a-zA-Z0-9\-]+$`.
- It fits all 44 templates today — but the repo''s own breaking-change guidelines close the door permanently:

> "Resource naming rules should not change… **Even if the rules become less strict**, clients relying on earlier name constraints to perform local validation will fail."

So post-GA you could never add `ZoneDown-Extended` or `Aks_ZonalFailure`. Cheapest possible moment to widen this is now.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,146 @@
...
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-provisioning-state" "Read-only catalog resource served from the resource provider; there is nothing to provision."
+@added(Microsoft.Chaos.Versions.v2026_11_01)
+model ScenarioTemplateProperties {
+  /**
+   * The version of the template being described. On the template resource this is its newest
+   * published version.
+   */
+  @visibility(Lifecycle.Read)
+  version: string;
```
franklixuefei: **🟡 Duplicates the child resource''s `name`, and means two different things depending on the carrier.**

On `ScenarioTemplateVersion` this is byte-identical to `name` (both are the version). On `ScenarioTemplate` it means "newest published". One field, two semantics, one of them redundant.

Style Guide: *"DO NOT add fields to the JSON if the value is easily computable from other fields to avoid bloating the body."*

This is a side-effect of one properties model serving both resources. Options: drop it from the child, or split into `ScenarioTemplateProperties` (with `latestVersion`) and `ScenarioTemplateVersionProperties`.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,146 @@
...
+   */
+  @visibility(Lifecycle.Read)
+  description?: string;
+
+  /**
+   * ISO 8601 duration bounding how long a scenario created from this template may run
+   * (for example PT3H for three hours).
+   */
+  @visibility(Lifecycle.Read)
+  maxTimeoutDuration?: string;
```
franklixuefei: **🟡 Documented as ISO-8601 but typed as a bare `string`.**

Slightly ironic in the PR that adds `ParameterFormatType.Duration` so *parameters* can declare duration-ness — this field can''t.

```diff
-  maxTimeoutDuration?: string;
+  maxTimeoutDuration?: duration;
```

`duration` emits `format: duration`. (Style Guide''s stated preference is actually a fixed-unit number with the unit in the name, e.g. `maxTimeoutInMinutes` — either beats bare `string`.)

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * The actions that define the template's orchestration. These are the same model a scenario
+   * publishes, and carry unresolved `%%\{...\}%%` macros; the macros are substituted when the
+   * scenario runs.
+   */
+  @visibility(Lifecycle.Read)
+  @minItems(1)
+  @identifiers(#["name"])
+  actions: ScenarioAction[];
```
franklixuefei, line 114: **🟠 The contract serves three macro namespaces but only models two.** ⭐

The doc says actions carry `%%{...}%%` macros, and `resourceSelectors[]` / `parameters[]` describe what they bind to. But this PR''s own `examples/2026-11-01/ScenarioTemplateVersions_List.json` returns:

```
%%{parameters.duration}%%              ← documented
%%{resourceTargeting.include.zones}%%  ← NOT modelled, NOT documented
%%{filters.zones}%%                    ← legacy, NOT documented, NOT deprecated
```

All four namespaces are implemented in `TemplateMacroPatterns.cs`, mirroring BE''s resolvers. And per your **D17** correction on AB#39047718, `resourceTargeting.include.*` is a first-class binding that survives to run time and resolves from the configuration snapshot.

The plan already decided this (**Q8**): the contract *"exposes **both**"*, and *"Annotating `filters.` as deprecated is **T003-8**"* — which is unaddressed here.

It matters concretely: **9 template versions still carry `filters.`** (`ComputeZoneDown-1.0/1.1`, `ZoneDown-0.5/1.0`, the SQL/PostgreSQL failover `1.0/1.1`s), and `ScenarioTemplateVersions.list` is explicitly documented as returning superseded versions — so these reach customers on day one.

A customer who `GET`s a template sees `%%{resourceTargeting.include.zones}%%` and has no contract-discoverable way to learn what to bind. Please either model the third namespace alongside `resourceSelectors`, or at minimum enumerate all namespaces in this doc comment and mark `filters.` deprecated (T003-8).

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.tsp
```diff
@@ -0,0 +1,61 @@
...
+   * Get a scenario template, described by its newest published version.
+   */
+  get is ArmResourceRead<ScenarioTemplate>;
+
+  /**
+   * Get a list of the scenario templates published by Microsoft.
+   */
+  list is ArmResourceListByParent<
+    ScenarioTemplate,
+    Response = ScenarioTemplateListResult
```
franklixuefei, line 25: **🟡 The `Response =` override is redundant and moves away from the ARM idiom.**

`ArmResourceListByParent<T>` already returns `ArmResponse<ResourceListResult<T>>` — the standard ARM list envelope. Overriding it with `Azure.Core.Page<T>` re-specifies the default in a non-ARM-idiomatic type.

```diff
-  list is ArmResourceListByParent<
-    ScenarioTemplate,
-    Response = ScenarioTemplateListResult
-  >;
+  list is ArmResourceListByParent<ScenarioTemplate>;
```

Acknowledged this matches `Action`/`ActionVersion`, so it''s pre-existing convention rather than a regression — but a brand-new GA surface is the cheapest place to stop propagating it. If you keep it, worth double-checking `GetCollectionResponseSchema` given parent and child share one properties model.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/main.tsp
```diff
@@ -59,6 +60,10 @@ enum Versions {
   /**
    * The 2026-08-01-preview API version.
    */
-  @Azure.Core.previewVersion
   v2026_08_01_preview: "2026-08-01-preview",
+
+  /**
+   * The 2026-11-01 API version.
+   */
+  v2026_11_01: "2026-11-01",
```
franklixuefei, line 68: **🟠 This GA is the *first* appearance of `scenarioTemplates`, not a promotion of it.**

To be clear up front: **removing `@previewVersion` from the line above is correct** — it''s the literal documented step in the [stable-after-preview howto](https://github.com/Azure/typespec-azure/blob/main/website/src/content/docs/docs/howtos/Versioning/03-stable-after-preview.md). Please leave that.

The issue is what''s being added. Verified across every preview folder:

```
scenarioTemplates   in preview/2026-08-01-preview → 0 occurrences
ParameterFormatType in preview/2026-08-01-preview → 0 occurrences
```

Every new model is `@added(v2026_11_01)`. But the PR description says this "promote[s] the complete current `2026-08-01-preview` API surface" — for `scenarioTemplates` that isn''t a promotion. The sanctioned workflow only describes *graduation* (`@added(preview)` → `@added(stable)`); it has no path for originating types at `@added(stable)`.

Mechanically it''s allowed (a new resource type is non-breaking). The cost is that the shape gets zero preview bake, and post-GA every correction needs a new api-version plus Breaking Change Board — which is exactly where H2/H3/H4 in my summary bite.

AB#39047718 is still **New**, "ARM API review approved" is unmet, and T003-1 (resolve the ARM Review Risks *before* authoring `.tsp`) looks to have happened after. Suggest landing `scenarioTemplates` in `2026-08-01-preview` and letting this GA promote it — which also makes the PR description true.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/readme.md
```diff
@@ -31,6 +31,15 @@ openapi-type: arm
 tag: package-2025-01
 ```
 
+### Tag: package-2026-11
```
franklixuefei, line 34: **🟡 New GA tag added, but the default `tag:` still points at `package-2025-01`.**

Line 31 still reads `tag: package-2025-01`, so AutoRest consumers and the docs pipeline keep resolving to the old GA. Convention is to advance the default to the newest stable when a GA ships:

```diff
-tag: package-2025-01
+tag: package-2026-11
```

Worth confirming against the SDK release plan in case the lag is deliberate.

franklixuefei (commented):
## Review — `2026-11-01` GA + `Microsoft.Chaos/scenarioTemplates`

Reviewed against four things, not just the diff:

- **ARM guidance** — the TypeSpec-Azure [stable-after-preview workflow](https://github.com/Azure/typespec-azure/blob/main/website/src/content/docs/docs/howtos/Versioning/03-stable-after-preview.md), the [Azure API Style Guide](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md), the [breaking-change guidelines](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/Breaking%20changes%20guidelines.md), and the LintDiff rule set.
- **This RP's own precedent** — `Action` / `ActionVersion`, the closest existing read-only parent/child proxy pair.
- **The team's plan** — `built-in-scenario-templates.plan.md`, which makes this PR **EPIC-003 / AB#39047718**.
- **The GW implementation** — ADO `!16703100` (catalog, `TemplateMacroPatterns`, `ScenarioTemplateResourceId`) and the 44 embedded template files it serves.

**Verdict: strong modeling, but not mergeable yet — 3 CI blockers and 4 issues that GA makes permanent.** The core risk is that for `scenarioTemplates` this GA is simultaneously the contract's *first* and *final* published form.

> **First, a non-finding, so it doesn't get "fixed":** removing `@Azure.Core.previewVersion` from `v2026_08_01_preview` is **correct**. It is the literal documented step in the stable-after-preview howto. Please leave it as-is.
>
> Likewise the `%%{...}%%` macro form is **right**, and I'd push back on anyone asking for structured bindings here — that's **D-TPL-1**, and the projector was deliberately deleted. Note the ADO PR description and AB#39047739's title ("macro to structured binding projection") are now *stale* relative to D-TPL-1, not this spec.

---

### 🔴 Blockers

**B1 — `service.yaml` is missing its trailing newline. This is the entire `TypeSpec Validation` failure.**
Not a nit — I pulled the CI diff and it is the only file the compiler reports as modified:

```
modified: specification/.../Chaos/service.yaml
-      - stable/2026-11-01/openapi.json
\ No newline at end of file
+      - stable/2026-11-01/openapi.json
Files have been changed after `tsp compile`.
```
One character unblocks the check.

**B2 — LintDiff introduces 15 must-fix errors on the new stable version.**

| Rule | × | Note |
|---|---|---|
| `PostResponseCodes` | 8 | RPC-Async-V1-11/14, RPC-POST-V1-02/03 |
| `TrackedExtensionResourcesAreNotAllowed` | 3 | `targets` — RPC-Uri-V1-12 |
| `TrackedResourcePatchOperation` | 1 | `Target` needs a tags PATCH |
| `RequestSchemaForTrackedResourcesMustHaveTags` | 1 | |
| `AvoidAdditionalProperties` | 1 | RPC-Policy-V1-05 |
| `GuidUsage` | 1 | `recommendedRoles: uuid[]` — needs API-board sign-off |

These are inherited from the preview surface, but LintDiff scores a new stable file as all-new — **promoting the surface promotes its violations, and GA freezes them.** They need fixes or recorded `suppressions.yaml` entries with ARM concurrence. (84 × `LatestVersionOfCommonTypesMustBeUsed` also suggests moving to common-types **v6** while the version is still editable.)

**B3 — merge gate.** The bot requires either the **`PublishToCustomers`** label or retargeting to a `release-*` branch. Also worth chasing the **`BreakingChange-Python-Sdk`** label — a purely additive GA normally shouldn't trip it, so it's worth confirming *what* was flagged rather than just getting it approved away.

---

### 🟠 GA-permanence issues

**H1 — `scenarioTemplates` debuts directly in a stable version, with no preview lineage.**
Verified: `scenarioTemplates` and `ParameterFormatType` occur **0 times** in *every* preview folder, including `2026-08-01-preview`. Every new model is `@added(v2026_11_01)`.

This is worth pausing on because the PR description says the change "promote[s] the complete current `2026-08-01-preview` API surface" — but `scenarioTemplates` isn't a promotion, it's a debut. The sanctioned workflow describes only *graduation* (`@added(preview)` → `@added(stable)`); it has no story for originating types at `@added(stable)`. Mechanically legal (adding a resource type is non-breaking), but it means the shape gets zero preview bake and every later correction costs a new api-version + Breaking Change Board.

Given AB#39047718 is still **New**, with "ARM API review approved" unmet and T003-1 ("resolve the five ARM Review Risks *before* authoring any `.tsp`") apparently done after the fact — **I'd land `scenarioTemplates` in `2026-08-01-preview` first and let the GA promote it**, which also restores the PR description's stated intent.

**H2 — The contract does not model two of the three macro namespaces it actually serves.** ⭐ *most important finding*

`ScenarioTemplateProperties` documents only `%%{parameters.*}%%` and `%%{resourceSelectors.*}%%`. But this PR's **own examples** return more than that — from `ScenarioTemplateVersions_List.json`:

```
%%{parameters.duration}%%              ← documented
%%{resourceTargeting.include.zones}%%  ← NOT modelled, NOT documented
%%{filters.zones}%%                    ← legacy, NOT documented, NOT deprecated
```

Both extra namespaces are real and load-bearing:
- `TemplateMacroPatterns.cs` implements all four patterns, mirroring BE's resolvers.
- Plan **Q8** answers this explicitly — the contract "exposes **both**", and *"Annotating `filters.` as deprecated is T003-8"*. **T003-8 is unaddressed.**
- Your own **D17** correction on AB#39047718 makes `resourceTargeting.include.*` a first-class binding that survives to run time.
- **9 template versions still carry `filters.`** (`ComputeZoneDown-1.0/1.1`, `ZoneDown-0.5/1.0`, the SQL/PostgreSQL failover 1.0/1.1s) — all reachable, because `ScenarioTemplateVersions.list` is documented as returning superseded versions.

Net effect: a customer can `GET` a template, see `%%{resourceTargeting.include.zones}%%`, and have **no contract-discoverable way to learn what to bind** — `resourceSelectors[]` and `parameters[]` won't mention it. The contract isn't self-describing, and GA makes that permanent. Please either model the third namespace (e.g. a `resourceTargeting`/`includes` descriptor array alongside `resourceSelectors`) or, at minimum, document all four namespaces on `actions` and mark `filters.` deprecated per T003-8.

**H3 — `properties` is optional here, but required on this RP's sibling catalog resource.**
From the generated `stable/2026-11-01/openapi.json`:

```
Action                   required=[properties]
ActionVersion            required=[properties]
ScenarioTemplate         required=[]      ← 
ScenarioTemplateVersion  required=[]      ← 
```
`Action` uses `ProxyResource<ActionProperties, false>`; this PR uses the defaulted `ProxyResource<ScenarioTemplateProperties>`. A read-only catalog entry *always* has properties, so this forces a null-check into every SDK for something the service always returns — and diverges from the pattern two resources over. **Fix: `ProxyResource<ScenarioTemplateProperties, false>`.**

**H4 — `NamePattern` is more restrictive than guidance, tighter than the sibling resource, and unfixable later.**
`^[A-Za-z][A-Za-z0-9]*$` forbids `-`, `.`, `_`, `~`. The Style Guide says you *should* allow `0-9 A-Z a-z - . _ ~`; `Action` already allows hyphens (`^[a-zA-Z0-9\-]+$`). It fits all 44 templates today, but the breaking-change guidelines are explicit that this is a one-way door:

> "Resource naming rules should not change… **Even if the rules become less strict**, clients relying on earlier name constraints to perform local validation will fail."

So `ZoneDown-Extended` or `Aks_ZonalFailure` can never be added. Recommend widening now.

---

### 🟡 Modeling / consistency

- **`Response =` override is redundant.** `ArmResourceListByParent<T>` already returns the ARM-idiomatic `ResourceListResult<T>`; overriding it with `Azure.Core.Page<T>` is non-idiomatic for ARM. (Acknowledged: `Action` does the same, so this is pre-existing — but a new GA is the cheapest place to stop propagating it.)
- **`properties.version` duplicates the child's `name`.** On `ScenarioTemplateVersion` the same value appears twice, and the field means two different things depending on which resource carries it ("newest published" vs "this one"). Style Guide: *"DO NOT add fields to the JSON if the value is easily computable from other fields."* Consider dropping it from the child, or splitting the two models.
- **`maxTimeoutDuration?: string`** is documented as ISO-8601 but typed as a bare string — in the very PR that adds `ParameterFormatType.Duration` to declare duration-ness. Use TypeSpec `duration` (emits `format: duration`), or a fixed-unit `maxTimeoutInMinutes`.
- **Version format is ambiguous and being ratified.** `^\d+\.\d+(\.\d+)?$` admits both `1.2` and `0.9.0`, and the catalog genuinely mixes them. Under `System.Version` (what `BuiltInTemplateCatalog` sorts with) `1.2` and `1.2.0` are *distinct and unequal*, so they'd be two resources with a non-obvious "latest". Worth settling on one documented form before GA.
- **`displayName` / `description` optional** on a Microsoft-published read-only catalog — nullable in every SDK for data the RP always emits.
- **`readme.md`**: the default `tag: package-2025-01` isn't advanced to `package-2026-11`.
- **Hidden templates**: the contract (correctly) has no `visibility` field, but the read-path filter is **EPIC-005 T005-9**, which is still blocked. `MockScenario` and the prerelease templates must not appear in tenant-scoped LIST on day one.

---

### ✅ Genuinely right

- `@armResourceOperations` + `@tenantResource` rather than hand-rolled `@route`, and `@visibility(Lifecycle.Read)` rather than the string form — ARM Review Risks #1 and #2 from AB#39047718, both resolved.
- Tenant-scope operations only, no subscription-scoped interfaces — exactly T003-2.
- Examples for all four GET/LIST operations — T003-3 done.
- Read-only interfaces correctly model an immutable Microsoft-published catalog; the `provisioningState` suppression is the documented escape hatch and matches `Action`'s.
- `union ParameterFormatType { string, ... }` is the correct extensible-enum pattern.
- Reusing `ScenarioAction` so a template action is valid scenario input is a genuinely good call, and it's what makes D-TPL-1 coherent.

---

### Suggested order

1. `service.yaml` newline → green `TypeSpec Validation`. *(1 min)*
2. `ProxyResource<..., false>`, widen `NamePattern`, drop `Response=`, type the duration. *(cheap, all one-way doors)*
3. Decide **H2** — model or document `resourceTargeting.include.*`, deprecate `filters.` (T003-8).
4. Decide **H1** — preview-first vs. straight to GA. This one is worth a deliberate call, not a default.
5. Then LintDiff (B2) + labels (B3), and get the ARM review queued early — T003-5 flags it as having real lead time.

Happy to pair on H2 — it's the one where the contract and the catalog genuinely disagree today.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/service.yaml
juliagao3: Fixed in `5d1d487` -- added the trailing newline.

Confirming the mechanism for anyone reading later: `tsp compile` regenerates this file, and it also **destructively rewrites it**, dropping every legacy swagger-sourced version (2021-09-15-preview through 2024-03-22-preview, ~97 lines). So the recovery is `git checkout -- service.yaml` and then re-append the newline, not a re-compile. The diff on this file is now exactly the one byte you identified.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Fixed in `5d1d487`, and I went one step further than you asked.

Both patterns are now the Style Guide set:

```diff
-  NamePattern = ^[A-Za-z][A-Za-z0-9]*\$      // scenarioTemplateName
-  NamePattern = ^\d+\.\d+(\.\d+)?\$          // version
+  NamePattern = ^[A-Za-z0-9\-._~]+\$         // both
```

I widened `version` too because your argument is symmetric and you only happened to flag the name. `^\d+\.\d+(\.\d+)?\$` fits all 44 files today, but it would have permanently locked out `1.2-beta` or `2.0-preview` -- and the version segment is *more* likely to need a suffix than the name is, since it is the axis we actually publish along.

The cost is that the pattern no longer documents "this looks like a version." I think that is fine: a malformed version is a 404 from the catalog either way, so the pattern was validating something the service already validates, in exchange for a door that only closes.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Fixed in `5d1d487` -- both resources now pass `false`.

Verified in the regenerated `openapi.json`:

```
Action                   required=[properties]
ActionVersion            required=[properties]
ScenarioTemplate         required=[properties]   <- was []
ScenarioTemplateVersion  required=[properties]   <- was []
```

Agreed on the reasoning: these are read-only catalog entries served from the RP's own embedded catalog, so `properties` is never absent. Defaulting it optional would have put a null check in every generated SDK for a value the service always returns.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Fixed in `5d1d487` -- now `duration`, emitting `format: duration`.

I checked this against the catalog before changing it, because the neighbouring `ScenarioAction.duration` is deliberately a bare `string` -- it has to accept `%%{parameters.duration}%%`, which is not valid ISO-8601. If `maxTimeoutDuration` had the same requirement the stricter type would have broken it.

It does not. All 44 template files carry a literal:

```
PT3H   x41
PT1H   x1    (PilotfishPowerLoss-0.1)
PT30M  x1    (MockScenario-0.1)
```

It is a service-authored ceiling, not customer input, so no macro can appear there. Stricter type costs nothing.

On the fixed-unit alternative (`maxTimeoutInMinutes`) -- agreed it is the Style Guide's stated preference, but the value is already ISO-8601 on the wire from the existing backend, so that one would be a data change rather than a type change. `duration` gets the format annotation without touching the catalog.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Fixed in `5d1d487` -- took your second option and split the models.

```
ScenarioTemplateVersionProperties  { displayName, description, maxTimeoutDuration,
                                     resourceSelectors, parameters, actions }
ScenarioTemplateProperties         { latestVersion, ...ScenarioTemplateVersionProperties }
```

Generated result:

```
parent props: latestVersion,displayName,description,maxTimeoutDuration,resourceSelectors,parameters,actions
child  props:               displayName,description,maxTimeoutDuration,resourceSelectors,parameters,actions
```

I also renamed it `version` -> `latestVersion`. Your "one field, two semantics" point is the real defect and dropping it from the child only fixes half of it -- the surviving field on the parent still read as "the version" when it means "the newest one." `latestVersion` says it outright, and it now cannot collide with the child's `name`.

Spread rather than inheritance so the two stay independently versionable: if the parent ever needs a field the version does not have (a deprecation flag, say), it can take one without touching the child's schema. Examples updated on both sides.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Took the documentation half in `5d1d487`; pushing back on the modelling half.

**Done:** `actions` now enumerates all four namespaces and marks `filters.` deprecated in prose:

| macro | resolves from | declared by |
|---|---|---|
| `%%{parameters.<name>}%%` | parameter values supplied at configure time | `parameters` |
| `%%{resourceSelectors.<name>}%%` | resources discovered for that selector | `resourceSelectors` |
| `%%{resourceTargeting.include.<dimension>}%%` | the scenario configuration's targeting | -- nothing to declare |
| `%%{filters.<dimension>}%%` | same resolver, deprecated spelling | -- |

That closes the discoverability gap you described: a caller reading a template can now learn what every macro binds to without reading our source.

**Not doing:** modelling `resourceTargeting` as a sibling array of `resourceSelectors`. The first two namespaces have declaration lists because the template *constrains* them -- it decides that `parameters.duration` exists and `parameters.frobnicate` does not. `resourceTargeting.include.*` is not like that. It reads dimensions off the configuration the customer supplies at run time (`zones`, `locations`, ...), and that vocabulary belongs to the configuration model, not to the template. A per-template list would either restate a fixed enum on all 44 templates or imply a constraint the template does not actually impose.

Confirmed against the resolvers, which take it from the targeting policy rather than from anything the template declares:

```
FilterTemplateMacroResolver.cs:22  %%\{\s*filters\.(?<name>[a-zA-Z0-9_]+)\s*\}%%
FilterTemplateMacroResolver.cs:25  %%\{\s*resourceTargeting\.include\.(?<name>[a-zA-Z0-9_]+)\s*\}%%
```

Both spellings hit the *same* resolver, which is why the 1.1/1.2 difference is cosmetic.

**On T003-8** -- correct that it is unaddressed, and it stays unaddressed here deliberately: it is sequenced after this PR. Worth being precise about what it can and cannot do, though. `filters.` is not a field we can `@deprecated` -- it is a substring inside an opaque `string` value in `actions`. There is no schema node to hang the annotation on. So T003-8 is (a) this prose, which is now in, and (b) re-spelling the 9 remaining `filters.` versions in the catalog. On (b): those versions are immutable by design (D-TPL-2), so they keep their original spelling permanently -- which is exactly why the newest version of every template already uses `resourceTargeting.include.` and why the prose tells callers not to depend on the old one.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.tsp
juliagao3: Keeping it, but you are right that it is worth stating why rather than just pointing at precedent.

The named list results are load-bearing for us in a way the inline default is not: `ScenarioTemplateListResult` and `ScenarioTemplateVersionListResult` are stable schema names that the generated SDKs and our own GW model generation bind to. Dropping the override renames those types in every downstream generator for no behavioural gain, and it would make this one resource pair inconsistent with `Action`/`ActionVersion` sitting directly above it in the same file.

Your `GetCollectionResponseSchema` concern is now moot, and it is a good catch that it *was* a concern -- it was a real consequence of the shared properties model, which the split above removed. After `5d1d487`:

```
ScenarioTemplateListResult.value.items         -> #/definitions/ScenarioTemplate
ScenarioTemplateVersionListResult.value.items  -> #/definitions/ScenarioTemplateVersion
ScenarioTemplate.properties                    -> #/definitions/ScenarioTemplateProperties
ScenarioTemplateVersion.properties             -> #/definitions/ScenarioTemplateVersionProperties
```

Parent and child now resolve to distinct schemas at every level, so there is no ambiguity for the collection emitter to get wrong.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/readme.md
juliagao3: Holding this one, and it is downstream of the thread on `main.tsp` L68.

If `scenarioTemplates` lands in `2026-08-01-preview` instead, there is no new GA in this PR and the default tag correctly stays at `package-2025-01`. Advancing it now and reverting would churn every AutoRest consumer twice.

Flagging one thing that is true either way, so it does not get lost: the default tag should not advance ahead of the SDK release plan. Moving `package-2025-01` -> `package-2026-11` repoints the docs pipeline and every `readme.md`-driven generator at a version whose SDKs are not published yet. So even in the "this stays GA" branch, this is a separate change gated on the SDK cut, not something to fold into the spec PR.

Will resolve once the placement thread is settled.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/main.tsp
juliagao3: Fair, and this one is genuinely open -- I am not going to resolve it in this round.

Accepting your factual correction: this is an origination, not a promotion. `scenarioTemplates` and `ParameterFormatType` appear zero times in `preview/2026-08-01-preview`, every model here is `@added(v2026_11_01)`, and so the PR description overstates it. That wording is wrong regardless of where the resource lands, and I will fix it.

Noting for the record that `@previewVersion` stays removed, per your first paragraph.

Where I am: the preview-vs-GA placement is still under discussion on our side, so I would rather not churn the folder twice. The reason I am comfortable landing the review fixes ahead of that decision is that **none of them depend on it.** `ProxyResource<T, false>`, the widened name patterns, the properties split, `format: duration`, and the macro documentation are all shape corrections -- they move verbatim into `2026-08-01-preview` if that is where this ends up. Deferring them until the placement is settled would only mean fixing the same six things later against a bigger diff.

Your H2/H3/H4 point cuts the other way too, and it is the strongest argument in the comment: the fixes above are exactly the class of correction that becomes a Breaking Change Board item once this ships stable. Landing them now is cheap precisely because the placement is not yet final.

I will come back to this thread with the decision. If it lands in preview, the GA promotion becomes a mechanical `@added` bump and the PR description becomes true as written.

juliagao3:
Thanks -- this was a thorough pass. Pushed `5d1d487` addressing six of the nine.

| # | Comment | Status |
|---|---|---|
| 🔴 | `service.yaml` trailing newline | **Fixed** -- diff on that file is now exactly the one byte |
| 🟠 | `properties` optional | **Fixed** -- `ProxyResource<T, false>`, matches `Action`/`ActionVersion` |
| 🟠 | Name pattern too restrictive | **Fixed** -- and widened `version` too, same argument |
| 🟠 | Third macro namespace undocumented | **Docs fixed**; pushing back on modelling it -- see thread |
| 🟡 | `latestVersion` duplicates child `name` | **Fixed** -- properties model split, renamed `latestVersion` |
| 🟡 | `maxTimeoutDuration` bare string | **Fixed** -- `duration`, verified no macros in the catalog |
| 🟡 | `Response =` override | **Keeping** -- named list results are load-bearing; your `GetCollectionResponseSchema` concern is moot after the split |
| 🟠 | GA vs preview origination | **Open** -- under discussion, deliberately not resolved this round |
| 🟡 | `readme.md` default tag | **Held** -- downstream of the placement decision |

Two things worth pulling out of the threads.

**Your factual correction on the placement comment is accepted.** This is an origination, not a promotion -- `scenarioTemplates` appears zero times in `2026-08-01-preview`. The PR description is wrong on that point regardless of where the resource lands, and I will fix the description.

**None of the six fixes depend on the placement decision.** They are all shape corrections and move verbatim into a preview folder if that is where this ends up. That is why I landed them ahead of the decision rather than after it -- and your H2/H3/H4 argument is the reason it would be a mistake to wait, since these are exactly the corrections that turn into Breaking Change Board items once the surface ships stable.

One correction to flag on **T003-8**, since it came up: `filters.` cannot be `@deprecated` in TypeSpec. It is a substring inside an opaque `string` value in `actions` -- there is no schema node to annotate. So T003-8 reduces to the prose that just landed, plus the fact that the newest version of every template already uses `resourceTargeting.include.`. The 9 versions still carrying `filters.` keep it permanently, by design (D-TPL-2 immutability).

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/readme.md
juliagao3: Reversing my earlier "holding" call on this — you were right, and CI proves it.

Swagger Avocado is failing on exactly this line:

- `NOT_LATEST_API_VERSION_IN_DEFAULT_TAG` x2
- `MISSING_APIS_IN_DEFAULT_TAG` x66

So this is a blocking bug regardless of where `scenarioTemplates` ends up, not something downstream of that decision. I had it backwards.

Fixed in 5b24fb5: `tag: package-2026-11`.

Before flipping it I checked that the new default is not narrower than the old one — `stable/2026-11-01/openapi.json` is a strict superset of `stable/2025-01-01/openapi.json` (51 paths vs 18, zero dropped), so every operation reachable under the previous default is still reachable.

If `scenarioTemplates` moves to `2026-08-01-preview`, this line stays correct anyway: `package-2026-11` remains the newest **stable** tag in the readme, which is what the default is supposed to point at. Only the contents of the 2026-11 tag would change.

juliagao3:
## CI status update, and a LintDiff finding that bears on the GA-vs-preview question

Two more pushes since the review round:

| Commit | What |
|---|---|
| `d7b6297` | `npx tsp format` — my hand-edits had left `);@@doc(` collapsed onto one line, which is what turned **TypeSpec Validation** red. Recompiled and confirmed `openapi.json` is byte-identical, so formatting-only. |
| `5b24fb5` | Default AutoRest tag -> `package-2026-11` (fixes **Swagger Avocado**, see that thread). |

**TypeSpec Validation is now green.**

### The LintDiff finding

`Swagger LintDiff` is failing with 17 errors. **Zero of them are on `scenarioTemplates`** — I grepped the job log to be sure. They all land on the legacy surface this PR promotes into a new stable version:

| Code | Count | Where |
|---|---|---|
| `PostResponseCodes` | 10 | `experiments/{}/start`, `/cancel`, `workspaces/{}/discover`, `/evaluate`, scenario `/execute`, `/validate`, `/fixResourcePermissions`, run `/cancel` |
| `TrackedExtensionResourcesAreNotAllowed` | 3 | `targets/{targetName}` GET + PUT (extension resource returning `location`) |
| `RequestSchemaForTrackedResourcesMustHaveTags` | 1 | `targets/{targetName}` PUT |
| `TrackedResourcePatchOperation` | 1 | `Target` |
| `AvoidAdditionalProperties` | 1 | `Target.properties` |
| `GuidUsage` | 1 | `Azure.Core.uuid` |

Two of these also fire against the `before/` document (`stable/2025-01-01`), which confirms they are long-standing Chaos v1 design, not regressions introduced here. They surface now only because a **new stable version** re-lints the whole surface.

### Why this matters for the placement decision

This is the strongest argument I have seen so far for **not** originating `scenarioTemplates` in a GA version in this PR. The GA promotion drags 17 pre-existing ARM violations into a stable contract, and clearing them is either a batch of `suppressions.yaml` entries with ARM board sign-off, or real breaking changes to `targets` and the POST actions. That is a much bigger conversation than the new read-only resource, and it should not gate it.

Concretely, I would rather split this:

1. `scenarioTemplates` originates in `2026-08-01-preview` — four read-only operations, no legacy surface, nothing for LintDiff to complain about.
2. The `2026-11-01` GA promotion becomes its own PR, where the 17 violations get triaged properly (suppress vs. fix) without a feature waiting behind them.

That is your call, not mine — but if you agree, I will do the split, and it also resolves the `main.tsp` L68 thread. Otherwise tell me which of the 17 you want suppressed and I will add them to `suppressions.yaml` with justifications.

juliagao3:
## Withdrawing the split proposal above — resolved a simpler way

Disregard my suggestion to move `scenarioTemplates` into `2026-08-01-preview`. We are keeping it in `2026-11-01` GA, and the LintDiff failure turned out not to need it.

The 17 errors were not a real triage problem — the `package-2026-11` tag simply had **no suppressions block at all**. The identical suppressions already exist and are ARM-approved on the adjacent `package-2026-08-01-preview` tag; my new tag just never inherited them.

Fixed in ae4f36f by porting that block verbatim, so the GA tag asserts exactly what the preview tag already asserts — no new justifications authored, nothing newly waived.

Coverage against the 15 errors on the `after` document:

| Code | Where |
|---|---|
| `TrackedExtensionResourcesAreNotAllowed` | `targets` GET / PUT |
| `RequestSchemaForTrackedResourcesMustHaveTags` | `targets` PUT |
| `TrackedResourcePatchOperation` | `Target` |
| `AvoidAdditionalProperties` | `Target.properties.properties` |
| `GuidUsage` | `Azure.Core.uuid` |
| `PostResponseCodes` | `discover`, `evaluate`, `execute`, `validate`, `fixResourcePermissions`, `runs/{runId}/cancel` |

The other two (`PostResponseCodes` on `experiments/{experimentName}/start` and `/cancel`) also fire on the **before** document (`stable/2025-01-01`), so LintDiff treats them as pre-existing rather than newly introduced. No tag in this readme suppresses them today and this change keeps it that way.

I also confirmed `2026-11-01` and `2026-08-01-preview` differ by exactly the four new `scenarioTemplates` paths, so the suppression set transfers 1:1 with nothing missing and nothing extra needed.

**Net: `scenarioTemplates` stays in `2026-11-01` GA.** That resolves the placement question on the `main.tsp` L68 thread as well.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/main.tsp
juliagao3: Closing this one out: **decision is to keep `scenarioTemplates` in `2026-11-01` GA.**

Your framing was correct and I have stopped describing it as a promotion — the PR description now says explicitly that this resource originates here and does not exist in any published preview version. I verified that: `scenarioTemplate` appears 43 times in `stable/2026-11-01/openapi.json` and **0 times** in `preview/2026-08-01-preview`, `preview/2026-05-01-preview`, and `preview/2024-11-01-preview`.

I did briefly float moving it into `2026-08-01-preview`, mostly because LintDiff looked like it was going to force a large triage. That turned out to be a false alarm — the `package-2026-11` tag was just missing the suppressions block that the preview tag already has (details in the PR comment). So there is no longer a CI-driven reason to relocate it, and I have withdrawn that suggestion.

On the substance of preview-vs-GA for a new resource: what makes this one defensible directly in GA is that there is no write surface to get wrong. Four read-only operations over a service-owned, immutable catalog — no customer-supplied state, no PUT/PATCH/DELETE, no LRO. The shape we would learn from a preview cycle is mostly the mutation surface, and there isn't one. The addressing rule (no version segment -> latest, version segment -> that exact version) is the only real design choice, and it is documented on the operations.

Happy to keep debating it, but I did not want it holding up the rest of the PR. If you feel strongly it should originate in preview, say so and I will do the move as a follow-up rather than reworking this PR.

franklixuefei (commented):
## ✅ Sign-off — all three blockers cleared, shape fixes verified in the generated output

Re-reviewed at `ae4f36f`. I verified every claim against the tree and the emitted `openapi.json` rather than taking the replies at face value — **everything checks out**, including both points you pushed back on. Not voting since this is a draft; consider this my sign-off on the spec changes.

### CI

**44 checks passing, zero failing.** All three of my blockers are green:

| Was | Now |
|---|---|
| `TypeSpec Validation` ❌ (the one-byte `service.yaml` newline) | ✅ pass |
| `Swagger LintDiff` ❌ 15 must-fix | ✅ pass |
| `Swagger Avocado` ❌ | ✅ pass |

`Swagger BreakingChange`, `ModelValidation`, `SemanticValidation`, `ARM Modeling Review` and `ARM Auto SignOff` all pass. Only `Automated merging requirements met` is red, and that is purely the `PublishToCustomers` acknowledgement — correct to leave off while this is a draft.

### On the LintDiff suppressions — I checked these hard, because "silence the linter" is exactly the shape of a workaround

It isn't one. Verified mechanically:

```
package-2026-11 (new GA tag)                : 16 suppressions
package-2026-08-01-preview (in this PR)     : 16
package-2026-08-01-preview (on origin/main) : 16
PR's preview block identical to main's      : True
codes in GA tag but not in preview tag      : (none — all carried)
any suppression mentioning scenarioTemplate : NO
```

So they are genuinely *carried* from a block that already exists on `main`, not authored to make my B2 finding go away — and **the new surface introduces zero suppressions of its own**. That distinction is the whole ballgame, and it's the right outcome: the debt stays attached to the legacy surface that owns it (`targets` extension-resource `location`, the LRO POSTs, `Target.additionalProperties`), and `scenarioTemplates` arrives clean.

Worth stating for the record, though: a suppression that was uncontroversial on a **preview** tag carries more weight on a **stable** one, because GA freezes it. The `GuidUsage` and `AvoidAdditionalProperties` reasons both already anticipate that ("Approved by ARM API reviewer", "Our RP will release a V2 in the future"), so I'm satisfied — but that's the set I'd expect an ARM reviewer to ask about, and it's worth having those answers ready rather than discovered.

### Shape fixes — verified in the emitted swagger, not just the `.tsp`

| # | Item | Verified |
|---|---|---|
| H3 | `properties` required | `ScenarioTemplate` / `ScenarioTemplateVersion` now `required=[properties]`, matching `Action` / `ActionVersion` ✅ |
| H4 | Name patterns | both now `^[A-Za-z0-9\-._~]+$` ✅ |
| — | Properties split | parent `latestVersion,displayName,description,maxTimeoutDuration,resourceSelectors,parameters,actions`; child the same minus `latestVersion` ✅ |
| — | Distinct schemas | `ScenarioTemplate.properties` → `ScenarioTemplateProperties`; `…Version.properties` → `ScenarioTemplateVersionProperties` ✅ |
| — | `duration` | `type=string format=duration` on both models ✅ |
| — | Default tag | `tag: package-2026-11` ✅ |
| H2 (docs) | Macro namespaces | all four enumerated, `filters.` marked deprecated ✅ |

Two of these you took further than I asked, and both were the better call:

- **`latestVersion` over dropping the field.** You're right that dropping it from the child only fixed half the defect — the survivor still read as "the version" while meaning "the newest one". Renaming fixes the semantics, not just the collision.
- **Widening the `version` pattern too.** Applying my own argument symmetrically is correct: loosening a name pattern later is breaking *in both directions* per the guidelines, so the version axis — the one you actually publish along — had the same one-way door. Losing "this looks like a version" from the regex costs nothing real, since a malformed version is a 404 from the catalog regardless.

I also spot-checked the `duration` change, since it would break if any template put a macro there. It can't: across all 44 catalog files `maxTimeoutDuration` is a literal every time — **42×`PT3H`, 1×`PT1H`, 1×`PT30M`**. Your reasoning about `ScenarioAction.duration` needing to stay a bare `string` for `%%{parameters.duration}%%` is exactly right, and it's a good catch that the two fields differ. *(Minor: your reply says `PT3H` ×41; it's ×42. Immaterial — the "no macro can appear there" conclusion is what mattered and it holds.)*

### Both pushbacks accepted

**`resourceTargeting` as a sibling array — you're right, I was wrong.** I verified against `FilterTemplateMacroResolver.cs` and your argument is stronger than the code comment conveys:

```csharp
var include = targetingContext.ResourceTargeting?.Include;
// "locations" -> include?.Locations
// "zones" -> include?.Zones
// "physicalZones" -> include?.PhysicalZones ?? []
```

Resolution reads off the *configuration*, and the dimension vocabulary is a **closed set of exactly three** hardcoded in the resolver. So a per-template array would have restated a fixed enum 44 times or implied a constraint the template genuinely doesn't impose. Modelling it as a sibling of `resourceSelectors` would have been wrong. Withdrawn.

> **One small thing that falls out of this** — non-blocking. Your doc says the dimensions are *"(for example `zones`, `locations`)"*, but it's a closed set of three and **`physicalZones` is missing**. Since the vocabulary is closed, "for example" undersells what you can promise: naming all three exactly is strictly more useful, and `physicalZones` has non-obvious semantics (returns `[]`, not `null`, when unconfigured). Cheap to tighten while the file is open.

**Named list results — keeping them is fine.** The SDK/GW-codegen binding argument is legitimate, and consistency with `Action`/`ActionVersion` directly above in the same file matters more than idiom purity for a two-resource addition. More to the point, my actual technical concern was `GetCollectionResponseSchema` ambiguity from the shared properties model, and the split resolved it — verified that parent and child now resolve to distinct schemas at every level:

```
ScenarioTemplateListResult.value.items        -> ScenarioTemplate
ScenarioTemplateVersionListResult.value.items -> ScenarioTemplateVersion
```

**T003-8** — agreed and well-argued. `filters.` is a substring inside an opaque `string`, so there's no schema node to hang `@deprecated` on; prose is the only lever the ARM contract has. Combined with D-TPL-2 immutability, "document it and stop publishing it" is the only coherent answer.

### On GA-vs-preview (H1)

Your decision, and I'm not going to block it — the "no write surface to get wrong" argument is the right frame, and `ARM Modeling Review` + `ARMModelingAutoSignedOff` both pass. I'll register the residual honestly rather than pretend it's zero: the six corrections in this round are precisely the class of change that's free in preview and a Breaking Change Board item after GA. That cuts your way as much as mine — they got caught *before* merge, which is the system working — but it's the argument I'd expect an ARM reviewer to make, so it's worth having the answer ready. Thanks for updating the description to say it originates here; that was the part that actually mattered, since it's what a reviewer reads first.

### Open, not blocking

1. **`PublishToCustomers`** — the only red gate. Deliberate human ack; correct to hold while draft.
2. **`BreakingChange-Python-Sdk`** label is still applied even though `Swagger BreakingChange` and all six `SDK Validation` legs pass. Probably stale, but it needs a reviewer to clear it and it'll block merge if left.
3. **`physicalZones`** doc nit above.
4. **Hidden-template gate** — unchanged and not this PR's problem, but the GA contract correctly has no `visibility` field, so **EPIC-005 T005-9** remains the only thing keeping the 19 `Hidden` templates out of tenant-scoped LIST. Worth a hermetic test pinning LIST to the 25 `Public` versions before EPIC-006 rollout.

Nice turnaround on this — the replies did the thing I most want from a review round, which is to answer with verifiable specifics (the `service.yaml` destructive-rewrite mechanism, the resolver line numbers, the regenerated `required=[properties]`) rather than just "done".

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
```diff
@@ -0,0 +1,173 @@
...
+   */
+  @visibility(Lifecycle.Read)
+  description?: string;
+
+  /**
+   * ISO 8601 duration bounding how long a scenario created from this template may run
+   * (for example PT3H for three hours).
+   */
+  @visibility(Lifecycle.Read)
+  maxTimeoutDuration?: duration;
```
RenzoPrettoMS: Is this already supported by the BE and other templates? This property is new to me haha

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.tsp
```diff
@@ -0,0 +1,61 @@
...
+import "./scenarioTemplate.models.tsp";
+
+using Azure.ResourceManager;
+using TypeSpec.Versioning;
+
+namespace Microsoft.Chaos;
+
+@added(Microsoft.Chaos.Versions.v2026_11_01)
+@armResourceOperations
+interface ScenarioTemplates {
```
RenzoPrettoMS, line 14: I imagine there is a lot of overlap between `Scenario` and `ScenarioTemplates` is there enough overlap to re-use models? Or do we think there is going to be enough drift to warrant separate models? Re-using models can help us keep this coupled in case one is changed and the other isn't. However, if we think there will be intentional drift, then keeping them separate is ok!

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.models.tsp
juliagao3: Good catch — you pushed on the right thing, and it turned out to be more than "new property." I've removed it.

To your literal question: it isn't new. Every built-in template declares it and the backend requires it. But when I traced it end to end, the value is only ever *written* and range-checked (1 minute to 7 days) when the execution plan is built — nothing reads it to actually bound a run. Run timing is governed by per-action durations and stop conditions instead. So it was a declared cap with no enforcement behind it.

That's harmless as internal metadata, but the name promises a cap, and someone reading `maxTimeoutDuration: PT3H` would reasonably expect their run to stop at three hours. Shipping that in GA makes the wrong meaning permanent. Adding the property back once enforcement exists isn't a breaking change; correcting the meaning of one already in GA would be.

Removed in 635b422.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/scenarioTemplate.tsp
juliagao3: Good instinct, and we do reuse — just at the leaf level rather than the envelope.

`scenarioTemplate.models.tsp` imports `scenario.models.tsp` and uses `ScenarioAction` and `ScenarioParameter` unchanged. Those are the models carrying the real complexity (the action shape, the macro syntax, the parameter definitions), so they're exactly the ones where drift would hurt, and they can't drift — there's only one definition.

What's separate is the properties envelope, and that's because the two differ on every field they share:

| Property | `ScenarioProperties` | `ScenarioTemplateVersionProperties` |
| --- | --- | --- |
| `description` | writable | read-only |
| `parameters` | **required**, writable | optional, read-only |
| `actions` | required, **writable** | required, read-only |
| `provisioningState`, `createdFrom`, `version`, `recommendation` | read-only | — |
| `displayName`, `resourceSelectors` | — | read-only |

Three shared names, and all three differ in visibility or cardinality. Sharing them would mean a visibility-transform wrapper rather than a plain spread, which is more machinery than the three lines it would save.

The non-shared fields are also non-shared for a reason rather than by omission: the scenario-only ones are per-instance state (what provisioned, what it was created from, what version it's on), while the template-only ones are catalog metadata. `resourceSelectors` in particular has no scenario counterpart at all — a scenario gets its targeting from `resourceTargeting` on the configuration, not from the scenario resource.

So my read is that this is intentional divergence rather than drift. That said, if you'd still prefer tighter coupling I'm happy to pull `description` + `parameters` + `actions` into a shared mixin and apply the visibility transforms on each side — just say the word and I'll do it in this PR.

juliagao3:
## Breaking change justification — `ReadonlyPropertyChanged` on `createdFrom`

The `Swagger BreakingChange` check reports rule [1029 `ReadonlyPropertyChanged`](https://github.com/Azure/openapi-diff/tree/master/docs/rules/1029.md) against two already-published **preview** versions:

| Version | Status |
| --- | --- |
| `2026-05-01-preview` | `createdFrom` readOnly `true` → `false` |
| `2026-08-01-preview` | `createdFrom` readOnly `true` → `false` |
| `2026-11-01` (GA) | not flagged — new version, no prior content to compare |

### Why the change was made

`ScenarioProperties.createdFrom` records the scenario template version a scenario was created from. The service already accepts this value on create and has since the property was introduced — the resource provider deserializes it from the request body and persists it. Only the contract marked it read-only, which meant the value was silently dropped and "create a scenario from this published template version" could not be expressed through the API at all.

Marking it `Lifecycle.Create, Lifecycle.Read` (`x-ms-mutability: ["read", "create"]`) rather than fully writable is deliberate: the lineage pointer is intentionally not mutated on update, so a scenario cannot be re-pointed at a different template version after it is created.

### Why this is safe for the flagged preview versions

- **No behavior change.** The service accepts `createdFrom` on create in both preview versions today. This corrects the contract to describe what is already implemented; it does not change what the service does.
- **Relaxing, not restricting.** A property moving from read-only to settable-on-create widens what a request may contain. Existing clients that omit it are unaffected — it remains optional, and every previously valid request stays valid.
- **Preview versions.** Both flagged versions are previews and are within the window where contract corrections are expected.

The alternative — leaving it read-only in GA — would ship a stable contract that permanently cannot express template-derived scenario creation, which is the core scenario this API version exists to support.

Requesting the `BreakingChangeReviewRequired` sign-off on that basis.

specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-08-01-preview/openapi.json
```diff
@@ -6994,7 +6994,10 @@
         "createdFrom": {
           "type": "string",
           "description": "Resource ID of the template version this scenario was created from (optional).",
-          "readOnly": true
+          "x-ms-mutability": [
+            "read",
+            "create"
```
RenzoPrettoMS, line 6999: Is this intentional? Shouldnt we only ever read "createdFrom"?

juliagao3: Intentional — `createdFrom` is `read` + `create`, i.e. write-once, never updatable.

It has to be writable at create because it's the *input* that says which template to instantiate: the client PUTs the scenario with `createdFrom` set to a template version's resource ID, and the service resolves the template behind it. Read-only would leave no way to express that — the server has nothing to infer it from.

It's deliberately out of `update`, so it behaves as read-only for the rest of the resource's life.

PR #45312, opened 2026-08-06
[Storage] List Blobs Arrow Support For Rust TypeSpec

vincenttran-msft:
For proof of concept now:

- Omit all others
- Standup a list blobs that returns raw binary, unpaged, and accepts `accept`

TODO:
- [x] Implement for hierarchical
- [x] Iron out details for definition of this custom list blobs route


specification/storage/data-plane/BlobStorage/client.tsp
heaths: You could just put `@scope("rust")` on the method in `routes.tsp`.

heaths: You'll still need this, but change it to be internal e.g.,

```typespec
@@access(Storage.Blobs.Container.listBlobs, Access.internal, "rust");
@@access(Storage.Blobs.Container.ListBlobsResponse, Access.public, "rust");
```

The reason is that 1) you need to keep `ListBlobsResponse` as public, and 2) there is an optimization we made for Storage in the emitter for pageable models so you need the model to be returned by a pageable; otherwise, `values: Vec<BlobItem>` will changes to `values: Option<Vec<BlobItem>>` which is a breaking changes that will impact callers.

Alternatively, I suppose, you might be able to change your `Page` trait implementation to handle that change, treating `None` as "done".

heaths (changes requested):

heaths:
Might be simpler: https://github.com/Azure/azure-rest-api-specs/pull/45316

specification/storage/data-plane/BlobStorage/tspconfig.yaml
heaths, line 44: Why not just elide this line? None of the other languages have it.

specification/storage/data-plane/BlobStorage/client.tsp
heaths: If you're only adding this for rust, why are you renaming it? Even if you want to use it for other languages in the future, why not just call it the final thing rather than this bifurcated maintenance burden?

specification/storage/data-plane/BlobStorage/tspconfig.yaml
vincenttran-msft: New feature being added now to the typespec definition are now feature gated, and this helps to unblock merging things into the `main` typespec definition, but then having a fine granular control (i.e. version-gated) to manage what actually gets emitted in cases where the SDK feature release set is lagging behind typespec.

specification/storage/data-plane/BlobStorage/client.tsp
vincenttran-msft: Fair point, I totally was just abit heavy-handed with `client.tsp` changes since I was really just trying to get it to work (final blocker was multiple references so I just wanted a quick and dirty rename). But will take this feedback when we land on the final def here (and will duplicate this feedback when standing up `ListBlobsHierarchicalInternal`)

specification/storage/data-plane/BlobStorage/client.tsp
```diff
@@ -207,6 +207,23 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
 @@access(Storage.Blob.BlockBlob.upload, Access.internal, "rust");
 @@clientName(Storage.Blob.Blob.download, "download_internal", "rust");
 @@access(Storage.Blob.Blob.download, Access.internal, "rust");
+
+@@clientName(Storage.Blob.Container.listBlobs, "listBlobsXml", "rust");
+@@access(Storage.Blob.Container.listBlobs, Access.internal, "rust");
```
heaths: You need to mark the two `*Internal` operations as `Access.internal`, though you could also probably just do that in `routes.tsp` since they are already scoped to `rust`.

heaths (changes requested):

specification/storage/data-plane/BlobStorage/client.tsp
```diff
@@ -208,20 +208,10 @@ interface PageBlobClient extends Storage.Blob.PageBlob {}
...
-  Storage.Blob.Container.listBlobHierarchySegment,
-  "listBlobsHierarchicalXml",
-  "rust"
-);
-@@access(
-  Storage.Blob.Container.listBlobHierarchySegment,
-  Access.internal,
-  "rust"
-);
+@@scope(Storage.Blob.Container.listBlobHierarchySegment, "!rust");
```
heaths: IIRC, if you do this you also need to explicitly set a `@@usage` for both `ListBlobsResponse` and `ListBlogsHierarchicalResponse` like I did in my TypeSpec PR. `@@access` was not enough. If you get rid of the operations that use them, the models are elided but `@@usage` prevents that.

heaths (changes requested):

heaths (approved):
Actually, signing off to keep you unblocked since I'm OOF next week. But you'll need to change that regardless. At least this way someone else can approve and you can merge if you're ready next week. Still, I need to see changes to the SDK PR closer to what mine had before I had to remove them due to a verify difficult rebase or even straight merge.

specification/storage/data-plane/BlobStorage/client.tsp
vincenttran-msft: This is a good callout. Original my plan was to get rid of these `xml` variants (aka the original implementations) but this caused the following issues:

**Removing the operations**
This in itself wasn't problematic, but as you had mentioned, the models were elided: (`ListBlobsResponse`, `ListBlobsHierarchicalResponse`, `BlobItem`, `BlobProperties`, `BlobPrefix`, `BlobHierarchyList`, `ObjectReplicationMetadata`)

**Keeping them with `@@usage`**

1. **`Page` impls gone** : only emitted for models returned by a paged operation, so `Pager<ListBlobsResponse>` no longer satisfies `T: Page` (10 errors).
2. **Field shapes changed** : generated from the model alone, `blob_items` became `Option<Vec<BlobItem>>` and `hierarchical_list` became `Option<BlobHierarchyList>` (previously bare `Vec`/struct), breaking the Arrow decoder (3 errors).
3. **XML unwrap serde gone** : The flat response nests as
 `<Blobs><Blob>…</Blob></Blobs>`. The paged op generated a `Blob_itemsBlobItem::unwrap` (`deserialize_with`) to flatten that into `Vec<BlobItem>`. Without it the field is a bare `#[serde(rename = "Blobs")] Option<Vec<BlobItem>>`, which quick-xml treats as repeated `<Blobs>` elements rather than a wrapper of `<Blob>` children. So the XML fallback deserializes to an empty list with no error or panic, silently dropping every blob. And since it's a `#[derive(Deserialize)]` on generated code, the only workaround is a shadow struct manually kept in sync with `BlobItem`/`BlobProperties`.


Let me know your thoughts. It seems to be that the maintenance burden of handwriting back in these gaps may not be worth it versus just having some unreachable un-callable `xml` variants around in the generated client.

heaths (approved):

jalauzon-msft (approved):

seanmcc-msft (approved):

PR #45313, opened 2026-08-06
Data-plane API stewardship review - pool-assignment automation

samvaity:
## What this does

Automates data-plane API stewardship review requests using GitHub's native team review-request. When the `APIStewardshipBoard-ReviewRequested` label is applied to a data-plane PR, this workflow requests a review from the `@Azure/azure-data-plane-api-reviewers` team. GitHub's native review-request delegation then assigns one member of the team as the reviewer. When the `APIStewardshipBoard-SignedOff` label is applied by an authorized signer, the workflow clears the fulfilled `APIStewardshipBoard-ReviewRequested` label. It follows the existing `package-name-approval` / `protected-labels` `pull_request_target` label-trigger pattern.

## Behavior

- Request review on `APIStewardshipBoard-ReviewRequested`: request the team via `requestReviewers({ team_reviewers: [...] })`. Skips if the team is already a pending reviewer. No comment.
- Sign-off on `APIStewardshipBoard-SignedOff`: if applied by an authorized signer, clear the fulfilled `APIStewardshipBoard-ReviewRequested` label via the shared `removeLabelIfPresent`. No comment.
- The workflow subscribes to the `labeled` event only. Removing a label is a no-op by design, so it does not subscribe to `unlabeled`.

## Notes

- The reviewer team is a single constant (`REVIEWER_TEAM = "azure-data-plane-api-reviewers"`) in `assign-reviewers.js`.
- Native one-of-team delegation requires the team's `reviewRequestDelegationEnabled` setting to be on. Without it, GitHub requests the whole team instead of one member.
- Sign-off authorization still lists individual logins in `.github/protected-labels.yml`, manually mirrored to the team roster. The default Actions `GITHUB_TOKEN` is repo-scoped and cannot read org team membership at runtime. See #45487 for the rationale.

Part of #45451
Part of #45437


.github/cspell.yaml
```diff
@@ -28,6 +28,7 @@ words:
   - tsgs
   - unstub
   - vitest
+  - weidongxu
```
heaths: This might be untenable. Instead, you might just turn cspell off for a list of users. See below...

.github/protected-labels.yml
```diff
@@ -17,6 +17,16 @@ global-approvers:
...
 
+# Data-plane REST stewardship reviewers. Gated here as the only users who may apply the
+# sign-off label — the approval that clears the merge gate (protected-labels exists to stop
+# teams self-signing-off). This same list is the pool auto-requested/assigned when the
+# intake workflow applies `APIStewardshipBoard-ReviewRequested` (that label is a
+# machine-applied signal, not a human-gated action, so it is intentionally not protected;
+# see .github/workflows/src/data-plane-review/assign-reviewers.js). Add a login to grow it.
+APIStewardshipBoard-SignedOff:
+  - heaths
+  - weidongxu-microsoft
```
heaths, line 33: You could do something like this instead since usernames are often going to be full of "misspellings":

```suggestion
# cspell:disable
  - heaths
  - weidongxu-microsoft
# cspell:enable
```

See https://cspell.org/docs/Configuration/document-settings for more options.

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,352 @@
...
+  const content = await readFile(path, "utf8");
+  const config = /** @type {Record<string, unknown>} */ (yaml.load(content) ?? {});
+  const entry = config[label];
+
+  /** @type {unknown[]} */
+  let logins = [];
+  if (Array.isArray(entry)) {
+    logins = entry;
+  } else if (entry && typeof entry === "object") {
+    logins = Object.values(/** @type {Record<string, unknown>} */ (entry)).flatMap((value) =>
```
xirzec, line 57: nit: instead of using jsdoc everywhere for type checking, let's take advantage of the fact that Node 22+ supports .ts files directly. Especially since it's not clear to me that any infra is in place right now to actually valildate these type comments.

.github/protected-labels.yml
```diff
@@ -17,6 +17,16 @@ global-approvers:
   - samvaity
   - praveenkuttappan
 
+# Data-plane REST stewardship reviewers. Gated here as the only users who may apply the
+# sign-off label — the approval that clears the merge gate (protected-labels exists to stop
+# teams self-signing-off). This same list is the pool auto-requested/assigned when the
+# intake workflow applies `APIStewardshipBoard-ReviewRequested` (that label is a
+# machine-applied signal, not a human-gated action, so it is intentionally not protected;
+# see .github/workflows/src/data-plane-review/assign-reviewers.js). Add a login to grow it.
+APIStewardshipBoard-SignedOff:
```
xirzec, line 30: I get that we're repeating the casing of the label here, but it feels a little odd that this is the only key in the yaml file with capital letters

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * The pool is the authorized-users list of {@link POOL_LABEL}. That value is either a flat
+ * array of logins or an object keyed by plane (`{ "management-plane": [...], "data-plane":
+ * [...] }`); both shapes are flattened here.
+ *
+ * @param {string} [path]
+ * @param {string} [label]
+ * @returns {Promise<string[]>} Deduplicated list of reviewer logins.
+ */
+export async function loadReviewerPool(path = REVIEWERS_CONFIG_PATH, label = POOL_LABEL) {
```
xirzec: this method gets called twice, once in loadAuthorizedSigners and again in assignReviewers. We should cache the result after the first call to avoid hitting the disk twice in a row for the same file

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,352 @@
...
+    core.info(`Label '${targetLabel}' is not '${TRIGGER_LABEL}' or '${SIGNOFF_LABEL}', skipping.`);
+    return;
+  }
+
+  const { owner, repo, issue_number } = await extractInputs(github, context, core);
+  const prNumber = issue_number;
+
+  const pool = await loadReviewerPool();
+  if (pool.length === 0) {
+    core.warning("Data-plane review pool is empty; nothing to assign.");
```
xirzec: message should indicate this is API reviewers and complain about the specific empty key inside of protected-labels.yml as well as reference that file so someone debugging knows what needs to be fixed

xirzec (commented):
Some comments

.github/protected-labels.yml
samvaity: The flag wasn't from `protected-labels.yml` (that file is in cspell `ignorePaths`). It came from the test referencing real handles. I switched the test to dummy names, so no cspell exception is needed. Real handles stay only in the ignored `protected-labels.yml`.

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,280 @@
...
+  const targetLabel = payload.label?.name;
+
+  // Sign-off completes the review: clear the now-fulfilled review-request label. Assignees are
+  // intentionally left in place (in a 2-person assign-all pool the second reviewer lingering is
+  // acceptable). Removing the label via the default token does not re-trigger this workflow.
+  if (payload.action === "labeled" && targetLabel === SIGNOFF_LABEL) {
+    const { owner, repo, issue_number } = await extractInputs(github, context, core);
+    return await completeReviewOnSignOff({
+      github,
+      core,
```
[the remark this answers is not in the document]
samvaity: Handled in `completeReviewOnSignOff`: the signer is authorized before any cleanup. It proceeds only for a trusted bot (`ALLOWED_BOT_LOGINS`) or an authorized signer (pool plus `global-approvers`). Unauthorized sign-offs are skipped, leaving protected-labels to revert the bogus label. Covered by two tests.

.github/protected-labels.yml
samvaity: Yeah, this is a known issue. Current labels in repos, use mixed conventions:  PascalCase  ( BreakingChangeReviewRequired ,  ARMSignedOff ),  kebab-case  ( auto-sdk-build-fix ,  package-name-<lang>-approved ), and hybrid ( ARM-Review-Required ). There is a future work planned to standardize to  kebab-case  for new labels.

.github/workflows/src/data-plane-review/assign-reviewers.js
samvaity: `loadAuthorizedSigners` now reads the config once, factored into `loadConfig` plus `extractLogins`. Added a test asserting `readFile` is called once.

.github/workflows/src/data-plane-review/assign-reviewers.js
samvaity: The warning now names the key and file: `No data-plane API reviewers found: 'APIStewardshipBoard-SignedOff' is empty or missing in '.github/protected-labels.yml'. Add at least one reviewer login under that key.` Added a test for it.

.github/workflows/src/data-plane-review/assign-reviewers.js
samvaity: Keeping this as `.js` for consistency. All of `.github/workflows/src` is JS with JSDoc (no `.ts`), and `.github/tsconfig.json` states the folder should contain no TypeScript sources. The types are validated by `npm run lint:tsc` (`checkJs`). A repo-wide move to `.ts` could be a separate change if the team wants it.

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,217 @@
...
+export const SIGNOFF_LABEL = "APIStewardshipBoard-SignedOff";
+
+/**
+ * GitHub team requested as reviewer on intake. The team has code-review auto-assignment
+ * enabled, so GitHub replaces the team request with one of
+ * its members using the team's routing algorithm; the workflow never picks the individual,
+ * it only requests the team. Kept as a constant here; unifying assignment + sign-off
+ * authorization onto this team in config is tracked separately (issue #45487).
+ */
+export const REVIEWER_TEAM = "azure-data-plane-api-reviewers";
```
xirzec, line 31: It doesn't appear that this team has any access to the specs repository: https://github.com/orgs/Azure/teams/azure-data-plane-api-reviewers/repositories

This will cause the call  to `requestReviewers` to fail

.github/workflows/data-plane-review-assignment.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+permissions:
+  contents: read
+  pull-requests: write
+
+# Serialize runs per PR so racing label events do not interleave. Do not cancel
+# in-progress runs: a cancelled run could drop a team request mid-flight.
+concurrency:
+  group: data-plane-review-assignment-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
```
xirzec, line 27: Adding `queue :max` could help avoid problems where the same pr queues up multiple times

```suggestion
  cancel-in-progress: false
  queue: max
```

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -0,0 +1,217 @@
...
+    if (!includesLogin(authorizedSigners, signer)) {
+      core.info(
+        `'${signer}' is not authorized to apply '${SIGNOFF_LABEL}'; leaving '${TRIGGER_LABEL}' ` +
+          `for the protected-labels workflow to reconcile.`,
+      );
+      return;
+    }
+  }
+
+  const currentLabels = (payload.pull_request.labels ?? []).map((label) => label.name);
```
xirzec: might be worth re-fetching the current labels here to ensure that `APIStewardshipBoard-SignedOff` is still present instead of using what is cached off the event. Not a common case, but it is possible that the sign off is removed before this automation completes.

xirzec (approved):
Overall looks pretty good, left a few comments

.github/workflows/src/data-plane-review/assign-reviewers.js
samvaity: @JonathanGiles, @heaths, @weidongxu-microsoft could you add the repo access to the team. We need maintainer privileges to do so.

xirzec (approved):

PR #45327, opened 2026-08-07
[Foundry] Add invocations_moderation to RaiConfig for hosted-agent invocations content safety

amitbhave10:
Adds an optional `invocations_moderation` property to `RaiConfig`, plus its supporting model and unions, so the platform can locate user/agent text inside an **agent-defined** hosted-agent `invocations` request/response body and apply content-safety (AACS) moderation to it.

## Why

`RaiConfig` today only carries `rai_policy_name`. That is sufficient for the `/responses` protocol, whose body schema is fixed and platform-owned, so the platform already knows where the text lives. The hosted-agent **invocations** protocol is different: the request and response bodies are defined by the *agent author*, so the platform has no way to find the text to moderate.

This property lets the author declare where that text lives, which is the missing piece needed to moderate the invocations path.

## What an agent definition looks like

Given an agent whose request is `{"input": "...", "session_state": {...}}`, whose non-streaming response is `{"output": "...", "usage": {...}}`, and whose SSE stream emits `{"type":"text.delta","delta":"..."}`:

```jsonc
{
  "definition": {
    "kind": "hosted",
    "image": "contosoacr.azurecr.io/support-agent:v3",
    "cpu": "1",
    "memory": "2Gi",
    "container_protocol_versions": [{ "protocol": "invocations", "version": "v1" }],

    "rai_config": {
      "rai_policy_name": "contoso-strict",        // existing
      "invocations_moderation": {                 // NEW - all of this block
        "response_mode": "both",
        "input_content_type": "json",
        "output_content_type": "json",
        "input_paths":  ["$.input"],
        "output_paths": ["$.output"],
        "stream_selectors": [
          { "event_type": "text.delta", "text_field": "delta" }  // text_field may be omitted; defaults to "delta"
        ]
      }
    }
  }
}
```

Only `$.input` / `$.output` and the SSE `delta` values are sent to AACS - not `session_state`, `usage`, JSON keys, or SSE framing.

Minimal non-streaming agent (no `stream_selectors` needed):

```jsonc
"invocations_moderation": {
  "response_mode": "non_streaming",
  "input_paths":  ["$.messages[*].content"],
  "output_paths": ["$.choices[*].message.content"]
}
```

Non-JSON agent - the whole (size-capped) body is moderated and paths are ignored:

```jsonc
"invocations_moderation": {
  "response_mode": "non_streaming",
  "input_content_type": "text",
  "output_content_type": "text"
}
```

## What's added

| Type | Purpose |
|---|---|
| `RaiConfig.invocations_moderation` | New **optional** property |
| `RaiInvocationModeration` | The declaration itself (paths, stream selectors, content types, response mode) |
| `RaiSseTextSelector` | SSE `event_type` -> `text_field` selector for streaming output |
| `RaiInvocationContentType` | Extensible union: `json` \| `text` |
| `RaiInvocationMode` | Extensible union: `non_streaming` \| `streaming` \| `both` |

### Field reference

| Field | Type | Required | Default |
|---|---|---|---|
| `response_mode` | `RaiInvocationMode` | **yes** | - |
| `input_content_type` | `RaiInvocationContentType` | no | `json` |
| `output_content_type` | `RaiInvocationContentType` | no | `json` |
| `input_paths` | `string[]` | when `input_content_type=json` | - |
| `output_paths` | `string[]` | when `response_mode` is `non_streaming`/`both` and output is `json` | - |
| `stream_selectors` | `RaiSseTextSelector[]` | when `response_mode` is `streaming`/`both` and output is `json` | - |
| `stream_selectors[].event_type` | `string` | yes (per selector) | - |
| `stream_selectors[].text_field` | `string` | no (per selector) | `delta` |

Every default in the table is **declared in the spec** (`= RaiInvocationContentType.json`, `= "delta"`), so each one emits `"default"` in the generated schema rather than living only in prose.

Paths use a deliberately constrained grammar - `$.a.b`, `[0]`, `[*]` only. No filters, no recursive descent.

## Naming rationale

Names were chosen deliberately after review feedback; recording the reasoning so it does not get re-litigated.

**`invocations_moderation`** (not `invocations_text_map` / `invocations_text` (earlier drafts)). An earlier draft used a `_map` suffix. "Map" describes the implementation (a lookup from field to location) rather than the intent, and a `Map` postfix on a model type is not idiomatic. `invocations_text` was the next iteration, but it reads as though the property *holds* text when it actually declares *how this protocol gets moderated* - and one of its fields (`response_mode`) is not a text-location concern at all. `invocations_moderation` is the only name under which every current field genuinely belongs, and it leaves room for future non-extraction moderation knobs. The `invocations_` prefix is load-bearing: it scopes the block to the invocations protocol, since `rai_config` also applies to `/responses`, which needs no such declaration.

**`RaiInvocationModeration`, `RaiInvocationContentType`, `RaiInvocationMode`** - singular `Invocation*`, applied consistently across the three invocation-scoped types. All four new types (these plus `RaiSseTextSelector`) additionally carry the `Rai` prefix so they sort and read with the existing `RaiConfig`.

**Property names deliberately do not repeat their type names** (`response_mode: RaiInvocationMode`, not `rai_invocation_mode`). Type names are global across the spec - which is why they carry the `Rai` prefix - while property names are positionally scoped, so the container already supplies that context; `rai_config.invocations_moderation.rai_invocation_mode` would say "rai" and "invocation" twice each. Structurally the two cannot match anyway: `RaiInvocationContentType` is the value *domain* for **two** properties (`input_content_type`, `output_content_type`), which are distinguished by *role*. This matches the rest of `agents/models.tsp`, where no property repeats its type's full name and every prefixed type sheds that prefix at the property (`AgentState` -> `state`, `AgentVersionStatus` -> `status`, `AgentEndpointProtocol` -> `protocol`). `response_mode` is also more precise than `invocation_mode` would be: the request is always a single non-streaming body, so the field governs only which *output* gate runs. Naming the type `RaiResponseMode` instead was rejected as actively misleading, since `rai_config` also applies to the sibling `/responses` protocol.

**`input_paths` / `output_paths` are plural, deliberately.** The `[*]` wildcard already handles multiple *matches*, so plural is not about cardinality - it is about **polymorphism**. The path grammar has no union/alternation operator, so "the text is either here or there" cannot be expressed in a single path. This is common in practice: an OpenAI-compatible agent whose `content` is either a string or an array of parts needs `["$.messages[*].content", "$.messages[*].content[*].text"]`, and an agent accepting both a legacy and a current request shape needs one path each. All matches are concatenated. Restricting this to a single path would leave union-shaped bodies unmoderatable.

**`stream_selectors[].event_type` is singular, deliberately.** `stream_selectors` is already a list, so an earlier `event_types: string[]` was a second, redundant level of plurality: `{event_types: [a, b], text_field: f}` is exactly equivalent to two selectors, one per event type. Flattening it removes a nesting level and a validation rule while keeping everything expressible. Unlike the path fields, nothing is lost.

## Compatibility

- **Additive and non-breaking** - the new property is optional, and all new models/unions are new types.
- Rides the current preview version; no `@added` decoration, matching this spec's existing convention for additive fields.
- Omitting the property is valid and simply leaves the invocations path without content-safety moderation, which is the behavior today.
- Unions are extensible (`string` variant included), so new content types / response modes stay non-breaking.

## Contents

1. `src/agents/models.tsp` - the spec change.
2. Regenerated `openapi3/` artifacts (`v1` + `virtual-public-preview`, `.json` + `.yaml`) - pure additions.

`npx tsp compile .` succeeds with no errors or warnings.

## Consumer

The Azure ML "Vienna" Agents service reads this property on the hosted-agent invocations path and syncs these contracts via `tsp-location.yaml`. It currently carries a hand-written stopgap shim reading `invocations_moderation` out of the additional-properties bag; that shim is deleted once this lands and syncs.

## API review follow-ups (addressed)

- **`Rai` prefix** on all new model names, matching the existing `RaiConfig` (@dargilco).
- **Removed `allow_empty`** — its only effect was suppressing an internal service monitor counter, which is the wrong contract for a public API: customers never see our alerts. It was also redundant, since an author whose agent legitimately has text-free turns can already express that by omitting `invocations_moderation` entirely.
- **Service-side defaults declared in TypeSpec** — applied to *every* optional field that has one: `input_content_type`/`output_content_type` carry `= RaiInvocationContentType.json` and `text_field` carries `= "delta"`. Each emits a `"default"` in the generated schema.
- **Doc wording** — "When omitted, the service defaults to X" instead of "Defaults to X", so generated clients don't imply a client-side null default.
- **Dropped the "fail-open" jargon** from the `invocations_moderation` doc.
- **`InvocationResponseMode` -> `RaiInvocationMode`** (@ankitsul).
- **Bare `string` kept in the two unions**: they are reachable from `AgentVersionObject` (returned by the version GET/PUT/LIST routes) via `definition -> rai_config -> invocations_moderation`, so they appear in service *responses* and must stay extensible. 18 of 21 unions in `agents/models.tsp` follow the same pattern.


amitbhave10:
### CI triage: all red gates are pre-existing on `feature/foundry-release`

This PR is additive-only (one optional property `invocations_text_map` on `RaiConfig`, plus the
models it references) and regenerated OpenAPI3 artifacts. Diff is **+414 / -0**. I verified each
failing gate below; none are caused by this change.

**`TypeSpec Validation`** — fails on three folders this PR does not touch:

```
specification/ai-foundry/data-plane/Foundry/src/sdk-csharp-azure-ai-voice-agents
specification/ai-foundry/data-plane/Foundry/src/sdk-js-azure-ai-voice-agents
specification/ai-foundry/data-plane/Foundry/src/sdk-python-azure-ai-voice-agents
```

The failing rule is `SdkTspConfigValidation`, which reads only `tspconfig.yaml`:

```
- Failed to find "options.@azure-tools/typespec-java.emitter-output-dir".
- Failed to find "options.@azure-tools/typespec-ts.package-details.name".
- Failed to find "options.@azure-tools/typespec-python.emitter-output-dir".
```

Those projects **compile successfully**; only their `tspconfig.yaml` is non-conformant. This PR
modifies no `tspconfig.yaml` and no voice-agents file. Reproduced locally on a clean checkout of
`origin/feature/foundry-release` with zero changes applied — identical failure, confirming it is
pre-existing branch state. The project this PR actually affects passes:

```
$ npx tsv specification/ai-foundry/data-plane/Foundry
Compilation completed successfully.
Executing rule: Format          -> unchanged
Executing rule: SdkTspConfigValidation -> skipped
exit 0
```

**`Swagger Avocado`, `BreakingChange`, `LintDiff`, `ModelValidation`, `SemanticValidation`** —
these Swagger-2.0-oriented gates fail identically on already-merged PRs on this branch
(verified against #44772 and #45081, both merged red on the same five gates). This spec emits
OpenAPI 3, so these gates do not apply cleanly.

**Green:** `TypeSpec Requirement`, `TypeSpec Suppressions`, `TypeSpec APIView`, `SpellCheck`,
`Swagger PrettierCheck`, `Breaking Change(Cross-Version)`, `Protected Files`, `license/cla`,
`Automated merging requirements met`, and SDK Validation for Go / JS / Python / Rust.

No suppressions were added for any of the above.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+@doc("Declares where request/response text lives so content-safety guardrails can extract it.")
+model InvocationsTextMap {
+  @doc("How the REQUEST body is parsed. Defaults to `json`.")
+  input_content_type?: InvocationContentType;
+
+  @doc("How the RESPONSE body is parsed. Defaults to `json`.")
+  output_content_type?: InvocationContentType;
+
+  @doc("Author-declared response shape; drives which output gate runs and which fields are required.")
+  response_mode: InvocationResponseMode;
```
ankitsul: Just `InvocationMode`?

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
 model RaiConfig {
   @doc("The name of the RAI policy to apply.")
   rai_policy_name: string;
+
+  @doc("""
+    Author-declared map telling the platform where user/agent text lives in the agent-defined
+    invocations request/response bodies, so content-safety guardrails can extract and moderate it.
+    Optional; a rai_config without it leaves the invocations path without content-safety
+    moderation (fail-open).
```
dargilco: Not sure what "fail-open" means. I think it can be removed since the sentence before it describes it well.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+  event_types: string[];
+
+  @doc("The field on a matched event holding the text delta (defaults to `delta`).")
+  text_field?: string;
+}
+
+@doc("Declares where request/response text lives so content-safety guardrails can extract it.")
+model InvocationsTextMap {
+  @doc("How the REQUEST body is parsed. Defaults to `json`.")
+  input_content_type?: InvocationContentType;
```
dargilco: Specify service default while declaring the property: Add this at the end `= InvocationContentType.json;`

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+  @doc("SSE event `type` values that carry text.")
+  event_types: string[];
+
+  @doc("The field on a matched event holding the text delta (defaults to `delta`).")
+  text_field?: string;
+}
+
+@doc("Declares where request/response text lives so content-safety guardrails can extract it.")
+model InvocationsTextMap {
+  @doc("How the REQUEST body is parsed. Defaults to `json`.")
```
dargilco: Instead of "Defaults to `json`." say "When omitted, the service defaults to `json`". It's important to say that this is a service-side default, as the Python client adds a string "Defaults to None".

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+  input_content_type?: InvocationContentType;
+
+  @doc("How the RESPONSE body is parsed. Defaults to `json`.")
+  output_content_type?: InvocationContentType;
```
dargilco: same as above

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+    Author-declared map telling the platform where user/agent text lives in the agent-defined
+    invocations request/response bodies, so content-safety guardrails can extract and moderate it.
+    Optional; a rai_config without it leaves the invocations path without content-safety
+    moderation (fail-open).
+    """)
+  invocations_text_map?: InvocationsTextMap;
+}
+
+@doc("How an invocations request/response body is parsed to locate text for content-safety moderation.")
+union InvocationContentType {
```
dargilco: Please Add `Rai` prefix to all model names, so it's clear by the name what group of operations they are associated with

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -288,6 +288,72 @@ union AgentKind {
...
+  string,
```
dargilco, line 342: "sting" is only needed in unions that are used to define service-response (or both service-response and service-request). To allow the service to return a new value in the future that is not yet defined here. But if the union only defines input (request), then no need to add "string". Just list the supported values.

This comment applies to other unions in this PR.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Agreed — done. Renamed `InvocationResponseMode` -> `InvocationMode` (and it is now `RaiInvocationMode` after @dargilco's prefix comment below). Note the branch was a couple of commits behind when you looked; those are pushed now, so the diff also renames the property from `invocations_text_map` to `invocations_moderation`.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Done — dropped the `(fail-open)` parenthetical. You are right that it is internal jargon and the preceding sentence already conveys it.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Done. Renamed all four to match the existing `RaiConfig`:
- `InvocationModeration` -> `RaiInvocationModeration`
- `InvocationContentType` -> `RaiInvocationContentType`
- `InvocationMode` -> `RaiInvocationMode`
- `SseTextSelector` -> `RaiSseTextSelector`

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Done — now reads "When omitted, the service defaults to `json`." Thanks for the Python-client context, that is a good reason to be explicit. I applied the same wording to `RaiSseTextSelector.text_field` (was "(defaults to `delta`)") for consistency.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Done — same change applied to `output_content_type`.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Done — both properties now declare `= RaiInvocationContentType.json`, which emits `"default": "json"` into the generated schema.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
amitbhave10: Thanks — that rule makes sense, but I think these two unions do land in a service **response**, so I have left the `string` in for now. Please correct me if I am reading the graph wrong.

`RaiInvocationContentType` and `RaiInvocationMode` are reachable from `AgentVersionObject`, which is returned by the version routes (GET / PUT / LIST in `agents/routes.tsp`):

```
AgentVersionObject.definition -> AgentDefinition.rai_config -> RaiConfig.invocations_moderation
    -> RaiInvocationModeration.{input_content_type, output_content_type}  -> RaiInvocationContentType
    -> RaiInvocationModeration.response_mode                              -> RaiInvocationMode
```

So although the author *writes* these values on create/update, the service also *echoes them back* on read. By the criterion you described (extensible where the service may return a value the client does not yet know), that puts them in the "both request and response" bucket — a client on an older API version reading an agent authored with a newer content type would otherwise fail to deserialize.

That also matches the prevailing convention in this file: 18 of the 21 unions in `agents/models.tsp` carry the bare `string`.

Happy to drop it if you still disagree — just confirm and I will remove it from both.

amitbhave10:
Removed `allow_empty` from `RaiInvocationModeration` (commit 8312f811a7).

Rationale: its only effect was suppressing an internal service monitor counter, which is the wrong contract for a public API — customers never see our alerts, so the field asked an agent author to reason about Microsoft-internal telemetry and the doc string could not be expressed in customer-meaningful terms.

It was also redundant: an author whose agent legitimately has text-free turns can already express that by omitting `invocations_moderation` entirely, which leaves the invocations path without content-safety moderation.

The generated `openapi3` outputs were regenerated with `tsp compile` (verified byte-identical to the hand edit); the only change is the removal of the schema property. The matching service-side removal is Vienna PR 2203301.

glecaros (approved):

PR #45340, opened 2026-08-07
[Docs] Update `PathRenameMode` Description

vincenttran-msft:
Updates to include information about the scenario of renaming to an existing name in the destination.
>+++
When renaming a directory to a destination directory that already exists, the source directory is moved into the destination directory as a subdirectory.


vincenttran-msft:
Now should render as:
image

PR #45347, opened 2026-08-07
Preserve Confidential Ledger protocol-only API surface

JoshLove-msft:
## Summary
- disable convenience method generation for Azure.Security.ConfidentialLedger
- preserve the protocol-only API surface used by the legacy emitter
- avoid adding unintended CancellationToken and typed-model overloads during generator migration

Follow-up for Azure/azure-rest-api-specs#45346 and Azure/azure-sdk-for-net#61836.

--generated by Copilot


JoshLove-msft:
Closing because the new convenience overloads are intentional additive features for this migration. The model-factory rename will be documented separately as the breaking API change.

--generated by Copilot

PR #45377, opened 2026-08-10
ProgramEnrollment: retire 2026-03-01-preview, introduce 2026-08-10-preview

tthung1997:
Retires the `2026-03-01-preview` api-version for `Microsoft.ProgramEnrollment` and introduces `2026-08-10-preview` in its place.

Supersedes #45338, which applied the same redesign *in place* to `2026-03-01-preview` and was correctly flagged `VersioningReviewRequired`.

Mirrors Azure/azure-rest-api-specs#45378 for `Microsoft.BillingTrust`, so both resource providers retire their prior preview and land on the same `2026-08-10-preview` date.

## Why the version is retired rather than retained

`eduEnrollments` moves from a subscription/resource-group `TrackedResource` to a billing-account extension resource addressed via `{resourceUri}`. The two versions therefore share **no API paths** apart from `/operations`:

| | `2026-03-01-preview` | `2026-08-10-preview` |
|---|---|---|
| Resource kind | `TrackedResource` | `ExtensionResource` (proxy) |
| Scope | subscription / resource group | `{resourceUri}` → billing account |
| PATCH | yes | removed |
| LIST | `listBySubscription`, `listByResourceGroup` | single `{resourceUri}` LIST |
| `location` / `tags` | present | removed |

Because no paths carry over, retaining the previous version leaves API paths in the service that the new default tag cannot contain, which Avocado reports as `MISSING_APIS_IN_DEFAULT_TAG`. Retiring the previous version is therefore the representable shape, and it matches the service's actual behaviour: `2026-03-01-preview` is being removed from the ProviderHub manifest and will no longer be served.

The previous version was never deployed to public production and was not announced. A beta SDK was generated from it (`azure-mgmt-programenrollment==1.0.0b1`, `Azure.ResourceManager.ProgramEnrollment==1.0.0-beta.1`); a replacement will be generated from `2026-08-10-preview`.

## What this PR does

- Renames `preview/2026-03-01-preview/` to `preview/2026-08-10-preview/` and `examples/2026-03-01-preview/` to `examples/2026-08-10-preview/`.
- Updates `main.tsp` to the billing-account extension contract and bumps the `Versions` enum. The `client.tsp` import and its `@@clientName` customizations are preserved unchanged, so generated SDK type names are unaffected.
- Updates the `readme.md` tag to `package-preview-2026-08-10`.
- `service.yaml` is regenerated by `tsp compile` from the `Versions` enum.

## Contract in the new version

| Operation | Path |
|---|---|
| GET / PUT / DELETE | `/{resourceUri}/providers/Microsoft.ProgramEnrollment/eduEnrollments/{enrollmentName}` |
| LIST | `/{resourceUri}/providers/Microsoft.ProgramEnrollment/eduEnrollments` |
| Operations | `/providers/Microsoft.ProgramEnrollment/operations` |

`eduEnrollments` is a proxy extension resource. At runtime the service constrains `{resourceUri}` to `Microsoft.Billing/billingAccounts/{billingAccountName}`; the generic `{resourceUri}` scope follows the ARM-approved `Microsoft.BillingTrust` precedent.

## Related

- `RPSaaSMaster`: Azure/azure-rest-api-specs-pr#29948 (merged) — same contract, same version date
- `RPSaaSDev`: Azure/azure-rest-api-specs-pr#29941 (merged)
- `Microsoft.BillingTrust`: Azure/azure-rest-api-specs#45378 — same retirement, same version date

## Validation

- `npx tsp compile . --warn-as-error` — succeeds; `openapi.json` generated with `info.version: 2026-08-10-preview`
- `npx tsp format "**/*.tsp"` — clean
- `npx tsv <spec path>` — passes
- `Swagger Avocado` — passes in this shape

## Labels requested

This PR removes a published api-version, so it needs the same approvals granted to the sibling `Microsoft.BillingTrust` PR:

- `Versioning-Approved-Retired`
- `Approved-Suppression` — for the pre-existing `GuidUsage` suppression carried over unchanged (covers `tenantId` on `DomainGroup`, an Entra tenant identifier and therefore a GUID by definition)


tthung1997:
Thanks - the blocking finding was correct, and it has been fixed in `b1dd838`.

**Blocking: published version removed despite the description claiming otherwise.** Accurate, and the description was written for the original state rather than the state at `c549db4`. Sequence of events: the PR was opened as a pure add-alongside (`3fa8f28`), which matched the description. `Swagger Avocado` then failed with `MISSING_APIS_IN_DEFAULT_TAG`, because this is a scope migration - `2026-03-01-preview` exposes `/subscriptions/{subscriptionId}/...` paths while `2026-08-10-preview` exposes `/{resourceUri}/...`, so the two versions share no paths apart from `/operations` and the new default tag cannot contain the old version's APIs. The version folder was removed in an attempt to satisfy Avocado, which was the wrong trade: it introduced `VersioningReviewRequired` and contradicted the stated intent.

The branch has been reset back to add-alongside. `preview/2026-03-01-preview/` and `examples/2026-03-01-preview/` are once again byte-identical to `main`, and the `package-preview-2026-03-01` tag is restored. Avocado is instead handled with a `suppressions.yaml` entry scoped to that one file, following the existing precedent in `Microsoft.AlertsManagement` (older limited-private-preview version) and `Microsoft.HybridCompute` (operations intentionally retired during TypeSpec migration).

**Info: `GuidUsage` suppression lacks an approval label.** Noted. The suppression is carried over unchanged from the existing tag rather than newly introduced, and the `suppressions.yaml` entry above is also new - both will need `Approved-Suppression` from a reviewer. We are not self-applying them.

**Suggestion: async header inconsistency between PUT and DELETE.** Taking this back to the service team rather than answering from the spec alone. The asymmetry is inherited from the previous version rather than introduced here, but `final-state-via: azure-async-operation` on PUT versus `location` on DELETE is worth confirming against what the RP actually implements. Will follow up.

tthung1997:
Updated to match the approved shape of the sibling `Microsoft.BillingTrust` PR #45378, which has `Versioning-Approved-Retired` and `Approved-Suppression` applied by @MSEvanhi.

**What changed since the earlier blocking review:** the PR now retires `2026-03-01-preview` deliberately, and the description says so rather than claiming the version is untouched. The earlier inconsistency - description saying "completely untouched" while the diff removed the folder - is gone; the intent and the diff now agree.

**Why retirement rather than retaining the previous version:** this is a scope migration, so the two versions share no API paths apart from `/operations`. `2026-03-01-preview` exposes `/subscriptions/{subscriptionId}/...`; `2026-08-10-preview` exposes `/{resourceUri}/...`. Retaining the old version leaves API paths in the service that the new default tag cannot contain, which Avocado reports as `MISSING_APIS_IN_DEFAULT_TAG`. An add-alongside variant was attempted and required an Avocado suppression to work around exactly that; retiring is the representable shape and matches what the service is doing - `2026-03-01-preview` is being removed from the ProviderHub manifest and will no longer be served.

**Disclosure, since it is relevant to the retirement decision:** beta SDK packages were generated from `2026-03-01-preview` (`azure-mgmt-programenrollment==1.0.0b1`, `Azure.ResourceManager.ProgramEnrollment==1.0.0-beta.1`). The version was never announced and never deployed to public production, and a replacement SDK will be generated from `2026-08-10-preview`. Noting it explicitly rather than leaving it implicit. `Microsoft.BillingTrust` is in the same position (`azure-mgmt-billingtrust` is published) and was approved for retirement.

Requesting the same two labels: `Versioning-Approved-Retired` and `Approved-Suppression` (the latter for the pre-existing `GuidUsage` suppression carried over unchanged).

PR #45378, opened 2026-08-10
BillingTrust: retire 2026-03-17-preview, introduce 2026-08-10-preview

tekond:
## Summary
Retires the `Microsoft.BillingTrust` preview api-version `2026-03-17-preview` and introduces `2026-08-10-preview`. The API contract matches the private-repo (`RPSaaSMaster`) retired state; **public's SDK configuration is preserved**.

## Why we are retiring `2026-03-17-preview` (design change, no consumers)
The assessment resource was **redesigned**: the singleton `assessments/default` became the keyed `assessments/{assessmentName}` model (plus async DELETE and `$expandCustomerData` PII gating). Under ARM uniform-versioning (**RPC-BestPractice-14**), a contract-shape change requires a **new** api-version rather than mutating the published `2026-03-17-preview`.

Because **`2026-03-17-preview` has no consumers**, it is **retired outright** (folder + served version removed) rather than kept side-by-side. Preview / pre-release APIs are out of scope of the formal 36-month Azure retirement timeline (ARM `APIDeprecationPolicy.md`), so removal is direct.

## Changes
- `Versions` enum → `2026-08-10-preview`.
- Removed the old `preview/2026-03-17-preview/` (+ examples); added the new keyed `preview/2026-08-10-preview/` (+ examples). openapi is byte-identical to the validated private output.
- AutoRest readme → `tag: package-preview-2026-08-10` + `input-file: preview/2026-08-10-preview/openapi.json`; `service.yaml` → `2026-08-10-preview`.

## Preserved public SDK config (not reverted)
- `tspconfig.yaml` SDK naming kept as-is (`billing-trust` / `Billing.Trust` / `@azure/arm-billing-trust`, `sdk/billing`).
- `.cspell.yaml` kept.
- `client.tsp` kept and **reconciled to the new shape**: removed the two `@@clientName` entries for `AssessmentType` and `RegistrationRequirement` (models the new shape no longer defines); all other curated client names (including the `ProvisioningState` → `BillingTrustProvisioningState` de-collision) retained.

## Validation
`tsp compile` passes; the generated openapi is byte-identical to the private repo's validated output. (Full `tsv` / breaking-change / SDK validation run in CI.)


specification/billingtrust/resource-manager/Microsoft.BillingTrust/BillingTrust/readme.md
```diff
@@ -86,7 +106,7 @@ directive:
       - $.definitions.AssessmentProperties.properties.initialValues
     reason: |
       `initialValues` is an intentional transient seed parameter forwarded by
-      the parent RP at assessment creation. The values are forwarded to the
+      the parent resource provider at assessment creation. The values are forwarded to the
```
gary-x-li: **[NEW] 🟠 Warning** **[[TSP-Suppression-Accuracy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#4-suppressions-review-enforcement)]** — this `WriteOnlyProperties` suppression block (lines 102–121) — the reason is factually inaccurate after the `initialValues` schema change.

**Classification reasoning:** NEW — the base version carried this suppression text unchanged, but base `initialValues` was create-only, where the reason was accurate. This version's `+Lifecycle.Read` on `initialValues` is what introduces the mismatch.

The reason states that `initialValues` is a create-only seed that "cannot be returned on Read," is "mirrored by the `x-ms-mutability: [\"create\"]` annotation in the schema," and that "the absence of `initialValues` in the GET response cannot cause What-If false drift."

That no longer matches this version. `initialValues` is now declared `@visibility(Lifecycle.Create, Lifecycle.Read)` (`main.tsp` line 221), so the generated `openapi.json` emits `"x-ms-mutability": ["read","create"]` for `$.definitions.AssessmentProperties.properties.initialValues`, and the property's own doc (`main.tsp` line 216) says *"Returned on read."* The property **is** returned on read and its mutability is `["read","create"]`, not `["create"]`. Because the property is now readable, the `WriteOnlyProperties` rule likely no longer fires, so the suppression may also be an orphan.

Per the suppression-accuracy rule (reasons must be factually accurate; flag the inconsistency when the PR changes the relevant model), please do one of:

- **(a)** Remove the `WriteOnlyProperties` suppression if it no longer fires against the now-readable property; or
- **(b)** Correct the reason to describe the actual `["read","create"]` mutability — the structure is returned on read; only the customer PII within it (e.g. `domainNames`) is gated behind `?$expandCustomerData=true`; or
- **(c)** If a write-only seed was truly intended, revert to `@visibility(Lifecycle.Create)` only.

gary-x-li (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`bf09ce3`](https://github.com/Azure/azure-rest-api-specs/pull/45378/commits/bf09ce39b8b87a0404ddd7654cb43dac3d963e06). See the inline comment for the single finding (1 Warning, `[NEW]`).

Note: the API contract in this PR (`openapi.json`, `readme.md` suppressions, examples, `service.yaml`) is byte-identical to the private-repo PR Azure/azure-rest-api-specs-pr#29950; this finding applies to both.

Approval labels observed: `Approved-Suppression`, `Versioning-Approved-Retired`.

specification/billingtrust/resource-manager/Microsoft.BillingTrust/BillingTrust/readme.md
tekond: Thanks for reviewing. Fixed in 4a05c48c - removed the orphan  WriteOnlyProperties  suppression;  initialValues  is intentionally  ["read","create"]  (returned on read, PII gated by  $expandCustomerData ), so the rule no longer fires. Kept the visibility as-is by design.

gary-x-li (approved):

PR #45380, opened 2026-08-10
Remove minItems constraint from RunAfter.items (Chaos preview versions)

juliagao3:
### Summary

Removes `minItems: 1` from `RunAfter.items` in `2026-05-01-preview` and `2026-08-01-preview`.

### Why this is a spec bug fix

The spec declared `minItems: 1` on `RunAfter.items`, but the Chaos service backend **never enforced this constraint**. The generated backend baseline did not validate it, so clients have always been able to send `runAfter` with an empty `items` array successfully.

This divergence was latent until a recent model simplification regenerated validation directly from the spec. That regeneration began enforcing `minItems`, which broke the Azure portal for scenarios sending an empty `items` array.

This change **realigns the spec with actual, long-standing service behavior**. It does not change what the service accepts - it corrects the spec to describe what the service has always done.

### Impact

- **Constraint relaxation only.** No request that was previously valid becomes invalid.
- **Preview versions only.** Stable `2025-01-01` is unaffected - `RunAfter` does not exist in that version.
- No client that worked before stops working. Clients that were incorrectly blocked by spec-derived validation are unblocked.

### Requested approval

`BreakingChange-Approved-BugFix` - the spec described a constraint the service never implemented.

### Changes

- `scenario.models.tsp` - removed `@minItems(1)` from `RunAfter.items`
- Regenerated `openapi.json` for both preview versions

`Scenario.actions` `minItems: 1` is intentionally unchanged.

### DIFF
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-05-01-preview/openapi.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-05-01-preview/openapi.json
@@ -6273,7 +6273,6 @@
         "items": {
           "type": "array",
           "description": "Array of action dependencies.",
-          "minItems": 1,
           "items": {
             "$ref": "#/definitions/ActionDependency"
           },
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-08-01-preview/openapi.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-08-01-preview/openapi.json
@@ -6739,7 +6739,6 @@
         "items": {
           "type": "array",
           "description": "Array of action dependencies.",
-          "minItems": 1,
           "items": {
             "$ref": "#/definitions/ActionDependency"
           },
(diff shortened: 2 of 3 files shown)


juliagao3:

juliagao3:
Applying `Versioning-Approved-BugFix` per [pr-brch-deep](https://eng.ms/docs/products/azure-developer-experience/design/specs-pr-guides/pr-brch-deep), which allows the PR author to apply this label when the change corrects the spec to accurately describe actual service behavior.

**Justification:** The spec declared `minItems: 1` on `RunAfter.items`, but the Chaos service backend has never enforced it — the same payload succeeds via raw REST and fails only through spec-generated clients. A recent model simplification began generating enforcement of this phantom constraint, which is currently breaking the Azure portal. This change removes the constraint so the spec matches long-standing actual service behavior.

- Constraint relaxation only: no previously-valid request becomes invalid.
- The breaking change analyzer returned `BreakingChangeReviewRequired=false`; only `VersioningReviewRequired` was applied by automation.
- Both affected API versions (2026-05-01-preview, 2026-08-01-preview) are preview.
- Precedent: #43833 ("Removing max length constraint"), approved as `Versioning-Approved-BugFix`.

RenzoPrettoMS (approved):

PR #45396, opened 2026-08-11
Add model_selection_details TypeSpec for Model Router response details

smriti28:
Adds TypeSpec definitions for the `model_selection_details` response field on Chat Completions and Responses API, enabling visibility into Model Router routing decisions and fallback behaviour.

## Changes

- **CognitiveServices Chat Completions** (`common.tsp`, `chat_completions.tsp`): Added `ModelSelectionDetails`, `ModelRouterDetails`, `ModelRouterStep` models, `ModelRouterMode` and `ModelRouterStepResult` enums, and `modelSelectionDetails` field on `ChatCompletions` response.
- **Foundry Responses API** (`models.tsp`): Added equivalent types with snake_case and `model_selection_details` field on `AzureCreateResponseDetails`.
- **Preview gating** (`servicepatterns.tsp`): Added `model_router_controls_v1_preview` to `FoundryFeaturesOptInKeys` requiring `Foundry-Features: ModelRouterControls=V1Preview` header.

## SDK Spec
https://github.com/coreai-microsoft/foundrysdk_specs/pull/253


specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
```diff
@@ -93,6 +93,76 @@ namespace Azure.AI.Projects {
...
+
+    /** The model returned a transient server error. */
+    internal_server_error: "internal_server_error",
+
+    /** The model timed out. */
+    timeout: "timeout",
+
+    /** The model rejected the request payload. */
+    invalid_request: "invalid_request",
+  }
```
baywet: we're partially re-modeling http status here, this feels brittle

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
```diff
@@ -93,6 +93,76 @@ namespace Azure.AI.Projects {
...
+    `model`: string;
+
+    /** How this attempt concluded — whether the model served the request successfully or encountered an error. */
+    result: ModelRouterAttemptResult;
+  }
+
+  /** A single entry in the routing trace representing one inference turn. Each entry contains the latency and the ordered list of attempts for that turn. */
+  model RoutingTraceEntry {
+    /** Time in milliseconds spent in the routing decision for this turn. */
+    latency_ms: int32;
```
baywet: uint maybe? is there a case for negative values?

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
```diff
@@ -93,6 +93,76 @@ namespace Azure.AI.Projects {
...
+
+    /** Ordered list of routing attempts. First element is the initial selection; subsequent elements are fallback attempts. */
+    attempts: ModelRouterAttempt[];
```
baywet, line 165: what are the bounds here? is that array ever null? empty?

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
smriti28: The idea is to convey to the user on why a request failed for a llm. There can be multiple status codes there but we wanted to group the reasons so all >=400 and <499 except 429 is invalid_req category.  504 as timeout, >=500 as internal server error, 429 as too many requests. To keep into account for each http status and reason out seems hard maintenance. What is your suggestion here? The goal is to convey user that X llm failed with this reason so we are retrying to next best.

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
smriti28: yes will make it uint

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
smriti28: no, if the call has reached to model router service from apidispatcher and the routing layer call succeeds, it is never going to be empty. if model selection call fails even at routing layer (because of any dependency such as manifold issue etc) we have failover mechanism to decide models based on cost order.

baywet: would it be worth adding a min items decorator then?

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
baywet: the goal of http errors is to empower the user to recover by themselves to some degree.

For all 4XX errors, this usually means the client request was malformed to some degree. Providing the response body and headers would probably go a long way for the user to investigate.

For 429/503, are we planning to provide the retry-after header? so the client might implement some kind of backoff?

Also I'm not sure that grouping the status codes is a good idea, it "hides information" from the client, and now they are stuck guessing.

The feedback I'm trying to convey here is that between the grouping, and the lack of response body and headers, the customer is likely to feel this is a black box in case of failure, and won't be able to properly investigate the failure cause. I don't believe this is a great developer experience.

Model routing *could be compared to batching*, a pattern that's widely available in REST APIs. And if you look at [Microsoft Graph batching](https://learn.microsoft.com/en-us/graph/json-batching?tabs=http#example-json-batch-response) you'll notice that status codes, response headers and body are provided to the user for that exact same concern.

baywet (approved):
Thank you for making the changes!

specification/ai-foundry/data-plane/Foundry/src/openai/responses/models.tsp
```diff
@@ -93,6 +93,89 @@ namespace Azure.AI.Projects {
...
+    result: ModelRouterAttemptResult;
+  }
+
+  /** A single entry in the routing trace representing one inference turn. Each entry contains the latency and the ordered list of attempts for that turn. */
+  model RoutingTraceEntry {
+    /** Identifier of the output item/turn this routing decision is associated with. This is omitted for Chat Completions. */
+    output_id?: string;
+
+    /** Time in milliseconds spent in the routing decision for this turn. */
+    latency_ms: uint32;
```
glecaros:
```suggestion
    @encode(DurationKnownEncoding.milliseconds, uint32)
    latency_ms: duration;    
```

glecaros (approved):

PR #45419, opened 2026-08-11
Add deprecation notice for Workflow Agents

dargilco:
The same notice already appears here: https://learn.microsoft.com/en-us/azure/foundry/agents/concepts/workflow#migration-guide

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
@@ -279,7 +279,7 @@ union AgentKind {
   hosted: "hosted",
   external: "external",
   
-  #deprecated "Deprecated. Use Microsoft Agent Framework instead"
+  #deprecated "deprecated"
   workflow: "workflow",
 
   @removed(Versions.v1)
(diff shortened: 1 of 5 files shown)


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -359,7 +359,11 @@ model PromptAgentDefinitionTextOptions {
   format?: OpenAI.TextResponseFormatConfiguration;
 }
 
-@doc("The workflow agent definition.")
+/**
+* The workflow agent definition. Microsoft Foundry is retiring workflows on December 1, 2026. 
+* If you're looking to build new workflows, use Microsoft Agent Framework. To migrate existing workflows,
+* see the [Migration guide](https://learn.microsoft.com/azure/foundry/agents/concepts/workflow#migration-guide).
+*/
```
glecaros, line 366: should we also mark with the #deprecated directive?

dargilco: Good idea. Done!

glharper (approved):

PR #45442, opened 2026-08-12
Attestation: keep /certs and OpenID metadata endpoints unversioned

ksapathy:
## Summary

Restores `/certs` and `/.well-known/openid-configuration` to being **unversioned**, matching the behavior of every previously published Attestation API version.

## Problem

The TypeSpec migration declared both discovery endpoints using `Azure.Core.Foundations.Operation<>`, which injects a required `api-version` query parameter. That was an incidental side effect of the wrapper, not an intentional contract change.

As a result, generated clients began sending:

```
GET https://<instance>.attest.azure.net/certs?api-version=2025-06-01
```

where previously they sent:

```
GET https://<instance>.attest.azure.net/certs
```

This surfaced as a recording mismatch in the .NET SDK migration (Azure/azure-sdk-for-net#61741), but it affects every language generated from this TypeSpec.

## Evidence that unversioned is correct

`/certs` declares **zero** parameters in all four previously published versions, while `/policies/{attestationType}` declares `ApiVersionParameter` in those same documents:

| API version | `/certs` params |
| --- | --- |
| 2018-09-01-preview | 0 |
| 2020-10-01 | 0 |
| 2022-08-01 | 0 |
| 2022-09-01-preview | 0 |
| 2025-06-01 (pre-fix) | 1 |

The service registers both routes with an empty api-version alongside the versioned values, so the unversioned form is explicitly supported. These are discovery endpoints - clients call them to learn what the service supports, before negotiating an API version.

## Change

Declares the two operations explicitly rather than via `Azure.Core.Foundations.Operation<>`, **retaining the `default` error response** so error modeling is unchanged, and suppresses `operation-missing-api-version` with a rationale.

## Emitted swagger diff

Limited to removing the `api-version` parameter from the two routes. `200` and `default` responses are unchanged:

```diff
-        "parameters": [
-          { "$ref": "#/parameters/Azure.Core.Foundations.ApiVersionParameter" }
-        ],
+        "parameters": [],
```

Final emitted contract now matches 2020-10-01 and 2022-08-01 exactly:

| route | params | responses |
| --- | --- | --- |
| `/certs` | (none) | 200, default |
| `/.well-known/openid-configuration` | (none) | 200, default |
| `/policies/{attestationType}` | api-version | 200, default |

`tsp compile` completes with no warnings.

## Validation

- TypeSpec compiles cleanly (0 warnings)
- Emitted swagger diff reviewed and limited to the intended change
- Fix verified against the .NET SDK test suite: 20/20 passing on net8.0 and net10.0, with zero public API surface diff and no re-recording required

### DIFF
--- a/specification/attestation/data-plane/Attestation/examples/2025-06-01/Get_MetadataConfiguration.json
+++ b/specification/attestation/data-plane/Attestation/examples/2025-06-01/Get_MetadataConfiguration.json
@@ -1,6 +1,5 @@
 {
   "parameters": {
-    "api-version": "2025-06-01",
     "instanceUrl": "https://instance.attest.azure.net"
   },
   "responses": {
--- a/specification/attestation/data-plane/Attestation/routes.tsp
+++ b/specification/attestation/data-plane/Attestation/routes.tsp
@@ -235,16 +235,16 @@ interface SigningCertificates {
    * Retrieves metadata signing certificates in use by the attestation service
    */
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API - must use current operation signature for backward compatibility"
+  #suppress "@azure-tools/typespec-azure-core/operation-missing-api-version" "Unversioned discovery endpoint - clients call /certs before negotiating an API version, and every prior API version declares no parameters on this route"
   @summary("Retrieves the attestation signing keys in use by the attestation service")
   @route("/certs")
   @get
-  get is Azure.Core.Foundations.Operation<
-    {},
-    JsonWebKeySet & {
-      @header
-      contentType: SigningCertificatesContentType;
-    }
-  >;
+  get():
+    | (JsonWebKeySet & {
+        @header
+        contentType: SigningCertificatesContentType;
+      })
+    | Azure.Core.Foundations.ErrorResponse;
 }
 
 interface MetadataConfiguration {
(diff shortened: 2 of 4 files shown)


ksapathy:
Closing this — it is no longer needed for the work that motivated it, and merging it now would cost more than it returns.

## Why it was opened

`/certs` and `/.well-known/openid-configuration` use `Azure.Core.Foundations.Operation`, so the emitter injects an `api-version` query parameter into both. That broke 11 playback tests in the .NET SDK (Azure/azure-sdk-for-net#61741), whose recordings had no `api-version` on `/certs`.

## Why it is being closed

The .NET issue was resolved from the other direction. `Azure.Security.Attestation` 1.1.0 had to move its wire API version to `2025-06-01` anyway — `TdxVm` does not exist in the service's `AttestationType` before that version — and the tests were re-recorded. The recordings now contain `/certs?api-version=2025-06-01`, which matches what the client sends, so the SDK is green without this change.

That also inverts the cost/benefit here: merging this would remove the parameter and invalidate those fresh recordings, forcing another re-provision and re-record.

## The underlying observation still stands

This was never a runtime bug. The service registers both routes as `[ApiVersion("", ...)]`, i.e. genuinely unversioned, and accepts the parameter when sent. But the TypeSpec still describes them inaccurately, and every language generated from it will inject `api-version` on two endpoints that are conceptually unversioned — `/.well-known/openid-configuration` in particular is a discovery endpoint that should not be versioned.

So the fix remains correct; it is just spec hygiene rather than a blocker, and it is best folded into a future revision where a regenerate/re-record is happening regardless.

## Notes for whoever picks this up

The change is small, but two non-obvious things bit me:

1. Replacing the operation with a bare `op get()` also drops the `default` error response from the emitted swagger. Union `| Azure.Core.Foundations.ErrorResponse` back in to keep the emitted shape unchanged.
2. That then trips the `operation-missing-api-version` linter diagnostic, which needs an explicit `#suppress` with a justification.

The commits on this branch (`ef5c6a458be` through `b97a76d1195`) implement all of the above and were validated end to end against the .NET SDK — 20/20 tests passed locally with the parameter removed.

PR #45454, opened 2026-08-13
Add Microsoft.Chaos/scenarioTemplates to 2026-11-01

juliagao3:
## Summary

Adds `scenarioTemplates` to the `2026-11-01` stable API version — two tenant-level resources and four read-only operations.

Unlike the rest of the `2026-11-01` work, this is **not** a promotion of existing preview surface. It originates here, which is why it is split into its own PR.

| Operation | Path |
|---|---|
| `ScenarioTemplates_List` | `/providers/Microsoft.Chaos/scenarioTemplates` |
| `ScenarioTemplates_Get` | `/providers/Microsoft.Chaos/scenarioTemplates/{scenarioTemplateName}` |
| `ScenarioTemplateVersions_List` | `.../scenarioTemplates/{scenarioTemplateName}/versions` |
| `ScenarioTemplateVersions_Get` | `.../scenarioTemplates/{scenarioTemplateName}/versions/{version}` |

These are tenant-level, service-owned, immutable catalog resources — no write surface, no customer-supplied state. Addressing rule: **no version segment returns the latest version; a version segment returns that exact version.**

## Stacked on the GA promotion

| PR | Contents | Base |
|---|---|---|
| **#45311** | `2026-11-01` GA promotion (straight promotion of `2026-08-01-preview`) | `main` |
| **this PR** | `scenarioTemplates` only | `juliagao-microsoft-add-chaos-2026-11-ga` |

Review #45311 first. Once it merges, this PR retargets to `main` automatically and its diff is unchanged.

## Design notes

**Two resources, not one.** A published version is immutable — corrections ship as a new version rather than by editing an existing one, so a scenario that records the version it was created from always resolves to the same content. That requires each version to be individually addressable by URL.

**Two properties models.** `ScenarioTemplateVersionProperties` holds the blueprint content; `ScenarioTemplateProperties` spreads it and adds `latestVersion`. So `GET` on the template returns the same shape as `GET` on a version, plus a label saying which version it is — one client code path renders both.

**Macros are not resolved on read.** `actions` carry unresolved `{...}` macros; they are substituted when the scenario runs. Four namespaces are documented on the `actions` field, including `filters.` marked as the deprecated spelling of `resourceTargeting.include.`.

## Still open

Whether this should originate in a preview version (e.g. `2026-08-01-preview`) rather than directly in GA. Raised on the previous review threads and not yet settled.

## Review history

Earlier review feedback from @franklixuefei was addressed on #45311 before the split (commit `Address PR review feedback on scenarioTemplates`), covering: `PropertiesOptional`, resource name patterns widened to the Azure API Style Guide character set, the properties-model split, `maxTimeoutDuration` typed as `duration`, macro namespace documentation, and the `service.yaml` trailing newline. Those inline threads remain on #45311; the files they anchor to now live here.

Refs: AB#39047718


juliagao3:
Superseded: keeping the scenarioTemplates changes on #45311 so the existing review threads stay anchored. #45311 is now retargeted to the GA-promotion branch and contains only scenarioTemplates; the GA promotion moves to its own PR.

PR #45458, opened 2026-08-13
add serviceEntitlements resource for Microsoft.Security

gangaken:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

- [ ] New resource provider.
- [ ] Other, please clarify:

## Summary

Adds the `2026-09-01-preview` API for the generic `Microsoft.Security/billingPlans` proxy resource.

## Changes

- Adds required immutable `serviceType` and required updateable `state` properties.
- Adds an optional create/read-only `serviceProperties` extensibility bag.
- Models billing plans as non-regional configuration resources without location or tags.
- Adds read-only provisioning state and asynchronous PUT, PATCH, and DELETE operations.
- Adds GET, list-by-resource-group, and list-by-subscription operations.
- Adds TypeSpec examples, generated OpenAPI, service manifest, AutoRest package tag, and validation wiring.

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/security/resource-manager/Microsoft.Security/Security/preview/2026-09-01-preview/examples/BillingPlans/CreateOrUpdateBillingPlan.json
```diff
@@ -0,0 +1,57 @@
+{
+  "operationId": "BillingPlans_CreateOrUpdate",
+  "title": "Creates a billing plan",
+  "parameters": {
+    "api-version": "2026-09-01-preview",
+    "subscriptionId": "20ff7fc3-e762-44dd-bd96-b71116dcdc23",
+    "resourceGroupName": "myResourceGroup",
+    "billingPlanName": "MyBillingPlan1",
+    "resource": {
+      "location": "eastus",
```
gangaken, line 10: remove location.

gangaken: changed to ProxyResource

specification/security/resource-manager/Microsoft.Security/Security/readme.md
```diff
@@ -24,6 +24,8 @@ To see additional help and options, run:
 
 ``` yaml
 directive:
+  - suppress: MULTIPLE_API_VERSION
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-SUPPRESS-SCOPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppression-scoping-rules-rpc-suppress-scope)]** `specification/security/resource-manager/Microsoft.Security/Security/readme.md` - line 27 - The new `MULTIPLE_API_VERSION` suppression is blanket-scoped: it has neither `from:` nor `where:`, and the current Swagger Avocado run still reports `MULTIPLE_API_VERSION` for `package-composite-v3`.

**Classification reasoning:** Introduced in this PR at base SHA `ad48a6c`; the base readme has no `MULTIPLE_API_VERSION` directive.

**Approval context:** No `Approved-Suppression` label was observed at the session SHA. If this suppression has already been approved, ensure that label is applied and resolve this conversation; otherwise obtain suppression approval or address the underlying package-version violation.

**Suggested fix:** First fix the default package composition and verify Swagger Avocado no longer reports `MULTIPLE_API_VERSION`; only then is removing this directive safe. If the violation still fires, fix the underlying composition or use an approved suppression with specific `from:` and `where:` scopes rather than this blanket directive.

specification/security/resource-manager/Microsoft.Security/Security/readme.md
```diff
@@ -139,6 +141,16 @@ tag: package-composite-v3
...
 The following packages may be composed from multiple api-versions.
 
+### Tag: package-preview-2026-09-01-preview
+
+These settings apply only when
+`--tag=package-preview-2026-09-01-preview` is specified on the command line.
+
+```yaml $(tag) == 'package-preview-2026-09-01-preview'
+input-file:
+  - preview/2026-09-01-preview/billingPlans.json
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-Operations-V1-TAG](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#10a3-operations-api-must-be-in-package-tag-rpc-operations-v1-tag)]** `specification/security/resource-manager/Microsoft.Security/Security/readme.md` - line 151 - The new `package-preview-2026-09-01-preview` tag contains only `billingPlans.json`, so SDKs generated from this tag omit the required Microsoft.Security Operations API.

**Classification reasoning:** Introduced in this PR; the new package tag and billingPlans API version do not exist at base SHA `ad48a6c`.

**Suggested fix:** Add `2026-09-01-preview` to the sibling `OperationsAPI` TypeSpec project, regenerate `preview/2026-09-01-preview/security-Operations.json`, and include that same-version file beside `billingPlans.json` in this tag.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/tspconfig.yaml
```diff
@@ -0,0 +1,14 @@
+emit:
+  - "@azure-tools/typespec-autorest"
+options:
+  "@azure-tools/typespec-autorest":
+    omit-unreachable-types: true
+    emitter-output-dir: "{project-root}"
+    azure-resource-provider-folder: "resource-manager"
+    output-file: "../{version-status}/{version}/billingPlans.json"
+    arm-types-dir: "{project-root}/../../../../../common-types/resource-management"
+    emit-lro-options: "all"
```
gary-x-li: **[NEW] 🔴 Blocking** **[[Section 6.7](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#67-polling-behavior-and-final-state-via)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/tspconfig.yaml` - line 10 - `emit-lro-options: "all"` emits `final-state-via: azure-async-operation` on the PUT (generated `billingPlans.json` line 245). That tells SDKs to deserialize the operation-status monitor as the final `BillingPlan`, although the standard ARM PUT pattern must perform the final GET on the resource URI.

**Classification reasoning:** Introduced with this new TypeSpec project; no BillingPlans LRO configuration exists at base SHA `ad48a6c`.

**Suggested fix:** Remove `emit-lro-options: "all"` and regenerate the OpenAPI so standard PUT/PATCH/DELETE operations carry `x-ms-long-running-operation: true` without `x-ms-long-running-operation-options`. Add explicit final-location modeling only for an operation whose non-default final-result source is intentional.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
```diff
@@ -0,0 +1,44 @@
...
+interface BillingPlans {
+  /** Gets a billing plan. */
+  get is ArmResourceRead<BillingPlan>;
+
+  /** Creates or replaces a billing plan. */
+  createOrUpdate is ArmResourceCreateOrReplaceAsync<BillingPlan>;
+
+  /** Updates a billing plan. */
+  @patch(#{ implicitOptionality: false })
+  update is ArmCustomPatchAsync<BillingPlan, BillingPlanUpdate>;
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-Async-V1-08](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#62-async-patch-model-rpc-async-v1-08)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp` - line 34 - The async PATCH emits only `Location` and `Retry-After` on its `202` response (`billingPlans.json` lines 291-303); ARM requires both `Location` and `Azure-AsyncOperation`.

**Classification reasoning:** The PATCH operation is new in this PR; no BillingPlans operation exists at base SHA `ad48a6c`.

**Suggested fix:** Customize the async PATCH response to use `ArmCombinedLroHeaders<FinalResult = BillingPlan>` while retaining the required final `200` response, regenerate the OpenAPI, and add `Azure-AsyncOperation` to the example's `202.headers`. Note: removing `emit-lro-options: "all"` (see the tspconfig finding) likely restores both headers via the default combined-header template.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
```diff
@@ -0,0 +1,44 @@
...
+
+  /** Deletes a billing plan. */
+  delete is ArmResourceDeleteWithoutOkAsync<BillingPlan>;
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-Async-V1-09](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#63-async-delete-model-rpc-async-v1-09)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp` - line 37 - The async DELETE emits only `Location` and `Retry-After` on its `202` response (`billingPlans.json` lines 349-360); ARM requires both `Location` and `Azure-AsyncOperation`.

**Classification reasoning:** The DELETE operation is new in this PR; no BillingPlans operation exists at base SHA `ad48a6c`.

**Suggested fix:** Customize the async DELETE response to use combined ARM LRO headers, regenerate the OpenAPI, and add `Azure-AsyncOperation` to the example's `202.headers`. Note: removing `emit-lro-options: "all"` (see the tspconfig finding) likely restores both headers via the default combined-header template.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp
```diff
@@ -0,0 +1,77 @@
...
+   * Service-specific billing plan configuration. The schema is owned by the
+   * service identified by serviceType.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Billing plans provide a service-owned extensibility bag so new Microsoft.Security products can onboard without changing this API contract."
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  serviceProperties?: Record<unknown>;
+
+  /** The status of the most recent resource operation. */
+  @visibility(Lifecycle.Read)
+  provisioningState?: BillingPlanProvisioningState;
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-Async-V1-02](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#must-be-read-only) / [SCHEMA-REF-SIBLINGS](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#241-ref-must-not-have-sibling-keywords-schema-ref-siblings)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp` - line 64 - The emitted `provisioningState` is `{ "$ref": ..., "readOnly": true }` (`billingPlans.json` lines 452-455). OpenAPI 2.0 tooling ignores siblings of `$ref`, so the status is not reliably read-only to validators and generators.

**Classification reasoning:** The property and its emitted shape are new in this PR; no BillingPlans model exists at base SHA `ad48a6c`.

**Suggested fix:** Add `use-read-only-status-schema: true` under the `@azure-tools/typespec-autorest` options in `tspconfig.yaml`, then regenerate the OpenAPI and verify that the referenced provisioning-state definition itself is read-only.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp
```diff
@@ -0,0 +1,77 @@
...
+  @visibility(Lifecycle.Create, Lifecycle.Read, Lifecycle.Update)
+  state: BillingPlanState;
+
+  /**
+   * Service-specific billing plan configuration. The schema is owned by the
+   * service identified by serviceType.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Billing plans provide a service-owned extensibility bag so new Microsoft.Security products can onboard without changing this API contract."
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  serviceProperties?: Record<unknown>;
```
gary-x-li: **[NEW] 🟠 Warning** **[[PLCY005](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/policy-compatibility.md#plcy005-avoid-free-form-objects)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp` - line 60 - `serviceProperties` is modeled as `Record<unknown>` and emits unrestricted `additionalProperties`. The comment says this schema is service-owned, so it does not meet the pass-through-data exception; its fields cannot receive Azure Policy aliases or stable SDK types.

**Classification reasoning:** Introduced with the new BillingPlans model; no corresponding property exists at base SHA `ad48a6c`.

**Suggested fix:** Replace the bag with a discriminated, typed configuration model per `serviceType`, or define explicit service-owned properties. If arbitrary keys are truly user-defined pass-through data that the service neither validates nor interprets, document that narrower ownership contract instead.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace BillingPlansAPI;
+
+/** A Microsoft Defender for Cloud billing plan. */
+model BillingPlan is ProxyResource<BillingPlanProperties> {
+  ...ResourceNameParameter<
+    Resource = BillingPlan,
+    KeyName = "billingPlanName",
+    SegmentName = "billingPlans",
+    NamePattern = "^.+$"
```
gary-x-li: **[NEW] 🟠 Warning** **[[Section 2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp` - line 20 - `NamePattern = "^.+$"` accepts every non-empty character sequence and has no maximum length, so generated clients and ARM preflight cannot enforce the service's actual billing-plan name rules.

**Classification reasoning:** The resource-name parameter is new in this PR; no BillingPlans path exists at base SHA `ad48a6c`.

**Suggested fix:** Derive an allowlist and maximum length from the service implementation, for example `^(?![.-])[A-Za-z0-9_.-]{1,128}$` only if that exactly matches server-side validation.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/main.tsp
```diff
@@ -0,0 +1,31 @@
...
+using TypeSpec.Rest;
+using TypeSpec.Versioning;
+using Azure.Core;
+using Azure.ResourceManager;
+
+/**
+ * API specification for Microsoft Defender for Cloud billing plans.
+ */
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "The provider-wide Microsoft.Security operations endpoint is defined by the separate OperationsAPI TypeSpec project."
+#suppress "@azure-tools/typespec-azure-core/casing-style" "Microsoft.Security TypeSpec projects use PascalCase namespaces with an uppercase 'API' suffix as a service-wide convention."
```
gary-x-li: **[NEW] 🟠 Warning** **[[Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/main.tsp` - line 19 - The new casing suppression is justified only by saying other Microsoft.Security TypeSpec projects use the same convention. Suppression reasons must stand on their own; matching another project does not explain why the casing rule is inapplicable to this namespace.

**Classification reasoning:** Introduced with the new BillingPlans TypeSpec project; this suppression is absent at base SHA `ad48a6c`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the session SHA. If this suppression has already been approved, ensure that label is applied and resolve this conversation; otherwise obtain approval or address the warning.

**Suggested fix:** Rename the namespace to satisfy the casing rule, or replace the reason with a project-specific technical justification explaining why this namespace cannot comply and obtain TypeSpec suppression approval.

gary-x-li (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`2350457`](https://github.com/Azure/azure-rest-api-specs/pull/45458/commits/23504578bc6014104c1743c5389f2fc54bde2ad8). See inline comments for findings 1-9.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp
gary-x-li: **CI note:** Swagger LintDiff flags this exact construct as `AvoidAdditionalProperties` at **error** level on this commit (one of the new violations currently failing the LintDiff check), so it is a **blocking CI failure**, not only a Recommended (PLCY005) warning. Applying the suggested fix resolves both.

gangaken: removing this object alltogether

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp
gangaken: Added `use-read-only-status-schema: true` under the `@azure-tools/typespec-autorest` options in commit `ef5471b`, as suggested. Note: `provisioningState` here uses a custom `@lroStatus` union (`ServiceEntitlementProvisioningState`), so the emitter still renders `{ $ref, readOnly: true }` (readOnly is applied via `@visibility(Lifecycle.Read)`). This readOnly-sibling-of-`$ref` is the standard ARM emitter output for union/enum-typed provisioning state. Flagging for your review before resolving.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
gangaken: Fixed in commit `ef5471b`. `NamePattern` tightened from `^.+$` to `^[a-z0-9]{3,24}$` (Microsoft.Storage `storageAccounts` convention). Example resource names were lowercased to match.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/tspconfig.yaml
gangaken: Fixed in commit `ef5471b`. Set `emit-lro-options: "none"` in `tspconfig.yaml`. The regenerated `serviceEntitlements.json` no longer emits any `x-ms-long-running-operation-options` / `final-state-via`; PUT/PATCH/DELETE now carry only `x-ms-long-running-operation: true`. (Project renamed to `ServiceEntitlementsAPI`.)

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/main.tsp
gangaken: Fixed in commit `ef5471b`. Renamed the namespace to `ServiceEntitlementsApi` to satisfy the `casing-style` rule, and removed the suppression entirely (no longer needed).

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
gangaken: Fixed in commit `ef5471b`. The async DELETE now uses `LroHeaders = ArmCombinedLroHeaders & Azure.Core.Foundations.RetryAfterHeader`; the 202 returns both `Location` and `Azure-AsyncOperation`. The Delete example was updated with the `Azure-AsyncOperation` header.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/models.tsp
gangaken: Fixed. The free-form `serviceProperties: Record<unknown>` bag was removed entirely. Properties are now explicitly typed (`serviceType`, `state`, and read-only `provisioningState`). This also resolves the LintDiff `AvoidAdditionalProperties` error noted below.

specification/security/resource-manager/Microsoft.Security/Security/BillingPlansAPI/BillingPlan.tsp
gangaken: Fixed in commit `ef5471b`. The async PATCH now uses `LroHeaders = ArmCombinedLroHeaders<ArmOperationStatus, ServiceEntitlement> & Azure.Core.Foundations.RetryAfterHeader`, so the 202 returns both `Location` and `Azure-AsyncOperation` (plus `Retry-After`). The Update example was updated with the `Azure-AsyncOperation` header.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/examples/2026-09-01-preview/ServiceEntitlements/DeleteServiceEntitlement.json
```diff
@@ -0,0 +1,19 @@
+{
+  "operationId": "ServiceEntitlements_Delete",
```
gangaken, line 2: is there a way to not do delete?

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/models.tsp
```diff
@@ -0,0 +1,71 @@
...
+  /** The service entitlement is enabled. */
+  Enabled: "Enabled",
+
+  /** The service entitlement is disabled. */
+  Disabled: "Disabled",
+}
+
+/** The provisioning state of a service entitlement. */
+@lroStatus
+union ServiceEntitlementProvisioningState {
```
adityagoel123, line 26: Could we clarify the intended state-transition semantics before we finalize the ARM contract? 

1.) I see that state (Enabled/Disabled) and provisioningState (Accepted/Creating/Updating/Deleting/...) represent two different dimensions, which makes sense. It would be helpful to explicitly define the expected transitions and their relationship — for example, how a resource transitions from provisioning to Enabled, how Enabled → Disabled and Disabled → Enabled are handled, and what should happen if another update/delete request arrives while an operation is still in progress.

2.) This is particularly important for the backend implementation, since we will need to coordinate these states with Billing Platform success/failure semantics. Agreeing on this now will help ensure the ARM contract and backend state machine remain consistent.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/models.tsp
```diff
@@ -0,0 +1,71 @@
...
+  Deleting: "Deleting",
+
+  string,
+}
+
+/** Properties of a service entitlement. */
+model ServiceEntitlementProperties {
+  /** The service associated with the service entitlement. */
+  @visibility(Lifecycle.Create, Lifecycle.Read)
+  serviceType: string;
```
adityagoel123, line 48: What is the intended role of this property in the resource model? If the resource itself represents a service entitlement, is serviceType primarily identifying which product/service the entitlement belongs to (e.g. Perception, ISOC), or is it intended to carry additional service classification?

If this is the primary discriminator for the entitlement, should we consider constraining it through an extensible enum/union rather than an unrestricted string? This could give us stronger contract validation while still allowing additional service types to be introduced over time.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
```diff
@@ -0,0 +1,81 @@
...
+  listByResourceGroup is ArmResourceListByParent<ServiceEntitlement>;
+
+  /** Lists service entitlements in a subscription. */
+  listBySubscription is ArmListBySubscription<ServiceEntitlement>;
+
+  /** Lists service entitlements in the tenant. */
+  #suppress "@azure-tools/typespec-azure-core/no-openapi" "Static tenant-scoped route is required; ResourceListResult natively emits x-ms-pageable."
+  @list
+  @listsResource(ServiceEntitlement)
+  @route("/providers/Microsoft.Security/serviceEntitlements")
```
adityagoel123: Could we clarify the intended need/semantics for the tenant-scoped LIST API? The resource itself is modeled as a subscription/resource-group scoped ARM resource, while this endpoint is tenant scoped.

I'm fine with supporting tenant-level discovery if there is a concrete consumer requirement, but I'd like to understand the expected authorization/visibility semantics here — e.g. whether this returns entitlements across all subscriptions the caller can access, and whether that behavior is something our backend contract is prepared to support.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
```diff
@@ -0,0 +1,81 @@
...
+  @get
+  @armResourceList(ServiceEntitlement)
+  listByTenant(...ApiVersionParameter):
+    | ArmResponse<ResourceListResult<ServiceEntitlement>>
+    | ErrorResponse;
+
+  /** Gets the result of a long-running operation performed on a service entitlement. */
+  @get
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Security/serviceEntitlements/{serviceEntitlementName}/operationResults/{operationId}")
+  getOperationResult(
```
adityagoel123, line 56: Could we clarify the relationship between the custom operationResults/{operationId} endpoint and the standard Azure-AsyncOperation status endpoint returned by the LROs?

From the examples, the client is directed to the provider-level operationStatuses URL, while we also expose a resource-scoped operationResults GET. I want to make sure we have a clear and intentional story for which endpoint the ARM client is expected to poll and whether the custom operation-result API is required for any additional scenario beyond the standard LRO status.

If the standard ARM LRO pattern is sufficient for the scenarios we're supporting, could we consider relying on that rather than introducing a second operation-status contract?

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
```diff
@@ -0,0 +1,81 @@
...
+
+namespace ServiceEntitlementsApi;
+
+/** A Microsoft Defender for Cloud service entitlement. */
+model ServiceEntitlement is TrackedResource<ServiceEntitlementProperties> {
+  ...ResourceNameParameter<
+    Resource = ServiceEntitlement,
+    KeyName = "serviceEntitlementName",
+    SegmentName = "serviceEntitlements",
+    NamePattern = "^[a-z0-9]{3,24}$"
```
adityagoel123, line 20: Is the 3–24 lowercase alphanumeric constraint based on an existing ARM, Microsoft.Security, or backend/storage requirement? If this is a new resource type, I'd like to make sure we're not introducing a narrower naming restriction than necessary.

If the constraint is intentional for compatibility with an existing platform or backend requirement, it would be helpful to capture that rationale here.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
```diff
@@ -0,0 +1,81 @@
...
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using Azure.Core;
+using Azure.ResourceManager;
+
+namespace ServiceEntitlementsApi;
+
+/** A Microsoft Defender for Cloud service entitlement. */
+model ServiceEntitlement is TrackedResource<ServiceEntitlementProperties> {
```
adityagoel123, line 15: Is location expected to have actual resource semantics for a Service Entitlement, or is this primarily required because we're modeling this as a TrackedResource?

If the entitlement is logically tenant/subscription scoped rather than region-bound, it would be great to document here, what the location represents and whether any backend behavior or validation is expected to depend on it.

adityagoel123 (commented):
Thanks Gangadhar — I went through the ARM contract end-to-end. Overall the resource model and standard ARM CRUD surface look reasonable to me. I have a few questions mainly around locking down the state-machine semantics, LRO behavior, and the scope/semantics of the tenant-level LIST before we finalize the contract.

Nothing fundamentally blocking from my side at this point; I'd just like to make sure we get these semantics right now since this becomes the contract that our backend implementation will need to honor.

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-09-01-preview/examples/ServiceEntitlements/CreateOrUpdateServiceEntitlement.json
```diff
@@ -0,0 +1,57 @@
...
+          "environment": "test"
+        },
+        "properties": {
+          "serviceType": "Perception",
+          "state": "Enabled",
+          "provisioningState": "Creating"
+        }
+      },
+      "headers": {
+        "Azure-AsyncOperation": "https://management.azure.com/subscriptions/20ff7fc3-e762-44dd-bd96-b71116dcdc23/providers/Microsoft.Security/locations/eastus/operationStatuses/11111111-1111-1111-1111-111111111111?api-version=2026-09-01-preview"
```
gangaken: Recommended: we can use entitlementOperationStatuses

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/examples/2026-09-01-preview/ServiceEntitlements/DeleteServiceEntitlement.json
gangaken: For Tracked resource delete is necessary.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
gangaken: Scenario: Security Admin wants to look at all the securityEntitlements created in the tenant across subscriptions.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
gangaken: we will have resourceType Specific operationStatuses Endpoint. I will push a new iteration with updated examples based on ARM reviewer feedback. They recommended OperationStatuses as a child resource

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
gangaken: used the standard naming convention for storage accounts. I am not expecting a whole lot of resources for this resourceType to include Capitals, Underscore and hyphen. Keeping it simple. Let me know your thoughts.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/models.tsp
gangaken: Yes, serviceType is to represent the product that is being billed like Perception, defender, etc..

gangaken: I would like the union with modelAsString approach, but we need to keep updating api version for every product launch. With string, we would not need to rev up api versions, additionally for discovery we could add a catalog/discovery endpoint that lists all product types.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
gangaken: Security Products have location associated with them even in the syntex resource.

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-09-01-preview/serviceEntitlements.json
```diff
@@ -0,0 +1,570 @@
...
+  "tags": [
+    {
+      "name": "BillingPlans"
+    }
+  ],
+  "paths": {
+    "/subscriptions/{subscriptionId}/providers/Microsoft.Security/billingPlans": {
+      "get": {
+        "operationId": "BillingPlans_ListBySubscription",
+        "tags": [
```
gangaken: need tags and Location as it is Tracked resource

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/models.tsp
gangaken: these are provided by Nehal in the document

gangaken:
Requesting approval for the `Approved-Avocado` label to override `MULTIPLE_API_VERSION` on `package-composite-v3`.

The Microsoft.Security default package intentionally composes the latest API version for each resource type. This PR adds `preview/2026-09-01-preview/serviceEntitlements.json` to that existing composite so the new resource is included in the default Security SDK package. Removing it would omit the resource from the default package. An AutoRest `directive` cannot suppress Avocado and has therefore been removed.

The dedicated `package-preview-2026-09-01-preview` tag remains single-version. The Avocado error is limited to the intentionally multi-version default composite.

specification/security/resource-manager/Microsoft.Security/Security/readme.md
gangaken: published new iteration without the suppress

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/models.tsp
gangaken: Keeping `serviceType` as an open string because supported products will be dynamically discoverable through a catalog endpoint. The contract now documents the currently supported value, `Perception`, while remaining forward-compatible with additional service types.

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-09-01-preview/examples/ServiceEntitlements/CreateOrUpdateServiceEntitlement.json
gangaken: Superseded by the current resource-scoped `operationResults` LRO polling contract and updated examples.

specification/security/resource-manager/Microsoft.Security/Security/preview/2026-09-01-preview/examples/ServiceEntitlements/UpdateServiceEntitlement.json
```diff
@@ -0,0 +1,42 @@
...
+        "properties": {
+          "serviceType": "Perception",
+          "state": "Disabled",
+          "provisioningState": "Succeeded"
+        }
+      }
+    },
+    "202": {
+      "headers": {
+        "Location": "https://management.azure.com/subscriptions/20ff7fc3-e762-44dd-bd96-b71116dcdc23/resourceGroups/myResourceGroup/providers/Microsoft.Security/serviceEntitlements/myserviceentitlement1/operationResults/22222222-2222-2222-2222-222222222222?api-version=2026-09-01-preview",
```
gangaken, line 37: check why location is needed when we already have Azure-AsyncOperation.

psah434:
required check swagger avocado is failing. Please fix

gangaken:
> required check swagger avocado is failing. Please fix

@psah434 
Microsoft.Security intentionally uses this composite tag to generate one shared SDK from the latest API version of each independently versioned resource type. This PR adds `serviceEntitlements.json` at `2026-09-01-preview` to that existing composite. Aligning every Microsoft.Security resource to one API version is outside this resource owner’s scope, while excluding Service Entitlements would omit it from the shared SDK.


The finding is therefore an expected consequence of the existing Microsoft.Security packaging model, not an API-contract defect. No effective readme directive can suppress Avocado, so requesting the `Approved-Avocado` override label.
{"code":"MULTIPLE_API_VERSION","message":"The default tag contains multiple API versions swaggers.","readMeUrl":"/home/runner/work/azure-rest-api-specs/c93b354fd9c14905bb574a8834c4d69b/specification/security/resource-manager/Microsoft.Security/Security/readme.md","tag":"package-composite-v3","path":"/home/runner/work/azure-rest-api-specs/c93b354fd9c14905bb574a8834c4d69b/specification/security/resource-manager/Microsoft.Security/Security/readme.md","level":"Error"}

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
psah434: On the tenant-scoped `listByTenant` (`GET /providers/Microsoft.Security/serviceEntitlements`): the "Security Admin wants to see all entitlements in the tenant" scenario is a valid requirement, but it does not justify this API shape, and it raises an authorization concern that needs explicit ARM + security review sign-off before it ships.

`serviceEntitlements` is a resource-group-scoped `TrackedResource`. A collection GET with no subscription or resource group in the path is authorized by ARM at the **tenant root scope (`/`)**, not per-subscription. If the RP aggregates entitlements across every subscription in the tenant without applying per-subscription authorization itself, it would bypass the per-subscription / per-RG RBAC boundary that `ListBySubscription` and `ListByResourceGroup` (both already present) enforce — which is exactly why this needs sign-off. In practice either the caller must hold a tenant-root role (very few principals do), or the RP has to implement bespoke cross-subscription authorization and per-item filtering — a custom security boundary that ARM does not enforce for you. For a security product, exposing the existence and state of entitlements across subscription boundaries is a real confidentiality risk.

The sanctioned mechanism for a cross-subscription "all X in the tenant" view is **Azure Resource Graph**, which enforces per-resource read RBAC at query time and already indexes tracked resources.

**Recommendation:** remove `listByTenant` and rely on `ListBySubscription` + ARG for the tenant-wide view, or keep it only with explicit ARM API review + security review sign-off that documents the authorization model. Please treat this as blocking pending that sign-off.

gangaken: removed list by tenant.

specification/security/resource-manager/Microsoft.Security/Security/ServiceEntitlementsAPI/ServiceEntitlement.tsp
```diff
@@ -0,0 +1,70 @@
+import "@typespec/http";
```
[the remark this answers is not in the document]
gangaken: Several other resource types under microsoft.security use the same pattern of operationstatus/operationresults and we are following the same for new resource type.

	"resourceProviderAuthorizationRules": {
		"asyncOperationPollingRules": {
			"authorizationActions": [
				"Microsoft.Security/governanceRules/operationResults/read",
				"Microsoft.Security/serviceEntitlements/operationResults/read",
				"Microsoft.Security/serverVulnerabilityAssessments/operationResults/read",
				"Microsoft.Security/securityConnectors/devops/operationResults/read",
				"Microsoft.Security/defenderForStorageSettings/operationStatus/read"
			],
			"authorizationRule": "ProtectedAsyncOperationPolling"
		}
	},

psah434 (approved):

PR #45468, opened 2026-08-13
Add DPP Cross-Tenant Restore (CTR) feature - API version 2026-08-31-preview

zubairabid:

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


gary-x-li:
Please resolve branch conflicts and fix LintDiff errors first.

zubairabid:
> Please resolve branch conflicts and fix LintDiff errors first.

Hi @gary-x-li I've resolved these.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
```diff
@@ -49,6 +49,25 @@ semantic-validator: true
...
+input-file:
+  - preview/2026-08-31-preview/dataprotection.json
+suppressions:
+  - code: GetCollectionOnlyHasValueAndNextLink
+    from: dataprotection.json
+    reason: DPP list models compose value + DppResourceList.nextLink via allOf and satisfy the value+nextLink condition; the LintDiff rule does not follow the allOf composition (tool false-positive). Consistent with the existing suppression on prior DPP tags (package-preview-2024-02).
+  - code: AvoidAdditionalProperties
+    from: dataprotection.json
+    reason: There are objects that need a generic key-value pair in contract.
+  - code: PostResponseCodes
```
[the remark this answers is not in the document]
zubairabid: Requesting approval as these are existing suppressions that the new tag inherits.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+  is Azure.ResourceManager.ProxyResource<CrossTenantVaultMappingProperties> {
+  ...ResourceNameParameter<
+    Resource = CrossTenantVaultMapping,
+    KeyName = "mappingName",
+    SegmentName = "crossTenantVaultMappings",
+    NamePattern = ""
+  >;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation" "Cross-tenant vault mappings are removed via the POST 'remove' action (RBAC separation-of-duties + Resource Guard / MUA body semantics), not a DELETE verb."
```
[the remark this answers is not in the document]
zubairabid: As mentioned in review comment, remove is modelled as `POST` rather than `DELETE` as it's a ResourceGuard-guarded operation that carries `resourceGuardOperationRequests` in the request body (RBAC separation of duties). This is how the operation has been modelled in the service. We have made an identical design decision in our sister service, which was reviewed and merged under PR #41694 .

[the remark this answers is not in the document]
psah434: **[NEW] 🟠 Warning** — **[[`no-resource-delete-operation`](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#25-arm-resource-definitions-review-enforcement) / [TSP §4.1 suppression accuracy](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `CrossTenantVaultMappingResource.tsp` — remove action (line 150-151), suppression at line 120.

**Classification:** [NEW] — `CrossTenantVaultMapping` is introduced in this PR; not present in the previous version (`stable/2026-06-01`).

**Issue:** The `remove` action is modelled as `ArmResourceActionSync<CrossTenantVaultMapping, void, OkResponse>` — a `void` request body. The emitted `POST .../remove` has **no body parameter** and returns 200-empty. But the justification for POST-instead-of-DELETE claims a request body exists:

- The `no-resource-delete-operation` suppression reason (line 120) cites _"Resource Guard / MUA **body semantics**"_.
- The PR reply on this thread states the remove _"carries `resourceGuardOperationRequests` in the **request body**"_.

Both are contradicted by the actual contract — `resourceGuardOperationRequests` lives in the resource **properties** (set at PUT time), not the remove body. **The PR even contradicts itself:** the `readme.md` line 68 `PostResponseCodes` suppression reason correctly states the remove _"returns 200 with **no body**"_, the opposite of the tsp:120 claim.

Per TSP §4.1, suppression reasons **MUST be factually accurate**. So either (a) the `remove` action is missing the MUA request body it is supposed to carry (a functional gap for a Resource-Guard-protected operation), or (b) the justification and suppression reason are inaccurate and should be corrected.

**Suggested fix:** Either add a request body to the `remove` action carrying `resourceGuardOperationRequests` (consistent with other DPP MUA-guarded operations such as `StopProtection`), or correct the line 120 suppression reason and the design justification so they do not claim body semantics the operation does not have.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+interface DppCrossTenantVaultMappings {
+  /**
+   * Gets a cross-tenant vault mapping by name from the target vault.
+   */
+  get is ArmResourceRead<CrossTenantVaultMapping, Error = CloudError>;
+
+  /**
+   * Creates or updates a cross-tenant vault mapping under the target vault.
+   */
+  createOrUpdate is ArmResourceCreateOrReplaceSync<
```
psah434, line 149: **[NEW] 🔵 Suggestion** — **[[RPC-Async design consideration](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md)]** `CrossTenantVaultMappingResource.tsp` — createOrUpdate (line 132) + provisioningState (line 73).

**Classification:** [NEW] — introduced in this PR; not present in the previous version (`stable/2026-06-01`).

**Issue:** `createOrUpdate` is `ArmResourceCreateOrReplaceSync` (synchronous PUT, emits 200/201, no `x-ms-long-running-operation`), yet the resource exposes a read-only `provisioningState`. Given the cross-tenant semantics (Active/Inactive states, a separate source-side status check), please confirm creation is genuinely synchronous. If a cross-tenant handshake happens during create, the PUT should be an async LRO. _(Low confidence — sync-create with `provisioningState` is permitted; this is a clarification.)_

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+ */
+@added(Versions.v2026_08_31_preview)
+@parentResource(BackupVaultResource)
+model CrossTenantVaultMapping
+  is Azure.ResourceManager.ProxyResource<CrossTenantVaultMappingProperties> {
+  ...ResourceNameParameter<
+    Resource = CrossTenantVaultMapping,
+    KeyName = "mappingName",
+    SegmentName = "crossTenantVaultMappings",
+    NamePattern = ""
```
psah434: **[NEW] 🔵 Suggestion** — **[[arm-resource-name-pattern / TSP §2.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `CrossTenantVaultMappingResource.tsp` — line 116.

**Classification:** [NEW] — introduced in this PR; not present in the previous version (`stable/2026-06-01`).

**Issue:** The new resource uses `NamePattern = ""` (empty), so `mappingName` has no `@pattern`/length constraint in the emitted swagger. Consistent with existing DPP resources (`vaultName` is also unconstrained), so low priority — consider adding a name pattern for the new resource. _(Low confidence.)_

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
zubairabid: @pshah434 Confirmed: creation is genuinely synchronous in the service case. We will be keeping this definition unchanged - please resolve.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
zubairabid: Noted. We're updating this to use the NamePattern used in the equivalent recoveryservices implementation. Please resolve.

zubairabid:
Can we get the ARMSignedOff label? Not merging yet, but we will need the label for manifest rollout.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
```diff
@@ -0,0 +1,10704 @@
...
+            "description": "The name of the CrossTenantVaultMapping",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Azure operation completed successfully.",
+            "schema": {
+              "$ref": "#/definitions/BackupInstanceResourceList"
```
sandipsh: **[NEW] 🔴 Blocking** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json` - line 3166 - The new cross-tenant backup-instance collection response exposes a secret-capable credential model.

**Classification reasoning:** Introduced in this PR. `BackupInstancesFromDppCrossTenantVault_List` is a new cross-tenant operation that is absent from the previous version at base SHA `120cdc984d09afe0cf1d79b29ecfce1bc3bc90b0`.

**Issue:** The response returns `BackupInstanceResourceList`; its items reach `BackupInstance.datasourceAuthCredentials` at line 6492, described as credentials, and the `SecretStoreBasedAuthCredentials` variant reaches `SecretStoreResource.value` at line 9966. This makes a credential/secret-capable model available through a new cross-tenant LIST route.

**Suggested fix:** Use a cross-tenant response model that excludes `datasourceAuthCredentials`, or return a non-secret reference. Secret values must be omitted or null in GET/LIST responses.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
```diff
@@ -0,0 +1,10704 @@
...
+          },
+          {
+            "name": "$filter",
+            "in": "query",
+            "description": "OData filter options.",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "$skipToken",
```
sandipsh: **[NEW] 🔴 Blocking** **[[RPC-Get-V1-15](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#22-tracked-resource-requirements)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json` - line 3509 - The new recovery-points collection GET exposes `$skipToken` as an explicit query parameter.

**Classification reasoning:** Introduced in this PR. `RecoveryPointsFromDppCrossTenantVault_List` is a new cross-tenant operation absent from the previous version at base SHA `120cdc984d09afe0cf1d79b29ecfce1bc3bc90b0`.

**Issue:** ARM collection GET operations may only use `api-version` and OData `$filter`. This operation also declares `$skipToken`, despite already returning an `x-ms-pageable` `nextLink`.

**Suggested fix:** Remove `$skipToken` from the TypeSpec list parameters and use the returned `nextLink` for continuation.

zubairabid: ### Context
**What this feature is.** Cross-Tenant Restore (CTR) lets a customer restore backup data from a vault in a *different Entra tenant*, through a first-class secured resource — a **`crossTenantVaultMapping`** — that the customer creates on their **target** vault pointing at a specific **source** vault. That mapping is the trust boundary: creating it is gated by a **linked-access-check on `sourceVaultId`** (caller must prove access to the source vault) **and Resource Guard / MUA approval**. Nothing is reachable cross-tenant until this authorized, auditable handshake exists.

**What the "new" operations are.** Once a mapping exists, the operations under `.../crossTenantVaultMappings/{name}/...` (list backup instances, list/get recovery points, restore, jobs) are **not new capabilities** — they are the **existing vault operations invoked across the authorized mapping** (in the service, literally `SwapToCrossTenantSourceVaultForRead` + the same `List/GetBackupInstancesInternal` / `ListRecoveryPointsInternal`). In the spec they reuse the **exact same request/response models and parameters** (via `LegacyOperations`), so the cross-tenant response is **byte-for-byte identical** to the in-tenant one — we intentionally do **not** introduce a divergent model or change the response.

### Response
This cross-tenant recovery-points list is the existing **`RecoveryPoints_List`** operation reached across an authorized vault mapping (see context) — not a new operation shape. It reuses that operation's exact list signature, including `$filter` + `$skipToken`, which `RecoveryPoints_List` (and `FetchSecondaryRecoveryPoints_List`) have declared since the **first DataProtection api-version, `2021-01-01`**, and which the service wire accepts. Dropping `$skipToken` only on the cross-tenant route would diverge from the operation it mirrors and from the service contract. If `$skipToken` were to be removed, that would be a shared, breaking change to the existing (shipped) DPP recovery-point lists across all versions — not specific to these cross-tenant routes.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
zubairabid: The cross-tenant backup-instance list is the existing **`BackupInstances_List`** operation reached across an authorized vault mapping (see context). It returns the **same `BackupInstance` model** returned since api-version **2021-06-01-preview** (first GA 2021-07-01) — reused verbatim via `LegacyOperations`, with **no change to the response shape** (in the service the cross-tenant read is literally the same internal read with a source-vault context swap). So `datasourceAuthCredentials` → `SecretStoreResource.value` is neither introduced nor newly exposed here; it's the pre-existing shared model, and we intentionally did **not** fork a cross-tenant-specific model (that would diverge a response that is otherwise identical to the in-tenant one).

On behavior: DPP's secret-store auth is **Key-Vault-URI-based** — the service uses `SecretStoreResource.uri` (a reference) and never reads or populates the inline `value` (zero code references), so no secret value is materialized in the response, in-tenant or cross-tenant; cross-tenant the `uri` resolves inside the *source* tenant's Key Vault. This is a pre-existing shared DataProtection model, unchanged by this PR; how it's annotated is a DataProtection-wide, cross-version matter independent of these cross-tenant routes.

zubairabid:
Hi @sandipsh, appreciate the feedback. I've added detailed replies to both threads explaining why these are pre-existing patterns reused by the cross-tenant routes rather than new exposures. Happy to discuss further if anything's unclear.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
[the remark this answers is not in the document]
sandipsh: please provide where clauses rather than blanket suppression to whole file.

sandipsh: PostResponseCodes: suppression is new for package-2026-08-31-preview; it is not inherited from the previous package-2026-06-01 tag. The reason accurately describes the emitted behavior, but “matching the frozen BMS wire” does not by itself explain why the new DPP action must violate the ARM POST response-code contract.

Please document the concrete compatibility constraint that requires 200 with an empty body, add a where: clause scoped specifically to the DppCrossTenantVaultMappings_Remove response, and obtain Approved-Suppression. If no such constraint applies, change the operation to a compliant response shape and remove this suppression.

sandipsh:
**[NEW] 🔴 Blocking** **[[`StaleApiVersionPin`](https://github.com/Azure/azure-rest-api-specs/blob/main/eng/tools/typespec-validation/src/rules/stale-api-version-pin.ts#L16-L45)]** TypeSpec Validation is still failing because this PR adds `2026-08-31-preview`, while the C# management emitter remains pinned to `2025-09-01` in `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/tspconfig.yaml` line 45.

**Classification reasoning:** The pin predates this PR, but this PR introduces the newer API version that makes the pin stale. I could not find an author explanation in the PR discussion for why the pin must remain or why the validator-prescribed fix is not feasible.

Please either remove the C# emitter's `api-version` setting and regenerate the affected artifacts, as the validator directs, or document the concrete SDK compatibility or release constraint that requires retaining the older pin and explain how the required TypeSpec Validation check will be resolved. The required check must be green before merge.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
zubairabid: @sandipsh Accepting change: returning 204 now, instead of 200 with an empty body. Suppression removed.

zubairabid:
@sandipsh the @azure-typespec/http-client-csharp-mgmt emitter has intentionally been pinned to stable 2025-09-01. Updating it is currently out of scope for the preview 2026-08-31-preview API version release.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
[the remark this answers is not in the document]
gary-x-li: Acknowledged that these suppressions are carried forward from older tags. That explains their origin, but it does not make a whole-file suppression safe for a new tag: it can also hide newly introduced violations elsewhere in the generated file.

I agree with the requested resolution already stated in this thread: add a precise `where:` clause for each affected JSON path and obtain suppression approval for the new tag. Carrying forward existing technical debt is acceptable when explicitly scoped and reviewed; inheritance alone is not sufficient.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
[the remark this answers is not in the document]
gary-x-li: I reviewed the Recovery Services precedent in PR 41694. That PR received ARM sign-off and suppression approval for using a POST remove action rather than standard DELETE, so I am comfortable with POST-as-remove here as an explicitly approved exception.

The remaining Warning is narrower: the precedent's remove operation accepts `RemoveCrossTenantVaultMappingRequest`, whose body contains `resourceGuardOperationRequests`. This operation uses `ArmResourceActionSync<CrossTenantVaultMapping, void, ...>`, and the emitted Swagger has no request body. Therefore the suppression text and the reply's claim that this operation carries Resource Guard/MUA data in its request body contradict the current contract.

Please either add the intended MUA request body, or revise the suppression reason to explain the actual bodyless POST design and obtain the corresponding suppression approval. The POST verb itself is not the objection.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
gary-x-li: Thanks for the clarification. I agree that a Key Vault URI is a resource reference, not itself a secret. The remaining concern is the separate `SecretStoreResource.value` field: the response schema contains both `uri` and an unconstrained `value`, so the public GET/LIST contract permits a secret value even if the current implementation only populates `uri`.

Reusing the old model explains how this happened, but the cross-tenant route is a new public exposure of that model. Please encode the runtime guarantee in the contract by excluding `value` from GET/LIST responses (for example, through response visibility or a response projection). Until the schema prevents `value` from being returned, I consider this a valid Blocking contract-modeling issue, not evidence that the Key Vault URI itself is secret.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
gary-x-li: Thanks for the context. It is valid for a server-generated `nextLink` to point to the same LIST URL and contain an opaque `$skipToken`; clients should follow that URL verbatim. The issue is declaring `$skipToken` as an independent input parameter on the LIST operation, which exposes the continuation mechanism in generated SDK signatures and invites callers to construct or modify it.

The older in-tenant route is a legacy contract, but the cross-tenant URL is new and is therefore expected to meet the current collection-GET rule. Removing the declared parameter from this new route does not prevent the service from accepting the token when it arrives through `nextLink`, and it does not change the existing route. I therefore consider the Blocking finding justified for the new operation.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/main.tsp
```diff
@@ -68,6 +69,11 @@ enum Versions {
    * The 2026-06-01 API version.
    */
   v2026_06_01: "2026-06-01",
+
+  /**
+   * The 2026-08-31-preview API version.
+   */
+  v2026_08_31_preview: "2026-08-31-preview",
```
gary-x-li, line 77: **[NEW] 🔴 Blocking** **[[Section 3 - Versioning](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-project.instructions.md#3-versioning)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/main.tsp` - line 76 - The new preview API version is missing the required `@Azure.Core.previewVersion` decorator.

**Classification reasoning:** Introduced in this PR. The `2026-08-31-preview` enum member is new at base SHA `befd168`, and preview members MUST be explicitly identified to TypeSpec and downstream tooling.

**Suggested fix:**

```tsp
@Azure.Core.previewVersion
v2026_08_31_preview: "2026-08-31-preview",
```

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/examples/2026-08-31-preview/CrossTenantVaultMappingOperations/RestoresFromDppCrossTenantVault_Trigger.json
```diff
@@ -0,0 +1,67 @@
...
+  "responses": {
+    "200": {
+      "body": {
+        "jobId": "c60cb49-63e8-4b21-b9bd-26277b3fdfae",
+        "objectType": "OperationJobExtendedInfo"
+      }
+    },
+    "202": {
+      "headers": {
+        "Azure-AsyncOperation": "https://management.azure.com/subscriptions/04cf684a-d41f-4550-9f70-7708a3a2283b/resourceGroups/000pikumar/providers/Microsoft.DataProtection/backupVaults/PratikPrivatePreviewVault1/operationStatus/ZTA4YjQ0ZGYtYmNkNS00YTk1LWFjZTMtOTc1MjNmZWIxYWZlO2Y1ODg1MzA3LWJkNjItNDQ2OC05ZjZlLTJkMGM2NjNiNmJmNg==?api-version=2026-08-31-preview",
```
gary-x-li: **[NEW] 🟠 Warning** **[[EX-LRO-HEADERS](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#225-lro-example-headers-ex-lro-headers)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/examples/2026-08-31-preview/CrossTenantVaultMappingOperations/RestoresFromDppCrossTenantVault_Trigger.json` - line 59 - The `Location` polling URL does not match the cross-tenant operation-result route defined by this PR.

**Classification reasoning:** Introduced in this PR. The same malformed route is present in `ValidateRestoreFromDppCrossTenantVault_Trigger.json`. Examples drive generated documentation and customer implementations, so polling URLs must identify the actual endpoint for the operation.

**Suggested fix:** Use the matching route in both examples:

```text
.../backupVaults/{vaultName}/crossTenantVaultMappings/{mappingName}/backupInstances/{backupInstanceName}/operationResults/{operationId}?api-version=2026-08-31-preview
```

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+   */
+  sourceVaultId: Azure.Core.armResourceIdentifier<[
+    {
+      type: "Microsoft.DataProtection/backupVaults";
+    }
+  ]>;
+
+  /**
+   * Timestamp when the cross-tenant vault mapping was created.
+   */
```
gary-x-li: **[NEW] 🔴 Blocking** **[[Section 2.5 - Top-Level ARM Properties](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#25-top-level-arm-properties-must-not-be-reused-inside-the-properties-bag)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp` - line 54 - `createdTime` duplicates the standard ARM envelope field `systemData.createdAt`.

**Classification reasoning:** Introduced in this PR with the new `CrossTenantVaultMapping` resource. The property is described as the mapping creation timestamp, which is the same event represented by `systemData.createdAt`. Publishing both creates redundant SDK surface and leaves their behavior undefined if the values diverge.

**Suggested fix:** Remove `createdTime` and return the creation timestamp through the inherited top-level `systemData.createdAt` field.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+model CrossTenantOperationResultAcceptedResponse {
+  /**
+   * The HTTP 202 status code indicating the cross-tenant operation is still in progress.
+   */
+  @statusCode statusCode: 202;
+
+  /**
+   * The URL to poll for the operation result while it is still in progress.
+   */
+  @header("Location") location?: string;
```
gary-x-li: **[NEW] 🔴 Blocking** **[[RPC-Async-V1-14](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#66-202-response-and-polling-headers-rpc-async-v1-07-rpc-async-v1-06-rpc-async-v1-14)]** `specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp` - line 488 - The `202 Accepted` polling response makes the required `Location` header optional.

**Classification reasoning:** Introduced in this PR with the new cross-tenant operation-result poller. ARM requires every `202` response to include `Location`; otherwise a client may receive an in-progress response without a contractually defined URL for its next poll. The corresponding example currently demonstrates that invalid missing-header state.

**Suggested fix:**

```tsp
@header("Location") location: string;
```

Also include the matching cross-tenant operation-result URL in every `202` example.

gary-x-li (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`fb969de`](https://github.com/Azure/azure-rest-api-specs/pull/45468/commits/fb969de364139674dbc310acbeb60456b22f4d83). See inline comments for findings 5-8.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
```diff
@@ -0,0 +1,560 @@
...
+  is Azure.ResourceManager.ProxyResource<CrossTenantVaultMappingProperties> {
+  ...ResourceNameParameter<
+    Resource = CrossTenantVaultMapping,
+    KeyName = "mappingName",
+    SegmentName = "crossTenantVaultMappings",
+    NamePattern = "^[A-Za-z][A-Za-z0-9]{1,99}$"
+  >;
+}
+
+#suppress "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation" "Cross-tenant vault mappings are removed via the POST 'remove' action (RBAC separation-of-duties + Resource Guard / MUA body semantics), not a DELETE verb."
```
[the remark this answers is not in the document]
gary-x-li: Duplicate as what I posted. Can be ignored.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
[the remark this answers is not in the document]
gary-x-li: Duplicate as what I posted. Can be ignored.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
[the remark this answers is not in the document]
gary-x-li: Duplicate as what I posted. Can be ignored.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
```diff
@@ -49,6 +49,22 @@ semantic-validator: true
...
+```yaml $(tag) == 'package-2026-08-31-preview'
+input-file:
+  - preview/2026-08-31-preview/dataprotection.json
+suppressions:
+  - code: GetCollectionOnlyHasValueAndNextLink
+    from: dataprotection.json
+    reason: DPP list models compose value + DppResourceList.nextLink via allOf and satisfy the value+nextLink condition; the LintDiff rule does not follow the allOf composition (tool false-positive). Consistent with the existing suppression on prior DPP tags (package-preview-2024-02).
+  - code: AvoidAdditionalProperties
+    from: dataprotection.json
+    reason: There are objects that need a generic key-value pair in contract.
```
gary-x-li: Please add `where` to each suppression.

Also, the reason here is too generic. You need to justify each unique scenario.

gary-x-li:
Please also fix the SpellCheck error.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/readme.md
zubairabid: I've replaced the generic reasoning and broad suppression with `where`s and more specific reasons.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
zubairabid: Change accepted.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/CrossTenantVaultMappingResource.tsp
zubairabid: Change accepted.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/examples/2026-08-31-preview/CrossTenantVaultMappingOperations/RestoresFromDppCrossTenantVault_Trigger.json
zubairabid: Fixed.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/main.tsp
zubairabid: Done.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
zubairabid: Noted, made the change.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
zubairabid: @gary-x-li That is fair, I have some reservations around the cost of encoding the change as a projection given the somewhat minimal gains.

In order to not exclude `value` from GET/LIST responses, we'll have to fork every model on the shared chain of ` SecretStoreResource → SecretStoreBasedAuthCredentials → AuthCredentials  (a discriminated base)  → BackupInstance.datasourceAuthCredentials → BackupInstanceResource`. into parallel CrossTenant* copies, duplicating 105-odd lines of TypeSpec, ~70 of which are for `BackupInstance`.

`BackupInstance` is one of core models for our service and among the most frequently updated and extended. The fork will have to be hand-updated for every future change, and a missed sync causing the in-tenant and cross-tenant `BackupInstance` models to diverge would be an easy bug to introduce while being very hard to catch. 

And as mentioned earlier, it reduces one new public exposure of this model, but the model already exists for every in-tenant `BackupInstance` GET/LIST across all released versions since 2021.  We'd be adding a substantial level of complexibility and maintenance trouble for not much of a gain.

zubairabid:
All feedback addressed, awaiting further feedback.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
gary-x-li: Thanks, the maintenance concern is valid. I agree that copying the entire model chain would create substantial duplication and synchronization risk.

However, a cross-tenant model fork may not be necessary. Since you indicated that the service never reads or populates `SecretStoreResource.value`, the cleaner fix may be to remove that property beginning with `2026-08-31-preview`, using TypeSpec versioning so previously published API versions remain unchanged.

This would be a breaking change relative to the preceding version and may require approval, but it would make the new contract accurately reflect service behavior without duplicating the model hierarchy.

Before choosing that approach, please confirm that clients cannot legitimately send `value` either. If it is accepted on writes but never returned, it should instead have create/update-only visibility. Adding only `@secret` would not prevent it from appearing in GET/LIST responses.

Until the published response schema excludes `value`, the new cross-tenant GET/LIST routes still permit returning an inline secret. Although the shared model predates this PR, exposing it through the new cross-tenant routes is new and crosses an additional authorization boundary. I therefore believe this remains Blocking.

gary-x-li:
Verified all other issues are fixed. I responded to the only remaining issue: https://github.com/Azure/azure-rest-api-specs/pull/45468#discussion_r3909175280

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
zubairabid: Noted. Implementing this approach.

zubairabid:
Adopted suggestion as described in https://github.com/Azure/azure-rest-api-specs/pull/45468#discussion_r3909175280, leading to breaking change review. Please check the conversation there and approve as appropriate.

specification/dataprotection/resource-manager/Microsoft.DataProtection/DataProtection/preview/2026-08-31-preview/dataprotection.json
zubairabid: Update: after reviewing with the team, I found an alternate way to exclude `value` from only the new Cross-tenant API surface, without causing a breaking change in the existing APIs as well. Proceeding with this approach, please confirm that the PR can be unblocked.

PR #45492, opened 2026-08-13
Fix GitHub test alignment after TypeSpec package bump

xirzec:
The GitHub Actions test workflow began failing because the root TypeSpec Azure packages were bumped to 0.71.0 without running the `.github` alignment checks. The next `.github` change exposed stale data-plane linter coverage metadata.

This change:

- re-verifies the coverage map for TypeSpec Azure 0.71.0, replacing `no-unnamed-union` with `no-unnamed-types` and adding `no-openapi-client-extensions`
- removes the abandoned `no-dollar-prefixed-query-params` proposal and its upstream PR references
- triggers GitHub Actions tests for root `package.json` changes so future dependency bumps cannot bypass the alignment invariant

## Validation

- `npm run check` from `.github`

### DIFF
--- a/.github/workflows/github-test.yaml
+++ b/.github/workflows/github-test.yaml
@@ -6,9 +6,11 @@ on:
       - main
     paths:
       - .github/**
+      - package.json
   pull_request:
     paths:
       - .github/**
+      - package.json
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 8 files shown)


mikeharder (approved):

.github/workflows/github-test.yaml
```diff
@@ -6,9 +6,11 @@ on:
       - main
     paths:
       - .github/**
+      - package.json
```
mikeharder, line 10: @copilot: Add a comment explaining why this workflow needs to trigger on root package.json now.

PR #45500, opened 2026-08-14
Simplify the data-plane API reviewer

xirzec:
## Why

The data-plane reviewer had accumulated more than 8,000 lines of critic, eval, alignment, and synchronization machinery around a small manually invoked workflow. That machinery introduced repeated integration failures and made TypeSpec package updates brittle, while production usage showed that the highest-value findings came from a much smaller semantic scope.

## What changed

- Narrow the unattended reviewer to secrets and visibility, stable breaking changes, resource/action modeling, error semantics, and LRO/paging correctness.
- Replace the separate critic and version-pinned linter interlock with one fail-closed pre-emit verification gate that re-fetches every cited line at the pinned head or base SHA.
- Remove the data-plane vally suite, synthetic fixtures, alignment workflow/script/tests, report-format contract, and low-signal naming/docs and design-question guidance.
- Keep the label-first gh-aw workflow, `workflow_dispatch` testing path, trusted checkout, read-only tools, threat detection, and safe outputs, while removing cross-run cache memory.
- Cap output at five findings and add a generic check requiring changed gh-aw Markdown sources to include their regenerated lock files.

This deliberately trades some review coverage for a smaller, more reliable advisory reviewer. It removes 46 files and about 9,400 net lines; naming/documentation findings and speculative design questions are no longer part of unattended review.

## Validation

- `npm run check` from `.github`
- `gh aw compile data-plane-api-review --no-check-update`
- `git diff --check`


.github/workflows/github-test.yaml
```diff
@@ -38,6 +35,32 @@ jobs:
...
+              pull_number: context.payload.pull_request.number,
+              per_page: 100,
+            });
+            const changed = new Set(files.map(({ filename }) => filename));
+            const sources = [...changed].filter((file) =>
+              /^\.github\/workflows\/[^/]+\.md$/.test(file),
+            );
+            const missing = sources.filter(
+              (source) => !changed.has(source.replace(/\.md$/, ".lock.yml")),
+            );
```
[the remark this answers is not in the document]
xirzec: Fixed in 020b6dcd9b2. The guard now tracks each changed file's status: a non-removed workflow source requires a non-removed lock, while removing a workflow source requires removing its lock. I also exercised the modified/removed source and lock combinations.

.github/agents/README.md
```diff
@@ -5,13 +5,12 @@ level is auto-registered by Visual Studio Code.
...
+| ------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------- | --------------- |
+| [`arm-api-reviewer.agent.md`](./arm-api-reviewer.agent.md)               | Reviews Azure REST API spec PRs; findings verified by the Critic before posting.                              | Yes             |
+| [`arm-api-review-critic.agent.md`](./arm-api-review-critic.agent.md)     | Internal subagent invoked by the Reviewer at Step 7 to re-verify findings.                                    | No              |
+| [`data-plane-api-reviewer.agent.md`](./data-plane-api-reviewer.agent.md) | Reviews data-plane TypeSpec for high-value semantic defects outside deterministic compiler and linter checks. | Yes             |
+| [`hero-scenarios.agent.md`](./hero-scenarios.agent.md)                   | Suggests hero scenarios from TypeSpec to seed SDK samples, documentation, and tests.                          | Yes             |
```
heaths, line 13: If reducing tokens was a goal, this could be much smaller. Don't pad columns, and only minimal separator which, I believe, is `---` e.g.,

```markdown
File | Purpose | User-invocable
--- | --- | ---
1 | 2 | 3
a | b | c
```

Notice how I also left off outer lines. Also valid markdown. Small, but things like this add up.

.github/agents/README.md
```diff
@@ -35,6 +30,6 @@ subdirectory keeps reference files out of Visual Studio Code's agent scan path.
 - Agents that run unattended (from a `.github/workflows/*.md` gh-aw workflow)
   get **no** mutating GitHub tools at all; `safe-outputs` is their only write
   channel. Behavioral gating in prose is not a substitute for an allowlist.
-- No emoji or non-ASCII chars except where required by agent output.
-  Severity glyphs (🔴 / 🟡 / 💡) are part of agent output and are expected in
-  report templates and severity tables.
+- No emoji or non-ASCII chars except where required by an agent's output
```
heaths: I think for at-a-glance review, I'd leave the emoji. It's much more obvious.

.github/skills/azure-api-review/references/data-plane-lro-and-paging.md
```diff
@@ -1,243 +1,84 @@
...
+## DP-PAGE-02: Sibling list operations need one paging shape
 
-What the linter does **not** check, and you therefore own:
+- **Severity:** Warning.
 
-| Decorator         | Applies to                     | Marks                           |
-| ----------------- | ------------------------------ | ------------------------------- |
-| `@lroCanceled`    | `EnumMember` \| `UnionVariant` | the terminal cancellation state |
-| `@lroResult`      | `ModelProperty`                | the result on success           |
-| `@lroErrorResult` | `ModelProperty`                | the error on failure            |
```
heaths, line 128: I think we do need to encode TypeSpec info into these instructions. Don't make the agent fetch things. Not only does that taker longer, but it'll have to load all those tokens from docs written for humans.

.github/skills/azure-api-review/references/data-plane-lro-and-paging.md
```diff
@@ -1,243 +1,84 @@
...
 > [status monitor structure](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-status-monitor-structure),
 > [polling behavior](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#lro-poll),
 > [server-driven paging](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections-support-server-driven-paging),
-> [consistent paging options](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections-consistent-options-with-pagination),
 > and
-> [`maxpagesize`](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections-maxpagesize-param)
-> in the Azure REST API Guidelines. TypeSpec-specific status-monitor metadata is
-> defined by the
-> [Azure.Core LRO decorators](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/decorators/#Azure.Core.lroErrorResult).
-> The upstream wire contract and TypeSpec decorator reference take precedence.
```
heaths, line 31: These need to be encoded here. Making the agent fetch guidelines every time not only wastes time, but is more tokens. Those guidelines are written and maintained for humans. This is why I had the Azure SDK custom agent turn the guidelines into LLM-optimized guidelines. Did I go too far? Maybe. We need to measure performance against a test corpus. But we certainly don't need all the explanation of text e.g., the "why"s. We can and should link to guidelines for people to learn why.

.github/workflows/data-plane-api-review.md
```diff
@@ -91,17 +64,11 @@ safe-outputs:
...
     target: "${{ github.event.pull_request.number || inputs.item_number }}"
   messages:
-    footer: "> 🔍 *Automated data-plane API review by [{workflow_name}]({run_url}). Advisory only.*"
-    run-started: "🔍 [{workflow_name}]({run_url}) is reviewing the data-plane TypeSpec changes in this PR…"
-    run-success: "🔍 [{workflow_name}]({run_url}) finished. ✅"
-    run-failure: "🔍 [{workflow_name}]({run_url}) {status}. ❌"
+    footer: "> Automated data-plane API review by [{workflow_name}]({run_url}). Advisory only."
+    run-started: "[{workflow_name}]({run_url}) is reviewing the data-plane TypeSpec changes in this PR."
+    run-success: "[{workflow_name}]({run_url}) finished."
+    run-failure: "[{workflow_name}]({run_url}) {status}."
```
heaths: Again, at a glance the emoji are actually useful indicators. Both here and above, they are for us during review of comments/status.

.github/agents/README.md
xirzec: Unfortunately, this one gets enforced by Prettier, so we'd have to reconfigure it (at least for the .github folder.) Can tackle that for all files as a separate item.

.github/agents/README.md
xirzec: Added these back

.github/skills/azure-api-review/references/data-plane-lro-and-paging.md
xirzec: Brought this back

heaths (approved):
All these rules more or less copied from the guidelines, though: how are those updated? Might be good to have a skill checked in[^1] that can redo those.

[^1]: That said, let's talk offline. I'm working on that for cross-language SDK repos - a more generic version of what I did for the Rust SDK.

PR #45507, opened 2026-08-14
Add TypeSpec branch sync runbook

iscai-msft:
## Summary
- Add a prompt-based runbook for the approval-gated `main` <-> `typespec-next` sync process.
- Keep `main` on stable TypeSpec packages and keep `typespec-next` on `"next"` package pins while preserving merge ancestry.
- Require the existing **TypeSpec Validation - All** matrix to pass before landing either phase.
- Clarify that the routine sync is `main` into `typespec-next`, and that the reverse direction is only for intentional promotion of `typespec-next` changes into `main`.
- Use draft PRs for the sync steps so reviewers can iterate without notifying the full repo audience.

## Validation gate
The runbook requires every **TypeSpec Validation - All** matrix job to pass for each sync PR: all three specification shards on both Ubuntu and Windows. It explicitly requires investigating every failed job, not just the first failure.

### DIFF
--- a/.github/prompts/bidirectional-spec-upgrade.prompt.md
+++ b/.github/prompts/bidirectional-spec-upgrade.prompt.md
@@ -1,0 +1,159 @@
+---
+mode: agent
+---
+
+Perform the approval-gated TypeSpec branch sync between `main` and `typespec-next`.
+
+The routine sync is `main` into `typespec-next`: merge `main`, fix TypeSpec Validation - All
+(TSV) failures, and keep `typespec-next` on its `"next"` package pins while preserving commit
+ancestry. Only use the reverse direction when a specific `typespec-next` change is intentionally
+being promoted into `main`.
+
+`main` must use the latest compatible stable TypeSpec packages. `typespec-next` intentionally uses
+the `"next"` TypeSpec packages to find upcoming compatibility issues. Preserve that difference while
+keeping the branches' commit histories connected.
+
+## Non-negotiable rules
+
+- Use real merge commits. Do not use squash merge or rebase merge.
+- Preserve ancestry. After each phase, `git merge-base --is-ancestor <merged-source> <candidate-tip>`
+  must succeed.
+- Do not copy files with `git checkout <ref> -- <path>`; that copies content without preserving
+  source-branch ancestry.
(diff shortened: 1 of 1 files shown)


.github/prompts/bidirectional-spec-upgrade.prompt.md
```diff
@@ -0,0 +1,137 @@
...
+4. If any job fails, inspect logs for every failed shard and operating system:
+
+   ```bash
+   gh run view <run-id> --log-failed
+   ```
+
+   Fix every reported specification failure, rerun the complete matrix, and repeat until all jobs
+   pass. A failure in one shard does not prove the other shards are clean.
+
+## Phase 1: merge `typespec-next` into `main`
```
catalinaperalta: I see one of the last steps mentions opening a pr, should we have a 1 sentence summary under the phase 1 title that states the the goal of this phase is to create a pr that merged typespec-next into main? also IMO those PRs should be opened as draft since they tag all repo authors. So that whoever is attempting to keep the sync can iterate without pinging all affected authos

.github/prompts/bidirectional-spec-upgrade.prompt.md
```diff
@@ -0,0 +1,137 @@
...
+
+   The first command must not show TypeSpec package dependencies. The second command must succeed.
+
+4. Run `npm install`, commit the resulting lockfile, complete the required validation gate, and
+   create a PR targeting `main`.
+5. Wait for approval and landing. Record the landed `main` SHA before continuing.
+
+## Phase 2: merge approved `main` into `typespec-next`
+
+1. Fetch the `main` commit containing the approved Phase 1 PR, then create a working branch from
```
catalinaperalta: This doesnt seem completely true if you want this prompt to also be used for between release sync prs. It should just be simplified to merge main into typespec-next and fix all tsv errors

catalinaperalta (approved):
lgtm. Just one nit if you want to address now, I would make the merge main into typespec-next as phase 1 and typespec next into main as phase 2 since that is the less frequent path. I just wonder how an agent would take it that we're typically skipping phase 1

catalinaperalta (approved):

PR #45514, opened 2026-08-14
Add shell tool in toolbox type spec

ylxlpl:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -144,6 +146,130 @@ model WebSearchToolboxTool extends ToolboxTool {
...
+
+  ...OmitProperties<
+    OpenAI.ContainerAutoParam,
+    "type" | "network_policy"
+  >;
+
+  /**
+   * The network access policy for the container.
+   */
+  network_policy?: ToolboxShellNetworkPolicy;
```
dargilco, line 195: Do we need to update doc string to discuss what happens when network_policy is not provided? "When omitted, the service defaults to ..."

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -144,6 +146,130 @@ model WebSearchToolboxTool extends ToolboxTool {
...
+   * The type of the tool. Always `shell`.
+   */
+  type: ToolboxShellToolType;
+
+  ...OmitProperties<ToolFields<OpenAI.FunctionShellToolParam>, "environment">;
+
+  /**
+   * The environment in which shell commands are executed.
+   */
+  environment?: ToolboxShellEnvironment;
```
dargilco: Do we need to update doc string to discuss what happens when environment is not provided? "When omitted, the service defaults to ..."

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
```diff
@@ -72,6 +72,7 @@ alias ToolFields<Source> = {
 alias ToolboxCodeInterpreterToolType = "code_interpreter";
 alias ToolboxFileSearchToolType = "file_search";
 alias ToolboxWebSearchToolType = "web_search";
+alias ToolboxShellToolType = "shell";
```
dargilco, line 75: Why don't we have a preview version first?

specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp
dargilco: Per our chat, since at the moment it's not optional on the service, please remove the "?". In the future we can change this to be optional

ylxlpl: Customer can omit this. If they omit this, the default behavior will be disabled.

dargilco: That's for the correction. I understand that this property needs to remain optional. But Please update the doc string with a sentence to describe service-side default, such as "When omitted the service defaults to ..."

glecaros (approved):

PR #45545, opened 2026-08-17
Make resources fields read-only and add priority to CommandJob

swarupecenits:
## Summary

Aligns the Foundry jobs data-plane contract with the agreed service behavior for compute selection, per work item 5475147.

Two changes:

1. **`resources` fields are now read-only.** `instanceType`, `properties`, `shmSize` and `dockerArgs` on `JobResourceConfiguration` are marked `@visibility(Lifecycle.Read)`. The service infers these from the target cluster and ignores them on create, so accepting them in the request contract was misleading. They are still returned on GET and List. `instanceCount` stays writable - CPU jobs use it to request whole nodes.

2. **`priority` added to `CommandJob`.** The field already exists in the live service but was never modeled. Typed as the extensible `JobPriority` union with values `Low`, `Mid`, `High`.

### Read-only resources fields

| Field | Before | After |
|---|---|---|
| `instanceCount` | writable | **writable** (unchanged) |
| `instanceType` | writable | read-only |
| `properties` | writable | read-only |
| `shmSize` | writable | read-only |
| `dockerArgs` | writable | read-only |

Confirmed with the service team: these are removed from the job creation flow only, and deliberately kept for read operations because the service infers and populates them.

### `priority`

Wire contract confirmed independently by two service owners: field name `priority`, string type, values `Low` / `Mid` / `High`. The service-side C# enum uses a string converter, so the ordinals never reach the wire.

Modeled as an extensible union to match the existing convention in this file (`JobType`, `AssetTypes`, `InputOutputModes`, `ListViewType`) and so that future values are not a breaking change.

The service applies `Low` when the field is omitted. This is documented in the property description rather than emitted as an OpenAPI `default`, so generated clients do not materialize and send `"Low"` on every request. That preserves the distinction between "caller omitted" and "caller explicitly chose Low", and keeps the service authoritative if the default ever changes.

## Files changed

- `src/jobs/models.tsp` - visibility decorators, `JobPriority` union, `priority` property
- `examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json` - read-only fields removed from the request body, `priority` added to request and both responses
- `examples/v1/Jobs_Get_MaximumSet_Gen.json` - `priority` added to the response
- `examples/v1/Jobs_List_MaximumSet_Gen.json` - `priority` added to the response
- `openapi3/**` - regenerated

No versioning decorators are needed: the jobs surface is gated behind the required `Foundry-Features: Jobs=V1Preview` header and this contract is not yet published.

## Validation

- `npx tsp compile . --emit "@typespec/openapi3"` - clean
- `npx tsp compile . --warn-as-error` - clean, zero linter warnings
- All modified example files re-parse as valid JSON
- Verified in the emitted schema: exactly 4 `readOnly: true` under `JobResourceConfiguration`, `instanceCount` unaffected, `priority` present on `CommandJob` and absent from `required[]`, no `default` emitted for `priority`

Related: work item [5475147](https://msdata.visualstudio.com/Vienna/_workitems/edit/5475147)

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json
+++ b/specification/ai-foundry/data-plane/Foundry/examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json
@@ -11,8 +11,12 @@
         "environmentImageReference": "mcr.microsoft.com/azureml/openmpi4.1.0-cuda11.8-cudnn8-ubuntu22.04:latest",
         "displayName": "My Training Job",
         "description": "A sample command job for training",
-        "tags": { "framework": "pytorch" },
-        "properties": { "experimentName": "my-experiment" },
+        "tags": {
+          "framework": "pytorch"
+        },
+        "properties": {
+          "experimentName": "my-experiment"
+        },
         "codeId": "azureai:my-training-code:1",
         "computeId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.MachineLearningServices/workspaces/my-ws/computes/my-cluster",
         "inputs": {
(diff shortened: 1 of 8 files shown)


saanikaguptamicrosoft:
There are some failing checks. It should not be a merge blocker as this is not main branch, but in future when we try to merge this branch into main, we might run into these failures then. So if it's feasible, pls resolve checks which make sense. For checks failing due to service team API architecture, let it be..

saanikaguptamicrosoft (approved):

Samiulla (approved):

sharvin2187 (approved):

sharvin2187 (approved):

PR #45557, opened 2026-08-17
Suppress swagger checks

catalinaperalta:


catalinaperalta:
this was already implemented in the spec

PR #45455, opened 2026-08-13
Add Chaos 2026-11-01 stable API version

juliagao3:
## Summary

- add the `2026-11-01` stable API version for Microsoft.Chaos
- carry over the existing `2026-08-01-preview` API surface unchanged
- add versioned source examples and generated OpenAPI output
- add the `package-2026-11` AutoRest tag
- carry the already-approved LintDiff suppressions forward into the `package-2026-11` tag

## Notes

This establishes the GA baseline while `2026-08-01-preview` finishes publishing. It should be updated from `main` after that preview lands so the GA surface and examples reflect the final published preview.

PR #45591, opened 2026-08-18
fix(chaos): allow template-based scenario creation and author-managed version (AB#39305402)

franklixuefei:
Fixes the two `ScenarioProperties` fields named in **AB#39305402**, both of which contradict what the Backend already implements.

## What changed

**1. `parameters` and `actions` are optional from 2026-11-01**

They were required with `@minItems(1)`, so a template-based create — supply `createdFrom`, let the template define the structure — could not be expressed at all. This is not a hypothetical: BE has implemented template-only creation *deliberately* since `ScenariosController.cs:85-92`, whose own comment reads *"the user's request body actions are intentionally not used."* The contract was the only thing blocking it.

**2. `version` is no longer `readOnly`**

BE accepts a caller-supplied version on **both** paths — `ScenariosController.cs:113` (`Version: requestBody.Version ?? "1.0"`) on create, and `:141` → `Scenario.UpdateProperties(..., string version, ...)` on update. Scenarios are customer-owned instances whatever they were stamped from, so their version is author-managed. Note this is the *opposite* direction from `createdFrom`: one was read-only and needed create, this one was read-only and needed full write.

**3. The template example no longer sends fields the service discards**

`Scenarios_CreateOrUpdate_From_Template.json` sent `createdFrom` **and** `parameters` **and** `actions`. BE ignores the latter two. The request is reduced to what is actually honoured; the responses still show the fully resolved scenario, which is the point of template-based creation.

## Why the obvious fix would have been wrong

The obvious change is to drop `required` and edit the docs. Both parts have a trap.

- **`required` must be version-gated.** An ungated edit silently removed `required: [parameters, actions]` from `2026-05-01-preview` and `2026-08-01-preview` — the first of which is the pinned Portal contract. `@madeOptional(...v2026_11_01)` confines it to the GA version. Verified: `required` is gone from 2026-11-01 and still present in both previews.
- **Doc text cannot be version-gated at all** — TypeSpec's versioning library exposes only `@added`, `@removed`, `@renamedFrom`, `@madeOptional`, `@madeRequired`, `@typeChangedFrom`, `@returnTypeChangedFrom`. There is no visibility or doc equivalent. So the docs here describe **behaviour**, which is version-invariant ("when `createdFrom` is supplied the template is authoritative and inline structure is not used"), rather than **schema** ("omit `parameters` and `actions`"), which would be false for the previews where they remain required.

## Blast radius

The two preview files change by 4 insertions / 5 deletions each: the `readOnly` removal on `version`, plus description text.

The `version` change cannot be confined to 2026-11-01 for the language reason above. There is direct precedent on this branch — #45311 flipped `createdFrom` to `read, create` ungated, and both previews carry `x-ms-mutability: [read, create]` today. It is also a relaxation rather than a restriction, and it makes the previews *more* accurate: BE has always accepted a caller-supplied version, so `readOnly: true` was already false for those versions.

## What this deliberately does not do

- **No format or identity constraint on `createdFrom`.** Which wire form is canonical, and where resolution happens, is open — see **AB#39312160**. Constraining the shape here would prejudge it.
- **No BE change.** None is required; `@minItems(1)` was the entire blocker.
- **`@minItems(1)` is retained**, so an inline create still cannot send an empty `actions` array.

## Reading order

`scenario.models.tsp` first — it is the only hand-edited file. The other five are emitter output.

## Verification

`tsp compile` clean (compiler v1.14.0). Emitted swagger inspected directly rather than assumed: `required: [parameters, actions]` removed from 2026-11-01 only; `readOnly` removed from `version`. `service.yaml` is deliberately **not** included — a bare `tsp compile` regenerates it and drops every swagger-sourced version from 2021-09-15-preview through 2024-03-22-preview, which is unrelated collateral.

## Rigor gate

Reviewed before implementation by `claude-opus-5` and `gpt-5.6-sol` independently, read-only, against pinned baselines — Chaos `ee187c044312e754ef771ee7fbb1a6c6e30aa2cd`, specs `c764f40c93fdbf36156134c6b0a019f0f649c02d`. Neither returned clean and each found blocking issues the other missed; the original design was discarded and the direction kept. The surviving warning both raised independently: **do not make `ScenarioValidationService.cs:71-81` conditional** — that guard is what turns a structureless scenario into a 400 rather than a silent empty resource.

**Could not verify:** that `azure-rest-api-specs-pr` agrees with the public repo (not cloned locally); that Julia's `!16703100` behaves as its description claims (read the description, not the diff).

**One correction to the record this PR rests on.** The bug originally claimed that stripping `createdFrom` produces an empty scenario with HTTP 200. That is false, and it was repeated by four sessions before being checked. It fails 400 at three independent layers — contract `@minItems(1)`, `ScenarioValidationService.cs:71-81`, and `ScenariosController.cs:150-154`. The genuine defect nearby is a **500**, not a silent success, and it is tracked separately as AB#39312160.

### DIFF
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/examples/2026-11-01/Scenarios_CreateOrUpdate_From_Template.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/examples/2026-11-01/Scenarios_CreateOrUpdate_From_Template.json
@@ -8,39 +8,7 @@
     "resource": {
       "properties": {
         "createdFrom": "/providers/Microsoft.Chaos/scenarioTemplates/ComputeZoneDown/versions/1.2",
-        "description": "Simulates a datacenter/availability zone failure by shutting down VMSS instances and standalone VMs in the target zone. Zone failures account for ~25% of all cloud incidents — this scenario validates your application's cross-zone resilience.",
-        "parameters": [
-          {
-            "name": "duration",
-            "type": "string",
-            "default": "PT15M",
-            "required": false,
-            "description": "The duration of the zone down scenario.",
-            "format": {
-              "formatType": "Duration"
-            }
-          }
-        ],
-        "actions": [
-          {
-            "name": "vmssZoneShutdown",
-            "actionId": "urn:csci:microsoft:virtualMachineScaleSet:shutdown/1.0.0",
-            "description": "Shut down VMSS instances in target zone",
-            "duration": "%%{parameters.duration}%%",
-            "parameters": [
-              {
-                "key": "Zones",
-                "value": "%%{resourceTargeting.include.zones}%%"
-              }
-            ]
-          },
-          {
-            "name": "vmZoneShutdown",
-            "actionId": "urn:csci:microsoft:compute:shutdown/1.0.0",
-            "description": "Shut down VMs in target zone",
-            "duration": "%%{parameters.duration}%%"
-          }
-        ]
+        "description": "Simulates a datacenter/availability zone failure by shutting down VMSS instances and standalone VMs in the target zone. Zone failures account for ~25% of all cloud incidents — this scenario validates your application's cross-zone resilience."
       }
     }
   },
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-05-01-preview/openapi.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-05-01-preview/openapi.json
@@ -6540,7 +6540,7 @@
     },
     "ScenarioProperties": {
       "type": "object",
-      "description": "Model that represents the properties of the scenario.",
+      "description": "Model that represents the properties of the scenario.\n\nA scenario's structure comes from exactly one of two sources, and they are\nmutually exclusive:\n\n- **Template-based**: `createdFrom` references a scenario template version,\nand that template is the authoritative source for the scenario's structure.\nInline `parameters` and `actions` are not used.\n- **Inline**: `createdFrom` is absent, and the scenario's structure is taken\nfrom the `parameters` and `actions` supplied in the request.\n\nSupplying neither source is rejected. This constraint cannot be expressed in\nthe schema and is enforced at request validation.",
       "properties": {
         "provisioningState": {
           "$ref": "#/definitions/ProvisioningState",
@@ -6557,8 +6557,7 @@
         },
         "version": {
           "type": "string",
-          "description": "Version of the scenario.",
-          "readOnly": true
+          "description": "Version of the scenario, supplied by the author. Defaults to `1.0` when not\nsupplied on create. This is the scenario's own version and is independent of\nthe version of any template it was created from."
         },
         "description": {
           "type": "string",
@@ -6566,7 +6565,7 @@
         },
         "parameters": {
           "type": "array",
-          "description": "Parameter definitions for the scenario.",
+          "description": "Parameter definitions for the scenario. Used as the scenario's parameter\ndefinitions when `createdFrom` is not supplied; not used when it is,\nbecause the referenced template is then the authoritative source for the\nscenario's structure.",
           "items": {
             "$ref": "#/definitions/ScenarioParameter"
           },
@@ -6576,7 +6575,7 @@
         },
         "actions": {
           "type": "array",
-          "description": "Array of actions that define the scenario's orchestration.",
+          "description": "Array of actions that define the scenario's orchestration. Used as the\nscenario's actions when `createdFrom` is not supplied; not used when it is,\nbecause the referenced template is then the authoritative source for the\nscenario's structure. When supplied, at least one action must be present.",
           "minItems": 1,
           "items": {
             "$ref": "#/definitions/ScenarioAction"
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-08-01-preview/openapi.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/preview/2026-08-01-preview/openapi.json
@@ -6984,7 +6984,7 @@
     },
     "ScenarioProperties": {
       "type": "object",
-      "description": "Model that represents the properties of the scenario.",
+      "description": "Model that represents the properties of the scenario.\n\nA scenario's structure comes from exactly one of two sources, and they are\nmutually exclusive:\n\n- **Template-based**: `createdFrom` references a scenario template version,\nand that template is the authoritative source for the scenario's structure.\nInline `parameters` and `actions` are not used.\n- **Inline**: `createdFrom` is absent, and the scenario's structure is taken\nfrom the `parameters` and `actions` supplied in the request.\n\nSupplying neither source is rejected. This constraint cannot be expressed in\nthe schema and is enforced at request validation.",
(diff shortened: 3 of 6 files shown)


franklixuefei:
> **Updated after merge.** The "asking for" at the end of this comment was wrong when first posted — breaking-change sign-off is **deferred to the promotion branch's own merge to `main`**, not requested here. The technical analysis below stands and is intended for that review. See the merge-rationale comment at the bottom of this thread for the reasoning.

## `Swagger BreakingChange` — 4 × rule 1029, and where they actually get resolved

### Half of these are not from this PR

The tool diffs against **`main`**, not against this PR's base. The promotion branch already carries two of them.

| | 1029 errors | which |
|---|---|---|
| **#45455** @ `c764f40c93` (this PR's base) | **2** | `createdFrom` × both previews |
| **#45591** (this PR) | **4** | those same 2, **plus `version` × both previews** |

`c764f40c93` is exactly the commit this branch was cut from, and #45455's `Swagger BreakingChange` was already failing there. The `createdFrom` pair comes from #45311 and would need Julia's change reverted to clear. **This PR adds two, both on `version`.**

### The two that are mine cannot be version-gated

`@visibility` is not version-aware, and this is a language limitation rather than a convention we could opt out of. `@typespec/versioning` exports exactly seven decorators:

`@added` · `@removed` · `@renamedFrom` · `@madeOptional` · `@madeRequired` · `@typeChangedFrom` · `@returnTypeChangedFrom`

There is no visibility equivalent, and none for doc text either. `@madeOptional` is why the `parameters`/`actions` half of this PR produces **zero** breaking errors — that change *is* confined to 2026-11-01. The `version` change has no such instrument available.

Declaring the property twice under `@removed`/`@added` is not a workaround: duplicate property names in one model are illegal regardless of versioning decorators.

### Suppression is not an available route either

No spec in this repository suppresses rule 1029, and `BreakingChange` does not appear as a suppressible tool in any `suppressions.yaml`. (Verified against the repo, with a control query confirming the search returns results for entries that do exist.) Chaos's own `suppressions.yaml` carries only `TypeSpecRequirement` entries.

### Why the change is still correct

Keeping `readOnly: true` on `version` would leave the published contract asserting something untrue in *every* version, including the two previews: BE has always accepted a caller-supplied `version` — `ScenariosController.cs:113` on create, `:141` → `UpdateProperties` on update. The previews are not being made wrong by this change; they are being made to match behaviour that already shipped.

Both are also relaxations rather than restrictions — a field clients could not send becomes sendable. No request that is valid today becomes invalid.

### Where this gets decided

At the **ARM review when the promotion branch merges to `main`**, covering `createdFrom` and `version` together — same rule, same two fields, same two files, same underlying cause. These findings are recorded here so that review has them, not because they need resolving on this PR.

RenzoPrettoMS (commented):
Whats the context of making ScenarioParams and ScenarioActions optional? Its fine by me, but want to know :)

franklixuefei:
@RenzoPrettoMS Sure! Short version: today you can't actually express a template-based create. `actions` is required with `@minItems(1)`, so to say "give me ZoneDown 1.0 and nothing else" you have to invent a dummy action — which the Backend then throws away, because it already treats the template as authoritative and ignores inline `actions` when `createdFrom` is set (`ScenariosController.cs:85-92`, comment there literally reads "the user's request body actions are intentionally not used"). So the contract was forcing callers to send something meaningless. This just lets them omit it.

Two guardrails: `@madeOptional` confines the change to `2026-11-01` (the previews keep `required`, so the pinned Portal contract is untouched), and `@minItems(1)` stays, so an inline create still can't send an empty `actions` array. The `version` change in here is a separate but similar mismatch — it's marked read-only while the Backend accepts a caller-supplied value on both create and update.

franklixuefei:
> **Updated after merge.** This comment originally framed `BreakingChange-Python-Sdk` as the gate on *this* PR. That was wrong: both gates are computed against `main`, which this PR does not touch, so both are deferred to the promotion branch's merge to `main`. The findings below stand and are recorded for that review.

## There are **two** breaking-change gates, and they resolve in the same place — later

My earlier comment in this thread analysed `Swagger BreakingChange` (rule 1029). There is a second, separate gate, and the bot distinguishes them:

| | gate | bot's wording |
|---|---|---|
| ❌ | `BreakingChange-Python-Sdk` | *"next steps that **must** be taken to merge"* → step 3 of the [PR workflow diagram](https://aka.ms/azsdk/pr-diagram) |
| ⚠️ | `Swagger BreakingChange` | *"as of today they are **not blocking** this PR"* → [aka.ms/brch](https://aka.ms/brch) |

Both are computed **against `main`**. This PR targets the 2026-11-01 GA accumulation branch, so neither is evaluated against what this PR actually changes — and both will be **recomputed on the promotion PR** when that merges. That is where they get decided, once, with ARM.

### What is verified

- **`SDK Validation - Python` passes — 0 errors, 0 warnings.** So does every other language: .NET, JS, Go, Java, Rust all succeeded. This is not a build or generation failure.
- The label therefore comes from the **generated-API-surface comparison** (the APIView revision linked earlier in this thread), not from SDK validation.
- **Python is the only language flagged.** No `BreakingChange-*-Sdk` label was applied for Go, JS, .NET, Java or Rust.

### What is *not* verified

I have **not** read the specific Python API-surface delta — the APIView revision needs interactive access I don't have. So I am not asserting which member changed shape or in which direction.

The only plausible source is this PR's sole non-additive changes: `readOnly` removal on `version` and `createdFrom`, and `required` removal on `parameters`/`actions` in 2026-11-01 only. In Python's ARM codegen, `readOnly` members are emitted as documented-but-not-settable *Variables* rather than `__init__` parameters, so removing `readOnly` moves a member between those categories. **That is inference, not verification** — the single-language asymmetry supports it, since a language-agnostic contract break would light up several generators.

### For the ARM review

Both gates share one root cause — the `readOnly` removals — so they warrant one decision rather than two. If someone with APIView access can paste the actual Python delta at that point, the inference above should be confirmed or corrected rather than left standing.

franklixuefei:
## Merging into the GA accumulation branch — breaking-change sign-off deferred by design

Recording why this merges with a failing `Swagger BreakingChange` check and no formal approval vote, so the decision is on the record rather than implied.

`juliagao-microsoft-chaos-2026-11-ga-promotion` is the **base branch for all 2026-11-01 GA work**. Changes accumulate into it as they become ready; around **11/01 it is locked down** and merged to `main` **after ARM team approval**. It is not a stalled PR chain — it is an accumulation branch working as intended.

That places both breaking-change gates at the wrong point in the pipeline for *this* PR:

- Both `Swagger BreakingChange` (4 × rule 1029) and `BreakingChange-Python-Sdk` are computed **against `main`**.
- This PR does not touch `main`. The promotion branch does, later.
- The same breaking changes will be **recomputed on the promotion PR** when it merges.

So the authoritative review is the **ARM review at lock-down**, and the per-PR gates here are advisory duplicates of a decision that gets made once, at the end. Resolving them now would mean paying the same tax on every GA PR into this branch for one decision.

Worth noting for whoever handles that review: **two of the four 1029 findings are already on the base branch** (`createdFrom`, from #45311) and are not introduced by this PR. This PR adds two, both on `version`.

Approval status: **@RenzoPrettoMS** reviewed and said *"Its fine by me"* — a favourable review comment rather than a recorded Approve vote. Merging on that basis is a deliberate call for an accumulation branch, not an oversight.

PR #45601, opened 2026-08-18
Add validation for multiple new API versions

### LINKED ISSUE #16731: Add a CI check that detects the presence of two new API versions in the same spec PR
When a spec PR adds two new API versions, the check would fail if the older api-version is not explicitly set in 'tspconfig.yaml' for each language emitter. The reason of setting the older API version is the older version is supposed to release at first. 

Check result should clarify like:
  - This PR will ship a preview package from API version 2026-01-02-preview
  - In order to release API version 2026-01-01, you will need to set 'api-version' in tspconfig.yaml

raych1:
resolves https://github.com/Azure/azure-sdk-tools/issues/16731
## Summary

- add a `multiple-new-api-versions` TypeSpec validation rule: doc is at https://github.com/Azure/azure-rest-api-specs/wiki/TypeSpec-Validation#multiplenewapiversions
- detect and report multiple newly introduced API versions
- add shared TypeSpec metadata support, tests, and documentation
- integrate the rule into TypeSpec validation workflows

## Testing

- added unit tests for the validation rule and metadata utilities

### DIFF
--- a/.github/shared/package.json
+++ b/.github/shared/package.json
@@ -26,6 +26,7 @@
     "./swagger-suppressions": "./src/swagger-suppressions.js",
     "./tag": "./src/tag.js",
     "./time": "./src/time.js",
+    "./typespec-metadata": "./src/typespec-metadata.js",
     "./test/examples": "./test/examples.js"
   },
   "bin": {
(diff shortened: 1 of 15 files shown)


mikeharder (changes requested):
blocking until we design how "static" vs "dynamic" rules work for local dev and CI

mikeharder (approved):

mikeharder (approved):

PR #45617, opened 2026-08-19
[Recovery Services Backup] Migrate CRR 2023-01-15 to TypeSpec

bharatpurwar:
## Summary

Migrates the Recovery Services Backup Cross Region Restore `2023-01-15` specification from Swagger to a sibling TypeSpec project.

## Changes

- Adds the `RecoveryServicesBackupCrr` TypeSpec project and generated Swagger.
- Preserves the historical CRR resource routes and parent hierarchy.
- Restores operation tags, paging metadata, and `x-ms-odata` metadata.
- Restores discriminator mappings and required-property compatibility.
- Preserves the access-token summary and bodyless `400` response.
- Preserves `CrossRegionRestore_Trigger` response codes.
- Preserves existing SDK request-body parameter names.

## Validation

- TypeSpec compilation completed successfully.
- Contract comparison was performed against the existing `2023-01-15` Swagger.

## Status

Draft while the remaining migration compatibility review is completed.


specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/examples/2023-01-15/AzureIaasVm/BackupAadProperties_Get.json
```diff
@@ -0,0 +1,21 @@
+{
+  "parameters": {
+    "api-version": "2023-01-15",
```
hiaga, line 3: Let's keep the folder name RecoveryServicesBackupCRR

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/examples/2023-01-15/AzureIaasVm/BackupAadProperties_Get.json
```diff
@@ -0,0 +1,21 @@
...
+    "azureRegion": "southeastasia",
```
hiaga, line 4: Also check if this "specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr" and will not conflict with the version increment in the other folder

hiaga:
question: main.tsp forces common-types v3 with a FIXME: Common type version v1 is not supported for now — does this change the emitted envelope vs the original hand-written swagger?

hiaga:
CrrJobRequest, CrossRegionRestoreRequest, and RecoveryPointGetAccessTokenBadRequest end in "Request"/"BadRequest" and will likely trip the TCGC csharp-model-suffix rule in TypeSpec Validation CI. Common fix used elsewhere in this repo (e.g. specification/ai/Face/models.common.tsp, specification/cognitiveservices/OpenAI.Inference/models.tsp): add #suppress "@azure-tools/typespec-client-generator-core/csharp-model-suffix" "Name kept for backward compatibility" right above each model.

hiaga:
nit: @@ doc(BackupResourceConfigResource.name, ") sets an explicitly empty doc string — consider a short doc or a documented suppression instead for clarity to future readers.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    fix-const-stuttering: true
+    flavor: "azure"
+    generate-samples: true
+    generate-fakes: true
+    head-as-boolean: true
+    inject-spans: true
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/resource-manager"
+  disable:
```
markcowl, line 50: You should not disable entire rules - this prevents these rules from being enforced going forward when this api evolves

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-core/casing-style": "Faithful migration of the frozen stable/2023-01-15 CRR contract. The original hand-written swagger uses model/property/interface names that are not PascalCase/camelCase (e.g. KEKDetails, IaasVMRestoreRequest, OperationStatusProvisionILRExtendedInfo, deferredDeleteSyncTimeInUTC, BackupUsageSummariesCRR). These names are part of the published wire contract and must be preserved byte-for-byte."
+    "@azure-tools/typespec-azure-core/composition-over-inheritance": "Faithful migration. The original swagger models a discriminated type hierarchy via allOf/extends (RestoreRequest, ProtectedItem, Job, OperationStatusExtendedInfo, etc.). Preserving the inheritance shape is required to emit an identical bms.json."
+    "@azure-tools/typespec-azure-core/documentation-required": "Faithful migration. A few frozen 2023-01-15 operations (e.g. AadProperties_Get, vaultstorageconfig reads) do not carry descriptions in the original contract; adding docs would change the emitted swagger."
```
markcowl, line 53: No reason whatsoever not to provide documentation.  Documentation updates are not breaking changes.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    generate-samples: true
+    generate-fakes: true
+    head-as-boolean: true
+    inject-spans: true
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/resource-manager"
+  disable:
+    "@azure-tools/typespec-azure-core/casing-style": "Faithful migration of the frozen stable/2023-01-15 CRR contract. The original hand-written swagger uses model/property/interface names that are not PascalCase/camelCase (e.g. KEKDetails, IaasVMRestoreRequest, OperationStatusProvisionILRExtendedInfo, deferredDeleteSyncTimeInUTC, BackupUsageSummariesCRR). These names are part of the published wire contract and must be preserved byte-for-byte."
+    "@azure-tools/typespec-azure-core/composition-over-inheritance": "Faithful migration. The original swagger models a discriminated type hierarchy via allOf/extends (RestoreRequest, ProtectedItem, Job, OperationStatusExtendedInfo, etc.). Preserving the inheritance shape is required to emit an identical bms.json."
```
markcowl, line 52: This rule does not apply to discriminated types - you are likely modeling discriminated types incorrectly so that emitters will not support them if you do this.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    flavor: "azure"
+    generate-samples: true
+    generate-fakes: true
+    head-as-boolean: true
+    inject-spans: true
+linter:
+  extends:
+    - "@azure-tools/typespec-azure-rulesets/resource-manager"
+  disable:
+    "@azure-tools/typespec-azure-core/casing-style": "Faithful migration of the frozen stable/2023-01-15 CRR contract. The original hand-written swagger uses model/property/interface names that are not PascalCase/camelCase (e.g. KEKDetails, IaasVMRestoreRequest, OperationStatusProvisionILRExtendedInfo, deferredDeleteSyncTimeInUTC, BackupUsageSummariesCRR). These names are part of the published wire contract and must be preserved byte-for-byte."
```
markcowl, line 51: model property capitalization should be preserved because it is part of the api.  model and interface names are not part of the api and should use appropriate capitalization - if this causes an issue for any sdks, you can use client decoration to resolve the difference.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-core/composition-over-inheritance": "Faithful migration. The original swagger models a discriminated type hierarchy via allOf/extends (RestoreRequest, ProtectedItem, Job, OperationStatusExtendedInfo, etc.). Preserving the inheritance shape is required to emit an identical bms.json."
+    "@azure-tools/typespec-azure-core/documentation-required": "Faithful migration. A few frozen 2023-01-15 operations (e.g. AadProperties_Get, vaultstorageconfig reads) do not carry descriptions in the original contract; adding docs would change the emitted swagger."
+    "@azure-tools/typespec-azure-core/no-legacy-usage": "Faithful migration. RecoveryPointResource and BackupResourceConfigResource carry a legacy optional 'location' and are modeled with Azure.ResourceManager.Legacy.TrackedResourceWithOptionalLocation to reproduce the original envelope exactly. They are read-only/singleton proxy resources, not true tracked resources."
```
markcowl, line 54: This should be inline suppressions.  It is likely yopu are using legacy apis much more than is necessary.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-core/no-openapi-client-extensions": "Faithful migration. The original CRR list operations emit x-ms-pageable with nextLinkName:null. This is reproduced via @@extension to keep the generated bms.json identical to stable/2023-01-15."
```
markcowl, line 55: Any meaningful thing you can put in openapi extensions can be represented in a higher levelkr typespec concept.  You should *never* use OpenAPI.  OpenAPI decoration ensures that the typespec view and the openapi view of your spec differs.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-core/no-string-discriminator": "Faithful migration. The original contract uses plain-string discriminators (objectType, jobType, protectedItemType). Converting to an extensible union would alter the emitted schema."
```
markcowl, line 56: There is no difference between a string enumeration and a string from an API perspective.  Use of open enumerations allows smooth evolution of the APIs.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/beyond-nesting-levels": "Faithful migration. The frozen CRR paths are subscription/location-scoped singletons whose nesting depth exceeds the ARM-standard limit; the legacy path shape must be preserved."
```
markcowl, line 57: This should have inline rather than global suppressions.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/lro-location-header": "Faithful migration. The original CRR trigger/operation-result operations (CrossRegionRestore_Trigger, CrrOperationResults_Get) do not emit a Location header on 202; adding one would change the contract."
```
markcowl, line 58: This needs point suppression.  Is it the case that location header was not emitted by the service (as is generally required for correct functioning), or was this just omitted from the original swagger?

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint": "Faithful migration. The CRR passive-stamp surface does not expose the standard Microsoft.RecoveryServices/operations list; that endpoint is owned by the main RecoveryServicesBackup spec, not this CRR-only project."
```
markcowl, line 59: This only requires one inline suppression

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/missing-x-ms-identifiers": "Faithful migration. The original array properties (e.g. details, recoveryPointTierDetails, clientScripts, errorDetails, tasksList) have no x-ms-identifiers in stable/2023-01-15; emitting identifiers would change the contract."
+    "@azure-tools/typespec-azure-resource-manager/no-empty-model": "Faithful migration. The original RecoveryPoints_GetAccessToken 400 response (AccessTokenBadRequestResponse) is an empty model in the frozen contract."
```
markcowl, line 61: Who uses this 400 endpoint?  Is this actually used by downstream consumers?  Does it provide any information that isn't standard for a BadRequest response, which is emitted by literally every api?

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/no-resource-delete-operation": "Faithful migration. BackupResourceConfigResource and RecoveryPointResource are read-only/singleton proxy resources; the original contract intentionally has no delete operation."
```
markcowl, line 62: Note that 'faithful reproduction' or 'textual equivalence' of swagger is a non-goal of conversion.  The API needs to be faithfully reproiduced *in typespec* and where alternate encodings that are standard produce equivalent API descriptiosn in swagger, we need to favor standard encodings over textual equivalence.

In this case - an inline suppression is required rather than this bl;anket suppression which will also impact new api versions.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    "@azure-tools/typespec-azure-resource-manager/no-response-body": "Faithful migration. Certain frozen CRR operations return success status codes with no body, matching the original stable/2023-01-15 contract."
```
markcowl, line 63: These need point suppressions

markcowl (changes requested):

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
```diff
@@ -0,0 +1,63 @@
...
+    namespace: "Azure.ResourceManager.RecoveryServices.CrossRegionRestore"
+    emitter-output-dir: "{output-dir}/sdk/recoveryservices-backup/{namespace}"
+  "@azure-tools/typespec-python":
+    service-dir: "sdk/recoveryservices"
+    emitter-output-dir: "{output-dir}/{service-dir}/azure-mgmt-recoveryservicescrossregionrestore"
+    namespace: "azure.mgmt.recoveryservicescrossregionrestore"
+    generate-test: true
+    generate-sample: true
+    flavor: "azure"
+  "@azure-tools/typespec-java":
```
weidongxu-microsoft, line 24: service-dir for Java should be `recoveryservicesbackup` (the 3rd segment of your package name)

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Done — removed the global  disable:  block.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Added the missing documentation and removed all  documentation-required  suppressions.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Fixed. The discriminated models are now modeled as discriminated types, and the global rule disable is gone. The remaining inline suppressions only apply to non-discriminated resource/list wrappers.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Done. The global casing disable is removed, model and interface symbols now use TypeSpec casing, and the existing SDK names are preserved with client decorators. The remaining inline casing suppressions are only for wire property names.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Removed the global suppression. Each remaining legacy API use now has its own inline suppression.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Removed the global suppression and replaced the paging and discriminator extensions with TypeSpec concepts. The five remaining extensions are  x-ms-odata  references to query models, and couldn’t find a higher-level TypeSpec equivalent

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: moved these suppressions inline, but kept the discriminators as strings because open unions add  enum / x-ms-enum  to the generated Swagger. The active backup spec handles these the same way.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Done — removed the global suppression and added it inline

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: added it to both 202 responses and removed the suppression.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Done — there’s now a single inline suppression on the service namespace in  main.tsp

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Removed the custom 400 response since it didn’t provide anything beyond the standard error response.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: Updated this to a single inline suppression on  BackupResourceConfigResources

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackupCrr/tspconfig.yaml
bharatpurwar: updated.

PR #45635, opened 2026-08-19
Fix Hub virtual network IPv6 peering contract

carenlim20:
## Summary

- correct `HubVirtualNetworkConnectionProperties.enableOnlyIPv6Peering` to match the deployed NRP JSON contract
- model the property as `boolean` instead of the `Enabled`/`Disabled` string union
- regenerate the 2025-09-01 and 2026-01-01 Swagger and examples from TypeSpec

## Versioning rationale

The property was introduced in API version 2025-09-01, so both 2025-09-01 and 2026-01-01 must describe the same deployed wire contract. The PR targets the `release-microsoft-network-2026-01-01` branch while correcting both generated stable projections on that branch.

## Validation

- `npx tsp compile .`
- generated `virtualWan.json` uses `enableOnlyIPv6Peering` with `type: boolean` in both versions
- generated examples use `true`/`false`
- pre-2025-09-01 API-version files are unchanged
- `git diff --check`


carenlim20:
Superseded by a PR from the existing intended branch nableOnlyIPv6Peering-boolean.

PR #45649, opened 2026-08-19
[Microsoft.Web] Add builderApps resource type in 2026-09-01-preview

mishapos:
## Summary

Adds a new top-level tracked resource type **`Microsoft.Web/builderApps`** in a new preview API version, **`2026-09-01-preview`**.

A builder app binds a source repository and branch to a managed build-and-host pipeline; one resource represents one app.

## What's in this PR

- `AppService/BuilderApp.tsp` — the new resource, its models, and the `BuilderApps` interface
- `AppService/main.tsp` — new `Versions` enum member + import
- `AppService/examples/2026-09-01-preview/` — six examples (CreateOrUpdate, Get, Update, Delete, ListByResourceGroup, ListBySubscription)
- `specification/web/cspell.yaml` — `builderapps`
- `AppService/readme.md` — `package-2026-09-01-preview` tag (the default tag is unchanged)
- Generated `preview/2026-09-01-preview/openapi.json` + copied examples

Every new declaration is gated on `@added(Versions.v2026_09_01_preview)`.

## Operations

| Verb | Operation | Shape |
| --- | --- | --- |
| PUT | `BuilderApps_CreateOrUpdate` | async LRO, `final-state-via: azure-async-operation` |
| GET | `BuilderApps_Get` | sync |
| PATCH | `BuilderApps_Update` | sync, body limited to `tags` + `variables` |
| DELETE | `BuilderApps_Delete` | async LRO, `final-state-via: location`, 202/204 |
| GET | `BuilderApps_ListByResourceGroup` | pageable |
| GET | `BuilderApps_ListBySubscription` | pageable |

## Day-0 properties

- **`source`** — immutable identity of the app (`x-ms-mutability: ["read", "create"]`): `type`, `repositoryId`, `repositoryFullName`, `branch`.
- **`provisioningState`** — read-only; `ResourceProvisioningState` plus `Creating` / `Updating` / `Deleting`.
- **`runtime`** — read-only reflection of the deployment currently being served: `url`, `activeRunId`, `activeDeploymentId`, `components[]`. Re-materialized by the platform on every deployment; never supplied by the caller.
- **`variables[]`** — configuration injected as environment variables. Values are `@secret` and write-only (`x-ms-secret: true`, mutability `create`/`update`), so they are never returned on GET or on the PUT response. Read-back, if ever needed, would be a separate `listSecrets`-style POST in a later version.

### Why `2026-03-01-preview` is retained

This PR adds `2026-09-01-preview` **alongside** the existing `2026-03-01-preview` rather than replacing it. TypeSpec-Azure guidance generally prefers a single live preview version, but retiring `2026-03-01-preview` would move the AI Gateway resource's API version, which is not acceptable. `v2026_03_01_preview` and `AIGateway.tsp` are untouched. The generated `preview/2026-03-01-preview/openapi.json` and the `stable/**` outputs do carry a small diff, but only the additive `readOnly: true` stamping described under *LintDiff* below — no other change to those versions.

### No child resources

Deployments and runs are deliberately *not* modelled as ARM resources. That history is unbounded, system-generated, and automatically garbage-collected, which fits poorly with ARM's PUT/DELETE model (RPC-Put-V1-16/17). It lives below ARM on the runtime APIs; the ARM resource carries only a thin read-only pointer (`runtime.activeRunId`). Everything else is inline under `properties`, which also keeps the resource a single cloneable document.

### Arrays, not maps

`runtime.components[]` is a typed array with `x-ms-identifiers: ["name"]`, per RPC-Put-V1-23.

### Deliberately held out of the day-0 surface

- **`linkedServices[]`** — omitted on the project lead's explicit instruction. The design doc still lists it as Phase 0, so the doc is stale on this point. Re-adding it later is purely additive (a new optional property) and non-breaking.
- **`settings`** (`autoDeploy`, `idleTimeoutHours`, `components[]`) — deferred to a later phase.
- **`resources[]`** and **`previewOverrides`** — still in draft; held out until the shape settles.

### Diff size

Adding an API version regenerates the entire Microsoft.Web AppService surface into a new `openapi.json` (~57.5k lines) plus copied example files. The hand-authored change is `BuilderApp.tsp`, the six examples, and three config lines — that's where review effort is best spent.

## Validation

- `tsp compile` — clean
- `tsv specification/web/resource-manager/Microsoft.Web/AppService` — all rules pass (FolderStructure, NpmPrefix, EmitAutorest, FlavorAzure, LinterRuleset, Compile, Format, SdkTspConfigValidation)

## CI status

All checks pass except `Swagger LintDiff`, which is explained below.

Three failures on the first push shared a single root cause: the branch was behind `main`, which had since added `stable/2026-03-15` and `stable/2026-07-15`. An earlier `git pull` merged those in without regenerating the emitted swagger, so the committed `preview/2026-09-01-preview/openapi.json` and `service.yaml` were stale. Merging latest `main` and recompiling fixed:

- **TypeSpec Validation** — CI regenerated the swagger and found it differed from what was committed.
- **Breaking Change (Cross-Version)** — was diffing against `stable/2026-07-15`, a version the stale generation did not contain, producing 46 phantom `RemovedPath` entries.
- **SDK Validation - Rust** — unrelated transient outage; the "Install Azure SDK Tools CLI" step received a GitHub timeout page. Passes on re-run.

The regeneration touched only `preview/2026-09-01-preview/**` and `service.yaml`. No `stable/**` or `preview/2026-03-01-preview/**` churn.

### LintDiff

LintDiff reports 343 "new" violations. This tag is brand new, so it has no before-baseline and the entire legacy Microsoft.Web surface is attributed to it. Attributing each violation to its enclosing path or definition:

| Count | Rule | Level | Belongs to `builderApps`? |
| --- | --- | --- | --- |
| 314 | `XmsExamplesRequired` | warning | No — pre-existing Microsoft.Web operations |
| 24 | `LatestVersionOfCommonTypesMustBeUsed` | warning | Yes — repo-wide common-types version; 2,146 pre-existing in this spec |
| 4 | `ProvisioningStateMustBeReadOnly` | **error** | Yes — see below |
| 1 | `EnumInsteadOfBoolean` | warning | Yes — read-only `internal` flag; 256 pre-existing in this spec |

The four errors were **real**, and are now fixed properly rather than suppressed.

The rule is evaluated against the *resolved* document. `$ref` siblings are discarded by the resolver, so emitting `readOnly` next to the `$ref` is not enough:

```json
"provisioningState": {
  "$ref": "#/definitions/BuilderAppProvisioningState",
  "readOnly": true   // <- dropped on resolution, rule never sees it
}
```

The fix is `use-read-only-status-schema: true` in `tspconfig.yaml`, which makes the emitter stamp `readOnly: true` onto the **status schema definition itself** — which is what survives resolution and what the rule actually inspects. Verified on the generated swagger: `definitions.BuilderAppProvisioningState` now carries `readOnly: true`. The `readme.md` suppression added in an earlier revision has been removed.

One caveat worth flagging for review: this option is **project-wide** and recognises status schemas by member *name* (`Succeeded` / `Failed` / `Canceled`). Regenerating therefore also adds `readOnly: true` to eight pre-existing enums in every API version, including four shipped stable ones — `BackupItemStatus`, `CustomDnsSuffixProvisioningState`, `KubeEnvironmentProvisioningState`, `OperationStatus`, `ProvisioningState`, `WorkflowProvisioningState`, `WorkflowStatus`, `WorkflowTriggerProvisioningState`. Note that `BackupItemStatus`, `WorkflowStatus` and `OperationStatus` are not provisioning states; they match only by member name. The change is purely additive (nothing removed), and it also clears 35 pre-existing instances of this same violation in the spec — but it is a surface change to shipped stable versions, so please confirm that is acceptable.

### Target branch

This PR targets **`release-Microsoft.Web-2026-09-01-preview`** rather than `main`, so the new preview version can be staged and reviewed before it lands on `main`.

The branch is kept current with that base. Falling behind it caused a spurious `Swagger Avocado` failure that reported `MISSING_README` against unrelated `specification/riskiq/**` folders — an artifact of unmerged folder-refactor commits in the base, not a problem with this change. Merging the base cleared it.


sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration(s), converged) against commit [`025ed52`](https://github.com/Azure/azure-rest-api-specs/pull/45649/commits/025ed526f5af6fbc98a9459fe46f14258d40bf1e). See inline comments for findings 1.

specification/web/resource-manager/Microsoft.Web/AppService/readme.md
```diff
@@ -102,6 +102,24 @@ directive:
...
 
+### Tag: package-2026-09-01-preview
+
+These settings apply only when `--tag=package-2026-09-01-preview` is specified on the command line.
+
+```yaml $(tag) == 'package-2026-09-01-preview'
+input-file:
+    - preview/2026-09-01-preview/openapi.json
+suppressions:
+    - code: ProvisioningStateMustBeReadOnly
```
sandipsh: **[NEW] 🟠 Warning** **[[ProvisioningStateMustBeReadOnly](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/provisioning-state.md#provisioningstatemustbereadonly----emitter-workaround-use-read-only-status-schema)]** `specification/web/resource-manager/Microsoft.Web/AppService/readme.md` - line 113 - This suppression hides a real emitted-schema problem for the new `BuilderApp`: `provisioningState` is emitted at `$.definitions.BuilderAppProperties.properties.provisioningState` as a `$ref` with sibling `readOnly: true` (openapi.json lines 40825-40828), and OpenAPI 2.0 tooling ignores `$ref` siblings. The result is that clients and validators can treat the LRO status as writable. The project supports the emitter fix, so suppressing this on a new resource preserves the broken shape instead of correcting it.

**Classification reasoning:** New in this PR: `BuilderApp`, its provisioning state, and this suppression do not exist in the 2026-07-15 base version.

**Approval context:** No `Approved-Suppression` label was observed at the session SHA. Please confirm whether suppression approval exists. If already approved, apply the label and resolve this conversation; otherwise fix the emitted shape or obtain approval.

**Suggested fix:** Add `use-read-only-status-schema: true` under the existing `@azure-tools/typespec-autorest` options in `tspconfig.yaml`, regenerate the OpenAPI, verify `BuilderAppProvisioningState` is emitted as read-only without the `$ref`-sibling shape, then remove this suppression. Removing it is safe only after that emitted check passes.

```yaml
options:
```

mishapos: Thanks — you were right, and my earlier justification for the suppression was wrong. I've applied the fix as suggested and pushed it, but it has a project-wide side effect that I'd like your call on before going further.

## The fix works

`use-read-only-status-schema: true` is now set in `tspconfig.yaml` and the `readme.md` suppression has been removed entirely.

My earlier reasoning was wrong because I only inspected the *property*:

```json
"provisioningState": {
  "$ref": "#/definitions/BuilderAppProvisioningState",
  "readOnly": true
}
```

…which is unchanged by the option, and whose sibling `readOnly` is discarded by the resolver. What I missed is that the emitter stamps `readOnly: true` onto the **status schema definition itself** — `definitions.BuilderAppProvisioningState` — which is what survives resolution and what the rule actually inspects. Verified on the generated swagger.

**Swagger LintDiff now passes with no suppression**, and `SuppressionReviewRequired` has dropped off the PR.

## The side effect

The option is project-wide and recognises status schemas by member *name* (`Succeeded` / `Failed` / `Canceled`), so regenerating stamped `readOnly: true` onto **8 pre-existing enums in every API version** — including four shipped stable ones:

`BackupItemStatus`, `CustomDnsSuffixProvisioningState`, `KubeEnvironmentProvisioningState`, `OperationStatus`, `ProvisioningState`, `WorkflowProvisioningState`, `WorkflowStatus`, `WorkflowTriggerProvisioningState`

Note that `BackupItemStatus`, `WorkflowStatus` and `OperationStatus` are not provisioning states at all — they match purely on member names.

CI now reports **140 `ReadonlyPropertyChanged` findings** across `stable/2025-03-01`, `stable/2025-05-01`, `stable/2026-03-15`, `stable/2026-07-15` and `preview/2026-03-01-preview`.

## Why I think those 140 are benign

I checked every property in the spec that references those 8 enums. Only **two** were not already `readOnly: true` at the property level:

| Property | Model used in a request body? |
| --- | --- |
| `Operation.status` | No — response-only (2 responses, 0 request bodies) |
| `OperationResultProperties.status` | No — never appears in a request body |

For every other affected property the definition-level `readOnly` is redundant with an existing property-level `readOnly`, so it is semantically a no-op. No property that a client could previously write becomes unwritable. The breaking-change tool flags these because it compares resolved schemas literally.

## Your call

1. **Keep the fix** and take the 140 findings to the breaking-change board on the "functionally a no-op" argument above.
2. **Go back to the narrowly-scoped suppression** and leave the shipped versions untouched. There's precedent for this in [`Microsoft.Management/ServiceGroups/suppressions.yaml`](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/management/resource-manager/Microsoft.Management/ServiceGroups/suppressions.yaml), which records hitting this exact problem: *"Adding `use-read-only-status-schema` to the tspconfig.yaml file resulted in a breaking `VersioningReviewRequired` in the existing preview API version. Recommended workaround is to suppress 'ProvisioningStateMustBeReadOnly'."*

I've left the fix in place on the branch so you can see the actual diff. Happy to revert to option 2 immediately if you'd prefer not to touch the shipped versions.

PR #45659, opened 2026-08-20
ContentSafety: add Unified Moderate in 2026-09-01-preview

huliang-microsoft:
Adds a new public preview api-version **`2026-09-01-preview`** to the Content Safety data plane, introducing the synchronous **Unified Moderate** operation. It lets applications and agent runtimes enforce an Agent Control Specification (ACS) policy against a prompt, completion, proposed tool call, or completed tool result. It ships as part of the existing [Content Safety](https://learn.microsoft.com/azure/ai-services/content-safety/overview) product.

## What's new in `2026-09-01-preview`

New `UnifiedOperations` interface (synchronous):

- `POST /content:unifiedModerate` resolves an explicit policy or an existing resource binding, evaluates the selected ACS intervention point, and returns the enforced caller verdict plus the complete ACS verdict.

New models: `UnifiedModerateOptions`, `UnifiedModerateContext`, `UnifiedModerateSource`, `UnifiedModerateResult`, `UnifiedModerateVerdict`, `AcsDecision`, `AcsVerdict`, `AcsHarmResult`, `AcsHarmDetail`, `AcsApproval`, `AcsTransform`, and `AcsEvidence`.

The public contract intentionally:

- keeps APIM-derived account and partner metadata out of public request headers;
- uses body property `targetResourceId` for bound-resource policy resolution;
- carries plain text or JSON-encoded structured content as a string, matching existing RAI and streaming contracts; and
- exposes `harmResults` as a strongly typed map keyed by policy-configured harm category.

`2026-09-01-preview` is a superset of `2026-07-01-preview`: it retains text/image analysis, shield prompt, protected material detection, text blocklists, and Content Provenance Detection, and adds Unified Moderate. The shipped `2026-07-01-preview` contract is unchanged.

Release plan: **TODO - add release plan link**.

## API Info: The Basics

Is this review for:

- [x] a public preview

### Change Scope

* Design Document: Unified Moderate - Azure Public API Review (internal Word document; upload/link pending)
* New API version: `2026-09-01-preview`
* New path:
  * `POST /content:unifiedModerate`

### Notes for reviewers

* TypeSpec source lives in `specification/cognitiveservices/ContentSafety/` (unified Content Safety spec); Unified Moderate is gated to `2026-09-01-preview` via `@added`.
* The `2026-07-01-preview` generated contract remains byte-identical to `main`.
* The operation ID is `UnifiedOperations_UnifiedModerate`, matching the Platform APIM2 and Platform Resource Provider registrations.
* The change includes inherited examples for the new version plus one Unified Moderate policy example with Hate, Sexual, SelfHarm, Violence, Blocklist, and Jailbreak results.
* Generated with the repo-pinned emitters (`@typespec/compiler` 1.15.0 / `@azure-tools/typespec-autorest` 0.71.0).
* Related implementation changes:
  * Platform APIM2: https://msazure.visualstudio.com/Cognitive%20Services/_git/Platform-APIM2/pullrequest/16651600
  * Platform Resource Provider: https://msazure.visualstudio.com/Cognitive%20Services/_git/Platform-ResourceProvider/pullrequest/16651595
  * AACS backend: https://msazure.visualstudio.com/Cognitive%20Services/_git/Service-ContentModerator-vnext/pullrequest/16651632


specification/cognitiveservices/ContentSafety/examples/2026-09-01-preview/UnifiedModerate.json
```diff
@@ -0,0 +1,75 @@
...
+    }
+  },
+  "responses": {
+    "200": {
+      "body": {
+        "verdict": "blocked",
+        "reason": "content_policy_violation",
+        "acsVerdict": {
+          "decision": "deny",
+          "reason": "content_policy_violation",
```
lycroy-work, line 28: should we keep one top-level "reason" field only?

specification/cognitiveservices/ContentSafety/models.tsp
```diff
@@ -275,6 +275,224 @@ model TextProtectedMaterialAnalysisResult {
   detected: boolean;
 }
 
+@added(ContentSafety.Versions.v2026_09_01_preview)
+@doc("The workload event source evaluated by Unified Moderate.")
+union UnifiedModerateSource {
+  string,
+
+  @doc("A prompt submitted to a model or agent.")
```
liamcrumm: I was originally thinking this might have PreRun and PostRun mapping to agent_startup / agent_shutdown. Don't think it will be immediately useful for customers but might make sense for future cases to conform to ACS hooks?

specification/cognitiveservices/ContentSafety/models.tsp
```diff
@@ -275,6 +275,224 @@ model TextProtectedMaterialAnalysisResult {
...
+@doc("A request to evaluate and enforce an applicable policy for one workload event.")
+model UnifiedModerateOptions {
+  @doc("An optional built-in or account-scoped policy identifier. Omit this property to resolve the policy bound to targetResourceId.")
+  policyId?: string;
+
+  @doc("The Azure resource identifier whose policy binding is resolved. Required when policyId is omitted. The target resource must already be bound under the Content Safety account serving the request.")
+  targetResourceId?: string;
+
+  @doc("The workload event source. The source selects the policy intervention point.")
+  source: UnifiedModerateSource;
```
liamcrumm, line 332: Should we have optional role here to handle AOAI AppliedFor? I think it scopes for (source,role). If we have a policy scoped to (SYSTEM, Prompt) for example it'd default to the role implied by the source

specification/cognitiveservices/ContentSafety/models.tsp
lycroy-work: Need an update in FD side for this as well once decided. Currently it is accepting Input/Output only

specification/cognitiveservices/ContentSafety/models.tsp
```diff
@@ -275,6 +275,224 @@ model TextProtectedMaterialAnalysisResult {
...
+  @doc("Optional approval metadata for a deny decision. Unified Moderate does not invoke an approval workflow, so the content remains blocked.")
+  approval?: AcsApproval;
+
+  @doc("Canonical transform metadata. Required when the decision is transform.")
+  transform?: AcsTransform;
+
+  @doc("Optional policy evidence and verification pointers.")
+  evidence?: AcsEvidence[];
+
+  @encodedName("application/json", "harm_results")
```
lycroy-work: Should we make this camelCase?

specification/cognitiveservices/ContentSafety/examples/2026-09-01-preview/UnifiedModerate.json
huliang-microsoft: @liamcrumm  said that the fd reason may differ from acs reason/decision so let's keep them both.

specification/cognitiveservices/ContentSafety/models.tsp
huliang-microsoft: Yes I think so

specification/cognitiveservices/ContentSafety/models.tsp
huliang-microsoft: > I was originally thinking this might have PreRun and PostRun mapping to agent_startup / agent_shutdown. Don't think it will be immediately useful for customers but might make sense for future cases to conform to ACS hooks?

Yes, but the order is to have the implementation supported in AACS first then we can update the API. We cannot add anything which is not usable here

specification/cognitiveservices/ContentSafety/models.tsp
huliang-microsoft: I think we may have but I dont think that's fully functional in AACS. e.g. no role check for completion. We can add this support but we need to make change in AACS first and adjust the code as we need

specification/cognitiveservices/ContentSafety/main.tsp
```diff
@@ -43,6 +43,9 @@ enum Versions {
   v2024_09_01: "2024-09-01",
 
   @doc("Public preview that adds asynchronous Content Provenance Detection operations.")
-  @Azure.Core.previewVersion
   v2026_07_01_preview: "2026-07-01-preview",
```
weidongxu-microsoft: Recommended to retain only the last preview (and stables).

specification/cognitiveservices/ContentSafety/routes.tsp
```diff
@@ -83,6 +83,25 @@ interface ImageOperations {
   >;
 }
 
+@added(ContentSafety.Versions.v2026_09_01_preview)
+interface UnifiedOperations {
+  @summary("Unified Moderate")
+  @doc("A synchronous API that evaluates input, output, a proposed tool call, or a completed tool result against an applicable Agent Control Specification policy and returns the enforced content and policy verdict.")
+  @route("/unifiedModerate")
```
weidongxu-microsoft: Could this follow the service's existing `resource:action` route pattern instead of introducing the standalone verb path `/unifiedModerate`? Existing Content Safety actions use routes such as `/image:analyze`, `/text:analyze`, `/text:shieldPrompt`, and `/provenance:detect`. The [Azure REST API Guidelines for performing an action](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#actions) recommend appending `:{action}` to a resource or collection URL. For example, consider `/content:unifiedModerate`, or another appropriate resource noun.

For reference only, here is the guidance on [repeatability headers](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#repeatability). I am not requesting those headers here because this appears to be a synchronous evaluation without a documented non-idempotent side effect, consistent with the service's existing analysis operations.

huliang-microsoft: Changed to content:unifiedModerate

specification/cognitiveservices/ContentSafety/main.tsp
huliang-microsoft: done

weidongxu-microsoft:
> ## Data-Plane API Review
> _Automated advisory review by Copilot. This is one non-deterministic pass over a narrow semantic scope and does not replace complete human API review._
> 
> **Findings:** 🔴 0 · 🟡 1 · 💡 0
> 
> **Index:** DP-VERSION-03 — Preserve the published preview version (`specification/cognitiveservices/ContentSafety/main.tsp:47`)
> 
> **Not reviewed:** deterministic compiler/linter concerns, naming and documentation quality, grey-area design questions, runtime behavior, SDK customizations, and unchanged API surface.
> 
> > 🔍 _Automated data-plane API review by [Data-Plane API Review](https://github.com/Azure/azure-rest-api-specs/actions/runs/32934506364). Advisory only._

We may need to update the agentic reviewer -- Mark C requires that data-plane only keep the latest preview api-version. So when adding new preview,  deleting the prior preview api-version is expected.

weidongxu-microsoft (approved):

weidongxu-microsoft:
> **Agent-generated follow-up review:** The notes below are considerations for the author, not requested changes.

I reviewed the API once more from a customer's perspective. The basic text-moderation example is understandable, but a few areas may be worth clarifying so customers can construct the other request variants without relying on out-of-band knowledge:

1. **Policy selection:** `policyId` and `targetResourceId` are both optional in the schema. Consider documenting what happens when both or neither are supplied, how customers discover valid built-in or account-scoped policy IDs, which target resource types are supported, and how the prerequisite policy binding is created.
2. **Source-specific request shapes:** The contract defines `input`, `output`, `pre_tool_call`, and `post_tool_call`, but provides only an `input` example. Consider adding a per-source field table or examples that explain what belongs in `content` versus `toolArguments`, which tool fields apply to each source, the structured-content JSON shape, and relevant size limits.
3. **Transforms:** `AcsTransform.path` is described as a canonical path without naming its grammar. Consider identifying the syntax, such as JSON Pointer, and adding a transformed-response example showing how `path`, `value`, and returned `content` relate.

The rest of this pass looked good: `/content:unifiedModerate` follows the action-route convention, the preview-version lifecycle is consistent with Mark C, prior emitted versions remain unchanged, the standard Azure.Core error envelope is used, and repeatability headers do not appear necessary for the documented synchronous evaluation.

Connor-Wood (approved):

weidongxu-microsoft (approved):

PR #45663, opened 2026-08-20
[AzureResilienceManagement] Add new API version 2026-08-31-preview

arpja:

## Summary

Adds new API version `2026-08-31-preview` for `Microsoft.AzureResilienceManagement`, introducing optional Azure Health Model (AHM) and SLI monitoring on Drills.

> Note: these changes were originally authored in `Azure/azure-rest-api-specs-pr` on branch `user/jainkapil/ahm-sli-integration` (PR Azure/azure-rest-api-specs-pr#29996). That PR was raised against the wrong repository; this PR ports the identical change set to the public `Azure/azure-rest-api-specs` repo. The `-pr` PR should be closed.

### Changes

- New API version `2026-08-31-preview` added to `main.tsp` (`Versions` enum, common-types v6), `readme.md` (`package-2026-08-31-preview` tag) and `service.yaml`.
- New models: `HealthModelMonitoringProperties` (single AHM via mandatory `discoveryRuleId`), `SliMonitoringProperties` (identity + max 2 typed SLIs), `SliSelection` (`sliId`, `type`), `SliAttentionStatus` (per-SLI exists / typeMatch / per-AMW RBAC).
- New enums: `SliType` (`Availability` | `Latency`), `SliTypeMatchState` (`Matched` | `Mismatched`).
- `DrillProperties`: added `healthModelMonitoringProperties`, `sliMonitoringProperties`.
- `DrillPatchProperties`: nullable variants to support null-to-clear (replace-whole-list) semantics.
- `AttentionReason`: added `monitoringSourceNotConfigured`, `healthModelExists`, `discoveryRuleExists`, `drillRbacOnHealthModel`, `rbacNeededForDrillOnHealthModel`, `drillRbacOnSli`, plus `sliAttentionStatuses[]`.
- Examples: new `2026-08-31-preview` example set; Drill examples carry AHM/SLI payloads.
- `cspell.yaml`: added `amws`, `slis`.

All changes are additive and scoped to the new preview API version; no prior versions are modified.

### Local validation

- `tsp compile . --warn-as-error` — passes.
- Checked-in `preview/2026-08-31-preview/openapi.json` is byte-identical to freshly generated output.
- `tsp format --check` and `prettier --check` — clean.

## Purpose of this PR

- [x] New API version for an existing resource provider.

## Due diligence checklist

  [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md).
- [ ] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created.


harshb-15 (approved):

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drill.models.tsp
```diff
@@ -270,6 +278,89 @@ model MonitoringPropertiesOfDrill {
...
+/**
+ * Model that represents the SLI monitoring properties of a Drill.
+ */
+@doc("SLI monitoring properties of a Drill. At most two SLIs may be selected: at most one Availability and one Latency.")
+@added(Versions.v2026_08_31_preview)
+model SliMonitoringProperties {
+  @doc("Identity that the Drill uses to read evaluated SLI results from each SLI's destination Azure Monitor Workspace. The Drill is granted Monitoring Reader on every destination AMW of every selected SLI for this identity.")
+  identity: AssociatedIdentity;
+
+  @doc("The SLIs selected for Drill monitoring. Maximum of two entries: at most one Availability and one Latency. Duplicate types or duplicate SLI Ids are rejected.")
```
umpatel, line 307: Nit: this doc states the validation constraints (max 2 SLIs, no duplicate 	ype, no duplicate sliId) in prose only. The @maxItems(2) is schema-enforced, but the 'no duplicates' rule isn't expressible in TypeSpec/JSON Schema and will need service-side validation (with a corresponding 400 error). Worth double-checking the service implementation actually rejects duplicates, since API consumers will only see this behavior at runtime, not from the spec.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+  @visibility(Lifecycle.Read)
+  report?: DrillRunReport;
+}
+
+@added(Versions.v2026_08_31_preview)
+@doc("Public, read-only view of the report generated for a Drill Run. Exposes status and pointers only - internal storage locations are never surfaced.")
+model DrillRunReport {
+  @doc("Overall report generation status for the Drill Run.")
+  @visibility(Lifecycle.Read)
+  generationStatus?: ReportGenerationStatus;
```
umpatel: Per repo convention (e.g. AttentionReason fields elsewhere in this file), every newly introduced property should carry its own @added(Versions.v2026_08_31_preview) decorator, not just the containing model. Even though DrillRunReport itself is new in this version, tagging each property individually (generationStatus, stageStatuses, vailableFormats, lastGeneratedTimestamp, schemaVersion, isFinalized, lastError) keeps versioning metadata self-describing and future-proof if a property is ever added to an older model via merge, and matches how the rest of this PR versions new fields on existing models.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drill.models.tsp
arpja: This validation is already added in service

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
```diff
@@ -422,3 +422,66 @@ union ResourceTypeCategories {
...
+  @doc("Report generation failed. Details are captured in the lastError field.")
+  Failed: "Failed",
+}
+
+/**
+ * Enum for the formats a Drill Run report can be downloaded in.
+ */
+@doc("Format of a Drill Run report.")
+@added(Versions.v2026_08_31_preview)
+union ReportFormat {
```
umpatel: Consider renaming this to DrillReportFormat (or moving it to a common/shared enums file) — as written, ReportFormat is a fairly generic name that could collide with or be confused by other report-generating scenarios (e.g. recovery plan reports, recovery job reports) that may want to reuse the same concept in the future. Scoping the name to Drill now avoids a breaking rename later, or alternatively promoting it to a common location makes it reusable as-is.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
```diff
@@ -422,3 +422,66 @@ union ResourceTypeCategories {
...
+  @doc("Declared type does not match the SLI's actual category.")
+  Mismatched: "Mismatched",
+}
+
+/**
+ * Enum for the report generation status of a Drill Run report.
+ */
+@doc("Report generation status.")
+@added(Versions.v2026_08_31_preview)
+union ReportGenerationStatus {
```
umpatel: Same consideration as ReportFormat below: either rename this to DrillReportGenerationStatus to scope it clearly to the Drill Run report feature, or move it to a common enums location so it can be reused by other report-generation scenarios in the future without a naming collision or breaking rename.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+
+  @doc("Report generated for this Drill Run.")
+  @added(Versions.v2026_08_31_preview)
+  @visibility(Lifecycle.Read)
+  report?: DrillRunReport;
+}
+
+@added(Versions.v2026_08_31_preview)
+@doc("Public, read-only view of the report generated for a Drill Run. Exposes status and pointers only - internal storage locations are never surfaced.")
+model DrillRunReport {
```
umpatel: Consider renaming DrillRunReport to DrillGenerateReportResult (or DrillGenerateReportResponse) to better reflect that this is the result/response shape returned by the generateReport action (and surfaced via DrillRunProperties.report), following the *Result/*Response naming convention used elsewhere for action outputs in this spec (e.g. ReportDownloadUrlResponse, and the @@clientName(..., "*Result", "csharp") renames applied to other action responses in client.tsp).

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+  lastAttemptTimestamp?: utcDateTime;
+
+  @doc("Error from the last failed report generation attempt for this stage.")
+  @visibility(Lifecycle.Read)
+  lastError?: ErrorDetails;
+}
+
+@added(Versions.v2026_08_31_preview)
+@doc("Request to mint a short-lived, read-only download URL for a Drill Run report.")
+model ReportDownloadUrlRequest {
```
umpatel: "ReportDownloadUrlRequest"/"ReportDownloadUrlResponse" read ambiguously (as if downloading a URL, rather than a URL used to download the report). Since every other action request/response in this file mirrors its action name exactly (e.g. MarkAsCompleteRequest/MarkAsCompleteResponse for markAsComplete, DrillRunFailoverRequest for failOver), consider renaming these to ListReportDownloadUrlRequest / ListReportDownloadUrlResponse to match the listReportDownloadUrl action name and remove the ambiguity, rather than keeping a name derived from the noun phrase "report download URL" alone.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+@added(Versions.v2026_08_31_preview)
+@doc("Response containing a short-lived, read-only download URL for a Drill Run report.")
+model ReportDownloadUrlResponse {
+  @doc("Format of the report the download URL refers to.")
+  @visibility(Lifecycle.Read)
+  format?: ReportFormat;
+
+  @doc("Short-lived, read-only URL to download the report.")
+  @visibility(Lifecycle.Read)
+  downloadUrl?: url;
```
umpatel, line 169: Two suggestions for this property: 1) Rename downloadUrl to reportUrl for consistency with the proposed rename above (and to avoid "download" appearing redundantly given the operation is already listReportDownloadUrl). 2) Since this URL embeds a SAS signature (per the example: sv=...&sr=b&sig=...&se=...&sp=r) that grants time-limited read access to the report, consider marking it @secret so it is never logged/cached by generic tooling and is handled consistently with other credential-bearing properties in ARM specs, even though it is intentionally returned in a 200 response (RPC012 secret-retrieval-via-POST list* action pattern already covers that part).

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+  @visibility(Lifecycle.Read)
+  lastGeneratedTimestamp?: utcDateTime;
+
+  @doc("Schema version of the generated report content.")
+  @visibility(Lifecycle.Read)
+  schemaVersion?: string;
+
+  @doc("Indicates whether the report is finalized and immutable.")
+  @visibility(Lifecycle.Read)
+  isFinalized?: boolean;
```
umpatel: Consider avoiding boolean-typed fields per Azure REST API Guidelines - booleans are hard to extend later (a third state inevitably shows up, e.g. "Finalizing"/"PendingFinalization") and are less self-descriptive in API docs/SDKs than a named enum. Suggest replacing isFinalized with an extensible union/enum instead, e.g. finalizationState: ReportFinalizationState { NotFinalized, Finalized } (open string enum so more states can be added later without a breaking change), consistent with how this same PR already models other tri-state/extensible concepts as unions (ReportGenerationStatus, SliTypeMatchState, RBACState) rather than booleans.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
```diff
@@ -169,6 +169,34 @@ interface DrillRuns {
...
+      ArmLroLocationHeader &
+      Azure.Core.Foundations.RetryAfterHeader
+  >;
+
+  @route(ServiceGroupRoute)
+  @added(Versions.v2026_08_31_preview)
+  @armResourceAction(DrillRun)
+  @post
+  @doc("This returns a short-lived, read-only URL to download the report for this Drill Run. The URL expires at the returned expiryTimestamp and grants access to that single report only.")
+  listReportDownloadUrl is ArmResourceActionSync<
```
umpatel, line 193: Question/naming check: this action name uses "list", which per RPC012 is the ARM convention for POST actions that retrieve secret/credential-like data - it does not imply the response must be a collection (e.g. listKeys also returns a single object). Confirming here: ReportDownloadUrlResponse returns a single downloadUrl/format/expiryTimestamp object, not an array/value+nextLink collection. If that is correct (one URL per call, scoped to the requested format), the current singular response shape is fine as-is. If instead the intent is ever to return multiple report format URLs in one call, the response should switch to a plural array (e.g. reportUrls: ReportDownloadUrlResponse[]) to match. Please confirm the intended cardinality so the name/shape stay consistent.

umpatel (changes requested):

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Done in 1587896caba - every property on `DrillRunReport`, `ReportStageStatus`, `ListReportDownloadUrlRequest` and `ListReportDownloadUrlResponse` now carries its own `@added(Versions.v2026_08_31_preview)`.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
arpja: Done in 1587896caba - renamed to `DrillReportGenerationStatus`.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
arpja: Done in 1587896caba - renamed to `DrillReportFormat`.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Done in 1587896caba - renamed to `ListReportDownloadUrlRequest` / `ListReportDownloadUrlResponse` to mirror the `listReportDownloadUrl` action name. `client.tsp` C# renames updated to match.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Done in 1587896caba - replaced `isFinalized: boolean` with `finalizationState?: DrillReportFinalizationState` (extensible union: `NotFinalized` / `Finalized`), so additional states can be added later without a breaking change. Example payload updated accordingly.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
arpja: Confirming: cardinality is intentionally singular. One call returns one short-lived SAS URL for the single requested format (defaulting to `Html`); a caller wanting a different format issues another call. The `list` prefix is purely the RPC012 convention for POST actions returning credential-like data, as you note. We are not planning to return multiple format URLs in one call, so the singular response shape stays.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Pushing back on this one, if that's ok. `DrillRunReport` is not only the `generateReport` action result - it is also surfaced as a read-only resource property, `DrillRunProperties.report`, and is projected from the same stored entity in both places. Naming it `DrillGenerateReportResult` would make the resource property read as `drillRun.properties.report: DrillGenerateReportResult`, which implies it is the transient output of one action invocation rather than the current durable state of the report for that Drill Run. The `*Result`/`*Response` convention you cite applies to types that exist only as action outputs (`ListReportDownloadUrlResponse`, `MarkAsCompleteResponse`), which is not the case here. `DrillRunReport` also reads consistently with the other resource-state models in this file. Happy to revisit if you still feel strongly, but the dual use is the reason I'd prefer to keep the noun form.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Two parts here:

**Rename `downloadUrl` -> `reportUrl`:** happy to do this if you'd still like it, though note the containing model is now `ListReportDownloadUrlResponse`, so `downloadUrl` is no longer ambiguous in context and `reportUrl` arguably loses the signal that it is a download link rather than a portal/blob address. Let me know and I'll apply it.

**`@secret`:** could you help me understand the value add here? Concretely - what does marking it `@secret` buy us beyond the work of adding it? My understanding is that `@secret` mainly affects SDK/emitter handling (redaction in `ToString`/logging), and:

- the URL is already short-lived and single-report scoped, with `expiryTimestamp` returned alongside it;
- it is the entire point of the response - the caller must read it, pass it to an HTTP client, and typically hand it to a browser or download utility, so redacting it in logs does not remove it from the places it actually flows;
- RPC012 (`list*` POST for credential-bearing data) already covers the transport/RBAC side, which is the protection that matters;
- `@secret` on the only meaningful field of the response risks making the object unhelpfully opaque in diagnostics without materially reducing exposure.

If there is a concrete downstream behaviour I'm missing (a specific SDK/telemetry/ARM pipeline that treats `@secret` differently in a way that matters here), I'm glad to add it - I just want to make sure it is buying real protection rather than ceremony. Also worth noting the alternative framing: if the concern is the SAS signature specifically, that is a property of how the service mints the URL, not of the contract shape, and would not change by annotating the field.

arpja: Following up on the `@secret` question with what we found when we actually traced it.

**What `@secret` emits.** It is a real built-in compiler decorator, and the autorest emitter does exactly two things with it:

```js
if (isSecret(program, typespecType)) {
    newTarget.format = "password";
    newTarget["x-ms-secret"] = true;
}
```

**First concern - it conflicts with the fix we just made.** `format` is assigned unconditionally, so on a `url`-typed property `@secret` overwrites `format: "uri"`. We compiled it to confirm rather than assume:

| | `downloadUrl?: url` | `@secret` + `url` |
|---|---|---|
| emitted | `"type": "string", "format": "uri"` | `"type": "string", "format": "password"` |

So adding `@secret` here would revert the URI-properties fix the ARM review bot asked for earlier in this same PR. We cannot have both.

**Second concern - we could not establish that it changes anyone's behaviour.** This is the honest position: we do not know of a service, SDK or pipeline that would actually stop logging this field because of the annotation, and we could not find one. What we checked:

- `azure-sdk-for-net`, `-java`, `-python`, `-js`, `-go` and `autorest.csharp`: **no references to `x-ms-secret` at all**. No generator appears to read it.
- `oav` is the main real consumer, and it uses it to *relax* validation, not to protect anything - "Ignore required validation rule in response when x-ms-secret is true and the httpMethod is POST", plus round-trip and `required` checks being skipped. Its effective meaning is closer to "this value is not echoed back, do not fail validation when it is absent".
- The SAS redaction that genuinely exists in the SDKs is pipeline-level and generic: `HttpLoggingPolicy` allowlists query parameters, so `sig` is already redacted by default, independent of any spec annotation.

If you know of a concrete downstream consumer that behaves differently when `x-ms-secret` is set, we would genuinely like to know - that would change the calculus and we will add it.

**On splitting the field.** We did consider returning the base URL and the SAS token as separate properties so only the token carries `@secret`. Two problems:

1. It is unnecessary extra work on both sides. The service would have to decompose a URL it just constructed, and every client would have to stitch it back together (`reportUrl` + `?` + `sasToken`), with the usual separator and encoding pitfalls. The stitched result is byte-identical to what we would have returned as one field, so the exposure surface is unchanged the moment the caller uses it - we would be paying real implementation and usability cost for no measurable protection.
2. It would *create* a rule obligation rather than satisfy one. A property named `sasToken` ends with "token", so LintDiff `XMSSecretInResponse` and the TSP `secret-prop` (RPC-v1-13) rule would then require `x-ms-secret`. Neither rule fires on `downloadUrl` today - both match only names ending in access/credential/secret/password/key/token/auth/connection **and** `type: string`, and ours is the `url` scalar.

**Precedent.** Billing and CostManagement both return a full SAS download URL as a single `downloadUrl` property with no `x-ms-secret` and no splitting (`Microsoft.Billing/.../InvoicesGet.json`, `Microsoft.CostManagement/.../client.tsp`). We did not find an ARM spec that pairs a split `sasUri`/`sasToken` with `x-ms-secret`.

So our proposal is to leave this as a single `url`-typed property, keeping `format: "uri"`. RPC012 (`list*` POST for credential-bearing data), the short expiry, the single-report scope and the returned `expiryTimestamp` are what actually constrain exposure here. Very happy to revisit if you can point at a consumer that would behave differently - we just want to avoid imposing parsing and stitching work on ourselves and on every client for an annotation whose practical effect we could not demonstrate.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Confirmed - verified in the latest commit (1587896c). Thanks for the quick turnaround.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
umpatel: Confirmed - DrillReportFormat looks good. Thanks.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/enums.tsp
umpatel: Confirmed - DrillReportGenerationStatus looks good. Thanks.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Confirmed - ListReportDownloadUrlRequest/Response resolves the ambiguity, and matches the action name. Thanks.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Confirmed - finalizationState with the extensible NotFinalized/Finalized union is exactly the right shape. Thanks for making it extensible.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
umpatel: Understood and agreed - singular is correct for the documented behavior (one format per call). No change needed, thanks for confirming.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Fair point, and I agree - the dual usage (action result + durable DrillRunProperties.report state) is a real distinction I hadn't weighed enough. DrillGenerateReportResult would indeed misrepresent the resource-state usage. Keeping DrillRunReport as-is, no change needed.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Thanks for tracing this all the way through the emitter/SDKs - that's convincing. On @secret: agreed, don't add it - the format conflict with url/uri plus the lack of any consumer that treats x-ms-secret differently means it would cost real correctness (breaking the URI-properties fix) for no measurable protection, and RPC012 + short expiry + expiryTimestamp already constrain exposure appropriately. On the rename: given downloadUrl now lives inside ListReportDownloadUrlResponse the ambiguity is resolved by context, and reportUrl would arguably lose the 'this is a download link, not a canonical address' signal as you noted - let's leave it as downloadUrl. No further changes needed on either point.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Re-raising this one - I don't think the dual-usage argument actually resolves the naming problem, it just relocates it.

Looking at the model's own contents (`generationStatus`, `stageStatuses`, `availableFormats`) and its doc comment ("Exposes status and pointers only - internal storage locations are never surfaced"), this type never contains the report itself. It contains metadata *about* report generation - state and download-format availability. So calling it `DrillRunReport` is misleading in **both** places it's used: as the `generateReport` action response, a consumer would reasonably expect "the report," not a status descriptor; and as `DrillRunProperties.report`, it reads as if the resource embeds the report, when it embeds report-generation status.

A generic name that reflects what the type actually models - status/metadata, not content - would fit both use sites correctly, e.g. `DrillReportStatus` or `DrillReportSummary`. That avoids the action-result-vs-resource-state tension entirely, since neither name implies "this is the output of one specific action call."

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
umpatel: Thanks for the detailed trace - the `format: "uri"` vs `format: "password"` conflict is real and confirmed. But that conflict is specific to the `@secret` decorator, not to emitting `x-ms-secret` itself - there's a way to get one without the other.

`@secret` unconditionally does two things in the autorest emitter: `format = "password"` and `x-ms-secret = true`. If you instead use `@extension("x-ms-secret", true)` (from `@typespec/openapi`, already usable in this spec) directly on the property, you get the `x-ms-secret: true` marker without touching `format`, so `format: "uri"` is preserved:

```tsp
@extension("x-ms-secret", true)
downloadUrl?: url;
```

That resolves the technical conflict you found. On whether it's worth doing at all, given nothing currently consumes `x-ms-secret`: RPC-v1-13 / LintDiff's `XMSSecretInResponse` rule exists specifically so ARM and reviewers can *identify* secret-bearing response fields from the spec itself, independent of whether an SDK currently acts on it - that's the point of a static declaration ahead of consumption. Since this field literally carries a SAS-signed URL (i.e., a bearer credential in the query string), I'd still mark it, using `@extension` rather than `@secret` to sidestep the format clash. Open to leaving it out if the team has a documented exception for this pattern (as with Billing/CostManagement), but wanted to flag that a conflict-free option exists.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/recoveryPlan/recoveryActions.tsp
```diff
@@ -119,3 +125,90 @@ model RecoveryActionRequest {
...
+
+  string,
+}
+
+@added(Versions.v2026_08_31_preview)
+@doc("SKU details for a resource feasibility review, used for both the current target SKU and the recommended alternate SKUs.")
+model SkuDetails {
+  @added(Versions.v2026_08_31_preview)
+  @doc("The Azure SKU name.")
+  sku: string;
```
umpatel: Naming/grammar check: all newly-added properties in this diff do have @added(Versions.v2026_08_31_preview) tags - verified across every changed .tsp file (drill.models.tsp, drillRun.models.tsp, enums.tsp, resourceProtection.tsp, recoveryActions.tsp, common.tsp). No gaps found there.

On naming: inside SkuDetails, naming this field sku is redundant with the model name (SkuDetails.sku reads like DrillRunReport.report). Since this model already scopes the concept, 
ame (or skuName if you want to disambiguate from a generic 
ame) would read more naturally as SkuDetails.name.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/recoveryPlan/recoveryActions.tsp
```diff
@@ -119,3 +125,90 @@ model RecoveryActionRequest {
...
+
+  @added(Versions.v2026_08_31_preview)
+  @doc("Number of virtual CPUs for the SKU. Absent when SKU specifications are unavailable.")
+  vCpu?: int32;
```
umpatel: Cpu?: int32 is a count, not an identifier - per its own doc (Number of virtual CPUs for the SKU), the value is a quantity. As named it reads like a version/id field. Azure precedent for this exact concept (Compute VirtualMachineSize) uses 
umberOfCores; suggest CpuCount (or 
umberOfVCpus) to make the cardinality explicit in the name itself.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/recoveryPlan/recoveryActions.tsp
```diff
@@ -119,3 +125,90 @@ model RecoveryActionRequest {
...
+
+  @added(Versions.v2026_08_31_preview)
+  @doc("Memory in GiB for the SKU. Absent when SKU specifications are unavailable.")
+  ram?: int32;
```
umpatel: Same pattern here: 
am?: int32 doesn't self-describe its unit, and the doc only clarifies it as GiB. Azure precedent (Compute VirtualMachineSize.memoryInMB) embeds the unit directly in the property name so callers don't have to consult docs to interpret the raw number correctly, and it avoids ambiguity if the unit ever needs to change across API versions. Suggest memoryInGiB instead of 
am.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Agreed - you're right, and my earlier reply didn't actually address your point. The dual-usage argument explains why `*Result` is wrong, but it doesn't rescue `DrillRunReport`, which misleads at both use sites for exactly the reason you give: the type models report-generation status and format availability, never report content.

Renamed to **`DrillReportSummary`** in `7fd6f196`. I picked `Summary` over `Status` to avoid collision with the existing `DrillReportGenerationStatus` enum and `ReportStageStatus` model, and because the type carries more than status (`availableFormats`, `schemaVersion`, `lastGeneratedTimestamp`).

Also tightened the docs so the name and the contract agree:
- `DrillRunProperties.report` -> "Summary of report generation for this Drill Run."
- model doc -> "Public, read-only summary of report generation for a Drill Run. Exposes status and pointers only - the report content and internal storage locations are never surfaced."

Recompiled; no other references to the old name remain.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
arpja: Good call-out on `@extension` — I hadn't considered decoupling the marker from the decorator, so I went and tested it. Unfortunately the toolchain blocks that path explicitly.

Adding it to `downloadUrl`:

```tsp
@TypeSpec.OpenAPI.extension("x-ms-secret", true)
downloadUrl?: url;
```

fails to compile:

```
error @azure-tools/typespec-azure-core/no-openapi-client-extensions:
Do not use the @typespec/openapi @extension decorator to emit the client-altering
"x-ms-secret" extension. It only affects the OpenAPI output, so client SDKs and other
emitters will produce an incorrect representation of the API.
Use the corresponding TypeSpec construct instead.
```

This isn't incidental — `x-ms-secret` is a hardcoded member of the `clientAlteringExtensions` set in `@azure-tools/typespec-azure-core/dist/src/rules/no-openapi-client-extensions.js` (line 26), alongside `x-ms-enum`, `x-ms-parameter-grouping`, etc. The rule fires on any `TypeSpec.OpenAPI.extension` whose first argument is one of those keys. Its whole purpose is to stop specs from emitting these markers via `@extension` and to force the TypeSpec construct — i.e. `@secret` — precisely because `@extension` reaches the OpenAPI document without informing the other emitters.

(Bare `@extension` doesn't resolve here at all, since `@typespec/openapi` isn't imported in this file; the fully-qualified form above is what produces the diagnostic.)

So the two options collapse back to a single either/or:

- **`@secret`** — emits `x-ms-secret: true` *and* overwrites `format: "uri"` with `format: "password"`, which reverts the URI-properties fix requested earlier in this same review.
- **Nothing** — keeps `format: "uri"`.

There's no conflict-free middle option. Worth adding that the rule's severity is `warning`, but this repo compiles with `--warn-as-error`, so taking the `@extension` route would also require a `#suppress` plus the `Approved-TypeSpecSuppression` gate — and this PR currently carries **zero** suppressions, which I'd rather preserve.

On the substantive point: your static-declaration argument is a fair one, and I agree a spec-level marker can have value ahead of consumers existing. But given the choice is now strictly "lose a confirmed, emitter-honoured correctness signal (`format: uri`) in exchange for a marker that no SDK generator currently reads", I'd keep `format: "uri"`. That also matches the Billing and CostManagement precedent, where full SAS URLs are returned in a single `downloadUrl` with no `x-ms-secret`. Exposure stays bounded by the short TTL, `expiryTimestamp`, and per-tenant authorization on the mint path.

Happy to revisit if ARM ships a way to express both, or if you'd like me to raise it upstream against the typespec-azure rule — but I don't think it should hold this PR.

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/models/drill/drillRun.models.tsp
```diff
@@ -77,6 +77,86 @@ model DrillRunProperties extends JobProperties {
...
+@added(Versions.v2026_08_31_preview)
+@doc("Response containing a short-lived, read-only download URL for a Drill Run report.")
+model ReportDownloadUrlResponse {
+  @doc("Format of the report the download URL refers to.")
+  @visibility(Lifecycle.Read)
+  format?: ReportFormat;
+
+  @doc("Short-lived, read-only URL to download the report.")
+  @visibility(Lifecycle.Read)
+  downloadUrl?: string;
```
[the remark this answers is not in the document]
arpja: Addressed in `c3a13829` — `downloadUrl` is now typed with the TypeSpec `url` scalar, which emits `format: "uri"`. Resolving; this thread is pinned to a superseded commit. (The model has since been renamed to `ListReportDownloadUrlResponse`.)

specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
```diff
@@ -169,6 +169,35 @@ interface DrillRuns {
       ArmLroLocationHeader &
       Azure.Core.Foundations.RetryAfterHeader
   >;
+
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-post-operation-response-codes" "Per Azure REST API guidelines and the Spectral PostResponseCodes rule, LRO POST operations with no final body must return 202 + 204 + default. This TSP ARM library rule conflicts with that requirement; we follow the official guideline."
```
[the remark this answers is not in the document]
arpja: No longer applicable — the `arm-post-operation-response-codes` suppression was removed entirely in `7616309d`. `generateReport` now returns `202` + `200`-with-schema + `default`, the one shape both the TSP ARM rule and Swagger LintDiff accept, so no suppression is needed. This PR currently carries **zero** suppressions. Resolving; thread is pinned to a superseded commit.

umpatel (approved):

PR #45668, opened 2026-08-20
Add experimentName to CommandJob

swarupecenits:
## Summary

Models `experimentName` on `CommandJob` as a typed top-level property, and corrects the examples that misrepresented where it lives on the wire.

Two changes:

1. **`experimentName` added to `CommandJob`.** The field already exists in the live service. it is accepted on create and returned on read, but was never modeled, so it did not surface in the generated clients. Callers could only reach it through the untyped dictionary passthrough.

2. **Examples corrected.** Every hand-written jobs example placed `experimentName` inside the `properties` user-property bag. The service does not return it there; it is a sibling of `properties` at the top level. The examples were describing a shape the service never produces.

### `experimentName`

Placement confirmed with the service team: top level is the recommended and correct position. Verified against live GET responses, which return `experimentName` alongside `properties`, not inside it.

Modeled as a plain optional `string`:

- **No `@visibility`** - unlike the `resources` fields, this is genuinely round-trip. The service accepts it on create and returns it on read, so restricting either direction would be wrong.
- **No OpenAPI `default`**, even though the service applies `"Default"` when the field is omitted. Same rationale as `priority` in #45545: emitting a default would make generated clients materialize and send `"Default"` on every request, erasing the distinction between "caller omitted" and "caller explicitly chose Default", and pinning the value if the service default ever changes. Documented in the property description instead.

### Example corrections

| Example | Before | After |
|---|---|---|
| `Jobs_CreateOrUpdate` (request + 200 + 201) | nested in `properties` | top level |
| `Jobs_Get` (response) | nested in `properties` | top level |
| `Jobs_List` (response) | nested in `properties` | top level |

The `properties` bag was left in place with a genuine user-defined property so it still demonstrates its actual purpose rather than appearing empty.

`Jobs_List` also used `"properties": "experimentName=my-experiment"` as its query filter. Since `experimentName` is not a `properties` entry, filtering on it through that parameter would not work - changed to a real user property.

## Files changed

- `src/jobs/models.tsp` - `experimentName` property on `CommandJob`
- `examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json` - moved to top level in the request and both responses
- `examples/v1/Jobs_Get_MaximumSet_Gen.json` - moved to top level
- `examples/v1/Jobs_List_MaximumSet_Gen.json` - moved to top level, query filter corrected
- `openapi3/**` - regenerated

No versioning decorators are needed: the jobs surface is gated behind the required `Foundry-Features: Jobs=V1Preview` header and this contract is not yet published.

## Validation

- `npx tsp compile . --emit "@typespec/openapi3" --warn-as-error` - clean, zero linter warnings
- All modified example files re-parse as valid JSON
- Asserted zero remaining occurrences of `experimentName` nested inside a `properties` bag
- `prettier --check` - already conformant, no reformatting needed
- Verified in the emitted schema: `experimentName` present on `CommandJob`, typed `string`, absent from `required[]`, no `default` emitted
- Confirmed the emitted diff is limited to this one property across the four generated files - no incidental drift

Related: work item [5540988](https://msdata.visualstudio.com/Vienna/_workitems/edit/5540988/?view=edit)

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json
+++ b/specification/ai-foundry/data-plane/Foundry/examples/v1/Jobs_CreateOrUpdate_MaximumSet_Gen.json
@@ -92,13 +92,7 @@
             "framework": "pytorch"
           },
           "experimentName": "my-experiment",
-<<<<<<< HEAD
           "properties": {},
-=======
-          "properties": {
-            "team": "ml-platform"
-          },
->>>>>>> 687ab77bed593311b2ca9f8c5f4729b197e6bd1a
           "codeId": "azureai:my-training-code:1",
           "computeId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/my-rg/providers/Microsoft.MachineLearningServices/workspaces/my-ws/computes/my-cluster",
           "inputs": {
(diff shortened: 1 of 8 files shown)


saanikaguptamicrosoft (approved):

babu-namburi (approved):
Reviewed how the experiment name is set. LGTM

PR #45670, opened 2026-08-20
Add telemetryConfig to SandboxGroup

bhanuprakash-1:

## Purpose of this PR

- [ ] New resource provider.
- [x] Other, please clarify:
  - This is a minor additive update to the unreleased `2026-07-01` API on its release branch. It adds `telemetryConfig` as a read/write SandboxGroup property, available on both create-or-update (PUT) and update (PATCH) requests, so customers can configure where telemetry for sandboxes in a sandbox group is sent.

## Due diligence checklist

- [x] I have reviewed the [Resource Provider guidelines](https://aka.ms/rpguidelines), including the [ARM resource provider contract](https://aka.ms/azurerpc) and [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md).
- [ ] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created.

## What this adds

`SandboxGroupProperties.telemetryConfig` is optional and read/write. It is also picked into
`SandboxGroupPatchProperties`, so PUT and PATCH share a single type and description.

`TelemetryConfig` carries `endpoints` (at least one) and an optional `metricsIntervalSeconds`
(minimum 2, defaults to 15 when omitted).

`TelemetryEndpoint` is polymorphic on a `kind` discriminator, with three endpoint kinds:

| `kind` | Destination | Authentication |
|---|---|---|
| `OTLP` | An endpoint that accepts the OpenTelemetry Protocol. | Optional secret reference sent as an HTTP header. |
| `ApplicationInsights` | An Application Insights resource. | Secret reference holding the connection string. |
| `LogAnalytics` | An Azure Monitor Log Analytics workspace, written through the Logs Ingestion API. | User-assigned or system-assigned managed identity. |

Every endpoint declares the telemetry categories it receives (`data`), and optionally a custom log
output schema (`columns`) and `dynamicJsonColumns`.

Supporting types: `TelemetryData`, `TelemetryProtocol`, `TelemetryHeaderAuth`,
`TelemetryApplicationInsightsAuth`, `TelemetryAuth` (`ManagedIdentity` /
`SystemAssignedManagedIdentity`), `LogColumn` (`Value` / `Ref`), `LogColumnRef`, and
`DynamicJsonColumnsMode`.

## Relationship to the Azure Dev Compute data-plane specification

This contract was derived from the ADC data-plane telemetry specification so that customers see one
coherent feature across both planes. The shapes below intentionally differ, because the ARM linter
applies rules that do not run against data-plane specifications, and because ARM has its own
modelling conventions. The service maps between the two representations.

| Area | Data-plane shape | ARM shape in this PR | Reason |
|---|---|---|---|
| `columns` | Map keyed by output column name | Array of objects carrying a `name` property | ARM does not allow open maps outside `tags`, `delegatedResources` and `userAssignedIdentities` (`RPC-Policy-V1-05`, `RPC-Put-V1-23`) |
| Secret key property | `secretKey` | `secretKeyName` | The ARM linter requires any property whose name ends in a sensitive keyword to be annotated `x-ms-secret`, which would make it write-only. This value is a key *name*, not a secret, and must be returned on read, so the property is named to reflect that |
| `dynamicJsonColumns` | Boolean | Extensible enum `Enabled` / `Disabled` | ARM prefers extensible enums over booleans so further modes can be added without a breaking change |
| `endpoints` | Required | Optional, `minItems: 1` | Properties reachable from a PATCH request body must not be required. The minimum item count keeps the "at least one endpoint" requirement enforceable |
| Polymorphism | Discriminated union | `@discriminator("kind")` base model with derived models | Emits `discriminator` + `allOf` + `x-ms-discriminator-value`, which ARM tooling and the SDK generators consume |
| Generated type names | `LogColumnDefValueLogColumnDef`, `TelemetryEndpointOtlpTelemetryEndpoint`, and similar | `ValueLogColumn`, `OtlpTelemetryEndpoint`, and similar | Readable names in the generated SDKs |

## Intentional exclusions

- The `LogAnalyticsLegacy` endpoint kind is deliberately **not** included. Because it was the only
  consumer of the polymorphic telemetry secret reference, the `AppSecret` / `secretRef` variant and
  the `workspaceId` property are likewise absent from this contract.
- `publicNetworkAccess` remains out of scope for this specification work.

## Additional information

- **No suppressions are introduced by this PR.** Every linter finding was resolved by changing the
  contract rather than silencing the rule.
- No secret material is placed in the contract. `secretId` names a sandbox group secret and
  `secretKeyName` names a key within that secret; both are returned in read responses.
- `TelemetryManagedIdentityAuth.identity` is typed as an ARM resource identifier restricted to
  `Microsoft.ManagedIdentity/userAssignedIdentities`.
- `endpoints` and `data` both carry `minItems: 1`, matching the service behaviour, which rejects an
  empty list in either position.
- Descriptions are written for external Azure customers and describe the contract only. The
  sandbox-group-level configuration is described neutrally as the telemetry configuration for
  sandboxes in the group; sandbox-level precedence is not asserted.
- The stable OpenAPI document and the linked `SandboxGroups_Update` example are regenerated from
  TypeSpec. The example demonstrates configuring telemetry through PATCH.

### DIFF
--- a/specification/app/resource-manager/Microsoft.App/DevCompute/examples/2026-07-01/SandboxGroups_Update.json
+++ b/specification/app/resource-manager/Microsoft.App/DevCompute/examples/2026-07-01/SandboxGroups_Update.json
@@ -14,15 +14,29 @@
           "endpoints": [
             {
               "kind": "LogAnalytics",
-              "dceEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
-              "dcrImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
+              "dataCollectionEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
+              "dataCollectionRuleImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
               "tableName": "Custom-SandboxLogs_CL",
-              "auth": {
-                "kind": "SystemAssignedManagedIdentity"
+              "authentication": {
+                "type": "UserAssigned",
+                "identityResourceId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/exampleidentity"
               },
               "data": [
                 "ContainerStdoutStderr",
                 "Metrics"
+              ]
+            },
+            {
+              "kind": "OTLP",
+              "endpoint": "https://example-collector.contoso.com/otlp",
+              "protocol": "Grpc",
+              "authentication": {
+                "headerName": "Authorization",
+                "secretId": "telemetry-secrets",
+                "secretKeyName": "collectorToken"
+              },
+              "data": [
+                "ContainerOtel"
               ]
             }
           ],
@@ -48,15 +62,29 @@
             "endpoints": [
               {
                 "kind": "LogAnalytics",
-                "dceEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
-                "dcrImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
+                "dataCollectionEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
+                "dataCollectionRuleImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
                 "tableName": "Custom-SandboxLogs_CL",
-                "auth": {
-                  "kind": "SystemAssignedManagedIdentity"
+                "authentication": {
+                  "type": "UserAssigned",
+                  "identityResourceId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/exampleidentity"
                 },
                 "data": [
                   "ContainerStdoutStderr",
                   "Metrics"
+                ]
+              },
+              {
+                "kind": "OTLP",
+                "endpoint": "https://example-collector.contoso.com/otlp",
+                "protocol": "Grpc",
+                "authentication": {
+                  "headerName": "Authorization",
+                  "secretId": "telemetry-secrets",
+                  "secretKeyName": "collectorToken"
+                },
+                "data": [
+                  "ContainerOtel"
                 ]
               }
             ],
--- a/specification/app/resource-manager/Microsoft.App/DevCompute/examples/2026-07-01/SandboxGroups_Update.json
+++ b/specification/app/resource-manager/Microsoft.App/DevCompute/examples/2026-07-01/SandboxGroups_Update.json
@@ -9,7 +9,25 @@
         "environment": "test"
       },
       "properties": {
-        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv"
+        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv",
+        "telemetryConfig": {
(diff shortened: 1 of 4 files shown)


specification/app/resource-manager/Microsoft.App/DevCompute/stable/2026-07-01/examples/SandboxGroups_Update.json
```diff
@@ -9,7 +9,25 @@
...
+        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv",
+        "telemetryConfig": {
+          "endpoints": [
+            {
+              "kind": "LogAnalytics",
+              "dceEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
+              "dcrImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
+              "tableName": "Custom-SandboxLogs_CL",
+              "auth": {
+                "kind": "SystemAssignedManagedIdentity"
```
BlackRider97: In update let's use different endpoint type example

specification/app/resource-manager/Microsoft.App/DevCompute/stable/2026-07-01/examples/SandboxGroups_Update.json
```diff
@@ -9,7 +9,25 @@
...
-        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv"
+        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv",
+        "telemetryConfig": {
+          "endpoints": [
+            {
+              "kind": "LogAnalytics",
+              "dceEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
+              "dcrImmutableId": "dcr-8c1d2e3f4a5b6c7d8e9f0a1b2c3d4e5f",
+              "tableName": "Custom-SandboxLogs_CL",
+              "auth": {
```
vturecek: let's use this standardized format for auth:

```
{
  "authentication": {
    "type": "UserAssigned",
    "identityResourceId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/example-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/example-mi"
  }
}
```

specification/app/resource-manager/Microsoft.App/DevCompute/examples/2026-07-01/SandboxGroups_Update.json
```diff
@@ -9,7 +9,25 @@
         "environment": "test"
       },
       "properties": {
-        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv"
+        "environmentId": "/subscriptions/34adfa4f-cedf-4dc0-ba29-b6d1a69ab345/resourceGroups/examplerg/providers/Microsoft.App/managedEnvironments/exampleenv",
+        "telemetryConfig": {
+          "endpoints": [
+            {
+              "kind": "LogAnalytics",
+              "dceEndpoint": "https://example-dce-a1b2.eastus-1.ingest.monitor.azure.com",
```
shivamkm07: As per teams discussion, let's update to use `dataCollectionEndpoint` and `dataCollectionRuleImmutableId`

bhanuprakash-1: Done

specification/app/resource-manager/Microsoft.App/DevCompute/stable/2026-07-01/examples/SandboxGroups_Update.json
bhanuprakash-1: Done

shivamkm07 (approved):

bhanuprakash-1:
Not making any changes to the Control plane GA API version (2026-07-01). Its frozen for now and any changes will go to next API version.

 Hence closing this PR

PR #45681, opened 2026-08-20
TypeSpec for new ARM Foundry Cost Control APIs

dargilco:
## Summary

Add the `2026-09-15-preview` Cognitive Services Management API version and the Cost Controls control-plane contract.

Cost controls let an account define reusable cost limits and attach them directly to the owning account or its deployments. Account-level connections publish cost control audit telemetry and alerts. The TypeSpec reflects the current Platform Resource Provider implementation after the service renamed Budget Policies to Cost Controls.

## Service implementation PRs

- [Add Budget Policies ARM contracts](https://msazure.visualstudio.com/Cognitive%20Services/_git/Platform-ResourceProvider/pullrequest/16756995) introduced the original resource and rule contracts, attachment properties, validation limits, serialized names, and API-version gating.
- [CRUD APIs for budget policies](https://msazure.visualstudio.com/Cognitive%20Services/_git/Platform-ResourceProvider/pullrequest/16781932) implemented and tested the account-child CRUD/list endpoints, account and deployment attachments, ETag concurrency, feature gating, and delete-in-use behavior.
- [Rename Budget Policy CRUD APIs to Cost Control](https://msazure.visualstudio.com/Cognitive%20Services/_git/Platform-ResourceProvider/pullrequest/16870346) renamed the resource path, public contracts, attachments, validators, operation metadata, feature gate, and tests from Budget Policy to Cost Control.

## API changes

- Add preview API version `2026-09-15-preview`.
- Add account-child proxy resource `Microsoft.CognitiveServices/accounts/costControls`.
- Add synchronous operations:
  - `CostControls_Get`
  - `CostControls_CreateOrUpdate`
  - `CostControls_Update`
  - `CostControls_Delete`
  - `CostControls_List`
- Add Cost Control resource, patch, rule, counter-dimension, match, threshold, unit, period, connection, and list-result models.
- Add `costControlIds` to account and deployment properties, with up to 20 full Cost Control resource IDs.
- Add nullable account-level `costControlConnections` for Application Insights audit telemetry and Event Grid alerts.
- Model ETag response headers and conditional `If-Match`/`If-None-Match` request headers.
- Support the service-defined counter dimensions `agent`, `identity`, `session`, `project`, `account`, and `custom`; custom dimensions accept an `attribute` of up to 128 characters.
- Model the four supported match attributes, including the service wire name `foundry.project.id`.
- Support the threshold actions `audit`, `alert`, and `block`.
- Generate the `2026-09-15-preview` Swagger under `preview/2026-09-15-preview`.

## Contract decisions

- Cost controls are proxy resources nested under accounts; they have no tags, location, SKU, identity, or provisioning state.
- PUT fully replaces cost control properties. PATCH distinguishes an omitted `displayName` from explicit `null`, allowing callers to clear it.
- Cost control and rule names use the service identifier pattern and 64-character limit.
- This preview supports exactly one counter dimension per rule. The `account` dimension is counter-key only, and `custom` requires the `attribute` property.
- Match rules accept up to four trusted request attributes with up to 20 values per attribute.
- `usd` is the only cost control unit in this preview.
- Recurring rules default `recurring` to `true` and use a calendar-aligned period; the service validates the recurring/period relationship.
- Thresholds may be omitted, `null`, or empty to define an audit-only rule. Explicit `audit` thresholds require a value of zero. `alert` and `block` thresholds require a value greater than zero; only `block` enforces a limit.
- Account-level `costControlConnections` configures Application Insights audit telemetry and Event Grid alerts. Application Insights is optional for attachment, accounting, and enforcement and can be configured later. Event Grid must be configured before attaching a cost control with an `alert` threshold.
- The connection object and its connection IDs distinguish omission from explicit `null`, allowing callers to clear configured connections when service validation permits.
- Attachments are supported on accounts and deployments.
- Operations are synchronous. DELETE returns `200 OK`, matching the current service implementation.
- The specification intentionally remains on `2026-09-15-preview`. The service currently uses an earlier API-version gate, which will be updated separately.

## Validation

- `npx tsp compile .`: passed with no diagnostics.
- `npx tsp format --check CostControl.tsp main.tsp models.tsp`: passed.
- VS Code TypeSpec diagnostics: no errors.
- `git diff --check`: passed.
- Emitted Swagger review: verified the API version, Cost Control paths and operation IDs, path/body/header parameters, requiredness, defaults, constraints, response codes, ETag and conditional headers, match wire names, dimension values, threshold behavior and actions, account/deployment attachment ARM ID types, and account-level connection ARM ID types.
- Confirmed no stale Budget Policy identifiers remain in the TypeSpec or generated `2026-09-15-preview` Swagger.

## Unresolved questions or known failures

- `Swagger LintDiff - Analyze Code` is currently failing and requires investigation.
- `Breaking Change(Cross-Version) - Analyze Code` is currently failing and requires investigation.


fmabroukmsft:
## Blocking

### 1. `costControlIds` is constrained to 20; the service allows 1

`models.tsp` sets `@maxItems(20)` on `costControlIds` in both `AccountProperties` and
`DeploymentProperties`. The emitted swagger carries `"maxItems": 20` for both.

The service caps attachments per scope at 1:

- `CostControlLimits.MaxAttachedCostControlsPerScope = 1` (`CostControl.cs:29`)
- enforced in `CostControlAttachmentValidator.cs:84` — "A scope cannot attach more than 1 cost controls."
- pinned by `CostControlContractTests.cs:465` (`Assert.AreEqual(1, MaxAttachedCostControlsPerScope)`)
- documented on both properties (`AccountProperties.cs:279`, `AccountDeploymentProperties.cs:28`)

20 is the value of the adjacent constant `MaxCostControlsPerAccount` (`CostControl.cs:22`), which
bounds how many cost control definitions an account owns, not how many can be attached to a scope.

Any request carrying 2 or more IDs is accepted by the contract and rejected by the service with a
400. Relaxing `maxItems` from 1 to 20 in a later version is non-breaking; tightening 20 to 1 is
breaking. `@maxItems(1)` on both properties matches the current service limit.

### 2. `CostControl.etag` has no `@@alternateType` in `client.tsp`

`client.tsp` is unchanged in this PR and contains no references to `CostControl`. Every other
etag-bearing resource in this spec maps the property to the standard ARM type for .NET:

```
@@alternateType(Account.etag, eTag, "csharp");                    // :191
@@alternateType(CommitmentPlan.etag, eTag, "csharp");             // :205
@@alternateType(Deployment.etag, eTag, "csharp");                 // :207
@@alternateType(EncryptionScope.etag, eTag, "csharp");            // :259
@@alternateType(Project.etag, eTag, "csharp");                    // :312
@@alternateType(RaiPolicy.etag, eTag, "csharp");                  // :779
@@alternateType(DefenderForAISetting.etag, eTag, "csharp");       // :782
@@alternateType(Workbench.etag, eTag, "csharp");                  // :1141
@@alternateType(ManagedComputeDeployment.etag, eTag, "csharp");   // :1201
```

18 in total. `CostControl` is the only etag-bearing resource without one.

The .NET SDK will expose `CostControl.Etag` as `string` while every sibling resource exposes
`Azure.ETag`. This compiles cleanly, so no CI check reports it. Changing the type after the SDK
ships is a compile-breaking change.

```tsp
@@alternateType(CostControl.etag, eTag, "csharp");
```

`CostControl` also carries no `@@clientName` prefix for C#, while other resources in this spec are
renamed to avoid collisions (`Account` -> `CognitiveServicesAccount`, `Deployment` ->
`CognitiveServicesAccountDeployment`).

## Requires a decision

### 3. `amount` and `threshold.value` emit as `format: double`; the service uses `decimal`

`CostControlRule.Amount` and `CostControlThreshold.Value` are `decimal` in the service contracts
and `float64` in TypeSpec, emitting `"type": "number", "format": "double"`.

`double` for currency is common in ARM specs and the precision impact at typical budget values is
small. Changing the emitted type after the SDKs ship is breaking, so the choice is cheaper to
settle in this version than in a later one.

## Non-blocking

4. `CostControlDimension.attribute` has no `@pattern`. The emitted schema carries only
   `maxLength: 128`. The service additionally enforces `^[a-zA-Z][a-zA-Z0-9_.-]{0,127}$`
   (`CostControlValidator.cs:197-199`). Rule and resource names already carry their patterns.
   Failure mode is a 400 from the service.

5. `@minValueExclusive(0)` on `amount` is absent from the emitted swagger. The `amount` schema has
   no `minimum` and no `exclusiveMinimum`, and `exclusiveMinimum` does not appear anywhere in the
   generated document. `@minValue(0)` on `threshold.value` did emit `"minimum": 0`, so the
   exclusive variant is dropped by the emitter. The service rejects non-positive amounts
   (`CostControlValidator.cs:164`), so behavior is correct and the constraint is unexpressed.

6. The remaining Breaking Change (Cross-Version) error is `ChangedParameterOrder` on `raiPolicies`
   PUT (`raiPolicy` body parameter moved from index 3 to index 5). It originates in the base
   branch and is unrelated to cost controls. The description still lists `Swagger LintDiff` as
   failing; it currently passes.

## Verified, no action

- The 217 removed lines in `cognitiveservices.json` are the alphabetical relocation caused by
  inserting `costControls`. Added and removed counts balance exactly.
- `Microsoft.MachineLearningServices/workspaces/connections` is correct for the connection IDs.
  The examples confirm the `@AML` workspace projection.
- `@visibility(Lifecycle.Read, Lifecycle.Update)` on `costControlConnections` matches
  `AccountsHandler.cs:624`, which rejects the property at account creation.
- The `no-nullable` suppressions are accurate. `AccountsHandler.cs:1712-1861` inspects the raw
  `JToken` for `JTokenType.Null`, and `CostControlPatchProperties` tracks `IsDisplayNameSpecified`,
  so omitted and explicit null are distinguishable.
- `x-ms-client-name` is emitted for all four dotted match keys, so the
  `DefinitionsPropertiesNamesCamelCase` suppression reason holds.
- `If-None-Match` on PUT is handled via `GenericNestedResourceTypeHandler:191`.
- No path in this spec declares 412, so omitting it is consistent.
- Dimension types, threshold types and actions, match attribute names, and the limits 50 / 4 / 20 /
  64 / 128 match the service. `foundry.account.id` is correctly counter-key-only and absent from
  `CostControlMatch`.
- The default tag remains `package-2026-09-01`, the most recent stable version.

Platform-ResourceProvider line numbers are from commit `47da6c1743`.

fmabroukmsft:
Re-reviewed at `71379b82b`.

## Addressed

- `costControlIds` is now `@maxItems(1)` on both `AccountProperties` and `DeploymentProperties`,
  and the regenerated swagger carries `"maxItems": 1`.
- `client.tsp` now has `@@alternateType(CostControl.etag, eTag, "csharp")` and
  `@@clientName(CostControl, "CognitiveServicesAccountCostControl", "csharp")`.
- `CostControlDimension.attribute` now carries `@pattern("^[a-zA-Z][a-zA-Z0-9_.-]{0,127}$")`,
  emitted to the swagger. The existing `x-tenant-id` examples still satisfy it.

## New: TypeSpec Validation fails on this commit

`TypeSpec Validation` passed on `b6c6a92e8` and fails on `71379b82b`. The cause is a missing
trailing newline on `client.tsp`:

```
Rule Format failed
- Formatting
✔ 1 formatted, 144 unchanged

@@ -1237,4 +1237,4 @@ op regenerateKeyCustom(
 @@clientName(CostControl, "CognitiveServicesAccountCostControl", "csharp");
-@@alternateType(CostControl.etag, eTag, "csharp");
\ No newline at end of file
+@@alternateType(CostControl.etag, eTag, "csharp");

Files have been changed after `tsp format`. Run `tsp format` and ensure all files are included in your change.
```

`client.tsp` is the only modified file, but the job reports a failure for seven cognitiveservices
projects (`AnomalyDetector`, `CognitiveServices.Management`, `ContentSafety`, `OpenAI.Inference`,
`Speech.Transcription`, `Speech.VideoTranslation`), since all of them format the shared folder.

```
npx tsp format "**/*.tsp"
```

## Breaking Change (Cross-Version): the failure cause has changed

The previous run had one error, `ChangedParameterOrder` on `raiPolicies` PUT. This run has that
plus eight `ConstraintChanged` errors on `connectionName`, across the account-level and
project-level connections paths (GET, PUT, PATCH, DELETE on each).

These do not originate in this commit or in the head branch:

| Source | `connectionName` constraint |
| --- | --- |
| PR head `71379b82b` | `pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]{2,32}$`, no minLength/maxLength |
| `main` `stable/2026-07-01` (the comparison target) | `pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]{2,32}$`, no minLength/maxLength |
| Base branch `swagger-agent/CogSvc-2026-09-15-preview-typespec` | `pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]{2,127}$`, minLength 3, maxLength 128 |
| PR merge commit (`refs/pull/45681/merge`, what CI validates) | `pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]{2,127}$`, minLength 3, maxLength 128 |

Head and target are identical, so nothing here changed the constraint. The check runs against the
merge commit, and the head is 18 commits behind its base. The tightening comes from the base
branch, so it belongs to that PR rather than this one.

Also note the runner logged `errorCnt: 0, oadViolationsCnt: 9, process.exitCode: 0` and
"Prevented spurious failure of breaking change check", then exited 1, so part of this may be
check infrastructure rather than spec content.

## Not addressed, previously marked non-blocking

- `amount` and `threshold.value` still emit `format: double` against `decimal` in the service contracts.
- `@minValueExclusive(0)` on `amount` is still absent from the emitted swagger.

Neither blocks merge. Flagging only because both are cheaper to change before the SDKs ship.

fmabroukmsft (approved):

PR #45682, opened 2026-08-20
Add RAI policy control plane contracts to 2026-09-15-preview

lycroy-work:

## Purpose of this PR

- [ ] New resource provider.
- [x] Update an existing private preview API version for a new feature.
- [ ] Update an existing version to fix OpenAPI specification quality issues.
- [ ] Convert an existing OpenAPI specification to TypeSpec.
- [ ] Other.

## Due diligence checklist

- [x] This PR modifies Azure Resource Manager specifications, not data-plane specifications.
- [x] The changes were reviewed against the ARM resource provider contract and Azure REST API guidelines.
- [x] Release planning for `2026-09-15-preview` is tracked by the parent API-version PR, #45556.

## Summary

This PR extends the Cognitive Services `2026-09-15-preview` management-plane TypeSpec with account-scoped Responsible AI policy authoring resources.

- Adds `raiRegos` GET, PUT, DELETE, and LIST operations for reusable Rego source.
- Extends `raiPolicies` with the optional `format`, typed `acs` manifest, same-account `acsRegos` references, and optional blocklist and external safety-provider references.
- Adds `raiBindings` GET, PUT, DELETE, and LIST operations for associating an ARM resource ID with a Responsible AI policy.
- Adds ETag response headers and optional `If-Match` / `If-None-Match` request preconditions.
- Adds Rego and binding paging contracts, examples, and generated OpenAPI.

## Public contract details

- Existing ContentFilters policy requests remain supported. When creating a policy, omitting `format` selects the ContentFilters representation; ACS policy requests use `format: "ACS"`.
- `agent_control_specification_version` is a required string in the ACS manifest. Examples use the current `0.4.0-alpha.1` manifest value without defining a public version enum.
- Rego source is a required nonempty string. The optional `encoding` property defaults to `None` and also supports `Base64`.
- Rego and binding resource names use the `2-64` character ARM name contract.
- Rego LIST defaults `$top` to 10; binding LIST defaults `$top` to 50. Both require positive values and return the standard `value` / `nextLink` shape.
- Policy LIST keeps the existing collection route and can return both ContentFilters and ACS policy representations.
- Existing `customBlocklists` SDK model and source enum are preserved.
- The nested ACS document preserves canonical snake_case wire names and name-keyed map shapes. TypeSpec and generated SDK projections retain idiomatic client property names.

## Compatibility

- Only generated `2026-09-15-preview` artifacts change. All earlier generated versions, including `2026-07-15-preview`, remain unchanged.
- Existing `RaiPolicies_*` operation IDs and the `RaiPolicies` SDK client name are preserved.
- The new payload cardinality and operational capacity policies are not encoded as public OpenAPI maxima.
- Scoped Swagger exceptions document the externally defined ACS wire names and map-shaped manifest fields.

## Examples

Adds create, replace, get, list, and delete examples for Rego, ACS policies, and bindings where applicable. Authored TypeSpec examples and emitted OpenAPI example copies are byte-identical.

## Validation

- `npx tsp compile specification/cognitiveservices/CognitiveServices.Management/main.tsp`
- TypeSpec formatting check
- Cognitive Services scoped spell check
- OAV semantic validation
- OAV example validation
- Direct TypeSpec SDK projection validation
- Pre-`2026-09-15-preview` generated-output isolation check
- Authored/emitted example hash comparison
- All seven authored ACS request/response manifests validated against the Agent Control Specification `liamcrumm/release-0.4.0-alpha.3` JSON Schema

## Dependency

This PR is stacked on #45556, which introduces the Cognitive Services `2026-09-15-preview` API version.


rkuthala (approved):

fmabroukmsft:
`RaiBindingProperties.targetPolicyName` (models.tsp) has no length bounds. The pattern `^[a-zA-Z0-9][a-zA-Z0-9_.-]*$` accepts a single character and has no upper bound, but the field references a `RaiPolicy` name, which is constrained to 2–64 characters. The sibling `RaiRegoReference.regoName` bounds this to 2–64, so this appears to be an inconsistency.

Suggested change:

```typespec
@pattern("^[a-zA-Z0-9][a-zA-Z0-9_.-]*$")
@minLength(2)
@maxLength(64)
targetPolicyName: string;
```

lycroy-work:
> `RaiBindingProperties.targetPolicyName` (models.tsp) has no length bounds. The pattern `^[a-zA-Z0-9][a-zA-Z0-9_.-]*$` accepts a single character and has no upper bound, but the field references a `RaiPolicy` name, which is constrained to 2–64 characters. The sibling `RaiRegoReference.regoName` bounds this to 2–64, so this appears to be an inconsistency.
> 
> Suggested change:
> 
> ```tsp
> @pattern("^[a-zA-Z0-9][a-zA-Z0-9_.-]*$")
> @minLength(2)
> @maxLength(64)
> targetPolicyName: string;
> ```

Thanks for flagging this and I intentionally did this. Let me explain to you:

RaiPolicy names have never had a 2–64 length constraint: both old versions and this new API version emit only ^[a-zA-Z0-9][a-zA-Z0-9_.-]*$. We intentionally did not add bounds to RaiPolicy because the route is shared by existing ContentFilters policies and new ACS-format policies. Adding a maximum now would be a cross-version breaking change: an existing legacy policy whose name exceeds 64 characters would no longer conform when a customer uses the new API version to GET, PUT, or DELETE it.

For the same reason, I propose leaving RaiBindingProperties.targetPolicyName aligned with the referenced RaiPolicy name contract. Constraining only the binding property would allow creation of an ACS policy that could not be referenced by a schema-conforming binding. RaiRegoReference.regoName is different because both the Rego resource and its reference are newly introduced and constrained together. If we decide to impose 2–64 for newly created ACS policies, we should first add ownership-aware RP validation that grandfathers existing legacy and PMS-backed policies, then align the binding contract separately in the runtime path.

fmabroukmsft (approved):

PR #45686, opened 2026-08-20
TypeSpec Suppressions: link to the source line in the PR from the rendered comment

catalinaperalta:
Example PR with the link: https://github.com/catalinaperalta/azure-rest-api-specs/pull/11


iscai-msft (approved):
looks good in the demo! i'm wondering how getting rid of the sha still has it work across diff commits in the same pr, but more as a question

catalinaperalta:
> looks good in the demo! i'm wondering how getting rid of the sha still has it work across diff commits in the same pr, but more as a question

Well before as you noticed it's the line in the specific commit, now the source is the PR. With this update we calculate the hash from the file path in the PR and show the line in the PR diff instead.

PR #45687, opened 2026-08-20
Add Voice Live VAD profiles to Foundry Voice Agents

shuzhancnjx:
## Summary

- expose the Voice Live Azure semantic VAD `profile` contract in the Foundry Voice Agents TypeSpec
- support `classic`, `eager`, `normal`, and `patient`, with omitted values defaulting to `classic`
- preserve explicit `profile: null` to disable profile expansion without disabling turn detection
- document the profile-bearing `session.update` path and related VAD lifecycle events in `src/voice-agents/protocol.tsp`
- regenerate the Foundry v1 and virtual-public-preview OpenAPI documents

## Scope

The profile is inherited only by the Voice Live-managed Azure semantic VAD variants:

- `azure_semantic_vad`
- `azure_semantic_vad_en`
- `azure_semantic_vad_multilingual`

`protocol.tsp` explicitly documents `audio.input.turn_detection.profile` on `session.update` and the existing `input_audio_buffer.speech_started`, `input_audio_buffer.speech_stopped`, and `input_audio_buffer.timeout_triggered` server events. Event payloads remain unchanged. Assistant-speaking and TTS-conditioned VAD tuning remains private and is not included in the SDK contract.

## Validation

- `npx tsp format -c specification/ai-foundry/data-plane/Foundry/src/voice-agents/{agents,protocol}.tsp`
- `npx tsp compile specification/ai-foundry/data-plane/Foundry`

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
@@ -558,9 +558,33 @@ model VoiceEndOfUtteranceDetection {
   timeout_ms?: duration;
 }
 
+@doc("A bundled Voice Live profile for Azure semantic voice activity detection.")
+@extension("x-ms-foundry-meta", VoiceAgentRequiredPreview)
+union VoiceAzureSemanticVadProfile {
+  string,
+
+  @doc("Disables profile expansion and uses explicit turn-detection settings.")
+  none: null,
+
+  @doc("Preserves the current default turn-detection behavior.")
+  classic: "classic",
+
+  @doc("Prioritizes fast turn detection.")
+  eager: "eager",
+
+  @doc("Balances turn-detection latency and tolerance for pauses.")
+  normal: "normal",
+
+  @doc("Prioritizes tolerance for longer pauses.")
+  patient: "patient",
+}
+
 /** Options shared by Azure semantic voice activity detection configurations. */
 #suppress "@azure-tools/typespec-azure-core/known-encoding" "The existing voice-agent wire contract requires integer-millisecond durations."
 alias VoiceAzureSemanticVadOptions = {
+  @doc("The bundled VAD profile. Defaults to classic. Set to null to disable profile expansion.")
+  profile?: VoiceAzureSemanticVadProfile = VoiceAzureSemanticVadProfile.classic;
+
   @doc("Activation threshold for voice activity detection, from 0 to 1.")
   @minValue(0)
   @maxValue(1)
--- a/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
+++ b/specification/ai-foundry/data-plane/Foundry/openapi3/v1/microsoft-foundry-openapi3.json
@@ -90123,6 +90123,11 @@
               "azure_semantic_vad_en"
             ]
           },
+          "profile": {
+            "$ref": "#/components/schemas/VoiceAzureSemanticVadProfile",
+            "description": "The bundled VAD profile. Defaults to classic. Set to null to disable profile expansion.",
+            "default": "classic"
+          },
           "threshold": {
             "type": "number",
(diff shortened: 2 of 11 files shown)


shuzhancnjx:
Updated in 48cde8d20c: the VAD profile contract now lives in `specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp`, following the Foundry Voice Agents format referenced in the review. The standalone VoiceLive changes were removed. Existing VAD protocol events remain unchanged, and assistant-speaking/TTS-conditioned tuning is not exposed.

shuzhancnjx:
Updated in 093af4291b to address the protocol-file requirement directly. `src/voice-agents/protocol.tsp` now documents the `session.update` VAD profile path and the existing speech-started, speech-stopped, and timeout events. Event payloads are unchanged, and no assistant-speaking/TTS-private parameters are exposed.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -558,9 +558,33 @@ model VoiceEndOfUtteranceDetection {
...
+@doc("A bundled Voice Live profile for Azure semantic voice activity detection.")
+@extension("x-ms-foundry-meta", VoiceAgentRequiredPreview)
+union VoiceAzureSemanticVadProfile {
+  string,
+
+  @doc("Disables profile expansion and uses explicit turn-detection settings.")
+  none: null,
+
+  @doc("Preserves the current default turn-detection behavior.")
+  classic: "classic",
```
yulin-li, line 570: for v1, maybe we can remove classic or at lest set default to normal

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
```diff
@@ -558,9 +558,33 @@ model VoiceEndOfUtteranceDetection {
   timeout_ms?: duration;
 }
 
+@doc("A bundled Voice Live profile for Azure semantic voice activity detection.")
+@extension("x-ms-foundry-meta", VoiceAgentRequiredPreview)
+union VoiceAzureSemanticVadProfile {
+  string,
+
+  @doc("Disables profile expansion and uses explicit turn-detection settings.")
+  none: null,
```
yulin-li, line 567: customer could still override it so I think we don't need a "none"?

shuzhancnjx: `none` is only the TypeSpec variant name for the wire value `null`; the generated OpenAPI does not expose a `"none"` string. Per-field values override individual profile defaults, but they do not provide a way to disable expansion for every omitted profile-owned field. Explicit `profile: null` is the bundle-level opt-out while keeping turn detection enabled and field-by-field configuration available.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
shuzhancnjx: Keeping `classic` and the `classic` default is intentional for v1 compatibility. Omitting `profile` must preserve today's turn-detection behavior; changing the default to `normal` would silently apply the latency-tuned normal bundle to existing clients. `normal` remains an explicit opt-in profile.

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
yulin-li: then we should use `? VoiceAzureSemanticVadProfile | null` instead of putting null into the enum?

specification/ai-foundry/data-plane/Foundry/src/voice-agents/agents.tsp
yulin-li: this should be a part of migration as this is a new service/api version

PR #45700, opened 2026-08-21
Add Maintenance 2025-11-01 stable API version

dvarshney29:

> [!TIP]

## Summary

Adds the new stable Microsoft Maintenance API version `2025-11-01`.

The new stable version carries forward the complete `2025-10-01-preview` API
surface while introducing the validated route-casing and compatibility
corrections in a new version. The already-published `2023-10-01-preview` and
`2025-10-01-preview` OpenAPI files are not modified.

### Background

This PR originally applied the corrections directly to the two published
preview versions. Based on API review feedback that published API versions are
immutable, the change now introduces `2025-11-01` instead.

The underlying compatibility work was previously validated in:

- [Azure/azure-rest-api-specs-pr#29926](https://github.com/Azure/azure-rest-api-specs-pr/pull/29926)
- [Azure/azure-rest-api-specs-pr#30088](https://github.com/Azure/azure-rest-api-specs-pr/pull/30088)

### Changes

- Adds stable API version `2025-11-01`.
- Carries forward every API, schema, property, and example from
  `2025-10-01-preview`.
- Uses lowercase `resourcegroups` for the eight affected resource-group-scoped
  route shapes in the new stable version.
- Preserves the published camel-case route for the legacy maintenance
  configuration resource-group list operation.
- Splits configuration-assignment read and write parameter contracts so GET
  retains `^.+$`, while PUT, PATCH, and DELETE omit that write constraint.
- Adds `@armResourceList(MaintenanceConfiguration)` for the explicit legacy
  list operation required by the current TypeSpec ARM library.
- Retains scoped validation suppressions and explanatory compatibility
  comments.
- Keeps the published `2023-10-01-preview` and `2025-10-01-preview`
  `Maintenance.json` files byte-for-byte identical to `main`.

### Validation

- TypeSpec validation succeeds.
- `tsp compile . --warn-as-error` succeeds.
- The generated `stable/2025-11-01/Maintenance.json` is structurally identical
  to the corrected `2025-10-01-preview` contract after normalizing the API
  version.
- All 40 source examples are identical after API-version normalization.
- The generated stable examples are emitted successfully.
- Both published preview OpenAPI files match `main` exactly.

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


dvarshney29:
## `2025-11-01` stable-version compatibility justification

This PR no longer modifies the published `2023-10-01-preview` or
`2025-10-01-preview` specifications. It introduces the corrections in the new
stable `2025-11-01` version.

The generated stable contract intentionally matches the corrected
`2025-10-01-preview` contract after API-version normalization. The expected
cross-version findings against stable `2023-04-01` are inherited differences
already investigated for the corresponding private-repository change:

- Type-format metadata introduced by TypeSpec/common-types v5.
- Read-only metadata on the Operations API response model.
- The common-types Operations model's legacy property and enum metadata
  differences.
- The intentional resource-group route casing change from `resourceGroups` to
  `resourcegroups`.

The route checker represents each casing change as an old path removal and a
new path addition because paths are compared case-sensitively. No API operation
is being removed.

The resource-group maintenance-configuration collection route remains
camel-case because it is a published legacy exception. The eight corrected
route shapes use lowercase `resourcegroups`, matching the service's new-route
convention.

The configuration-assignment write operations also omit the previous
`configurationAssignmentName` regex constraint. This is a relaxation, while GET
retains the existing constraint.

Validation confirms:

- The two published preview OpenAPI files match `main` byte-for-byte.
- The new stable Swagger matches the corrected preview after API-version
  normalization.
- All 40 examples match after API-version normalization.
- TypeSpec validation and compilation succeed.

We request versioning and breaking-change approval for introducing these
previously reviewed compatibility corrections in the new stable version.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/routes.tsp
```diff
@@ -177,7 +177,7 @@ interface ApplyUpdateForResourceGroupOperationGroup {
   #suppress "@azure-tools/typespec-azure-resource-manager/use-api-version" "FIXME: review suppression and update justification"
   #suppress "@azure-tools/typespec-azure-resource-manager/use-interface" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
   @summary("Get Configuration records within a subscription and resource group")
-  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Maintenance/applyUpdates")
+  @route("/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/Microsoft.Maintenance/applyUpdates")
```
rajabadmanaban, line 179: **[BLOCKING] This line is a regression, not a casing correction, and it changes a route that has been camelCase in GA-stable since 2021-05-01.**

I pulled every published version of `Maintenance.json` and diffed the casing of this exact route shape (`ApplyUpdateForResourceGroup_List`):

| API version | Casing of `.../{resourceGroupName}/providers/Microsoft.Maintenance/applyUpdates` |
|---|---|
| 2020-04-01 (stable) | absent |
| **2021-05-01 (stable GA)** | **`resourceGroups` (camelCase)** |
| 2022-07-01-preview | `resourceGroups` |
| 2022-11-01-preview | `resourceGroups` |
| 2023-09-01-preview | `resourceGroups` |
| 2023-10-01-preview (published) | `resourceGroups` |
| 2025-10-01-preview (published) | `resourceGroups` |
| **This PR** | **`resourcegroups` (lowercase)** |

This route has never once been lowercase in any shipped version. The PR description and the justification comment both characterise all eight route changes as "the same textual route-casing correction" restoring legacy casing. That statement is accurate for seven of the eight; it is factually wrong for this one. Compare with the other three affected shapes, which genuinely were regressed by the TypeSpec conversion and are correctly restored here:

- `maintenanceConfigurations/{resourceName}`: lowercase in 2020-04-01, 2021-05-01, 2023-04-01, 2022-07, 2022-11, 2023-09 -> camelCase only in the two TypeSpec-generated previews. Restoring to lowercase is correct.
- `configurationAssignments/{configurationAssignmentName}` (RG): lowercase in 2023-04-01 GA and 2023-09-01-preview. Restoring to lowercase is correct.
- `maintenanceConfigurations` (RG list): camelCase everywhere including GA, and this PR correctly goes to significant lengths to preserve it.

The inconsistency is the problem: `MaintenanceConfiguration.tsp` gets a brand-new interface plus two suppressions purely to preserve its legacy camelCase, while this route, which has exactly the same legacy camelCase and a stronger claim to it (it is in a GA-stable version), is silently flipped by the global segment override. The PR applies its own stated rule to one route and not the other.

Required: revert this to `resourceGroups` and preserve it with an explicit `@route` exactly as `MaintenanceConfigurationsForResourceGroupRoutes` does. That also drops the `RemovedPath`/`AddedPath` finding count from 8 to 7 per version.

If the intent genuinely is to normalise this route to lowercase, that is an API change and belongs in a new API version with an explicit ARM breaking-change exception, not folded into a PR whose stated purpose is a TypeSpec 1.15 `use-operation-decorator` build fix.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/back-compatible.tsp
```diff
@@ -143,6 +143,13 @@ using Microsoft.Maintenance;
 @@clientLocation(UpdatesOperationGroup.listParent, "Updates");
 @@clientLocation(UpdatesOperationGroup.list, "Updates");
 
+// All current and future Maintenance resource-group-scoped routes intentionally use lowercase
+// `resourcegroups`. Operations that must preserve legacy camelCase declare an explicit route.
+@@TypeSpec.Rest.segment(
```
rajabadmanaban, line 150: **[BLOCKING] This global override is the mechanism behind the `applyUpdates` regression, and it is unguarded against recurrence.**

`@@TypeSpec.Rest.segment` on `Azure.ResourceManager.ResourceGroupParameter.resourceGroupName` rewrites the segment for every resource-group-scoped route in the Maintenance namespace at once, across every API version compiled from this TypeSpec, in a single statement. Two consequences:

1. It silently rewrote `ApplyUpdateForResourceGroup_List` away from the casing it has carried since GA 2021-05-01 (see my comment on `routes.tsp:180`). Nothing in the diff makes that visible at the call site, which is exactly why it slipped through.
2. The comment above states the intent is that "all current and future" RG-scoped routes use lowercase. That makes lowercase the silent default for every route added from now on. Any future operation that needs to match a legacy camelCase contract will inherit lowercase and ship wrong unless the author happens to remember this override exists in a different file. There is no compile-time signal.

Note also that this is applied unversioned. It is not gated behind `@added`/`@removed` or scoped to a version range, so it retroactively rewrites the already-published `2023-10-01-preview` and `2025-10-01-preview` simultaneously. In the ARM spec world an API version is the rollout control, and this change deliberately edits two shipped versions in place. That is precisely why the PR is carrying `VersioningReviewRequired`. There is no staged-rollout mechanism here, so correctness has to be established up front rather than mitigated after merge.

Two asks:

- State explicitly in the comment which route shapes are intended to be lowercase, and confirm the set matches stable 2023-04-01 exactly. I verified the intended end state is 1 camelCase + the rest lowercase in both previews; that matches GA except for `applyUpdates`.
- Add a regression guard so this cannot silently re-break: a checked-in casing assertion over the generated `paths` keys, or at minimum a comment block in `readme.md` enumerating the camelCase exceptions. A convention enforced only by reviewer memory will not survive the next conversion bump.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/ConfigurationAssignment.tsp
```diff
@@ -263,44 +266,74 @@ interface ConfigurationAssignmentsForSubscriptions {
...
+model ConfigurationAssignmentsForResourceGroupWriteNameParameters {
+  ...Extension.ExtensionProviderNamespace<ConfigurationAssignment>;
+
+  // Stable 2023-04-01 omits the name pattern on PUT, PATCH, and DELETE.
+  // Keep the empty pattern so writes match that contract while GET retains `^.+$`.
+  ...KeysOf<ResourceNameParameter<
+    Resource = ConfigurationAssignment,
+    KeyName = "configurationAssignmentName",
+    SegmentName = "configurationAssignments",
+    NamePattern = ""
```
rajabadmanaban, line 303: **[SHOULD-FIX] The emitted output is correct, but the comment above is factually wrong and the server-side contract is unverified.**

First, the part that is fine, since it is worth recording that it was checked rather than assumed: `NamePattern = ""` emits **no** `pattern` key at all, not `"pattern": ""`. I confirmed this by diffing the generated swagger. Had it emitted an empty pattern, every generated SDK would have rejected all PUT/PATCH/DELETE calls client-side. It does not. The only schema delta in the entire generated output of this PR is the removal of three `"pattern": "^.+$"` entries per version, and I verified the GET retains `^.+$` while PUT, PATCH and DELETE drop it, matching stable 2023-04-01. That part of the change does what it claims.

Now the problems:

1. The comment says "Keep the empty pattern so writes match that contract". There is no empty pattern being kept; the pattern is being **removed**. A future reader chasing a validation bug will search the generated JSON for an empty pattern, find nothing, and lose time. Reword to "Omit the name pattern on writes to match stable 2023-04-01, which declares none."

2. The resulting asymmetry is now a permanent, undocumented property of the API: GET rejects an empty name at the SDK layer, PUT/PATCH/DELETE do not. Whatever the historical reason, that asymmetry is user-visible and belongs in the operation `@doc`, not only in a TypeSpec-side comment that never reaches customers.

3. This is the part I actually want answered. Removing the client-side constraint means `configurationAssignmentName` on the write path is now validated only by the RP. `^.+$` was the sole guard against an empty or whitespace-only name reaching the service. Please confirm, with a pointer to the MRP-side validator, that:
   - the name is normalised (trim, case-fold as appropriate) before it is used as a storage key, rather than passed through verbatim, and
   - empty and whitespace-only names are rejected server-side with a 400 rather than persisted.

   If the RP was relying on the swagger pattern for that guard, this change moves an input-validation boundary without a compensating server-side control, and it does so on the write path, which is the direction where it matters. A "the pattern was always cosmetic" answer is fine, but it needs to be stated explicitly with a code reference rather than inferred from the fact that stable omits it.

4. Minor design point: two near-identical interfaces (`...ReadOps` / `...WriteOps`) and four new parameter models are introduced to express a single-attribute difference. The scope-parameter and parent-parameter models are genuinely shared and that extraction is a real readability win. The read/write name split is the minimum viable way to express this in TypeSpec, so I am not asking for a rewrite, but the two interfaces should carry a one-line `/** */` doc each stating that the only difference is the name pattern, so nobody later "tidies up" by collapsing them and silently reintroduces the pattern on writes.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/MaintenanceConfiguration.tsp
```diff
@@ -164,6 +156,28 @@ interface MaintenanceConfigurationOperationGroup {
   >;
 }
 
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "This interface intentionally uses an explicit route to preserve legacy path casing."
+interface MaintenanceConfigurationsForResourceGroupRoutes {
```
rajabadmanaban, line 160: **[SHOULD-FIX] This is the right pattern. Apply it consistently, and confirm the response shape is unchanged.**

Preserving the legacy camelCase route via an explicit `@route` on a dedicated interface, with narrowly-scoped suppressions and a comment explaining why, is the correct approach. My objection is not to this block; it is that the same treatment was not given to `ApplyUpdateForResourceGroup_List`, which has an equally strong and in fact older claim to its camelCase route. See my comment on `routes.tsp:180`. Whatever is decided there, the two routes must be handled by the same rule.

Points on this block specifically:

1. I verified that moving `list` out of `MaintenanceConfigurationOperationGroup` into this new interface produced **no** change to the generated swagger for this operation: the path, `operationId` (`MaintenanceConfigurationsForResourceGroup_List`), tag, parameters and responses are byte-identical between base and head. The `@armResourceList(MaintenanceConfiguration)` addition likewise emits nothing. The claim in the PR description holds up. Recording that here so it does not have to be re-verified.

2. The interface name `MaintenanceConfigurationsForResourceGroupRoutes` is doing two jobs at once and the `Routes` suffix suggests a general-purpose container. Given the comment says future RG-scoped routes should be lowercase, this interface should be explicitly named for its single purpose, for example `MaintenanceConfigurationsLegacyCasedRoutes`, so it is obvious that adding an unrelated operation here would be wrong. Right now the name invites exactly that mistake.

3. `list` here returns `ArmResponse<ListMaintenanceConfigurationsResult> | MaintenanceError` as a raw union rather than going through the ARM operation templates, which is the reason two suppressions are needed. That is an acceptable trade for route preservation, but it does mean this operation no longer inherits future ARM template fixes (error-response shape, pagination conventions, common-types bumps) that the rest of the spec picks up for free. Please add that to the comment as a known cost so it is a deliberate decision on record rather than a silent divergence.

4. The `#suppress` justification on line 159 says "This interface intentionally uses an explicit route to preserve legacy path casing." Good. The one on line 167 (`no-openapi`) says only "Preserve the existing operation ID for backward compatibility." Name the operation ID in the justification, since a reviewer approving the `SuppressionReviewRequired` label should not have to read the code to learn which ID is being pinned.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/readme.md
```diff
@@ -197,6 +197,48 @@ directive:
   - suppress: RequestSchemaForTrackedResourcesMustHaveTags
     from: maintenance.json
     reason: False positive. ConfigurationAssignments is proxy resource at subscription/resourceGroup level.
+  - suppress: TrackedExtensionResourcesAreNotAllowed
```
rajabadmanaban, line 228: **[BLOCKING] Seven new suppressions are unrelated to this PR's stated purpose, and all of them are silently coupled to the casing change.**

I validated every new `where` clause by resolving it against the generated `2023-10-01-preview` and `2025-10-01-preview` output from this PR's head commit. All eight path-based selectors and all four definition-based selectors resolve. Mechanically these suppressions work. Two serious problems remain.

**1. Hidden coupling to the casing change.** Every path-based `where` clause here hard-codes lowercase `resourcegroups`. JSONPath matching is case-sensitive. These selectors resolve **only because** the casing change in this same PR landed. If ARM review requires reverting any of the casing changes, or if a future change restores camelCase on one of these routes, the corresponding suppression silently stops matching. It does not error, it does not warn; the suppression simply becomes a no-op and the underlying finding reappears, or worse, is masked as "already suppressed" in a stale S360 view. Two suppressions in this block (`TrackedExtensionResourcesAreNotAllowed`, `PatchBodyParametersSchema`) are pinned to routes this PR is actively re-casing. That is a fragile coupling with no signal on breakage, and it needs to be called out explicitly in the reasons.

**2. Scope creep on a PR that is nominally a build fix.** The stated purpose of this PR is to resolve a TypeSpec 1.15 `use-operation-decorator` failure and align with the private repo. `TrackedExtensionResourcesAreNotAllowed`, `PathForTrackedResourceTypes`, `AllTrackedResourcesMustHaveDelete`, `AvoidAdditionalProperties`, `OperationsApiSchemaUsesCommonTypes`, `PatchBodyParametersSchema` and `RequiredPropertiesMissingInResourceModel` are pre-existing API-quality findings. None of them is caused by the TypeSpec version bump. Bundling seven suppressions of long-standing ARM/S360 findings into a compatibility PR means the ARM reviewer clearing `SuppressionReviewRequired` is being asked to sign off on seven independent API-design concessions in a PR whose description frames them as incidental. Split these into their own PR with per-suppression justification so each gets reviewed on its merits.

On the individual justifications:

- `AllTrackedResourcesMustHaveDelete`: "the shared read-only public route causes this false positive" asserts a false positive without demonstrating it. State which route the validator is keying on and why DELETE cannot exist there. As written this is not reviewable.
- `AvoidAdditionalProperties` on `extensionProperties`: "intentional open string dictionary for extension-defined keys" is the right shape of argument, but an open dictionary on a control-plane resource is a durable API-quality liability, not a false positive. Say so, and link the tracking item for bounding it, rather than framing it as intended-and-closed.
- `PatchBodyParametersSchema`: the justification is correct that removing the `rebootSetting` default would change generated SDK behaviour. That is a genuine constraint. But it means PATCH carries a server-side default that will silently overwrite an unspecified value, which is the exact class of bug this rule exists to catch. Please confirm the RP treats an absent `rebootSetting` on PATCH as "leave unchanged" rather than "apply the default", and reference that behaviour here. If the RP does apply the default on PATCH, this is a live data-loss bug on a customer-facing write path and it should be filed regardless of the suppression.
- `RequiredPropertiesMissingInResourceModel` on `ListUpdatesResult`: "pageable response envelope, not an ARM resource" is correct and well-scoped. No concern.
- `TrackedExtensionResourcesAreNotAllowed`: the reason states the suppression is limited to the GET and PUT operations currently flagged, and that DELETE and collection operations produce no finding. "Currently flagged" pins this to today's validator behaviour. When the validator broadens, the omission will read as an oversight rather than a decision. State the invariant (only operations that echo the resource body carry `location`, hence only those are flagged) instead of the observation.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/ApplyUpdate.tsp
```diff
@@ -10,6 +10,9 @@ using TypeSpec.Http;
 using TypeSpec.OpenAPI;
 
 namespace Microsoft.Maintenance;
+
+// Common-types v5 emits this resource's id as arm-id, which is a known cross-version finding.
```
rajabadmanaban: **[NIT / SHOULD-FIX] This comment is duplicated verbatim four times, and in three of the four places it has no referent.**

The same two lines are added to `ApplyUpdate.tsp`, `ConfigurationAssignment.tsp`, `MaintenanceConfiguration.tsp` and `models.tsp`. In `models.tsp` it sits directly above the `id` property of `model Resource`, where "this resource's id" reads correctly. Here, and in the other two files, it is placed at namespace scope immediately after `namespace Microsoft.Maintenance;` and before the model's `/** */` doc block, so "this resource's id" points at nothing in particular. A reader cannot tell which resource or which `id` is meant.

Concretely:

- Keep the single well-placed instance in `models.tsp` on the `Resource.id` property, which is the one type these findings actually originate from.
- Delete the three namespace-scope copies, or if the intent is per-file discoverability, attach each to the specific model it describes.
- "Remove this note after 2025-10-01 is GA and becomes the v5 stable comparison baseline" is an action with no owner and no tracking item. Link an ADO work item. A dated cleanup instruction with no work item attached will still be here in three years, and by then nobody will know whether the underlying finding was resolved.
- Comments placed between a `namespace` declaration and a following doc comment are also fragile against formatter changes. Prefer attaching them to the declaration they describe.

This is cosmetic and does not affect the generated output, but four copies of a stale-dated note with no tracking item is exactly how spec files accumulate misleading commentary.

rajabadmanaban (changes requested):
## Review summary

I verified this PR by reconstructing the generated output independently rather than relying on the description: I downloaded the base and head `Maintenance.json` for both affected API versions, diffed them case-sensitively, and diffed the resulting path set against **every** previously published Maintenance API version (2020-04-01, 2021-05-01, 2023-04-01, 2022-07/2022-11/2023-09/2023-10/2025-10 previews).

**Verdict: request changes.** The core intent is right and most of the execution is correct, but one of the eight route changes is a regression away from a GA-stable contract, and it is mischaracterised in both the PR description and the justification comment as a "correction".

### What I confirmed is correct

- The **only** schema-level delta in the entire generated output is the removal of three `"pattern": "^.+$"` entries per version. `NamePattern = ""` emits no `pattern` key rather than `"pattern": ""`. This matters: an empty pattern would have broken every generated SDK on all writes. It does not.
- GET retains `^.+$`; PUT, PATCH and DELETE drop it. That matches stable 2023-04-01, and as a relaxation it is not a breaking change.
- `@armResourceList(MaintenanceConfiguration)` emits nothing. Confirmed.
- Moving `list` into a new interface leaves path, `operationId`, tag, parameters and responses byte-identical.
- The author's accounting of the breaking-change check (8 `RemovedPath` + 8 `AddedPath` per version, all from casing) is accurate and complete. There are no unrelated path removals.
- Three of the four affected route shapes genuinely were regressed by the TypeSpec conversion and are correctly restored to the casing used by GA 2023-04-01.
- All 12 new suppression `where` selectors resolve against the generated output.

### Blocking

1. **`routes.tsp:180` — `ApplyUpdateForResourceGroup_List` casing regression.** This route has been `resourceGroups` (camelCase) in **stable GA 2021-05-01** and in every version since. It has never been lowercase. This PR flips it. That is not restoration, it is a new divergence from a GA contract, and it contradicts the rule the PR applies two files away, where a whole interface plus two suppressions exist purely to preserve the same legacy camelCase on a different route. Revert it and pin it with an explicit `@route`.
2. **`back-compatible.tsp:148` — the global segment override is unguarded.** It rewrites every current and future RG-scoped route in one statement, across both already-published API versions, with no compile-time signal at the call site. It is the mechanism by which finding 1 slipped through, and as written it makes lowercase the silent default for every route added from now on. Needs an explicit exception list and a regression guard.
3. **`readme.md:200` — seven unrelated suppressions bundled into a build fix, all coupled to the casing change.** None is caused by the TypeSpec 1.15 bump. Every path selector hard-codes lowercase `resourcegroups`, so if any casing change is reverted the suppression silently becomes a no-op with no warning. Split them out so each is reviewed on its merits.

### Should fix

4. `ConfigurationAssignment.tsp:303` — the comment says "Keep the empty pattern"; the pattern is removed, not kept. More importantly, removing `^.+$` from writes moves an input-validation boundary. **Please confirm with a code reference that MRP normalises `configurationAssignmentName` before using it as a storage key and rejects empty/whitespace names server-side.** This is the one open functional risk in the PR.
5. `MaintenanceConfiguration.tsp:160` — right pattern, inconsistently applied; interface name invites misuse; the raw-union operation no longer inherits ARM template fixes and that cost should be recorded.
6. `ApplyUpdate.tsp:14` — the same two-line note is duplicated four times, and in three places has no referent. Dated cleanup instruction with no work item.

### Rubric items, answered explicitly

- **Unit test coverage.** Not applicable in the conventional sense; this repo has no unit tests. The equivalent assurance is the generated-output diff, and I performed it directly rather than trusting the description. Note that the PR's own validation statement, "byte-for-byte identical to the corresponding validated private-repository output", establishes parity with the private repo but says nothing about the delta against public `main` — which is where the regression lives. That is precisely the gap that let finding 1 through, and it is why the description should state the delta against `main`, not against the private repo.
- **Flaky tests.** None introduced. No test assets in this PR.
- **Feature flags.** Not applicable, and worth being explicit about why: in an ARM spec repo the API version *is* the rollout control. This PR deliberately edits two already-published preview versions in place, so there is no staged rollout and no kill switch. Correctness has to be established before merge, not mitigated after. This is what `VersioningReviewRequired` is flagging.
- **Outgoing calls / metrics / logging / PII.** No service code, no new outgoing calls, no logging. Not applicable.
- **Input normalisation before use as a key.** This is the live concern; see item 4.
- **Single Responsibility.** The parameter-model extraction is a genuine improvement. The read/write interface split is justified. The global segment override in `back-compatible.tsp` is the violation: one statement with namespace-wide, cross-version, cross-route effect.
- **Scope modifiers.** Not applicable to TypeSpec here; no visibility concerns. The analogous concern, operation grouping and client location, is handled correctly via `@@clientLocation`, which was updated in step with the interface move.
- **Sovereign clouds (Mooncake, Fairfax) and outage risk.** No outage risk from the casing changes. ARM path matching is case-insensitive on these segments, the RP already serves the lowercase forms in GA, and no path, `operationId`, parameter or response shape is added or removed. The customer-visible impact is limited to regenerated SDK URL strings, which resolve identically. Sovereign clouds are not differentially affected: this spec is not cloud-partitioned and carries no per-cloud data files, so the Gold-Config-style failure mode of divergent per-cloud configuration does not apply here. The real risk in this PR is contractual and reviewer-facing, not availability-facing.
- **Gold-Config repo.** Not applicable; this is not a Gold-Config change.

### Not blocking, but worth stating

The PR carries `VersioningReviewRequired`, `SuppressionReviewRequired` and `NotReadyForARMReview`, and lacks `PublishToCustomers`. Fixing finding 1 reduces the breaking-change findings from 8 to 7 per version but does not eliminate them, so the `aka.ms/brch` exception process is still required regardless. That makes it worth getting the route set exactly right first, since the exception request should not have to be amended later.

I am not approving this. The single change I would insist on before re-review is the `applyUpdates` route revert; the rest is negotiable with adequate justification.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/routes.tsp
dvarshney29: Thank you for tracing the historical public versions. We confirmed that the currently validated private specifications use lowercase `resourcegroups` for this exact `ApplyUpdateForResourceGroup_List` route in both `2023-10-01-preview` and `2025-10-01-preview`.

We intend to keep the public specification in parity with those private source contracts. Lowercase `resourcegroups` is also the convention being followed for new Maintenance resource-group-scoped APIs. For example, the private `2026-02-01-preview` Maintenance Plans collection and item routes both use lowercase `resourcegroups`.

For that reason, this PR deliberately applies lowercase `resourcegroups` consistently to the affected Maintenance routes rather than preserving this operation as another camel-case exception. We acknowledge that the currently published public previews used camel-case and that the same-version Swagger checker therefore reports this as a removed/added path pair. The detailed PR comment documents that finding, and we are requesting the applicable versioning/breaking-change review for the intentional synchronization.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/back-compatible.tsp
dvarshney29: Updated in commit [`0673d827e6e`](https://github.com/Azure/azure-rest-api-specs/commit/0673d827e6e).

We are retaining the service-wide segment override so all new Maintenance resource-group-scoped APIs inherit lowercase `resourcegroups` by default. The inline comment now explicitly documents that policy, identifies `MaintenanceConfigurationsForResourceGroupRoutes.list` as the known legacy camel-case exception, and requires future exceptions to use an explicit `@route` rather than changing the default.

I also added a `Resource-group route casing` compatibility section to `readme.md` recording the same policy and exception for future authors. TypeSpec validation and compilation pass, and the generated 2023 and 2025 `Maintenance.json` hashes are unchanged.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/ConfigurationAssignment.tsp
dvarshney29: Addressed the comments/documentation items in commit [`726b614cfd0`](https://github.com/Azure/azure-rest-api-specs/commit/726b614cfd0):

- Reworded the inline note to state that writes **omit** the OpenAPI name pattern; it no longer implies that an empty pattern is emitted.
- Added one-line documentation to the read and write helper interfaces explaining that GET retains `^.+$`, while writes omit the pattern to match stable `2023-04-01`.

This is documentation-only: no operation-level `@doc`, decorators, parameter types, or behavior changed. TypeSpec validation and compilation pass, and the generated `2023-10-01-preview` and `2025-10-01-preview` `Maintenance.json` hashes are unchanged.

We are not claiming in this update that server-side validation has been verified. Confirming normalization and rejection of empty or whitespace-only names requires an implementation-side MRP validator reference outside this specification repository; that remains a separate follow-up item.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/MaintenanceConfiguration.tsp
dvarshney29: Documented the compatibility tradeoff in commit [`205adc96fe0`](https://github.com/Azure/azure-rest-api-specs/commit/205adc96fe0).

We are intentionally keeping `MaintenanceConfigurationsForResourceGroupRoutes`, its suppressions, and the TypeSpec implementation unchanged to maintain source parity with the validated private specification. No `.tsp` or generated `.json` files were changed.

The `readme.md` compatibility section now records that this operation:

- preserves the legacy camel-case `resourceGroups` route,
- pins operation ID `MaintenanceConfigurationsForResourceGroup_List`,
- uses an explicit response union instead of an ARM list-operation template, and
- will therefore require explicit review before adopting future ARM template changes.

This keeps the published contract and private/public TypeSpec alignment intact while putting the known maintenance cost on record for future authors.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/readme.md
dvarshney29: Thank you for validating that the selectors resolve against both generated preview versions. We are intentionally retaining these suppressions in this PR.

They mirror the suppressions already applied and validated in the private specification repository and are being carried into the public repository to keep the two specification sources aligned. These rules report pre-existing compatibility or API-quality findings; resolving them directly would require changes to established routes, models, response shapes, or generated SDK behavior. The suppressions prevent those findings from driving unrelated breaking contract changes while this specification is synchronized.

The lowercase `resourcegroups` selectors are intentional and match the route-casing policy retained in this PR and in the current private 2023 and 2025 preview specifications. We have documented that policy and the explicit legacy camel-case exception in `readme.md`.

For those reasons, we do not plan to remove or split these suppressions from this synchronization PR. No `.tsp` or generated `.json` change is being made in response to this comment.

specification/maintenance/resource-manager/Microsoft.Maintenance/Maintenance/ApplyUpdate.tsp
dvarshney29: Thank you for calling this out. We are intentionally retaining these comments unchanged in this PR to preserve source parity with the validated private TypeSpec specification.

The namespace-level copies are repeated for per-file discoverability around the affected compatibility definitions; they are comments only and do not alter the generated OpenAPI descriptions or contract. Consolidating or repositioning them only in the public repository would create unnecessary divergence between the public and private TypeSpec sources.

We will therefore leave them as-is for this synchronization PR and defer any cleanup until it can be applied consistently to both repositories. We are not creating a new ADO tracking item as part of this PR.

rajabadmanaban (approved):
## Re-review

I re-verified rather than relying on the resolution claims. Since my review commit `851c0142`, exactly three files changed: `ConfigurationAssignment.tsp`, `back-compatible.tsp` and `readme.md`. No `.tsp` route change and no regenerated `.json`. The generated `2023-10-01-preview` and `2025-10-01-preview` output is byte-identical to the version I validated, so the "hashes unchanged" claim holds and none of the follow-up commits introduced new surface. CI is now down to the single expected `Swagger BreakingChange` failure.

**Approving.** Three items were addressed with changes, three were declined with coherent written rationale, and the one item I originally said I would insist on has turned into a versioning-policy judgment that belongs to the ARM reviewer, not to me.

### On the `applyUpdates` route: I am withdrawing my objection, and I want to be explicit about why

My factual finding stands and I do not retract it: this route was camelCase in stable `2021-05-01` and in every version it has ever appeared in, so it is a change away from precedent, not a restoration. That mattered, and the record is now corrected. The PR no longer rests on the claim that all eight route changes are the same restoration.

What changed my assessment is evidence from my own analysis that I under-weighted the first time. The current GA version, `2023-04-01`, does **not** contain this route at all. The only GA that does is `2021-05-01`, and this PR does not modify any stable specification. So the affected surface is preview-only, preview versions carry no compatibility guarantee, ARM path matching is case-insensitive on this segment, and no `operationId`, parameter or response shape moves. I framed this as a GA-contract violation; on the evidence it is a preview re-casing with no runtime impact. That is a materially weaker objection than the one I made, and I should have weighted it correctly the first time.

The remaining question, whether re-casing two published previews to reach parity with the private source is acceptable, is exactly what `VersioningReviewRequired` and `aka.ms/brch` exist to decide. The author is routing it through that gate rather than around it. Holding a blocking review on a call reserved to the ARM reviewer would obstruct without adding protection.

**One dependency I could not verify.** The rationale rests on the private `2023-10-01-preview` and `2025-10-01-preview` already using lowercase for this route, and on `2026-02-01-preview` Maintenance Plans following lowercase. I have no access to the private repository, so I am taking that on trust. If it is inaccurate the justification collapses, and the ARM reviewer should confirm it directly rather than inheriting my assumption.

### Accepted as addressed

- **`back-compatible.tsp`.** The revised comment now names `MaintenanceConfigurationsForResourceGroupRoutes.list` as the known exception and directs future exceptions to explicit `@route` rather than changing the default. Combined with the `readme.md` section, this addresses the "future authors silently re-break this" concern. I had asked for a mechanical regression guard; I accept documentation here, because the generated `.json` is checked in, so any future casing drift surfaces as a visible diff at PR review time. That is a real guard, not a notional one.
- **`ConfigurationAssignment.tsp`.** The misleading "keep the empty pattern" wording is fixed, and both helper interfaces now document that the only difference is the name pattern. That was the specific trap I was worried about, where someone later collapses the two interfaces and silently reintroduces the pattern on writes.
- **`MaintenanceConfiguration.tsp`.** Declining the rename for source parity is reasonable. The `readme.md` entry records the pinned operation ID and, importantly, that the operation will not inherit future ARM template changes. That was the substance of the ask; the naming point was secondary.

### Accepted as the reviewer's call, not mine

- **Suppressions.** I still think seven pre-existing API-quality concessions travelling inside a synchronization PR is poor packaging, and I stand by that. But the PR carries `SuppressionReviewRequired`, so an ARM reviewer must clear each one regardless of how they are packaged. The governance is intact either way. My earlier concern about the lowercase JSONPath selectors silently becoming no-ops is moot now that the casing policy is being retained and documented. I re-confirmed all twelve selectors resolve against the current generated output.
- **Duplicated comments.** A nit, declined for parity. Fine.

### Two items I am recording as unresolved, not blocking

Neither is introduced by this PR and neither should hold it up, but both were raised and neither has an answer or a tracking item:

1. **`rebootSetting` default on PATCH.** The `PatchBodyParametersSchema` justification is correct that removing the default would change generated SDK behaviour, but that leaves a server-side default on a PATCH body. If the RP applies the default when the field is absent rather than leaving the value unchanged, that is a silent-overwrite bug on a customer-facing write path, and it exists independently of whether the finding is suppressed. Worth confirming and filing separately.
2. **Server-side name validation.** The author states plainly that normalization and rejection of empty or whitespace-only `configurationAssignmentName` has not been verified and needs an MRP-side reference. I appreciate the straight answer rather than an assertion. Since `^.+$` was the only client-side guard on the write path, this should become a tracked follow-up rather than an open loop.

### One small request

The PR description still reads "Updates client routing and the Apply Updates route to use the intended operation and route casing." Given the discussion above, "intended" understates it. Please reword to state that this route is being changed from the published camelCase to lowercase for parity, and that the change is covered by the versioning review. The inline discussion is now accurate; the description should match it so the ARM reviewer is not working from the softer framing.

Approving on the spec-synchronization mechanics. The versioning and suppression decisions remain with the ARM reviewer.

PR #45071, opened 2026-07-28
Foundry: Add Microsoft 365 publish API reference

MattSable:
## Summary

Documents the agent Microsoft 365 / Microsoft Teams publish API in the Foundry data-plane
API reference. This surface is exposed publicly through APIM passthrough, so it belongs in
the Foundry reference. Targeting `feature/foundry-release` since this is a newly added API.

Adds a new `agents-microsoft365` TypeSpec project describing three operations:

| Verb | Route | Description |
| ---- | ----- | ----------- |
| POST | `/agents/{agent_name}/microsoft365/publish` | Publish an agent to Microsoft 365 and return the title id and Teams app id. |
| POST | `/agents/{agent_name}/microsoft365/zip` | Generate the Microsoft Teams app package (`application/zip`) without publishing. |
| GET  | `/agents/{agent_name}/microsoft365/publishdefaults` | Return default and previously-published values to pre-populate a publish request. |

## Changes

- New `src/agents-microsoft365/models.tsp` and `src/agents-microsoft365/routes.tsp`
  describing the request/response models and the three operations. Field names use the
  camelCase wire contract the service actually serializes.
- Registered the `Agent Microsoft 365 Publishing` tag under `Agents Root` in
  `tag-definitions.tsp` and imported the new routes in `main.tsp`.
- Regenerated the OpenAPI descriptions (`openapi3/v1` and
  `openapi3/virtual-public-preview`) via `tsp compile .`. The change is purely additive.

## Validation

- `tsp compile .` completes successfully with no new warnings from the added project.
- Generated OpenAPI diff is additive only (new paths, schemas, and tag); no changes to any
  existing operation or schema.

PR #45710, opened 2026-08-21
Foundry: Add publish approval status to agent endpoint config

MattSable:
Adds an optional `publish_approval_status` property to `AgentEndpointConfig`, backed by a new `PublishApprovalStatus` extensible union (`NotPublished`, `Pending`, `Approved`, `Rejected`, `NoApprovalNeeded`).

The Agents service now reports the Microsoft Agent Certification review status of an agent's Microsoft 365 store title on the agent endpoint config returned by Get Agent. The value is server-populated and optional: it is absent when the status could not be determined, and an absent value must not be read as "not published". No value is terminal, because publishing a new version of an agent reuses the same store title and sends it back through review.

Related context: the publish API in #45071.

Regenerated `openapi3/v1` and `openapi3/virtual-public-preview` (json + yaml) via `npx tsp compile .` from `specification/ai-foundry/data-plane/Foundry`.

Open question: the values here are PascalCase to match what the service actually returns, whereas the neighbouring `AgentIdentityStatus` union uses lowercase values. Is that inconsistency acceptable, or should the wire format be changed?


MattSable:
Marked `publish_approval_status` as read-only via `@visibility(Lifecycle.Read)`, matching the existing precedent in `src/agents/models.tsp` (`state`, `state_source`, `instance_identity`, `blueprint`, `agent_guid`).

The property is server-populated: the service computes it and ignores any client-supplied value. Previously it was emitted onto `AgentEndpointConfigUpdate`, the PATCH request model, which incorrectly advertised it as settable. It now appears only on the `AgentEndpointConfig` response schema (with `readOnly: true`) and is gone from `AgentEndpointConfigUpdate` in both `openapi3/v1` and `openapi3/virtual-public-preview`.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -726,6 +726,30 @@ model AgentEndpointConfig {
...
+  @visibility(Lifecycle.Read)
+  publish_approval_status?: PublishApprovalStatus;
+}
+
+@doc("The Microsoft Agent Certification review status of the Microsoft 365 store title published for an agent.")
+union PublishApprovalStatus {
+  string,
+
+  @doc("The agent has never been published to the Microsoft 365 store, so there is nothing to review.")
+  NotPublished: "NotPublished",
```
dargilco: please use lower letter snake_case for union strings (e.g. "not_published")

MattSable:
Updated `PublishApprovalStatus` to use lowercase snake_case for both the union member identifiers and the string values, matching the neighbouring `AgentIdentityStatus` union: `not_published`, `pending`, `approved`, `rejected`, `no_approval_needed`. Regenerated the OpenAPI 3 output for both `v1` and `virtual-public-preview`. The Agents service will normalize to these values, so the emitted contract and the service response agree.

glecaros (approved):

PR #45721, opened 2026-08-21
Added API version 2025-01-23-preview for OTLP support

qinjoshua:
> [!IMPORTANT]
> **Action needed from reviewer: please apply the `Approved-Avocado` label.**
>
> The required `Swagger Avocado` check reports two errors, both expected for this PR. Per
> [aka.ms/azsdk/pr-suppressions](https://aka.ms/azsdk/pr-suppressions), Avocado has no
> `suppressions.yaml` support — the `Approved-Avocado` label is the only override, and I do not
> have permissions to add labels on this repo.
>
> 1. **`MULTIPLE_API_VERSION`** — Pre-existing and **not introduced by this PR**. The default tag
>    `package-2024-08-01` has referenced 12 input files spanning 9 API versions (`2015-05-01`,
>    `2020-02-02`, `2020-03-01-preview`, `2020-11-20`, `2021-03-08`, `2021-10-14`, `2022-06-15`,
>    `2023-06-01`, `2024-02-01-preview`) on the base branch already. This PR does not modify the
>    default tag; it only *adds* an opt-in `package-2025-01-23-preview` tag. Avocado re-reports this
>    rule for any PR touching the RP folder. Consolidating would require re-versioning every
>    ApplicationInsights sub-API, or splitting the composite readme — already deferred in
>    `suppressions.yaml` ("Not ready to split as independent services").
>
> 2. **`NOT_LATEST_API_VERSION_IN_DEFAULT_TAG`** — By design. The new API version is a **preview**
>    (`2025-01-23-preview`); the default tag intentionally stays on GA `stable/2020-02-02` so the
>    default SDK package remains stable. Promoting a preview into the default tag would ship a
>    preview-based GA SDK for all of ApplicationInsights.
>
> The GA surface is unchanged: `stable/2020-02-02/components_API.json` is untouched, and the new
> version adds only 6 properties + 1 definition with identical paths and no removals.
>
> See [this comment](https://github.com/Azure/azure-rest-api-specs/pull/45721#issuecomment-5375116316)
> for the full analysis.

> [!TIP]

## PR review workflow diagram

## Purpose of this PR

  - [ ] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

</details>

## Getting help

  and https://aka.ms/ci-fix.


qinjoshua:
## Requesting `Approved-Avocado` label

Both Avocado errors are expected for this PR and cannot be fixed within its scope. Details below for the reviewer.

### 1. `MULTIPLE_API_VERSION` — pre-existing, not introduced by this PR

The default tag `package-2024-08-01` has referenced 12 input files spanning **9 distinct API versions** since long before this PR:

| API version | Files |
| --- | --- |
| `2015-05-01` | `openapi.json`, `favorites_API.json`, `webTestLocations_API.json`, `analyticsItems_API.json` |
| `2020-02-02` | `components_API.json` |
| `2020-03-01-preview` | `componentLinkedStorageAccounts_API.json` |
| `2020-11-20` | `workbookTemplates_API.json` |
| `2021-03-08` | `workbookOperations_API.json` |
| `2021-10-14` | `livetoken_API.json` |
| `2022-06-15` | `webTests_API.json` |
| `2023-06-01` | `workbooks_API.json` |
| `2024-02-01-preview` | `deletedWorkbooks_API.json` |

This PR **does not modify the default tag** — the only `readme.md` change is an added, opt-in `### Tag: package-2025-01-23-preview` section. The error surfaces because Avocado's `isPRRelatedError()` special-cases `MULTIPLE_API_VERSION` so it is re-reported for any PR touching this RP folder, even when the condition already exists on the base branch.

Resolving it properly would require either re-versioning every ApplicationInsights sub-API (components, favorites, webTests, workbooks, analyticsItems, workbookTemplates, workbookOperations, livetoken, linkedStorageAccounts, deletedWorkbooks) onto one common version, or splitting the composite readme into per-service-group readmes. The latter is already tracked and deliberately deferred — see `suppressions.yaml` in this folder: _"Not ready to split as independent services."_

### 2. `NOT_LATEST_API_VERSION_IN_DEFAULT_TAG` — by design

This PR adds preview API version `2025-01-23-preview` for Components (OTLP support). The default tag intentionally remains on the GA version `stable/2020-02-02/components_API.json` so that the default SDK package stays stable; pulling a preview API version into the default tag would ship a preview-based GA SDK for all of ApplicationInsights.

The new version is exposed through a separate opt-in tag `package-2025-01-23-preview`, and is registered in `Components/service.yaml`.

This error is triggered because the default tag already contains preview inputs (`2020-03-01-preview`, `2024-02-01-preview`), which makes Avocado evaluate the default tag against the newest preview across the whole readme directory.

The existing stable API surface is unchanged: `stable/2020-02-02/components_API.json` is untouched, and the new version's swagger differs only by 6 added properties and 1 added definition, with identical paths and no removals.

specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp
```diff
@@ -283,6 +308,52 @@ model ApplicationInsightsComponentProperties {
...
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Name is consistent with the existing wire contract of this API."
+  @added(Versions.v2025_01_23_preview)
+  AzureMonitorWorkspaceResourceId?: Azure.Core.armResourceIdentifier;
+
+  /**
+   * Indicates the state of the OpenTelemetry ingestion into the linked Azure Monitor workspace.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Name is consistent with the existing wire contract of this API."
+  @added(Versions.v2025_01_23_preview)
+  AzureMonitorWorkspaceIngestionMode?: AzureMonitorWorkspaceIngestionMode;
```
sandipsh: **[NEW] 🔴 Blocking** **[[OAPI034](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034-clear-field-ownership)]** `specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp` - line 324 - `AzureMonitorWorkspaceIngestionMode` is writable and the PUT example sends `Enabled`, but `Disabled` is documented as a value only the resource provider sets. That gives one property both client-owned configuration and server-owned status semantics, so SDKs allow clients to submit a state the service says only it controls and ARM What-If can report drift.

**Classification reasoning:** Introduced in this PR; neither the property nor its union exists at base SHA `167da049737b838c6bea8fc07c7f223971bf77f9`.

**Suggested fix:** Split desired configuration from observed status: keep a client-owned setting with only client-selectable values, and expose the service-computed state (including `Disabled`) as a separate `@visibility(Lifecycle.Read)` property. If clients should not control this at all, make the existing property read-only instead.

specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp
```diff
@@ -283,6 +308,52 @@ model ApplicationInsightsComponentProperties {
    */
   #suppress "@azure-tools/typespec-azure-core/casing-style" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
   ForceCustomerStorageForProfiler?: boolean;
+
+  /**
+   * Resource Id of the Azure Monitor workspace which the OpenTelemetry data will be ingested to.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Name is consistent with the existing wire contract of this API."
```
sandipsh: **[NEW] 🟠 Warning** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp` - line 315 - These six new `casing-style` suppressions say they preserve an existing wire contract, but the properties are all introduced by this PR with `@added(Versions.v2025_01_23_preview)` and are absent from the prior API. An inaccurate suppression reason makes the exception unauditable and leaves non-idiomatic SDK names as permanent API surface.

**Classification reasoning:** All six suppressions and properties are introduced in this PR and absent at base SHA `167da049737b838c6bea8fc07c7f223971bf77f9`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the session SHA. If an approved compatibility exception covers these exact wire names, apply the label and link the shipped contract that requires them; otherwise address the suppressions.

**Suggested fix:** Use idiomatic camelCase TypeSpec property names and preserve required JSON wire names with the supported projection/wire-name mechanism. If these names already shipped outside this spec, replace each reason with a specific compatibility anchor showing where and when that wire name became immutable.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration(s), converged) against commit [`0b20fa7`](https://github.com/Azure/azure-rest-api-specs/pull/45721/commits/0b20fa702ecbae211d58bcd2c8db69298c8063dc). See inline comments for findings 1-2.

sandipsh:
Thanks, this rationale checks out. `MULTIPLE_API_VERSION` is inherited from the unchanged default tag, and keeping the new preview in an opt-in tag avoids moving the default SDK package off GA. I support adding `Approved-Avocado` for these two diagnostics.

specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp
qinjoshua: This design pattern is replicated from `IngestionMode`, which also has a `Disabled` mode that can be set by our service. Clients can set for `AzureMonitorWorkspaceIngestionMode` is `NotOptedIn` - the service side can never transition `NotOptedIn` to `Disabled`, because when it's not opted in there is nothing to disable. The only time that a client can transition to `Disabled` is when the workspace ingestion mode was previously `Enabled`, but if the workspace ingestion mode was previously `Enabled` the property will from then on function as a read-only property. Because of the limited set of states and transitions between states, we think that the risk of drift is minimal and most likely won't impact the usability of this API.

We prefer this API design because it matches our existing `IngestionMode` property, reducing confusion and improving customer usability. In addition, we have existing customers using OTLP this way in preview and we have no complaints from drift causing issues.

specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp
qinjoshua: Agreed — the suppression reason was inaccurate. These six properties are new to the spec, so there was no prior API version for them to be "consistent with." The intent was that the resource provider already serializes these exact JSON names, but the wording did not say that and was not auditable.

Fixed by adopting your suggested approach rather than rewording. The TypeSpec properties are now idiomatic camelCase, with the required JSON wire names preserved via `@encodedName`:

```tsp
/**
 * The OTLP endpoint to which OpenTelemetry metrics for this component are sent.
 */
@added(Versions.v2025_01_23_preview)
@encodedName("application/json", "OTLPMetricsEndpoint")
@visibility(Lifecycle.Read)
otlpMetricsEndpoint?: url;
```

All six `casing-style` suppressions are removed — none remain in `models.tsp`.

The emitted wire contract is unchanged: the JSON keys are still `AzureMonitorWorkspaceResourceId`, `AzureMonitorWorkspaceIngestionMode`, `DataCollectionRuleResourceId`, `OTLPMetricsEndpoint`, `OTLPLogsEndpoint` and `OTLPTracesEndpoint`, matching the deployed resource provider. The only delta in `components_API.json` is an added `x-ms-client-name` per property, which gives SDKs the idiomatic name.

Verified locally: `tsp compile --warn-as-error` succeeds with no suppressions, `client.tsp` compiles, `tsp format` reports no changes, and `oav validate-example` / `oav validate-spec` both pass.

specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/Components/models.tsp
qinjoshua: Following up on the rationale above — we would like to keep this design and are **requesting an exception via the `Approved-Suppression` label**.

A `suppressions.yaml` entry has been added at `specification/applicationinsights/resource-manager/Microsoft.Insights/ApplicationInsights/suppressions.yaml`, scoped to this API version only:

```yaml
- tool: OAPI034
  paths:
    - ./preview/2025-01-23-preview/components_API.json
  reason: >
    ... (full justification in the file)
```

The substance of the justification:

**1. Consistency with the existing property on the same resource.** `IngestionMode` on `ApplicationInsightsComponentProperties` is also a writable property that the resource provider can move to a service-set `Disabled` state (`ComponentsProvider.cs:3473`). Matching that shape keeps the two ingestion settings consistent for customers. Worth noting the new property is actually *more* discoverable than the precedent it follows: the shipped `IngestionMode` enum omits `Disabled` from the spec entirely (`["ApplicationInsights","ApplicationInsightsWithDiagnosticSettings","LogAnalytics"]`), whereas `AzureMonitorWorkspaceIngestionMode` documents the service-set value and says explicitly that clients cannot set it.

**2. The state machine is closed, so the drift the rule guards against cannot occur.** Clients may only set `NotOptedIn` or `Enabled`:

| Behavior | Source |
| --- | --- |
| Client sending `Disabled` is rejected with `ArgumentException` | `ComponentsProvider.cs:1752` |
| `Enabled` → `NotOptedIn` is rejected ("cannot opt out once enabled") | `ComponentsProvider.cs:1778` |
| `Disabled` set by the RP only when the AMW link breaks | `ComponentsProvider.cs:3455`, `:3478` |
| Read defaults to `NotOptedIn` when unset | `ComponentsProvider.cs:5007` |

The service never transitions `NotOptedIn` → `Disabled`, because a component that has not opted in has no Azure Monitor workspace link to break. `Disabled` is reachable only from `Enabled`, and once a component reaches `Enabled` the property is effectively read-only. A client can therefore never submit a value the service will contradict, so What-If has no state in which to report drift.

**3. Cost of the alternative.** The resource provider serializes a single wire field. Splitting desired configuration from observed status requires a Mgmt-AppInsights-CDS change to emit a second field, and `2025-01-23-preview` is already deployed to production with existing preview customers using OTLP this way — removing `Disabled` from the current field would break them. We would rather revisit the split at GA, where a contract change is expected.

Making the property read-only is not an option, since setting `Enabled` is the opt-in gesture for OTLP ingestion.

The ownership boundary is now also explicit in the contract itself, on the property and on each enum value — for example `Disabled` is documented as "set only by the service and is rejected if specified by a client". The two new examples in this PR (`ComponentCreateWithoutOtlp`, `ComponentCreateWithManagedWorkspaces`) exercise only the two client-settable values.

Could you apply `Approved-Suppression` if the justification is acceptable? Happy to adjust the wording in `suppressions.yaml` if you would like it framed differently.

qinjoshua:
## Review findings addressed — ready for another look

Pushed `c48f741e68` (and `0139195bc4`) addressing both findings from the [ARM API review](https://github.com/Azure/azure-rest-api-specs/pull/45721#issuecomment-5375116316) on `0b20fa7`.

**Finding 2 — `casing-style` suppressions (fixed as suggested).** The six new OTLP properties now use idiomatic camelCase TypeSpec names with `@encodedName("application/json", ...)` preserving the JSON wire names the resource provider already serializes. All six suppressions are removed. The wire contract is unchanged — the only swagger delta is an added `x-ms-client-name` per property. Details in [this reply](https://github.com/Azure/azure-rest-api-specs/pull/45721#discussion_r3876529626).

**Finding 1 — OAPI034 (exception requested).** We would like to keep the current design; a version-scoped `suppressions.yaml` entry documents why the closed state machine makes the What-If drift unreachable, and the ownership boundary is now explicit on the property and each enum value. Full justification in [this reply](https://github.com/Azure/azure-rest-api-specs/pull/45721#discussion_r3876530280) — **this still needs the `Approved-Suppression` label** if the rationale is acceptable.

**CI status:** LintDiff, ModelValidation, SemanticValidation, BreakingChange, BreakingChange(Cross-Version), TypeSpec Validation, TypeSpec Requirement, TypeSpec Suppressions, ARM Modeling Review and ARM Auto SignOff all pass on the new commits. `Swagger Avocado` remains covered by the `Approved-Avocado` label already applied — thank you for that.

I do not have write access, so I cannot re-trigger the ARM API reviewer myself. The PR currently carries `ARMSignedOff` / `ARMAutoSignedOff-IncrementalTSP` from the automated pass on Aug 26, which survived these pushes. If you would like the agent to re-verify the two fixes, please re-run `/arm-review`; otherwise the only outstanding item is the `Approved-Suppression` decision on Finding 1.

qinjoshua:
## Requesting re-review of the latest commits

The `ARMSignedOff` / `ARMAutoSignedOff-IncrementalTSP` labels were applied on 2026-08-26 against the state of the PR *before* the review findings were addressed. Two commits have landed since:

| Commit | Change |
| --- | --- |
| `0139195bc4` | Finding 2 fix (camelCase + `@encodedName`, all six `casing-style` suppressions removed) and Finding 1 (version-scoped `OAPI034` entry in `suppressions.yaml` plus explicit ownership documentation) |
| `c48f741e68` | SpellCheck fix for the new `OAPI` token |

Since these changed the reviewed API surface, we would like an ARM reviewer to validate the current state rather than rely on the earlier sign-off.

**Could a maintainer please re-queue this PR for ARM review?** Either:

- add `WaitForARMFeedback` and remove `ARMSignedOff`, or
- run `/arm-review`

I do not have write access on this repository, so I cannot change labels or trigger the reviewer myself.

For context on the cost: the ARM review queue currently holds 9 PRs, 4 of which were opened before this one (opened 2026-08-21), so this would re-enter at roughly position 5 — about 1 business day per the [documented ETA formula](https://eng.ms/docs/products/azure-developer-experience/design/specs-pr-guides/pr-arm-review).

All required checks are currently green, including `Swagger Avocado` via the `Approved-Avocado` label. The one open item from the previous review is the `Approved-Suppression` decision on Finding 1, discussed in [this thread](https://github.com/Azure/azure-rest-api-specs/pull/45721#discussion_r3876530280).

qinjoshua:
Looks like the changes were already approved for release, but I may need another review in order to merge due to diff with latest changes.

qinjoshua:
@sandipsh @alexeldeib — small request, blocked on permissions.

Could one of you add the **`WaitForARMFeedback`** label to this PR? (No need to remove `ARMSignedOff`.)

**Why:** the `ARMSignedOff` / `ARMAutoSignedOff-IncrementalTSP` labels were applied 2026-08-26, before the review findings were addressed. Two commits landed after that sign-off:

- `0139195bc4` — Finding 2 fix (camelCase + `@encodedName`, all six `casing-style` suppressions removed) and Finding 1 (version-scoped `OAPI034` entry in `suppressions.yaml` plus explicit ownership docs)
- `c48f741e68` — SpellCheck fix

Since those changed the reviewed API surface, we would like the current state validated rather than relying on the earlier sign-off. Running `/arm-review` instead works equally well if you prefer.

I do not have write access on this repo, so I cannot set labels or trigger the reviewer myself.

All required checks are green, including `Swagger Avocado` via `Approved-Avocado`. The one still-open item from the previous review is the `Approved-Suppression` decision on Finding 1, discussed in [this thread](https://github.com/Azure/azure-rest-api-specs/pull/45721#discussion_r3876530280) — happy to take that either way.

ramanujbhattacharjee93 (approved):

ramanujbhattacharjee93 (approved):

Issue #45521, opened 2026-08-14
Auto-apply + atomic flip to data-plane-review-* labels (build on new label, retire APIStewardshipBoard-*)


samvaity:
Part of #45437. Dependency of #45313, #45451, and #45450. Confirmed unblocked: `APIStewardshipBoard-ReviewRequested` is applied only manually today, nothing external (no ADO/openapi pipeline) applies it, so this repo controls 100% of both the applier and the readers. That lets the atomic label flip below happen here without external coordination.

## Why this is the keystone
`APIStewardshipBoard-ReviewRequested` is never applied automatically today, so everything keyed off it stays dormant until a human applies it by hand: native team assignment (#45313), the Project 1018 data-plane view (#45451, view now live), and the review-agent trigger (#45450). Building the auto-apply activates all three.

Because the auto-apply is greenfield (nothing in `prSummary.ts` references either label yet), it should be built directly on the NEW label `data-plane-review-requested`, not the old `APIStewardshipBoard-*` name we intend to delete. Wiring fresh automation onto a soon-to-be-deleted label would just create debt. So this issue also carries the atomic reader flip that #45452 originally bundled (that part is no longer 196-blocked now that the 1018 view exists and the applier is confirmed manual).

## Work (one atomic PR - all readers and the applier move together)
1. Auto-apply: in `prSummary.ts`, apply `data-plane-review-requested` when `impactAssessment.dataPlaneRequired && impactAssessment.isNewApiVersion`.
2. Merge gate: in `.github/workflows/src/summarize-checks/labelling.js` (`rulesPri0dataPlane`), switch `anyPrerequisiteLabels` to `["data-plane-review-requested"]` and `anyRequiredLabels` to `["data-plane-review-signoff"]`, and drop the `allPrerequisiteLabels: ["data-plane", "new-api-version"]` fallback (per the TODO at labelling.js line 832 - the fallback existed only because auto-apply was missing).
3. Assignment: in `.github/workflows/src/data-plane-review/assign-reviewers.js`, repoint `TRIGGER_LABEL` -> `data-plane-review-requested` and `SIGNOFF_LABEL` -> `data-plane-review-signoff` (and the test mock config key).
4. Sign-off protection: in `.github/protected-labels.yml`, move the auth-gated key from `APIStewardshipBoard-SignedOff` to `data-plane-review-signoff` (keep the same roster). Otherwise the new sign-off label is unprotected and anyone could clear the gate.
[...]

Issue #45450, opened 2026-08-12
Point review agent at data-plane-review-requested trigger


samvaity:
Part of #45437. Depends on the new labels existing (see the label-creation sub-issue).

Point the review agent at the real trigger:

- Change `.github/workflows/data-plane-api-review.md` to trigger on
  `data-plane-review-requested` instead of the prototype `data-plane-api-review-needed`
  (0 real runs).
- Delete the prototype `data-plane-api-review-needed` label.

Result: the agent and human intake share one signal.

samvaity:
Blocked until initial review agent seeding.
cc: @xirzec

Issue #45437, opened 2026-08-12
Proposal: data-plane review label consolidation


samvaity:
Scope: data-plane stewardship review labels only. Not ARM labels. Not merging stewardship
review with SDK/namespace review.

## Context

We are wiring up data-plane review now (reviewer assignment, board view, review agent). The labels these depend on use three different naming styles, and some of them are barely used or never used. Pick one naming scheme and drop the dead labels before we build more on top of them.

Styles in use today:
- PascalCase: `SuppressionReviewRequired`, `BreakingChangeReviewRequired`
- kebab: `data-plane`, `new-api-version`, `package-name-<lang>-approved`
- hybrid: `APIStewardshipBoard-ReviewRequested`, `APIStewardship-ChangesRequested`

## What the flow actually needs

The data-plane review flow needs exactly two labels:

| Label | Applied by | Purpose |
|---|---|---|
| `data-plane-review-requested` | intake (auto) | signals a review is needed; triggers reviewer assignment; the board/queue filters on it |
| `data-plane-review-signoff` | reviewer pool (protected) | reviewer sign-off; clears the merge gate |

Everything else is either a duplicate of these or a state that GitHub already tracks natively (a "changes requested" PR review).

## Today's labels and what we do with each

| Label today | PRs ever labeled | Used by automation? | Plan |
|---|---|---|---|
| `APIStewardshipBoard-SignedOff` | 405 | yes (merge gate + who may apply it) | rename to `data-plane-review-signoff` |
| `APIStewardshipBoard-ReviewRequested` | 60 | yes (assignment trigger) | rename to `data-plane-review-requested` |
| `data-plane-api-review-needed` | 0 | yes (review agent trigger) | replace with `data-plane-review-requested` (one trigger, not two) |
| `offline-review-requested` | 37 | no | drop; fold into `data-plane-review-requested` (redundant "wants review" marker; reviews are offline now) |
| `APIStewardship-ChangesRequested` | 9 | no | drop; use the native GitHub "changes requested" review state |
| `APIStewardshipBoard-InDesign` | 1 | no | drop (effectively unused) |

## The protected sign-off label

`data-plane-review-signoff` (today `APIStewardshipBoard-SignedOff`) is a protected label: only an authorized group of reviewers can apply it, and anyone else's attempt is removed automatically. This is what stops a service team from self-signing-off to clear the merge
[...]

Issue #45494, opened 2026-08-13
Consolidate ReviewRequested-clearing on sign-off into the trusted central label flow


samvaity:
### Background

Today the data-plane assignment workflow
(`.github/workflows/src/data-plane-review/assign-reviewers.js`,
`completeReviewOnSignOff`) removes the fulfilled
`APIStewardshipBoard-ReviewRequested` label when
`APIStewardshipBoard-SignedOff` is applied. Because that workflow runs as a
separate job racing the protected-labels enforcement job, it has to
**re-verify the signer's authorization itself** (`loadAuthorizedSigners`)
before clearing the request label. If it did not, an unauthorized
`SignedOff` (which protected-labels reverts using `GITHUB_TOKEN`, a removal
that does not re-trigger workflows) could still cause `ReviewRequested` to be
cleared and never restored, potentially unblocking the merge gate in the
`ReviewRequested`-only case.

That means the same authorization list is resolved in two places:
- `check-label.js` (protected-labels enforcement), and
- `assign-reviewers.js` (`loadAuthorizedSigners`) just to guard this side-effect.

### Proposal

Move the "clear the fulfilled `ReviewRequested` label" action into the trusted,
centralized label-reconciliation flow, matching how ARM already does it in
`.github/workflows/src/summarize-checks/labelling.js` (see the `ARMSignedOff`
handling around lines ~421-437, where applying the sign-off label removes the
now-satisfied request/feedback labels in a single trusted computation).

Doing so:
- removes the duplicate authorization check from `assign-reviewers.js`
  (`loadAuthorizedSigners`, `loadConfig`, `extractLogins`, `includesLogin` can go),
- eliminates the cross-workflow race and the "lost signal" hazard entirely,
- gives a single authorization source of truth.

### Chosen approach: reconcile from current state, do not react to the apply event

The key design point: the central flow must **reconcile the label from the
current, durable label state**, not react to the `SignedOff`-applied event.

Concretely, the rule becomes:

> IF `APIStewardshipBoard-SignedOff` is present on the PR now, THEN
> `APIStewardshipBoard-ReviewRequested` should be absent.

evaluated on every `labelling.js` run against the labels that are actually on
the PR at that moment (exactly like the ARM `ARMSignedOff` reconciliation).

Why this removes the need for any authorization check:
- An unauthorized `SignedOff` is reverted by protected-labels, so by the time
  reconciliation reads the PR it simply is **not present** -> `ReviewRequested`
[...]

PR #45724, opened 2026-08-21
Auto-apply + atomic flip to data-plane-review-* labels (#45521)

samvaity:
Implements #45521 (part of #45437). One atomic PR that moves data-plane stewardship review onto the new `data-plane-review-requested` / `data-plane-review-signoff` labels, so the merge gate never stops enforcing sign-off mid-flip.

## What changes
| Piece | Before | After |
|---|---|---|
| Intake label | manual `APIStewardshipBoard-ReviewRequested` | auto-applied `data-plane-review-requested` on new data-plane API versions |
| Merge gate | old label + `data-plane`/`new-api-version` fallback | requires `data-plane-review-signoff` when the request label is present |
| Requesting the reviewer team | fired only on the manual label | also fires on the auto-applied label (trigger fix below) |
| Sign-off protection | `APIStewardshipBoard-SignedOff` | `data-plane-review-signoff` (same roster) |

Reviewer selection is unchanged: the workflow requests the `@Azure/azure-data-plane-api-reviewers` team, and GitHub's team code-review auto-assignment picks the individual member. This PR only changes what triggers that team request.

## The trigger fix
The intake label is auto-applied by Summarize Checks with the default `GITHUB_TOKEN`, and labels set by that token do not re-trigger `labeled` workflows. So the team request now also runs on `workflow_run` of "Summarize Checks" and reconciles from the PR's live labels. Mirrors the existing `arm-auto-signoff-status.yaml` pattern.

## Not in this PR (follow-ups once this merges)
- Delete the old `APIStewardshipBoard-ReviewRequested` and `APIStewardshipBoard-SignedOff` labels, once nothing references them anymore.
- Review-agent trigger (#45450): on hold for seeding.

## Validation
45 tests passing; lint + prettier clean.

### DIFF
--- a/.github/workflows/data-plane-review-assignment.yaml
+++ b/.github/workflows/data-plane-review-assignment.yaml
@@ -5,12 +5,13 @@
 # picks the individual member; this workflow only requests the team.
 #
 # Two triggers, because the `data-plane-review-requested` label can arrive two ways:
-#   - `pull_request_target: labeled` — manual label, plus the sign-off that clears it.
+#   - `pull_request_target: labeled` — the label applied manually.
 #   - `workflow_run` on "Summarize Checks" — the auto-applied label. It is applied with the
 #     default GITHUB_TOKEN, which does not re-trigger `labeled` workflows, so this path
 #     reconciles the request from the PR's live labels instead.
 #
-# `unlabeled` is not handled: native delegation owns the individual reviewer.
+# Sign-off and `unlabeled` are not handled: the durable `data-plane-review-signoff` label
+# gates merge on its own, and native delegation owns the individual reviewer.
 name: "Data-Plane API Review - Assign Reviewers"
 
 on:
@@ -30,16 +31,9 @@ permissions:
   contents: read
   pull-requests: write
 
-# Serialize runs per PR so racing events do not interleave. `pull_request_target` keys on the
-# PR number directly; `workflow_run` keys on the triggering run's head SHA (the PR number is
-# only resolved inside the job). Do not cancel in-progress runs: a cancelled run could drop a
-# team request mid-flight.
-concurrency:
-  group: >-
-    data-plane-review-assignment-${{
-      github.event.pull_request.number || github.event.workflow_run.head_sha
-    }}
-  cancel-in-progress: false
+# No `concurrency` block: the job is idempotent, like the other workflow_run label-reaction
+# workflows here. A PR-scoped group isn't possible on the workflow_run path anyway
+# (workflow_run.head_sha is the base tip, shared across open PRs).
 
 jobs:
   assign-reviewers:
--- a/.github/workflows/data-plane-review-assignment.yaml
+++ b/.github/workflows/data-plane-review-assignment.yaml
@@ -1,12 +1,20 @@
 # Data-Plane API Review - Assign Reviewers
 #
-# When the `APIStewardshipBoard-ReviewRequested` label is applied to a data-plane PR,
-# request review from the `azure-data-plane-api-reviewers` team. The team has GitHub
-# code-review auto-assignment enabled, so GitHub delegates the request to one team
-# member using the team's routing algorithm. Removing the label is intentionally not
-# handled: native delegation owns the individual reviewer, so the workflow never removes
-# a reviewer it did not pick, and it does not subscribe to `unlabeled` events. Applying
-# `APIStewardshipBoard-SignedOff` clears the request label.
+# Requests review from the `azure-data-plane-api-reviewers` team when a data-plane PR needs
+# stewardship review. The team has GitHub code-review auto-assignment enabled, so GitHub
+# delegates the request to one team member using the team's routing algorithm.
+#
+# Two triggers, because the `data-plane-review-requested` intake label can arrive two ways:
+#   - `pull_request_target: labeled` handles labels applied by a human (manual request) and
+#     the `data-plane-review-signoff` sign-off that clears the request label.
+#   - `workflow_run` on "Summarize Checks" handles the auto-applied intake label: Summarize
+#     Checks applies it with the default GITHUB_TOKEN, and labels applied by that token do
+#     not re-trigger `labeled` workflows, so this reconciles the reviewer request from the
+#     PR's live labels after Summarize Checks completes.
+#
+# Removing the label is intentionally not handled: native delegation owns the individual
+# reviewer, so the workflow never removes a reviewer it did not pick, and it does not
+# subscribe to `unlabeled` events.
 name: "Data-Plane API Review - Assign Reviewers"
 
 on:
@@ -15,21 +23,40 @@ on:
       - labeled
     branches:
       - main
+  workflow_run:
+    workflows:
+      - "Summarize Checks"
+    types:
+      - completed
 
 permissions:
+  actions: read # extractInputs reads issue-number/head-sha handoff artifacts on workflow_run
   contents: read
   pull-requests: write
 
-# Serialize runs per PR so racing label events do not interleave. Do not cancel
-# in-progress runs: a cancelled run could drop a team request mid-flight.
+# Serialize runs per PR so racing events do not interleave. `pull_request_target` keys on the
+# PR number directly; `workflow_run` keys on the triggering run's head SHA (the PR number is
+# only resolved inside the job). Do not cancel in-progress runs: a cancelled run could drop a
+# team request mid-flight.
 concurrency:
-  group: data-plane-review-assignment-${{ github.event.pull_request.number }}
+  group: >-
+    data-plane-review-assignment-${{
+      github.event.pull_request.number || github.event.workflow_run.head_sha
+    }}
   cancel-in-progress: false
 
 jobs:
   assign-reviewers:
     name: Assign Data-Plane Review Team
-    runs-on: ubuntu-24.04
+    # For workflow_run, skip runs of Summarize Checks that did not succeed: a failed or
+    # cancelled run may not have applied labels, and there is nothing to reconcile.
+    if: >-
+      ${{ github.event_name == 'pull_request_target' ||
+          github.event.workflow_run.conclusion == 'success' }}
+    # ubuntu-slim: this job only calls the GitHub REST API via github-script (no build), and
+    # runs on every Summarize Checks completion, so keep the runner minimal. Matches the other
+    # label-reaction workflows (arm-auto-signoff-status, update-labels).
+    runs-on: ubuntu-slim
     steps:
       # IMPORTANT: This workflow uses pull_request_target, which runs with repo write
       # permissions. Only base-branch code is checked out below, and no PR-head code is
--- a/.github/workflows/data-plane-review-assignment.yaml
+++ b/.github/workflows/data-plane-review-assignment.yaml
@@ -1,20 +1,16 @@
 # Data-Plane API Review - Assign Reviewers
 #
-# Requests review from the `azure-data-plane-api-reviewers` team when a data-plane PR needs
+# Requests review from the `azure-data-plane-api-reviewers` team on data-plane PRs that need
 # stewardship review. The team has GitHub code-review auto-assignment enabled, so GitHub
-# delegates the request to one team member using the team's routing algorithm.
+# picks the individual member; this workflow only requests the team.
 #
-# Two triggers, because the `data-plane-review-requested` intake label can arrive two ways:
-#   - `pull_request_target: labeled` handles labels applied by a human (manual request) and
-#     the `data-plane-review-signoff` sign-off that clears the request label.
-#   - `workflow_run` on "Summarize Checks" handles the auto-applied intake label: Summarize
-#     Checks applies it with the default GITHUB_TOKEN, and labels applied by that token do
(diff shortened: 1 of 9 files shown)


samvaity:
Addressed the auto-apply / assignment gap (review comment 1) with a `workflow_run` handoff, following the established house pattern (`arm-auto-signoff-status.yaml`).

**Why it was needed:** `data-plane-review-requested` is auto-applied by *Summarize Checks* using the default `GITHUB_TOKEN`. Labels applied by that token do not re-trigger `labeled` workflows (GitHub's loop-prevention rule), so the assignment workflow only ever fired for *manually* applied labels.

**Fix:** the assignment workflow now reacts on both triggers, reading live PR state:
- `pull_request_target: labeled` — human label changes + sign-off (unchanged)
- `workflow_run` on "Summarize Checks" completed — the auto-applied label

Supporting changes:
- *Summarize Checks* now publishes `issue-number` / `head-sha` handoff artifacts (via the existing `add-empty-artifact` action) so the consumer can resolve the PR via `extractInputs` — required on the nested `workflow_run` run that first applies the label.
- `assignFromWorkflowRun` requests the team only when the PR is open, non-draft, still carries the intake label, is not signed off, and the team is not already requested (all idempotent).
- Runner set to `ubuntu-slim` + success-only job guard to keep this high-frequency reaction cheap, matching `arm-auto-signoff-status` / `update-labels`.

6 new tests cover the `workflow_run` path; full suite + lint green.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -550,6 +569,7 @@ export function processImpactAssessment(labelContext, impactAssessment) {
   }
 
   dataplaneLabel.applyStateChange(labelContext.toAdd, labelContext.toRemove);
+  dataPlaneReviewRequestedLabel.applyStateChange(labelContext.toAdd, labelContext.toRemove);
```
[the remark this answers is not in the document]
samvaity: Fixed in fb186db76c8. The assignment workflow now also triggers on `workflow_run` of "Summarize Checks" (alongside `pull_request_target: labeled`) and reconciles from live PR state, so the auto-applied label reaches assignment. Summarize Checks publishes `issue-number`/`head-sha` handoff artifacts for PR resolution. Mirrors the house pattern in `arm-auto-signoff-status.yaml`. 6 new tests cover the path.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -533,10 +544,18 @@ export function processImpactAssessment(labelContext, impactAssessment) {
...
+
+      // For data-plane PRs, a new API version requires API stewardship review. Apply the
+      // intake label unless the review has already been signed off: assign-reviewers.js
+      // clears this label on sign-off, and re-adding it here would re-request the reviewer
+      // team. See rulesPri0dataPlane below.
+      if (
+        impactAssessment.dataPlaneRequired &&
+        !labelContext.present.has("data-plane-review-signoff")
+      ) {
+        dataPlaneReviewRequestedLabel.shouldBePresent = true;
```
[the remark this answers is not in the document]
samvaity: Fixed in 19951fdc8b9. The intake label is now pure content classification. `dataPlaneReviewRequestedLabel.shouldBePresent` is set solely from `impactAssessment.isNewApiVersion && dataPlaneRequired` and never reads the protected `data-plane-review-signoff` label, so presence of the protected label can neither suppress nor remove the request label here. The merge gate (`rulesPri0dataPlane`) still requires sign-off whenever the request label is present, and only assign-reviewers.js clears it, on an authorization-verified, serialized sign-off. Both request states are covered by regression tests.

.github/workflows/data-plane-review-assignment.yaml
```diff
@@ -15,21 +19,38 @@ on:
...
-# in-progress runs: a cancelled run could drop a team request mid-flight.
+# Serialize runs per PR so racing events do not interleave. `pull_request_target` keys on the
+# PR number directly; `workflow_run` keys on the triggering run's head SHA (the PR number is
+# only resolved inside the job). Do not cancel in-progress runs: a cancelled run could drop a
+# team request mid-flight.
 concurrency:
-  group: data-plane-review-assignment-${{ github.event.pull_request.number }}
+  group: >-
+    data-plane-review-assignment-${{
+      github.event.pull_request.number || github.event.workflow_run.head_sha
```
xirzec: `workflow_run.head_sha` is the SHA of the Summarize Checks workflow, which runs from the default branch, rather than the PR head. Unrelated PRs therefore share this concurrency group. GitHub keeps at most one running and one pending run per group and replaces older pending runs even when `cancel-in-progress` is false, so activity on one PR can drop another PR's reviewer-assignment run. This needs a PR-scoped key, or the workflow-level concurrency should be removed in favor of idempotency after resolving the PR number.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -550,6 +567,7 @@ export function processImpactAssessment(labelContext, impactAssessment) {
   }
 
   dataplaneLabel.applyStateChange(labelContext.toAdd, labelContext.toRemove);
+  dataPlaneReviewRequestedLabel.applyStateChange(labelContext.toAdd, labelContext.toRemove);
```
xirzec, line 570: This reconciliation removes an existing `data-plane-review-requested` whenever the impact assessment is not a non-draft new data-plane API version. That conflicts with the documented manual-label path: applying the label itself triggers Summarize Checks, which can immediately remove it. Because the merge gate now keys only on this label, a manual stewardship request can become ungated. Please preserve manual requests until verified sign-off, or distinguish machine-derived requests from manual ones.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -533,10 +544,16 @@ export function processImpactAssessment(labelContext, impactAssessment) {
...
-      // See requiredLabelsRules.ts.
       newApiVersionLabel.shouldBePresent = true;
+
+      // Data-plane new API versions need stewardship review. Classify purely from PR
+      // content; never read the protected data-plane-review-signoff label here (its presence
+      // isn't proof of sign-off, and reading it would be racy). The gate requires sign-off
+      // whenever this label is present, and only assign-reviewers.js clears it, on a
+      // verified sign-off.
+      if (impactAssessment.dataPlaneRequired) {
+        dataPlaneReviewRequestedLabel.shouldBePresent = true;
```
xirzec, line 555: This makes the request label non-durable after sign-off. `assign-reviewers.js` removes it after verifying the signer, but the sign-off event also starts Summarize Checks, and this branch sets it back to true without representing the completed state. The race, or any later summary run, can re-add the request label, putting an approved PR back in the review queue. The state machine needs a durable verified-completion signal that reconciliation respects.

.github/workflows/src/data-plane-review/assign-reviewers.js
```diff
@@ -131,29 +137,90 @@ export default async function assignReviewers({ github, context, core }) {
...
  */
-async function requestReviewerTeam({ github, core, owner, repo, prNumber, payload }) {
+async function requestReviewerTeam({ github, core, owner, repo, prNumber, requestedTeams }) {
   // requested_teams lists teams whose request is still pending (before delegation swaps the
   // team for a member). If our team is already pending, another run already requested it.
-  const requestedTeams = /** @type {{ slug?: string }[]} */ (
-    /** @type {unknown} */ (payload.pull_request.requested_teams ?? [])
-  );
-  const alreadyRequested = requestedTeams.some(
+  const alreadyRequested = (requestedTeams ?? []).some(
```
xirzec: This only deduplicates while the team itself remains in `requested_teams`. GitHub team auto-assignment replaces the team request with an individual, so subsequent successful Summarize Checks runs see no matching team and call `requestReviewers` again. That can repeatedly notify or assign reviewers. Please track the completed team handoff durably, or otherwise recognize the already-assigned individual or review state.

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -823,19 +841,17 @@ const rulesPri0dataPlane = [
...
+  // data-plane-review-requested is auto-applied in processImpactAssessment above whenever a
+  // data-plane PR introduces a new API version, so the previous data-plane + new-api-version
+  // fallback prerequisite is no longer needed.
   {
     precedence: 0,
-    anyPrerequisiteLabels: ["APIStewardshipBoard-ReviewRequested"],
-    allPrerequisiteLabels: ["data-plane", "new-api-version"],
-    anyRequiredLabels: ["APIStewardshipBoard-SignedOff"],
+    anyPrerequisiteLabels: ["data-plane-review-requested"],
+    anyRequiredLabels: ["data-plane-review-signoff"],
```
xirzec, line 856: The flip stops accepting old sign-offs without migrating active PRs or temporarily accepting both labels. For example, #43941 is still open with `data-plane`, `new-api-version`, and `APIStewardshipBoard-SignedOff` only; its next reconciliation will add the new request label and this rule will block it until the new sign-off is manually reapplied. Please include a migration or dual-label transition in the rollout.

.github/workflows/src/summarize-checks/labelling.js
samvaity: Good catch. Let me do that.

.github/workflows/data-plane-review-assignment.yaml
samvaity: Good catch. Dropped the concurrency block and went with idempotency, which matches the other workflow_run workflows here (none of them use concurrency either). Fixed in 268d5334d.

.github/workflows/src/summarize-checks/labelling.js
samvaity: Done in 268d5334d. The request label is add-only now, so a manual request sticks and reconciliation never strips it.

.github/workflows/src/summarize-checks/labelling.js
samvaity: Fixed in 268d5334d by not clearing the request label on sign-off at all. The protected signoff label is the durable terminal state and satisfies the gate on its own, so there's no remove/re-add race. Bonus: that let me drop the in-workflow auth duplication (#45494).

.github/workflows/src/data-plane-review/assign-reviewers.js
samvaity: Right, requested_teams empties once GitHub delegates the team to a person. Switched to the durable review_requested timeline event keyed to our team, so we request exactly once no matter how many summarize runs fire. 268d5334d.

.github/workflows/src/summarize-checks/labelling.js
samvaity: Added a dual-accept in 268d5334d: the gate now takes data-plane-review-signoff or the legacy APIStewardshipBoard-SignedOff, so in-flight PRs like #43941 aren't re-blocked. The legacy acceptance gets removed once no open PR carries it (tracked on #45521).

.github/workflows/src/summarize-checks/labelling.js
```diff
@@ -823,19 +841,18 @@ const rulesPri0dataPlane = [
...
+  // data-plane-review-requested is auto-applied in processImpactAssessment above, so the old
+  // data-plane + new-api-version fallback is no longer needed. The legacy signoff is accepted
+  // as a transition bridge; remove it once no open PR carries it (tracked in #45521).
   {
     precedence: 0,
-    anyPrerequisiteLabels: ["APIStewardshipBoard-ReviewRequested"],
-    allPrerequisiteLabels: ["data-plane", "new-api-version"],
-    anyRequiredLabels: ["APIStewardshipBoard-SignedOff"],
+    anyPrerequisiteLabels: ["data-plane-review-requested"],
+    anyRequiredLabels: ["data-plane-review-signoff", "APIStewardshipBoard-SignedOff"],
```
xirzec: The legacy transition label is accepted by the merge gate here, but this PR removes `APIStewardshipBoard-SignedOff` from `.github/protected-labels.yml`. The protected-label workflow skips labels that are absent from that configuration, so any user who can apply labels could use the legacy label to satisfy this rule without an authorized stewardship reviewer. Please keep both sign-off labels protected for the migration window, or migrate existing approvals and stop accepting the legacy label.

samvaity: Good catch, went with the clean switch. Dropped `APIStewardshipBoard-SignedOff` from the gate in 1ea115f7114 so it only reads the protected `data-plane-review-signoff`. No dual-accept bridge, since an unprotected legacy label the gate still trusted is exactly the bypass you called out.

Strategy for the PRs currently carrying the legacy sign-off: only #43941 is a real in-flight one. The [other ](https://github.com/Azure/azure-rest-api-specs/issues?q=state%3Aopen%20label%3AAPIStewardshipBoard-SignedOff) are abandoned 2021-era PRs, so ignore/close. For #43941 an authorized signer re-stamps it with `data-plane-review-signoff` (it was already validly signed off, so this is just moving the stamp to the protected label). Nothing gets ungated and no unprotected label is trusted. Tracked on #45521.

tjprescott (approved):

PR #45736, opened 2026-08-22
Apply build-failed label to generated SDK PRs

JoshLove-msft:
## Summary
- apply the `BuildFailedLabel` emitted by `spec-gen-sdk-runner` when creating a generated SDK pull request
- do not apply the label when a rerun updates an existing pull request
- initialize the optional variable so successful builds and non-opted-in languages remain unlabeled
- add regression coverage for the creation-only wiring

## Root cause

#43760 implemented and tested the producer, but `archetype-spec-gen-sdk.yml` never consumed the emitted pipeline variable. As a result, release-planner SDK PRs were created without `auto-sdk-build-fix`.

The pipeline now passes `BuildFailedLabel` through `Submit-PullRequest.ps1`'s `-PRLabels` argument. That script applies `PRLabels` only in its new-PR branch; its existing-PR branch updates only the title and body. Therefore a rerun cannot add this eligibility label to an existing PR.

## Validation

- `npm run check` in `eng/tools/spec-gen-sdk-runner` (233 tests)
- parsed `archetype-spec-gen-sdk.yml` with the repository's `yaml` dependency

Related: Azure/azure-sdk-for-net#59651

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/test/pipeline-labeling.test.ts
+++ b/eng/tools/spec-gen-sdk-runner/test/pipeline-labeling.test.ts
@@ -34,27 +34,18 @@ describe("SDK PR build-failed labeling pipeline", () => {
     expect(initializationIndex).toBeLessThan(generationIndex);
   });
 
-  test("applies the emitted label to the submitted SDK pull request", () => {
-    const task = getTask("Add build-failed label");
+  test("passes the emitted label when creating the SDK pull request", () => {
+    const task = getTask("Create pull request");
 
-    expect(task).toContain("filePath: $(SdkRepoDirectory)/eng/common/scripts/Add-IssueLabels.ps1");
-    expect(task).toContain('-IssueNumber "$(Submitted.PullRequest.Number)"');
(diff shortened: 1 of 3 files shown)


jorgerangel-msft (approved):

raych1 (approved):

eng/tools/spec-gen-sdk-runner/test/pipeline-labeling.test.ts
```diff
@@ -0,0 +1,51 @@
+import fs from "node:fs";
```
raych1, line 1: Given the pipeline template tests are added, you will need to include pipeline template path in the test workflow trigger.

JoshLove-msft: Added `eng/pipelines/templates/stages/archetype-spec-gen-sdk.yml` to the `spec-gen-sdk-runner-test` pull-request path filter, so any change covered by `pipeline-labeling.test.ts` now triggers that workflow.

--generated by Copilot

raych1 (approved):

PR #45742, opened 2026-08-24
[Mission] Replace unnamed unions with named unions

HarveyLink:

## Purpose of this PR

- [ ] New resource provider.
- [x] Other, please clarify:
  - Replace eight unnamed extensible unions in the Microsoft.Mission TypeSpec with reusable named unions while preserving their wire values, optionality, defaults, and version behavior.

## Summary

- Define eight named unions centrally in `model.tsp` using their original specification names.
- Reference the named unions directly from governed-service, maintenance-mode, principal, and approval-callback models.
- Preserve the existing C# SDK names with C#-scoped `@@clientName` mappings in `client.tsp`.
- Remove the equivalent client-only replacement unions and `@@alternateType` mappings.
- Regenerate OpenAPI for all five existing preview API versions.

The generated Swagger changes structurally from inline schemas to named definitions referenced with `$ref`. The accepted values and runtime payload contract remain unchanged. Swagger definition names use the original TypeSpec names; only C# clients receive the `VirtualEnclave*` names.

## Due diligence checklist

- [ ] I have reviewed the Resource Provider guidelines, ARM resource provider contract, and REST guidelines.
- [ ] A release plan has been created.

## Validation

- Azure TypeSpec validation passed.
- `npx tsp compile .` passed for all versions.
- Generated OpenAPI was semantically compared across all five preview versions; literals, extensibility, requiredness, defaults, and version visibility are unchanged.
- `git diff --check` passed.

Three pre-existing unnamed-union warnings remain outside these eight types: two `approvalStatus` expressions and one community expression.


specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp
```diff
@@ -38,6 +38,161 @@ using TypeSpec.Versioning;
...
+
+  @doc("Enforce the policy.")
+  Enforce: "Enforce",
+
+  @doc("Do not apply a policy action.")
+  None: "None",
+}
+
+@doc("The maintenance mode for a virtual enclave.")
+union MaintenanceModeConfigurationModelMode {
```
psah434, line 84: **[NEW] 🔵 Suggestion** `specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp` - line 84 - Enum type-name ergonomics (advisory; no linter/rule ID).

**Classification reasoning:** [NEW] - these enums were previously unnamed inline unions with no `x-ms-enum.name`, so no prior enum type name existed for Python/Go/JS/TS SDKs. The named types are introduced in this PR.

**Issue:** Several new union names emitted as swagger `x-ms-enum.name` (and thus the enum type name in Python/Go/JS/TS SDKs) leak internal type/property-bag artifacts: `MaintenanceModeConfigurationModelMode` (line 84) and `MaintenanceModeConfigurationModelJustification` (line 104) embed `Model`; `CommunityPropertiesPolicyOverride` (line 186) embeds `Properties` and appears in the stable 2026-04-01 swagger. These diverge from the cleaner C# names applied via `@@clientName` (e.g., `VirtualEnclaveMaintenanceModeConfigurationMode`, `CommunityPolicyOverride`). Since these enums were previously unnamed, you can pick clean names uniformly now, before GA SDKs lock them in.

**Suggested fix:** Rename the unions to drop the internal artifacts, or apply `@@clientName` across all target emitters (not just `"csharp"`).

specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp
```diff
@@ -38,6 +38,161 @@ using TypeSpec.Versioning;
...
+
+  @doc("Delete the resource.")
+  Delete: "Delete",
+
+  @doc("Update the resource.")
+  Update: "Update",
+}
+
+@doc("The approval status for an approval callback request.")
+union ApprovalCallbackRequestApprovalStatus {
```
psah434, line 163: **[NEW] 🔵 Suggestion** `specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp` - line 163 - Duplicate identical enums (advisory).

**Classification reasoning:** [NEW] - both unions are introduced as named types in this PR (previously unnamed inline unions).

**Issue:** `ApprovalCallbackRequestApprovalStatus` (line 163) and `ApprovalActionRequestApprovalStatus` (line 174) are two separately-named unions with identical members (`Approved`/`Rejected`) and identical descriptions.

**Suggested fix:** Consider a single shared union referenced by both models. Do **not** name it `ApprovalStatus` - a semantically different `ApprovalStatus` enum (`Approved`|`Deleted`|`Expired`|`Pending`|`Rejected`) already exists in the spec. Use a non-colliding name such as `ApprovalDecision` or `ApprovalRequestStatus`.

psah434 (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`0d413e8`](https://github.com/Azure/azure-rest-api-specs/pull/45742/commits/0d413e8170edf1ae6d9a2a168e28cb10aaf760eb). See inline comments for findings 1-2.

specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp
ArcturusZhang: despite these unions are previously unnamed unions, the SDK will always generate them with a name produced by TCGC.
the names we chose here are the same as those TCGC generated names, therefore there should not be breaking changes for other languages' SDKs.
csharp needs those renames because the TCGC geneated names are not good enough for csharp because they are just names that TCGC puts together the model name and the property name.
Also if there would be a breaking change for SDKs, we would have those "BreakingChange-Language" labels, but currently there is none.

specification/mission/resource-manager/Microsoft.Mission/Mission/model.tsp
ArcturusZhang: this is something we no longer could change.
TCGC will model each unnamed unions independently by its enclosing type name and property name, therefore in the released SDK, we already shipped with two different enums with identical values. Merging them will be a breaking change.
On the other hand, this might help the service team in the future, because now they have identical values, we cannot say they would not diverge in the future :)

PR #45786, opened 2026-08-25
[Relay] Added Missing namespace properties in 2026-07-01-preview

schaudhari6254888:
# Choose a PR Template

### DIFF
--- a/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
+++ b/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
@@ -413,6 +413,18 @@ model RelayNamespaceProperties {
    */
   @added(Versions.v2026_07_01_preview)
   clusterArmId?: string;
+
+  /**
+   * This property disables SAS authentication for the Relay namespace.
+   */
+  @added(Versions.v2026_07_01_preview)
+  disableLocalAuth?: boolean;
+
+  /**
+   * Alternate name for namespace
+   */
+  @added(Versions.v2026_07_01_preview)
+  alternateName?: string;
 }
 
 /**
--- a/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
+++ b/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
@@ -407,6 +407,12 @@ model RelayNamespaceProperties {
    */
   @added(Versions.v2026_01_01)
   minimumTlsVersion?: TlsVersion;
+
+  /**
+   * Cluster ARM ID of the Namespace.
+   */
+  @added(Versions.v2026_07_01_preview)
+  clusterArmId?: string;
 }
 
 /**
--- a/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
+++ b/specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
@@ -419,12 +419,6 @@ model RelayNamespaceProperties {
    */
   @added(Versions.v2026_07_01_preview)
   disableLocalAuth?: boolean;
-
-  /**
-   * Alternate name for namespace
-   */
-  @added(Versions.v2026_07_01_preview)
-  alternateName?: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


KPIyer:
Approved as benign. This changes the current API which is in preview by adding an optional input param.

dsouzaarun (approved):

specification/relay/resource-manager/Microsoft.Relay/Relay/models.tsp
```diff
@@ -407,6 +407,12 @@ model RelayNamespaceProperties {
    */
   @added(Versions.v2026_01_01)
   minimumTlsVersion?: TlsVersion;
+
+  /**
+   * Cluster ARM ID of the Namespace.
+   */
+  @added(Versions.v2026_07_01_preview)
+  clusterArmId?: string;
```
psah434, line 415: **[NEW] 🟠 Warning** **[[TSP-ARM-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#63-use-armresourceidentifier-for-resource-references-tsp-arm-resource-id)]** `models.tsp` - line 415 - `clusterArmId` uses a plain `string` for an ARM resource identifier.

**Classification reasoning:** Introduced in this PR — the property carries `@added(Versions.v2026_07_01_preview)` and did not exist in the previous version.

Section 6.3 of the TypeSpec review rules: "Do not use plain `string` for ARM resource ID properties — `armResourceIdentifier` enables ARM validation and rich SDK typing." The generated `relay.json` emits a bare `"type": "string"` with no `"format": "arm-id"`. The sibling `RelayNamespaceReference` model ("A reference to a Relay namespace assigned to a cluster") confirms this is a genuine ARM reference. Best fixed now while the property is preview-only.

**Suggested fix:**

```tsp
clusterArmId?: armResourceIdentifier<[{ type: "Microsoft.Relay/clusters" }]>;
```

Replace the `type` constraint with the correct cluster RP type if it differs.

psah434, line 415: **[NEW] 🟠 Warning** **[[Naming — Azure terminology](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/naming-conventions.md#azure-terminology-in-client-facing-text)]** `models.tsp` - line 415 - property name `clusterArmId` embeds "ARM".

**Classification reasoning:** Introduced in this PR (`@added(Versions.v2026_07_01_preview)`).

The naming-conventions terminology table recommends property names without "ARM" (e.g., `armResourceId` → `monitorAccountResourceId`). This is a naming-hygiene recommendation rather than a literal prefix match, but since the property is preview-only, renaming now avoids a breaking rename after GA. Keep the `Id` suffix (required for resource-reference properties).

**Suggested fix:** rename to `clusterId` (or `clusterResourceId`).

psah434, line 415: **[NEW] 🔵 Suggestion** **[Property mutability]** `models.tsp` - line 415 - `clusterArmId` is fully mutable (read/create/update); confirm this matches service behavior.

**Classification reasoning:** Introduced in this PR (`@added(Versions.v2026_07_01_preview)`).

`clusterArmId` has no `@visibility` decorator, so it is read/create/update. For a cluster-assignment reference this is usually **not** correct — the assignment is typically fixed at creation. Please confirm which applies:

- Service reports the assigned cluster; caller can't set it → `@visibility(Lifecycle.Read)`.
- Caller assigns the cluster at create, fixed thereafter → `@visibility(Lifecycle.Read, Lifecycle.Create)` (most likely).
- Assignment can be changed via update → leave mutable, and document the semantics.

Tightening mutability (mutable → create-only/read-only) is a breaking change after GA, while loosening is not — prefer the narrowest correct visibility now while the API is preview.

**Suggested fix:**

```tsp
clusterArmId?: string;
```

psah434 (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 1 iteration, converged) against commit [`4ec828e`](https://github.com/Azure/azure-rest-api-specs/pull/45786/commits/4ec828e8df92a09dd3590cea05843594bd1157a0). See inline comments for findings 1-3. No blocking issues.

PR #45796, opened 2026-08-25
Add GitHub Copilot harness for prompt agents

viswabalaji:
## Summary
- add a preview-gated `github_copilot_preview` runtime selector to persisted prompt agent definitions
- add shared top-level skill references for prompt agents with or without a harness
- configure Copilot built-ins through the prompt-agent-only `github_copilot_toolset_preview` toolset
- keep harness compute settings service-managed and preserve the existing `PromptAgentDefinition.tools` client shape
- regenerate the Foundry v1 and virtual public-preview OpenAPI documents

## Source design
- coreai-microsoft/foundrysdk_specs#298

## Validation
- `npx tsp compile . --warn-as-error`
- TypeSpec Prettier check
- `git diff --check`


specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -303,12 +306,18 @@ model AgentDefinition {
 model PromptAgentDefinition extends AgentDefinition {
   kind: AgentKind.prompt;
 
+  @doc("The managed runtime and agent loop used to execute this prompt agent.")
+  harness?: AgentHarness;
```
glecaros, line 352: add the extension indicating this is a preview property, same to the one below

viswabalaji: Added property-level x-ms-foundry-meta preview requirements for both harness (GitHubCopilot=V1Preview) and skills (Skills=V1Preview), and regenerated the OpenAPI output in ec6232a7f37.

glecaros (approved):

PR #45801, opened 2026-08-25
[Microsoft.AzureResilienceManagement] Model listReportDownloadUrl as an async action (2026-08-31-preview)

arpja:

## Summary

Changes `listReportDownloadUrl` on `Microsoft.AzureResilienceManagement/drills/drillRuns` from a
synchronous to a long-running (async) POST action in `2026-08-31-preview`. This is a fix for a
live `412 HttpResponsePayloadAPISpecValidationFailed` — the action is currently unusable.

### Why

`drills/drillRuns` is registered in RPaaS with `enableAsyncOperation: true` and a POST
concurrency policy of `SynchronizeBeginExtension`. As a result ProviderHub routes **every** POST
action on the resource type through its async pipeline and emits `202 Accepted` with
`Location` / `Azure-AsyncOperation` headers *before* the request is ever dispatched to the RP.

`listReportDownloadUrl` was the only `ArmResourceActionSync` in the provider, so its generated
swagger declared only `200` and `default`. RPaaS response validation therefore rejected the
platform's own `202`:

```
412 HttpResponsePayloadAPISpecValidationFailed
APIValidationEngine.TraceErrorCodes: INVALID_RESPONSE_CODE
  at preview/2026-08-31-preview/openapi.json (responses block of listReportDownloadUrl)
```

Confirmed from RPaaS `Traces` telemetry (`ShouldCreateResourceManagementJob: 'True' -
isResourceManagementJobNeededForConcurrencyControl: 'True'`) and independently by the ICM
troubleshooting analysis on incident 856928156. `enableAsyncOperation` is **resource-type
scoped with no per-action override**, so the only correct fix is to declare the action async so
the contract matches the platform behaviour.

### What changed

| Change | Detail |
| --- | --- |
| `ArmResourceActionSync` → `ArmResourceActionAsync` | Adds `202` with `Azure-AsyncOperation`, `Location`, `Retry-After`; `x-ms-long-running-operation: true`; `final-state-via: location` |
| `ServiceGroupParameters` → `ServiceGroupAndRequestHeadersParameters` | Required for the async parameter set, matching the sibling `generateReport` action |
| Removed `OptionalRequestBody = true` | Body is now required — see below |

The action's contract is now identical in shape to the sibling `generateReport` action on the
same resource type, which has always been async and has always worked.

**On the body becoming required:** the ProviderHub controller emitter does not honour
`OptionalRequestBody = true` — it generates an unconditional null check that throws. An optional
body therefore meant a spec-compliant caller who omitted it received a `500` rather than the
documented response. Declaring the body required moves that rejection to ARM as a clean `400`.
This preview version has no wire callers today, so there is no compatibility impact.

No changes to any other API version, resource type, or provider.

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Update existing version for a new feature. (This is applicable only when you are revising a private preview API version.)
  - [ ] Other, please clarify:

> Note: `2026-08-31-preview` is an unreleased preview version. This PR corrects the contract of an
> action introduced in that same version, before it has any callers.

## Due diligence checklist

- [x] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including [ARM resource provider contract](https://aka.ms/azurerpc) and [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md) (estimated time: 4 hours). I understand this is required before I can proceed to the diagram Step 2, "ARM API changes review", for this PR.
- [x] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created.

### DIFF
--- a/specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
+++ b/specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/drillRun.tsp
@@ -190,12 +190,14 @@ interface DrillRuns {
   @armResourceAction(DrillRun)
   @post
   @doc("This returns a short-lived, read-only URL to download the report for this Drill Run. The URL expires at the returned expiryTimestamp and grants access to that single report only.")
-  listReportDownloadUrl is ArmResourceActionSync<
+  listReportDownloadUrl is ArmResourceActionAsync<
     DrillRun,
     ListReportDownloadUrlRequest,
     ListReportDownloadUrlResponse,
-    ServiceGroupParameters,
-    OptionalRequestBody = true
+    ServiceGroupAndRequestHeadersParameters,
+    LroHeaders = ArmAsyncOperationHeader &
+      ArmLroLocationHeader &
+      Azure.Core.Foundations.RetryAfterHeader
   >;
 }
 
--- a/specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/examples/2026-08-31-preview/DrillRuns_ListReportDownloadUrl_MaximumSet_Gen.json
+++ b/specification/azureresiliencemanagement/resource-manager/Microsoft.AzureResilienceManagement/AzureResilienceManagement/examples/2026-08-31-preview/DrillRuns_ListReportDownloadUrl_MaximumSet_Gen.json
@@ -4,6 +4,7 @@
   "parameters": {
     "serviceGroupName": "sampleServiceGroupName",
     "api-version": "2026-08-31-preview",
+    "operation-id": "3f2b1c4d-5e6f-4a7b-8c9d-0e1f2a3b4c5d",
     "drillName": "drill1",
     "drillRunName": "ca92602e-53bf-43d2-ae62-d3fc940474b3",
     "body": {
@@ -11,6 +12,11 @@
     }
   },
   "responses": {
+    "202": {
+      "headers": {
+        "location": "https://contoso.com/operationstatus"
+      }
+    },
     "200": {
       "body": {
         "format": "Html",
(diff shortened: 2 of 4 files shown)


arpja:
## Versioning review context — why this same-version amendment carries no compatibility risk

Flagging context for the versioning / breaking-change review (`VersioningReviewRequired`), since the 7 OAD violations here look alarming in isolation but are all mechanical consequences of one correction.

### The action has never worked

`Microsoft.AzureResilienceManagement/drills/drillRuns` is registered in RPaaS with `enableAsyncOperation: true` and a POST concurrency policy of `SynchronizeBeginExtension`. ProviderHub therefore routes **every** POST action on the resource type through its async pipeline and emits `202 Accepted` with `Location` / `Azure-AsyncOperation` headers **before the request is ever dispatched to the RP**.

`listReportDownloadUrl` was modelled as `ArmResourceActionSync`, so its swagger declared only `200` and `default`. RPaaS response validation consequently rejected the platform's own `202`:

```
412 HttpResponsePayloadAPISpecValidationFailed
APIValidationEngine.TraceErrorCodes: INVALID_RESPONSE_CODE
  at preview/2026-08-31-preview/openapi.json (responses block of listReportDownloadUrl)
```

Production telemetry over the observation window: **3 of 3 calls returned 412, 0 succeeded.** There is no working behaviour for this change to break. Root cause independently confirmed by ICM incident 856928156.

### Why introducing a new API version is not a viable alternative

- `2026-08-31-preview` is the **default** API version for this resource type, so callers would continue to hit the broken contract; a new version would leave the defect in the version everyone resolves to.
- `enableAsyncOperation` is **resource-type scoped with no per-action override**, so async is the only shape this action can legally take. The sibling `generateReport` action on the same resource type has always been async and has always worked — this change makes the two structurally identical.
- The version is unreleased preview with no SDK shipped from it.

### The 7 violations all stem from that single correction

| Rule | Violation | Cause |
| --- | --- | --- |
| 1044 `XmsLongRunningOperationChanged` | sync → async | the fix itself |
| 1011 `AddingResponseCode` | added `202` | required by the async contract |
| 1010 `AddingRequiredParameter` | added required `operation-id` header | part of the async parameter set |
| 1025 `RequiredStatusChange` | body `false` → `true` | see below |
| 1042 `ChangedParameterOrder` x3 | `drillName`, `drillRunName`, `body` shifted one index | purely positional, caused by inserting the `operation-id` header |

The three `ChangedParameterOrder` entries are not semantic changes at all — no parameter was renamed, retyped, or removed; they merely shifted position because a header parameter was inserted ahead of them.

On **1025**: the ProviderHub controller emitter does not honour `OptionalRequestBody = true` — it generates an unconditional null check that throws. An optional body therefore meant a spec-compliant caller who omitted it received a `500` rather than the documented response. Making the body required moves that rejection to ARM as a clean `400`. There are no wire callers of this preview action today.

Happy to provide the RPaaS `Traces` telemetry or the ICM analysis if useful for the review.

mearvind (approved):

PR #45805, opened 2026-08-25
Adding "updated_at" field to Toolbox and $latest support.

pjoshi23:
# Choose a PR Template


specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
dargilco, line 13: Why are the above two statements needed? We usually see these kind of decorators in the client-specific files in the "sdk-" folders, due to SDK specific hand-written code. But not in the TypeSpec files that define the REST API spec.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -67,6 +67,32 @@ interface Toolboxes {
...
+      @body request: unknown;
+    },
+    {
+      /** The content type of the MCP response body. */
+      @header contentType: string;
+
+      /** The MCP response body. */
+      @body response: unknown;
+    }
+  >;
```
dargilco, line 94: Is this operation going directly to GA or does it need a PuPr period? If it needs PuPr period, then it needs a "Foundry-Features" preview http request header, and the Python emitted configured to meit the method on the ".beta.toolboxes" subclient.

pjoshi23: This is going directly to GA.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
pjoshi23: I didn't add these lines, but my guess is they make sure the named model is kept as a public input model during SDK generation, since it is consumed by handwritten .NET code.

dargilco: Let me reach out to the team and ask to move these two lines to the appropriate client.tsp file(s). I'll resolve this comment.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
dargilco: okay

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -67,6 +67,32 @@ interface Toolboxes {
...
+  invokeLatestToolboxMcp is FoundryDataPlaneOperation<
+    {
+      /** The name of the toolbox. */
+      @path name: string;
+
+      /** The content type of the MCP request body. */
+      @header contentType: string;
+
+      /** The MCP request body. */
+      @body request: unknown;
```
dargilco: I think this should be `Record<unknown>` instead of just `unknown`, per other examples for generic JSON request in out TypeSpec. same for `response` below. @glecaros ?

pjoshi23: Record<unknown> makes sense for the request. However, I think the response can also be in the form of text/event-stream, which I don't think Record<unkown> allows for.

dargilco: Thanks for the clarification. So can you change just the request to Record<unknown>?

pjoshi23: yup did that.

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -67,6 +67,32 @@ interface Toolboxes {
     ToolboxVersionObject
   >;
 
+  /** Invokes the latest version of the specified toolbox through its MCP endpoint. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
+  @summary("Invoke the latest toolbox version through MCP")
+  @operationId("invokeLatestToolboxMcp")
+  @post
+  @route("/toolboxes/{name}/latest/mcp")
```
glecaros: given that `GET /toolboxes/{name}` retrieves the latest version of the toolbox, would it be better to have this call sit directly on that route?
In terms of the naming, probably it would be a good idea to make the action clear, so there's no confusion about it operating on a sub resource (e.g. `:invoke` or `:invoke_mcp`). Also, as per our guidelines, action separator is `:`.

pjoshi23: makes sense, just made the change. thanks for the feedback!

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -67,6 +67,32 @@ interface Toolboxes {
     ToolboxVersionObject
   >;
 
+  /** Invokes the latest version of the specified toolbox through its MCP endpoint. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "not yet an Azure operation"
+  @summary("Invoke the latest toolbox version through MCP")
+  @operationId("invokeLatestToolboxMcp")
+  @post
+  @route("/toolboxes/{name}:invokeMcp")
```
glecaros: from our naming guidelines should be `:invoke_mcp`

glecaros (commented):
Aside from the naming comment, this seems ok. Need to resolve the conflics though

glecaros (approved):

PR #45806, opened 2026-08-25
TrafficManager: add stable 2026-09-01 API version

ransinha1989:
## Summary
- Adds a new stable API version ` 2026-09-01 ` for Microsoft.Network/TrafficManager, generated from TypeSpec.
- Promotes ` RecordType `/` recordType ` (previously preview-only in ` 2024-04-01-preview `) to GA in ` 2026-09-01 `.
- The previous preview version is frozen as ` 2025-01-01-preview ` (renamed from ` 2024-04-01-preview `, removed from the active TypeSpec ` Versions ` enum, OpenAPI/examples left in place) so existing consumers referencing ` api-version=2025-01-01-preview ` continue to work.
- Updated ` readme.md ` with a new ` package-2026-09 ` tag (now default) and a ` package-preview-2025-01 ` tag.

## Validation
- ` tsp compile . ` succeeds with no warnings.
- TypeSpec validation passes.


ransinha1989:
## Breaking Change(Cross-Version) analysis

Investigated the failing `Breaking Change(Cross-Version)` check (run: https://github.com/Azure/azure-rest-api-specs/actions/runs/32908325684).

**Root cause:** All 32 flagged violations are the same rule — `TypeFormatChanged` (rule [1023](https://github.com/Azure/openapi-diff/tree/master/docs/rules/1023.md)): *"The new version has a different format 'uuid' than the previous one ''."*

This is triggered on standard ARM parameters (e.g. `SubscriptionIdParameter`) because:
- Old stable `2022-04-01` uses `armCommonTypesVersion` **v3** (no format on `subscriptionId`).
- New `2025-01-01-preview`/`2026-09-01` use **v6**, which adds `format: uuid` to that same parameter.

This is an expected, well-known consequence of bumping common-types from v3 → v6, not something introduced by the API design changes in this PR.

**Non-breaking additive change:** The new `recordType` property only produced `Info`-level messages (`AddedPropertyInResponse` / `AddedOptionalProperty`), correctly not flagged as breaking.

**Next steps:** Per the PR's "Next Steps to Merge" checklist, this requires the standard breaking-change review process (aka.ms/brch) and ARM sign-off, since this PR targets `main` and ships as GA. Following up with the reviewers for approval.

sandipsh:
**[EXISTING]** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md#proactive-secret-detection-sec-secret-detect)]** `specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp` - lines 739-740 - `key` is a credential signal, the examples return a key-like value from GET, and the `secret-prop` suppression itself says to use `@secret`. GET responses must not expose secrets.

**Classification reasoning:** This exact model and suppression are present at base SHA `befd1687368facdf2e5f0878725c66107c2cd8d5` and in stable `2022-04-01`, so this is inherited rather than introduced by this PR.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the session SHA. Security-rule suppressions are not an acceptable substitute for secret handling.

**Suggested fix:** Remove the suppression, mark the value `@secret`, and do not return it from GET/PUT responses. If callers need to retrieve it, expose a POST `listKeys` action with granular RBAC.

sandipsh:
**[EXISTING]** **[[Section 6 - Null Values](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#null-values)]** `specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp` - line 122 - The description says `null` means disabled, but response properties should be omitted rather than returned as `null`, and the disabled example already uses the explicit `Disabled` value. This leaves clients with two representations for the same state.

**Classification reasoning:** The same description and schema are present at base SHA `befd1687368facdf2e5f0878725c66107c2cd8d5` and in stable `2022-04-01`.

**Suggested fix:** Remove the `null` statement and document `Disabled` as the disabled representation. Ensure the service omits the property if it has no value.

specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp
```diff
@@ -482,7 +482,7 @@ model ProfileProperties {
   /**
    * When record type is set, a traffic manager profile will allow only endpoints that match this type.
    */
-  @added(Versions.v2024_04_01_preview)
+  @added(Versions.v2026_09_01)
   recordType?: RecordType;
```
sandipsh, line 486: **[NEW] 🟠 Warning** **[[Section 6 - Null Values](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#null-values)]** `specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp` - line 486 - Could you confirm the runtime behavior when `recordType` is unset? The earlier review said it would be returned as `null`, while the current examples omit it and this property is optional. For a stable API, these need to agree: please have the service omit the property when unset and update this description to say so. Returning `null` would introduce response behavior the schema does not describe and gives clients inconsistent expectations.

**Classification reasoning:** Introduced into the stable API by this PR; `recordType` is absent from stable `2022-04-01`. This carries forward the sole unresolved human thread from predecessor PR https://github.com/Azure/azure-rest-api-specs-pr/pull/17695#discussion_r1560214308.

**Suggested fix:** Update the doc comment to state: `The allowed endpoint DNS record type. This property is omitted when no record type is configured.` Preserve the current optional type and `@added(Versions.v2026_09_01)` decorator.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`98a4670`](https://github.com/Azure/azure-rest-api-specs/pull/45806/commits/98a46704c6b364a2bfb714d5c7102446b4cb91b8). See inline comments for finding 2. Findings 1 and 3 concern lines unchanged in this PR and are posted as top-level PR comments rather than inline comments.

Approval labels observed: `BreakingChange-Approved-Benign`.

specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp
ransinha1989: Confirmed runtime behavior via live testing. I created a Traffic Manager profile without recordType set (subscription f36efc75-a504-4bb0-8701-25904dc9e086, using the already-implemented 2024-04-01-preview api-version) and checked both the PUT and GET responses:
```
"recordType": null
```

To match the above, we updated the recordType doc comment in models.tsp to clarify the behavior accourately.
```
"When record type is set, a traffic manager profile will allow only endpoints that match this type. If it is not set, traffic manager profile will allow adding all types of endpoints. It is returned as null when this is not set."
```

Also added a new example showing the above.

Please let us know if this is acceptable.

ransinha1989:
> **[EXISTING]** **[[SEC-SECRET-DETECT](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/secret-detection.md#proactive-secret-detection-sec-secret-detect)]** `specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp` - lines 739-740 - `key` is a credential signal, the examples return a key-like value from GET, and the `secret-prop` suppression itself says to use `@secret`. GET responses must not expose secrets.
> 
> **Classification reasoning:** This exact model and suppression are present at base SHA `befd1687368facdf2e5f0878725c66107c2cd8d5` and in stable `2022-04-01`, so this is inherited rather than introduced by this PR.
> 
> **Approval context:** No `Approved-TypeSpecSuppression` label was observed at the session SHA. Security-rule suppressions are not an acceptable substitute for secret handling.
> 
> **Suggested fix:** Remove the suppression, mark the value `@secret`, and do not return it from GET/PUT responses. If callers need to retrieve it, expose a POST `listKeys` action with granular RBAC.

Created WI to be handled ASAP:
https://dev.azure.com/msazure/One/_workitems/edit/39493961

ransinha1989:
> **[EXISTING]** **[[Section 6 - Null Values](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#null-values)]** `specification/trafficmanager/resource-manager/Microsoft.Network/TrafficManager/models.tsp` - line 122 - The description says `null` means disabled, but response properties should be omitted rather than returned as `null`, and the disabled example already uses the explicit `Disabled` value. This leaves clients with two representations for the same state.
> 
> **Classification reasoning:** The same description and schema are present at base SHA `befd1687368facdf2e5f0878725c66107c2cd8d5` and in stable `2022-04-01`.
> 
> **Suggested fix:** Remove the `null` statement and document `Disabled` as the disabled representation. Ensure the service omits the property if it has no value.

Created WI to be handled ASAP.
https://dev.azure.com/msazure/One/_workitems/edit/39493958

tundwed (approved):

PR #45812, opened 2026-08-26
Add Microsoft.GcpConnector stable API version 2027-01-01 (GCP connector GA)

alexmahonic:

> [!TIP]

## Summary of changes

Adds the **Microsoft.GcpConnector** TypeSpec projects to the public specs repo at stable api-version **2027-01-01** (GCP public cloud connector GA). These specs previously lived only in `Azure/azure-rest-api-specs-pr` on `RPSaaSMaster` at `2025-12-01-preview` https://github.com/Azure/azure-rest-api-specs-pr/tree/RPSaaSMaster/specification/gcpconnector; this PR brings them public directly at GA, so the preview version is intentionally **not** carried over.

| TypeSpec project | Resource | Kind |
| --- | --- | --- |
| `Gcpconnector.Management` | `Operations` | — |
| `BigQueryDataset.Management` | `BigQueryDataset` | TrackedResource |
| `CloudFunction.Management` | `CloudFunction` | TrackedResource |
| `SQLAdminInstances.Management` | `SQLAdminInstance` | TrackedResource |
| `StorageBucket.Management` | `StorageBucket` | TrackedResource |
| `ComputeInstance.Management` | `ComputeInstance` | ExtensionResource |
| `ContainerClusters.Management` | `ContainerCluster` | ExtensionResource |

All resources, operations, models and properties from `2025-12-01-preview` are carried over unchanged. There are no `@added` / `@removed` / `@renamedFrom` decorators — `2027-01-01` is the sole entry in each `Versions` enum. Generated OpenAPI lands under `resource-manager/Microsoft.GcpConnector/stable/2027-01-01/` and `readme.md` exposes tag `package-2027-01-01`.

**SDK generation:** `suppressions.yaml` keeps the `SdkTspConfigValidation` waiver, reworded from the temporary "preview version, SDK not required" note into a **permanent GA exemption** — no SDKs are generated from `Microsoft.GcpConnector`. The customer-facing SDKs for this scenario are generated from `Microsoft.HybridConnectivity` (see #45480).

**Validation performed:** `tsp compile .` clean for all 7 projects; `azsdk_run_typespec_validation` passed for all 7 projects; prettier formatting applied; no remaining `2025-12-01-preview` references under `specification/gcpconnector`.

## PR review workflow diagram

## Purpose of this PR

  - [x] New resource provider.
  - [ ] Other, please clarify:

## Due diligence checklist

## Additional information

<details>

</details>
<details>

This PR carries a `SdkTspConfigValidation` suppression in `specification/gcpconnector/suppressions.yaml`, reworded as a permanent GA exemption because no SDKs are generated from this resource provider.

</details>

## Getting help

  and https://aka.ms/ci-fix.


safeermohammed (approved):

alexmahonic:
Applying `Approved-Avocado`. The `MISSING_README` errors are a false positive from an Avocado bug affecting PRs that onboard a brand-new RP folder.

specification/gcpconnector/suppressions.yaml
```diff
@@ -0,0 +1,10 @@
+- tool: TypeSpecValidation
+  path: "**"
```
ravimeda, line 2: **[NEW] 🔴 Blocking** **[[RPC-SUPPRESS-SCOPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppression-scoping-rules-rpc-suppress-scope)]** `specification/gcpconnector/suppressions.yaml` - line 2 - `path: "**"` makes the `SdkTspConfigValidation` waiver blanket the entire service instead of only the project configuration files covered by the exemption.

**Classification reasoning:** This is new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers this blanket scope, and resolve this conversation; otherwise narrow the suppression before approval.

**Suggested fix:** Scope the waiver to the seven project `tspconfig.yaml` files, using a supported narrow glob such as `resource-manager/Microsoft.GcpConnector/*/tspconfig.yaml` or an explicit list of those paths. Keep the specific no-SDK justification.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/main.tsp
```diff
@@ -0,0 +1,634 @@
...
+
+  /**
+   * Output only. The Cloud Build name of the latest successful deployment of the function.
+   */
+  build?: string;
+
+  /**
+   * Docker Registry to use for this deployment. This configuration is only applicable to 1st Gen functions, 2nd Gen functions can only use Artifact Registry. If `docker_repository` field is specified, this field will be automatically set as `ARTIFACT_REGISTRY`. If unspecified, it currently defaults to `CONTAINER_REGISTRY`. This field may be overridden by the backend for eligible deployments.
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-types" "Maintaining compatibility with current SDKs."
```
ravimeda, line 86: **[NEW] 🔴 Blocking** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/main.tsp` - line 86 - This and 62 other `no-unnamed-types` suppressions use the reason “Maintaining compatibility with current SDKs,” although this RP generates no SDK and has no public baseline.

**Classification reasoning:** These 63 suppressions are new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers these unsupported rationales, and resolve this conversation; otherwise correct the suppressions before approval.

**Suggested fix:** Replace the anonymous extensible string unions with named open unions while retaining `string` as the open variant. If a real compatibility constraint exists, identify the exact SDK/package and incompatible generated API instead of using generic boilerplate.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/ContainerClusters/main.tsp
```diff
@@ -0,0 +1,3414 @@
...
+}
+
+/**
+ * Logging configuration for the cluster
+ */
+model LoggingConfig {
+  /**
+   * Configuration of cluster logging components
+   */
+  #suppress "@azure-tools/typespec-azure-core/no-unnamed-types" "inline anonymous model"
```
ravimeda, line 2552: **[NEW] 🔴 Blocking** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/ContainerClusters/main.tsp` - line 2552 - This and 26 other `no-unnamed-types` suppressions use “inline anonymous model” as the reason, which merely restates the lint violation and provides no technical justification.

**Classification reasoning:** These 27 suppressions are new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers these suppressions, and resolve this conversation; otherwise correct the suppressions before approval.

**Suggested fix:** Extract each anonymous object into a named model and remove its suppression, or document a concrete technical constraint proving why a named model cannot be used.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp
```diff
@@ -0,0 +1,430 @@
...
+  @pattern("^(?=.{0,259}[^\\s.]$)(?!.*[<>%&\\?/#])")
+  @key("name")
+  @path
+  @segment("bigQueryDatasets")
+  name: string;
+}
+
+#suppress "@azure-tools/typespec-azure-core/casing-style" "Suppress casing style as the property name is defined in GCP."
+#suppress "@azure-tools/typespec-azure-core/no-unknown" "Suppress no-unknown to handle the datatype object used in GCP"
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-no-record" "Suppress arm-no-record to handle the datatype object used in GCP"
```
ravimeda, line 52: **[NEW] 🔴 Blocking** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp` - lines 51-52 - Forty-four `arm-no-record` and six `no-unknown` suppressions use the generic “datatype object used in GCP” rationale without identifying the affected property, why an open shape is required, or why a compliant named model cannot represent it.

**Classification reasoning:** These 50 suppressions are new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers these generic rationales, and resolve this conversation; otherwise correct the suppressions before approval.

**Suggested fix:** Remove suppressions where named models can represent the shape. Where arbitrary GCP keys are contractually required, move the suppression to the exact `Record`/`unknown` property and explain why its key set is open and cannot be enumerated.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp
```diff
@@ -0,0 +1,430 @@
...
+import "./shared.tsp";
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using TypeSpec.Versioning;
+using OpenAPI;
+using Azure.ResourceManager;
+using Azure.Portal;
+
+#suppress "@azure-tools/typespec-azure-resource-manager/missing-operations-endpoint" "Suppress missing operations endpoint as the same is defined ../operations/main.tsp."
```
ravimeda, line 18: **[NEW] 🔴 Blocking** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp` - line 18 - This and five sibling `missing-operations-endpoint` suppressions cite nonexistent `../operations/main.tsp`; the shared Operations interface is defined in `../GcpConnector/main.tsp` at line 30.

**Classification reasoning:** The six suppressions are new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`. The other affected lines are CloudFunction 17, ComputeInstance 16, ContainerClusters 20, SQLAdminInstances 18, and StorageBucket 17.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers these inaccurate reasons, and resolve this conversation; otherwise correct the reasons before approval.

**Suggested fix:** Update each reason to cite the actual shared project, for example: “The Microsoft.GcpConnector operations endpoint is defined by the shared Operations interface in `../GcpConnector/main.tsp`.” Keep the suppressions because the separate projects share the RP namespace.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/readme.md
```diff
@@ -0,0 +1,55 @@
...
+openapi-type: arm
+openapi-subtype: rpaas
+tag: package-2027-01-01
+```
+
+### Suppression
+
+```yaml
+directive:
+  - suppress: OperationsAPIImplementation
```
ravimeda, line 37: **[NEW] 🔴 Blocking** **[[RPC-SUPPRESS-SCOPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppression-scoping-rules-rpc-suppress-scope)]** `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/readme.md` - line 37 - All 14 new readme directives omit mandatory `from:` selectors. The six `OperationsAPIImplementation` entries additionally omit `where:`, so they suppress the rule for each entire package.

**Classification reasoning:** These directives are new because `specification/gcpconnector` does not exist at base SHA `b02520492e1b99f101d785e6333524891942c154`. The affected directives are at BigQueryDataset 37/39, CloudFunction 37/39/48, ComputeInstance 37/39, ContainerClusters 37/39/53, SQLAdminInstances 37/39, and StorageBucket 37/39.

**Approval context:** No `Approved-Suppression` label was observed at the reviewed head SHA. If approval exists, apply the label, confirm it covers these blanket scopes, and resolve this conversation; otherwise narrow the directives before approval.

**Suggested fix:** Add exact `from:` and `where:` selectors matching the emitted OpenAPI and LintDiff diagnostic paths. Retain the reasons explaining the RP-wide operations project and the specific additional-properties/secret false positives.

ravimeda (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`661b871`](https://github.com/Azure/azure-rest-api-specs/pull/45812/commits/661b871cdb2ca939b95d156c51872cc97431a3fc). See inline comments for findings 1-6.

specification/gcpconnector/suppressions.yaml
alexmahonic: Fixed in 0917d11.

`path: "**"` was far broader than intended — it would have applied to any spec folder, not just this one. Narrowed to:

```yaml
path: "resource-manager/Microsoft.GcpConnector/*/tspconfig.yaml"
```

which matches exactly the seven `tspconfig.yaml` files in this RP and nothing else.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/main.tsp
alexmahonic: You're right, and the stated reason was simply false — this is a new, unreleased API version with no SDK and no public baseline, so there was nothing to stay compatible with. All 63 suppressions are gone as of 5456e5b.

Each anonymous union is now a named open union following the convention already used by the 19 hand-named unions in `ContainerClusters`, with `string` retained so values Google adds later stay representable:

```tsp
/**
 * BuildConfig.dockerRegistry values, as defined by the Google Cloud API.
 */
union DockerRegistry {
  string,

  /** Unspecified. */
  DOCKER_REGISTRY_UNSPECIFIED: "DOCKER_REGISTRY_UNSPECIFIED",

  /** Docker images will be stored in multi-regional Container Registry repositories named `gcf`. */
  CONTAINER_REGISTRY: "CONTAINER_REGISTRY",

  /** Docker images will be stored in regional Artifact Registry repositories. ... */
  ARTIFACT_REGISTRY: "ARTIFACT_REGISTRY",
}
```

Per-variant documentation comes from the `enumDescriptions` in Google's published discovery documents rather than being invented, which also let me diff every value against upstream. That surfaced a set of transcription errors in the original draft, now corrected:

| Property | Was | Now (per discovery doc) |
|---|---|---|
| `GatewayApiConfig.channel` | `RAPID`, `REGULAR`, `STABLE` (these are `ReleaseChannel`'s values) | `CHANNEL_DISABLED`, `CHANNEL_EXPERIMENTAL`, `CHANNEL_STANDARD` |
| `EnterpriseConfig.enterpriseTier` | `ADVANCED`, `PREMIUM` | `STANDARD`, `ENTERPRISE` |
| `CloudRunConfig.loadBalancerType` | `EXTERNAL`, `INTERNAL` | `LOAD_BALANCER_TYPE_EXTERNAL`, `LOAD_BALANCER_TYPE_INTERNAL` |
| `NetworkConfig.datapathProvider` | `LEGACY`, `ADVANCED` | `LEGACY_DATAPATH`, `ADVANCED_DATAPATH` |
| `ConfidentialNodes.confidentialInstanceType` | `CONFIDENTIAL_VM`, `CONFIDENTIAL_SPACE` | `SEV`, `SEV_SNP`, `TDX` |
| `DatabaseEncryption.currentState` | `..._ENCRYPTING`, `..._DECRYPTING` | `..._ENCRYPTION_PENDING`, `..._DECRYPTION_PENDING` |
| `SecurityPostureConfig.mode` | `STANDARD` | `DISABLED`, `BASIC`, `ENTERPRISE` |

I also restored 16 enums that had been flattened to plain `string` (SQLAdmin `Operation.operationType`, `Settings.*`, `IpMapping.type`, and others).

All seven projects compile with `--warn-as-error`, `npx tsv` passes on all seven, and `oav validate-example` passes on the changed specs.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/ContainerClusters/main.tsp
alexmahonic: Agreed — "inline anonymous model" restated the rule rather than justifying it. Addressed in 4808d56, split two ways.

**Extracted into named models (4), suppression removed:** `LoggingComponentConfig`, `MonitoringComponentConfig`, `NodeConfigDefaults` in `ContainerClusters`, and `OperationErrors` in `SQLAdminInstances`. These are cases where Google's own discovery document defines the shape as a named schema and `$ref`s it, so a named model is the faithful representation. (A fifth was absorbed as a nested member of one of these.)

**Kept (22), with a real justification replacing the boilerplate:** these are shapes that Google's discovery document defines *inline* rather than as a `$ref`'d schema — most of them in `storage/v1`. Extracting them would invent a named type that has no counterpart upstream and would diverge our wire contract's structure from the API we are projecting. The reason now says that, rather than restating the diagnostic.

The distinction is meaningful here because the discovery format *does* support `$ref` for object schemas, so Google inlining a shape is a deliberate authoring choice rather than a limitation of the format. (That is not true of enums — discovery has no mechanism for a named enum — which is why I did not use the same argument for the `no-unnamed-types` unions in the sibling thread and named all 63 instead.)

Happy to extract the remaining 22 as well if you'd prefer structural consistency on our side over fidelity to Google's schema layout — just say the word.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp
alexmahonic: Fixed in 0917d11: **50 model-level suppressions removed, 28 property-scoped ones added.**

Before changing anything I tested whether each was actually load-bearing — I copied the projects to a scratch directory, stripped *every* `arm-no-record` and `no-unknown` suppression, and compiled:

- `BigQueryDataset`: 4 removed → 3 diagnostics
- `ComputeInstance`: 8 removed → 7 diagnostics
- **`no-unknown`: 0 diagnostics across the board — all six of those suppressions were dead** and are now deleted outright.

(The `no-unknown` ones were dead because the properties in question are `Record<unknown>[]`; both rules test `prop.type` directly, so the array wrapper means neither rule ever fired.)

The remaining 28 are now attached to the exact `Record<>` property rather than the enclosing model, and carry one of three concrete reasons instead of the single copy-pasted string:

1. **User-supplied key/value maps** — GCP labels, resource-manager tags, node `metadata`, `sysctls`. The key set is chosen by the customer at runtime; there is no enumerable schema.
2. **`google.protobuf.Any` / `Status.details`** — the payload type is selected by the service per error condition and is not enumerated in the discovery document.
3. **Keyed-by-identifier collections** — e.g. `perLocationOperations`, keyed by GCP location name, and `MaintenanceWindow.maintenanceExclusions`, keyed by a user-chosen exclusion name.

Worth noting for context: the reason strings were 100% uniform across all 50 sites (44 identical + 6 identical), which is what a bulk find-and-replace looks like rather than 50 individual judgements — so your read of it as generic boilerplate was correct.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/main.tsp
alexmahonic: Fixed in 0917d11. `../operations/main.tsp` never existed — thanks for catching it and for pointing at the real location. All six reasons now cite the shared project:

> The Microsoft.GcpConnector operations endpoint is defined by the shared `Operations` interface in `../GcpConnector/main.tsp`.

Updated at `BigQueryDataset` 18, `CloudFunction` 17, `ComputeInstance` 16, `ContainerClusters` 20, `SQLAdminInstances` 18, and `StorageBucket` 17 — the exact set you listed. Suppressions kept, as you suggested, since the projects share the RP namespace.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/BigQueryDataset/readme.md
alexmahonic: Fixed in 82a22cf. All 14 directives across the six service readmes now carry `from:`, and the six `OperationsAPIImplementation` directives additionally carry `where:`:

```yaml
directive:
  - suppress: OperationsAPIImplementation
    from: storageBucket.json
    where: $.paths
    reason: The Operations API for Microsoft.GcpConnector is defined once for the whole resource provider in the GcpConnector spec (GcpConnector/stable/2027-01-01/operations.json).
```

I used `where: $.paths` rather than guessing, based on the existing precedent in `specification/blueprint/resource-manager/Microsoft.Blueprint/Blueprint/readme.md`, which pairs `suppress: OperationsAPIImplementation` with `from:` and `where: $.paths`.

Each readme's YAML blocks parse cleanly and `npx tsv` passes on all seven projects.

alexmahonic:
## Enum values refreshed against Google's discovery documents

While addressing the suppression findings above, naming the 63 anonymous unions required deciding what each one's values should be. Rather than preserve the draft's values as-is, I diffed every enum against Google's published discovery documents (`container/v1`, `compute/v1`, `cloudfunctions/v2`, `sqladmin/v1`, `storage/v1`, `bigquery/v2`) and took upstream as authoritative.

**Service team: please sanity-check the following**, since these are behavioural corrections, not just cosmetics.

- **7 properties had values that do not exist in any published version of Google's API** (verified against the current doc plus 2025-08 and 2024-09 snapshots). Most look like transcription errors — `GatewayApiConfig.channel` was carrying `ReleaseChannel`'s values, for instance. Full list in the `no-unnamed-types` thread above.
- **19 properties were missing values Google has added**, now included.
- **16 properties had been flattened to plain `string`** despite being enums upstream; these are restored as named unions (notably SQLAdmin `Operation.operationType`, `Settings.*`, `IpMapping.type`).
- Per-variant documentation is generated from Google's `enumDescriptions` rather than hand-written.

Every union keeps `string` as an open variant, so nothing is rejected at runtime and later upstream additions remain representable without a breaking change.

alexmahonic:
### Note on the two remaining red checks

Both are agent/tooling failures unrelated to this PR's content:

- **TypeSpec APIView** — `New-TypeSpecAPIViewTokens: The term 'pnpm' is not recognized as a name of a cmdlet, function, script file, or executable program.`
- **spec - api-doc-preview** — `npm error Missing script: "install-pnpm"`, followed by `gh: Bad credentials (HTTP 401)` when the job tried to report its own status.

Same root cause: the pipelines expect a pnpm bootstrap step that isn't resolving on the agent.

Everything that validates this PR's content is green: Swagger LintDiff, Swagger BreakingChange, Breaking Change (Cross-Version), Swagger Avocado, Swagger ModelValidation, Swagger SemanticValidation, Swagger APIView, TypeSpec Validation, TypeSpec Suppressions, SpellCheck, Protected Files, and all six SDK Validation jobs.

`Automated merging requirements met` is red pending the suppression approval labels and removal of `ARMChangesRequested`.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
```diff
@@ -0,0 +1,1291 @@
...
+          "$ref": "#/definitions/OnDeployUpdatePolicy",
+          "description": "OnDeployUpdatePolicy of BuildConfig."
+        },
+        "runtime": {
+          "type": "string",
+          "description": "The runtime in which to run the function. Required when deploying a new function, optional when updating an existing function. For a complete list of possible choices, see the [`gcloud` command reference](https://cloud.google.com/sdk/gcloud/reference/functions/deploy#--runtime)."
+        },
+        "serviceAccount": {
+          "type": "string",
+          "description": "[Preview] Service account to be used for building the container"
```
psah434: **[NEW] 🔵 Suggestion** *(no rule anchor — documentation-quality observation)* `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json` - line 429 - `BuildConfig.serviceAccount` carries a `[Preview]` marker in a stable GA api-version.

**Classification reasoning:** [NEW] — `Microsoft.GcpConnector` has no previous public api-version, so every finding on this version is new.

**Issue:** In a stable (GA) api-version, this property description literally begins with "[Preview]" — *"[Preview] Service account to be used for building the container"*. A GA contract should not label individual properties "[Preview]" in customer-facing docs: either the capability is GA (drop the marker) or it is genuinely preview (in which case it should not ship in a stable version). This is a documentation-quality observation, not a violation of a named rule.

**Suggested fix:** Remove the "[Preview]" prefix from the description, or move the property to a preview api-version if it is not GA-ready.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
```diff
@@ -0,0 +1,1291 @@
...
+          "$ref": "#/definitions/Environment",
+          "description": "Describe whether the function is 1st Gen or 2nd Gen."
+        },
+        "eventTrigger": {
+          "$ref": "#/definitions/EventTrigger",
+          "description": "An Eventarc trigger managed by Google Cloud Functions that fires events in response to a condition in another service."
+        },
+        "kmsKeyName": {
+          "type": "string",
+          "description": "[Preview] Resource name of a KMS crypto key (managed by the user) used to encrypt/decrypt function resources. It must match the pattern `projects/{project}/locations/{location}/keyRings/{key_ring}/cryptoKeys/{crypto_key}`."
```
psah434: **[NEW] 🔵 Suggestion** *(no rule anchor — documentation-quality observation)* `specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json` - line 698 - `kmsKeyName` carries a `[Preview]` marker in a stable GA api-version.

**Classification reasoning:** [NEW] — `Microsoft.GcpConnector` has no previous public api-version, so every finding on this version is new.

**Issue:** Same as the `serviceAccount` finding: in a stable (GA) api-version this description literally begins with "[Preview]" — *"[Preview] Resource name of a KMS crypto key (managed by the user) used to encrypt/decrypt function resources..."*. A GA contract should not surface `[Preview]` markers on individual properties.

**Suggested fix:** Remove the "[Preview]" prefix from the description, or move the property to a preview api-version if it is not GA-ready.

psah434 (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, manual decision) against commit [`711fb81`](https://github.com/Azure/azure-rest-api-specs/pull/45812/commits/711fb81f313f9097f79b9ea351ca36e2176bb1e6). See inline comments for findings 2-3; finding 1 (output-only properties missing `readOnly`, OAPI034) is a rule-citation correction posted as a reply on the existing unresolved thread rather than a new comment.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
```diff
@@ -0,0 +1,1291 @@
...
+      "type": "object",
+      "description": "Describes the Build step of the function that builds a container from the given source.",
+      "properties": {
+        "automaticUpdatePolicy": {
+          "$ref": "#/definitions/AutomaticUpdatePolicy",
+          "description": "AutomaticUpdatePolicy of BuildConfig."
+        },
+        "build": {
+          "type": "string",
+          "description": "Output only. The Cloud Build name of the latest successful deployment of the function."
```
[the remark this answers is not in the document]
psah434: Correction to the rule citation above: the precise rule for this defect is **OAPI034 — Clear Field Ownership** ([property-mutability.md](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/property-mutability.md#oapi034-clear-field-ownership)) — *"Server-owned properties are computed by the service and MUST be `readOnly`"* — not OAPI027, which concerns *write-only* properties (settable but not readable via GET). The `build` property here is output-only-but-readable, so the correct requirement is that it be marked `"readOnly": true` (via `@visibility(Lifecycle.Read)` in TypeSpec). The defect, its cross-resource scope (~79 output-only properties across the six resource projects), and the fix are all unchanged.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
alexmahonic: Fixed in b6842bb092f.

The `[Preview]` prefix was stale rather than deliberate: it was transcribed from Google's `cloudfunctions/v2` discovery document around Sept 2024, when the field was in preview at Google. Google has since GA'd it and dropped the marker, so the text was inaccurate as well as inappropriate for a stable api-version.

Synced to Google's current wording, which also adds the format note:

> Service account to be used for building the container. The format of this field is `projects/{projectId}/serviceAccounts/{serviceAccountEmail}`.

Your alternative suggestion (move to a preview api-version) doesn't apply here — the capability is GA at Google.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
alexmahonic: Fixed in b6842bb092f, same cause as the `serviceAccount` finding.

The marker came from Google's `cloudfunctions/v2` discovery document as of ~Sept 2024. Google's current document has no `[Preview]` marker on `Function.kmsKeyName` and the rest of the text is unchanged, so the fix is just dropping the prefix — our description now matches Google's current wording verbatim.

These were the only two `[Preview]` occurrences in the RP; the spec is now clean.

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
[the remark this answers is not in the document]
alexmahonic: Thanks for the rule correction — agreed that OAPI034 is the right anchor, not OAPI027.

I'd like to push back on the finding itself, though. Two pieces of evidence:

| Spec | mirrored properties | `readOnly` |
|---|---|---|
| `Microsoft.AwsConnector` `stable/2024-12-01` (GA, in main) | 4,802 | 112 — `provisioningState` only |
| `Microsoft.GcpConnector` `preview/2025-12-01-preview` (-pr) | 1,089 | 6 — `provisioningState` only |
| this PR `stable/2027-01-01` | 1,093 | 6 — `provisioningState` only |

The directly analogous multicloud-inventory RP shipped GA through ARM review with zero mirrored cloud properties marked `readOnly`, including several AWS documents as computed. This PR is consistent with both it and the private draft it was ported from.

On the substance: OAPI034 covers properties *computed by the service*. `"Output only."` in Google's discovery document is GCP's ownership annotation — it means Google computes the field and no GCP caller can set it. It does not make the field computed by `Microsoft.GcpConnector`, which transcribes it rather than deriving it. The one field this RP genuinely computes is `provisioningState`, and that is marked `readOnly` on every resource type.

The practical constraint is that ARM's `readOnly` is caller-agnostic. These resources are projections whose property bag is populated by the connector's discovery path via PUT; there's no way to express "read-only to customers, writable by ingestion." Marking these would strip 38 of 40 mirrored fields out of the ComputeInstance PUT body.

One note on scope: checked against Google's *current* discovery documents rather than our own descriptions, the set is 127, not ~79 — the smaller number is what our descriptions reveal, and some of ours are stale. So the finding would be larger than stated if we did apply it.

Also, minor: the `[EXISTING]` classification looks incorrect. All six findings on the earlier trusted ARM review thread concerned suppression hygiene; none touched property mutability. Your correction re-tags it `classification: new`, which I think is right.

Happy to reverse this if the ARM position is that the ingestion path should not use the public PUT surface — that's the question it turns on, and I'll confirm the write path with the service team. Leaving this thread open for your call rather than resolving it.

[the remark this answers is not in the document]
psah434: Agreed this turns on the write-path question, and thanks for the AwsConnector data — precedent alone isn't dispositive for me, but your substance argument is the stronger one: if `Microsoft.GcpConnector` only transcribes these fields (rather than computing them) and the connector's own ingestion writes them through the **public** ARM PUT, then `readOnly` is inapplicable here and OAPI034 doesn't fire — ARM's caller-agnostic `readOnly` would incorrectly block the ingestion write path.

Please confirm with the service team whether the discovery/ingestion path writes these properties via the same public PUT surface customers use:
- **If it does** → I'll withdraw the finding; these are legitimately writable, resolve the thread.
- **If ingestion uses a separate internal path** (public PUT is customer-only) → the mirrored fields are customer-read-only and should carry `readOnly` (the count is ~127, per your discovery-doc check), and we'd want a follow-up to apply it.

Leaving open pending that confirmation.

psah434:
please remove the armchangesrequested label once you have discussed with your team

specification/gcpconnector/resource-manager/Microsoft.GcpConnector/CloudFunction/stable/2027-01-01/cloudFunction.json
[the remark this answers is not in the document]
alexmahonic: Confirmed internally with our service bahavior: **ingestion writes these resources through the public ARM PUT.** The connector fetches the resource from the GCP API, places the returned JSON into `properties`, and PUTs it via MetaRP against the ARM endpoint. Customers don't write these resources at all — the write surface exists solely for ingestion.

That's your first branch. And because those PUTs are validated against the registered swagger, marking the mirrored fields `readOnly` would make the connector's own ingestion write illegal. Since customers write nothing, the consistent extension isn't 127 properties — it's all 1,093, leaving an empty PUT body.

This also matches `Microsoft.AwsConnector` `stable/2024-12-01`: PUT on 110/110 resource types, full mirrored bag writable.

Asking to resolve on that basis.

mpothuganti (approved):

PR #45822, opened 2026-08-26
Spec fixes for next GA release

jpalvarezl:
- Rename for Java
- Adding beta markers for header gated field for chat completions and responses
- Resorting union fields to avoid unnecessary breaking API change


glecaros (approved):

glecaros:
@jpalvarezl, please make sure the checks pass (for one, I think the files need formatting)

PR #45781, opened 2026-08-25
Add 2026-09-01 Storage DataRP API

sssash18:
## Summary
- Port the stable `2026-09-01` Storage DataRP contract from https://github.com/Azure/azure-rest-api-specs-pr/pull/29696 into the parent `Microsoft.Storage` TypeSpec project.
- Add Blob Access Point configuration CRUD/list and connection-test operations, models, client customizations, and examples.
- Apply the version-scoped Connector changes: remove `creationTime` and introduce the six camel-cased `dataSourceType` values.
- Add `allowStorageBlobAccessPointConfigurations` to the storage data collaboration policy for `2026-09-01`.
- Keep examples and suppressions scoped to DataRP resources only.

## Validation
- `npm exec --no -- tsp compile --list-files --warn-as-error .`
- `npm exec --no -- tsp compile --no-emit --warn-as-error client.tsp`
- `npx oav validate-spec specification/storage/resource-manager/Microsoft.Storage/stable/2026-09-01/openapi.json -l warn`
- Targeted `oav validate-example` for all 19 DataRP operation IDs

PR #45825, opened 2026-08-26
[Storage][SRP] Add BlobAccessPointConfiguration to Create Container API (2026-09-01)

gauravm043:
## Summary
Adds the `blobAccessPointConfiguration` property to the blob container (`ContainerProperties`) for api-version **2026-09-01**, plus a new `BlobAccessPointConfigurationConnection` model and a `BlobContainersPutBlobAccessPointConfiguration` example.

This is the public-spec (TypeSpec) counterpart of the SRP control-plane change (Storage-SRP PR [16577214 - "Rename Storage Connector to Blob Access Point Configuration in container API"](https://msazure.visualstudio.com/One/_git/Storage-SRP/pullrequest/16577214)).

## Scope (only the Blob Access Point container change)
- `Storage.Management/models.tsp`: `ContainerProperties.blobAccessPointConfiguration` (`@added(Versions.v2026_09_01)`) + `BlobAccessPointConfigurationConnection` model.
- `Storage.Management/examples/2026-09-01/BlobContainersPutBlobAccessPointConfiguration.json`.

## Dependency / sequencing
Targets `feature/storage/spr-2026-09-01`. The property is gated to api-version `2026-09-01`, introduced separately by #45594 (Add API version 2026-09-01). This PR is intentionally scoped to only the Blob Access Point container change and does not duplicate the version scaffolding; it should merge after api-version 2026-09-01 is present on the feature branch. Draft until then.

### DIFF
--- a/specification/storage/Storage.Management/examples/2026-09-01/AdvancedPlatformMetricsCRUD/AdvancedPlatformMetricsRules_CreateOrUpdate_AllContainers.json
+++ b/specification/storage/Storage.Management/examples/2026-09-01/AdvancedPlatformMetricsCRUD/AdvancedPlatformMetricsRules_CreateOrUpdate_AllContainers.json
@@ -1,0 +1,59 @@
+{
+  "operationId": "AdvancedPlatformMetrics_CreateOrUpdate",
+  "title": "AdvancedPlatformMetricsRules_CreateOrUpdate_AllContainers - Create advanced platform metrics rule with all containers filter",
+  "parameters": {
+    "subscriptionId": "00000000-0000-0000-0000-000000000000",
+    "resourceGroupName": "res6977",
+    "accountName": "sto2527",
+    "advancedPlatformMetricsRuleType": "ContainerLevelCapacityMetrics",
+    "api-version": "2026-09-01",
+    "resource": {
+      "properties": {
+        "enabled": true,
+        "ruleConfig": {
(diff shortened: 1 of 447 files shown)


nateprewitt (commented):
Thanks for the PR @gauravm043. This looks like it may have some overlap with https://github.com/Azure/azure-rest-api-specs/pull/45781. Is this PR meant to be additive onto the other?

It looks like there's some variation where things are referred to here as `blobAccessPointConfiguration` but `blobAccessPointConfigurations` in #45781. Is that distinction intended?

gauravm043:
> Thanks for the PR Gaurav Mishra (@gauravm043). This looks like it may have some overlap with #45781. Is this PR meant to be additive onto the other?
> 
> It looks like there's some variation where things are referred to here as `blobAccessPointConfiguration` but `blobAccessPointConfigurations` in #45781. Is that distinction intended?

Hi @nateprewitt yes this is an additive PR [part of the same feature work]. #45781 introduces control plane CRUD for blobAccessPointConfiguratio**ns**/blobAccessPointConfiguratio**n**  resource type. My PR adds the capability for attaching a single configuration to a container using the Create Container API, so the "s" is intentional.

nateprewitt (approved):
Thank @gauravm043, this should be fine to merge into the PR branch. Please make sure you've addressed any issues in CI in the main PR (#46102) and filed any of the forms required in the overview.

PR #45827, opened 2026-08-26
Add configuration_state field to AgentObject

daniel-rodrigues809:
Adds a new optional, read-only `configuration_state` field to `AgentObject` using the existing `AgentState` union.

This field reports the agent's explicit administrative enable/disable setting independently from the composite operational `state`, which may also be disabled because an Entra identity is disabled. Together with `state_source`, clients can distinguish administrative intent from identity-derived availability.

- `configuration_state: enabled` — the agent was administratively enabled.
- `configuration_state: disabled` — the agent was administratively disabled.
- absent — backward-compatible behavior for older responses.

`specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp`

Validation: `npx tsp compile .` succeeds and emits `configuration_state` as optional and read-only in both configured OpenAPI outputs.


mattchenderson (approved):

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -159,6 +159,10 @@ model AgentObject {
   @visibility(Lifecycle.Read)
   state: AgentState;
 
+  @doc("The administrative configuration state of the agent. This reflects whether the agent was explicitly enabled or disabled, independently of identity-derived operational state.")
+  @visibility(Lifecycle.Read)
+  configuration_state?: AgentState;
+
```
glecaros, line 165: Is this field preview? If it is we probably want to annotate with the extension 
```typespec
@extension("x-ms-foundry-meta", #{ required_previews: #[<the-opt-in-key>] })
```

daniel-rodrigues809: @matthew henderson I don't think this needs to be a preview field since technically this enable / disable is already available and in use? we are just moving it from "state" to "configuration_state"

mattchenderson: The state transition controls themselves are GA. This is us swizzling the properties around to allow another GA experience to drive behavior on an existing one (`state`), in a way that is intentionally NOT opt-in. Think of this as acting like the existing `state` property under current behavior. I think we need to return it in all cases.

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
```diff
@@ -159,6 +159,10 @@ model AgentObject {
   @visibility(Lifecycle.Read)
   state: AgentState;
 
+  @doc("The administrative configuration state of the agent. This reflects whether the agent was explicitly enabled or disabled, independently of identity-derived operational state.")
+  @visibility(Lifecycle.Read)
```
dargilco, line 178: Can you say something about the state of the Agent when this property is omitted? "When omitted the Agent is ... "

glecaros: based on the comment below, I would assume this shouldn't be optional

specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp
glecaros: got it, thanks!

glecaros (approved):

PR #45831, opened 2026-08-26
Add skuDistributionMode property to SkuProfile for 2026-09-01

wenhuan-zhai-msft:
Add SkuDistributionMode enum (Heterogeneous, Homogeneous) and skuDistributionMode property to SkuProfile in ComputeRP.json. Include example for creating a VMSS with homogeneous SKU distribution mode.


audreyttt (changes requested):
The example also needs to be copied over to the stable folder which should happen automatically when you run tsp compile

wenhuan-zhai-msft:
> The example also needs to be copied over to the stable folder which should happen automatically when you run tsp compile
Rerun tsp compile and now example file is correctly copied over
specification/compute/resource-manager/Microsoft.Compute/Compute/stable/2026-09-01/examples/virtualMachineScaleSetExamples/VirtualMachineScaleSet_Create_WithSkuProfile_SkuDistributionMode.json

audreyttt (approved):

PR #45766, opened 2026-08-24
Centurion WAF: add WAF policy tier + Basic_v2/Basic_WAF_v2 SKUs (2026-01-01)

baoqihuang0326:
## Summary
Adds Centurion WAF (Basic WAF v2) API contract changes to api-version 2026-01-01:

1. **Application Gateway SKU names and tiers** - adds Basic_v2 and Basic_WAF_v2 to ApplicationGatewaySkuName and ApplicationGatewayTier.
2. **Reserved capacity** - adds reservedCapacity for the Basic_v2 and Basic_WAF_v2 SKU tiers.
3. **WAF Policy tier** - adds the WebApplicationFirewallPolicyTier enum (Basic / Standard) and optional tier property on WebApplicationFirewallPolicyPropertiesFormat.
The WAF policy tier values are ordered to align generated SDK enum ordinals with Standard = 0 and Basic = 1.

All changes are authored in TypeSpec (`Network/Network/models.tsp`) and gated with `@added(Versions.v2026_01_01)`. The 2026-01-01 Swagger was regenerated with `tsp compile`, and the WAF policy get/create-or-update examples demonstrate `tier`.

Adapted from #45330 for the 2026-01-01 API version with the corrected Basic_v2/Basic_WAF_v2 SKU set.

Validation: TypeSpec compile, OAV validate-spec, and OAV validate-example pass.

PR #45844, opened 2026-08-26
Add AppGw Advanced Routing Rule in NRR API 2026-01-01

paawankohli:
# Summary

Introduces Advanced Routing for Application Gateway in API version 2026-01-01: a rules engine that lets a request routing rule select a backend based on evaluated request properties (headers, query string, path, client IP, HTTP method) instead of just host/path.

Entry point is a new AdvancedRouting value on the existing ApplicationGatewayRequestRoutingRuleType enum. A rule of that type points at an advanced routing map, which owns an ordered set of rules; each rule references a condition set and a backend/redirect target, with map-level defaults for the no-match case.

All changes are authored in TypeSpec (Network/Network/models.tsp) and gated with `@added(Versions.v2026_01_01)`. The 2026-01-01 Swagger was regenerated with tsp compile, and the AppGw  get/create examples were added.

### DIFF
--- a/specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
+++ b/specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
@@ -563,6 +563,45 @@ union ApplicationGatewayRequestRoutingRuleType {
    * PathBasedRouting
    */
   PathBasedRouting: "PathBasedRouting",
+
+  /**
+   * AdvancedRouting
+   */
+  @added(Versions.v2026_01_01)
+  AdvancedRouting: "AdvancedRouting",
+}
+
+/**
+ * The type of request property that an advanced routing condition is evaluated against.
+ */
+@added(Versions.v2026_01_01)
+union ApplicationGatewayAdvancedRoutingConditionType {
+  string,
+
+  /**
+   * The condition is evaluated against an HTTP request header identified by propertyName.
+   */
+  Header: "Header",
+
+  /**
+   * The condition is evaluated against a query string argument identified by propertyName.
+   */
+  QueryString: "QueryString",
+
+  /**
+   * The condition is evaluated against the request path. propertyName is not applicable.
+   */
+  Path: "Path",
+
+  /**
+   * The condition is evaluated against the client IP address. propertyName and propertyValueMatcher are not applicable.
+   */
+  ClientIP: "ClientIP",
+
+  /**
+   * The condition is evaluated against the HTTP method. propertyName and propertyValueMatcher are not applicable.
+   */
+  Method: "Method",
 }
 
 /**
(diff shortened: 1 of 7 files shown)


paawankohli:
Have not made any changes to existing model, have only added new properties as part of 20260101. Can someone review why **BreakingChange-Go-Sdk** and **BreakingChange-JavaScript-Sdk** labels have been added?

According to co-pilot, the change is not caused by my PR. The actual breaking changes are:

Go (armnetwork v12.0.0, build 6749912) — Total 1 breaking change(s), 36 additive change(s):
```
### Breaking Changes
- Type of `ManagedRuleSetRuleGroup.Rules` has been changed from `[]*string` to `[]*int32`
```

JavaScript (@azure/arm-network, build 6749913):
```
### Breaking Changes
  - Type of parameter rules of interface ManagedRuleSetRuleGroup is changed from string[] to number[]
  - Operation WebApplicationFirewallPoliciesOperations.createOrUpdate has a new signature
  - Operation WebApplicationFirewallPoliciesOperations.get has a new signature
  - Operation WebApplicationFirewallPoliciesOperations.list has a new signature
  - Operation WebApplicationFirewallPoliciesOperations.listAll has a new signature
```

All of it traces to one root: ManagedRuleSetRuleGroup.rules becoming an integer array. That came from [PR 45621](https://github.com/Azure/azure-rest-api-specs/pull/45621) which is already merged into release-microsoft-network-2026-01-01 and carrying BreakingChange-Approved-Benign. The four JS signature changes are downstream fallout of the same WAF policy type change.

spec-gen-sdk generates the SDK from your head SHA and diffs it against the last published package, not against the base branch. Since release-microsoft-network-2026-01-01 hasn't shipped, every PR into it inherits the branch's entire accumulated breaking diff. This also impacts other PRs. Eg: https://github.com/Azure/azure-rest-api-specs/pull/45766

paawankohli:
**Regarding [Swagger LintDiff](https://github.com/Azure/azure-rest-api-specs/actions/runs/33048055984):** 

Linter has flagged that node name, and description can't be the same but for **ApplicationGatewayRequestRoutingRuleType**, but we are following the precedent and adhering to the pattern that is used by existing rule types

image

paawankohli:
**Regarding [Breaking Change(Cross-Version)](https://github.com/Azure/azure-rest-api-specs/actions/runs/33048711131):**

Workflows are bringing up some errors related to changes which are not in my PR - [Add AppGw Advanced Routing Rule · Azure/azure-rest-api-specs@28bdbcc](https://github.com/Azure/azure-rest-api-specs/actions/runs/33048711131)

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -7881,6 +7932,255 @@ model ApplicationGatewayPathRulePropertiesFormat {
...
+model ApplicationGatewayAdvancedRoutingMap extends SubResource {
+  /**
+   * Properties of the application gateway advanced routing map.
+   */
+  properties?: ApplicationGatewayAdvancedRoutingMapPropertiesFormat;
+
+  /**
+   * Name of the advanced routing map that is unique within an Application Gateway.
+   */
+  name?: string;
```
ajaygit158, line 7951: Shouldn't be optional

paawankohli: Following the precedent set by similar feature URLPathMap, please refer line 7791 of the same file.

**Please note:** Advanced Routing feature is in a way a super set of Path Routing feature and hence the API design (as well as the product offering) is strongly inspired by that feature. 

AdvancedRoutingMap is inspired by UrlPathMap
AdvancedRoutingRule is inspired by PathRule

santoshgh317 (approved):

PR #45853, opened 2026-08-27
update customization for storage

kazrael2119:
update customization for storage

### DIFF
--- a/specification/storage/Storage.Management/client.tsp
+++ b/specification/storage/Storage.Management/client.tsp
@@ -1443,5 +1443,11 @@ enum ProvisioningStateEnum {
 @@clientName(
   Azure.ResourceManager.CommonTypes.Encryption,
   "StorageAccountEncryption",
-  "python,java,go,javscript"
+  "python,java"
+);
+
+@@clientName(
+  Azure.ResourceManager.CommonTypes.Encryption,
+  "ContextCacheEncryption",
+  "go,javascript"
 );
--- a/specification/storage/Storage.Management/client.tsp
+++ b/specification/storage/Storage.Management/client.tsp
@@ -1448,6 +1448,6 @@ enum ProvisioningStateEnum {
 
 @@clientName(
   Azure.ResourceManager.CommonTypes.Encryption,
-  "ContextCacheEncryption",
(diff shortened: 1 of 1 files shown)


specification/storage/Storage.Management/client.tsp
```diff
@@ -1445,3 +1445,9 @@ enum ProvisioningStateEnum {
   "StorageAccountEncryption",
   "python,java"
 );
+
+@@clientName(
+  Azure.ResourceManager.CommonTypes.Encryption,
+  "ContextCacheEncryption",
```
kazrael2119:
```suggestion
  "ArmEncryption",
```

specification/storage/Storage.Management/client.tsp
```diff
@@ -1445,3 +1445,9 @@ enum ProvisioningStateEnum {
   "StorageAccountEncryption",
   "python,java"
 );
+
+@@clientName(
+  Azure.ResourceManager.CommonTypes.Encryption,
+  "ArmEncryption",
+  "go,javascript"
+);
```
kazrael2119, line 1453: @msyyc @weidongxu-microsoft 
could you help comfirm if using a common name like `ArmEncryption` is better to your languages? 
`Azure.ResourceManager.CommonTypes.Encryption` is only used by both ContextCacheProperties and ContextCachePropertiesUpdate. Renaming it to StorageAccountEncryption may cause confusion with the existing Encryption model used by StorageAccountProperties.

msyyc: Python already used name `StorageAccountEncryption` so we don't want to change it.

JiaqiZhang-Dev (approved):

specification/storage/Storage.Management/client.tsp
weidongxu-microsoft: Sorry, second thought, Java never call something `ArmXXX`. So I guess we keep `StorageAccountEncryption`

kazrael2119: got it

JialinHuang803 (approved):

PR #45856, opened 2026-08-27
feat(chaos): add required region to connection properties in 2026-11-01

franklixuefei:
> **Opened for @juliagao-microsoft's decision, not as a fait accompli.** This adds a member to your in-flight GA promotion, so please treat it as a proposal — happy to withdraw it or reshape it.

## What this adds

A **required** `region` property on `ConnectionProperties`, scoped to **`2026-11-01` only**.

Chaos Studio assigns each connection's regional data-plane endpoint (`dataPlaneEndpoint`, read-only) from the target's region, so the service must know that region at create time.

## Why the client supplies it rather than the service resolving it

This is the part most likely to draw ARM-review scrutiny, so the argument up front:

**An ARM resource ID does not encode the location of the resource it names.** `targetResourceId` gives us subscription, resource group, provider and name — not region. Resolving it service-side would require a read against the customer's target resource, and the caller has not necessarily granted us that.

The pattern is established for exactly this situation — resources whose **DNS or endpoint configuration is regional**:

- **`Microsoft.Search/searchServices/sharedPrivateLinkResources.resourceRegion`** — its own documentation says it is *"only required for those resources whose DNS configuration are regional (such as Azure Kubernetes Service)"*. Our target here is AKS.
- **Azure Front Door** does the same with `privateLinkLocation`.
- **`Microsoft.Chaos/targets`** already accepts a client-supplied `location` in this same RP.

The discriminator across Azure is consistent: where the region is needed for **endpoint provisioning at create**, the client supplies it; where it is needed only to validate a policy rule, the service resolves it. We assign a regional DNS endpoint, so we are on the first side.

## Why required, and why that is safe here

The endpoint cannot be assigned without it, so an optional field would only move the failure later and make it less clear. Requests naming a region where Chaos Studio operates no data plane are rejected with `400`.

**It is added only in `2026-11-01`.** Connections were introduced in `2026-08-01-preview`, so a member without `@added` would have applied from that version onward and **silently changed a shipped preview**. Verified after compiling:

- `stable/2026-11-01/openapi.json` — `region` present and listed in `required`
- `preview/2026-08-01-preview/openapi.json` — `region` **absent**, `required` unchanged

A client moving preview → GA must start supplying it. That is a deliberate preview-to-GA change, stated here rather than left to be discovered.

## Mechanics

- Type is `Azure.Core.azureLocation`.
- Swagger and example output **regenerated with `tsp compile`**; nothing hand-edited.
- Examples updated for the three operations with a body; `region: "eastus"` is consistent with the existing `dataplane.eastus.chaos-prod.azure.com` endpoint in those same examples.

## Context

Backs `Microsoft.Chaos/workspaces/connections`, the trust anchor authorizing an AKS cluster to reach the Chaos Studio data plane for a workspace. Internal design reference: D32 in the workspace-connections decision log; work item AB#38851037.

### DIFF
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/connection.models.tsp
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/connection.models.tsp
@@ -48,6 +48,23 @@ model ConnectionProperties {
   targetResourceId: Azure.Core.armResourceIdentifier;
 
   /**
+   * The Azure region of the target resource. Chaos Studio assigns this connection's
+   * regional data-plane endpoint from this value, so it must name a region in which
+   * Chaos Studio operates a data plane; the service rejects the request otherwise.
+   *
+   * This is supplied by the client because an Azure Resource Manager resource ID does
+   * not encode the location of the resource it names, and resolving it service-side
+   * would require a read against the target resource that the caller may not have
+   * granted. This mirrors the established pattern for resources whose DNS or endpoint
+   * configuration is regional — for example
+   * `Microsoft.Search/searchServices/sharedPrivateLinkResources.resourceRegion`, whose
+   * own documentation calls it out as required specifically for targets such as Azure
+   * Kubernetes Service.
+   */
+  @added(Microsoft.Chaos.Versions.v2026_11_01)
+  region: Azure.Core.azureLocation;
+
+  /**
    * The Microsoft Entra principal (object) ID of the identity used by the connection.
    */
   principalId?: Azure.Core.uuid;
--- a/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/examples/2026-11-01/Connections_CreateOrUpdate.json
+++ b/specification/chaos/resource-manager/Microsoft.Chaos/Chaos/examples/2026-11-01/Connections_CreateOrUpdate.json
@@ -9,6 +9,7 @@
       "properties": {
         "kind": "AksExtension",
         "targetResourceId": "/subscriptions/6b052e15-03d3-4f17-b2e1-be7f07588291/resourceGroups/exampleRG/providers/Microsoft.ContainerService/managedClusters/exampleCluster",
+        "region": "eastus",
         "principalId": "1a2b3c4d-5e6f-7a8b-9c0d-1e2f3a4b5c6d",
         "tenantId": "72f988bf-86f1-41af-91ab-2d7cd011db47"
       }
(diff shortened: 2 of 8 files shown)


franklixuefei:
**On hold — please don't spend review time on this yet.**

An internal design question has reopened that may make this property unnecessary. The `region` here anchors a connection's data-plane endpoint to the **target cluster's** region. We're re-examining whether it should instead follow the **workspace's** region — which ARM already carries, since `Microsoft.Chaos/workspaces` is a `TrackedResource` with `location`.

If we go that way, **no ARM contract change is needed at all** and I'll close this.

The current shape exists because a shipped, blocking conformance gate in our own AKS extension onboarding asserts the endpoint's region equals the cluster's location. That gate is ours, and whether its rationale still holds is exactly what we're checking.

Leaving it open rather than closing, since the change itself is complete and verified if we do proceed. Apologies for the noise on the promotion branch.

franklixuefei:
Closing this — the design decision it implements has been **reversed**, so the member is no longer needed.

**What changed.** This PR added a required `region` to `ConnectionProperties` because `dataPlaneEndpoint` was to be assigned from the **target AKS cluster's** region, and an ARM resource id does not encode location — so the client had to supply it. That basis no longer holds.

Two findings reversed it:

1. **The forcing constraint was circular.** The decision cited a shipped, blocking conformance gate as what "forced" cluster-anchoring. On inspection, that gate's own stated rationale is test-harness hygiene — it exists so a cross-region test pair does not produce "a failure that looks like a product defect but is a setup mistake." Its load-bearing claim, that a cross-region pair "is not a valid configuration", is asserted and nowhere supported. The sibling data-plane design states the opposite: the control plane assigns a region "nearest, capacity-based, etc. — not necessarily the cluster's own region."

2. **Cluster-anchoring required a cross-region storage path that does not exist.** The per-region authorization artifact for the cluster's region would have to be written by the backend that owns the workspace, sitting in a different region. That storage is regional.

**The replacement.** The connection's region is the region of the **workspace that owns it**, which the service already knows without asking anyone. Every service-owned hop becomes co-regional; the only cross-region hop is the customer's own in-cluster subscriber reaching its assigned endpoint over Front Door — the same shape already used for agent-based faults.

So `region` stays **server-derived and off the contract**. `dataPlaneEndpoint` remains read-only and server-assigned, exactly as it ships today. No spec change is required, and the GA surface is unaffected.

Apologies for the churn, and thanks to anyone who spent time on it.

PR #45862, opened 2026-08-27
Fix Container Apps 2026-03-02-preview CI findings

najian:
Removes the restored Source Control parameters that caused cross-version rule 1042 (ChangedParameterOrder), cleans the corresponding examples, carries forward pre-existing preview lint suppressions, and suppresses Avocado for the intentionally retired internal Functions invoke operation.\n\nThis updates #45788 without introducing a breaking-change approval requirement.


najian:
Superseded by consolidated PR #45866, which targets main directly and includes this validated non-breaking CI fix.

PR #45868, opened 2026-08-27
Add SecurityTest RP initial API spec (2026-01-01)

hasskooelhadi:
Adding initial API specification for Microsoft.SecurityTest resource provider.

- New RP namespace: Microsoft.SecurityTest
- API version: 2026-01-01 stable
- ARM resource type


hasskooelhadi:
@microsoft-github-policy-service agree

PR #45869, opened 2026-08-27
Add SecurityTest resource provider API spec (2026-01-01)

hasskooelhadi:
Adds the initial REST API specification for the SecurityTest resource provider.

This PR adds:
- Resource manager API specification
- Initial swagger definition
- SDK generation configuration


hasskooelhadi:
@microsoft-github-policy-service agree

PR #45764, opened 2026-08-24
Add AuthenticationPolicy resource and Application Gateway authConfigs binding for 2026-01-01

baoqihuang0326:
## Summary

Adds the **AuthenticationPolicy** resource (`Microsoft.Network/authenticationPolicies`) and binds it to Application Gateway request routing rules, targeting the `stable/2026-01-01` API version.

### Changes
- **New** `stable/2026-01-01/authenticationPolicy.json` — standalone `Microsoft.Network/authenticationPolicies` resource with `ListAll` / `List` / `Get` / `CreateOrUpdate` / `Delete` operations and definitions (`AuthenticationPolicy`, `AuthenticationPolicyPropertiesFormat`, `AuthenticationProviderProperties`, enums `UserTrustProviderType`, `OnUnauthenticatedRequest`).
- **New** 5 examples under `stable/2026-01-01/examples/` (CreateOrUpdate, Get, Delete, List, ListAll).
- **Modified** `stable/2026-01-01/applicationGateway.json` — added `authConfigs` array on the request routing rule properties plus the `ApplicationGatewayAuthConfig` definition (links a named config to an authentication policy and/or authorization policy).
- **Modified** `Network/readme.md` — wired `authenticationPolicy.json` into the `package-2026-01-01` input-file list.

### Local validation performed
- JSON parse — pass (all files)
- Prettier (`specification/**/*.json` swagger plugin config) — pass
- `oav validate-spec` (semantic) — pass (authenticationPolicy.json and applicationGateway.json)
- `oav validate-example` (model/example) — pass

> LintDiff (Spectral/AutoRest) was not reproducible locally due to an AutoRest core-package bootstrap limitation in the dev environment; it will run in CI on this PR.

## Contribution checklist
- [x] I commit to follow the [Breaking Change Policy](https://aka.ms/AzBreakingChangesPolicy/) of "no breaking changes".
- [x] I have reviewed the [documentation](https://aka.ms/ContributeToAzureRestAPISpecs) for the workflow.
- [ ] Design review (if required) has been completed / linked.

PR #45892, opened 2026-08-27
Correct AuthenticationPolicy contract for 2026-01-01

baoqihuang0326:
## Summary

Follow-up to #45764 that aligns the `2026-01-01` AuthenticationPolicy contract with the implemented Network RP service behavior. `2026-01-01` is unreleased — it exists only on the `release-microsoft-network-2026-01-01` branch and has no counterpart on `main` — so these corrections land before the version ships rather than changing a published contract.

### Application Gateway authentication bindings

- Move the authentication bindings to the attachment points the service actually implements: `defaultAuthConfigs` on `ApplicationGatewayAdvancedRoutingMap` and `authConfigs` on `ApplicationGatewayAdvancedRoutingRule`.
- Remove `defaultAuthConfigs` from `ApplicationGatewayUrlPathMap` and `authConfigs` from `ApplicationGatewayRoutingRule`; those are not advanced routing attachment types and were never implemented.
- Retain `authConfigs` on `ApplicationGatewayRequestRoutingRule`.
- All bindings use `maxItems: 1` because NRP supports one auth config per attachment point.
- Document that auth config names must be unique across the gateway, that a gateway can reference at most 100 distinct authentication policies, and that bindings require the Standard_v2 or WAF_v2 SKU.

### AuthenticationPolicy resource

- Add the user-assigned `identity` envelope property used by user sign-in policies to reach their Key Vault client secret.
- Replace the tags-only `AuthenticationPolicies_UpdateTags` PATCH with `AuthenticationPolicies_Update`, backed by a resource-specific `AuthenticationPolicyUpdateParameters` model carrying both `tags` and `identity`. AuthenticationPolicy is a tracked resource, so ARM requires PATCH for tag updates, and the MSI onboarding guidance requires the control plane to accept identity updates on PATCH. Using a dedicated update model rather than the shared `TagsObject` satisfies both at the source instead of via suppression.
- Keep `UserTrustProviderType` extensible (`modelAsString: true`) with the `Entra` SDK constant mapped to the lowercase wire value `entra`.
- Document that JWT validation and user sign-in fields are mutually exclusive within a single policy.

### Constraints and documentation

- Service limits: issuer 2048 characters, client secret URL 4096 characters, session cookie name 256 characters, scope item 128 characters, and scope collection 10 items.
- `issuer` is typed as `url` with an HTTPS pattern, consistent with `jwksUri` and `clientSecret`.
- `sessionTimeout` remains a string to match the implemented wire shape, with its accepted syntax and units documented and enforced via pattern and length.
- `clientId` is documented and constrained as a GUID.
- Normalize AuthenticationPolicy example locations to lowercase and align example payloads with service-tested values.

### Suppressions

Net change for AuthenticationPolicy is two fewer suppressions.

- Scope the `ProvisioningStateMustBeReadOnly` suppression to the specific GET, PUT, and PATCH response schemas.
- Make the `arm-resource-invalid-envelope-property` suppression reason self-contained for the `identity` property.
- No `PatchIdentityProperty` suppression is needed, because the PATCH body accepts `identity`.
- Remove the `TrackedResourcePatchOperation` suppression, which no longer applies now that PATCH exists.
- Remove the obsolete `SchemaDescriptionOrTitle` suppression now that the description is emitted.

## Clarifications

- The TypeSpec union member is named `Entra`, while its serialized wire value remains lowercase `entra`. This is intentional and matches the existing NRP contract.
- `ApplicationGatewayAuthConfig` is not renamed in this PR. The shared Azure Firewall naming question can be resolved before a Firewall binding is introduced, avoiding an unrelated change to the currently published Application Gateway model.
- Renaming `AuthenticationPolicies_UpdateTags` to `AuthenticationPolicies_Update` is safe because `2026-01-01` is unreleased, and the former name is no longer accurate now that the body is not tags-only.

## Validation

- TypeSpec compilation
- OAV semantic validation for `authenticationPolicy.json` and `applicationGateway.json`
- OAV example validation for `authenticationPolicy.json`
- Prettier and diff checks


baoqihuang0326:
Addressed in 49afe4cd7f. `UserTrustProviderType` now includes the open `string` variant, matching `OnUnauthenticatedRequest`. Generated Swagger restores `x-ms-enum.name: UserTrustProviderType`, `modelAsString: true`, and the `Entra` SDK constant mapped to the current lowercase wire value `entra`. I also removed the obsolete `SchemaDescriptionOrTitle` suppression because the description is now emitted.

baoqihuang0326:
Addressed in 9d1040b363:

- Documented that JWT validation and user sign-in fields are mutually exclusive and cannot be combined in one policy.
- Documented that auth config names must be unique across the gateway and that a gateway can reference at most 100 distinct authentication policies.
- Scoped the `ProvisioningStateMustBeReadOnly` suppression to the GET 200 and PUT 200/201 response schemas.
- The PATCH body parameter naming concern is no longer applicable: this PR removes the unsupported AuthenticationPolicy PATCH operation and its example entirely.

baoqihuang0326:
Updated in e90beb847c using the service-tested payloads:

- Added the AuthenticationPolicy user-assigned identity property used by user sign-in policies to access their Key Vault client secret.
- Updated the user sign-in create/update example with the tested `eastus`, identity, issuer, client ID, and client secret values.
- Added a separate JWT-validation create/update example using only JWT fields.
- Updated the Basic_v2 Application Gateway example with a user-assigned identity, HTTPS listener hostname, and the supported request-routing-rule `authConfigs` binding.

The advanced-routing fragment is intentionally not included because its advanced routing map/rule models are not public in this API version; adding those values to the existing URL path map/routing rule types would recreate the unsupported placement issue this PR removes.

baoqihuang0326:
Corrected in 4909de55b6 based on the updated service information:

- Restored the tags-only `AuthenticationPolicies_UpdateTags` PATCH operation supported by the generic controller.
- Restored the source and generated PATCH examples.
- Renamed the generated body parameter to `parameters` using the established Network RP `@@clientName` pattern.
- Removed the no-PATCH waiver.
- Added PATCH 200 to the scoped `ProvisioningStateMustBeReadOnly` suppression.
- Documented the 256-character combined limit across all scope entries.

baoqihuang0326:
Addressed in 502f1541ed:

- Removed `authConfigs` from the Basic_v2 request and both response bodies, restoring that example to the release-branch version. The binding example is now on the Standard_v2 `ApplicationGatewayCreate.json` example and references the JWT validation policy.
- Added `defaultAuthConfigs` to `ApplicationGatewayAdvancedRoutingMapPropertiesFormat` and `authConfigs` to `ApplicationGatewayAdvancedRoutingRulePropertiesFormat`, both with `maxItems: 1` and `x-ms-identifiers: ["name"]`.
- Added the returned user-assigned identity, including service-populated `principalId` and `clientId`, to both AuthenticationPolicy PUT response bodies.
- Added machine-readable constraints for client ID GUID shape, HTTPS/2048-character JWKS URI, 2048-character audience, nonempty/patterned scope values, and session cookie name pattern.
- Removed the incorrect 4096-character limit from the Key Vault URL. The description now states the URL requirements and documents that the retrieved secret value is limited to 4096 characters.

The branch also incorporated the concurrent release-branch merge and PATCH identity-property suppression before this commit was pushed.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
```diff
@@ -74,45 +79,60 @@ model AuthenticationPolicyPropertiesFormat {
...
 
-  /** The scopes used by an application during authentication to authorize access to a user's details. A maximum of 10 scopes is supported. */
+  /** The scopes used by an application during authentication to authorize access to a user's details. A maximum of 10 scopes is supported, each scope can contain up to 128 characters, and all scopes can contain up to 256 characters combined. */
   @maxItems(10)
-  scope?: string[];
+  scope?: AuthenticationScope[];
 
-  /** The timeout of the session cookie used for user authentication, in seconds. Applicable to the Application Gateway post-OIDC workflow. */
+  /** The timeout of the session cookie used for user authentication, in seconds. The supported range is 1 to 604800 seconds, and the default is 86400 seconds. Applicable to the Application Gateway post-OIDC workflow. */
   sessionTimeout?: string;
```
sandipsh, line 133: **[EXISTING]** **[[TSP-NUMERIC-TYPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#64-numeric-properties-must-use-numeric-types-tsp-numeric-type)]** `specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp` - line 115 - `sessionTimeout` is still a `string`, so the documented numeric range (`1` to `604800`) and default (`86400`) cannot be expressed or validated in the schema.

**Classification reasoning:** The property is also `string` at base SHA `8864214`; this PR adds the numeric range/default documentation but leaves the existing wire type unchanged.

**Suggested fix:** Please confirm the implemented wire shape. If it is numeric, use `int32` with `@minValue(1)` and `@maxValue(604800)` in a new API version or obtain the required versioning approval before changing this published stable contract. If the service intentionally sends a string, document the accepted string syntax and units so clients can validate it.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
```diff
@@ -74,45 +79,60 @@ model AuthenticationPolicyPropertiesFormat {
...
- * Properties for authentication provider configuration.
+ * Properties for authentication provider configuration. A policy must configure either JWT validation fields or user sign-in fields; the two configurations are mutually exclusive and cannot be combined.
  */
 #suppress "@azure-tools/typespec-azure-core/no-legacy-usage" "Legacy feature decorator required for Network RP feature gating, consistent with other Network RP property models"
 @added(Versions.v2026_01_01)
 @Azure.ResourceManager.Legacy.feature(Features.authenticationPolicy)
 model AuthenticationProviderProperties {
   /** The URL of the Secure Token Service. Include a trailing slash at the end of the value. Example: https://login.microsoftonline.com/{Microsoft Entra Tenant ID}/ */
+  @maxLength(2048)
   issuer: string;
```
sandipsh: **[EXISTING]** **[[TypeSpec Section 5 - URI properties](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp` - line 90 - `issuer` is documented as the Secure Token Service URL but remains plain `string`, so the emitted schema lacks `format: uri` while adjacent `jwksUri` now uses `url`.

**Classification reasoning:** `issuer` is also `string` at base SHA `8864214`; this PR changes its constraints but not its type.

**Suggested fix:** Confirm compatibility before changing this published stable contract. Prefer `url` in a new API version so SDKs validate and type the URI, or document why the service intentionally accepts non-URI strings.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
```diff
@@ -42,6 +43,10 @@ model AuthenticationPolicy extends Common.Resource {
   #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "etag is an established read-only envelope property across Network RP resources"
   @visibility(Lifecycle.Read)
   etag?: string;
+
+  /** The user-assigned identity used by a user sign-in policy to access its Key Vault client secret. */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "AuthenticationPolicy uses the established Network RP identity envelope property pattern."
```
sandipsh: **[NEW] 🟠 Warning** **[[TypeSpec Section 4.1](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#41-warning-suppression-policy)]** `specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp` - line 48 - The new `arm-resource-invalid-envelope-property` suppression reason relies on an “established Network RP ... pattern” instead of explaining why this specific identity property legitimately belongs in the resource envelope.

**Classification reasoning:** The identity property and this suppression are introduced by this PR and are absent at base SHA `8864214`.

**Approval context:** No `Approved-TypeSpecSuppression` label was observed at the session SHA. Please confirm whether TypeSpec suppression approval already covers this exact suppression. If it does, ensure `Approved-TypeSpecSuppression` is applied and resolve this conversation; otherwise make the reason self-contained or obtain approval.

**Suggested fix:** State the technical exception directly, for example: `AuthenticationPolicy requires this user-assigned identity in the resource envelope so user sign-in policies can access their Key Vault secret reference; the property uses the standard ManagedServiceIdentity shape.`

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`502f154`](https://github.com/Azure/azure-rest-api-specs/pull/45892/commits/502f1541ed39f4ee3bd8c6a0986321e697dea52d). See inline comments for findings 2, 3, and 5. Findings 1 and 4 concern lines outside the current diff and are posted as top-level comments.

sandipsh:
**[EXISTING]** **[[ARM Section 26.2](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#262-forbidden-schema-changes)]** `specification/network/resource-manager/Microsoft.Network/Network/stable/2026-01-01/applicationGateway.json` - line 5052 / JSON path `$.definitions.ManagedRuleSetRuleGroup.properties.rules.items` - Cross-version analysis reports this item changing from `string` in `2025-09-01` to `integer`/`int32` in `2026-01-01` (`TypeChanged` and `TypeFormatChanged`). The aggregate status is currently green, but the underlying API type change still needs an explicit disposition.

**Classification reasoning:** This shape is already present at PR base SHA `8864214`, so it was inherited from the release branch rather than introduced by the AuthenticationPolicy commits.

**Approval context:** No `BreakingChange-Approved-*` or `Versioning-Approved-*` label was observed at the session SHA; the Go/JavaScript SDK approval labels do not approve the REST wire change. Please confirm whether API-level approval already covers this exact type change. If it does, ensure the applicable API approval label is applied and resolve this conversation; otherwise restore the prior wire type or obtain approval.

**Suggested fix:** Restore the prior `string` item type in the owning TypeSpec and regenerate, or complete the API breaking-change approval process for this stable-version type change. If the release branch is fixed separately, merge/rebase that fix and rerun validation.

sandipsh:
**[EXISTING]** **[[TSP-ARM-RESOURCE-ID](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#63-use-armresourceidentifier-for-resource-references-tsp-arm-resource-id)]** `specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp` - line 70 - `associatedResources` is documented as a collection of ARM resource IDs but is modeled as `string[]`, so ARM-aware validation and SDK resource-ID typing are lost.

**Classification reasoning:** The property is also `string[]` at base SHA `8864214` and is unchanged in this PR.

**Suggested fix:** Preserve compatibility for this published stable version. In the next API version, model each item as `armResourceIdentifier` with the applicable resource-type constraints; if the collection intentionally accepts arbitrary strings, document that explicitly instead of calling every value a resource ID.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
```diff
@@ -309,13 +314,17 @@ suppressions:
...
     from: authenticationPolicy.json
     reason: Network RP resources in this package intentionally use the legacy shared Resource envelope, which does not expose systemData.
-  - code: SchemaDescriptionOrTitle
-    from: common.json
-    where: $.definitions.UserTrustProviderType
-    reason: The TypeSpec union is documented, but the Swagger 2.0 emitter does not carry that documentation to this closed single-value union in the shared common.json file.
   - code: XMSSecretInResponse
     from: authenticationPolicy.json
     reason: clientSecret holds a Key Vault secret URL reference (e.g. https://myvault.vault.azure.net/secrets/mysecret), not the secret value itself, so it is safe to return in responses and is not marked x-ms-secret.
+  - code: PatchIdentityProperty
```
sandipsh: **[NEW] 🟠 Warning** **[[RPC-Patch-V1-11](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/arm-api-review.instructions.md#43b-patch-should-support-updating-envelope-properties-rpc-patch-v1-09-rpc-patch-v1-11) / [RPC-SUPPRESS-SCOPE](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/skills/azure-api-review/references/suppression-review-criteria.md#suppression-scoping-rules-rpc-suppress-scope)]** `specification/network/resource-manager/Microsoft.Network/Network/readme.md` - line 320 - The new `PatchIdentityProperty` suppression accepts PUT-only identity updates even though RPC-Patch-V1-11 says PATCH should include identity when updates are supported, and it has no `where:` clause, so it suppresses this rule across the entire `authenticationPolicy.json` file.

**Classification reasoning:** Introduced in this PR. Identity and this AuthenticationPolicy-specific suppression are absent at base SHA `8864214`; the base already had the tags-only PATCH.

**Approval context:** No `Approved-Suppression` label was observed at the session SHA. Please confirm whether an existing approval covers this exact exception. If it does, ensure the label is applied and resolve this conversation; otherwise obtain ARM peer review after narrowing the suppression or use a compliant PATCH/mutability design.

**Suggested fix:** First add a `where:` targeting the exact `PatchIdentityProperty` JSON path reported by LintDiff for the `AuthenticationPolicies_UpdateTags` body parameter. Then either include identity in an AuthenticationPolicy-specific PATCH model, mark identity create/read-only if immutable, or document the PUT-only update exception and obtain ARM peer approval.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
```diff
@@ -7990,6 +7982,14 @@ model ApplicationGatewayAdvancedRoutingMapPropertiesFormat {
    */
   defaultRewriteRuleSet?: SubResource;
 
+  /**
+   * Default authentication configuration bindings of the advanced routing map. Only one authentication configuration is supported. Authentication configuration names must be unique across the Application Gateway, and an Application Gateway can reference at most 100 distinct authentication policies. Authentication policies can only be bound to Application Gateways using a v2 SKU.
+   */
+  @added(Versions.v2026_01_01)
+  @TypeSpec.OpenAPI.extension("x-ms-identifiers", #["name"])
+  @maxItems(1)
+  defaultAuthConfigs?: ApplicationGatewayAuthConfig[];
```
sandipsh, line 7991: **[NEW] 🔴 Blocking** **[[Section 2 - API Versioning](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/openapi-review.instructions.md#2-api-versioning)]** `specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp` - line 7991 - This moves published authentication bindings to different models in stable `2026-01-01`: `ApplicationGatewayUrlPathMap.defaultAuthConfigs` and `ApplicationGatewayRoutingRule.authConfigs` are removed, while similarly named properties are added to the advanced routing map/rule models. The new properties are not wire-compatible replacements for the removed properties.

**Classification reasoning:** Introduced in this PR. Both removed properties are present in the PR base, including `defaultAuthConfigs` at base source line 7843, and absent at the current head; the advanced-model properties are newly added at current lines 7991 and 8081.

**Approval context:** No `Versioning-Approved-*` label was observed at the session SHA. If API versioning approval already covers this exact published-contract correction, apply the matching label and resolve this conversation; otherwise preserve the old properties or introduce the corrected placement in a new API version.

**Suggested fix:** Restore the two published properties in `2026-01-01` and add the corrected advanced-model bindings in a new API version. If changing the published stable version is unavoidable, obtain explicit versioning approval for both removals and apply the matching `Versioning-Approved-*` label.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iteration(s), converged) against commit [`502f154`](https://github.com/Azure/azure-rest-api-specs/pull/45892/commits/502f1541ed39f4ee3bd8c6a0986321e697dea52d). See inline comments for findings 1-2.

baoqihuang0326:
Addressed in ea8ce5e033:

- Removed the `entraJWTValidationConfigs` collection and `entraJWTValidationConfig` rule reference from the Basic_v2 request and both response bodies. The Basic_v2 example now contains no authentication feature unsupported by that SKU.
- Updated request-routing-rule, advanced-routing-map, and advanced-routing-rule auth binding descriptions to explicitly allow only Standard_v2 and WAF_v2.
- Changed the JWKS URI and Key Vault URL patterns to `^[Hh][Tt][Tt][Pp][Ss]://`, matching NRP case-insensitive URI scheme parsing.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
baoqihuang0326: Addressed in 972432be6b6. The suppression reason is now self-contained and states the technical exception rather than pointing at an RP pattern:

> AuthenticationPolicy requires this user-assigned identity in the resource envelope so a user sign-in policy can read its client secret from Key Vault; the property uses the standard ManagedServiceIdentity shape and is an envelope-level credential of the resource rather than a routing/authentication setting, so it cannot move under properties.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
baoqihuang0326: Addressed in 972432be6b6. `issuer` is now typed as `url` with the same HTTPS pattern already used by `jwksUri` and `clientSecret`, so the emitted schema is `"type": "string", "format": "uri", "pattern": "^[Hh][Tt][Tt][Pp][Ss]://", "maxLength": 2048`. The description was tightened to "The absolute HTTPS URL of the Secure Token Service."

On the versioning concern: `2026-01-01` is not published. It exists only on the `release-microsoft-network-2026-01-01` release branch (this PR's base) and there is no `stable/2026-01-01` directory on `main`. `issuer` itself was introduced in that same unreleased version by #45764, so tightening it here does not change any shipped contract.

specification/network/resource-manager/Microsoft.Network/Network/Network/AuthenticationPolicy.tsp
baoqihuang0326: Addressed in 972432be6b6. Confirmed the implemented wire shape is a string: the service-tested payloads used to build these examples send `"sessionTimeout": "86400"`, so `int32` would not match what NRP accepts or returns.

Taking the second option in your suggested fix, the accepted string syntax and units are now documented and enforced:

- Description: "The service accepts and returns this value as a string containing a base-10 unsigned integer number of seconds with no sign, decimal point, unit suffix, or whitespace (for example `86400`). The supported range is 1 to 604800 seconds, and the default is 86400 seconds."
- `@pattern("^[1-9][0-9]{0,5}$")` and `@maxLength(6)`, so the schema now rejects non-numeric strings, leading zeros, and values outside the digit width of the supported range.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
baoqihuang0326: Addressed in 972432be6b6 for the RPC-SUPPRESS-SCOPE part. The suppression is no longer file-wide; it now carries a `where:` targeting only the AuthenticationPolicy PATCH operation:

```yaml
- code: PatchIdentityProperty
  from: authenticationPolicy.json
  where: $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/authenticationPolicies/{authenticationPolicyName}"].patch
```

That matches the rule's `given` of `$[paths,'x-ms-paths'].*.patch`, so it suppresses exactly the one reported occurrence and leaves the rest of `authenticationPolicy.json` covered.

On RPC-Patch-V1-11 itself: `AuthenticationPolicies_UpdateTags` is a tags-only PATCH served by the generic Network RP controller, which does not process `identity` on this operation. Identity is set through PUT. Adding `identity` to the PATCH body would advertise a capability the service does not implement, and marking it create/read-only is not correct either since it is mutable via PUT. Requesting ARM peer review of the narrowed suppression.

specification/network/resource-manager/Microsoft.Network/Network/Network/models.tsp
baoqihuang0326: This one I would like to push back on, because the premise that `2026-01-01` is a published stable contract does not hold.

`2026-01-01` has never shipped. It exists only on the `release-microsoft-network-2026-01-01` release branch, which is the base branch of this PR. There is no `stable/2026-01-01` directory on `main`:

```
$ git ls-tree --name-only origin/main specification/network/resource-manager/Microsoft.Network/Network/stable/
...
specification/network/resource-manager/Microsoft.Network/Network/stable/2025-07-01
specification/network/resource-manager/Microsoft.Network/Network/stable/2025-09-01
```

The latest stable version on `main` is `2025-09-01`. `ApplicationGatewayUrlPathMap.defaultAuthConfigs` and `ApplicationGatewayRoutingRule.authConfigs` were both introduced on this same release branch by #45764 and have never been released, so removing them is not a change to a published contract and there is no client that could depend on them. Introducing the corrected placement in "a new API version" would mean shipping `2026-01-01` with bindings the service does not implement and then immediately deprecating them.

That is exactly what this follow-up PR is for: correcting the unreleased `2026-01-01` contract before it is merged to `main`, so that the released shape matches the implemented NRP behavior. `ApplicationGatewayAdvancedRoutingMap` / `ApplicationGatewayAdvancedRoutingRule` are the service's actual advanced routing attachment points; `ApplicationGatewayUrlPathMap` and `ApplicationGatewayRoutingRule` are not.

The "not wire-compatible replacements" observation is correct and intentional. They are not replacements, they are the correct attachment points. Please let me know if you would still like a `Versioning-Approved-*` label applied for tracking, but no versioning exception should be required for a version that has not been released.

specification/network/resource-manager/Microsoft.Network/Network/readme.md
baoqihuang0326: Follow-up in c866d62223f: the `where:` is now a two-entry list so it matches whichever path LintDiff reports, while still staying scoped to this single operation.

```yaml
where:
  - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/authenticationPolicies/{authenticationPolicyName}"].patch
  - $.paths["/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/authenticationPolicies/{authenticationPolicyName}"].patch.parameters[4]
```

`parameters[4]` is the `parameters` body parameter of `AuthenticationPolicies_UpdateTags`, matching the existing `PatchIdentityProperty` suppressions elsewhere in this readme, which all use the `.patch.parameters[N]` form.

baoqihuang0326: Resolved at the source in cdd0092c93a: the PATCH operation has been removed rather than suppressed.

The service supports only GET, PUT, and DELETE for `authenticationPolicies`. The tags-only `AuthenticationPolicies_UpdateTags` was carried over from #45764 and does not reflect an operation the RP meaningfully implements for this resource. This matches `ApplicationGatewayWebApplicationFirewallPolicies` in this same RP and API version, which also ships with only GET / PUT / DELETE / LIST and no PATCH.

Because `PatchIdentityProperty` is `given: $[paths,'x-ms-paths'].*.patch`, removing the operation means the rule no longer fires at all. The suppression is deleted, not narrowed:

- Removed `AuthenticationPolicies_UpdateTags` plus its source and generated examples.
- Deleted the `PatchIdentityProperty` suppression entirely.
- Restored the `TrackedResourcePatchOperation` suppression: "AuthenticationPolicy intentionally does not expose a PATCH operation because the service supports only GET, PUT, and DELETE for this resource."

`identity` remains on the resource envelope and is set through PUT, so there is no longer an operation that accepts a body while silently ignoring identity. Net effect on suppression count is zero, and the remaining one is a truthful statement about the service contract rather than a waiver of an ARM rule.

Worth noting this is only possible cleanly because `2026-01-01` is unreleased: the PATCH exists on the release branch but has never shipped, so removing it breaks no client.

baoqihuang0326: You're right, and I've reversed the previous change. Thanks for pushing back on this.

I had concluded AuthenticationPolicy wasn't a tracked resource, which was wrong. It extends the Network RP `Resource` base type in `common.json`, and that base supplies `location`, `tags`, and `x-ms-azure-resource: true`. It is tracked by every definition ARM uses, so RPC-Patch-V1-01 applies and PATCH has to exist. The WebApplicationFirewallPolicies comparison I drew earlier also doesn't hold up, since that resource has no `identity` and so was never in the same situation.

`516655408d6` restores PATCH, but with a resource-specific update model rather than the shared `TagsObject`:

```
model AuthenticationPolicyUpdateParameters {
  tags?: Record<string>;
  identity?: Common.ManagedServiceIdentity;
}
```

This is the separate PATCH object you originally asked about. Because the body now carries `identity`, the MSI onboarding requirement that the control plane accept identity updates on PATCH is met at the source, so `PatchIdentityProperty` no longer needs a suppression, and that suppression is gone rather than re-added. The `TrackedResourcePatchOperation` suppression is also removed, since it no longer applies.

Operation is `AuthenticationPolicies_Update` (renamed from `AuthenticationPolicies_UpdateTags`, which is safe because `2026-01-01` is unreleased and now inaccurate given the body is no longer tags-only). Added an example that exercises both a tag update and an identity update.

Net suppression change for this resource is -2.

ajaygit158 (approved):

PR #39677, opened 2026-01-15
Refactor(sql): migrate to unified folder structure

v-xuto:
This PR migrates the **sql** specification to the new unified folder structure according to Azure folder structure guidelines.

**Change:**
- Consolidated TypeSpec and Swagger files under 
```
sql/resource-manager/Microsoft.Sql/
├── SQL
└── Legacy   #Include "stable/2014-04-01-legacy" and "stable/2014-04-01"
```
- Fixed all common-types references and input-file paths in `readme.md`
- Added `suppressions.yaml` for legacy API versions
- Verified TypeSpec compilation works correctly
- Updated the relevant paths in `cspell.yaml` file

PR #45900, opened 2026-08-27
Remove internal endpoints from SQL examples

lfraleigh:
## Summary

Remove internal endpoint entries from the SQL managed instance outbound network dependency examples across all affected API versions.

The remaining certificate verification endpoints are unchanged.


mikeharder:
i don't think this change should cause modelvalidation to fail.  let me see if it was already failing in main.

mikeharder:
Failures of ModelValidation on old APIs are pre-existing (https://github.com/Azure/azure-rest-api-specs/pull/39677).  Added bypass label.

PR #45050, opened 2026-07-28
Refactor(providerhub): migrate to unified folder structure

v-xuto:
This PR migrates the **providerhub** specification to the new unified folder structure according to Azure folder structure guidelines.
	
**Change:**
- Consolidated TypeSpec and Swagger files under `resource-manager/Microsoft.ProviderHub/ProviderHub`
- Updated `tspconfig.yaml` with correct paths and ARM types directory
- Fixed all common-types references and input-file paths in `readme.md`
- Added `suppressions.yaml` for legacy API versions
- Verified TypeSpec compilation works correctly
- Updated the relevant paths in `specification/suppressions.yaml` file
- Add rpaas `readme.md` under resource-manger

**Benefits:**
- Unified structure for better maintainability
- Aligns with Azure folder structure standards
- Maintains backward compatibility
- Supports future TypeSpec migration efforts

PR #45913, opened 2026-08-28
[Microsoft.ProviderHub] Remove internal-only identifiers and retire preview API versions older than one year

wwendyc:

> [!IMPORTANT]
> **Purpose:** an internal compliance review identified internal-only Microsoft
> system names and endpoints present in the public `Microsoft.ProviderHub` API
> surface. This was raised as an incident and requires prompt remediation, so
> prioritised review would be appreciated.

## Purpose of this PR

  - [ ] New resource provider.
  - [x] Other, please clarify:
    - Remediation of internal-only identifiers exposed in public specs,
      generated SDKs, and the published ARM template reference.
    - Removal of preview API versions past the retirement window defined in the
      [Azure versioning guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/VersioningGuidelines.md).

Rebased onto the unified folder structure introduced by #45050.

## Release plan

[Release plan 36336](https://azsdk-releaseplan-dashboard-hveph5aqhhcfhtgu.westus-01.azurewebsites.net/?releaseplan=36336)
— *GA release plan for ProviderHub*

| | |
| --- | --- |
| API release type | GA |
| SDK release type | stable |
| Target | September 2026 |
| TypeSpec project | ProviderHub (management plane) |

Packages in scope: `Azure.ResourceManager.ProviderHub` (.NET),
`@azure/arm-providerhub` (JS), `azure-mgmt-providerhub` (Python),
`azure-resourcemanager-providerhub` (Java),
`sdk/resourcemanager/providerhub/armproviderhub` (Go).

This PR introduces no new API version. The plan covers the SDK re-release
required by the breaking spec changes below, across all five languages.

## Commits

Three independent changes, reviewable commit by commit.

### 1. Remove internal-system references (`00d36b1812d`)

Removes the affected models, properties and enum values across all API
versions, along with the internal placeholders and one preproduction hostname
that appeared in example payloads. Internal service and role placeholders are
replaced with `Contoso` equivalents matching the existing conventions in those
files.

One enum was previously closed (`modelAsString: false`). Because the service may
still return the retired values, leaving it closed would break SDK
deserialization on GET, so it is now modelled as an extensible union and unknown
values round-trip as strings.

**Operation removals** (`2024-09-01`, `2025-10-01`, `2026-02-01-preview`): the
`NewRegionFrontloadRelease` feature is removed in full — `GenerateManifest`,
`CreateOrUpdate`, `Get`, `Stop` — with the models and examples it owned. One of
the removed properties was **required** on the request model shared by the two
write operations, so they could no longer form a request the service accepts.
`Get` and `Stop` are removed alongside them rather than leaving a resource that
can be read but never created, which would violate ARM RPC and fail linting.

> [!NOTE]
> Because the feature is removed in full, no operation in the spec still depends
> on the removed property, and there is no request the generated SDK can no
> longer form. **No service-side change is required for this PR to be correct.**
>
> The open question for reviewers is a deprecation one: the service continues to
> serve these operations, but they will disappear from the generated SDKs, Bicep
> types and ARM template reference. Please confirm new-region frontload has no
> first-party callers depending on the public surface.

### 2. Remove preview versions older than one year (`cae58deb5b2`)

Removes `2021-05-01-preview`, `2021-06-01-preview` and `2021-09-01-preview`
(~5 years old), superseded by `2024-09-01`, `2025-10-01` and
`2026-02-01-preview`. Per the versioning guidelines: *"DO NOT keep a preview
feature in preview for more than 1 year"*, *"DO retire all prior preview API
versions 90 days after a new GA or preview API version is released"* and *"DO
remove retired API versions from the azure-rest-api-specs repo."*

`R4024 PreviewVersionOverOneYear` flags these, but only at `warning` severity,
which is why they persisted. For each version this removes the folder, its
`readme.md` tag block, its `service.yaml` entry and its `suppressions.yaml`
entry. The default tag (`package-2025-10-01`) and all GA versions are unchanged.

> [!NOTE]
> Please confirm these versions are already deprovisioned at the service. Policy
> is deprovision first, then remove from the repo.

### 3. Add missing resource id to 2020-11-20 examples (`8458c44aebc`)

Model validation reports `MISSING_RESOURCE_ID` for 18 response bodies across 16
examples in `2020-11-20`. These gaps are pre-existing, but commit 1 modifies that
spec, which brings the version into the changed set CI validates, so they now
surface as failures. Each id is backfilled from the equivalent `2024-09-01`
example. Examples only; no schema or wire change.

## Validation performed locally

| Check | Result |
| --- | --- |
| TypeSpec compile | Passes; checked-in swagger is emitter output, not hand-edited |
| LintDiff (ARM ruleset) | **New violations: 0** |
| Semantic validation (oav) | Passes on all remaining API versions |
| Model validation (oav) | **0 errors** on all remaining API versions |
| Avocado | No new errors (194 before → 181 after) |
| Prettier / cspell | Pass |

`Swagger BreakingChange` is expected to fail: this removes operations,
definitions and API versions that are published in shipped stable versions.
Requesting Breaking Change review.

## Due diligence checklist

- [ ] I have reviewed following [Resource Provider guidelines](https://aka.ms/rpguidelines), including [ARM resource provider contract](https://aka.ms/azurerpc) and [REST guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md).
- [x] A [release plan](https://aka.ms/azsdkdocs/release-plans) has been created. (Release plan 36336)


KPIyer:
Approved. User impact is minimal (12 customers likely internal). Wendy to send email to callers and cc Aarthi Murugan.

image

gary-x-li (approved):

wwendyc:
For the preview API versions, the ProviderHub API specs do not have direct impact on partners. We are working with the PMs to notify RPs using the older preview API versions to switch to the newer versions.

PR #45923, opened 2026-08-28
[PaloAltoNetworks.Cloudngfw] Add 2026-08-27-preview with session handling properties

Prajjwal-Singh12:
Adds six session handling properties to the firewall contract as a new preview
version. Purely additive — no changes to existing paths, and no existing
definition loses or changes a property.

Purely additive: diff vs 2026-07-29-preview shows 0 new or removed paths (71
unchanged), 2 new definitions (TcpReset, LoadDistribution), and 6 new optional
properties spread across 3 existing definitions. 125 of 128 existing definitions
are byte-identical; 0 definitions or properties were removed.

New properties (all optional, supplied on firewall create/update via PUT):

| ARM path | Type | Range | Default |
|---|---|---|---|
| `properties.sessionDrainTimeoutInMinutes` | int32 | 15–120 | 15 |
| `properties.networkProfile.outboundIdleTimeoutInMinutes` | int32 | 4–120 | 4 |
| `properties.networkProfile.outboundTcpReset` | TcpReset | — | Enabled |
| `properties.frontEndSettings[].idleTimeoutInMinutes` | int32 | 4–100 | 4 |
| `properties.frontEndSettings[].tcpReset` | TcpReset | — | Disabled |
| `properties.frontEndSettings[].loadDistribution` | LoadDistribution | — | Default (5-tuple) |

The outbound idle maximum (120) intentionally differs from the inbound maximum
(100), and the outbound TCP reset default (Enabled) intentionally differs from
the inbound default (Disabled). Both asymmetries are confirmed by the partner and
are not oversights.

All six properties are optional so the resource provider can distinguish an
unspecified field from an explicitly set one; the partner resets any field absent
from a PUT, so the control plane carries forward existing customer values.

TypeSpec changes (source of truth; swagger + examples are generated):
- main.tsp: add v2026_08_27_preview to the Versions enum.
- models.tsp: add TcpReset (Enabled/Disabled) and LoadDistribution (Default =
  five-tuple hash, SourceIP = two-tuple hash) as extensible unions, and add the
  six properties above to FirewallDeploymentProperties, NetworkProfile and
  FrontendSetting, all @added(v2026_08_27_preview) with @minValue/@maxValue
  bounds on the integer fields. TcpReset is a dedicated type rather than a reuse
  of the existing EnableStatus: the wire values coincide but the semantics are
  unrelated.
- service.yaml / readme.md: register the version and add the
  package-2026-08-27-preview tag.
- examples/2026-08-27-preview: full example set for the new version; the six
  properties are exercised in the Firewalls MaximumSet examples.

Commits are split for review convenience: the first commit is a mechanical copy of
2026-07-29-preview to 2026-08-27-preview (generated swagger and all 208 examples
identical apart from the api-version string), and the second commit contains the
contract change.


Prajjwal-Singh12:
Ran the standard TypeSpec validation pipeline for the Cloudngfw project.

- `npx tsp compile ... --warn-as-error` → Pass
- `npx tsp compile ... --no-emit --warn-as-error` → Pass
- `npx oav validate-example preview/2026-08-27-preview/PaloAltoNetworks.Cloudngfw.json` → Validation completes without errors

No errors, no warnings. Compiler v1.15.0, emitter `@azure-tools/typespec-autorest`.

specification/paloaltonetworks/resource-manager/PaloAltoNetworks.Cloudngfw/Cloudngfw/models.tsp
```diff
@@ -1350,6 +1410,26 @@ model NetworkProfile {
...
+  // intentionally differs from the Disabled default on FrontendSetting.tcpReset. Both
+  // asymmetries were confirmed by the partner - do not normalize them.
+  /**
+   * The duration, in minutes, for which an outbound source NAT session is permitted
+   * to stay inactive prior to termination. If not specified, a default of 4 is applied.
+   */
+  @added(Versions.v2026_08_27_preview)
+  @minValue(4)
+  @maxValue(120)
+  outboundIdleTimeoutInMinutes?: int32 = 4;
```
sandipsh: **[NEW] 🔴 Blocking** **[[TypeSpec Section 5](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#5-common-anti-patterns)]** `specification/paloaltonetworks/resource-manager/PaloAltoNetworks.Cloudngfw/Cloudngfw/models.tsp` - line 1425 - `outboundIdleTimeoutInMinutes` has a TypeSpec default and is reachable from `Firewalls_Update` through `FirewallResourceUpdateProperties.networkProfile`. The emitted PATCH schema therefore supplies `4` when the client omits the field, which conflicts with JSON Merge Patch semantics. The same issue applies to `FrontendSetting.idleTimeoutInMinutes` at line 1662 through `frontEndSettings`.

**Classification reasoning:** Introduced in this PR; neither property exists in `2026-07-29-preview` at base SHA `d96052c6d8ad733fcd738805f72ed60d98e5acd0`.

**Suggested fix:** Remove the TypeSpec `= 4` defaults from both properties and apply defaults server-side for PUT. PATCH omission must preserve the current value. Regenerate the OpenAPI and examples afterward.

sandipsh (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 2 iteration(s), converged) against commit [`8a5a5f5`](https://github.com/Azure/azure-rest-api-specs/pull/45923/commits/8a5a5f535d4fc9a01a865ee2178afcf4c78d6ef8). See inline comments for finding 1.

specification/paloaltonetworks/resource-manager/PaloAltoNetworks.Cloudngfw/Cloudngfw/models.tsp
Prajjwal-Singh12: Fixed in the new commit.

Prajjwal-Singh12:

Prajjwal-Singh12:
Ran `review_typespec`, below are the results:

## TypeSpec ARM Review — PaloAltoNetworks.Cloudngfw `2026-08-27-preview` (delta only)

**Scope reviewed:** `41ea1c669b..HEAD` — everything after the "Copy 2026-07-29-preview to new 2026-08-27-preview version" commit. That's 6 commits touching `models.tsp`, `client.tsp`, the generated swagger, and 5 MaximumSet examples. Pre-existing content excluded.

**Verdict: ✅ Ready, with 3 warnings to confirm before GA** — 0 critical, 3 warnings, 5 info.

---

## 🟡 Warnings

**1. PATCH array-replace will silently reset the three new `FrontendSetting` fields.**

`frontEndSettings` is carried in `FirewallResourceUpdateProperties` and is declared `@identifiers(#[])`, so under merge-patch the array is replaced wholesale. A client that PATCHes `frontEndSettings` without echoing `idleTimeoutInMinutes` / `tcpReset` / `loadDistribution` will wipe a customer's tuning back to the service default. Same risk, lower severity, for `networkProfile` (`models.tsp` L1398-L1410). Confirm the RP behavior (preserve vs. reset) and state it in the property descriptions — this is the kind of thing that bites on the first PATCH after the feature ships.

**2. TCP-only properties on a model that can be UDP.**

`FrontendSetting.protocol` is `TCP | UDP`, but `idleTimeoutInMinutes` and `tcpReset` (`models.tsp` L1628-L1640) only apply to TCP. Commit `774c2cd70e` correctly scoped the wording to "inbound TCP session", but the spec still doesn't say what happens if they're set on a UDP frontend. ARM reviewers flag conditionally-applicable properties (OAPI020/OAPI029 family). Add "Ignored when protocol is UDP" or "Rejected with 400 when protocol is UDP" — whichever the RP actually does. Worth clarifying whether `loadDistribution` applies to UDP too.

**3. `sessionDrainTimeoutInMinutes` retains a client-side default while the idle timeouts had theirs removed.**

`models.tsp` L1286-L1292 keeps `= 15`. The reasoning is sound — PUT is full-replace so echoing the default is harmless, whereas a PATCH-reachable default causes SDKs to silently overwrite the server value — and it matches the existing `firewallSku = "STANDARD"` precedent. Two things to confirm anyway:

- Generated SDKs (Python/JS especially) initialize the property, so every SDK-originated create explicitly carries `15`. If PAN ever changes the service-side default, older SDKs pin the old value. If the service should own the default, drop `= 15` and keep the prose.
- `@minValue(15)` equals the default, so 15 is both floor *and* default. Nobody can ever request a faster drain. Lowering that floor later in a stable version is a breaking change — settle it now.

---

## 🔵 Info

- **Description wording.** "Default will be 15." → "Defaults to 15." reads better. The drain description is a comma splice: *"Can only be set on PUT, not updatable via PATCH, Default will be 15."* → *"Can only be set on PUT; not updatable via PATCH. Defaults to 15."* ("Default will be X" matches the existing local style, so this is optional.)
- **Naming asymmetry.** `NetworkProfile.outboundTcpReset` / `outboundIdleTimeoutInMinutes` vs. `FrontendSetting.tcpReset` / `idleTimeoutInMinutes`. Defensible — a frontend setting is inherently inbound — but if symmetry is wanted, rename now; post-GA renames are breaking.
- **Generic type names outside C#.** `TcpReset` and `LoadDistribution` become top-level swagger definitions and public SDK types. `FirewallTcpReset` / `FirewallLoadDistribution` were added for C# (`client.tsp` L142 and L255), matching this spec's existing pattern. Python/Java/JS/Go still get the generic names — consider naming them `FirewallTcpReset`/`FirewallLoadDistribution` in the TypeSpec itself and dropping the C# override so all languages benefit.
- **`TcpReset` duplicates the shape of `EnableStatus`** (Enabled/Disabled, `models.tsp` L307-L313). A distinct named type is the better call here for docs and independent evolution — just be ready for a reviewer to ask.
- **`LoadDistribution.Default`** as a value name ages poorly if the default ever changes, but Azure Load Balancer uses exactly `Default`/`SourceIP`/`SourceIPProtocol`, so precedent supports it. Extensible enum means `SourceIPProtocol` can be added later non-breaking.

---

## ✅ What's good

- All 6 properties and both enums are gated with `@added(Versions.v2026_08_27_preview)`, all optional. Zero impact on `2026-07-29-preview` and earlier — **non-breaking across the board**.
- **The PUT-only modeling for `sessionDrainTimeoutInMinutes` is correct.** It's in `FirewallDeploymentProperties` but deliberately absent from the hand-authored `FirewallResourceUpdateProperties`, so PATCH can't set it while GET still returns it. Removing the create-visibility in `87bdfee88b` was the right call — `@visibility(Lifecycle.Read, Lifecycle.Create)` would have emitted `x-ms-mutability: ["read","create"]` and wrongly declared it immutable after create, blocking legitimate PUT updates.
- Extensible `union … string` enums rather than booleans (RPC-BestPractice-12), and every member has a real doc comment — better than most pre-existing enums in this file, which carry `documentation-required` suppressions.
- Range constraints mirror Azure Standard Load Balancer exactly (inbound idle 4–100, outbound idle 4–120). Good alignment with the platform underneath.
- Examples are consistent and genuinely self-documenting: in `Firewalls_Update_MaximumSet_Gen.json` the PATCH request omits `sessionDrainTimeoutInMinutes` while the 200 response returns it — demonstrating PUT-only + readable in one file. All 5 MaximumSet examples updated; MinimumSet correctly untouched.
- `tsp compile` is clean under the ARM + client-sdk rulesets, and the checked-in `PaloAltoNetworks.Cloudngfw.json` is exactly in sync with the TypeSpec (recompile produced no diff).
- New `client.tsp` entries inserted in correct alphabetical position.

---

**Breaking-change assessment:** all changes are `@added`-gated into a brand-new preview version → non-breaking (no Rule 1020/1034/1033/1026/1045 hits).

**Verification caveat:** the repo's LintDiff (`lint-diff`) was run against the copy commit as baseline, but AutoRest could not install `@autorest/core` in the local environment, so the ARM linter result is unverified locally — CI covers it. The TypeSpec-level ARM linter did run clean.

specification/paloaltonetworks/resource-manager/PaloAltoNetworks.Cloudngfw/Cloudngfw/models.tsp
```diff
@@ -312,6 +312,42 @@ union EnableStatus {
...
+
+  /**
+   * Five-tuple hash: source IP, source port, destination IP, destination port and protocol
+   */
+  Default: "Default",
+
+  /**
+   * Two-tuple hash: source IP and destination IP
+   */
+  SourceIP: "SourceIP",
```
sorabhtandon, line 348: The comment mentions that its a hash but the field name suggest its source IP ?

Prajjwal-Singh12: `SourceIP` is Azure Load Balancer's value name, kept verbatim: `Microsoft.Network`'s `LoadDistribution` enum uses the same values (`Default`, `SourceIP`, `SourceIPProtocol`).

Per the [Load Balancer docs](https://learn.microsoft.com/en-us/azure/load-balancer/distribution-mode-concepts), `SourceIP` hashes source IP + frontend IP; and since the frontend IP is fixed, the source IP is what actually varies, hence the name.

nikhgup (approved):

PR #45932, opened 2026-08-28
Create Session + Get Layout (TypeSpec)

nickliu-msft:
Merge of these TypeSpec PRs: https://github.com/Azure/azure-rest-api-specs/pull/45371 & https://github.com/Azure/azure-rest-api-specs/pull/45476

### DIFF
--- a/specification/storage/data-plane/BlobStorage/Common/routes.tsp
+++ b/specification/storage/data-plane/BlobStorage/Common/routes.tsp
@@ -446,6 +446,28 @@ interface Container {
     {}
   >;
 
+  /** The Create Session operation enables users to create a session scoped to a container. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
+  @added(Versions.v2026_10_06)
+  @post
+  @route("?restype=container&comp=session")
+  createSession is StorageOperation<
+    {
+      ...TimeoutParameter;
+
+      /** The configuration used to create the session. */
+      @body
+      createSessionConfiguration: CreateSessionConfiguration;
+    },
+    {
+      @statusCode statusCode: 201;
+
+      /** The created session. */
+      @body
+      createSessionResponse: CreateSessionResponse;
+    }
+  >;
+
   /** Allows multiple API calls to be embedded into a single HTTP request. */
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Existing API"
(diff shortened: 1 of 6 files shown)


specification/storage/data-plane/BlobStorage/Common/routes.tsp
```diff
@@ -1592,6 +1616,65 @@ interface Blob {
...
+        ...ApiVersionHeader;
+        ...DateResponseHeaderPrivate;
+
+        /** The layout of the blob. */
+        @body
+        blobLayout: BlobLayout;
+      }
+    | {
+        #suppress "@azure-tools/typespec-azure-core/no-closed-literal-union" "Following standard pattern" Expand annotationCheck warning on line R1663
+        @statusCode statusCode: 204;
```
vincenttran-msft, line 1664: What is the expectation for when a 204 is hit?
Edit: I see in the dev-doc guide that a 204 is supposed to be fallback to default behavior (which is pretend like it's not a layout). Just confirming that this is true / if there is a case that the service really does return a 204?

nickliu-msft: 204 means that the service did not return a layout in the response body, but the property headers were still returned. As per the FE spec:

> "If it succeeds with HTTP 204, no layout information is available for the blob, and the API is functionally equivalent to Get Blob Properties.  If no layout information is available, clients should use standard GetBlob calls to download the blob."

You can read more about 204 in the [Data Locality Spec](https://microsoft.sharepoint.com/:w:/r/teams/AzureStorage/_layouts/15/Doc.aspx?sourcedoc=%7B10D0C1CE-C69C-467B-800E-3823E76C80A0%7D&file=Data%20Locality%20REST%20APIs.docx&nav=eyJjIjoxMDg0Njk4NDMzfQ&action=default&mobileredirect=true&share=IQHOwdAQnMZ7RoAOOCPnbICgAZFjPmG2yYtD2GGBzX1Em64&CID=019CD80E-6BFF-4EC4-8A09-56650C289AAA)

vincenttran-msft: Thanks for the confirmation. This has been a sticky point in Rust as the way pageable responses are modeled in `typespec-rust`, it currently eagerly attempts to XML deserialize any 2xx, and so this is actually is bubbling up as a `DataConversion` error and masks the underlying `204`, so my status code classifier isn't able to discern that this should be a "no layout" scenario.

PR #45938, opened 2026-08-28
Make TypeSpec Suppressions check required

catalinaperalta:
## Summary

Makes the `TypeSpec Suppressions` check merge-blocking instead of advisory/testing-mode.

Changes:
- Added `--fail-on-approval` to the `npm exec -- typespec-suppressions` invocation in `typespec-suppressions-code.yaml`, so the analyzer now exits non-zero when any suppression requires review/approval.
- Added `"TypeSpec Suppressions - Set Status"` to the `workflow_run.workflows` trigger list in `summarize-checks.yaml`, so the merge-summary recomputes when the suppression status changes.
- Added a `"TypeSpec Suppressions"` entry to `CHECK_METADATA` in `summarize-checks.js` (with the existing `Approved-TypeSpecSuppression` override label and the default troubleshooting guide), so once the check is added to the branch ruleset it will be surfaced correctly in the "Next Steps to Merge" comment and factored into "Automated merging requirements met".
- Removed "testing mode / non-blocking" language from `suppressions-comment.js`'s PR review comment and updated the approval-state/status text to reflect that the check now blocks merging.
- Updated `documentation/ci-fix.md` to remove the testing-mode callout, describe the check as blocking, and add `--fail-on-approval` to the local repro example.
- Updated existing unit tests (`post-results.test.js`, `suppressions-comment.test.js`) to match the new wording.

The existing `Approved-TypeSpecSuppression` label override mechanism (in `set-status.js` / `_reusable-set-check-status.yaml`) is unchanged and already works: applying the label force-succeeds the status; removing it re-evaluates.

## Required follow-up (not part of this PR's code changes)

For this check to actually **gate** merging on `main`, `"TypeSpec Suppressions"` must also be added to the `required_status_checks` list of the `Required PullRequest Checks` ruleset (id `912053`) — the same mechanism that already makes `TypeSpec Validation`, `Swagger BreakingChange`, etc. required. This is a manual, admin-only repo-settings change (not something a workflow-code PR can do) and should be done **after this PR merges**, so that:
1. The updated check-status workflow is live on `main` before it starts being enforced, and
2. Open PRs get a chance to re-run and produce a status before the check becomes required (to avoid PRs getting stuck on "Expected — waiting for status").

## Testing

- Validated the full change end-to-end on a personal fork with a demo PR:
  - Confirmed `--fail-on-approval` causes the analyzer to exit non-zero and the `TypeSpec Suppressions` commit status to be set to `failure` when suppressions require review.
  - Confirmed the PR review comment posts correctly with the updated (non-testing-mode) wording.
  - Confirmed the `Approved-TypeSpecSuppression` label override still works as expected.
- Ran the `.github` workflow test suite (`npm run check` in `.github/`) after updating test string expectations.

### DIFF
--- a/.github/workflows/src/typespec-suppressions/labels.js
+++ b/.github/workflows/src/typespec-suppressions/labels.js
@@ -1,0 +1,27 @@
+/**
+ * Remove a label from an issue/PR, ignoring 404 (label not present).
+ *
+ * Mirrors the package-name-approval helper of the same name
+ * (.github/workflows/src/package-name-approval/labels.js).
+ *
+ * @param {import("@actions/github-script").AsyncFunctionArguments["github"]} github
+ * @param {string} owner
+ * @param {string} repo
+ * @param {number} issueNumber
+ * @param {string} label
+ */
+export async function removeLabelIfPresent(github, owner, repo, issueNumber, label) {
+  try {
+    await github.rest.issues.removeLabel({
+      owner,
+      repo,
+      issue_number: issueNumber,
+      name: label,
+    });
+  } catch (error) {
+    if (error instanceof Error && "status" in error && error.status === 404) {
+      return;
+    }
+    throw error;
+  }
+}
--- a/.github/workflows/src/summarize-checks/labelling.js
+++ b/.github/workflows/src/summarize-checks/labelling.js
@@ -1193,6 +1193,27 @@ const rulesPri1Namespace = [
   },
 ];
 
+/** @type {RequiredLabelRule[]} */
+const rulesPri1TypeSpecSuppressions = [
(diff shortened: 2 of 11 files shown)


.github/workflows/src/typespec-suppressions/labels.js
catalinaperalta, line 1: this might benefit from being in a common dir

mikeharder: i'd put it here:

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/workflows/src/label.js

mikeharder (approved):

catalinaperalta:
@copilot resolve the merge conflicts in this pull request

PR #45955, opened 2026-08-31
Add AppContainers provisioning metadata and restore job resource name

ArcturusZhang:
## Changes
- add AppContainers provisioning name constraints and managed-environment RBAC metadata
- pin the .NET management emitter to API version `2026-01-01`
- restore `JobDetectorPropertyOps.Read` for `Jobs.proxyGet` so `Microsoft.App/jobs/detectorProperties` retains its distinct `ContainerAppJobDetectorProperty` resource name

## Context
The direct `ArmResourceRead<Job>` conversion removed the `ResourceName` supplied by `JobDetectorPropertyOps`, causing both `Microsoft.App/jobs` and `Microsoft.App/jobs/detectorProperties` to generate as `ContainerAppJobResource`. This overwrote the main job resource and produced duplicate .NET factory methods.

See the [original review comment on the change that introduced the regression](https://github.com/Azure/azure-rest-api-specs/pull/42504/changes#r3892136749).

The resource-name correction is also available as the standalone focused review [#45957](https://github.com/Azure/azure-rest-api-specs/pull/45957). This PR intentionally retains the same change so provisioning generation remains usable from this branch.

## Validation
- TypeSpec compilation succeeds
- .NET management generation is pinned to `2026-01-01`
- .NET regeneration produces `ContainerAppJobResource` for `Microsoft.App/jobs`
- .NET regeneration produces `ContainerAppJobDetectorPropertyResource` for `Microsoft.App/jobs/detectorProperties`
- the duplicate `GetContainerAppJobResource(ResourceIdentifier)` compilation errors are eliminated

The regenerated package still has an unrelated customization-drift error because `ArmAppContainersModelFactory` references the removed `ContainerAppRuntimeJavaAgent` type.

## Follow-up
We will introduce linters to prevent future modifications from replacing resource operation templates in a way that drops explicit resource naming metadata.


specification/app/resource-manager/Microsoft.App/ContainerApps/client.tsp
```diff
@@ -159,7 +159,25 @@ union OpenIdConnectClientCredentialMethod {
 
 @@alternateType(ManagedEnvironmentProperties.staticIp, ipV4Address, "csharp");
 
-@@alternateType(RegistryInfo.registryUrl, url, "csharp");
```
ArcturusZhang, line 162: why do we remove this alternateType?
maybe we should not change this.

ArcturusZhang:
The suppression is only added for client side options.

ArthurMa1978 (approved):

PR #46006, opened 2026-09-01
[CodeTransparency] Correct /jwks public verification-key contract

ryazhang-microsoft:
## Summary

Corrects the Code Transparency `/jwks` wire contract for stable API version `2026-03-26` so its public receipt-verification-key response does not advertise private or symmetric JWK members.

The service emits EC public JWKs (`kty`, `kid`, `crv`, `x`, `y`), while the previous schema also exposed `d`, `dp`, `dq`, `k`, `p`, `q`, and `qi`.

## Changes

- Preserves the existing `2025-01-31-preview` contract unchanged.
- Adds the stable-only `getPublicKeysV09` operation with canonical operation ID `GetPublicKeys`.
- Models `Content-Type` as a response header and the JWK set as the response body.
- Adds public-asymmetric JWK wire models containing only public key material.
- Regenerates the stable OpenAPI document and example without private-key fields.
- Keeps C# wire models internal and disables generated convenience methods so Azure/azure-sdk-for-net#62471 can expose normalized verification-key APIs while preserving protocol overloads.
- Removes Python SDK generation because Code Transparency does not plan to ship a Python SDK. The scoped `SdkTspConfigValidation` suppression covers the intentionally absent Python emitter options.

No new API version is introduced.

## Validation

- `npx tsv specification/confidentialledger/Microsoft.CodeTransparency`: passed
- TypeSpec compilation: passed
- Preview OpenAPI and example: unchanged from `main`

## Related

- Supersedes #46003, which was mistakenly stacked on the Confidential Ledger PR.
- .NET SDK PR: Azure/azure-sdk-for-net#62471


ryazhang-microsoft:
## Breaking-change review justification

This PR corrects the published contract for stable API version `2026-03-26`; it does not change deployed service behavior or the JSON payload returned by `GET /jwks`.

The checker reports two review-required changes:

1. **Remove output properties:** `d`, `dp`, `dq`, `k`, `p`, `q`, and `qi` are removed from the modeled response. These private or symmetric key fields have never been emitted by the service. The endpoint returns public receipt-verification keys only.
2. **Make `kid` required:** every deployed service response already includes `kid`, because it identifies the verification key. The specification previously modeled the existing runtime requirement as optional.

Therefore, existing callers receive the same payload before and after this change. The impact is limited to correcting the OpenAPI/TypeSpec contract and generated SDK surface so clients no longer expose impossible private-key fields or treat the always-present key identifier as nullable.

Additional scope controls:

- `2025-01-31-preview` remains unchanged.
- No endpoint, content type, JSON property name, or data type changes at runtime.
- No new API version is introduced.
- The paired .NET SDK correction is Azure/azure-sdk-for-net#62471.

Requested classification: **`Versioning-Approved-BugFix`**.

JoshLove-msft (approved):

specification/confidentialledger/Microsoft.CodeTransparency/main.tsp
```diff
@@ -63,6 +65,21 @@ op getPublicKeys is Foundations.Operation<
...
 
+#suppress "@azure-tools/typespec-azure-core/response-schema-problem" "cannot assign error to directive to cbor binary responses"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "IETF SCITT/SCRAPI operation exchanges raw COSE/CBOR binary payloads and does not map to Azure.Core standard resource operations."
+#suppress "@azure-tools/typespec-azure-core/no-openapi" "Preserving canonical GetPublicKeys operationId across versions"
+@added(Versions.`2026-03-26`)
+@operationId("GetPublicKeys")
+@sharedRoute
+@doc("Get the public keys used by the service to sign receipts, mentioned in IETF SCITT draft as part of jwks_uri implementation")
+@route("/jwks")
+op getPublicKeysV09 is Foundations.Operation<
```
JoshLove-msft, line 76: Can we rename this in the client.tsp?

specification/confidentialledger/Microsoft.CodeTransparency/main.tsp
```diff
@@ -63,6 +65,21 @@ op getPublicKeys is Foundations.Operation<
   AnyCborError
 >;
 
+#suppress "@azure-tools/typespec-azure-core/response-schema-problem" "cannot assign error to directive to cbor binary responses"
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "IETF SCITT/SCRAPI operation exchanges raw COSE/CBOR binary payloads and does not map to Azure.Core standard resource operations."
```
JoshLove-msft, line 69: Do we still need this suppression?

ryazhang-microsoft: Yes. `@@clientName` only controls the generated SDK-facing name; it does not preserve the OpenAPI `operationId`. The stable Swagger already uses `GetPublicKeys`, so removing `@operationId("GetPublicKeys")` would change that published metadata to the version-suffixed TypeSpec operation name. I kept the narrowly scoped suppression to avoid introducing an additional operationId change.

specification/confidentialledger/Microsoft.CodeTransparency/main.tsp
ryazhang-microsoft: Implemented in `client.tsp`. The stable operation is now exposed as `getPublicKeys` for C#, while the removed preview operation is assigned `getPublicKeysPreview` in the C# client graph to avoid a duplicate-name diagnostic. The TypeSpec symbols and wire contract remain unchanged.

ivarprudnikov (approved):

KPIyer:
Approved as bening.

PR #46017, opened 2026-09-01
Delay Foundry voice-agent APIs from v1

glecaros:
## Summary
- exclude the delayed Foundry voice-agent operations, models, discriminator values, and protocol event extensions from v1
- preserve the existing `VoiceAgents=V1Preview` opt-in key and restore all ten voice-agent examples
- comment out the OpenAI conversation-item `@@extension`/`@@copyProperties` customization section instead of materializing versioned copied members
- regenerate committed Foundry OpenAPI artifacts; v1 removes the delayed surface, while virtual preview reflects the disabled OpenAI conversation-item augmentations

## Validation
- Foundry `tsp compile --warn-as-error`
- `tsp format --check` for all 148 Foundry TypeSpec files
- v1 C#, Java, and Python/JavaScript SDK customization compiles with `--no-emit --warn-as-error`
- idempotent OpenAPI regeneration
- semantic v1 comparison: 13 delayed routes and 88 dependent schemas removed, with no paths or schemas added
- all ten voice-agent examples exactly restored
- separate Voice Live specification content unchanged

Full `tsv` was not rerun at the user's direction because its repository-wide Git status step hangs in this worktree; the targeted checks above passed.

ADO: https://msdata.visualstudio.com/Vienna/_workitems/edit/5578115


rhurey (approved):
:shipit:

nick863 (approved):

PR #46028, opened 2026-09-01
Add managed compute adapter deployment APIs

mdahamiwal:
## Summary
- add the `Microsoft.CognitiveServices/accounts/adapterDeployments` TypeSpec resource for the `2026-09-15-preview` API
- add create/replace, get, list, cancel, and delete operations with conditional request support
- use Foundry Project Models identifiers (`azureai://accounts/{account}/projects/{project}/models/{model}/versions/{version}`) and same-account `targetDeploymentName`
- add operation examples and regenerate the Cognitive Services OpenAPI

## Naming scope
- this PR defines the ARM management resource, whose path remains `adapterDeployments` under ARM naming conventions
- the PR 319 suggestion to use `adapter_deployments` applies only to the separate data-plane inference route, which is not defined by this management-plane PR

## Specification
- https://github.com/coreai-microsoft/foundrysdk_specs/pull/319

## Validation
- Cognitive Services TypeSpec compilation succeeds with warnings treated as errors
- repository TypeSpec validation succeeds


fmabroukmsft:
Automated analysis of this PR. Observations only, roughly by impact.

**`ifMatch` is required on DELETE.** It emits as `"required": true`, so callers that omit `If-Match` — portal, Terraform, `az`, ARM templates — will fail, and ARM cascade deletes (resource group / parent account) don't send it. Compounding this, the `201` branch of `CreateOrUpdateAdapterDeployment.json` returns no `etag`, so a client that just created an adapter needs a GET before it can delete. The PUT conditions alias makes both headers optional; the DELETE alias doesn't.

**The description lists a `cancel` operation that isn't here.** The interface defines `get`, `createOrUpdate`, `list`, `delete` — no cancel operation, example, or `cancel` member in `AdapterDeploymentOperationType`. Since `ConvergenceTimedOut` and `DrainTimedOut` are modeled, a re-target can run long, so it's worth confirming whether cancel was dropped or never intended.

**Enum casing is split within a single payload.** The four new unions use lowercase/snake_case (`"create"`, `"running"`, `"not_required"`, `"invalid_adapter_source"`) while `ProvisioningState` alongside them uses PascalCase. `GetAdapterDeployment.json` shows `"provisioningState": "Succeeded"` next to `"status": "succeeded"`. ARM management-plane convention is PascalCase, and this propagates into all six SDKs — cheap to change while in preview.

**`UpdateAdapterDeployment.json` documents a `201` for an existing resource.** It sends `If-Match: etag`, so `200` is the applicable branch; `201` is already covered by the create example.

**`AdapterDeploymentErrorDetail` duplicates common-types `ErrorDetail`** (`code`/`message`/`target`/`details[]`), differing only by constraining `code`. It adds a second error type to the SDKs alongside the `ErrorResponse` already on every `default` response.

**Example coverage is happy-path only** — nothing exercises `lastOperation.error` (21 error codes, zero examples), `rollbackResult`, `nextLink`, or the `If-None-Match: *` conflict.

Smaller items:

- `If-Match` example values are the literal string `"etag"`; response bodies use `"\"0x8D1234567890ABC\""`.
- The DELETE `202` example omits the `Azure-AsyncOperation` header the spec declares on that response.
- `sourceModelId` is documented as immutable and as an `azureai://` URI, but is a plain mutable `string` on a `CreateOrReplace` PUT with no `@pattern` — both constraints are runtime-only.
- `$filter` has no documented grammar or `@example`, and no error code covers a malformed filter.
- Gating note for a later cycle: `AdapterDeployment` has `@added(v2026_09_15_preview)` only, which is right today, but under this spec's single-track rule it graduates into the next stable automatically unless `@removed` is added when that version is cut.

fmabroukmsft (approved):

PR #46044, opened 2026-09-02
RecoveryServicesBackup: add 2026-10-01 API version with Instant Access snapshot policy props

Neelav-0211:
## Summary
 
Adds a new stable API version `2026-10-01` for `recoveryservicesbackup` (Microsoft.RecoveryServices/RecoveryServicesBackup), which is a superset of the existing `2026-08-01` stable version. This version introduces Instant Access snapshot policy support, additional recovery point diagnostic details, workload-snapshot filesystem metadata, and a few smaller additive fixes to existing workload-backup contracts.
 
### Instant Access (IA) snapshot policy support
- `AzureIaaSVMProtectionPolicy`: added `instantAccessSnapshotEnabled` (bool) and `instantAccessDurationMinutes` (int32, bounded 60–300) to enable and configure retention for Instant Access snapshots.
- `RecoveryPointTierType`: added a new tier value `IASnapshotRP` for recovery points retained under the policy's `instantAccessDurationMinutes`.
  - Converted `RecoveryPointTierType` from a closed `enum` to an open, string-extensible `union` (`modelAsString: true`), matching the existing `RecoveryPointTierStatus` pattern in the same file, so this and future tier values do not break existing SDK clients. Applied uniformly across all API versions (metadata-only change, no wire-value impact).
 
### New recovery point "extended info" operations (IaaS VM)
- New `backupFabrics/{fabricName}/getRPExtendedInfo` (POST) and
  `getRPExtendedInfoOperationResult/{operationId}` (GET) operations to trigger and
  retrieve additional recovery-point details (e.g., OS/data disk info) not returned
  by the existing recovery point GET API.
- New supporting models: `RPExtendedInfo` (+ `IaasVMRPExtendedInfo` subtype),
  `RPExtendedInfoResource(List)`, `GetRPExtendedInfoRequest(Resource)`.
- `DiskInformation`: added `diskSizeInGb` and `storageType`.
- `recoveryPointIds` constrained to exactly one id (`minItems`/`maxItems` = 1),
  matching what the service accepts today. The list shape is retained so the limit
  can be raised in a future api-version without a breaking change.
 
### Fabric-level operation status endpoint
- New `backupFabrics/{fabricName}/operationsStatus/{operationId}` (GET),
  `ProtectionContainerRefreshOperationStatuses_Get`, returning `OperationStatus`.
- Both operations above return this path in their `Azure-AsyncOperation` header
 
### Existing Basic VM protection handling
- `AzureIaaSVMProtectedItemProperties`: added `existingBasicVMProtection` (`ExistingBasicVMProtection` union) to control how pre-existing Basic-tier VM protection is handled when configuring protection (e.g., `DisableWithDeleteRPsNow`).
 
### SQL instance snapshot / workload restore changes
- Re-added (carried forward into 2026-10-01) several SQL-instance-snapshot properties that were previously scoped to `2026-05-31-preview` only: `ProtectionLevel`, `InstanceProtectionReadiness`, `RestorePointQueryType` (`Snapshot`, `FullDifferentialAndSnapshot`), `AzureVmWorkloadSQLInstanceProtectedItem`, `AzureVmWorkloadSQLDatabaseProtectedItem.parentProtectedItem`/`protectionLevel`, `DatabaseInRP`, `BmsrpQueryObject` (`baseSnapshotRPId`, `baseRPDatasourceId`, `getLatestRPOnly`).
- `SnapshotRestoreParameters`: added `disksToDetachOnClash` (disk ARM IDs to detach on filesystem clash).
- `AzureWorkloadSQLRecoveryPointExtendedInfo`: added `snapshotRecoveryPointInfo` with new supporting models `SnapshotRecoveryPointInfo`, `FilesystemInfo`, `StorageSpaceInfo`, `DiskInfo` describing per-volume filesystem/Storage Spaces/disk metadata captured on the VM for snapshot restore points.
- Added `ValidateAzureWorkloadRestoreOperationRequest` as a distinct discriminated subtype for Azure Workload (SQL/HANA/SAP ASE/AnyDatabase) restore validation.
 
### Container discovery
- `BMSRefreshContainersQueryObject`: added `containerSubscriptionId` to support cross-subscription container discovery.


specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/readme.md
```diff
@@ -41,7 +41,7 @@ tag: package-passivestamp-2023-01-15
 ```
 
 ```yaml $(package-activestamp)
-tag: package-2026-08-01
+tag: package-2026-10-01
```
gary-x-li, line 44: **[EXISTING] 🔴 Blocking** **[[MULTIPLE_DEFAULT_TAGS](https://github.com/Azure/avocado/blob/main/README.md#multiple_default_tags)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/readme.md` - line 44 - The AutoRest readme has more than one default tag.

**Classification reasoning:** The same three `tag:` blocks were already grouped under the `Basic Information` heading at base SHA `e16457ec7e3a852b849042ac03d802c3d9e777f7` (lines 31, 40, and 44), so this is pre-existing. This PR changes the values at lines 31 and 44, and the current Swagger Avocado check fails with `MULTIPLE_DEFAULT_TAGS` because all three blocks are still associated with `Basic Information`.

**Suggested fix:** Add a separate heading before the stamp-specific conditional blocks so only the global block is under `Basic Information`, for example:

````markdown
### Stamp Overrides

```yaml $(package-passivestamp)
tag: package-passivestamp-2023-01-15
```

```yaml $(package-activestamp)
tag: package-2026-10-01
```
````

File: specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp [hunk 2]
```diff
@@ -4426,6 +4426,18 @@ model AzureIaaSVMProtectionPolicy extends ProtectionPolicy {
...
+   * Specifies whether Instant Access snapshot is enabled for the policy.
+   */
+  @added(Versions.v2026_10_01)
+  instantAccessSnapshotEnabled?: boolean;
+
+  /**
+   * Duration in minutes for which the Instant Access snapshot is retained.
+   */
+  @added(Versions.v2026_10_01)
+  instantAccessDurationMinutes?: int32;
```
gary-x-li, line 4439: **[NEW] 🟠 Warning** **[[TypeSpec Section 2.2 - Type Constraints](https://github.com/Azure/azure-rest-api-specs/blob/main/.github/instructions/typespec-review.instructions.md#22-type-constraints-review-enforcement)]** `specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp` - line 4439 - `instantAccessDurationMinutes` has no lower-bound constraint.

**Classification reasoning:** This optional property is introduced in this PR and does not exist in `2026-08-01`. The generated schema therefore accepts zero and negative retention durations. The same-named Instant Access duration contract in `specification/compute/resource-manager/Microsoft.Compute/Compute/ComputeDisk/models.tsp` uses `@minValue(1)`, providing concrete precedent that the duration is positive.

**Suggested fix:** Add the positive minimum to the new property:

```tsp
@added(Versions.v2026_10_01)
@minValue(1)
instantAccessDurationMinutes?: int32;
```

gary-x-li (commented):
## ARM API Review

Posting findings from the [ARM API Reviewer agent](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/api-reviewer-agent.md) (critic-verified, 3 iterations, converged) against commit [`9689aba`](https://github.com/Azure/azure-rest-api-specs/pull/46044/commits/9689aba88a19d631fb01aad20e67a576f2b604a2). See inline comments for findings 1-2.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/readme.md
Neelav-0211: Stamp Overrides heading inserted before the two conditional stamp blocks

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Neelav-0211: Added checks for minimum : 60 and maximum : 300,  as values as per business requirement.

Neelav-0211:
> Neelav Bhatiya (@Neelav-0211) please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement
> ### Contribution License Agreement
> This Contribution License Agreement (**“Agreement”**) is agreed to by the party signing below (**“You”**), and conveys certain license rights to Microsoft Corporation and its affiliates (“Microsoft”) for Your contributions to Microsoft open source projects. This Agreement is effective as of the latest signature date below.
> 
> 1. **Definitions**.
>    **“Code”** means the computer software code, whether in human-readable or machine-executable form,
>    that is delivered by You to Microsoft under this Agreement.
>    **“Project”** means any of the projects owned or managed by Microsoft and offered under a license
>    approved by the Open Source Initiative ([www.opensource.org](http://www.opensource.org)).
>    **“Submit”** is the act of uploading, submitting, transmitting, or distributing code or other content to any
>    Project, including but not limited to communication on electronic mailing lists, source code control
>    systems, and issue tracking systems that are managed by, or on behalf of, the Project for the purpose of
>    discussing and improving that Project, but excluding communication that is conspicuously marked or
>    otherwise designated in writing by You as “Not a Submission.”
>    **“Submission”** means the Code and any other copyrightable material Submitted by You, including any
>    associated comments and documentation.
> 2. **Your Submission**. You must agree to the terms of this Agreement before making a Submission to any
>    Project. This Agreement covers any and all Submissions that You, now or in the future (except as
>    described in Section 4 below), Submit to any Project.
> 3. **Originality of Work**. You represent that each of Your Submissions is entirely Your original work.
>    Should You wish to Submit materials that are not Your original work, You may Submit them separately
>    to the Project if You (a) retain all copyright and license information that was in the materials as You
>    received them, (b) in the description accompanying Your Submission, include the phrase “Submission
>    containing materials of a third party:” followed by the names of the third party and any licenses or other
>    restrictions of which You are aware, and (c) follow any other instructions in the Project’s written
>    guidelines concerning Submissions.
> 4. **Your Employer**. References to “employer” in this Agreement include Your employer or anyone else
>    for whom You are acting in making Your Submission, e.g. as a contractor, vendor, or agent. If Your
>    Submission is made in the course of Your work for an employer or Your employer has intellectual
>    property rights in Your Submission by contract or applicable law, You must secure permission from Your
>    employer to make the Submission before signing this Agreement. In that case, the term “You” in this
>    Agreement will refer to You and the employer collectively. If You change employers in the future and
>    desire to Submit additional Submissions for the new employer, then You agree to sign a new Agreement
>    and secure permission from the new employer before Submitting those Submissions.
> 5. **Licenses**.
> 
> * **Copyright License**. You grant Microsoft, and those who receive the Submission directly or
>   indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license in the
>   Submission to reproduce, prepare derivative works of, publicly display, publicly perform, and distribute
>   the Submission and such derivative works, and to sublicense any or all of the foregoing rights to third
>   parties.
> * **Patent License**. You grant Microsoft, and those who receive the Submission directly or
>   indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license under
>   Your patent claims that are necessarily infringed by the Submission or the combination of the
>   Submission with the Project to which it was Submitted to make, have made, use, offer to sell, sell and
>   import or otherwise dispose of the Submission alone or with the Project.
> * **Other Rights Reserved**. Each party reserves all rights not expressly granted in this Agreement.
>   No additional licenses or rights whatsoever (including, without limitation, any implied licenses) are
>   granted by implication, exhaustion, estoppel or otherwise.
> 
> 6. **Representations and Warranties**. You represent that You are legally entitled to grant the above
>    licenses. You represent that each of Your Submissions is entirely Your original work (except as You may
>    have disclosed under Section 3). You represent that You have secured permission from Your employer to
>    make the Submission in cases where Your Submission is made in the course of Your work for Your
>    employer or Your employer has intellectual property rights in Your Submission by contract or applicable
>    law. If You are signing this Agreement on behalf of Your employer, You represent and warrant that You
>    have the necessary authority to bind the listed employer to the obligations contained in this Agreement.
>    You are not expected to provide support for Your Submission, unless You choose to do so. UNLESS
>    REQUIRED BY APPLICABLE LAW OR AGREED TO IN WRITING, AND EXCEPT FOR THE WARRANTIES
>    EXPRESSLY STATED IN SECTIONS 3, 4, AND 6, THE SUBMISSION PROVIDED UNDER THIS AGREEMENT IS
>    PROVIDED WITHOUT WARRANTY OF ANY KIND, INCLUDING, BUT NOT LIMITED TO, ANY WARRANTY OF
>    NONINFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
> 7. **Notice to Microsoft**. You agree to notify Microsoft in writing of any facts or circumstances of which
>    You later become aware that would make Your representations in this Agreement inaccurate in any
>    respect.
> 8. **Information about Submissions**. You agree that contributions to Projects and information about
>    contributions may be maintained indefinitely and disclosed publicly, including Your name and other
>    information that You submit with Your Submission.
> 9. **Governing Law/Jurisdiction**. This Agreement is governed by the laws of the State of Washington, and
>    the parties consent to exclusive jurisdiction and venue in the federal courts sitting in King County,
>    Washington, unless no federal subject matter jurisdiction exists, in which case the parties consent to
>    exclusive jurisdiction and venue in the Superior Court of King County, Washington. The parties waive all
>    defenses of lack of personal jurisdiction and forum non-conveniens.
> 10. **Entire Agreement/Assignment**. This Agreement is the entire agreement between the parties, and
>     supersedes any and all prior agreements, understandings or communications, written or oral, between
>     the parties relating to the subject matter hereof. This Agreement may be assigned by Microsoft.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -1545,6 +1549,10 @@ enum RecoveryPointTierType {
   InstantRP,
   HardenedRP,
   ArchivedRP,
+
+  /** Instant Access snapshot tier, retained for the policy's instantAccessDurationMinutes. */
+  @added(Versions.v2026_10_01)
+  IASnapshotRP,
```
dhruvesh: **Breaking change: adding a member to a closed enum.**

`RecoveryPointTierType` is emitted with `"modelAsString": false` in both `stable/2026-08-01/bms.json` and the new `stable/2026-10-01/bms.json`. Adding `IASnapshotRP` to a closed enum breaks existing generated SDK clients — they cannot deserialize the new value — and this should trip the breaking-change gate.

Two things to resolve:
1. Either flip this to an open enum (`union` with a `string` variant, i.e. `modelAsString: true`), which is itself a type-shape change requiring breaking-change approval, or obtain an explicit exemption for extending the closed enum.
2. Independently of the spec: confirm the service will not return `IASnapshotRP` to clients calling **older** API versions, otherwise those clients break regardless of what this file says.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
```diff
@@ -758,3 +759,100 @@ interface TieringCostOperationStatus {
     operationId: string,
   ): ArmResponse<OperationStatus> | ErrorResponse;
 }
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Legacy BMS route-based operation group; it is not an ARM resource operation group."
+@added(Versions.v2026_10_01)
+interface RecoveryPointsExtendedInfo {
```
dhruvesh, line 765: This adds two new operations to a **stable/GA** version that are not mentioned in the PR description (which only mentions the two Instant Access policy properties). Please expand the PR description so reviewers and the ARM review board can see the full GA surface being introduced here — two fabric-level operations plus ~15 new models.

On the operations themselves: the LRO shape looks correct (202 with `Location`, `Azure-AsyncOperation` and `Retry-After`, `x-ms-long-running-operation: true`, `final-state-via: location`, and no body on the 202), so no concern there.

One naming note: `getRPExtendedInfo` is a `POST` action whose name begins with `get`. ARM action naming generally expects a verb that reflects the mutation/trigger semantics. Given the description says this *triggers* a fetch, something like `triggerRPExtendedInfo` would read better — though I appreciate the name may already be fixed by the service implementation.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -5190,6 +5235,84 @@ model SnapshotRestoreParameters {
...
+   */
+  message?: string;
+}
+
+/**
+ * AzureWorkload-specific filesystem clash validation result. Carries aggregated
+ * managed disk IDs and the list of all detected clashes in a single error response.
+ */
+@added(Versions.v2026_10_01)
+model AzureWorkloadFilesystemClashValidationResult {
```
dhruvesh: **This model is unreachable and becomes a dead definition in the GA contract.**

`AzureWorkloadFilesystemClashValidationResult` appears exactly once in the generated `stable/2026-10-01/bms.json` — its own `definitions` entry — and nothing references it. `FilesystemClash` (and transitively `FilesystemClashType`) is reachable only through this model, so that whole cluster is dead weight.

Based on the doc comment ("Carries aggregated managed disk IDs and the list of all detected clashes in a single error response") the intent seems to be to surface this on the validate-restore error path. Either wire it up to that operation, or drop it from this version — once 2026-10-01 ships, an orphaned definition is permanent.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -4086,6 +4107,13 @@ alias AzureIaaSVMProtectedItemProperties = {
    */
   extendedProperties?: ExtendedProperties;
 
+  /**
+   * Specifies how existing Basic VM protection is handled when configuring protection.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "The property name matches the existing BMS wire contract."
+  @added(Versions.v2026_10_01)
+  existingBasicVMProtection?: ExistingBasicVMProtection;
```
dhruvesh, line 4120: **The suppression justification is not accurate.**

The justification states *"The property name matches the existing BMS wire contract."* I grepped every `bms.json` in this service across all `stable/` and `preview/` versions, and `existingBasicVMProtection` occurs **only** in the new `stable/2026-10-01/bms.json`. This is a brand-new property, so there is no pre-existing wire contract it needs to match.

Please either rename it to satisfy the `casing-style` rule (`existingBasicVmProtection`) or replace the justification with the real reason — e.g. consistency with the sibling `AzureIaaSVMProtectedItem` naming. A suppression whose stated reason is untrue is unlikely to pass ARM review.

Separately, `ExistingBasicVMProtection` is a single-valued union (`DisableWithDeleteRPsNow`) — an imperative command modelled as a resource property that will be echoed back on `GET`. At minimum mark it `@visibility(Lifecycle.Create, Lifecycle.Update)` so it is not returned on read; ideally move it to the configure-protection action input rather than the protected-item body.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -5587,6 +5851,113 @@ model DiskInformation {
   lun?: int32;
   #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
   name?: string;
+
+  /**
+   * Size of the disk in GB.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "Wire name is fixed by the existing BMS data contract, which spells the unit as GB."
+  @added(Versions.v2026_10_01)
+  diskSizeInGB?: int32;
```
dhruvesh, line 5794: **The suppression justification is not accurate.**

The justification states *"Wire name is fixed by the existing BMS data contract, which spells the unit as GB."* `diskSizeInGB` occurs **only** in the new `stable/2026-10-01/bms.json` — it does not appear in any earlier stable or preview version of this service. There is no existing data contract fixing this name.

Please either rename to `diskSizeInGb` to satisfy `casing-style`, or correct the justification to the actual rationale.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -5587,6 +5851,113 @@ model DiskInformation {
...
+  properties?: RPExtendedInfo;
+}
+
+/**
+ * List of the additional details of recovery points.
+ */
+#suppress "@azure-tools/typespec-azure-core/casing-style" "RP is an established acronym in the BMS contract; the model name matches the service data contract."
+#suppress "@azure-tools/typespec-azure-core/composition-over-inheritance" "Matches the existing BMS list envelope pattern used by RecoveryPointResourceList and ProtectedItemResourceList."
+@added(Versions.v2026_10_01)
+model RPExtendedInfoResourceList extends ResourceList {
```
dhruvesh, line 5860: **Returns a `nextLink` that clients cannot follow.**

`RPExtendedInfoResourceList extends ResourceList`, which contributes a `nextLink` property documented as *"The URI to fetch the next page of resources, with each API call returning up to 200 resources per page."* However, neither `getRPExtendedInfo` nor `getRPExtendedInfoOperationResult` is marked `x-ms-pageable` in the generated swagger, so callers receive a `nextLink` with no generated way to page.

Given the request accepts exactly one recovery point ID (see `GetRPExtendedInfoRequest.recoveryPointIds`), pagination looks unintended here. Consider not inheriting `ResourceList` so `nextLink` is not emitted at all. If paging genuinely is intended, add `x-ms-pageable`.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -5587,6 +5851,113 @@ model DiskInformation {
...
+ */
+#suppress "@azure-tools/typespec-azure-core/casing-style" "RP is an established acronym in the BMS contract; the model name matches the service data contract."
+#suppress "@azure-tools/typespec-client-generator-core/csharp-model-suffix" "Name kept in sync with the BMS service data contract."
+@added(Versions.v2026_10_01)
+model GetRPExtendedInfoRequest {
+  /**
+   * ARM id of the recovery point whose additional details are to be fetched. Exactly one recovery point id can be
+   * specified.
+   */
+  recoveryPointIds: string[];
```
dhruvesh, line 5894: The doc comment says *"Exactly one recovery point id can be specified"*, but the array is unconstrained, so the contract does not express that and nothing validates it at the API layer.

Please add `@minItems(1) @maxItems(1)`, or change this to a scalar `recoveryPointId: string` if exactly one is the permanent intent. Worth deciding deliberately, since relaxing to many later is easy but tightening is not.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -4426,6 +4454,20 @@ model AzureIaaSVMProtectionPolicy extends ProtectionPolicy {
...
+  @added(Versions.v2026_10_01)
+  instantAccessSnapshotEnabled?: boolean;
+
+  /**
+   * Duration in minutes for which the Instant Access snapshot is retained.
+   */
+  @added(Versions.v2026_10_01)
+  @minValue(60)
+  @maxValue(300)
+  instantAccessDurationMinutes?: int32;
```
dhruvesh, line 4474: Could you document the interdependency between these two properties? Specifically:

- What happens when `instantAccessDurationMinutes` is supplied while `instantAccessSnapshotEnabled` is `false` or absent — ignored, or a validation error?
- What is the default duration when `instantAccessSnapshotEnabled` is `true` but the duration is omitted?

The `@minValue(60)`/`@maxValue(300)` bounds are clear, but the 60–300 minute window itself is worth stating in the description so callers do not have to infer it from the generated schema.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -5379,8 +5502,146 @@ model AzureWorkloadSQLRecoveryPointExtendedInfo {
...
+   * Friendly name of the virtual disk.
+   */
+  virtualDiskFriendlyName?: string;
+}
+
+/**
+ * Identifiers for a physical disk that backs a volume.
+ */
+@added(Versions.v2026_10_01)
+model DiskInfo {
```
dhruvesh, line 5539: `DiskInfo` overlaps substantially with the existing `DiskInformation` model in this same spec — both carry LUN, size, and name/identifier style fields. Shipping two near-duplicate disk models in a GA version is likely to confuse SDK consumers.

Could you either extend/reuse `DiskInformation`, or clarify in the doc comment what distinguishes the two and why both are needed?

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -7207,6 +7579,21 @@ model ValidateRestoreOperationRequest extends ValidateOperationRequest {
   ...ValidateRestoreOperationRequestProperties;
 }
 
+/**
+ * Validate restore request for AzureWorkload (SQL/HANA/SAP ASE/AnyDatabase).
+ */
+#suppress "@azure-tools/typespec-client-generator-core/csharp-model-suffix" "Name kept for backward compatibility"
+@added(Versions.v2026_10_01)
+model ValidateAzureWorkloadRestoreOperationRequest
```
dhruvesh, line 7522: This appears structurally identical to the existing `ValidateRestoreOperationRequest` — same base type, same `...ValidateRestoreOperationRequestProperties` spread — differing only in the discriminator value.

If the sole purpose is to give AzureWorkload restores a distinct discriminator value, please say so explicitly in the doc comment. Otherwise it reads as an accidental duplicate, and it is permanent once this version ships.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
```diff
@@ -8435,6 +8822,12 @@ model BMSRefreshContainersQueryObject {
    * Backup management type for this container.
    */
   backupManagementType?: BackupManagementType;
+
+  /**
+   * Subscription containing the containers to discover.
+   */
+  @added(Versions.v2026_10_01)
+  containerSubscriptionId?: string;
```
dhruvesh, line 8765: This enables cross-subscription container discovery, which is worth being deliberate about in a GA contract:

1. The property is an unvalidated `string`. Consider a `@pattern` for a subscription GUID so callers get a clear validation error rather than a downstream failure.
2. Please confirm the authorization model — the caller's permissions are evaluated against the vault's subscription, but the containers being discovered live in a different subscription. It would be good to document what RBAC is required on `containerSubscriptionId`.

There is also no example covering this property (see the review summary).

dhruvesh (commented):
## ARM API review — 2026-10-01 (stable)

Thanks for the change. The headline Instant Access feature looks reasonable, but because **2026-10-01 is a stable/GA version** and therefore effectively immutable once shipped, I would like a few things resolved before merge.

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | Critical | `IASnapshotRP` added to a **closed** enum (`modelAsString: false`) — breaking for existing SDK clients |
| 2 | High | PR description materially understates the change |
| 3 | High | `AzureWorkloadFilesystemClashValidationResult` / `FilesystemClash` are unreachable — dead definitions in the GA contract |
| 4 | High | Two `casing-style` suppression justifications are factually incorrect |
| 5 | Medium | `RPExtendedInfoResourceList` returns `nextLink` with no `x-ms-pageable` |
| 6 | Medium | `ExistingBasicVMProtection` models an imperative command as a resource property returned on `GET` |
| 7 | Medium | `recoveryPointIds` documented as "exactly one" but unconstrained |
| 8 | Medium | Missing example coverage for several new properties |
| 9–12 | Low | `DiskInfo`/`DiskInformation` duplication; duplicate validate-request type; unvalidated `containerSubscriptionId`; undocumented property interdependency |

### On scope

The description says this adds `instantAccessSnapshotEnabled` and `instantAccessDurationMinutes`. It also adds two fabric-level operations (`getRPExtendedInfo`, `getRPExtendedInfoOperationResult`) and roughly 15 new models (`RPExtendedInfo`, `IaasVMRPExtendedInfo`, `SnapshotRecoveryPointInfo`, `FilesystemInfo`, `StorageSpaceInfo`, `DiskInfo`, `FilesystemClash`, …), plus new properties on `DiskInformation`, `SnapshotRestoreParameters` and `BMSRefreshContainersQueryObject`. Please expand the description accordingly.

### Example coverage

These new elements have **no** example coverage: `IASnapshotRP`, `disksToDetachOnClash`, `snapshotRecoveryPointInfo`, `containerSubscriptionId`. `instantAccessSnapshotEnabled` and `instantAccessDurationMinutes` each appear in only one example file.

### Checked and clear

- **LRO shape is correct** on the new POST: 202 with `Location`, `Azure-AsyncOperation`, `Retry-After`; `x-ms-long-running-operation: true`; `final-state-via: location`; no body on the 202.
- The `Invalid` sentinel in `FilesystemClashType` matches ~55 pre-existing occurrences in this spec, so it is consistent with local convention rather than a new defect.
- `readme.md` tag wiring, `service.yaml`, and the `main.tsp` version enum are correct and mutually consistent.
- The `@added(Versions.v2026_10_01)` re-additions of properties removed in 2026-07-01 generate valid swagger. Worth a sanity check with the service team that the 2026-07-01 removal was the error, not this restoration.

Every claim above was verified against the files in this PR and the generated `bms.json` rather than inferred.

dhruvesh (commented):
## Follow-up: additional findings (second review pass)

A deeper pass over the generated `bms.json`, the example set, and the PR's CI status surfaced several items my first review missed — including **two blocking issues**. Everything below was verified directly against the files and the live check results.

### New blocking issues

| # | Finding |
|---|---|
| A | `Azure-AsyncOperation` on both new operations targets `.../backupFabrics/{fabricName}/operationsStatus/{operationId}` — **a path that does not exist** in any of the 67 paths in `stable/2026-10-01/bms.json` |
| B | **`Swagger Avocado` is currently RED** on this PR — 10 × `MISSING_APIS_IN_DEFAULT_TAG` triggered by the default-tag advance (pre-existing gap, but it blocks merge) |

### Additional notes

**Stale `api-version` in example polling URLs.** Across the new `stable/2026-10-01/examples/` set, **101 embedded `api-version=` values in `Location` / `Azure-AsyncOperation` header URLs still reference older versions** (56 files contain embedded api-versions). Distribution:

| Count | api-version |
|---|---|
| 21 | `2016-12-01` |
| 20 | `2017-07-01` |
| 16 | `2026-08-01` |
| 12 | `2016-06-01` |
| 10 | `2025-02-01` |
| 8 | `2020-09-01` |
| 5 | `2020-07-01-preview` |
| 4 | `2020-06-01` |
| 4 | `2022-03-01` |
| 1 | `2019-05-13-preview` |

These are copy-paste carryovers. They do not fail validation, but they are customer-facing documentation showing a client polling a different API version than the one it called. Worth a sweep to normalize them to `2026-10-01`.

**Empty `x-ms-identifiers`.** The new `@identifiers(#[])` annotations (on `filesystemClashes`, `sourceFilesystemInfo`, `diskInfoList`, `dataDisks`) emit `"x-ms-identifiers": []`, which tells tooling the array elements have no identity key and disables ARM array-merge semantics on PATCH. This is usually the emitter default rather than a deliberate choice — please confirm each array genuinely has no identifying property. `diskInfoList` in particular has an obvious candidate in `diskUniqueId`/`lun`.

### Corrections to my earlier review

In the interest of accuracy, two things I flagged previously deserve softening:

- **`Invalid` sentinel in `FilesystemClashType`** — I noted this was consistent with local convention; to be precise, **60 pre-existing enums/unions in `models.tsp` already contain an `Invalid` member**. The new one is entirely conventional here. No change needed; it is only worth avoiding in future versions.
- **Version churn (`@added` → `@removed` → `@added`)** — I can now confirm this produces correct swagger. Verified programmatically: definitions 404 → 423, paths 65 → 67, **zero** property removals, **zero** enum narrowings. `2026-10-01` **is a proper superset** of `2026-08-01`. The only residual point is that a client moving `2026-07-01` → `2026-08-01` → `2026-10-01` sees fields disappear and reappear — worth a changelog note so it is not mistaken for a regression.

### Also verified clean

`x-ms-examples` coverage (all 84 operations have examples) · zero secret-bearing properties across all 19 new definitions · `readme.md` suppression continuity (the only change is whitespace normalization of the pre-existing `ConsistentPatchProperties` entry — nothing added or dropped) · `final-state-via: location` correct · the 200+202 response pair on the async POST is permitted · `x-ms-long-running-operation` correctly omitted on the GET poller · Swagger LintDiff, BreakingChange and ModelValidation all **pass**.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
```diff
@@ -758,3 +759,100 @@ interface TieringCostOperationStatus {
...
+    | ArmResponse<RPExtendedInfoResourceList>
+    | ArmAcceptedResponse<
+        "Accepted",
+        ArmLroLocationHeader &
+          Azure.Core.Foundations.RetryAfterHeader & {
+            /**
+             * A link to the shared fabric level operation status endpoint, which reports the status of this long
+             * running operation. The result itself is retrieved from the URL in the Location header.
+             */
+            @header("Azure-AsyncOperation")
```
dhruvesh, line 807: **Blocking: `Azure-AsyncOperation` points at a path this spec does not define.**

Both new operations emit an `Azure-AsyncOperation` header (here at line 807, and again at line 853 on `getRPExtendedInfoOperationResult`), and both new example files hard-code the polling URL as:

```
.../backupFabrics/Azure/operationsStatus/{operationId}?api-version=2026-10-01
```

I enumerated all **67** paths in the generated `stable/2026-10-01/bms.json`. The only fabric-scope operation path defined is:

```
.../backupFabrics/{fabricName}/operationResults/{operationId}
```

There is **no `operationsStatus` at `backupFabrics` scope**. The `operationsStatus` paths that do exist are at different scopes entirely — `protectedItems/{protectedItemName}/`, `backupTieringCost/default/`, and `privateEndpointConnections/{...}/`.

So a client that follows `Azure-AsyncOperation` polls an endpoint the contract never describes, and no SDK will generate a poller for it. ARM requires the `Azure-AsyncOperation` URL to return `200` with a status object whose terminal values include `Succeeded`/`Failed`/`Canceled` — an undefined path cannot satisfy that.

The RP already has the right precedent: `backupTieringCost/default/` defines **both** halves — the trigger *and* its `operationsStatus` endpoint (`TieringCostOperationStatus` in this file, returning `ArmResponse<OperationStatus>`).

**Fix — pick one:**
- **(a) Preferred.** Add a `.../backupFabrics/{fabricName}/operationsStatus/{operationId}` GET returning an ARM async-operation status object, mirroring `TieringCostOperationStatus`.
- **(b)** Drop the `Azure-AsyncOperation` header from both operations and rely on `Location` alone — which *does* correctly target the defined `getRPExtendedInfoOperationResult` path — and fix both example files to match.

Note the `Location` header and `final-state-via: location` are correct as-is; this finding is only about the `Azure-AsyncOperation` target.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/stable/2026-10-01/examples/AzureIaasVm/TriggerGetRPExtendedInfo.json
```diff
@@ -0,0 +1,55 @@
...
+                }
+              ]
+            }
+          }
+        ]
+      }
+    },
+    "202": {
+      "headers": {
+        "Azure-AsyncOperation": "https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRG/providers/Microsoft.RecoveryServices/vaults/testVault/backupFabrics/Azure/operationsStatus/00000000-0000-0000-0000-000000000000?api-version=2026-10-01",
```
dhruvesh, line 47: This `Azure-AsyncOperation` value (and the matching one at line 41 of `GetRPExtendedInfoOperationResult.json`) points at `.../backupFabrics/Azure/operationsStatus/{operationId}`, which is **not a path defined in `stable/2026-10-01/bms.json`**. See the detailed comment on `routes.tsp` line 807.

Whichever way that is resolved, both example files need to be updated to match — right now they document a polling flow a client cannot follow.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/readme.md
```diff
@@ -28,20 +28,22 @@ These are the global settings for the RecoveryServicesBackup API.
 title: Recovery Services Backup Client
 description: Open API 2.0 Specs for Azure RecoveryServices Backup service
 openapi-type: arm
-tag: package-2026-08-01
+tag: package-2026-10-01
```
dhruvesh, line 31: **Blocking: this default-tag advance leaves 10 API paths uncovered and the required `Swagger Avocado` check is currently RED.**

[Run 35053695027](https://github.com/Azure/azure-rest-api-specs/actions/runs/35053695027) — job "Swagger Avocado - Analyze Code" — fails with `errors: 10`, all `MISSING_APIS_IN_DEFAULT_TAG`. Advancing the default tag here (and at line 46) to `package-2026-10-01` makes `stable/2026-10-01/bms.json` the default, and it does not contain 10 paths that older shipped versions still define:

| Path | Genuinely absent? |
|---|---|
| `.../locations/{}/backupaadproperties` | Yes |
| `.../locations/{}/backupcrossregionrestore` | Yes |
| `.../locations/{}/backupcrrjob` | Yes |
| `.../locations/{}/backupcrrjobs` | Yes |
| `.../locations/{}/backupcrroperationresults/{}` | Yes |
| `.../locations/{}/backupcrroperationsstatus/{}` | Yes |
| `.../protecteditems/{}/recoverypoints/{}/accesstoken` | Yes |
| `.../vaults/{}/registeredidentities/{}` | Yes |
| `.../protecteditems/{}/recoverypoints/` | No — trailing-slash artifact |
| `.../vaults/{}/backupprotecteditems/` | No — trailing-slash artifact |

The last two are legacy authoring artifacts: `2026-10-01` defines both paths *without* the trailing slash; the trailing-slash forms exist only in `2018-12-20`, `2021-11-15` and `2023-01-15`.

**To be fair on classification:** this is a **pre-existing** gap, not one this PR introduces. All 10 are equally absent from `stable/2026-08-01`. The TypeSpec conversion never carried the CRR / accessToken / registeredIdentities APIs forward, so moving the default tag from one incomplete version to another merely re-triggers the same failure. It is still a red required check that blocks merge.

**Fix — pick one and state which in the PR:**
- **(a) Preferred.** Carry the 8 genuinely-absent operations forward into the TypeSpec so `2026-10-01` is a true superset of the live surface. Reconcile the last two by normalizing the legacy trailing-slash declarations, not by adding duplicate paths.
- **(b)** Leave the default tag on a tag that does cover all APIs.
- **(c)** If these APIs are intentionally retired, document that and obtain deprecation approvals — dropping GA operations from the default tag is customer-visible.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
dhruvesh: Following up on my earlier comment here: the reason this enum is closed in the first place is worth looking at.

The `#suppress "@azure-tools/typespec-azure-core/no-enum"` on the `enum RecoveryPointTierType` declaration (line 1546, just above this hunk) still carries the generated placeholder text — *"FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"*. So the closed-enum shape (`"modelAsString": false`) was never a deliberate design decision; it is unreviewed scaffolding left over from the Swagger→TypeSpec conversion, and this PR is now adding a value on top of it.

Note also that this PR carries the **`BreakingChange-JavaScript-Sdk`** label, and this closed-enum widening is the most likely cause. For contrast, `RestorePointQueryType` also gains values in this PR but is a `union` (`modelAsString: true`), so it is safe — this is the only closed-enum widening here.

**Fix:** convert `RecoveryPointTierType` to a `union` so it emits `modelAsString: true` and future values are non-breaking, and replace the FIXME with a real justification. This is the moment to do it — once `2026-10-01` ships GA, the closed shape is frozen permanently.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
```diff
@@ -758,3 +759,100 @@ interface TieringCostOperationStatus {
     operationId: string,
   ): ArmResponse<OperationStatus> | ErrorResponse;
 }
+
+#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-interface-requires-decorator" "Legacy BMS route-based operation group; it is not an ARM resource operation group."
```
dhruvesh, line 763: Two smaller points on this new interface.

**1. Suppression justification.** Credit where due — this is one of the few suppressions in this file with real text rather than the generated `FIXME` placeholder (the other 23 interfaces here all still carry `"FIXME: Update justification..."`). That said, "consistent with 23 unreviewed FIXMEs" is thin support for a new GA addition, so it is worth making sure the stated reason genuinely holds for this interface.

**2. Fabric-scope placement — a question, not a defect.** `getRPExtendedInfo` acts on recovery points but is routed at `backupFabrics/{fabricName}` scope, above the protected item that actually owns them. That is defensible if the operation genuinely spans containers, but it means RBAC is evaluated at fabric scope for an operation whose subject is a single recovery point (the request accepts exactly one id). Could you confirm fabric-scope permissions are the intended authorization boundary? This is not provable from the spec alone.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
iwrinshow: containerSubscriptionId is part of the $filter expression rather than a standalone wire parameter. BMSRefreshContainersQueryObject is retained because AutoRest uses this filter model to generate SDK and PowerShell parameters. Therefore, a TypeSpec pattern would not validate the actual $filter value; validation is performed by the service after parsing the expression. We added a cross-subscription refresh example.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
iwrinshow: This subtype provides the distinct objectType discriminator used for Azure Workload restore validation, updated the comment to explicitly specify this.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
iwrinshow: DiskInfo describes physical disks backing filesystem volumes, including GUIDs, location, byte size, friendly name, and managed-disk ID, while DiskInformation identifies included/excluded recovery-point disks using only lun and name. Clarified the same in the comment.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
iwrinshow: Removed it.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Neelav-0211: Done. Flipped to open.

Neelav-0211:
Thanks for the detailed context - done.

**Fix applied** (commit a4068550437): converted `RecoveryPointTierType` from a closed `enum` to an open `union` (with a `string` catch-all variant), matching the existing `RecoveryPointTierStatus` pattern in the same file. `IASnapshotRP` stays gated behind `@added(Versions.v2026_10_01)`. Verified via `tsp compile .` that this flips `modelAsString: false to true` in all 11 emitted `bms.json` files (only that one line changes per file) - a non-breaking relaxation, no other schema content changed.

**New issue this surfaced**: because `RecoveryPointTierType` is a shared model, `tsp compile` necessarily regenerates every historical version's `bms.json` (not just 2026-10-01). That pulls files like `stable/2026-02-01/bms.json` into LintDiff's "affected swaggers" scope for this PR, even though their only change is the same 1-line `modelAsString` flip. Once in scope, LintDiff's before/after violation diff reports dozens of pre-existing, unrelated legacy violations (e.g. `AvoidAdditionalProperties` on schemas like `DpmJobExtendedInfo`, `MabJobExtendedInfo` - untouched by this PR) as "new," including several at Level: error, which fails the Swagger LintDiff check.

I have root-caused this as a LintDiff tooling artifact (verified: before/after content for `stable/2026-02-01/bms.json` is identical except the same 1-line flip, yet before reports 11 unique `AvoidAdditionalProperties` violations and after reports 22 for that file) rather than anything wrong with the enum fix itself.

Given this is a known consequence of widening a shared closed enum, could someone from the API review / breaking-change team advise on how to get past this - e.g. is there a precedent for requesting a LintDiff baseline exception in this situation, or a different approach we should take? Flagging for manual guidance since we do not see a clean fix on our end.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Neelav-0211: Done - converted `RecoveryPointTierType` to a `union` (open enum, `modelAsString: true`) in commit a4068550437, matching `RecoveryPointTierStatus`. Replaced the stale FIXME suppression justification too. `IASnapshotRP` remains gated via `@added(Versions.v2026_10_01)`.

One follow-up: since this is a shared model, regenerating it touched all 11 historical `bms.json` files (1-line `modelAsString` flip each), which pulled older files like `stable/2026-02-01/bms.json` into Swagger LintDiff's scope and it's now flagging pre-existing, unrelated legacy violations there as "new." Left a fuller writeup on the PR - would appreciate guidance on how to get past that check given it's an expected side effect of this fix.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
Prabhkiratnitp: As no action is being performed on recovery point. We are fetching metadata for that particular recovery point id. Hence, we went ahead with this name instead. 

WE deliberately kept this as Vault Scoped API, as we foresee this API being used for our upcoming features where we expect multiple recovery point ids being passed in request body. 

@dhruvesh

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Prabhkiratnitp: renamed it to diskSizeInGb

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Prabhkiratnitp: added @minItems(1) @maxItems(1)

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
Prabhkiratnitp: Linked Access Checks will be applied both on the Recovery Services Vault and PecoveryPoint

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/readme.md
Prabhkiratnitp: @dhruvesh - this is a pre-existing gap for CRR endpoints. CRR Swagger TypeSpec Migration is in progress. We can fix this post that. Should not be a blocker for this PR

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/routes.tsp
Prabhkiratnitp: added this GET endpoint

Prabhkiratnitp: I was re-using the existing endpoint, which was already implemented and shipped by the Service, not sure how this was missed

@dhruvesh

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/stable/2026-10-01/examples/AzureIaasVm/TriggerGetRPExtendedInfo.json
Prabhkiratnitp: added this endpoint

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
dipenSingh19: Hi @dhruvesh 

Updated the suppression with the actual rationale as `existingBasicVMProtection` is a new property in the 2026-10-01 API version and we intend to retain the spelling to preserve the established 'VM' acronym casing used by the Recovery Services Backup contracts.

The existing public configureProtection API is the protected-item CreateOrUpdate PUT. 
Introducing a separate action would require a new public API operation and corresponding service changes. 
We have therefore retained it as an input-only property on the existing PUT contract and marked it 
`@visibility(Lifecycle.Create, Lifecycle.Update)`

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Neelav-0211: Update: per an offline team decision, narrowed the scope of the fix further. `RecoveryPointTierType` now stays fully closed (`modelAsString: false`) for every version before 2026-10-01 — identical to their originally committed state (10 historical `bms.json` files revert cleanly, 1-line each). Only 2026-10-01 gets `IASnapshotRP` plus the `string` catch-all, both gated via `@added(Versions.v2026_10_01)` on the union variants within the same `RecoveryPointTierType` union — no shared model duplication needed.

This confines the open-enum shape (and the breaking-change surface) to 2026-10-01 alone, so no unrelated older swagger is touched. Commit `bcf5e43b423`.

On your second point: confirmed with the service team — `IASnapshotRP` will not be returned to clients calling API versions prior to 2026-10-01.

Neelav-0211: Update: after an offline discussion, we narrowed this further — rather than opening `RecoveryPointTierType` across all versions, only the 2026-10-01 projection is open now. The union's `string` catch-all and `IASnapshotRP` are both gated with `@added(Versions.v2026_10_01)` inside the same union, so all 10 pre-2026-10-01 `bms.json` files revert to their exact original committed state (closed enum, unchanged FIXME suppressions on the 4 pre-existing members untouched — that legacy conversion scaffolding is out of scope for this fix). Only `stable/2026-10-01/bms.json` shows `modelAsString: true`. Commit `bcf5e43b423`.

This should remove `RecoveryPointTierType` as a `BreakingChange-JavaScript-Sdk` trigger for the pre-2026-10-01 versions, since nothing in their swagger changes at all now, while 2026-10-01 — a brand-new version — is free to introduce the open shape without breaking an existing shipped contract.

specification/recoveryservicesbackup/resource-manager/Microsoft.RecoveryServices/RecoveryServicesBackup/models.tsp
Neelav-0211: Documented both, and pushed in commit `f4b5c0200e9`:

- `instantAccessDurationMinutes` is ignored when `instantAccessSnapshotEnabled` is `false` or omitted — no Instant Access snapshot is retained in that case.
- Default duration is **300 minutes** when `instantAccessSnapshotEnabled` is `true` but `instantAccessDurationMinutes` is omitted.
- The 60–300 minute window is now stated explicitly in both property doc comments (in addition to the existing `@minValue(60)`/`@maxValue(300)`), and the PR description has been updated to call out the default.

PR #46063, opened 2026-09-02
feat(evaluations): add latency and cost schemas

shrutiyer:
Define evaluation target latency percentiles and estimated inference cost breakdowns on evaluation run responses.

Regenerate the v1 and virtual preview OpenAPI artifacts so the new response structures are available to clients.

Authored-by: GitHub Copilot CLI v1.0.81-0
Model: GPT-5.6 Sol (gpt-5.6-sol)

# Choose a PR Template

### DIFF
--- a/specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
+++ b/specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
@@ -719,6 +719,96 @@ model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
   }
 );
 
+/** Latency measurements reported for an evaluation run. */
+model EvalRunLatency {
+  /** End-to-end latency percentiles for the evaluation target. */
+  target?: LatencySummary;
+}
+
+/** Run-wide end-to-end latency percentiles for the evaluation target. */
+model LatencySummary {
+  /** Median target latency in milliseconds. */
+  @minValue(0)
+  p50_ms: float64;
+
+  /** 95th percentile target latency in milliseconds. */
+  @minValue(0)
+  p95_ms: float64;
+
+  /** Number of evaluation rows that contributed a usable latency measurement. */
+  @minValue(1)
+  sample_count: int64;
+}
+
+/** Estimated costs reported for an evaluation run. */
+model EvalRunCost {
+  /** Estimated inference cost for the evaluation target. */
+  target?: InferenceCostSummary;
+}
+
+/** Indicates whether all target models attributed to the evaluation run were priced. */
+union InferenceCostCompleteness {
+  string,
+
+  /** All attributed target models were priced. */
+  complete: "complete",
+
+  /** At least one attributed target model could not be priced. */
+  partial: "partial",
+}
+
+/**
+ * Run-wide estimated inference cost for the evaluation target.
+ *
+ * The estimate is derived from reported token counts and list prices. It excludes evaluator
+ * and runtime usage and does not account for negotiated pricing, commitments, or discounts.
+ */
+model InferenceCostSummary {
+  /** Estimated target inference cost in the specified currency. */
+  @minValue(0)
+  estimated_cost: float64;
+
+  /** ISO 4217 currency code for the estimate. */
+  currency: string;
+
+  /** Whether all attributed target models were priced. */
+  completeness: InferenceCostCompleteness;
+
+  /** Identifier of the price-list snapshot used for the estimate. */
+  pricing_version?: string;
+
+  /** Estimated cost and token usage for each priced target model. */
+  model_costs: ModelCost[];
+
+  /** Target models for which no reliable price was available. */
+  unpriced_models?: string[];
+}
+
+/** Estimated inference cost and token usage for one target model. */
+model ModelCost {
+  /**
+   * Backing model name used for pricing. For a model-router target, this is the model
+   * attributed by the runtime. For a direct deployment, this is resolved from deployment metadata.
+   */
+  model_name: string;
+
+  /** Estimated inference cost for this model. */
+  @minValue(0)
+  estimated_cost: float64;
+
+  /** Number of non-cached input tokens. */
+  @minValue(0)
+  prompt_tokens: int64;
+
+  /** Number of cached input tokens. */
+  @minValue(0)
+  cached_tokens: int64;
+
+  /** Number of output tokens. */
+  @minValue(0)
+  completion_tokens: int64;
+}
+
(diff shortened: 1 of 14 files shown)


specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -744,6 +744,88 @@ model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
...
+}
+
+/** Estimated costs reported for an evaluation run. */
+model EvalRunCost {
+  /** Estimated inference cost for the evaluation target. */
+  target?: TargetInferenceCostSummary;
+}
+
+/** Indicates whether all target models attributed to the evaluation run were priced. */
+union TargetInferenceCostCompleteness {
```
posaninagendra: Could we drop the prefix "Target" on model names as they could generally be applicable for other model types used in the eval? Imagining if we might want to expand this usage for LLM as judge model etcs.,

shrutiyer: Updated!

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -744,6 +744,88 @@ model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
...
+
+/**
+ * Run-wide estimated inference cost for the evaluation target.
+ *
+ * The estimate is derived from reported token counts and list prices. It excludes evaluator
+ * and runtime usage and does not account for negotiated pricing, commitments, or discounts.
+ */
+model InferenceCostSummary {
+  /** Estimated target inference cost in the specified currency. */
+  estimated_cost: decimal128;
```
glecaros: is a decimal128 really needed here?

shrutiyer: Switched it and fixed away from this, reflecting the agents team's cost type float64 in  AgentInsightEstimatedCost.amount

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -744,6 +744,88 @@ model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
   }
 );
 
+/** Latency measurements reported for an evaluation run. */
+model EvalRunLatency {
+  /** End-to-end latency percentiles for the evaluation target. */
+  target?: LatencySummary;
```
glharper, line 725: (Beginner question) So, because target is optional in this model, the service could add an empty object to EvalRun.latency, instead of just not adding a latency field? Why would the empty object be preferable?

(Same question applies to EvalRunCost below, but I'm assuming the answer's the same.)

shrutiyer: Hi Glenn, if there are no costs of latencies to add, we will not add anything at all (no empty field)
```
  /** Latency measurements observed during the evaluation run. */
  latency?: EvalRunLatency;

  /** Estimated costs observed during the evaluation run. */
  estimated_cost?: EvalRunCost;
```
(lmk if I misunderstood the question or miswrote this spec, but the intention is to not add anything empty. We only add these fields if there are costs and latencies to add)

glharper: Thanks! So why is target optional here, if it will always be present?

shrutiyer: We have plans to add other types of cost here. Say for judge model, that will only appear if judge models were used. If target wasnt used in that scenario with judge, then target field should not appear. Hence target is optional. Hope my explanation makes sense for why it is still optional.

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
```diff
@@ -719,6 +719,88 @@ model AzureAIBenchmarkPreviewEvalRunDataSource extends EvalRunDataSource {
...
+/** Run-wide end-to-end latency percentiles for the evaluation target. */
+model LatencySummary {
+  /** Median target latency in milliseconds. */
+  p50_ms: float64;
+
+  /** 95th percentile target latency in milliseconds. */
+  p95_ms: float64;
+
+  /** Number of evaluation rows that contributed a usable latency measurement. */
+  sample_count: int64;
```
dargilco, line 740: Please add @minValue decorators where it makes sense. For example, for the above, not clear to me if the min value can be 0 or 1

glharper (approved):

specification/ai-foundry/data-plane/Foundry/src/openai/evaluations/models.tsp
shrutiyer: Added, thanks!

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
```diff
@@ -82,7 +82,9 @@ interface Toolboxes {
       @header contentType: string;
 
       /** The MCP request body. */
-      @body request: Record<unknown>;
+      @alternateType(unknown, "java")
```
dargilco, line 85: We usually put client decorators in the relevant client.tsp file. @kaylieee can you look into this?

specification/ai-foundry/data-plane/Foundry/src/sdk-java-azure-ai-agents/client.tsp
```diff
@@ -24,8 +24,6 @@ using TypeSpec.Http;
 using Azure.AI.Projects;
 
 namespace Azure.AI.Projects.Beta {
-  interface AgentEndpointConversations {}
```
dargilco, line 27: @kaylieee  to review

kaylieee: lgtm

specification/ai-foundry/data-plane/Foundry/src/toolboxes/routes.tsp
kaylieee: This is a problematic type for Java. We can't alternate type the parameter in the client.tsp so have to make manual customization for it currently. The decorator works here, if we can leave this customization here in the upstream that would be preferred.

kaylieee (approved):

glecaros (approved):

PR #46133, opened 2026-09-04
Dataplane app microsoft.app preview/2026 09 01 preview

RichardChen820:

> [!TIP]

## PR review workflow diagram

## API Info: The Basics

- [ ] a private preview
- [ ] a public preview
- [ ] GA release

### Change Scope

<sup>This section will help us focus on the specific parts of your API that are new or have been modified. <br/>Please share a link to the design document for the new APIs, a link to the previous API Spec document (if applicable), and the root paths that have been updated. </sup>
* Design Document:
* Previous API Spec Doc:
* Updated paths:

### Viewing API changes

### Suppressing failures

[Swagger-Suppression-Process](https://aka.ms/pr-suppressions) 
to get approval.

### Release planner

A [release plan](https://aka.ms/azsdkdocs/release-plans) should have been created. If not, please create one as it will help guide you through the REST API and SDK creation process. 

<details>

</details>

<details>

### Tooling

### Helpful Links

</details>

## Getting help

  and https://aka.ms/ci-fix.


specification/app/data-plane/ContainerApps/suppressions.yaml
```diff
@@ -0,0 +1,8 @@
+- tool: TypeSpecValidation
+  reason: Python and JavaScript SDK generation are intentionally excluded until their package namespaces are approved and reserved.
+  rules: [SdkTspConfigValidation]
+  sub-rules:
+    - options.@azure-tools/typespec-python.*
+    - options.@azure-tools/typespec-ts.*
```
RichardChen820, line 6: We need them, but the CI complains that python package name is not registered on PyPI.org yet, need to reserve the name first. So I just excluded it temporarily.

Do we need to start the reservation process at this point?

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,137 @@
...
+        @query labels?: string;
+      },
+      SandboxListResult,
+      DataPlaneStandardTraits,
+      ListErrors
+    >;
+
+    #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Azure.Core StandardResourceOperations.ResourceCreateOrReplace cannot preserve the existing SandboxContent body, 201 response, and custom error status shapes without changing the established wire contract."
+    @doc("Creates a sandbox in a scope.")
+    @put
```
RichardChen820: From @weidongxu-microsoft 

Please keep this as `PUT`, but model it as a standard resource create-or-replace operation. Add the sandbox identifier as the resource `@key`, include that key as a path parameter (`.../sandboxes/{id}`), and use the same resource model shape for the request and response bodies, with lifecycle visibility for response-only properties. The current collection-level `PUT` and separate `SandboxRequest`/`SandboxFields` body shapes do not provide standard resource-addressed `PUT` semantics.

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
```diff
@@ -0,0 +1,1887 @@
...
+  action: EgressPolicyActionType;
+
+  /** protocol value. */
+  protocol: TransportProtocol;
+
+  /** destination value. */
+  destination: string;
+
+  /** port value. */
+  port: int32;
```
RichardChen820, line 1848: From @weidongxu-microsoft 

The model documentation says port may be null to match every port, but port: int32 emits as required and non-null. Please align the contract and documentation. If null is the wildcard, model it as nullable and add the documented 1..65535 constraint for integer values; otherwise remove the nullable behavior from the documentation.

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,137 @@
+using Http;
+using Azure.Core;
+
+namespace ContainerAppsSandbox {
+  /** Successful Sandbox list response. */
+  model SandboxListResult {
+    /** HTTP response status code. */
+    @statusCode statusCode: 200;
```
RichardChen820, line 9: From @weidongxu-microsoft 

200 does not need to be specified at all for normal GET

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,137 @@
...
+  alias GetErrors = ErrorResult<401> | ErrorResult<404> | ErrorResult<500>;
+
+  /** Errors returned when deleting a sandbox. */
+  alias DeleteErrors = ErrorResult<401> | ErrorResult<500>;
+}
+
+namespace ContainerAppsSandbox {
+  /** Azure-scoped Sandbox CRUD operations. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Azure.Core StandardResourceOperations.ResourceList cannot preserve the existing Page, PageSize, and labels query parameter names or the array response without changing the established wire contract."
+  #suppress "@azure-tools/typespec-azure-core/use-standard-names" "This list operation uses a custom response model to preserve the existing bare-array wire contract, which the rule misclassifies as a single-object response."
```
RichardChen820: From @weidongxu-microsoft 

These should be doable via resource and trait.

And you should not use SandboxListResponse as 200 is the default status code for GET.

It requires sufficient reason if you don't use standard op.

Same for other suppress on use-standard-operations

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
```diff
@@ -0,0 +1,1887 @@
...
+}
+
+/** Pod-level content package declaration. Declares a content package that the pod uses. Per-container downloads reference content packages by ContentPackageId PodContentPackage.ContentPackageId. */
+model PodContentPackage {
+  /** contentPackageId value. */
+  contentPackageId: ContentPackageId;
+}
+
+/** Pod-level volume declaration. Declares a named volume source that can be referenced by containers via ContainerVolumeMount. */
+#suppress "@azure-tools/typespec-autorest/union-unsupported" "The existing API requires this polymorphic volume union."
```
RichardChen820, line 767: Should it be framed as polymorphism models (there is a somewhat equivalent "union extends a common base model" but it not yet available in typespec)?

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
```diff
@@ -0,0 +1,1887 @@
...
+/** Sandbox source disk image wire model. */
+model SandboxSourceDiskImage {
+  /** id value. */
+  id?: DiskImageId;
+
+  /** Disk image name for public disk images (e.g., "nginx", "python3.12"). Mutually exclusive with Id. */
+  name?: string;
+
+  /** Indicates whether this references a public disk image. Must be true when using Name, must be false when using Id. */
+  isPublic?: boolean;
```
RichardChen820, line 1097: which are required? or what combination is allowed?

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
RichardChen820: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-response-body-is-resource-schema

ahmelsayed: The sandbox create API is a PUT just to follow the create semantics of RBAC/IAM permissions, but it doesn't take an id parameter nor does it support updating the sandbox

specification/app/data-plane/ContainerApps/.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
jorgerangel-msft: we shouldn't need this, can we please remove this file?

jorgerangel-msft (commented):
There are some CI failures related to missing examples that we should address:

```
  Rule FolderStructure failed
  Invalid folder structure: Spec folder with main.tsp must contain examples folder.
```

Since we have a generated swagger, we could should be able to follow [this link](https://azure.github.io/typespec-azure/docs/migrate-swagger/faq/x-ms-examples/#generate-the-examples) to get some initial examples added.

specification/app/data-plane/ContainerApps/.npmrc
RichardChen820: No problem

RichardChen820: Updated

RichardChen820:
> There are some CI failures related to missing examples that we should address:
> 
> ```
>   Rule FolderStructure failed
>   Invalid folder structure: Spec folder with main.tsp must contain examples folder.
> ```
> 
> Since we have a generated swagger, we could should be able to follow [this link](https://azure.github.io/typespec-azure/docs/migrate-swagger/faq/x-ms-examples/#generate-the-examples) to get some initial examples added.

Added

specification/app/data-plane/ContainerApps/models/common.tsp
```diff
@@ -0,0 +1,117 @@
...
+
+namespace ContainerAppsSandbox;
+
+/** Standard Azure Core traits for Azure Container Apps Sandbox operations without conditional, repeatable, or client request ID semantics. */
+alias DataPlaneStandardTraits = NoConditionalRequests &
+  NoRepeatableRequests &
+  NoClientRequestId;
+
+/** Azure Dev Compute error object. */
+@error
```
jorgerangel-msft, line 13: Is there a compatibility requirement for returning this Problem Details-style payload rather than the standard Azure error format?

The [Azure REST API error-handling guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#handling-errors) require:
- An `x-ms-error-code` response header containing a documented, stable error code.
- A response body with an `error` object containing required `code` and `message` fields.
- The body’s `error.code` to match the `x-ms-error-code` header.

The current `ProblemDetails`/`ErrorResult` definitions don't describe that envelope. Could we use `Azure.Core.Foundations.ErrorResponse` instead? If additional diagnostic fields are needed, we can extend `Azure.Core.Foundations.Error` and use `ErrorResponseBase<CustomError>`, following the [TypeSpec customization guidance](https://azure.github.io/typespec-azure/docs/getstarted/azure-core/step04/#error-response-customization).
\

specification/app/data-plane/ContainerApps/examples/2026-09-01-preview/AzureSandboxes_Create.json
```diff
@@ -0,0 +1,1277 @@
+{
```
jorgerangel-msft, line 1: it looks like we have duplicate examples in `containerapps/examples` and under the `preview` folder

specification/app/data-plane/ContainerApps/models/common.tsp
```diff
@@ -0,0 +1,117 @@
...
+model ErrorResult<Status extends int32> {
+  /** HTTP status code. */
+  @statusCode statusCode: Status;
+
+  /** Error response body. */
+  @body body: ProblemDetails;
+}
+
+/** Azure resource scope shared by Azure-scoped resource routes. */
+model AzureResourceScope {
```
jorgerangel-msft, line 42: Could we model these scopes as a resource hierarchy so the existing read/delete operations can use the standard Azure.Core templates without changing their URLs?

The existing `AzureResourceScope` spread can remain for operations that have not been migrated. For the standard resource operations, add `using Rest;` and route-only scope models:

```typespec
/** Subscription scope for resource routes. */
@resource("subscriptions")
model SubscriptionScope {
  /** Azure subscription identifier. */
  @key
  @visibility(Lifecycle.Read)
  subscriptionId: string;
}

/** Resource group scope for resource routes. */
@resource("resourceGroups")
@parentResource(SubscriptionScope)
model ResourceGroupScope {
  /** Azure resource group name. */
  @key
  @visibility(Lifecycle.Read)
  resourceGroupName: string;
}

/** Sandbox group scope for resource routes. */
@resource("sandboxGroups")
@parentResource(ResourceGroupScope)
model SandboxGroupScope {
  /** Sandbox group name. */
  @key
  @visibility(Lifecycle.Read)
  sandboxGroupName: string;
}
```

Then mark the resources with `@resource("<segment>")`, `@parentResource(SandboxGroupScope)`, and the appropriate `@key("<path parameter>")` on their existing identity property:

| Resource property | Collection segment | Path parameter |
| --- | --- | --- |
| `Sandbox.id` | `sandboxes` | `id` |
| `Connection.id` | `connections` | `id` |
| `ContentPackage.id` | `contentpackages` | `id` |
| `SandboxGroupCredential.name` | `credentials` | `credentialName` |
| `DiskImage.id` | `diskimages` | `id` |
| `PublicDiskImage.name` | `diskimages/public` | `name` |
| `NamedEgressPolicy.id` | `egressPolicies` | `policyId` |
| `Snapshot.id` | `snapshots` | `id` |
| `Volume.volumeName` | `volumes` | `volumeName` |
| `Secret.id` | `secrets` | `secretId` |

`Sandbox` already has its resource/key annotations, so only its parent needs adding. Keep existing property visibility. For the currently unannotated `ContentPackage.id` and `PublicDiskImage.name`, explicit `@visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)` preserves their existing read/write payload behavior while satisfying the key-visibility rule.

This describes the existing subscription/resource-group/sandbox-group URL hierarchy; it does not introduce new service endpoints or fields into request/response bodies. The templates derive path parameters from the keys, so remove the explicit operation-level `@route` from migrated operations. Any inherited interface route prefix must also be accounted for to avoid duplicating the path; for `AzureSandboxes`, move its existing full route onto the still-custom list/create/delete operations.

The emitted Swagger may inline previously shared parameter references and include unused scope definitions. Those are document-organization differences, not changes to the HTTP request. The current supported API is [`Azure.Core.ResourceOperations<Traits, ErrorResponse>`](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/interfaces/).

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/connections.tsp
```diff
@@ -0,0 +1,301 @@
...
+    DataPlaneStandardTraits,
+    ListConnectionErrors
+  >;
+
+  /** Gets a specific connection by ID. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Azure.Core StandardResourceOperations.ResourceRead cannot preserve the existing includeSandboxIds query parameter and custom success/error response shapes without changing the established wire contract."
+  @summary("Gets a specific connection by ID.")
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/sandboxGroups/{sandboxGroupName}/connections/{id}")
+  @get
+  getConnection is Foundations.Operation<
```
jorgerangel-msft, line 190: The suppression reason here is too broad: `ResourceRead` can preserve `includeSandboxIds` and the current success/error responses. `deleteConnection` can likewise preserve `force` and its existing `204` response.

With the resource hierarchy and `Connection.id` key modeled, add `using Azure.Core.Traits;` and these aliases outside the interface:

```typespec
alias ConnectionReadOperations = ResourceOperations<
  DataPlaneStandardTraits,
  GetConnectionErrors
>;
alias ConnectionDeleteOperations = ResourceOperations<
  DataPlaneStandardTraits,
  DeleteConnectionErrors
>;
```

The corresponding declarations inside the existing `ConnectionsOperations` interface can be:

```typespec
/** Gets a specific connection by ID. */
getConnection is ConnectionReadOperations.ResourceRead<
  Connection,
  QueryParametersTrait<
    {
      /** Whether to include identifiers of sandboxes using this connection. */
      @query includeSandboxIds?: boolean;
    },
    TraitContext.Read
  >
>;

/** Deletes a connection by ID. */
deleteConnection is ConnectionDeleteOperations.ResourceDelete<
  Connection,
  QueryParametersTrait<
    {
      /** Whether to force deletion. */
      @query force?: boolean;
    },
    TraitContext.Delete
  >
>;
```

Keep the existing operation summaries. The templates supply the HTTP verbs and routes; remove the old explicit routes and the two operation-specific `use-standard-operations` suppressions.

`QueryParametersTrait` retains the query names, Boolean types, and optionality. Passing `GetConnectionErrors` / `DeleteConnectionErrors` explicitly retains the existing status codes and `ProblemDetails` bodies; this does not require switching to the standard Azure error format. `DataPlaneStandardTraits` also preserves the current header/trait settings.

The old `GetConnectionResult` is an HTTP envelope, not a JSON `{ "body": ... }` wrapper. `ResourceRead<Connection>` therefore retains the existing `200` JSON body, while `ResourceDelete<Connection>` retains the bodyless `204`. The now-unused success-envelope models can be removed.

I prototyped these conversions locally and compared their emitted HTTP contracts: the paths, verbs, parameters, success/error responses, and operation IDs remain equivalent. This uses the current members of [`ResourceOperations`](https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/interfaces/), not the deprecated standalone `Azure.Core.ResourceRead` / `Azure.Core.ResourceDelete` helpers.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/contentpackages.tsp
```diff
@@ -0,0 +1,140 @@
...
+
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Custom operations preserve the source OpenAPI response status codes, bare arrays, and route shapes."
+#suppress "@azure-tools/typespec-azure-core/use-standard-names" "Operation names are derived from the established source routes to keep generated names deterministic."
+@tag("ContentPackages")
+interface ContentPackagesOperations {
+  /** Deletes a content package by ID. */
+  @summary("Deletes a content package by ID.")
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/sandboxGroups/{sandboxGroupName}/contentpackages/{id}")
+  @delete
+  deleteContentPackage is Foundations.Operation<
```
jorgerangel-msft, line 69: This delete and the corresponding read can use the current standard templates while preserving their existing contracts. With the resource hierarchy/key annotations in place, the relevant pieces are:

```typespec
alias ContentPackageReadOperations = ResourceOperations<
  DataPlaneStandardTraits,
  GetContentPackageErrors
>;
alias ContentPackageDeleteOperations = ResourceOperations<
  DataPlaneStandardTraits,
  DeleteContentPackageErrors
>;
```

Inside the existing interface, retain the operation documentation and replace the custom declarations with:

```typespec
getContentPackage is ContentPackageReadOperations.ResourceRead<ContentPackage>;
deleteContentPackage is ContentPackageDeleteOperations.ResourceDelete<ContentPackage>;
```

Remove their explicit operation-level routes and unused success-envelope models. The same pattern applies to the following 15 operations, in addition to the two connection operations:

| Resource | `ResourceRead` candidates | `ResourceDelete` candidates |
| --- | --- | --- |
| `Sandbox` | `AzureSandboxes.get` | None in this change |
| `ContentPackage` | `getContentPackage` | `deleteContentPackage` |
| `SandboxGroupCredential` | `getCredential` | `deleteCredential` |
| `DiskImage` | `getDiskImage` | `deleteDiskImage` |
| `PublicDiskImage` | `getDiskImagePublic` | None |
| `NamedEgressPolicy` | `getEgressPolicy` | `deleteEgressPolicy` |
| `Snapshot` | `getSnapshot` | `deleteSnapshot` |
| `Volume` | `getVolume` | `deleteVolume` |
| `Secret` | None | `deleteSecret` |

For each operation, instantiate `ResourceOperations<DataPlaneStandardTraits, ExistingOperationErrors>` with that operation's existing error union. The reads already return the resource as a `200` body, and these deletes already return `204` without a body, so no change to service success/error behavior is needed.

This recommendation intentionally excludes `AzureSandboxes.delete` (`200 | 204`), file deletion operations, lists, and create/update operations. Those need separate analysis rather than being forced into the same templates.

This addresses some of the [`use-standard-operations`](https://azure.github.io/typespec-azure/docs/libraries/azure-core/rules/use-standard-operations/) violations without changing the service contract.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
```diff
@@ -0,0 +1,127 @@
...
+/** Errors returned by listSnapshots. */
+alias ListSnapshotErrors = ErrorResult<401> | ErrorResult<500>;
+
+/** OK. */
+model GetSnapshotsCountResult {
+  /** HTTP response status code. */
+  @statusCode statusCode: 200;
+
+  /** Response body. */
+  @body body: int32;
```
jorgerangel-msft, line 37: It looks like the response body is a primitive and since there is no explicit content type set, this will default to text/plain. Generally, we should avoid this as it prevents the response body from being evolvable. Same thing for the other operation that retrieves a count.

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
```diff
@@ -0,0 +1,248 @@
...
+interface SandboxFilesOperations {
+  /** Deletes a file or directory in a running sandbox. If the target is a directory, the `recursive` query parameter must be set to true to delete it and its contents. */
+  @summary("Deletes a file or directory in a running sandbox. If the target is a directory, the `recursive` query\nparameter must be set to true to delete it and its contents.")
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/sandboxGroups/{sandboxGroupName}/sandboxes/{id}/files")
+  @delete
+  deleteSandboxFile is Foundations.Operation<
+    {
+      ...AzureResourceScope;
+
+      /** id parameter. */
```
jorgerangel-msft, line 109: a lot of these param's description are very vague and this will translate to the generated client. Can we please improve the parameter descriptions across the operations so it's easy for consumers to identify what they pertain to ?

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,138 @@
...
+  /** Azure-scoped Sandbox CRUD operations. */
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Azure.Core StandardResourceOperations.ResourceList cannot preserve the existing Page, PageSize, and labels query parameter names or the array response without changing the established wire contract."
+  @doc("Lists sandboxes in a scope.")
+  @tag("Sandboxes")
+  @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/sandboxGroups/{sandboxGroupName}/sandboxes")
+  interface AzureSandboxes {
+    @doc("Lists sandboxes in a scope.")
+    @listsResource(Sandbox)
+    @get
+    list is Foundations.Operation<
```
jorgerangel-msft, line 73: I am curious what the current experience is for this operation? Does this just return all of the sandboxes in a single call? Here are our guidelines regarding collection: [link](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#collections). We should be allowing to retrieve the next page using some sort of token or nextlink

specification/app/data-plane/ContainerApps/models/common.tsp
RichardChen820: @jorgerangel-msft  If we do so , will the auto-generator generate clients for the parent resources? `SubscriptionClient`, `ResourceGroupClient` ...

JonathanGiles:
From a naming perspective, `azure-containerapps-sandbox` / `com.azure.containerapps.sandbox` probably isn't the right name for Java. Our naming convention is to use a group name after azure - for Java we document this here: https://azure.github.io/azure-sdk/java_introduction.html#java-namespaces-approved-list

specification/app/data-plane/ContainerApps/models/common.tsp
RichardChen820: The action's implemented wire contract is indeed  Problem Details object, Switching directly to Azure.Core.Foundations.ErrorResponse would change the response body, would constitute a real breaking change to the wire contract.

RichardChen820: @jorgerangel-msft  The action's implemented wire contract is Problem Details object, Switching directly to Azure.Core.Foundations.ErrorResponse would constitute a real breaking change to the wire contract.

RichardChen820: @ahmelsayed

specification/app/data-plane/ContainerApps/routes/contentpackages.tsp
RichardChen820: Updated to use the standard operation

specification/app/data-plane/ContainerApps/routes/connections.tsp
RichardChen820: Updated to use the standard operation

specification/app/data-plane/ContainerApps/models/common.tsp
RichardChen820: Updated to use the standard operation

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
RichardChen820: This is the existing wire contract @jorgerangel-msft

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
RichardChen820: It's an existing wire contract, it returns set of sandboxes by page (using page/pageSize). I proposed this to be in line with the guideline: [Listing resource pagination uses continuation token support by RichardChen820 · Pull Request #46325 …](https://github.com/Azure/azure-rest-api-specs/pull/46325)

specification/app/data-plane/ContainerApps/tspconfig.yaml
```diff
@@ -0,0 +1,24 @@
...
+emit:
+  - "@azure-tools/typespec-autorest"
+options:
+  "@azure-tools/typespec-autorest":
+    emit-lro-options: "none"
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/containerappssandbox.json"
+  "@azure-typespec/http-client-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+    namespace: Azure.ContainerApps.Sandbox
```
jorgerangel-msft, line 15: @JoshLove-msft @m-nash for the package name, what are your thoughts on `Azure.Containers.AppContainersSandbox` or `Azure.Containers.AppContainers.Sandbox` to group this with the existing `Containers` group and related services?

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
RichardChen820: I'm going to update the descriptions in one shot after we settle everything else.

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
```diff
@@ -0,0 +1,1887 @@
...
+  kind: TdsAuthKind;
+
+  /** username value. */
+  username?: string;
+
+  /** secretRef value. */
+  secretRef?: EgressPolicySecretRef;
+
+  /** secret value. */
+  secret?: string;
```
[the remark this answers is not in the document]
jorgerangel-msft: Can we please address all of the flagged security concerns?

jorgerangel-msft (commented):
In general, we have a lot of suppressions that we should try and remove by simply moving to the approved templates for operations https://azure.github.io/typespec-azure/docs/libraries/azure-core/reference/interfaces/#Azure.Core.ResourceOperations

specification/app/data-plane/ContainerApps/models/volumes.tsp
```diff
@@ -0,0 +1,285 @@
...
+
+namespace ContainerAppsSandbox {
+  /** ARM resource ID for an Azure Storage blob container. */
+  scalar ArmStorageContainerId extends string;
+
+  /** ARM resource ID for an Azure user-assigned managed identity. */
+  scalar ArmUserAssignedIdentityId extends string;
+
+  /** Authentication for a bring-your-own (BYO) Azure Blob container volume: how ADC obtains credentials to access the customer-owned container. */
+  union AzureBlobByoVolumeAuth {
```
jorgerangel-msft, line 11: Question:

For user-provided blob volumes, is managed identity the only supported authentication mechanism, or are others planned, such as SAS or account keys? The union currently has only one variant, so this is resulting  in a `BinaryData` type for the Auth property in the generated .NET client. From a .NET API perspective, getting clarity on this will help decide if the library needs a single typed configuration or an extensible authentication hierarchy.

ahmelsayed: Currently (in this API version) it's only managed identity. We might add more in the future to future API versions.

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
ahmelsayed: Can we do this in a subsequent API version release?

specification/app/data-plane/ContainerApps/models/common.tsp
ahmelsayed: Can we do this in a subsequent API version release?

specification/app/data-plane/ContainerApps/tspconfig.yaml
ahmelsayed: I'm not sure how namespaces are managed across teams. the only `Azure.Containers` I see is `Azure.Containers.ContainerRegistry` is that right?

`ContainerApps` is our uber service name/branding in Azure, so I think `Azure.ContainerApps.*` makes more sense than `Azure.Containers.App*`. Under the `ContainerApps` umbrella, resources here are in a `SandboxGroup` Azure resource. So I think the full namespace could be `Azure.ContainerApps.SandboxGroups.{resourceName}

I know the dotnet sdk pr pushed back on reasonable names identified by their namespace, so I'm ok prefixing our object names with `SandboxGroup*`

specification/app/data-plane/ContainerApps/examples/2026-09-01-preview/AzureSandboxes_Create.json
ahmelsayed: Which path is the right place for these examples. I see in this repo some under 

- `specification/{name}/data-plane/{name}/examples/{api-version}` [example](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/search/data-plane/Search/examples/2025-11-01-preview)
- `specification/{name}/data-plane/{name}/preview/{api-version}/examples` [example](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/imds/data-plane/InstanceMetadataService/stable/2018-02-01/examples)

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
ahmelsayed: Updated to an application/json with `{ count: number }` type.

ahmelsayed: actually sorry. I don't seem to have permissions to push to this branch. I requested access

JoshLove-msft (commented):
I found three additional response-contract issues beyond the paging, error-envelope, count-response, and naming discussions: file downloads describe a JSON object instead of the documented binary body; storage-size fields use a binary-data scalar; and resource-usage byte counters cannot represent 2 GiB. Details are inline. These should be reconciled with the actual service payloads before publication, rather than assuming a later API version makes the corrections non-breaking.

For the sandbox-specific package, my revised naming preference is:

| Target | Proposed name |
| --- | --- |
| .NET package / root namespace | `Azure.Containers.Apps.Sandbox` |
| Java Maven coordinates | `com.azure:azure-containers-apps-sandbox` |
| Java namespace | `com.azure.containers.apps.sandbox` |

This follows the [group/service/optional-feature structure for .NET](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-namespace-naming) and [Java](https://azure.github.io/azure-sdk/java_introduction.html#namespaces). `Containers` supplies the category, `Apps` identifies Container Apps, and `Sandbox` identifies the feature, avoiding the repeated "Container" in my earlier recommendation. `SandboxGroups` describes a resource scope and fits better in the client/resource hierarchy than in the package name. This is a proposal for language-owner sign-off, not namespace approval.

This was a follow-up contract review of the TypeSpec and emitted Swagger, not verification of the deployed service. I did not repeat already-covered findings.

--generated by Copilot

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
```diff
@@ -0,0 +1,249 @@
...
+  | ErrorResult<404>
+  | ErrorResult<409>;
+
+/** OK. */
+model GetSandboxFileResult {
+  /** HTTP response status code. */
+  @statusCode statusCode: 200;
+
+  /** Response body. */
+  @body body: FileStreamResult;
```
JoshLove-msft: **Warning: Model the downloaded file as the response body, not a JSON result object.**

`getSandboxFile` explicitly says that the file is streamed directly to the response body, but `@body body: FileStreamResult` emits an `application/json` response with properties such as `fileStream`, `contentType`, and `fileDownloadName`. That describes JSON deserialization rather than reading the file's raw contents; a `contentType` property inside that JSON object does not describe the HTTP Content-Type header. `getVolumeFileDownload` uses the same result model.

Please model the successful response with a `bytes` body and the actual supported response media type(s), and model any returned download metadata as HTTP headers rather than serializing the server-side result object's properties. Keep the existing error responses separate. The [TypeSpec HTTP content-type guidance](https://typespec.io/docs/libraries/http/content-types/) distinguishes model/JSON bodies from binary `bytes` bodies. Confirm both download endpoints against the service's actual response before publishing this contract.

--generated by Copilot

specification/app/data-plane/ContainerApps/models/common.tsp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/** AbsoluteDirectoryPath wire model. */
+scalar AbsoluteDirectoryPath extends string;
+
+/** AbsoluteFilePath wire model. */
+scalar AbsoluteFilePath extends string;
+
+/** Bytes wire model. */
+scalar Bytes extends bytes;
```
JoshLove-msft: **Warning: Byte counts are being described as binary data.**

In TypeSpec, [`bytes` is a sequence of bytes](https://typespec.io/docs/language-basics/built-in-types/), not a numeric size. This scalar emits `type: string, format: byte`. It is then used for `ContentPackage.size`, `BlobVolumeUsage.usedBytes`, `DataDiskVolumeUsage.compressedBlobSizeBytes` / `usedSizeBytes`, and `VolumePathItem.sizeBytes`, so those size/usage fields become base64 data instead of numeric byte counts. The generated content-package example even has `"size": "SGVsbG8="`.

Please give these quantities a numeric scalar, such as `int64` with the intended JSON-number range, matching the actual service payload, and regenerate the affected schemas/examples. Keep `bytes` for actual file/package contents. This is separate from wrapping the count API response: it changes the types of size properties throughout the resource models and should not be left as a supposedly non-breaking cleanup after release.

--generated by Copilot

specification/app/data-plane/ContainerApps/models/sandbox-lifecycle.tsp
```diff
@@ -0,0 +1,196 @@
...
+    availableBytes?: int32;
+
+    /** Human-friendly display label (e.g., "container-0"). If set, UI should prefer this over MountPoint. */
+    label?: string;
+  }
+
+  /** Memory statistics in bytes. */
+  model MemoryStats {
+    /** totalBytes value. */
+    totalBytes?: int32;
```
JoshLove-msft: **Warning: Resource-usage counters cannot represent 2 GiB.**

These fields are explicitly measured in bytes, but [`int32` stops at 2,147,483,647](https://typespec.io/docs/language-basics/built-in-types/). A 2 GiB memory or filesystem capacity is already 2,147,483,648 bytes, outside the contract. `DiskStatsEntry.totalBytes` / `usedBytes` / `availableBytes` and the `NetworkStats` byte counters have the same problem; cumulative network traffic can exceed that threshold even for a small sandbox. The sandbox resource quantities have no corresponding sub-2-GiB limit.

Please use a sufficiently wide integer type, such as `int64`, consistently for these byte totals/counters and add a representative stats example above the 32-bit boundary. The current examples use `1`, which hides this range problem. Align the spec with the service's actual counter types before publishing; widening the documented response range later can break clients generated with 32-bit fields.

--generated by Copilot

specification/app/data-plane/ContainerApps/examples/2026-09-01-preview/AzureSandboxes_Create.json
jorgerangel-msft: I believe this needs to go under `preview` as https://azure.github.io/typespec-azure/docs/migrate-swagger/faq/x-ms-examples/#example-structure. CI should also gate if the incorrect directory is used

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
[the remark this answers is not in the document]
jorgerangel-msft: @ahmelsayed

specification/app/data-plane/ContainerApps/tspconfig.yaml
jorgerangel-msft: Please see the proposed name in https://github.com/Azure/azure-rest-api-specs/pull/46133#pullrequestreview-5215679849. This follows our existing guidance https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-namespace-naming and Azure.ContainerApps would be a new group name which would require an additional review process as outlined in the linked doc

kashifkhan:
For Python, keeping in with line with namespaces here `azure-containers-appsandbox`  can work https://azure.github.io/azure-sdk/python_design.html

The current package name doesnt fit in our established patterns and subnames

specification/app/data-plane/ContainerApps/suppressions.yaml
jorgerangel-msft: We should get alignment on the package names for the other languages and remove this file. @kashifkhan  has provided his guidance for python.

jorgerangel-msft: @kashifkhan  - do you know the answer to the pypi error regarding name registration ?

specification/app/data-plane/ContainerApps/models/sandbox-lifecycle.tsp
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/models/common.tsp
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/examples/2026-09-01-preview/AzureSandboxes_Create.json
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/tspconfig.yaml
ahmelsayed: Used `Azure.Containers.Apps.Sandbox` and `com.azure.containers.apps.sandbox`

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,118 @@
...
+  @tag("Sandboxes")
+  interface AzureSandboxes {
+    @doc("Lists sandboxes in a scope.")
+    list is SandboxListOperations.ResourceList<Sandbox, SandboxListTraits>;
+
+    #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "The established wire contract uses POST with CreateSandboxContent and returns the created Sandbox in a 201 response body."
+    @doc("Creates a sandbox in a scope.")
+    @route("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/sandboxGroups/{sandboxGroupName}/sandboxes")
+    @post
+    create is Foundations.Operation<
```
jorgerangel-msft: question: is this a long running operation?

specification/app/data-plane/ContainerApps/client.tsp
```diff
@@ -0,0 +1,770 @@
...
+
+  // Container image references can include registry names and tags rather than absolute URIs.
+  @@clientName(
+    CreateDiskImageSourceRegistrySource.imageUrl,
+    "imageReference",
+    "csharp"
+  );
+}
+
+@clientNamespace("Azure.Containers.ContainerApps.Sandbox", "csharp")
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/cspell.yaml
```diff
@@ -15,6 +15,10 @@ words:
   - occured
   - otlp
   - sreagent
+  - rootfs
+  - esan
+  - cifs
+  - upserts
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
ahmelsayed: No, it's intentionally synchronous  with a P99 target of ~1 second.

specification/app/data-plane/ContainerApps/preview/2026-09-01-preview/examples/VolumesOperations_PostVolumeFile.json
```diff
@@ -0,0 +1,75 @@
...
+  "responses": {
+    "201": {
+      "body": {
+        "itemName": "example-item",
+        "path": "/workspace/example.txt",
+        "isDirectory": true,
+        "sizeBytes": 1024,
+        "lastModifiedUtc": "2026-08-15T10:00:00Z",
+        "contentType": "example",
+        "eTag": "example-e-tag"
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/preview/2026-09-01-preview/examples/VolumesOperations_ListVolumes.json
```diff
@@ -0,0 +1,55 @@
...
+          {
+            "type": "Volume",
+            "volumeName": "wqf",
+            "labels": {
+              "key3961": "nhgorurqvrqct"
+            },
+            "provisioningState": "Provisioning"
+          }
+        ],
+        "nextLink": "https://microsoft.com/a"
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/preview/2026-09-01-preview/examples/ConnectionsOperations_PostConnectionAuthorize.json
```diff
@@ -0,0 +1,97 @@
...
+  "operationId": "ConnectionsOperations_PostConnectionAuthorize",
+  "parameters": {
+    "api-version": "2026-09-01-preview",
+    "subscriptionId": "B82004AA-DC11-4C6B-95FD-0A3A506CF6CA",
+    "resourceGroupName": "rgcontainerappssandbox",
+    "sandboxGroupName": "dqgidniquqbnwbcelbkijnrwo",
+    "id": "citmhfwyopfsxs",
+    "body": {
+      "parameterValues": {
+        "key8237": "vjinbboshdhnsfs"
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/preview/2026-09-01-preview/examples/AzureSandboxes_Get.json
```diff
@@ -0,0 +1,684 @@
...
+              {
+                "name": "xcrcnkyelvnfilofhrpwsjd",
+                "kind": "SqlPassword",
+                "username": "yjxbaifncuokfndfyyffqglrxfcscg",
+                "secretRef": {
+                  "secretId": "qeu",
+                  "secretKey": "kaloxmpmmz",
+                  "format": "xauryjy"
+                },
+                "secret": "szgtosxwqrbm"
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/models/sandboxes.tsp
```diff
@@ -0,0 +1,1873 @@
...
+  kind: TdsAuthKind;
+
+  /** The SQL username. */
+  username?: string;
+
+  /** A reference to the stored password or token. */
+  secretRef?: EgressPolicySecretRef;
+
+  /** Resolved credential value, when included by the service. */
+  secret?: string;
```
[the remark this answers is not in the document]
ahmelsayed: fixed

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
```diff
@@ -0,0 +1,96 @@
...
+  | NotFoundErrorResult
+  | InternalServerErrorResult;
+
+/** Shared operation traits for reading snapshots. */
+alias SnapshotGetOperations = ResourceOperations<
+  DataPlaneStandardTraits,
+  GetSnapshotErrors
+>;
+
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "The controller exposes a /count route, and DELETE may return either 202 or 204."
```
jorgerangel-msft: The updated suppression says snapshot deletion can return `202` or `204`, but `ResourceDelete` currently emits only `204`. Which behavior does the service implement?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
```diff
@@ -0,0 +1,248 @@
...
+      @query containerName?: string;
+
+      /** The directory path and creation options. */
+      @body body: MkDirContent;
+    },
+    PostSandboxFileMkdirResult
+  >;
+
+  /** Writes a file to a running sandbox. The file is streamed directly from the request body. */
+  #suppress "@azure-tools/typespec-azure-core/byos" "PUT /sandboxes/{id}/files reads raw file bytes directly from Request.Body."
```
jorgerangel-msft: The suppression describes `PUT /sandboxes/{id}/files`, but this `ResourceAction` emits `POST` for that route. Can we confirm the controller's supported verb and align the declaration or explanation? Generated clients will send `POST`, so this matters beyond the suppression wording.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
```diff
@@ -0,0 +1,120 @@
...
+using Azure.Core.Traits;
+
+namespace ContainerAppsSandbox {
+  /** Successful Sandbox create response. */
+  model SandboxCreatedResult {
+    /** The HTTP 201 status code. */
+    @statusCode statusCode: 201;
+
+    /** Created sandbox. */
+    @body body: Sandbox;
```
jorgerangel-msft, line 16: Both sandbox and disk-image creation return `201` without a modeled `Location` header. Can you please clarify how retrying after a lost response avoids duplicate resources, as required by the [POST creation guidance](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-post-must-be-idempotent) ?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/volumes.tsp
```diff
@@ -0,0 +1,358 @@
...
+alias VolumeForkOperations = ResourceOperations<
+  DataPlaneStandardTraits,
+  PostVolumeForkErrors
+>;
+
+/** The HTTP 201 response for the putVolume operation. */
+model PutVolumeResult {
+  /** The HTTP 201 status code. */
+  @statusCode statusCode: 201;
+
```
jorgerangel-msft, line 163: The operation still describes create-only behavior and exposes only `201`. What happens when a caller repeats the same PUT after losing the response, or sends a different configuration for an existing volume?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/models/connections.tsp
```diff
@@ -0,0 +1,105 @@
...
+    policyRules?: McpPolicyRule[];
+
+    /** Enabled tool groups. */
+    enabledToolGroups?: string[];
+  }
+
+  /** Options for generating a connection consent link. */
+  model GenerateConsentLinkContent {
+    /** The URI to redirect to after consent completes. */
+    redirectUrl?: string;
```
jorgerangel-msft: `redirectUrl` and `consentLink` are documented URIs but remain `string`. Please use the TypeSpec `url` scalar for these and the other confirmed URI-valued fields: sandbox app/management URLs, token-exchange endpoint, webhook/proxy URLs, and gateway runtime URLs.

This preserves the JSON string representation while enabling correct SDK types, such as .NET `Uri`. It is worth addressing before SDK publication because changing a public property from `string` to `Uri` later is breaking. Container image references should remain strings.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/models/connections.tsp
```diff
@@ -0,0 +1,105 @@
+using Rest;
+
+namespace ContainerAppsSandbox {
+  /** Parameters used to authorize a connection. */
+  model AuthorizeConnectionContent {
+    /** Provider-specific authorization parameters. */
+    parameterValues: Record<string>;
```
jorgerangel-msft, line 8: `AuthorizeConnectionContent.parameterValues` can carry credential material but still lacks `@secret`. Please mark the property as secret, consistent with the other sensitive credential inputs. This supplies the contract metadata for tooling; runtime redaction still needs to be handled separately.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/models/secrets.tsp
```diff
@@ -0,0 +1,43 @@
...
+  /** Keys stored in a secret. */
+  model SecretKeysResponse {
+    /** The key names stored in the secret. */
+    keys: string[];
+  }
+
+  /** Selected values from a secret. */
+  model SecretPeekResponse {
+    /** The sensitive key-value pairs stored in the secret. */
+    values: Record<string>;
```
jorgerangel-msft, line 17: `SecretPeekResponse.values` is explicitly documented as containing sensitive values but still lacks `@secret`, unlike `CreateSecretContent.values`. Please mark it as secret as well.

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/tspconfig.yaml
```diff
@@ -0,0 +1,24 @@
...
+  "@azure-tools/typespec-autorest":
+    emit-lro-options: "none"
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/containerappssandbox.json"
+  "@azure-typespec/http-client-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+    namespace: Azure.Containers.Apps.Sandbox
+    model-namespace: false
+  "@azure-tools/typespec-java":
+    emitter-output-dir: "{output-dir}/{service-dir}/azure-containerapps-sandbox"
```
jorgerangel-msft: Looks like java generation is failing. We can try changing the service-dir  value to `appcontainers` for java to match their existing structure at https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/appcontainers

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
ahmelsayed: The warning was supressing `GET ../Microsoft.App/snapshots/count`, changed it to a POST and removed the suppression.

specification/app/data-plane/ContainerApps/routes/sandbox-files.tsp
ahmelsayed: It should be a post. Updates the doc line.

specification/app/data-plane/ContainerApps/routes/volumes.tsp
ahmelsayed: Added a location header

specification/app/data-plane/ContainerApps/models/connections.tsp
ahmelsayed: Updated to `url` type

specification/app/data-plane/ContainerApps/models/connections.tsp
ahmelsayed: added `@secret`

specification/app/data-plane/ContainerApps/tspconfig.yaml
ahmelsayed: Updated

specification/app/data-plane/ContainerApps/models/secrets.tsp
ahmelsayed: Added `@secret`

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
ahmelsayed: Each POST will create a new sandbox/diskImage we don't have Repeatability-Request-ID & Repeatability-First-Sent headers. They sound like a good to have thing, but we don't have that today.

Added a location header.

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
jorgerangel-msft: @ahmelsayed - 
so getSnapshotsCount is actually a POST ? If so, what is the expected payload ? I think GET was fine, my question was tailored to the deleteSnapshot operation since the template only returned `204` and I just wanted to double check that matched service contract.

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
jorgerangel-msft: I think we should aim for supporting that. Not necessarily as part of this API version but maybe prior to GA ? @JoshLove-msft - do you mind sharing your thoughts ?

specification/app/data-plane/ContainerApps/models/connections.tsp
jorgerangel-msft: thanks. There's a couple more. can we please just update them all ? It's not a blocker but ideally we should align this in the service contract so emitters don't have to add client.tsp customizations down the line:

| TypeSpec property | Location | Suggested type |
|---|---|---|
| `Sandbox.appUrl` | [models/sandboxes.tsp:73](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L73) | `appUrl?: url` |
| `Sandbox.managementUrl` | [models/sandboxes.tsp:76](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L76) | `managementUrl?: url` |
| `EgressPolicyHookRef.endpoint` | [models/sandboxes.tsp:381](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L381) | `endpoint: url` |
| `EgressForwardProxy.url` | [models/sandboxes.tsp:504](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L504) | `url: url` |
| `GatewayConnection.mcpRuntimeUrl` | [models/sandboxes.tsp:614](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L614) | `mcpRuntimeUrl?: url` |
| `GatewayConnection.connectionRuntimeUrl` | [models/sandboxes.tsp:618](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L618) | `connectionRuntimeUrl?: url` |
| `SandboxGroupCredentialConnectionRefDetails.tokenExchangeEndpoint` | [models/credentials.tsp:45](https://github.com/Azure/azure-rest-api-specs/blob/11e263c88ad746e0b93d31edb9a30eb1034dd3f9/specification/app/data-plane/ContainerApps/models/credentials.tsp#L45) | `tokenExchangeEndpoint: url` |

specification/app/data-plane/ContainerApps/routes/volumes.tsp
jorgerangel-msft: Thanks for adding the `Location` header. Since this operation is now `POST /volumes`, can you clarify what happens when a caller retries after losing the response, or submits a request with a volume name that already exists? Does the behavior differ between identical and different configurations?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
jorgerangel-msft: In addition, The `Location` header was added to sandbox creation, but `PostDiskImageResult` still only defines the `201` status and response body. Can we also include the created disk image’s URL in a `Location` response header?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
ahmelsayed: the count operation is a GET today, but I was changing it to a POST because it's the one that was causing the use-standard-operations suppression. If I keep it as a GET I'll need to keep the suppression right?

The current API returns 204 if the resource was already deleted (or didn't exist) and 202 if it existed.

specification/app/data-plane/ContainerApps/routes/volumes.tsp
ahmelsayed: It really should be a `PUT ../volumes/{volumeName}`, but it's not an idempotent PUT today. It'll fail with a 409 if the volume exists. @RichardChen820 changed it to a POST and it now takes the name in the body. But if the name exists, it'll fail with 409. So if the caller loses the response, they'll get 409 because the volume would exist.

specification/app/data-plane/ContainerApps/tspconfig.yaml
ahmelsayed: maybe I misunderstood this. I did 

```diff
-    emitter-output-dir: "{output-dir}/{service-dir}/azure-containerapps-sandbox"
+    emitter-output-dir: "{output-dir}/appcontainers/azure-containerapps-sandbox"
```

but tsv fails with 


> - The path part "appcontainers/azure-containerapps-sandbox" in options.@azure-tools/typespec-java.emitter-output-dir does not match the required format "/^azure(-\w+)+$/". Please update the emitter-output-dir path to follow the SDK naming convention.

specification/app/data-plane/ContainerApps/models/connections.tsp
ahmelsayed: updated

specification/app/data-plane/ContainerApps/routes/sandboxes.tsp
ahmelsayed: updated

specification/app/data-plane/ContainerApps/tspconfig.yaml
```diff
@@ -0,0 +1,24 @@
...
+    emit-lro-options: "none"
+    emitter-output-dir: "{project-root}"
+    output-file: "{version-status}/{version}/containerappssandbox.json"
+  "@azure-typespec/http-client-csharp":
+    emitter-output-dir: "{output-dir}/{service-dir}/{namespace}"
+    namespace: Azure.Containers.Apps.Sandbox
+    model-namespace: false
+  "@azure-tools/typespec-java":
+    emitter-output-dir: "{output-dir}/appcontainers/azure-containerapps-sandbox"
+    namespace: com.azure.containers.apps.sandbox
```
weidongxu-microsoft: make sure the 2 matches

RichardChen820: updated

specification/app/data-plane/ContainerApps/tspconfig.yaml
RichardChen820: Fixed

specification/app/data-plane/ContainerApps/routes/volumes.tsp
RichardChen820: Returning 409 Conflict for a PUT request is a bit odd.

Typically, PUT semantics are:
- Resource does not exist: create it and return 201 Created
- Resource already exists: replace/update it and return 200 OK

Even if the same request is sent repeatedly, PUT naturally guarantees that the final state remains the same, which is exactly its idempotent behavior.

If the intention is to support "create only if the resource does not already exist", the standard approach is to use a conditional request with If-None-Match: *. In that case, if the resource already exists, the server should return 412, not 409 as well.

For this reason, so I think a create-only API is better modeled as a POST operation. Of course, using POST introduces idempotency concerns, so I think following the API Guidelines' recommendations for making POST requests idempotent is required

RichardChen820: One key difference between this API and the Sandbox/diskImage creation APIs is that users must specify the volume name. In that scenario, PUT is arguably a better fit. However, the current implementation is not fully idempotent because replaying the same request returns 409 Conflict.

There are two possible approaches:
- Keep the current behavior. If a client receives a 409 Conflict, it can issue a GET request to retrieve the existing resource.
- Perform a strict comparison. If the incoming request is identical to the existing resource on the server, return 200 OK; otherwise, return 409 Conflict.

RichardChen820: @jorgerangel-msft There're only createOrReplace and createOrUpdate standard operation for PUT, if we keep PUT, we also have to use the suppression, do we?

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
RichardChen820: Shift getSnapshotsCount  to be `POST` does not make much sense. We literally not gonna be able to avoid using suppressions anyway, the `/count` api is the one we have to tolerate as well.

RichardChen820: 202 is typically used for asynchronous operations. Our delete operations are all synchronous, and we don't have any asynchronous LRO processing implemented on the backend. In that case, returning 202 Accepted would not be appropriate. 
For both of the following scenarios, we should return 204 No Content:
- The resource existed and has just been deleted successfully.
- The resource had already been deleted or did not exist.

If we have plan to convert these deletion operations to LROs in the future, that would be a breaking change anyway. It doesn't justify returning 202 Accepted today.

jorgerangel-msft: I agree getSnapshotsCount should stay as GET since it's not updating any resource. The suppressions are there to help guide the APIs to our guidelines, but since this is an existing API, we need to understand the implications and breaking change policy if we don't follow them and change these after service GA. For getSnapshotsCount, we can keep the previous suppression, just please add a suppression message and please make sure the template we use accurately represents the status codes that can be returned. Otherwise, emitters may generate incorrect code.

Same goes for the delete operation(s). If the service returns 202 + 204, I don't think we can use the `ResourceDelete` template as that only returns 204. In that case, we would need to either adjust the backend or revert to the previous template with a suppression + justification. @RichardChen820's explanation makes sense to me.

specification/app/data-plane/ContainerApps/routes/volumes.tsp
jorgerangel-msft: Yes, we would have to add the suppression. I'd prefer if the spec accurately reflects the service vs favoring the approved templates and incorrectly reflecting the service behavior. For this operation, ideally we make this operation idempotent before GA, make it PUT, and use approved template. If that's not possible, then we can revert to the previous template with the suppression as long as it accurately describes the operation behavior and we understand the impact.

jorgerangel-msft: Also, `volumeName` is still required in the request body as well as the path. Does the service require both, and what happens if they differ? If the path supplies the name, can we make the body property read-only again?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/models/connections.tsp
jorgerangel-msft: Thanks for updating the URL types. One remaining field is [`EgressPolicyHookRef.endpoint`](https://github.com/Azure/azure-rest-api-specs/blob/c592d3da960a1c4273d19fa7cc83e86de77ba03e/specification/app/data-plane/ContainerApps/models/sandboxes.tsp#L381), which still uses `string`. Can we change this to `url` as well?

> _🤖 Generated by Jorge's Copilot_

specification/app/data-plane/ContainerApps/models/sandbox-files.tsp
```diff
@@ -0,0 +1,73 @@
+namespace ContainerAppsSandbox {
+  /** Files and directories returned for a directory listing. */
+  model DirListingResponse {
+    /** The directory path that was listed. */
+    path: string;
+
+    /** Files and directories contained in the path. */
+    entries: FileInfo[];
```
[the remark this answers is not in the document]
jorgerangel-msft: Is there a reason why we omitted pagination from this operation ?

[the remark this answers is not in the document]
ahmelsayed: It was never considered for this API to start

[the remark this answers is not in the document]
jorgerangel-msft: we should consider adding pagination support if we think any of the remaining collection operations will return large responses. Doing that in the future will be a breaking change.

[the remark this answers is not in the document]
RichardChen820: It's not an independent resource,  we also don't plan to do it in the future.

specification/app/data-plane/ContainerApps/routes/snapshots.tsp
RichardChen820: @ahmelsayed  We change the backend to only return 204, keep the spec to using the ResourceDelete template, are you good with it?

specification/app/data-plane/ContainerApps/routes/volumes.tsp
RichardChen820: Updated, remove the requirement from requestbody, it's read-only now.

specification/app/data-plane/ContainerApps/models/sandbox-files.tsp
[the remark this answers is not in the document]
RichardChen820: It's not an independent reource, we don't have plan to add pagination for it.

jorgerangel-msft (approved):

kashifkhan:
For Python, keeping in with line with namespaces here azure-containers-appsandbox can work https://azure.github.io/azure-sdk/python_design.html

The current package name doesnt fit in our established patterns and subnames

ahmelsayed:
> For Python, keeping in with line with namespaces here azure-containers-appsandbox can work [azure.github.io/azure-sdk/python_design.html](https://azure.github.io/azure-sdk/python_design.html)
> 
> The current package name doesnt fit in our established patterns and subnames

We haven't defined a python package name yet, right?

jorgerangel-msft (approved):

vturecek (approved):

PR #46143, opened 2026-09-05
Integrate azsdk-cli breaking change detection

raych1:
## Summary
- integrate `azsdk pkg detect-breaking-change` into the spec PR SDK generation flow
- use definitive azsdk-cli results to update the breaking-change label while preserving the existing result on errors or unknown responses
- keep the integration disabled by default behind a code-level feature flag
- add unit coverage for response handling, command construction, and the default feature-flag state

## Testing
- unit tests added for the new adapter and command-helper behavior

### DIFF
--- a/eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
+++ b/eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
@@ -506,6 +506,16 @@ function isAzsdkCliAvailable(): boolean {
 }
 
 /**
+ * Whether azsdk-cli SDK breaking-change detection is enabled for this run.
+ *
+ * Code-level feature flag: flip to `true` and merge to enable the whole
+ * integration.
+ */
+export function isBreakingChangeDetectionEnabled(): boolean {
+  return false;
+}
+
+/**
  * Prepare the azsdk pkg generate command arguments.
  *
  * @param commandInput - The spec-gen-sdk command input.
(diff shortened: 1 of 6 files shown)


eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
```diff
@@ -554,6 +564,25 @@ export function prepareAzsdkPackCommand(packagePath: string, outputPath?: string
...
+ * @param packagePath - Absolute path to the generated SDK package directory.
+ * @param tspConfigFullPath - Optional absolute path to the tspconfig.yaml file.
+ * @returns Array of arguments for the azsdk detect-breaking-change command.
+ */
+export function prepareAzsdkDetectBreakingChangeCommand(
+  packagePath: string,
+  tspConfigFullPath?: string,
+): string[] {
+  const args = ["pkg", "detect-breaking-change", "--package-path", packagePath];
+  if (tspConfigFullPath) {
```
raych1, line 579: @chunyu3 can you confirm whether the `tspconfig` path needs to be set?

chunyu3: When `tspConfigPath` is provided, `azsdk-detect-breaking-change` can analyze and classify SDK breaking changes more accurately because it can reference the underlying TypeSpec source. Therefore, providing the tspConfigPath is preferred whenever available.

In the azure-rest-api-specs repository, SDK code is generated from TypeSpec definitions, and the corresponding tspConfigPath can be determined. Given that, I think it would make sense to make tspConfigPath a **required** parameter in spec-gen-sdk-runner.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
```diff
@@ -554,6 +564,25 @@ export function prepareAzsdkPackCommand(packagePath: string, outputPath?: string
...
+    args.push("--tsp-config-path", tspConfigFullPath);
+  }
+  args.push("--output", "json");
```
chunyu3, line 577: 'detect-breaking-change' does not has parameter '--output'

chunyu3 (commented):
When we call azsdk cli, we may need to setup environment first.

raych1:
> When we call azsdk cli, we may need to setup environment first.

This is already done in the pipeline.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
raych1: Here in the SDK Validation checks, it would only generate the `SDK-changes.json`. That said, it is not necessary to set `tspConfigPath`.

eng/tools/spec-gen-sdk-runner/src/command-helpers.ts
raych1: It's supported by the base command.

chidozieononiwu (approved):

PR #46144, opened 2026-09-05
Adding API Version 2026-11-01 for Site Recovery

sisunkar:
# Adding SiteRecovery api-version 2026-11-01

Adds a new stable api-version **2026-11-01** to `recoveryservicessiterecovery`, covering two independent A2A feature areas: **Confidential VM (CVM) disk encryption** and **IPv6 dual-stack networking**. The change is **purely additive** (new version folder + `@added(Versions.v2026_11_01)`-guarded models); no existing version is modified.

The two Site Recovery feature areas share one API version. This PR also includes matching versioned contracts for vault management and Backup, with no new functionality in those two areas.

### A2A Confidential VM (CVM) disk encryption
- **models.tsp**: `ConfidentialDiskEncryptionInfo`, `UpdateConfidentialDiskEncryptionInfo` models and the `confidentialDiskEncryptionInfo` / `recoveryConfidentialDataDiskEncryptionIdentity` properties

### A2A IPv6 dual-stack networking
- **`ipVersion` on NIC ip-configs**: new `IPVersion` extensible enum (`IPv4` / `IPv6`, `modelAsString: true`) surfaced as `ipVersion` on `IPConfigDetails`. Declares the address family of a recovery NIC ip-config, orthogonal to the existing static/dynamic allocation type. A missing value is treated as `IPv4`, so existing clients are unaffected. The leaf is **response-only**: an ip-config's address family is derived from the source NIC during discovery, and the service update path resolves it from the stored configuration rather than from caller input, so it is deliberately not exposed on `IPConfigInputDetails`.
- **Vault-level recovery-network auto-sync**: new `RecoveryNetworkConfigAutoSyncPolicy` extensible enum (`Disabled` / `Enabled`) surfaced as `recoveryNetworkConfigAutoSync` on `VaultSettingProperties` (response) and `VaultSettingCreationInputProperties` (input). Controls whether the A2A recovery network configuration is automatically kept in sync as the source VM's network configuration changes. On the creation input an omitted value leaves the stored policy unchanged.
- **Examples**: `ReplicationVaultSetting_Create` / `_Get` / `_List` updated to show the new auto-sync leaf. New `ReplicationProtectedItems_Update_A2ADualStack` shows an A2A NIC carrying both an IPv4 and an IPv6 ip-config, on the update input and on the response, so the `ipVersion` wire shape is demonstrated end to end.

Both enums use `modelAsString: true` in line with the Azure ARM guidelines for new enums, so that a future service-side value does not break already-generated SDKs. This governs SDK/response forward-compatibility only, and is not a statement about request validation: the service validates supplied values and rejects an unrecognized `recoveryNetworkConfigAutoSync` with `InvalidParameter` (HTTP 400), which is the intended behaviour for a request enum.

### Shared / version plumbing
- **main.tsp**: `v2026_11_01` version enum entry
- **readme.md**: `package-2026-11-01` tag + default-tag update
- **service.yaml**: 2026-11-01 typespec entry
- **stable/2026-11-01/service.json** + **examples/** (158) and **stable/2026-11-01/examples/** (158)

### Vault management and Backup version alignment
- **recoveryServices**: adds `2026-11-01`, carrying forward the stable `2026-07-01` vault-management contract and its 45 examples without new fields or operations.
- **recoveryServicesBackup**: adds `2026-11-01`, carrying forward the stable `2026-08-01` Backup contract and its 116 examples without new fields or operations.
- Both areas include TypeSpec version declarations, source and emitted examples, generated-contract files, README tags/defaults, and `service.yaml` entries. Existing published versions and the ASR files are unchanged by this follow-up.
- Follows the sibling-version pattern of PR #43641 and supplies all three Swagger references requested for [RegionalRP PR 17243711](https://dev.azure.com/msazure/One/_git/Mgmt-RecoverySvcs-RegionalRP/pullrequest/17243711).
- The sibling Swagger files were carried forward from the checked-in stable contracts. Local TypeSpec compilation was not run because npm execution is restricted; the PR automated checks must confirm generated output matches the committed files.

## Notes for reviewers
- `stable/2026-11-01/service.json` is fully regenerated from the TypeSpec sources. This also picks up the pending `ReplicationProtectedItems_AddDisks` `x-ms-examples` rename that was already present in the `.tsp` source but missing from the committed swagger — that drift was previously failing TypeSpec Validation on this branch.
- The version alignment preserves the latest PR contract and examples, including response-only `ipVersion`; previously published API versions are unchanged.

## Contribution checklist
- [x] I have reviewed the [contribution guidelines](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/gitclone.md).
- [x] Purely additive new api-version (no breaking changes to existing versions).
- [x] Examples added and referenced by `service.json`.


specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -8427,6 +8547,12 @@ model IPConfigDetails {
   tfoPublicIPAddressId?: string;
   #suppress "@azure-tools/typespec-azure-core/documentation-required" "FIXME: Update justification, follow aka.ms/tsp/conversion-fix for details"
   tfoLBBackendAddressPoolIds?: string[];
+
+  /**
+   * The IP version (IPv4 / IPv6) of the ip-config. Orthogonal to the allocation type (static / dynamic); a missing value is treated as IPv4.
```
umpatel: `ipVersion` is added to `IPConfigDetails` but not to `IPConfigInputDetails`, so a caller has no way to set it - yet the doc says "a missing value is treated as IPv4", which reads as caller-settable. Either add `ipVersion` to `IPConfigInputDetails`, or mark it `@visibility(Lifecycle.Read)` and reword the doc to describe what the service returns.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -6469,6 +6512,12 @@ model VaultSettingCreationInputProperties {
    * VMware to Azure provider type.
    */
   vmwareToAzureProviderType?: string;
+
+  /**
+   * The vault level policy controlling whether the A2A recovery network configuration is automatically kept in sync with the source virtual machine. An omitted value leaves the stored policy unchanged.
```
umpatel: Doc says "An omitted value leaves the stored policy unchanged", but `replicationVaultSettings/{vaultSettingName}` exposes only GET and PUT. PUT is a full replace, so an omitted value is not a merge. Either add a PATCH operation, or reword to state the default applied on create/replace.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -6657,6 +6712,45 @@ model KeyEncryptionKeyInfo {
   keyVaultResourceArmId?: string;
 }
 
+/**
+ * Confidential disk encryption info for a CMK confidential VM managed disk.
+ */
+@added(Versions.v2026_11_01)
+model ConfidentialDiskEncryptionInfo {
```
umpatel, line 6731: InMageRcm and VMwareCbt model this as a flat `confidentialDiskEncryptionSetId` on the disk; A2A introduces a nested object with two differently-named IDs. Align the shape/naming with the existing providers, or document why A2A diverges - otherwise SDK users see two unrelated patterns for the same concept.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -9124,6 +9270,12 @@ model A2AVmManagedDiskUpdateDetails {
    * The target disk name for test failover operation.
    */
   tfoDiskName?: string;
+
+  /**
+   * The confidential disk encryption info to update for the managed disk. Only the recovery target confidential disk encryption set can be updated after protection is enabled; the replica confidential disk encryption set is immutable. Applicable to CMK confidential VMs.
```
umpatel: The replica-DES immutability holds only because this update model omits the field. `A2AVmManagedDiskInputDetails` still carries `recoveryReplicaConfidentialDiskEncryptionSetId` and is reachable after protection is enabled via `A2AAddDisksInput` and `A2ASwitchProtectionInput`. Narrow the doc to the update path, or enforce immutability server-side on those operations.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -6444,6 +6481,12 @@ model VaultSettingProperties {
    * VMware to Azure provider type.
    */
   vmwareToAzureProviderType?: string;
+
+  /**
+   * The vault level policy controlling whether the A2A recovery network configuration is automatically kept in sync with the source virtual machine.
+   */
+  @added(Versions.v2026_11_01)
+  recoveryNetworkConfigAutoSync?: RecoveryNetworkConfigAutoSyncPolicy;
```
umpatel, line 6489: No example illustrates `recoveryNetworkConfigAutoSync: "Disabled"` - Create/Get/List all show `Enabled`. Add a `Disabled` example so the off-state round-trip is covered.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
```diff
@@ -7825,6 +7929,12 @@ model A2AProtectedManagedDiskDetails {
    */
   @added(Versions.v2026_02_01)
   recoveryPublicNetworkAccess?: DiskPublicNetworkAccess;
+
+  /**
+   * The confidential disk encryption info for the managed disk. Applicable to CMK confidential VMs.
```
umpatel, line 7947: `A2AProtectedManagedDiskDetails` is a read-shaped `*Details` model, but it is reachable on the `replicationProtectionClusters` PUT body through `sharedDiskProperties.sharedDiskProviderSpecificDetails.protectedManagedDisks`, making `confidentialDiskEncryptionInfo` caller-settable there with no example covering it. Add `@visibility(Lifecycle.Read)` if it is output-only; otherwise add a cluster example exercising the new field.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
sisunkar: Updated the model description in [b138876ae7](https://github.com/Azure/azure-rest-api-specs/commit/b138876ae72e3fbca4f56e300e02db40c6a096be) to explain the two A2A roles: the replica confidential disk encryption set is used for the replica disk during replication, and the target confidential disk encryption set is used for the target disk after failover. The existing nested object and field names are retained to match the service contract.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
sisunkar: Narrowed the description in [b138876ae7](https://github.com/Azure/azure-rest-api-specs/commit/b138876ae72e3fbca4f56e300e02db40c6a096be) to this update operation: "Only the recovery target confidential disk encryption set can be changed through this update operation. Applicable to A2A CMK confidential VMs." It no longer claims replica-DES immutability across AddDisks or SwitchProtection. The request fields are unchanged.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
sisunkar: Added `@visibility(Lifecycle.Read)` to `A2AProtectedManagedDiskDetails.confidentialDiskEncryptionInfo` and the matching `readOnly: true` in the 2026-11-01 Swagger in [b138876ae7](https://github.com/Azure/azure-rest-api-specs/commit/b138876ae72e3fbca4f56e300e02db40c6a096be). This marks the property as output-only in the contract. The separate enable/add-disk/update input properties remain writable; the shared `ConfidentialDiskEncryptionInfo` model is unchanged.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
bhanuteja2625: Good catch - the asymmetry is intentional, but the spec wasn't saying so. `ipVersion` is projected from the discovered NIC configuration on the read path only; there is no counterpart on `IPConfigInputDetails`, and the service does not read one (the request contract has no such member, so an `ipVersion` sent by a caller is silently dropped by the deserializer rather than honoured).

Marked it `@visibility(Lifecycle.Read)` (`readOnly: true` in the 2026-11-01 swagger) and reworded the doc, dropping the misleading `a missing value is treated as IPv4`. The field is omitted entirely when IPv6 support is not enabled, so absence means `not reported`, not `IPv4`.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
bhanuteja2625: The doc wording was wrong, but the merge behaviour it described is correct and deliberate, so I've reworded rather than changed the contract.

`ReplicationVaultSetting_Create` persists a property only when it is supplied (`if (this.RecoveryNetworkConfigAutoSync.HasValue)`), exactly as `migrationSolutionId` and `vmwareToAzureProviderType` already behave on this same resource. Making the PUT a true full replace would reset a customer's stored policy on any unrelated vault-setting update - a regression for the existing properties too, not just the new one.

New wording ties the behaviour to the resource's established convention:

> As with the other properties on this resource, the stored value is updated only when the property is supplied; omitting it retains the previously stored policy rather than resetting it.

specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery/models.tsp
bhanuteja2625: Added `ReplicationVaultSetting_Create_Disabled.json` (in both `examples/2026-11-01/` and `stable/2026-11-01/examples/`) and wired it into `x-ms-examples` for `ReplicationVaultSetting_Create`. It carries `recoveryNetworkConfigAutoSync: `Disabled` ` in both the request body and the 200 response, so the `Disabled` round-trip is now covered.

sisunkar:
The InMage RCM extended-location changes are included in this PR. Please add the matching TypeSpec source changes:

**1. Add this property to both models in `models.tsp`:**
- `InMageRcmEnableProtectionInput` - after `targetCapacityReservationGroupId`.
- `InMageRcmReplicationDetails` - after `targetAvailabilityZone`.

```typespec
/**
 * The target extended location.
 */
@added(Versions.v2026_11_01)
targetExtendedLocation?: ExtendedLocation;
```

Reuse the existing `ExtendedLocation` model; no new shared type is needed. This follows the existing extended-location property pattern and keeps the addition limited to API version `2026-11-01`.

**2. Add the source copy of the existing example.** Relative to `specification/recoveryservicessiterecovery/resource-manager/Microsoft.RecoveryServices/SiteRecovery`, copy:

```text
From: stable/2026-11-01/examples/ReplicationProtectedItems_Create_InMageRcmEdgeZone.json
To:   examples/2026-11-01/ReplicationProtectedItems_Create_InMageRcmEdgeZone.json
```

Keep the existing content, including `operationId: "ReplicationProtectedItems_Create"` and this value in both the request and response:

```json
"targetExtendedLocation": {
  "name": "edge-zone-1",
  "type": "EdgeZone"
}
```

`tspconfig.yaml` reads examples from the source `examples` directory. The [current TypeSpec failure](https://github.com/Azure/azure-rest-api-specs/actions/runs/35808547715/job/107014678888) shows regeneration removing both properties and the example reference because these source additions are missing. Please ensure the generated output matches the Swagger included in the PR.

gary-x-li:
Please fix Swagger Avocado and TypeSpec Validation failures first.

sisunkar:
Thanks, @gary-x-li. In response to your [review comment](https://github.com/Azure/azure-rest-api-specs/pull/46144#issuecomment-5797260428), [TypeSpec Validation now passes](https://github.com/Azure/azure-rest-api-specs/actions/runs/35887729718/job/107272069089) after merging the InMage RCM TypeSpec source changes in `320a837e3e`.

The remaining [Avocado failure](https://github.com/Azure/azure-rest-api-specs/actions/runs/35887729671) concerns Backup's existing active/passive API-version configuration. Certain CRR operations use a separate passive API-version series and are selected separately for SDK generation, as explained by the service team in [PR #41997](https://github.com/Azure/azure-rest-api-specs/pull/41997#issuecomment-4251276403).

Our default tag now selects the active `2026-11-01` contract. Avocado reports nine paths from the passive `2023-01-15` contract and one legacy registered-identities path from `2016-06-01`. These paths were also absent from the previous active `2026-08-01` contract. Both active versions contain the same 65 paths, with no added or removed HTTP operations.

The same ten-path issue was documented in [PR #44639](https://github.com/Azure/azure-rest-api-specs/pull/44639#issuecomment-5176963594), and the passive-version configuration received explicit approval in [PR #44057](https://github.com/Azure/azure-rest-api-specs/pull/44057#issuecomment-5255824140).

Changing these separate selections affects SDK generation, so we propose coordinating that cleanup separately with the Backup/SDK owners rather than combining the contracts in this version update.

Could you please review this explanation for `Approved-Avocado` and continue the ARM review?

PR #46157, opened 2026-09-08
[AzureResilienceManagement] Remove GoalTemplate resource type from 20…

harshb-15:
…26-08-31-preview

Remove the GoalTemplate resource type (GoalTemplates operations plus the GoalTemplate / GoalTemplateProperties models) starting at 2026-08-31-preview, via @removed(Versions.v2026_08_31_preview).

Part of Goal Template deprecation: resiliency posture now lives directly on GoalAssignment and GoalResource, so the GoalTemplate resource type is no longer part of the API surface. Consumers still needing GoalTemplates should use an earlier preview api-version (2025-02, 2026-03, 2026-04, 2026-06), which are left unchanged.

Also removes the now-orphaned GoalTemplates_* examples for 2026-08-31-preview.

# Choose a PR Template


specification/azureresiliencemanagement/suppressions.yaml
```diff
@@ -23,3 +23,29 @@
     Older preview API version retained for compatibility. The default tag points to
     2026-04-01-preview; consumers needing 2025-02-01-preview should use the per-version
     tag package-2025-02-01-preview.
+
+- tool: Swagger Avocado
```
mkherani, line 27: Not sure if we can add this suppression. We need to check with Vivek for better understanding. What was the error for which we are supressing?

PR #46158, opened 2026-09-08
SDK Breaking Change Analysis workflow and SDK Breaking Change mitigation workflow

### LINKED ISSUE #16915: SDK breaking change analysis workflow in azure-rest-api-specs repo

### LINKED ISSUE #16916: SDK breaking change mitigation workflow in azure-rest-api-specs repo

chunyu3:
Fix https://github.com/Azure/azure-sdk-tools/issues/16915
Fix https://github.com/Azure/azure-sdk-tools/issues/16916

## SDK Breaking Change Analysis workflow

- implement a SDK breaking change analysis workflow. This workflow will detect and analyze the SDK breaking changes for a language and surface the result in the comment (see following example)

## SDK Breaking changes for Go

**Analyzed commit:** `6e9fca19c6df9cf66be7e753255d79251f1d4a91`

**Typespec Project:** specification/webpubsub/resource-manager/Microsoft.SignalRService/SignalRService

**SDK Package:** armwebpubsub

|  | Breaking change | Category | Suggested fix |
| --- | --- | --- | --- |
| ☐ | Struct ResourceInfo for TypeSpec model WebPubSubResource was renamed to Resource, changing all Client create, update, and response usages. | conversion-need resolve | Add @@clientName(WebPubSubResource, "ResourceInfo", "go") to client.tsp to preserve the legacy Go SDK name. |
| ☐ | Struct ResourceInfoList for TypeSpec model WebPubSubResourceList was renamed to ResourceList, changing both Client list response fields. | conversion-need resolve | Add @@clientName(WebPubSubResourceList, "ResourceInfoList", "go") to client.tsp to preserve the legacy Go SDK name. |
| ☐ | Field `Family`, `Tier` of struct `ResourceSKU` has been removed | spec change | - |

[SDK breaking change details (armwebpubsub/breaking-changes.json)](https://github.com/chunyu3/azure-rest-api-specs/actions/runs/35192392517#summary-105107677718)

- The workflow will be triggered by comment:  /azsdk sdk-breaking-analysis <sdk-language>
- When `SDK Breaking Change Labels` workflow completed and find out there are SDK breaking change detected for an language, it will comment '/azsdk sdk-breaking-analysis <tsp-config-pth> <sdk-language>' to trigger this 'SDK Breaking Change Analysis' workflow

## SDK Breaking Change Mitigation workflow

- implement a SDK breaking change analysis workflow. The workflow will mitigate the SDK breaking changes detected in 'SDK breaking change analysis' workflow
- The workflow will be triggered by comment:  /azsdk sdk-breaking-mitigate <sdk-language>


.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,590 @@
...
+  pull-requests: write
+
+jobs:
+  analyze:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request != null &&
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association) &&
+      startsWith(github.event.comment.body, '/azsdk sdk-breaking-analysis '))
```
samvaity: This introduces an operator-facing command and a new recovery/decision loop, but the consumer documentation does not expose it. `documentation/ci-fix.md` is the established entry point for `BreakingChange-{Language}-Sdk`: it identifies spec authors/SDK reviewers, the two-business-day ownership boundary, and suppression escalation, yet it currently contains no command syntax, supported language values, explanation of the generated table, rerun behavior after a spec fix, or failure/artifact recovery (the artifact is retained only seven days here). Those are concrete steps a PR author now needs to act on this workflow. Please extend the existing **SDK Breaking Change Review** section in `documentation/ci-fix.md` with this flow and link it from the result/failure text. A separate standalone design/process document is not necessary—the existing section is the authoritative home—but shipping the hidden slash-command contract without updating that process leaves its intended users unable to operate or recover it.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,590 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  analyze:
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+      github.event.issue.pull_request != null &&
+      contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association) &&
```
samvaity, line 33: The automatic trigger described by this PR cannot reach this job. The existing `sdk-breaking-change-labels.yaml` workflow only uploads label artifacts; it never creates the `/azsdk sdk-breaking-analysis ...` comment. In addition, a comment created with its `GITHUB_TOKEN` would be authored by `github-actions[bot]` with `author_association: NONE` (as can be verified on existing Actions-authored comments in this repository), so this allow-list would reject it anyway. Please wire the label workflow (or the downstream label applicator) to emit the command and authenticate that specific automation identity without allowing arbitrary untrusted commenters; otherwise only manual member comments/workflow dispatches work and the advertised automatic analysis never runs.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,590 @@
...
+        with:
+          python-version: "3.13"
+
+      - name: Set up Python SDK environment
+        if: steps.resolve-source.outputs.sdk-language == 'Python'
+        shell: bash
+        working-directory: repositories/${{ steps.resolve-source.outputs.sdk-repository }}
+        run: |
+          python -m pip install -U pip > /dev/null
+          python -m pip install eng/tools/azure-sdk-tools[ghtools] > /dev/null
```
samvaity: This unquoted extras specifier is parsed by ShellCheck as a bracket expression (`SC2102`), and the latest Ubuntu workflow test fails on this exact line. Quote the requirement, e.g. `python -m pip install 'eng/tools/azure-sdk-tools[ghtools]'`, so the required `.github` test check can pass.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,590 @@
...
+              }),
+            );
+
+            const report = [
+              `## SDK Breaking changes for ${process.env.SDK_LANGUAGE}`,
+              "",
+              ...projectReports.flatMap((projectReport) => [...projectReport, ""]),
+            ].join("\n");
+
+            if (process.env.TRIGGER_NAME === "issue_comment") {
```
samvaity, line 591: This report is not tied to the head that was analyzed. The job snapshots `pull.head.sha` near the start, then performs several installs plus generation/build/analysis, and finally updates the PR comment unconditionally; there is also no per-PR/language `concurrency` group. If the PR is synchronized while a run is in flight, the old run can publish (or overwrite the command with) results for the previous SHA after the new commit is already current, and the comment does not identify that SHA. Neighboring long-running review automation (`arm-api-review.md`) cancels superseded per-PR runs, while the SDK validation/status and label handoffs explicitly propagate `head-sha`. Please add PR/language concurrency with `cancel-in-progress`, and re-fetch the PR before publishing to verify that `pull.head.sha` still equals `steps.resolve-source.outputs.ref` (otherwise mark/skip the run as superseded). Include the analyzed SHA in any retained result for auditability.

.github/workflows/sdk-breaking-change-analysis.yaml
samvaity: I’d revise the placement recommendation: this is complex enough to warrant a dedicated, authoritative process document, not just a longer `ci-fix.md` section. The precedent is `.github/workflows/src/package-name-approval/PACKAGE-NAME-REVIEW-PROCESS.md`, which exists because package-name approval has multiple actors, automated handoffs, decision points, reruns, and recovery paths; this workflow has the same operational characteristics. The dedicated document should own the end-to-end contract (triggers, ownership, interpretation, mitigation, reruns/supersession, artifacts, failures, suppressions, and escalation), with `documentation/ci-fix.md` linking to it from **SDK Breaking Change Review**. That is different from an implementation README, which may explain code and artifact schemas but should not be the contributor-facing source of truth.

chunyu3: Thanks for pointing this out. I will add this document.

.github/workflows/sdk-breaking-change-analysis.yaml
chunyu3: Thanks for pointing this out. `sdk-breaking-change-labels.yaml` workflow will be updated to create the `/azsdk sdk-breaking-analysis ...` comment and trigger `sdk-breaking-change-analysis.yaml` automatically. @raych1 will work on this.

.github/workflows/sdk-breaking-change-analysis.yaml
chunyu3: Thanks for pointing this out. I’ve added per-PR/language `concurrency` and verify `pull.head.sha` before publishing to prevent stale analysis results. Also included the analyzed commit to the report.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,691 @@
...
+            }
+
+            core.setOutput("repository", pull.head.repo.full_name);
+            core.setOutput("sha", pull.head.sha);
+            core.setOutput("sdk-repository", languageConfig.repository);
+            core.setOutput("sdk-language", languageConfig.language);
+            core.setOutput("tsp-config-path", tspConfigPath ?? "");
+            core.setOutput("changed-typespec-files", JSON.stringify(changedTypespecFiles));
+
+      - name: Verify pull request was not updated after the trigger
```
raych1: This step does not seem necessary. The label workflow will be triggered when the PR is updated.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,691 @@
...
+        id: resolve-source
+        uses: actions/github-script@v9
+        env:
+          PR_NUMBER: ${{ github.event.issue.number || inputs.pr_number }}
+          SDK_LANGUAGE_INPUT: ${{ inputs.sdk_language }}
+          TSP_CONFIG_PATH_INPUT: ${{ inputs.tsp_config_path }}
+          TRIGGER_NAME: ${{ github.event_name }}
+          TRIGGER_COMMENT: ${{ github.event.comment.body }}
+        with:
+          script: |
```
raych1: Extract the inline scripts into individual script files so that they can be tested by unit test. Refer to: https://github.com/Azure/azure-rest-api-specs/tree/main/.github/workflows/src

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,691 @@
...
+
+      - name: Checkout target SDK repository
+        uses: actions/checkout@v7
+        with:
+          repository: Azure/${{ steps.resolve-source.outputs.sdk-repository }}
+          ref: main
+          path: repositories/${{ steps.resolve-source.outputs.sdk-repository }}
+          persist-credentials: false
+
+      - name: Set up Go
```
raych1, line 100: We should rely on the verify_setup tool in azsdk-cli to install language specific runtime and tools to avoid diverge of the env settings.

chunyu3: @raych1 We use `verify_setup` tool to setup SDK environment in step `Set up SDK development environment`. but `verify_setup` tool is not stable enough, and it has problem to set up environment for Go and Python. Current we have to define the special steps to setup environment for Go and Python. In the future, when we setup misc or `verify_setup` tool is stable, we will remove those additional setup step and rely on setup tool only.

.github/workflows/sdk-breaking-change-analysis.yaml
samvaity: The SHA/supersession fix is still incomplete: this checkout does not use the validated `head-sha`; it checks out GitHub's synthetic `refs/pull/<n>/merge` commit. That commit is different from the SHA recorded in the artifact/report, and GitHub can regenerate it against a newer base between resolution and checkout. The workflow can therefore analyze code other than the commit it claims to analyze. Please checkout `needs.resolve.outputs.head-sha` exactly (and keep the publish-time head check) so the recorded provenance matches the executed source.

.github/workflows/src/sdk-breaking-change/resolve-analysis-inputs.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  const language = SDK_LANGUAGES_BY_LABEL.get(
+    labelArtifact.labelName as (typeof SDK_LABELS)[number],
+  );
+  const languageConfig = resolveSdkLanguageConfig(language);
+
+  core.setOutput("pr-number", issueNumber);
+  core.setOutput("head-repository", pull.head.repo.full_name);
+  core.setOutput("head-sha", pull.head.sha);
+  core.setOutput("sdk-language", language);
```
samvaity, line 163: The automatic .NET handoff emits `.NET` here, but the workflow's setup condition compares this output to `DotNet`, and `azsdk verify setup --languages` validates against the `SdkLanguage` enum (`DotNet`, not `.NET`). A `BreakingChange-.Net-Sdk` artifact therefore skips `actions/setup-dotnet` and then fails setup before generation. Emit the normalized `languageConfig.language` value here (and in the dispatch branch) while retaining a separate display label if needed.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,345 @@
...
+              relative_results_path="$package_name-$project_number"
+            else
+              relative_results_path="$package_name"
+            fi
+            package_names["$package_name"]=1
+            project_results="$results_dir/$relative_results_path"
+            mkdir -p "$project_results"
+            mv "$generation_result" "$project_results/generate.json"
+
+            azsdk package build \
```
samvaity, line 240: The .NET sequence cannot produce the analysis promised by this workflow. The authoritative breaking-change skill requires this preparatory build to pass `--additional-arguments "/p:RunApiCompat=false"`; without it an ApiCompat break stops this `set -e` step before the dedicated detector runs. After that is fixed, the current `azure-sdk-for-net/eng/swagger_to_sdk_config.json` still has no `getSdkChangesScript`, so `detect-breaking-change` falls back to `LanguageService.DetectSdkBreakingChangeAsync` and returns `SDK Breaking Change Detection is not implemented for language DotNet.` Please both pass the required build argument and provide a supported .NET detector (or stop advertising/triggering .NET until one exists).

.github/workflows/sdk-breaking-change-mitigate.yaml
```diff
@@ -0,0 +1,313 @@
+name: SDK Breaking Change Mitigation
+
+on:
+  workflow_run:
```
samvaity, line 4: This automatic handoff bypasses the selection boundary in the authoritative workflow. The analysis comment renders unchecked boxes, but no code reads their state; every `breakingChanges[]` entry is passed to `customized-update` as soon as publishing succeeds. Thus a detection-only result immediately starts Copilot-driven spec edits, including changes the author may intend to accept/suppress. Keep analysis detection-only by default and require an explicit authorized mitigation request carrying the selected findings (or remove the selection UI and align the approved contract before enabling automatic mitigation).

.github/workflows/sdk-breaking-change-mitigate.yaml
```diff
@@ -0,0 +1,313 @@
...
+                  --output json | tee "$mitigation_result"; then
+                mitigation_succeeded=false
+              fi
+
+              if [[ "$mitigation_succeeded" == "true" ]] && jq -e . "$mitigation_result" > /dev/null; then
+                clean_result="$(
+                  jq -c \
+                    --argjson breakingChange "$breaking_change" \
+                    '(.result // .) as $result |
+                    ($result.success // false) as $isResolved | {
```
samvaity, line 218: `result.success` only says `customized-update` completed; that tool regenerates/builds but does not rerun `detect-breaking-change`. Marking the item resolved from this bit can therefore report a false resolution when the proposed customization did not remove the break, or when a later customization reintroduces an earlier one. Rerun the detector after mitigation (as the authoritative skill requires), and derive resolved/remaining findings from that result rather than from customization success.

.github/workflows/sdk-breaking-change-mitigation-status.yaml
```diff
@@ -0,0 +1,52 @@
+name: SDK Breaking Change Mitigation - Publish Results
+
+on:
+  workflow_run:
```
samvaity, line 4: This publisher is beyond GitHub's supported `workflow_run` chain depth. The automatic path is `SDK Breaking Change Labels` → `Analysis - Analyze` → `Analysis - Publish Results` → `Mitigation` → this workflow; GitHub documents that workflows after the third chained level do not run. Consequently automatic mitigation can finish but its result is never posted. Collapse a stage or start a new chain via an explicit dispatch/API event so publication is reachable.

.github/workflows/sdk-breaking-change-analysis.yaml
samvaity: The dedicated document and `ci-fix.md` link address the placement part, but this is only a partial resolution of the operational gap. The current document still describes only the happy-path automatic flow: it does not document the manual dispatch inputs/run-ID handoff, supported languages, stale/superseded runs, failed generation/build/detection recovery, the seven-day artifact window, or where to retrieve/apply a mitigation when no PR is actually created. Please add those operator steps here before resolving this thread; the example currently claims a “Mitigation pull request” even though the workflow only publishes a transient diff.

.github/workflows/sdk-breaking-change-mitigation-status.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+permissions:
+  actions: read
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  publish-results:
+    if: github.event.workflow_run.conclusion == 'success'
```
samvaity, line 16: Even if this publisher is made reachable, it explicitly skips failed mitigation runs. The mitigation job exits immediately when `customized-update` fails, before its non-`always()` artifact upload, and this `if` then suppresses the publisher entirely. A concrete customization failure therefore leaves the PR with the prior analysis comment and no indication that mitigation failed or where to recover. Produce/upload a failure result with `always()` and let this publisher handle `failure`, as the analysis publisher already does.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,345 @@
...
+
+      - name: Install Azure SDK CLI
+        shell: pwsh
+        run: ./eng/common/mcp/azure-sdk-mcp.ps1 -InstallDirectory $env:AZSDK_CLI_PATH
+
+      - name: Checkout target SDK repository
+        uses: actions/checkout@v7
+        with:
+          repository: Azure/${{ needs.resolve.outputs.sdk-repository }}
+          ref: main
```
samvaity, line 99: The SDK-side baseline is mutable and is not recorded. Analysis checks out `main` here, while the later mitigation workflow independently checks out `main` again; if the SDK repository advances between those runs, mitigation operates on a different package baseline than the one that produced the reported breaks. The artifact records only the spec SHA, so the result is not reproducible or auditable either. Resolve and record the SDK repository commit during analysis, include it in the artifact/report, and checkout that exact SHA during mitigation.

.github/workflows/sdk-breaking-change-analysis.yaml
raych1: I'd rather fixing the setup tool instead of adding own installation steps here.

.github/workflows/sdk-breaking-change-analysis.yaml
```diff
@@ -0,0 +1,691 @@
...
+          MANIFEST_PATH: ${{ runner.temp }}/sdk-breaking-change-results/projects.json
+          RESULTS_PATH: ${{ runner.temp }}/sdk-breaking-change-results
+          PR_NUMBER: ${{ github.event.issue.number || inputs.pr_number }}
+          SDK_LANGUAGE: ${{ steps.resolve-source.outputs.sdk-language }}
+          ANALYZED_SHA: ${{ steps.resolve-source.outputs.sha }}
+          TRIGGER_NAME: ${{ github.event_name }}
+        with:
+          github-token: ${{ github.token }}
+          script: |
+            const fs = await import("node:fs/promises");
```
timotheeguerin, line 541: this is a lot of inline js, can this not be moved to a tool we can call locally, having CI only check repeat the bad state we were in before

.github/workflows/sdk-breaking-change-mitigate.yaml
chunyu3: Thanks for pointing this out. The current mitigation workflow only attempts the fix locally and provides the generated customization code to the PR owner; it does not modify the PR directly.

The PR owner will apply the customization for the breaking changes they want to resolve and push a new commit. That commit will retrigger the `SDK validation pipeline` → `SDK breaking change labels` → `SDK breaking change analysis (if breaking changes remain)` → `mitigation workflow`. So the actual PR state will be validated again by the breaking-change detector.

.github/workflows/sdk-breaking-change-analysis.yaml
chunyu3: @samvaity  .NET will integrate to the breaking change and detection framework soon.
>The authoritative breaking-change skill requires this preparatory build to pass --additional-arguments "/p:RunApiCompat=false";

pass additional build command arguments will be supported in PR https://github.com/Azure/azure-sdk-tools/pull/16634. And I will update the common here for .NET after the skill PR is merged.

>After that is fixed, the current azure-sdk-for-net/eng/swagger_to_sdk_config.json still has no getSdkChangesScript, so detect-breaking-change falls back to LanguageService.DetectSdkBreakingChangeAsync

.NET will implemented this in PR https://github.com/Azure/azure-sdk-for-net/pull/62729

.github/workflows/sdk-breaking-change-mitigation-status.yaml
chunyu3: Thanks for pointing this out. I’ll review the workflow chain and the workflow_run depth limitation. If that’s the cause, we’ll consider collapsing a stage or using a different trigger mechanism.

.github/workflows/src/sdk-breaking-change/run-command.ts
```diff
@@ -0,0 +1,65 @@
...
+export type RunCommandOptions = {
+  command: string;
+  args: string[];
+  logPath: string;
+  outputPath: string;
+  env?: NodeJS.ProcessEnv;
+  throwOnFailure?: boolean;
+};
+
+export async function runCommand({
```
timotheeguerin: we already have helpers for that with `execFile` reuse that

.github/workflows/src/sdk-breaking-change/publish-mitigation-result.ts
```diff
@@ -0,0 +1,139 @@
...
+    breakingChanges: Array<{
+      breakingChange: string;
+      suggestedFix: string;
+      isResolved: boolean;
+      typespecChangesSummary?: string[];
+    }>;
+  }>;
+};
+
+function escapeTableCell(value: unknown, fallback = "-"): string {
```
timotheeguerin: i added some markdown helpers in shared/markdown, try to reuse those

.github/workflows/src/sdk-breaking-change/mitigate-sdk-breaking-changes.ts
```diff
@@ -0,0 +1,206 @@
...
+    const packagePaths = await Promise.all(
+      locationPaths
+        .map((locationPath) => dirname(locationPath))
+        .filter((packagePath) => basename(packagePath) === project.sdkPackage)
+        .map((packagePath) => realpath(packagePath)),
+    );
+    if (packagePaths.length !== 1 || !isWithin(sdkRepositoryPath, packagePaths[0])) {
+      throw new Error(
+        `Expected exactly one SDK package named ${project.sdkPackage}, found ${packagePaths.length}.`,
+      );
```
timotheeguerin: **[P1] Do not require existing TypeSpec metadata to locate a conversion's SDK package.**

Mitigation checks out a fresh SDK `main`, not the generated SDK checkout from analysis. For a Swagger-to-TypeSpec conversion, the existing package has no `tsp-location.yaml`; the file created during analysis is not transferred. This lookup therefore throws `Expected exactly one SDK package named ..., found 0` before attempting customization. Swagger-only packages such as `azure-sdk-for-go/sdk/resourcemanager/aad/armaad` have this shape.

Please carry the validated SDK-relative package path in the analysis artifact and initialize/regenerate the package from the analyzed spec before customization, rather than requiring the baseline package to have already migrated to TypeSpec.

.github/workflows/src/sdk-breaking-change/publish-analysis-result.ts
```diff
@@ -0,0 +1,88 @@
...
+    return;
+  }
+  const { data: pull } = await github.rest.pulls.get({
+    ...context.repo,
+    pull_number: result.prNumber,
+  });
+  if (pull.head.sha !== result.headSha) {
+    const staleMessage = `Analysis result for ${result.headSha} is stale; current PR head is ${pull.head.sha}.`;
+    core.warning(staleMessage);
+    await publishResultInComment({ github, context, core }, result.prNumber, command, staleMessage);
```
timotheeguerin: **[P2] Leave the current report untouched when an older run is superseded.**

This detects a stale SHA but still calls `publishResultInComment`, which updates the newest comment for the same language. If the current-head run has already published, a delayed older publisher replaces that valid report with a stale warning. The failure branch above is worse: it updates the comment without checking the head at all. Analysis-job concurrency does not serialize these separate publisher workflows.

Please check freshness before both success and failure publication, and return without replacing the current report when stale. Apply the same behavior to `publishMitigationResult`, which currently also publishes its stale warning through the shared updater.

.github/workflows/sdk-breaking-change-mitigate.yaml
```diff
@@ -0,0 +1,219 @@
...
+          set -euo pipefail
+          git add --intent-to-add --all
+          changed_code="$(git diff --no-ext-diff --no-color)"
+          max_snippet_length=12000
+          if [[ ${#changed_code} -gt $max_snippet_length ]]; then
+            changed_code="${changed_code:0:$max_snippet_length}"$'\n... diff truncated.'
+          fi
+          jq --arg customizationCode "$changed_code" \
+            '.customizationCode = $customizationCode' \
+            "$MITIGATION_RESULT_PATH" > "$MITIGATION_RESULT_PATH.tmp"
```
timotheeguerin: **[P2] Preserve the full mitigation patch in the artifact.**

The 12,000-character limit is applied before writing `customizationCode` to the only uploaded artifact. The workflow summary receives the same truncated value, so its link does not provide a complete version either. A larger mitigation is permanently lost when the runner is discarded, and truncation can cut through a diff hunk, making it impossible to apply.

Please upload the complete patch separately, or retain it intact in the artifact, and truncate only the comment/summary preview.

.github/workflows/src/sdk-breaking-change/resolve-analysis-inputs.ts
```diff
@@ -0,0 +1,321 @@
...
+    while (directory.startsWith("specification/")) {
+      const configPath = `${directory}/tspconfig.yaml`;
+      try {
+        await github.rest.repos.getContent({
+          ...context.repo,
+          path: configPath,
+          ref: headSha,
+        });
+        configPaths.add(configPath);
+        break;
```
timotheeguerin: **[P2] Resolve projects affected by shared-library changes, not only ancestor configs.**

A PR changing only `specification/contosowidgetmanager/Contoso.WidgetManager.Shared/main.tsp` reaches `No tspconfig.yaml could be resolved`, even though `Contoso.WidgetManager/main.tsp` imports that library and its sibling project has a config. This loop cannot discover sibling consumers; a mixed PR can also omit affected projects while reporting the ones it did find.

The SDK runner already handles shared libraries in `eng/tools/spec-gen-sdk-runner/src/spec-helpers.ts` through `searchSharedLibrary` and `searchRelatedTypeSpecProjectBySharedLibrary`. Please reuse/extract that discovery behavior so SDK validation and analysis agree on the affected projects, and cover a shared-library-only change.

.github/workflows/sdk-breaking-change-analysis-status.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Install dependencies for github-script actions
+        uses: ./.github/actions/install-deps-github-script
+
+      - name: Download SDK analysis results
+        uses: actions/download-artifact@v8
+        with:
+          name: sdk-breaking-change-results
+          path: ${{ runner.temp }}/sdk-breaking-change-results
+          github-token: ${{ github.token }}
+          run-id: ${{ github.event.workflow_run.id }}
```
timotheeguerin: **[P2] Distinguish an intentionally skipped analysis from a missing result artifact.**

When the resolver finds no supported label, or the label artifact says `false`, it sets `should-run=false`. The resolve job can succeed while the analyze job is skipped, so this publisher still runs for a successful workflow and unconditionally downloads an artifact that was never created. A normal no-op becomes a failed publishing workflow.

Please propagate/inspect the intentional skip outcome before downloading. Keep an unexpectedly missing artifact on a run that actually attempted analysis as an error rather than silently ignoring every missing artifact.

.github/workflows/src/sdk-breaking-change/publish-result-in-comment.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const body = `${command}\n\n${content}`;
+  const comments = await github.paginate(github.rest.issues.listComments, {
+    ...context.repo,
+    issue_number: pullNumber,
+    per_page: 100,
+  });
+  const existingComment = comments
+    .filter(({ body }) => body?.trimStart().toLowerCase().startsWith(command.toLowerCase()))
+    .sort((left, right) => right.id - left.id)[0];
```
timotheeguerin: **Reuse the existing comment updater.**

`../comment.ts` already provides `commentOrUpdate`, including pagination, a stable HTML-comment identifier, logging, and avoiding an update when the body is unchanged. Please use it with a language-specific identifier such as `sdk-breaking-analysis-Go` instead of introducing another upsert implementation keyed by a visible slash-command prefix. If a genuinely required behavior is missing, extend the existing helper rather than maintaining a parallel one.

.github/workflows/src/sdk-breaking-change/analyze-sdk-projects.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+  return matches.flat();
+}
+
+async function findGeneratedConfigs(repositoryPath: string, startedAt: number): Promise<string[]> {
+  const configs = await findFiles(repositoryPath, "tsp-location.yaml");
+  const generated = await Promise.all(
+    configs.map(async (path) => ((await stat(path)).mtimeMs >= startedAt ? path : undefined)),
+  );
+  return generated.filter((path): path is string => path !== undefined);
```
timotheeguerin: **Use the existing metadata-based package discovery rather than file modification times.**

`../../../shared/src/typespec-metadata.ts` already exposes `generateTypeSpecMetadata` with language-specific `packageName` and `outputDir`. The SDK runner uses that metadata and `resolvePackagePath` in `eng/tools/spec-gen-sdk-runner/src/command-helpers.ts` to locate generated packages.

Please reuse/extract that path-resolution logic instead of recursively scanning the whole SDK repository and inferring package identity from which `tsp-location.yaml` happened to be touched after `Date.now()`. Persist the validated relative package path in the handoff so mitigation does not have to rediscover it by basename.

.github/workflows/src/sdk-breaking-change/analyze-sdk-projects.ts
```diff
@@ -0,0 +1,155 @@
...
+  const matches = await Promise.all(
+    entries.map(async (entry): Promise<string[]> => {
+      const path = join(directory, entry.name);
+      if (entry.isDirectory()) {
+        return findFiles(path, name);
+      }
+      return entry.isFile() && entry.name === name ? [path] : [];
+    }),
+  );
+  return matches.flat();
```
timotheeguerin: **Avoid duplicating the filesystem helpers between analysis and mitigation.**

This recursive `findFiles` implementation and the `isWithin` function below are repeated in `mitigate-sdk-breaking-changes.ts`. Once package discovery uses explicit metadata, remove the scans that are no longer necessary. Any filesystem/path-containment logic still needed by both entry points should live in one shared helper so its behavior and edge-case coverage cannot drift.

.github/workflows/src/sdk-breaking-change/publish-mitigation-result.ts
```diff
@@ -0,0 +1,139 @@
...
+  github,
+  context,
+  core,
+  mitigationResultPath,
+  workflowSummaryUrl,
+}: Pick<AsyncFunctionArguments, "github" | "context" | "core"> & {
+  mitigationResultPath: string;
+  workflowSummaryUrl: string;
+}): Promise<{ report: string; result: MitigationResult }> {
+  const result = JSON.parse(await readFile(mitigationResultPath, "utf8")) as MitigationResult;
```
timotheeguerin: **Use one validated mitigation-result schema on both sides of the artifact handoff.**

`MitigationResult` is separately defined here and in `mitigate-sdk-breaking-changes.ts`, and the contracts already differ: this reader accepts a failure status while the producer type only permits success; the producer includes a schema version that this reader does not check. Casting `JSON.parse` to the local type does not validate the downloaded data.

Please follow the analysis artifact's `AnalysisResultSchema` pattern: define one shared Zod schema, infer its TypeScript type, and validate both the produced and consumed mitigation artifact against it.

.github/workflows/src/sdk-breaking-change/mitigate-sdk-breaking-changes.ts
```diff
@@ -0,0 +1,206 @@
...
+      if (succeeded) {
+        try {
+          cleanResult = cleanCustomizedUpdateResult(
+            JSON.parse(await readFile(commandResultPath, "utf8")),
+            change.breakingChange,
+            change.suggestedFix,
+          );
+        } catch {
+          cleanResult.isResolved = false;
+        }
```
timotheeguerin: **Do not silently turn malformed tool output into an ordinary unresolved finding.**

If the command exits successfully but its JSON is malformed or no longer matches the expected schema, this broad catch discards the parsing error and the overall result can still be published with `status: "success"`. That hides a tooling/contract failure as though customization merely could not resolve the break.

Please preserve the parsing/validation diagnostic and report an explicit failure or per-item error in the artifact. The operator needs to distinguish an unresolved breaking change from a response the workflow could not interpret.

.github/workflows/src/sdk-breaking-change/resolve-mitigation-inputs.ts
```diff
@@ -0,0 +1,41 @@
...
+    pull_number: analysis.prNumber,
+  });
+  if (pull.head.sha !== analysis.headSha) {
+    throw new Error(
+      `Pull request head ${pull.head.sha} does not match analyzed head ${analysis.headSha}.`,
+    );
+  }
+
+  core.setOutput("should-run", "true");
+  core.setOutput("pull-number", analysis.prNumber);
```
timotheeguerin: **Match the workflow's `pr-number` output contract.**

The YAML declares `pr-number: ${{ steps.resolve.outputs.pr-number }}` and later reads it into `PR_NUMBER`, but this function emits `pull-number`. The workflow output is consequently empty and the summary renders `**Pull request:** #`.

Please emit `pr-number` here and update the resolver assertion to match the consuming YAML rather than locking in the mismatched name.

.github/workflows/src/sdk-breaking-change/SDK-BREAKING-CHANGE-PROCESS.md
```diff
@@ -0,0 +1,87 @@
...
+| ----------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- |
+| Struct ResourceInfo for TypeSpec model WebPubSubResource was renamed to Resource, changing the model used by create, update, get, and their response types. | Add @@clientName(WebPubSubResource, "ResourceInfo", "go") to client.tsp to preserve the legacy Go SDK name.         |
+| Struct ResourceInfoList for TypeSpec model WebPubSubResourceList was renamed to ResourceList, changing the embedded list result in list responses.          | Add @@clientName(WebPubSubResourceList, "ResourceInfoList", "go") to client.tsp to preserve the legacy Go SDK name. |
+
+**Mitigation pull request:** https://github.com/chunyu3/azure-rest-api-specs/pull/59
```
timotheeguerin: **Document the mitigation output this workflow actually produces.**

This example promises a mitigation pull request, but the implementation only publishes customization code in a comment/summary and an artifact; it does not create or update a PR. Please replace this with the actual output and explicitly tell the owner how to retrieve/apply the proposed patch, push it to their existing PR, and obtain a fresh analysis. Otherwise the process sends contributors looking for a mitigation PR that will never appear.

PR #46171, opened 2026-09-08
[WIP] Add credential validation ARM service v5 - internal escalation

hasskooelhadi:
Adding new ARM spec for credential validation service v5.


hasskooelhadi:
Updating specification for v6 - mseng enumeration payload active

PR #46173, opened 2026-09-08
Add knowledge base tool type

islegare:


erjohnms (approved):

erjohnms (approved):

specification/ai-foundry/data-plane/Foundry/src/tools/models.tsp
```diff
@@ -183,6 +184,31 @@ namespace Azure.AI.Projects {
...
+  }
+
+  /**
+   * The input definition information for a knowledge base tool as used to configure an agent.
+   */
+  model KnowledgeBaseTool extends OpenAI.Tool {
+    /**
+     * The object type, which is always 'knowledge_base'.
+     */
+    type: "knowledge_base";
```
dargilco: Should this tool be in preview first, with "knowledge_base_preview" as the type discriminator?

dargilco: Please use the union literal instead of repeating the string. So it would be 
type: _AgentToolType.knowlege_base

islegare: I tried referencing `_PreviewAgentToolType.knowledge_base_preview` directly, but TypeSpec 1.15 reports `override-property-mismatch` and `missing-discriminator-property`. The variant is added to `OpenAI.ToolType` through `@@copyVariants`, and copied variants are not recognized as assignable union-member references during discriminator validation. I kept the string literal so `tsp compile .` succeeds while retaining the discriminator in `_PreviewAgentToolType`.

islegare: Updated the tool to use the `knowledge_base_preview` discriminator and moved the value from `_AgentToolType` to `_PreviewAgentToolType`.

PR #46022, opened 2026-09-01
[TEST - WILL DELETE] verify data-plane team reviewer resolution

samvaity:
Throwaway PR to test whether requesting the `azure-data-plane-api-reviewers` team via REST resolves (422 diagnosis for #45985). Will be closed immediately.

PR #46180, opened 2026-09-08
Fix data-plane reviewer assignment 422 via azure-sdk-automation App token (#45985)

samvaity:
Fixes the data-plane reviewer-assignment 422 (#45985) by minting an installation token from the **azure-sdk-automation** GitHub App for the assignment step.

## Problem
The assignment workflow requests the `azure-data-plane-api-reviewers` team via `requestReviewers({ team_reviewers })`. The default Actions `GITHUB_TOKEN` cannot read org team membership, so team resolution fails with **HTTP 422** and no reviewer is ever assigned.

## Fix 
Mint a short-lived token from the `azure-sdk-automation` App (installed on the Azure org with `members: read`) and use it as the `github-token` for the `github-script` step. The App token makes the `requestReviewers({ team })` call itself, so GitHub's native team auto-delegation still picks the member. `assign-reviewers.js` is unchanged.

The minted token is scoped to only what this job needs: `members:read`, `pull-requests:write`, `issues:read`, `actions:read`, `contents:read`.

## Confirmed this works
- A token with only `Organization -> Members: Read` resolves the closed team (200) - validated with a minimal fine-grained PAT.
- The same `requestReviewers({ team })` call returned 200 + auto-delegated to a member when run with an org-read token (throwaway PR #46022).

## Prerequisites before this activates (draft until then)
- [ ] Provision Actions secrets `AZURE_SDK_AUTOMATION_APP_ID` + `AZURE_SDK_AUTOMATION_PRIVATE_KEY` (repo or org).
- [ ] Confirm `azure-rest-api-specs` is in the App's selected repositories (needed for `pull-requests:write` on this repo).

## Safety
- Gated on the App ID secret: until it lands, the token step is skipped and the job falls back to `GITHUB_TOKEN` (team resolution 422s, i.e. no worse than today).
- `pull_request_target` runs base-branch code only; no PR-head code executes, so the token is never exposed to untrusted code.

### DIFF
--- a/.github/workflows/data-plane-review-assignment.yaml
+++ b/.github/workflows/data-plane-review-assignment.yaml
@@ -45,6 +45,12 @@ jobs:
     # ubuntu-slim: REST-only job (no build) that runs on every Summarize Checks completion,
     # matching the other label-reaction workflows (arm-auto-signoff-status, update-labels).
     runs-on: ubuntu-slim
+    # Gate the App-token step on the secret being present: the `env` context is available in a
+    # step `if`, but the `secrets` context is not. Until AZURE_SDK_AUTOMATION_APP_ID is
+    # provisioned, APP_ID is empty, the token step is skipped, and the job falls back to the
+    # default GITHUB_TOKEN (team resolution 422s on that path, i.e. no worse than today).
+    env:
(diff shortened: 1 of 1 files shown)


samvaity:
Close wrt https://github.com/Azure/azure-rest-api-specs/pull/46608

PR #46193, opened 2026-09-09
Add IdentityBindingProfile resource to Fleet 2026-11-02-preview

nibooy:
## Summary

Adds the **multi-cluster identity binding** resource to the Fleet RP in the `2026-11-02-preview` API version. This mirrors the per-cluster `Microsoft.ContainerService/managedClusters/identityBindings` resource so a single binding can be projected onto a selected set of fleet members instead of being configured cluster-by-cluster.

## What's added

New proxy resource `fleets/{fleetName}/identityBindingProfiles/{identityBindingProfileName}`:

| Operation | Route |
|---|---|
| `GET` (list) | `.../fleets/{fleetName}/identityBindingProfiles` |
| `GET` / `PUT` / `DELETE` | `.../identityBindingProfiles/{identityBindingProfileName}` |
| `POST` | `.../identityBindingProfiles/{identityBindingProfileName}/apply` |

The `apply` action follows the existing **Configure → Plan → Apply** pattern already used by `clusterMeshProfiles`. `Plan` is the CLI-side what-if on `apply` — it is deliberately not a separate API route, matching `clusterMeshProfiles`, which also exposes only `apply`.

## Parity with the per-cluster resource

Property names, patterns, constraints, and documentation strings are copied **verbatim** from `managedClusters/identityBindings` so validation behaves identically across the two APIs.

The models copied from AKS keep their upstream names rather than taking a Fleet-specific prefix, so `AllowedSubject`, `LabelSelector`, `LabelSelectorRequirement` and `Operator` emit definitions byte-identical to `managedClusters.json`. The fleet-specific models are named after the resource: `IdentityBindingProfile`, `IdentityBindingProfileProperties`, `IdentityBindingProfileStatus`, `IdentityBindingManagedIdentityProfile`, `IdentityBindingOidcIssuerProfile`.

`IdentityBindingManagedIdentityProfile` is schema-identical to the AKS `IdentityBindingManagedIdentityProfile`:

- `resourceId` — create-once (`x-ms-mutability: [read, create]`), required
- `objectId` / `clientId` / `tenantId` — `readOnly`, resolved by the RP from the user-assigned identity

The three GUIDs are read-only because they are derived from the referenced UAMI, and the OIDC issuer URL is built from `tenantId` + `clientId` — so they must be stable for the lifetime of the binding.

## Fleet-specific additions

- `memberSelector` — selects which fleet members the binding is projected onto
- `status` — surfaces `lastOperationId` / `lastOperationError` from the most recent apply

## Notes

- Scope is **fleet spec only**. The AKS-side `managedBy` change is intentionally excluded — it needs its own AKS release train.
- `preview/2026-11-02-preview/fleets.json` and the mirrored `examples/` files are emitter output from `tsp compile`, not hand-edited.
- Examples use `rg1` for every request parameter and `identitybinding1` as the resource name. This also corrects `Location` / `Azure-AsyncOperation` headers that previously pointed at `myResourceGroup`, a resource group that appeared nowhere in the request.
- The `Operator` description intentionally diverges from AKS. AKS documents "Valid operators are In and NotIn", but the RP accepts four: `trust-domain/ib-rp/internal/api/validation/allowed_subjects.go` rejects anything outside `In`/`NotIn`/`Exists`/`DoesNotExist`, and the generated `ib-apis` SDK constants expose all four. The enum values match AKS exactly; only the prose differs. The stale AKS text is worth a separate fix in `aks/LoadBalancer.tsp`.
- Compiles clean with `--warn-as-error`; `tsp format` and `prettier --check` both pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/identitybinding.tsp
```diff
@@ -0,0 +1,233 @@
...
+  @maxLength(36)
+  @minLength(36)
+  @pattern("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
+  tenantId?: string;
+}
+
+@doc("IdentityBinding OIDC issuer profile.")
+@added(Versions.v2026_11_01_preview)
+model FleetIdentityBindingOidcIssuerProfile {
+  #suppress "@azure-tools/typespec-client-generator-core/csharp-no-url-suffix" "Name kept for backward compatibility"
```
weng271190436: maybe just call this oidcIssuerUri?

It is a new property in a new file. Might be hard to justify that this is needed for backward compatibility

weng271190436 (approved):

specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/identitybinding.tsp
nibooy: We ideally want to keep the fields identical with the `managedCluster/identityBinding` API since this field is populated from the single-cluster scope resource and prevent confusion for customers.
https://github.com/Azure/azure-rest-api-specs/blob/22f6db4627b98c6fc2c00d22256ff79d6da77c65/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/IdentityBinding.tsp#L236

PR #46199, opened 2026-09-09
Add ACS tool catalog contract

lycroy-work:
## Summary
- add an extensible ACS `tools` catalog keyed by canonical tool name
- type canonical `security_labels` as `string[]` and `clearance` as `string`
- model `tool_name_from` only on pre/post tool-call intervention points, with documented selector constants
- add PUT, GET, conditional-update, post-tool-call, and no-tool-catalog examples

## Contract boundary
The `tools` map key is the canonical lookup key. An optional tool-entry `id` is host metadata and does not override the map key. Tool entries remain extensible through `additionalProperties` so custom policy metadata round-trips unchanged.

`security_labels` uses the canonical ACS array form and `clearance` uses the canonical scalar form. `id`, `type`, and `clearance` are non-empty when present; `tool_name_from` remains an extensible string selector whose empty-value rejection is owned by PMS write admission. `description` may be empty, and `security_labels` may be an empty array.

`tool_name_from` is exposed only on `pre_tool_call` and `post_tool_call`. The current Unified Moderate profile documents `$snap.tool_call.name` and its `$.tool_call.name` alias. A tools catalog may be omitted or empty when no selector is configured. A configured selector requires a non-empty catalog, and an unknown selected map key fails closed.

The GET example now reflects current RP behavior: RP returns the complete PMS-owned ACS object, including `metadata`, `tools`, `annotators`, and custom extension fields.

## Validation
- TypeSpec 1.14 compile with warnings as errors
- TypeSpec and JSON format checks
- OAV semantic validation
- OAV validation of every referenced example
- four source/generated example pairs are byte-identical
- generated-version isolation: only `2026-09-15-preview` changed
- explicit generated-contract assertions for label types, applicable inline non-empty constraints, selector values, and intervention-point specialization

## Follow-up gates
- generated SDK validation for C#, Java, JavaScript/TypeScript, Python, Go, and Rust in PR CI
- align PMS/Frontdoor admission with the canonical public label forms before Public activation; the current runtime branch temporarily accepts scalar-or-array forms for both fields
- complete the separately reviewed PMS/Frontdoor streaming runtime gates before enabling public catalog authoring

No RP semantic-validation or runtime-activation change is included in this PR.

### DIFF
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/GetRaiPolicyAcs.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/GetRaiPolicyAcs.json
@@ -27,6 +27,10 @@
               "input-guard": {
                 "type": "rego",
                 "query": "data.input_guard.verdict"
+              },
+              "tool-guard": {
+                "type": "rego",
+                "query": "data.tool_guard.verdict"
               }
             },
             "intervention_points": {
@@ -44,12 +48,47 @@
                     ]
                   }
                 }
+              },
+              "pre_tool_call": {
+                "policy_target": "$snap.tool_call.args",
+                "policy_target_kind": "tool_args",
+                "tool_name_from": "$snap.tool_call.name",
+                "policy": {
+                  "id": "tool-guard"
+                }
+              }
+            },
+            "tools": {
+              "web_search": {
+                "id": "web_search",
+                "type": "retrieval",
+                "description": "Search approved public documentation",
+                "security_labels": [
+                  "network_egress",
+                  "untrusted_content"
+                ],
+                "clearance": "public",
+                "allowed_domains": [
+                  "learn.microsoft.com"
+                ]
+              },
+              "wire_transfer": {
+                "id": "wire_transfer",
+                "type": "financial_action",
+                "security_labels": "financial_write",
+                "clearance": [
+                  "banking",
+                  "payments"
+                ]
               }
             }
           },
           "acsRegos": [
             {
               "regoName": "input-guard"
+            },
+            {
+              "regoName": "tool-guard"
             }
           ],
           "customBlocklists": [
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/GetRaiPolicyAcs.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/GetRaiPolicyAcs.json
@@ -23,6 +23,9 @@
           "format": "ACS",
           "acs": {
             "agent_control_specification_version": "0.4.0-alpha.1",
+            "metadata": {
+              "name": "agent-guard"
+            },
             "policies": {
               "input-guard": {
                 "type": "rego",
@@ -56,6 +59,14 @@
                 "policy": {
                   "id": "tool-guard"
                 }
+              },
+              "post_tool_call": {
+                "policy_target": "$snap.tool_result.value",
+                "policy_target_kind": "tool_result",
+                "tool_name_from": "$snap.tool_call.name",
+                "policy": {
+                  "id": "tool-guard"
+                }
               }
             },
             "tools": {
@@ -75,13 +86,13 @@
               "wire_transfer": {
                 "id": "wire_transfer",
                 "type": "financial_action",
-                "security_labels": "financial_write",
-                "clearance": [
-                  "banking",
-                  "payments"
-                ]
+                "security_labels": [
+                  "financial_write"
+                ],
+                "clearance": "confidential"
               }
-            }
+            },
+            "annotators": {}
           },
           "acsRegos": [
             {
--- a/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/PutRaiPolicyAcs.json
+++ b/specification/cognitiveservices/CognitiveServices.Management/examples/2026-09-15-preview/PutRaiPolicyAcs.json
@@ -17,6 +17,10 @@
             "input-guard": {
               "type": "rego",
               "query": "data.input_guard.verdict"
+            },
+            "tool-guard": {
+              "type": "rego",
+              "query": "data.tool_guard.verdict"
             }
           },
           "intervention_points": {
@@ -34,14 +38,48 @@
                   ]
                 }
               }
+            },
+            "pre_tool_call": {
+              "policy_target": "$snap.tool_call.args",
+              "policy_target_kind": "tool_args",
+              "tool_name_from": "$snap.tool_call.name",
+              "policy": {
+                "id": "tool-guard"
+              }
             }
           },
-          "tools": {},
+          "tools": {
+            "web_search": {
+              "id": "web_search",
+              "type": "retrieval",
+              "description": "Search approved public documentation",
+              "security_labels": [
+                "network_egress",
+                "untrusted_content"
+              ],
+              "clearance": "public",
+              "allowed_domains": [
+                "learn.microsoft.com"
+              ]
+            },
+            "wire_transfer": {
+              "id": "wire_transfer",
+              "type": "financial_action",
+              "security_labels": "financial_write",
+              "clearance": [
+                "banking",
+                "payments"
+              ]
+            }
+          },
           "annotators": {}
         },
         "acsRegos": [
           {
             "regoName": "input-guard"
+          },
+          {
+            "regoName": "tool-guard"
           }
         ],
         "customBlocklists": [
@@ -87,6 +125,10 @@
               "input-guard": {
                 "type": "rego",
                 "query": "data.input_guard.verdict"
+              },
+              "tool-guard": {
+                "type": "rego",
+                "query": "data.tool_guard.verdict"
               }
             },
             "intervention_points": {
@@ -104,14 +146,48 @@
                     ]
                   }
                 }
+              },
+              "pre_tool_call": {
+                "policy_target": "$snap.tool_call.args",
+                "policy_target_kind": "tool_args",
+                "tool_name_from": "$snap.tool_call.name",
+                "policy": {
+                  "id": "tool-guard"
+                }
               }
             },
-            "tools": {},
+            "tools": {
+              "web_search": {
+                "id": "web_search",
+                "type": "retrieval",
+                "description": "Search approved public documentation",
+                "security_labels": [
+                  "network_egress",
+                  "untrusted_content"
+                ],
+                "clearance": "public",
+                "allowed_domains": [
+                  "learn.microsoft.com"
+                ]
+              },
+              "wire_transfer": {
+                "id": "wire_transfer",
+                "type": "financial_action",
+                "security_labels": "financial_write",
+                "clearance": [
+                  "banking",
+                  "payments"
+                ]
+              }
+            },
             "annotators": {}
           },
           "acsRegos": [
             {
               "regoName": "input-guard"
+            },
+            {
+              "regoName": "tool-guard"
             }
           ],
           "customBlocklists": [
@@ -153,6 +229,10 @@
               "input-guard": {
                 "type": "rego",
                 "query": "data.input_guard.verdict"
+              },
+              "tool-guard": {
+                "type": "rego",
+                "query": "data.tool_guard.verdict"
               }
             },
             "intervention_points": {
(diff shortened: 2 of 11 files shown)


huliang-microsoft (approved):

fmabroukmsft:
## Review notes

### CI triage

Both red checks appear to be unrelated to this PR's content:

- **Swagger Avocado** — the run reports 56 errors, none under `specification/cognitiveservices`. They're
  repo-wide (`ai-foundry` MISSING_README ×30, `redhatopenshift` ×9, plus `vi`, `security`, `network`,
  `compute`). The job took 38m53s here vs 2m15s on #45556, which is consistent with Avocado falling back
  to a full-repo scan because the base ref is an unmerged stacked branch. #45556 passes Avocado.
- **Breaking Change (Cross-Version)** — the single Error-level finding is `ChangedParameterOrder` for the
  `raiPolicy` body parameter (index 3 → 5) against stable `2026-07-01`. None of the 65 findings reference
  `tools`, `tool_name_from`, `RaiAcsToolDefinition`, or `RaiAcsEmptyObject`. Same failure is present on
  #45556, so it needs sign-off there rather than here.

Remaining blockers are process, not content: `license/cla` is still unsigned, `SuppressionReviewRequired`
needs an ARM reviewer, and the stack (#46199 → #45556 → `CogSvc-2026-09-01-typespec`) has to merge
bottom-up — a squash or rebase of #45556 will invalidate this diff.

Validation claims in the description hold up on inspection: the three source/generated example pairs are
byte-identical at `050e95b`, request/response `acs` blocks echo exactly in both Put and Update (200 and
201), the default tag stays `package-2026-09-01`, and `RaiAcsEmptyObject` is still referenced by
`annotations`/`annotators`, so it isn't orphaned.

### Questions on the contract

**1. Precedence between the map key and `id`.** Every example uses `"web_search": { "id": "web_search" }`,
but nothing in the schema requires them to agree — `"web_search": { "id": "wire_transfer" }` is valid.
Which one does catalog lookup key off? Since `id` is optional and free-form, deferring this to PMS leaves
the resolution rule unverifiable from the spec. A sentence in the `@doc`, or dropping the redundant `id`,
would close it.

**2. `security_labels` / `clearance` as `unknown`.** The suppression justification explains why a
`string | string[]` union can't be emitted to OpenAPI v2, which is fair, but `unknown` is a much wider
fallback than the documented shape. The practical cost is in the generated SDKs — these surface as
`object` / `any` / `BinaryData` / `interface{}` and as untyped fields in the reference docs, which the
description's own follow-up gate flags as unverified. Given the model already carries
`additionalProperties: {}` for extensibility, typing the named fields as `string[]` and normalizing a
bare string in the RP would keep the extensibility story intact while preserving a typed SDK surface.
Preview makes this reversible, but it's the shape that tends to carry into GA.

### Smaller points

- `toolNameFrom` is a bare `string`, while `policyTarget` and `policyTargetKind` next to it are extensible
  unions. `$snap.tool_call.name` is the same class of token as the values in `RaiAcsPolicyTarget`, so an
  extensible union would be equally open and would generate constants and docs.
- `toolNameFrom` sits on `RaiAcsInterventionPoint`, which is shared by all four points, so it's also valid
  (and inert) on `input`, `output`, and `post_tool_call`. Worth noting in the `@doc` if it only applies to
  tool-call points.
- The new string fields (`id`, `type`, `description`, `toolNameFrom`) omit `@minLength(1)`, where the
  surrounding ACS models (`RaiAcsPolicyBinding.id`/`.query`, `RaiAcsRegoPolicyDefinition.query`,
  `RaiAcsHarmConfiguration.harmConfigId`) all use it. Consistency only.
- "add a non-empty ... tools catalog" in the description reads as a new constraint; the schema still allows
  `"tools": {}`. Removing `"tools": {}` from the Put example also drops the only coverage of that shape.
- Pre-existing from #45556, not introduced here: the GET example omits `metadata` and `annotators` that the
  Put example's own 200/201 responses return for the same resource, and `post_tool_call` isn't exercised in
  any example.

lycroy-work:
Thanks, these were good catches. Addressed in `9841bfe598`:

- Documented that the `tools` map key controls lookup; optional entry `id` is metadata only.
- Replaced the untyped named fields with canonical ACS shapes: `security_labels: string[]` and `clearance: string`. Custom fields remain supported through `additionalProperties`; RP does not normalize values.
- Added non-empty constraints to `id`, `type`, `clearance`, and `tool_name_from`; `description` remains unconstrained and an empty labels array remains valid.
- Added an extensible selector union for `$snap.tool_call.name` and `$.tool_call.name`.
- Restricted `tool_name_from` structurally to `pre_tool_call` and `post_tool_call` models.
- Clarified that `tools` may be omitted/empty without a selector, and added a focused no-tools example.
- Added `post_tool_call` to the tool examples.
- Added `metadata` and `annotators` to GET to match RP's complete PMS-owned ACS round trip.

Local TypeSpec compile/format, OAV semantic/example validation, source/generated parity, and generated-version isolation all pass. New CI is running. The planned PMS/Frontdoor admission currently accepts broader scalar-or-array label forms; aligning that runtime boundary to these canonical public shapes remains a pre-activation gate.

lycroy-work:
> ## Review notes
> ### CI triage
> Both red checks appear to be unrelated to this PR's content:
> 
> * **Swagger Avocado** — the run reports 56 errors, none under `specification/cognitiveservices`. They're
>   repo-wide (`ai-foundry` MISSING_README ×30, `redhatopenshift` ×9, plus `vi`, `security`, `network`,
>   `compute`). The job took 38m53s here vs 2m15s on [Add CognitiveServices preview API version 2026-09-15-preview #45556](https://github.com/Azure/azure-rest-api-specs/pull/45556), which is consistent with Avocado falling back
>   to a full-repo scan because the base ref is an unmerged stacked branch. [Add CognitiveServices preview API version 2026-09-15-preview #45556](https://github.com/Azure/azure-rest-api-specs/pull/45556) passes Avocado.
> * **Breaking Change (Cross-Version)** — the single Error-level finding is `ChangedParameterOrder` for the
>   `raiPolicy` body parameter (index 3 → 5) against stable `2026-07-01`. None of the 65 findings reference
>   `tools`, `tool_name_from`, `RaiAcsToolDefinition`, or `RaiAcsEmptyObject`. Same failure is present on
>   [Add CognitiveServices preview API version 2026-09-15-preview #45556](https://github.com/Azure/azure-rest-api-specs/pull/45556), so it needs sign-off there rather than here.
> 
> Remaining blockers are process, not content: `license/cla` is still unsigned, `SuppressionReviewRequired` needs an ARM reviewer, and the stack (#46199 → #45556 → `CogSvc-2026-09-01-typespec`) has to merge bottom-up — a squash or rebase of #45556 will invalidate this diff.
> 
> Validation claims in the description hold up on inspection: the three source/generated example pairs are byte-identical at `050e95b`, request/response `acs` blocks echo exactly in both Put and Update (200 and 201), the default tag stays `package-2026-09-01`, and `RaiAcsEmptyObject` is still referenced by `annotations`/`annotators`, so it isn't orphaned.
> 
> ### Questions on the contract
> **1. Precedence between the map key and `id`.** Every example uses `"web_search": { "id": "web_search" }`, but nothing in the schema requires them to agree — `"web_search": { "id": "wire_transfer" }` is valid. Which one does catalog lookup key off? Since `id` is optional and free-form, deferring this to PMS leaves the resolution rule unverifiable from the spec. A sentence in the `@doc`, or dropping the redundant `id`, would close it.
> 
> **2. `security_labels` / `clearance` as `unknown`.** The suppression justification explains why a `string | string[]` union can't be emitted to OpenAPI v2, which is fair, but `unknown` is a much wider fallback than the documented shape. The practical cost is in the generated SDKs — these surface as `object` / `any` / `BinaryData` / `interface{}` and as untyped fields in the reference docs, which the description's own follow-up gate flags as unverified. Given the model already carries `additionalProperties: {}` for extensibility, typing the named fields as `string[]` and normalizing a bare string in the RP would keep the extensibility story intact while preserving a typed SDK surface. Preview makes this reversible, but it's the shape that tends to carry into GA.
> 
> ### Smaller points
> * `toolNameFrom` is a bare `string`, while `policyTarget` and `policyTargetKind` next to it are extensible
>   unions. `$snap.tool_call.name` is the same class of token as the values in `RaiAcsPolicyTarget`, so an
>   extensible union would be equally open and would generate constants and docs.
> * `toolNameFrom` sits on `RaiAcsInterventionPoint`, which is shared by all four points, so it's also valid
>   (and inert) on `input`, `output`, and `post_tool_call`. Worth noting in the `@doc` if it only applies to
>   tool-call points.
> * The new string fields (`id`, `type`, `description`, `toolNameFrom`) omit `@minLength(1)`, where the
>   surrounding ACS models (`RaiAcsPolicyBinding.id`/`.query`, `RaiAcsRegoPolicyDefinition.query`,
>   `RaiAcsHarmConfiguration.harmConfigId`) all use it. Consistency only.
> * "add a non-empty ... tools catalog" in the description reads as a new constraint; the schema still allows
>   `"tools": {}`. Removing `"tools": {}` from the Put example also drops the only coverage of that shape.
> * Pre-existing from [Add CognitiveServices preview API version 2026-09-15-preview #45556](https://github.com/Azure/azure-rest-api-specs/pull/45556), not introduced here: the GET example omits `metadata` and `annotators` that the
>   Put example's own 200/201 responses return for the same resource, and `post_tool_call` isn't exercised in
>   any example.

Pushed fixes for critical findings in the new commit

fmabroukmsft:
Two follow-ups on `b7cb304`.

### 1. `minLength` on `tool_name_from` is not applied

The emitted schema is:

```json
"tool_name_from": {
  "$ref": "#/definitions/RaiAcsToolNameSelector",
  "description": "Selects the canonical tool catalog key from the raw Agent Hooks snapshot.",
  "minLength": 1,
  "x-ms-client-name": "toolNameFrom"
}
```

OpenAPI 2.0 resolves against JSON Schema draft-4, where sibling keywords of `$ref` are ignored, so the
`minLength` constraint has no effect here. `RaiAcsToolNameSelector` carries no `minLength` of its own and is
`modelAsString: true`, so `""` validates against the property. The other three fields (`id`, `type`,
`clearance`) are inline strings and do enforce correctly — this affects only the one property that changed
from a bare `string` to a union.

No linter catches this, since LintDiff doesn't flag `$ref`-sibling shadowing. Either move `@minLength(1)`
onto `RaiAcsToolNameSelector` itself, or drop the decorator and adjust the description, so the stated
"non-empty when present" guarantee matches what the contract actually enforces.

### 2. Direction of the spec/runtime divergence

Worth making explicit, because typing the labels inverted which side is permissive. Previously the schema
was `unknown` and the runtime accepted scalar-or-array, so the two agreed. Now the schema is `string[]` /
`string` while, per the follow-up gates, "the current runtime branch temporarily accepts scalar-or-array
forms for both fields."

The asymmetry matters on read rather than write: GET returns whatever PMS persisted, so a single entry
stored in the looser form produces a response that violates its own declared schema, which surfaces as a
deserialization failure in the strongly-typed SDKs (C#, Java, Go) rather than as a validation error the
caller can handle. Pinning the alignment to **write admission** rather than to activation would close this
— once no non-conforming value can be persisted, the GET contract holds regardless of activation timing.

Related, and probably just a documentation question: the original description stated that upstream ACS uses
`string | string[]` for both fields, and they have now been narrowed in opposite directions — labels to
array-only, clearance to scalar-only. The `AvoidAdditionalProperties` suppression in `readme.md` still cites
"ACS manifest portability and round-tripping" as its justification, but a manifest carrying
`clearance: ["banking", "payments"]` can no longer round-trip through this contract. If the scalar form is
genuinely canonical then this is fine and only the earlier claim was imprecise; confirming that would settle
it.

lycroy-work:
Thanks, agreed on both follow-ups.

1. Fixed in `3ec62474ec2`. I removed the ineffective property-level `@minLength(1)` from `tool_name_from` and clarified that `RaiAcsToolNameSelector` is an extensible string selector: the two listed snapshot paths are the values currently recognized by the Unified Moderate AACS host profile, while other strings remain allowed for forward compatibility. The emitted OpenAPI property no longer carries an ignored `minLength` sibling; `modelAsString: true` and both documented selector constants are preserved.

2. Agreed on the read-contract risk. I will fix this on the PMS side by enforcing the canonical public shapes during policy write admission, before persistence: `security_labels` must be an array of strings and `clearance` must be a scalar string. PMS will reject the broader scalar/array alternatives for this public profile. There are no existing nonconforming policies to migrate. This admission check will land before enabling the September public profile, so RP's pass-through GET remains schema-conformant without read-time normalization. The earlier scalar-or-array statement described temporary runtime parser compatibility, not the canonical alpha.3 public contract.

PR #46213, opened 2026-09-09
initial .net client.tsp updates

jorgerangel-msft:
# Choose a PR Template

### DIFF
--- a/specification/app/data-plane/ContainerApps/client.tsp
+++ b/specification/app/data-plane/ContainerApps/client.tsp
@@ -2,86 +2,118 @@ import "@azure-tools/typespec-client-generator-core";
 import "./main.tsp";
 
 using Azure.ClientGenerator.Core;
-using ContainerAppsSandbox;
 
-namespace ContainerAppsSandbox;
-
-// Client-surface customizations. Every rule here is scoped to "csharp", so none of them affect the
-// wire contract or the emitted OpenAPI under stable/ and preview/.
-//
-// Only types the emitter actually materializes can be renamed. HTTP envelope models ending in
-// Result are unwrapped into Response<T> / Response and never become C# classes, so @@clientName
-// rules on them are ineffective.
-
-// The service namespace is ContainerAppsSandbox, so the emitter derives the root client name
-// (ContainerAppsSandboxClient) on its own - no override needed for it.
-// The sub-client accessor preserves an existing Client suffix, yielding GetSandboxesClient().
-// The operationId in the OpenAPI stays AzureSandboxes_*.
-@@clientName(AzureSandboxes, "SandboxesClient", "csharp");
-
-// Qualify generic model names to avoid collisions with other libraries.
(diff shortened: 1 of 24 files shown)


specification/app/data-plane/ContainerApps/tspconfig.yaml
```diff
@@ -1,6 +1,6 @@
 parameters:
   "service-dir":
-    default: "sdk/containerappssandbox"
+    default: "sdk/containerapps"
```
jorgerangel-msft, line 3: `containerapps` is the existing service directory https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/containerapps.

RichardChen820: Make sense

specification/app/data-plane/ContainerApps/client.tsp
```diff
@@ -15,9 +15,18 @@ namespace ContainerAppsSandbox;
...
+@@clientName(AzureSandboxes, "SandboxesClient", "csharp");
+
+// Qualify generic model names to avoid collisions with other libraries.
+@@clientName(Connection, "SandboxConnection", "csharp");
+@@clientName(Secret, "SandboxSecret", "csharp");
+@@clientName(Sandbox, "ContainerAppsSandbox", "csharp");
+@@clientName(Snapshot, "SandboxSnapshot", "csharp");
+
+// SandboxVolume already describes a volume mount; this model is the group-scoped resource.
+@@clientName(Volume, "SandboxGroupVolume", "csharp");
```
RichardChen820: @jorgerangel-msft 
The namespace is able to distinguish them from others, is it? Can we just keep the names as they are?

jorgerangel-msft: In general, we want to avoid short names like `Volume`. We have a .NET analyzer that will flag that and prevent the library from building:

``Type name 'Volume' is too generic and has high chance of collision with BCL types or types from other libraries. Consider renaming to: 'SandboxVolume', 'ContainerAppsVolume', 'VolumeOptions'``

We just need to avoid the short name(s) for types, even if they are in a nested namespace.

specification/app/data-plane/ContainerApps/client.tsp
```diff
@@ -2,29 +2,580 @@ import "@azure-tools/typespec-client-generator-core";
...
+namespace ContainerAppsSandbox {
+  // Client hierarchy
+  // These namespaces declare no REST operations. C#-only placement decorators below
+  // populate the clients without changing the service interfaces or other languages.
+  /** Operations within one sandbox group. */
+  namespace Group {
+    /** Operations on one sandbox within the group. */
+    namespace Resource {
+      namespace Files {
+
```
RichardChen820, line 15: Useless blank line? Or it's required by format lint rule?

jorgerangel-msft: the format lint rule requires this

PR #46243, opened 2026-09-10
[DRAFT][2026-09-01] ContainerService: add LocalDNS failfastAllUnhealthyUpstreams and healthCheck

saewoni:
> **⚠️ DRAFT — early review only.** The target API version is **2026-09-01**, but the
> `dev-containerservice-Microsoft.ContainerService-2026-09` branch has not been cut yet
> (per the AKS API release train, swagger dev branches are created after the 2026-08 API
> release completes). This PR is temporarily based on the **2026-07** dev branch so the
> change can be reviewed early. It will be **retargeted to the 2026-09 dev branch** once it
> exists, at which point the `@added(Versions.v2026_09_01)` gate will compile.
> Please review the **model change only**; CI will not be green against this temporary base
> because `Versions.v2026_09_01` does not exist on the 2026-07 branch.

## Summary

Adds the CoreDNS forward-plugin health knobs to the AKS `LocalDNSOverride` model for the **2026-09-01** API version:

- `failfastAllUnhealthyUpstreams?: boolean` — when true, localDNS returns SERVFAIL immediately once all upstream DNS servers are unhealthy, instead of forwarding the query and waiting for a timeout.
- `healthCheck?: LocalDNSHealthCheck` — active upstream health checking:
  - `duration?: string` — health check interval as a duration string (e.g. `"500ms"`, `"1s"`).
  - `noRec?: boolean` — sets the RecursionDesired bit to false on health check queries.

These let customers configure the forward-plugin health behavior directly through the localDNS API payload, instead of enabling them out-of-band via a DaemonSet.

## Versioning

All new members are gated with `@added(Versions.v2026_09_01)`, so they appear only in `2026-09-01` and later. Earlier API versions are unchanged — no retroactive enablement.

## Related

- **aks-rp** datamodel + converters (ADO PR 17105774): adds the same fields to the `v20260901` frontend datamodel and wires them to the existing HCP proto fields and the agentbaker/nodeprovisioner paths.
- **agentbaker** (follow-up): emit `health_check` and `failfast_all_unhealthy_upstreams` inside the `forward` block of the localdns corefile templates.

## Remaining before this can merge

- [ ] Retarget onto `dev-containerservice-Microsoft.ContainerService-2026-09` (once cut).
- [ ] `tsp compile` + regenerate swagger on that base.
- [ ] Add/update a localdns example under `aks/examples/2026-09-01/` covering the new fields.


saewoni:
Closing — branch was based on fork main, producing a huge diff against the 2026-07 base. Recreating with the branch based directly on the target dev branch so the diff is limited to the LocalDNS model change.

